PR #369, opened 2017-06-15
Rename django folder

jmeakin:


jmeakin:
Code climate score went down as pep8 checks were enabled

Issue #435, opened 2018-03-20
Rename comic app to grandchallenge


jmsmkn:

jmsmkn:
Move manage.py up 1 level and namespace the application properly.

jmsmkn:
See the approach here: https://github.com/rtfd/readthedocs.org/blob/master/readthedocs/builds/models.py

jmsmkn:
Fixed

PR #447, opened 2018-03-22
Creates the grandchallenge namespace

jmsmkn:
As part of cleaning up the code all of the local apps are moved to a new namespace, and several instances of `comic` are removed.


jmsmkn:
Fixes #435

PR #552, opened 2018-10-05
WIP: Upgrade bootstrap

### LINKED ISSUE #506: Refresh the UI
We will have to decide what to do with the existing css options, especially for the heavily modified challenges such as tadpole. 

Suggestions:
- [x] Upgrade bootstrap3 to bootstrap4
- [x] Sidebar for challenge pages rather than 2nd navbar

jmsmkn:
Closes #506


app/grandchallenge/core/static/css/core.css
```diff
@@ -128,7 +128,7 @@ div.sidebar h1 {
 /* Little more spacing for checkbox */
 
 input[type=checkbox] {
-    margin-top: .3rem !important;
+    margin-right: .3rem !important;
```
jmsmkn: This also affects the checkboxes on forms, you should check those (log in as demo and then so to challenge settings)

app/grandchallenge/core/templates/site.html
```diff
@@ -28,8 +28,8 @@
 {% block jumbotron %}
     {% if site.banner %}
         <div id="headerTitle"
-             class="container">
-            <div class="row mt-3">
+             class="container mt-3">
+            <div class="row p-3">
```
jmsmkn: Check that the columns do not jump around when going between pages with banners, pages without banners, and full width pages (eg, the all challenges page)

HansPinckaers: What do you mean with jump around?

jmsmkn: The top left corner of the banner and the full width card should be at the same position on the screen, so that when you're clicking through pages (without scrolling in between) it doesn't look like the cards are moving left/right/up/down.

PR #576, opened 2018-10-30
Update dependencies and hold back django-celery-beat

jmsmkn:
django-celery-beat has a known bug so hold this package back. See celery/django-celery-beat#175


jmsmkn:
Builds are failing as docker hub is experiencing an outage: https://status.docker.com/

PR #584, opened 2018-10-31
Replace CKEditor with Summernote

jmsmkn:
CKEditor gives a bunch of problems with image uploads so switch to summernote which is built for bootstrap. 

SummernoteInlineWidget does not work with the fixed footer, so use the iframe version for now.


app/grandchallenge/core/templates/base.html
```diff
@@ -38,6 +38,10 @@
         {% block jquery_js %}
             <script type="text/javascript"
                     src="{% static "jquery-3.2.1/js/jquery-3.2.1.min.js" %}"></script>
+            {# jquery ui must be before bootstrap #}
```
jmsmkn:
```suggestion
            {# jquery ui must be before bootstrap, but is required by blueimp-file-upload #}
```

Issue #628, opened 2018-11-27
Travis build error - pipenv --dev flag


rzinkstok:
Travis build fails on building containers due to env variable collision: the PIPENV_DEV build-arg conflicts with an existing env var used by pipenv of the same name, but it is not compatible with a value of "--dev". Changing the build flag name to PIPENV_DEV_FLAG fixes the issue.

PR #629, opened 2018-11-27
Fix make build issue with pipenv env var collision; adds eyra deploy provider for travis.

rzinkstok:
Travis build fails on building containers due to env variable collision: the PIPENV_DEV build-arg conflicts with an existing env var used by pipenv of the same name, but it is not compatible with a value of "--dev". Changing the build flag name to PIPENV_DEV_FLAG fixes the issue.

Closes https://github.com/comic/grand-challenge.org/issues/628

We also added a new travis deploy provider for the eyra-master branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,11 @@ after_success:
   - codecov
 
 deploy:
-  provider: script
-  script: bash -c "docker login -u grandchallenge -p "$DOCKER_PASSWORD" && make push"
-  on:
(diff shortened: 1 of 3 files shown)


jmsmkn:
Looks good, we should still add a branch tag, probably use `TRAVIS_BRANCH`. Would need to integrate with the existing cycle_docker_compose and docker-compose.yml though.

Issue #630, opened 2018-11-27
Local development with Django devserver in PyCharm


rzinkstok:
In order to work with the PyCharm debugger etc it would be nice to be able to run the django development server in PyCharm itself instead of using the full docker-compose setup locally. It also makes for a quicker reload of the code in the server. We propose the following:
- creation of a .env.dev file containing all the environment variables needed for local development
- adding env vars for POSTGRES_HOST and CELERY_TASK_ALWAYS_EAGER
- adding a docker-compose_dev.yml file that contains only the database service; the postgres data folder is a folder in the repo (ignored in .gitignore) which is loaded as a volume to the DB container; the postgres port is shared with the host 
- setting CELERY_TASK_ALWAYS_EAGER removes the need for the celery workers and redis; caching is not needed either
- adding documentation to instruct people how to setup the local dev environment in PyCharm

Of course it is still possible to run the full docker-compose system, and that would still be needed to run all the tests.

rzinkstok:
You can see our changes in the eyra-dev branch (https://github.com/comic/grand-challenge.org/tree/eyra-dev)

jmsmkn:
Ok, all of the test that rely on celery are broken on that branch right now, my guess is that it's because `CELERY_TASK_ALWAYS_EAGER` should be None by default. Setting `CELERY_TASK_ALWAYS_EAGER` is not sufficient if you also want to not run redis and the workers, you also need to set the propagation, broker url and backend, eg:

```
        settings.task_eager_propagates = (True,)
        settings.task_always_eager = (True,)
        settings.broker_url = ("memory://",)
        settings.backend = "memory"
```

Note that you then lose the integration with celery so any views that have long tasks associated with them will then fail. There is also coupling between nginx and django as some extra routing is done there. 

Why do you need to maintain an exhaustive list of default environment variables? Defaults are handled with `os.environment.get`, so only store the ones that you explicitly need to override for running in this mode and don't pass through your system environment vars.

rzinkstok:
[...]

PR #634, opened 2018-11-30
Local Django dev server

rzinkstok:
These changes make it possible to use a local Django dev server, in combination with the full docker-compose stack. This enables e.g. the use of the PyCharm debugger. 

See also https://github.com/comic/grand-challenge.org/issues/630


.env.local
```diff
@@ -0,0 +1,30 @@
+CELERY_TASK_ALWAYS_EAGER=True
+CELERY_TASK_EAGER_PROPAGATES=True
```
jmsmkn: CELERY_TASK_ALWAYS_EAGER and CELERY_TASK_EAGER_PROPAGATES are not loaded from environment variables in settings, remove them here

.env.local
```diff
@@ -0,0 +1,30 @@
...
+CELERY_BROKER_URL=memory://
+CELERY_BACKEND=memory
```
jmsmkn: I thought that you were using redis now?

.env.local
```diff
@@ -0,0 +1,30 @@
...
+
+DEBUG=True
```
jmsmkn: Can be removed

.env.local
```diff
@@ -0,0 +1,30 @@
...
+SECRET_KEY=DEV
+
+EMAIL_HOST=LOCALHOST
+EMAIL_HOST_USER=USER
+EMAIL_HOST_PASSWORD=PASSWORD
+EMAIL_PORT=25
+EMAIL_USE_TLS=False
+DEFAULT_FROM_EMAIL=dev@localhost
+SERVER_EMAIL=dev@localhost
```
jmsmkn: These are all unnecessary for a development environment as they are ignored with EMAIL_BACKEND = "django.core.mail.backends.dummy.EmailBackend"

.env.local
```diff
@@ -0,0 +1,30 @@
...
+
+SITE_ID=3
+
+MAIN_PROJECT_NAME=comic_local
+MEDIA_ROOT=/tmp/comic/media_root/
+
+MAIN_HOST_NAME=http://localhost:8000
```
jmsmkn: This does not need to be set unless you're using SUBDOMAIN_IS_PROJECTNAME

.env.local
```diff
@@ -0,0 +1,30 @@
...
+
+EMAIL_HOST=LOCALHOST
+EMAIL_HOST_USER=USER
+EMAIL_HOST_PASSWORD=PASSWORD
+EMAIL_PORT=25
+EMAIL_USE_TLS=False
+DEFAULT_FROM_EMAIL=dev@localhost
+SERVER_EMAIL=dev@localhost
+
+SITE_ID=3
```
jmsmkn: Do you create this elsewhere? Probably remove this as it will break the one created by initcomicdemo

.env.local
```diff
@@ -0,0 +1,30 @@
...
+
+MAIN_PROJECT_NAME=comic_local
```
jmsmkn: See above

.env.local
```diff
@@ -0,0 +1,30 @@
...
+
+MAIN_HOST_NAME=http://localhost:8000
+
+SESSION_COOKIE_DOMAIN=127.0.0.1
+SECURE_HSTS_SECONDS=3600
+
+POSTGRES_HOST=localhost
+POSTGRES_PASSWORD=secretpassword
+POSTGRES_USER=comic
+POSTGRES_DB=comic
```
jmsmkn: Missing newline

.env.local
```diff
@@ -0,0 +1,30 @@
...
+MEDIA_ROOT=/tmp/comic/media_root/
+
+MAIN_HOST_NAME=http://localhost:8000
+
+SESSION_COOKIE_DOMAIN=127.0.0.1
+SECURE_HSTS_SECONDS=3600
+
+POSTGRES_HOST=localhost
+POSTGRES_PASSWORD=secretpassword
+POSTGRES_USER=comic
```
jmsmkn: You could add these as defaults in the settings file itself

.gitignore
```diff
@@ -88,3 +88,6 @@ app/celerybeat-schedule
 
 # No coverage files
 .coverage
+
+# Local postgres data
+postgres_data/
```
jmsmkn: Missing newline

rzinkstok:
You're right James, a lot turned out to be unneeded. In the end, only three items were required: POSTGRES_HOST, CELERY_BROKER_URL and SESSION_COOKY_DOMAIN.

Issue #616, opened 2018-11-16
Length of titles for cards


bramvanginneken:
Currently the space for the name of challenges on the cards is very limited:

![image](https://user-images.githubusercontent.com/1818749/48629751-c4d0b080-e9ba-11e8-9db4-e13a2d0c8ce8.png)

Could we have the titles on two lines? There is enough space next to the image. There is now also no way to see what is behind the ... (ok I'm fine with that).

If we do not want to change it, also fine, then we'll just shorten all titles.

PR #653, opened 2018-12-11
Truncate title when more than 3 words, also add full title as tooltip

HansPinckaers:
I checked on the grand-challenge page and truncating at 3 words seems the best option.

Relevant issue #616

### DIFF
--- a/app/grandchallenge/challenges/templates/challenges/challenge_list.html
+++ b/app/grandchallenge/challenges/templates/challenges/challenge_list.html
@@ -250,7 +250,7 @@
                     {% for challenge in challenges %}
                         <div class="col-12 col-md-6 col-lg-4 mb-3 projectlink {{ challenge.filter_classes|join:" " }} {{ challenge.year }}">
                             <div class="card" style="height:400px">
-                                <a id="{{ challenge.short_name }}" title="{{ challenge.title }}"
+                                <a id="{{ challenge.short_name }}" title="{% firstof challenge.title challenge.short_name %}"
                                    href="{{ challenge.get_absolute_url }}"
                                    style="display:block;">
                                     <div class="card-header small p-0 d-flex flex-row flex-nowrap align-items-center">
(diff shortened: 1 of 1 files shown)


app/grandchallenge/challenges/templates/challenges/challenge_list.html
```diff
@@ -250,15 +250,15 @@ <h2>{{ year }}</h2>
                     {% for challenge in challenges %}
                         <div class="col-12 col-md-6 col-lg-4 mb-3 projectlink {{ challenge.filter_classes|join:" " }} {{ challenge.year }}">
                             <div class="card" style="height:400px">
-                                <a id="{{ challenge.short_name }}"
+                                <a id="{{ challenge.short_name }}" title="{{ challenge.title }}"
```
jmsmkn:
```suggestion
                                <a id="{{ challenge.short_name }}" title="{% firstof challenge.title challenge.short_name %}"
```
Not all challenges will have titles, but they will have short names

app/grandchallenge/challenges/templates/challenges/challenge_list.html
```diff
@@ -250,15 +250,15 @@ <h2>{{ year }}</h2>
...
                                    href="{{ challenge.get_absolute_url }}"
                                    style="display:block;">
                                     <div class="card-header small p-0 d-flex flex-row flex-nowrap align-items-center">
                                         <img class="thumbnail rounded-top-left flex-shrink-0"
                                              alt="{{ challenge.short_name }} logo"
                                              src="{{ challenge.thumb_image_url }}">
-                                        <h3 class="m-0 px-3 text-nowrap text-truncate flex-shrink-1">
-                                            {% firstof challenge.title challenge.short_name %}
+                                        <h3 class="m-0 px-3 flex-shrink-1">
+                                            {% firstof challenge.title|truncatewords:3 challenge.short_name|truncatewords:3 %}
```
jmsmkn: I'm not too sure that we want to do a hard truncate without telling the user. What happens if we keep the text-truncate marker from bootstrap by allowing multiple lines?

HansPinckaers: Sadly, that doesn't work because multi-line wrapping of text is non-trivial in CSS.

HansPinckaers: Also, I'm not really sure how this is different from the current situation as we truncate the title now even more?

PR #661, opened 2018-12-14
Column visibility dropdown list

HansPinckaers:
*can be tested now*

There are some assumptions in this code:
- [x] the most important metric always is the first metric-column
- [x] Datatable crashes if there no column to be toggled.

### DIFF
--- a/app/grandchallenge/evaluation/templates/evaluation/result_list.html
+++ b/app/grandchallenge/evaluation/templates/evaluation/result_list.html
@@ -191,7 +191,7 @@
 
     <script type="text/javascript">
         $(document).ready(function () {
-            $('#resultsTable').DataTable({
+            var table = $('#resultsTable').DataTable({
                 {% comment %}
                     The column index of the default sort, must match the table
                     set up.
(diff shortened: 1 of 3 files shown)


jmsmkn:
I assume that we will only want the extra results columns to be toggleable, we know how many there are from `{{ extra_results_columns|length }}`, and they always start at column 4 (zero index).

HansPinckaers:
I was thinking only to allow the metrics to be toggable and always show the supplemental information / comments?

jmsmkn:
I agree - only for extra_results_columns though. The main result should always be visible.

HansPinckaers:
It looks quite clean now. Should be tested in staging environment.

toggable

app/grandchallenge/evaluation/templates/evaluation/result_list.html
```diff
@@ -199,11 +199,50 @@ <h2>Results</h2>
...
+                buttons: [
+                    {
+                        extend: 'colvis',
+                        text: 'Additional metrics',
+                        columns: '.toggable'
+                    },
+                    {
+                        text: 'Show all metrics',
+                        action: function ( e, dt, node, config ) {
+                            if($(node).text().indexOf('Hide') > -1) {
```
HansPinckaers: I do not really like using HTML of the button to determine 'state', maybe use an attribute of a certain element.

PR #714, opened 2019-02-20
Add protected minio storage

jmsmkn:
This PR allows protected data to be stored in a minio instance rather than on disk. Users can access the data by requesting it from django, which then checks the permissions and returns an X-Accel-Redirect if the user is able to access the file and the file exists. NGINX then sets up an **internal** location block at `/$PROTECTED_S3_STORAGE_BUCKET_NAME/`.


app/grandchallenge/core/storage.py
```diff
@@ -4,27 +4,43 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(
+            *args, config=settings.PRIVATE_S3_STORAGE_KWARGS, **kwargs
+        )
+
+
+@deconstructible
+class ProtectedS3Storage(S3Storage):
+    def __init__(self, *args, internal=False, **kwargs):
+        config = {**settings.PROTECTED_S3_STORAGE_KWARGS}
```
jmsmkn: @pkcakeout Should I use deepcopy here?

jmsmkn:
- [ ] use deepcopy
- [ ] check that the buckets are different at runtime
- [ ] add a test for the accelerated redirect header

app/grandchallenge/core/storage.py
pkcakeout: yes!

pkcakeout:
Looks good as far as I understand it. When the changes from your checklist were applied, this is ready to be merged, I would say. One more time: This is a very neat and elegant solution using  the internal-redirect feature of NGINX.

PR #725, opened 2019-03-01
Adds several workstation features

HarmvZ:
- Improved import views to allow upload of 'kappadata' (flat folder of images without metadata)
- Added several API viewsets for annotations
- Added annotation permission functionality for retina-graders / retina-admins


HarmvZ:
The build is failing because I broke the pipfile.lock when I merged master into this branch. I can't run the pipenv install on my machine because it will then add the windows dependencies again. I get a `PermissionError: [Errno 13] Permission denied: '/home/django'` when I try to run the command in the container. Could you run it on your machine @jmsmkn ?

jmsmkn:
Done! We should add a Makefile command to lock the pipfile for windows users

jmsmkn:
The PR build failed, I've restarted it, if it passes we have a new flakey test that needs to be fixed.

app/config/settings.py
```diff
@@ -685,3 +685,5 @@ def strtobool(val) -> bool:
 RETINA_ADMINS_GROUP_NAME = "retina_admins"
 RETINA_IMPORT_USER_NAME = "retina_import_user"
 RETINA_EXCEPTION_ARCHIVE = "Australia"
+# Default permission types that are set for retina users on annotation models
+PERMISSION_TYPES = ("view", "add", "change", "delete")
```
jmsmkn: Rather than hard coding this for all models here, wouldn't it be better to get this from the `default_permissions` of the models meta class? You only ever use this to assign all, right? See `create_permissions` in `django.contrib.auth.management.__init__.py`

app/grandchallenge/annotations/management/commands/copyannotations.py
```diff
@@ -59,12 +81,26 @@ def handle(self, *args, **options):
                 obj.grader_id = user_to.id
                 obj.pk = None
                 obj.save()
+                if options["add_permissions"]:
+                    for permission_type in PERMISSION_TYPES:
+                        assign_perm(
+                            f"annotations.{permission_type}_{model.__name__.lower()}",
```
jmsmkn: Why use `model.__name__` here and `model.__class__.__name__` below? Should we be using `instance._meta.app_label`, and `instance._meta.model_name` instead?

app/grandchallenge/annotations/models.py
```diff
@@ -14,12 +19,73 @@ class AbstractAnnotationModel(UUIDModel):
...
+
+        if (
+            self.grader.groups.filter(
+                name=settings.RETINA_GRADERS_GROUP_NAME
+            ).exists()
+            or self.grader.groups.filter(
+                name=settings.RETINA_ADMINS_GROUP_NAME
+            ).exists()
+        ):
+            model_name = self.__class__.__name__.lower()
```
jmsmkn: Again, this should come from `_meta`

app/grandchallenge/annotations/models.py
```diff
@@ -14,12 +19,73 @@ class AbstractAnnotationModel(UUIDModel):
...
     created = models.DateTimeField(default=timezone.now)
 
     class Meta:
         abstract = True
 
+    def save(self, *args, **kwargs):
+        """ Override save method to enable setting of permissions for retina users """
+        created = False
+        if self._state.adding:
+            created = True
```
jmsmkn: Lines 30-32 can be simplified to `created = self._state.adding`

app/grandchallenge/annotations/validators.py
```diff
@@ -0,0 +1,19 @@
...
+from django.conf import settings
+
+
+def validate_grader_is_current_retina_user(grader, context):
+    """
+    This method checks if the passed grader equals the request.user that is passed in the context.
+    Only applies to users that are in the retina_graders group.
+    """
+    request = context.get("request")
+    if request and request.user.is_authenticated:
```
jmsmkn: Do you need a try block here to catch user lookups and convert those exceptions to `ValidationErrors`?

app/grandchallenge/retina_api/mixins.py
```diff
@@ -45,13 +45,17 @@ class RetinaOwnerAPIPermission(permissions.BasePermission):
...
     def has_permission(self, request, view):
         user_id = view.kwargs.get("user_id")
         user = get_user_model().objects.get(id=user_id)
-        is_owner = request.user == user
-        return is_in_retina_admins_group(request.user) or is_owner
+        is_retina_user_and_owner = request.user == user and is_in_retina_group(
+            request.user
+        )
+        is_admin = is_in_retina_admins_group(request.user)
+        return is_admin or is_retina_user_and_owner
```
jmsmkn: Just a small style thing, line 54 was hard to grok, could you expand this out into 

```
if request.user == user and is_in_retina_group(request.user):
  return True
elif is_in_retina_admins_group(request.user):
  return True
else:
  return False
```

I know it's more verbose but I think it's much easier for a future dev to see whats going on immediately.

app/grandchallenge/retina_api/urls.py
```diff
@@ -1,5 +1,5 @@
 from django.urls import path, include
-from rest_framework.routers import DefaultRouter
+from rest_framework.routers import DefaultRouter, SimpleRouter
```
jmsmkn: What URLs does the default router provide now? Just the root page as there are no models registered with it?

app/tests/retina_api_tests/helpers.py
```diff
@@ -290,10 +326,18 @@ def test_load_save_data(self, client):
         # login client
         client, grader = client_login(client, user="retina_user")
 
-        for archive in ("Rotterdam", "Australia"):
+        for archive in ("Rotterdam", "Australia", "kappadata"):
```
jmsmkn: Could you make the archives an enum? Would make usages of `kappadata` easier to handle.

app/grandchallenge/retina_core/management/commands/setannotationpermissions.py
```diff
@@ -0,0 +1,117 @@
...
+    CoordinateListAnnotation,
+    SinglePolygonAnnotation,
+    SingleLandmarkAnnotation,
+)
+from django.contrib.auth.models import Group
+from django.conf import settings
+from guardian.shortcuts import assign_perm, remove_perm
+
+# Existing annotation (name, codename) as of annotations.0001_initial
+ANNOTATION_CODENAMES = (
```
jmsmkn: These can be found from the `_meta` class of the objects.

app/grandchallenge/retina_core/management/commands/setannotationpermissions.py
```diff
@@ -0,0 +1,117 @@
...
+        change_permission_func = assign_perm
+
+    olp_count = 0  # Object level permissions assigned/removed
+    mlp_count = 0  # Model level permissions assigned/removed
+    retina_admin_group = Group.objects.get(
+        name=settings.RETINA_ADMINS_GROUP_NAME
+    )
+    for (annotation_model, annotation_codename) in ANNOTATION_CODENAMES:
+        # Change user level object permissions to owners of annotations
+        for annotation in annotation_model.objects.all():
```
jmsmkn: Do we really want to set the permissions on each row here? Or are table permissions ok?

app/grandchallenge/retina_importers/views.py
```diff
@@ -219,6 +220,23 @@ def set_element_data_file_header(self, mhd_file, raw_file_name):
             sys.getsizeof(new_file),
         )
 
+    def save_mhd_and_raw_files(self, request, img):
+        # Save MHD and ZRAW files to Image.files model
+        uuid_name_str = str(uuid.uuid4())
```
jmsmkn: Everything so far is called `out.{mhd|zraw}` - is it better to stick to this convention? We will update it later anyway to `images/<image_pk>/<imagefile_pk>.<imagefile_type>`.

app/grandchallenge/annotations/management/commands/copyannotations.py
HarmvZ: Done in 3ce431a984d771bda25d6afd42fbf573f887e140

app/grandchallenge/annotations/models.py
HarmvZ: Done in 3ce431a984d771bda25d6afd42fbf573f887e140

app/grandchallenge/retina_api/mixins.py
HarmvZ: Done in c2f4ef79527de7ca30c3884cb11cf487f8a8269b

app/grandchallenge/retina_api/urls.py
HarmvZ: Yes, which is not used. I removed it in df95f72566d0b01be27342f0da42980d3d8a1b60

app/grandchallenge/retina_importers/views.py
HarmvZ: Ok, done in a7002e0bb5e4347d2d6ef223b6d168f80db8fe54

app/config/settings.py
HarmvZ: Done in a390029ac705996ea10dd8ef173b8bac8bd64e36

app/grandchallenge/retina_core/management/commands/setannotationpermissions.py
HarmvZ: Done in a390029ac705996ea10dd8ef173b8bac8bd64e36

app/grandchallenge/annotations/validators.py
HarmvZ: Added a check for if `request.user is None` and clarified the method with an additional comment

HarmvZ:
Made a few changes:
- Fixed all requested changes from the review.
- Added `stereoscopic_choice` and `field_of_view` attributes to the `Image` model, serializer, admin, importer, tests and created a migration for it.
- Changed importer to accept AREDS type data (mainly by adding the new attributes) and did some refactoring on it.
- Converted some the dynamically added batch tests for importers to parametrized tests
- Added endpoint and tests in importers to check if an image already exists
- Added `IS_RETINA_ADMIN` constant to retina index view context and template
- Added `UserSerializer` containing `id` and `username`
- Created an endpoint and tests in `retina_api` to get all users that have polygon annotations for a specific image

PR #754, opened 2019-03-22
Add Workstations

jmsmkn:
Workstations are applications that run in containers and offer server-side rendering.


app/config/settings.py
```diff
@@ -538,7 +539,10 @@ def strtobool(val) -> bool:
     "DEFAULT_PERMISSION_CLASSES": ("rest_framework.permissions.IsAdminUser",),
     "DEFAULT_AUTHENTICATION_CLASSES": (
         "rest_framework.authentication.TokenAuthentication",
+        "rest_framework.authentication.SessionAuthentication",
     ),
+    "DEFAULT_PAGINATION_CLASS": "rest_framework.pagination.LimitOffsetPagination",
+    "PAGE_SIZE": 100,
```
jmsmkn: @HarmvZ Could you check that if we added `DEFAULT_PAGINATION_CLASS` and `PAGE_SIZE` that the ophthalmology app would still work when you have more than 100 images?

PR #759, opened 2019-03-28
Add annotation link

HarmvZ:


app/grandchallenge/algorithms/templates/algorithms/algorithm_detail.html
```diff
@@ -67,6 +70,15 @@ <h2>Jobs for this algorithm</h2>
                     </span>
                     </td>
                     <td>{{ job.result.output }}</td>
+                    {% if job.result.images.all %}
+                        <td>
+                            {% for image in job.result.images.all %}
+                                <a href="{{ job.image.cirrus_link }}&{{ settings.CIRRUS_ANNOATION_QUERY_PARAM }}={{ image.pk }}">
```
jmsmkn: Just to check - this works ok? You don't need to use `&amp;`?

HarmvZ: It works in Chrome but `&amp;` might indeed be safer/better so I changed it. 

Also, it actually was not outputting the correct URL because the settings object is not exposed in the template so I added `CIRRUS_ANNOTATION_QUERY_PARAM` to the context. Now it should work, but I can't test it in my dev environment because CIRRUS tries to get the images from the production environment.

app/config/settings.py
```diff
@@ -613,7 +613,7 @@ def strtobool(val) -> bool:
 # CIRRUS Is an external application that can view images
 CIRRUS_APPLICATION = "https://apps.diagnijmegen.nl/Applications/CIRRUSWeb_master_98d13770/#!/?workstation=BasicWorkstation"
 CIRRUS_BASE_IMAGE_QUERY_PARAM = "grand_challenge_image"
-CIRRUS_ANNOATION_QUERY_PARAM = "grand_challenge_overlay"
```
jmsmkn: Thanks for correcting the typo, but could you check for other usages in the codebase?

HarmvZ: Yes I did. I checked by performing a find all.. on the setting name. It is not used anywhere else yet.

jmsmkn: Check your query - it's used in `datasets/models.py`

HarmvZ: Hmm. I'm not sure how I missed that. Maybe I used the fixed name to search for it. 😅 
Anyway, fixed it now.

PR #775, opened 2019-04-03
Pathology support

Karelger:
- Worklists
  - Adds Worklist models
  - Adds API view with permissions for each users own set
  - Adds form pages that allow full access for staff

- Patients
  - Adds API view that can access all patient entries and query based on image type or worklist
  - Adds form pages for CRUD operations, only accessible for staff

- Studies
  - Changes to string function to show its parent Patient
  - Adds API view that can access all study entries and query based on patient id
  - Adds form pages for CRUD operations, only accessible for staff


jmsmkn:
Closing in favour of #784

Issue #768, opened 2019-04-01
Make registrations load/save correctly


HarmvZ:
- [x] Create an endpoint for getting all registrations(LandmarkAnnotations) for a list of images
  - [x] Create APIListView with list of image ids and grader id as input
  - [x] Add tests
- [ ] ? Create an endpoint for saving the registrations more neatly
- [x] Create an endpoint for getting the OCT/obs registrations for an OCT/obs image
  - [x] Create APIDetailView with either OCT or OBS image as input
  - [x] Add tests
- [x] Adapt front-end to request registration data from new endpoints

HarmvZ:
Right now the ListAPIView for landmark annotations will return a 404 when there are no registrations for a set of images. This will probably result in a lot of 404's when graders load images that they have not registrated yet. It's probably better to just return an empty array for this.

PR #777, opened 2019-04-04
Adds registration and landmark annotation endpoints

HarmvZ:
- Adds an endpoint for getting all registrations(LandmarkAnnotations) for a list of images
- Adds an endpoint for getting the OCT/obs registrations for an OCT/obs image
See #768

### DIFF
--- a/app/grandchallenge/retina_api/views.py
+++ b/app/grandchallenge/retina_api/views.py
@@ -17,7 +17,9 @@ from rest_framework_guardian import filters
 from rest_framework.exceptions import NotFound
 
 from grandchallenge.core.serializers import UserSerializer
-from grandchallenge.registrations.serializers import OctObsRegistrationSerializer
+from grandchallenge.registrations.serializers import (
+    OctObsRegistrationSerializer
+)
 from grandchallenge.retina_api.mixins import (
     RetinaAPIPermission,
     RetinaAPIPermissionMixin,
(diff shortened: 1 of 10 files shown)


app/grandchallenge/retina_api/views.py
```diff
@@ -776,3 +781,47 @@ def get_queryset(self):
...
+        user_id = self.kwargs["user_id"]
+        image_ids = self.request.query_params.get("image_ids")
+        if image_ids is None:
+            raise NotFound()
+        image_ids = image_ids.split(",")
+        user = get_object_or_404(get_user_model(), id=user_id)
+        queryset = user.landmarkannotationset_set.filter(
+            singlelandmarkannotation__image__id__in=image_ids
+        ).distinct()
+        # if len(queryset) == 0:
```
jmsmkn: Please remove the commented out code

jmsmkn:
I think that you need to update the version of black that you're using. We've moved to 19.3b0 now. Also, I'd recommend installing the pre-commit hook which will do the formatting on your changesets automatically, would eliminate the `Run black` commits.

HarmvZ:
> I think that you need to update the version of black that you're using. We've moved to 19.3b0 now. Also, I'd recommend installing the pre-commit hook which will do the formatting on your changesets automatically, would eliminate the `Run black` commits.

I updated it already, but I think because this branch was made before the update it still runs on the [old version](https://github.com/comic/grand-challenge.org/blob/13d4f28122d0314a57e97319c2586565fb9d89fd/Pipfile#L12) in the `push` build.

I installed the pre-commit hook now, that will indeed make things a lot easier!

jmsmkn:
I see, thanks!

Issue #770, opened 2019-04-02
Make ETDRS grid annotations load/save correctly


HarmvZ:
- [x] #774 
- [x] Create endpoint for getting and saving ETDRS grid annotations
  - [x] Create API viewset with grader id as input
  - [x] Add tests
- [ ] Find out what the default pixel spacing/size is for different modalities (OCT, CFI, FAF, IR)
- [x] Figure out if current pixel sizes loaded from pixels per inch in JPG/TIF files is correct
- [x] Create endpoint for retrieving the pixel size/spacing
  - [x] Create APIDetailView to get ElementSpacing header from mhd file.
  - [x] Add tests
- [x] Adapt front-end to derive ETDRS grid size from pixel size/spacing and only require selecting the fovea location.
- [x] Adapt front-end to request ETDRS data from new endpoint
- [x] Adapt front-end to save ETDRS data to new endpoint
- [ ] Fix UnboundLocalError. Sentry.io: `GC-8EQ`

HarmvZ:
I've asked Timo for some details on the pixel size/spacing for all retina images that are in the database right now. I will wait for his response before I start implementing this. Also, see #774 for more info about this.

I also need to make a decision on how to create the endpoint. I either need to get all ETDRS grid annotations for a grader or all for a user and archive or one specific annotation for a user and image. This depends a bit on how I will implement the registration endpoints so I will start with that first.

PR #800, opened 2019-04-24
Adds element spacing endpoints

HarmvZ:
See #770 
* Adds an endpoint to change ElementSpacing ITK header in existing images
* Adds an endpoint to retrieve the ElementSpacing for a certain image

@jmsmkn I'm not really sure if the functionality in `SetElementSpacingForImage.set_mhd_element_spacing_header()` (app/grandchallenge/retina_importers/views.py:656) will work with the storage method on the production server. Do you know how I can check that?


jmsmkn:
You're using the right abstractions and are using the same filestore in development now. That function is covered by tests so should be ok.

app/grandchallenge/retina_importers/views.py
```diff
@@ -609,3 +613,61 @@ class UploadPolygonAnnotationSet(AbstractUploadView):
...
+        old_mhd = image.files.get(file__endswith=".mhd")
+        old_raw = image.files.get(file__endswith="raw")
+
+        sitk_image.SetSpacing(spacing)
+        with TemporaryDirectory() as tempdirname:
+            sitk.WriteImage(
+                sitk_image, str(Path(tempdirname) / Path("out.mhd")), True
+            )
+            for file in (("out.mhd", old_mhd), ("out.zraw", old_raw)):
+                fh = open(str(Path(tempdirname) / Path(file[0])), "rb")
```
jmsmkn: You should use 
```
with open(...) as fh:
  bio.write(fh.read())
```
Rather than
```
fh = open(...)
... do stuff ...
fh.close()
```

app/tests/retina_importers_tests/helpers.py
```diff
@@ -62,7 +62,7 @@ def create_test_images():
     files = {}
     for file_type in ("mhd", "zraw"):
         files[file_type] = BytesIO()
-        fh = open(RESOURCE_PATH / "image5x6x7.{}".format(file_type), "rb")
+        fh = open(RESOURCE_PATH / f"image5x6x7.{file_type}", "rb")
```
jmsmkn: Use `with open`

PR #815, opened 2019-05-13
Fix incorrect condition

HarmvZ:
I forgot to remove `"retina_admin"` from the first condition after adding a separate check below it (line 132-143) in #813. The test then fails when `polygonset2.created` is higher than `polygonset1.created`. This should fix it.


HarmvZ:
I think this fixed it. Not sure why the submission tests are failing now, however.

jmsmkn:
The tests should pass with the commit made to master now.

Issue #822, opened 2019-05-22
3D in-plane annotation structure


HarmvZ:
For the retina workstation, the graders have requested a feature that allows them to annotate 3D OCT images. They want to be able to annotate 2D polygons on a slice of the OCT. Currently, the polygon annotation models support only 2D annotations. I want to extend [that model](https://github.com/comic/grand-challenge.org/blob/c860be4b76efc54a0bb438fb9f38d6cb528d8c8c/app/grandchallenge/annotations/models.py#L173) to also allow for the new type of annotations. The easiest way to implement it for my use case is to add a `PositiveIntegerField` that references the slice index it belongs to. 

However, after briefly talking about this with @pkcakeout to make sure that the implementation is also useful in other 3D modalities, he came up with another implementation: 
- Add a field to indicate the coordinate values as either voxel or world coordinates (either `CharField` with the 2 choices or a `BooleanField` that indicates if it is world coordinates or not)
- Add two fields to define the orientation of the `x` and `y` axes (`ArrayField` of length 3 containing `FloatField`s)
- Add a field that indicates the slice index value in the `z` direction (`FloatField`, nullable so that it does not need to be defined for 2D images)
This allows for 2D in-plane polygon annotations in any direction, which is useful for other modalities.

With my use case in mind I would then define defaults like this: 
- voxel/world=voxel
- x = [1, 0, 0]
- y = [0, 1, 0]
- z = null

What do you think about this implementation @jmsmkn and @pkcakeout?

HarmvZ:
I added 2 pull requests with the proposed changes (easy fix: #829, and plane annotation fix: #828) to clarify this issue and the two proposed options.

pkcakeout:
What is stated here is true, but I talked to @HarmvZ a second time and while brainstorming, I think that our current model is overengineerd and it is bad practice to offer two representations for the same object in our database (world vs voxel coordinates). Instead, let's limit our representation to voxel-space only.

This limits the required parameters to only:
~~~~~~~~
x = [1, 0, 0]
y = [0, 1, 0]
z = null
polygon = [...]
~~~~~~~~

We should define that we use a standard right-handed coordinate system, since this is important to derive the correct z-axis. To convert coordinates to world-space one needs to go through the image definition, but this should not matter too much since annotations only have meaning with an image anyways. Speed should not be an issue as well (maybe important for longitudinal studies in the future) since mhd/mha files have headers that can quickly be used to assemble the voxel->world transform matrices.

PR #828, opened 2019-06-03
3d in plane annotation model

HarmvZ:
Adds changes as proposed in #822


jmsmkn:
@HarmvZ could you rebase onto master? It should solve those remaining tests

PR #831, opened 2019-06-06
Screenshot generation

xieweiyi:
requirements: 
opencv==3.4.2
SimpleITK==1.1.0
Pillow==5.3.0

main function:  draw_mask_tile_3view_lite
which will generate screenshots from multiple locations of a given prediction overlaying the corresponding scan.


jmsmkn:
Thanks for this! I'm going to close this as we still need to make a lot of changes to get this feature working, and we should do this on the screenshot_generation branch.

Issue #795, opened 2019-04-11
Fix read the docs generation


jmsmkn:
We're using autodoc which requires an install of the dev environment. Read the docs does not support pipenv yet, so we should be able to fix when this PR is released: https://github.com/rtfd/readthedocs.org/pull/4783

For the time being, local generation works fine.

PR #837, opened 2019-06-25
Update docs with additions for windows specific installation

HarmvZ:
@jmsmkn Could you please check if the docs build succeeds and the formatting is correct before merging this?

Building the docs fails on my machine with the error:
`ModuleNotFoundError: No module named 'djcelery_email'`

When I try to install that module it returns:
`Could not find a version that satisfies the requirement djcelery_email (from versions: ) No matching distribution found for djcelery_email`. 

I'm not sure what's going on there but I think that's probably some Windows issue again and it's probably not really useful to spend time on that right now.

### DIFF
--- a/docs/getting-started.rst
+++ b/docs/getting-started.rst
@@ -8,7 +8,14 @@ To develop the platform you need to have docker and docker-compose running on yo
 Installation
 ------------
 
-1. Install Docker_ and `Docker Compose`_ 
+1. Download and install Docker
+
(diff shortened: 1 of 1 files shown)


jmsmkn:
The docs generation are failing due to rtfd/readthedocs.org#4783 and we have an issue on it too (#795). I will double check this when upstream is fixed. This might not happen any time soon so we may need to setup an alternative to pipenv for managing our dependencies.

PR #854, opened 2019-07-23
Create DZI when tiff files are added to Grand Challenge

miriam-groeneveld:


app/tests/cases_tests/test_tiff.py
```diff
@@ -65,6 +70,48 @@ def test_tiff_validation(resource, expected_error_message):
...
+            "Image has incomplete tile information",
+        ),
+    ],
+)
+def test_dzi_creation(
+    source_dir, filename, expected_error_message, tmpdir_factory
+):
+    error_message = ""
+
+    try:
```
jmsmkn: A more pythonic way to do this would be
```
with pytest.raises(ValidationError) as err:
    ...

assert err.message == expected_message
```

app/grandchallenge/cases/models.py
```diff
@@ -351,3 +356,28 @@ class ImageFile(UUIDModel):
...
+
+
+class FolderUpload:
+    def __init__(self, image, folder):
+        self.image = image
+        self.folder = folder
+
+    def save(self):
+        for level in [d for d in os.scandir(self.folder) if d.is_dir()]:
+            for file in [f for f in os.scandir(level) if f.is_file()]:
```
jmsmkn: Can we simplify this?

app/grandchallenge/cases/image_builders/tiff.py
```diff
@@ -175,3 +198,19 @@ def create_tiff_image_entry(*, tiff_file: GrandChallengeTiffFile) -> Image:
...
+def create_dzi_images(*, tiff_file: GrandChallengeTiffFile) -> str:
+    # Creates the .dzi and corresponding tiles
+    dzi_filename = str(tiff_file.path.absolute()).replace(".tif", "")
+
+    try:
+        image = pyvips.Image.new_from_file(
+            str(tiff_file.path.absolute()), access="sequential"
+        )
+
+        pyvips.Image.dzsave(image, dzi_filename, tile_size=2560)
```
jmsmkn: Move the 2560 to settings

app/tests/cases_tests/test_tiff.py
```diff
@@ -103,10 +150,24 @@ def test_tiff_image_entry_creation(resource, expected_error_message):
...
     assert len(image_builder_result.consumed_files) == 1
     assert len(image_builder_result.new_images) == 1
-    assert len(image_builder_result.new_image_files) == 1
+    assert len(image_builder_result.new_image_files) == 2
+
+    # Asserts successful creation of files
+    assert os.path.isfile(temp_dir / "valid_tiff.dzi")
+    assert os.path.isdir(temp_dir / "valid_tiff_files")
+
+    for _, _, filenames in os.walk(temp_dir / "valid_tiff_files"):
```
jmsmkn: I think that this loop can be replaced with a glob, right?

app/tests/cases_tests/test_tiff.py
jmsmkn: Also, remove the lines that are not expected to raise the error out of the `try` or `with` block

app/grandchallenge/cases/tasks.py
```diff
@@ -281,7 +308,7 @@ def build_images(upload_session_uuid: UUID):
                         pass
 
             except Exception as e:
-                upload_session.error_message = str(e)
+                upload_session.error_message = str(e)[0:255]
```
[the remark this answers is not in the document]
jmsmkn: Actually, this would cause a migration so make it a variable in models.py

app/grandchallenge/cases/image_builders/tiff.py
```diff
@@ -175,3 +198,19 @@ def create_tiff_image_entry(*, tiff_file: GrandChallengeTiffFile) -> Image:
         stereoscopic_choice=Image.STEREOSCOPIC_UNKNOWN,
         field_of_view=Image.FOV_UNKNOWN,
     )
+
+
+def create_dzi_images(*, tiff_file: GrandChallengeTiffFile) -> str:
+    # Creates the .dzi and corresponding tiles
+    dzi_filename = str(tiff_file.path.absolute()).replace(".tif", "")
```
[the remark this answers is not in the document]
miriam-groeneveld: No, I actually need the path, so the dzi is created in the same temporary folder as the tif file. I changed the line to str(tiff_file.path.parent).

app/grandchallenge/cases/image_builders/tiff.py
```diff
@@ -126,10 +127,12 @@ def image_builder_tiff(path: Path) -> ImageBuilderResult:
     new_image_files = []
     consumed_files = set()
     invalid_file_errors = {}
+    new_folder_upload = []
 
     for file_path in path.iterdir():
         try:
             tiff_file = load_tiff_file(path=file_path)
+            dzi_output = create_dzi_images(tiff_file=tiff_file)
```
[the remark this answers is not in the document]
miriam-groeneveld: It's an ImageFile, not an Image.

app/grandchallenge/cases/models.py
```diff
@@ -339,8 +339,13 @@ class Meta:
 class ImageFile(UUIDModel):
     IMAGE_TYPE_MHD = "MHD"
     IMAGE_TYPE_TIFF = "TIFF"
+    IMAGE_TYPE_DZI = "DZI"
```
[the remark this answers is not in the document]
miriam-groeneveld: It's a property of the ImageFile.

app/tests/cases_tests/test_tiff.py
miriam-groeneveld: In that case I would have to split the test, because the first (valid) file that is tested doesn't raise an error. I could, make a separate test for valid files, and one for invalid files.

jmsmkn:
One last thing - could we use a smaller file for valid_tiff? 15MB is quite a lot to add to our git history.

PR #862, opened 2019-07-30
Adds Reader Studies

jmsmkn:
Reader studies allow users to gather annotations for a set of medical images. They support:

- setting a readers group, 
- uploading images, 
- using a workstation, 
- displaying those images in a particular order in a particular layout/hanging, 
- setting questions that require a text or boolean answer on a set of images,
- setting questions that require drawing of 2d bouding boxes on an image,
- setting the order of questions
- setting and downloading answers over a REST API

TODO:

- [x] Merge #878
- [x] Add help text on the reader study and workstation forms
- [x] Add validation for bounding box answers


app/grandchallenge/reader_studies/models.py
```diff
@@ -0,0 +1,310 @@
...
+        "required": ["main"],
+        "additionalProperties": False,
+        "properties": {
+            "main": {
+                "$id": "#/items/properties/main",
+                "type": "string",
+                "title": "The Main Schema",
+                "default": "",
+                "examples": ["im1.mhd"],
+                "pattern": "^(.*)$",
```
pkcakeout: Is this auto-generated? This is not necessary and quite verbose.

app/grandchallenge/reader_studies/signals.py
```diff
@@ -0,0 +1,40 @@
...
+    if action not in ["post_add", "post_remove", "pre_clear"]:
+        # nothing to do for the other actions
+        return
+
+    if reverse:
+        images = Image.objects.filter(pk=instance.pk)
+        try:
+            reader_studies = model.objects.filter(pk__in=pk_set)
+        except TypeError:
+            # When a clear method is used pk_set is none and a TypeError raised
```
pkcakeout: Why not do a `if pk_set is None` then? has `pk_set` not been retrieved yet and this is deferring the database interaction/optimizing the number of requests? Otherwise, branching with exceptions is quite naughty and can better be rewritten as an if-statement.

app/grandchallenge/reader_studies/models.py
jmsmkn: I think that it came from something that was auto generated. Feel free to change it but check that the JSON editor widget is happy with it (I remember getting problems there when I implemented a basic schema)

jmsmkn:
> You were working on the permissions-system, right? I remember that you said that this is undergoing changes right now!?

Yes, PR coming today.

app/grandchallenge/reader_studies/models.py
```diff
@@ -0,0 +1,310 @@
...
+
+    def is_answer_valid(self, *, answer):
+        return self.ANSWER_TYPE_VALIDATOR[self.answer_type](answer)
+
+
+class Answer(UUIDModel):
+    creator = models.ForeignKey(get_user_model(), on_delete=models.CASCADE)
+    question = models.ForeignKey(Question, on_delete=models.CASCADE)
+    images = models.ManyToManyField("cases.Image", related_name="answers")
+    answer = JSONField()
```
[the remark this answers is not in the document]
jmsmkn: The validation of `Answer.answer` depends on `Question.answer_type` - I think that this is hard to encapsulate in a single schema that you could pass to `JSONField` (correct me if I'm wrong). Instead, the validation is done on the serializer using the lambda functions defined here: https://github.com/comic/grand-challenge.org/pull/862/files/c8adae6cff593866fa1496949e2e049aa98cc583#diff-5e52064feade65c59ca2de9e78d45791R212

That data structure will need to be extended for bounding boxes or any other answer type that we add.

app/grandchallenge/reader_studies/signals.py
jmsmkn: Fixed. It didn't feel right to me (I don't really like all the work you need to do in this function, and that `pk_set is None` means you clear the entire through table, but I trust Django not to change this API and have it well tested). But, `pk_set` is already resolved, so doing this test saves actually saves a database trip in the `_clear` action case.

jmsmkn:
Merge #878 into here before merging this to master.

jmsmkn:
Support for the bounding box question has now been added.

PR #872, opened 2019-08-01
Algorithm job api post

haimasree:
Closes: #865

### DIFF
--- a/app/tests/algorithms_tests/test_api.py
+++ b/app/tests/algorithms_tests/test_api.py
@@ -1,5 +1,7 @@
 import pytest
 
+from grandchallenge.algorithms.models import Job
+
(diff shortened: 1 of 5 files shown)


app/tests/algorithms_tests/test_api.py
```diff
@@ -131,3 +134,27 @@ def test_result_api_permissions(client):
             content_type="application/json",
         )
         assert response.status_code == test[1]
+
+
+@pytest.mark.django_db
+def test_job_create(client):
```
jmsmkn: Could you add a test to check for the permissions of this view?

haimasree: Tests parameterized

PR #873, opened 2019-08-01
Adds WorkstationImage and Session model ordering

HarmvZ:
Orders the WorkstationImage and Session models by default so that the newest models are always on top.
I can add a test for this if you want, but I don't know if that is really necessary.


jmsmkn:
I will close this as we will take it up in another migration.

PR #877, opened 2019-08-12
Add design documentation

pkcakeout:
(see commit listing)


pkcakeout:
Updated the docs, did we need to change something else? I can do the apidocs at a later stage and open a new PR. Then this does not need to stay open forever.

PR #905, opened 2019-08-29
Fix cascading delete of archive

### LINKED ISSUE #900: Create a management command to remove a complete archive with all related patients, studies, images, annotations, registrations, etc
The current "Rotterdam Study 1" archive has to be removed and reuploaded. Although this might not happen often, doing this manually in the admin is not time efficient and error-prone. Therefore, I think we should develop a way in which this can be done easier. There are a few difficulties when doing this because of the way the models are set up:

- Removing an `Archive` will leave all other data intact.

- Removing a `Patient` or `Study` is not possible when it has related images since the study foreign key field on an image contains `on_delete=models.PROTECT`.

- Removing an `Image` will also remove all related annotations, but not any related `Study`/`Patient`/`Archive`/`ImageFile`.

I think the best way to achieve this is to make a management command with `Archive.id` as input that removes the Archive and all of the related models.

HarmvZ:
Closes #900


app/grandchallenge/archives/models.py
```diff
@@ -13,4 +14,9 @@ class Archive(UUIDModel):
     images = models.ManyToManyField(Image)
 
     def __str__(self):
-        return "<{} {}>".format(self.__class__.__name__, self.name)
+        return f"<{self.__class__.__name__} {self.name}>"
+
+    def delete(self, *args, **kwargs):
+        # Remove all related patients and other models via cascading
+        Patient.objects.filter(study__image__archive__id=self.id).delete()
```
jmsmkn: I think that .delete() should also get the `*args` and `**kwargs`.

What happens if the `Image` also belongs to another archive?

HarmvZ: It would have been deleted. I changed the method now.

HarmvZ:
I changed the method to only remove images that belong exclusively to the archive that is removed. Also expanded the test to test for: 
- archive-only images
- image with ImageFiles
- dual archive images
- image in study but not in same archive as other images in that study

I tried to find a nice way to do this with the Django ORM but ended up going with a more manual approach to remove the objects. The method may be a bit inefficient now, but I think that is not a big problem since it will probably only be used sporadically and only by superusers.

app/grandchallenge/archives/models.py
```diff
@@ -13,4 +17,42 @@ class Archive(UUIDModel):
...
+        related_patients = Patient.objects.filter(
+            study__image__in=images_to_remove
+        ).distinct()
+        related_studies = Study.objects.filter(
+            image__in=images_to_remove
+        ).distinct()
+        for patient in related_patients:
+            if patient.id not in protected_patient_ids:
+                patient.delete(*args, **kwargs)
+        for study in related_studies:
```
jmsmkn: I think these 2 for loops can be replaced with an extra filter in each of the queries, eg `.exclude(pk__in=protected_patient_ids)`

app/grandchallenge/archives/models.py
```diff
@@ -13,4 +17,42 @@ class Archive(UUIDModel):
...
+        exclusively to this archive
+        """
+        images_to_remove = (
+            Image.objects.annotate(num_archives=Count("archive"))
+            .filter(archive__id=self.id, num_archives=1)
+            .all()
+        )
+        protected_study_ids = set()
+        protected_patient_ids = set()
+        for image in images_to_remove:
```
jmsmkn: I don't really understand what this block is trying to do, could you split this out into a separate (maybe protected) function so that developers can easier understand the flow? Or, even a comment would do.

HarmvZ: I moved this functionality into a separate method and tried to add a comment that explains what it does. Please let me know if the explanation is clear!

PR #906, opened 2019-08-29
Allow archive only images in Archive API endpoint

HarmvZ:
Resolves https://sentry.io/organizations/grand-challenge/issues/1195527370/?project=303639

I thought I wrote the test for this so that it already tested if archive-only images are allowed. But I did not account for the fact that the ImageFactory will create a Study automatically when it is not specified. I explicitly declared the `study=None` in the fixture, so now it is tested.

### DIFF
--- a/app/grandchallenge/retina_api/views.py
+++ b/app/grandchallenge/retina_api/views.py
@@ -925,7 +925,8 @@ class ArchiveAPIView(APIView):
                         archive_only_images.append(image)
 
                 images = sorted(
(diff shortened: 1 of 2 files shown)


HarmvZ:
The push build on Travis fails with `curl: (7) Failed to connect to github.com port 443: Connection timed out` when trying to fetch docker-compose from GitHub. Probably requires a manual restart of the build to fix? The pr build passes.

PR #908, opened 2019-08-30
Adds Distance measurement option

### LINKED ISSUE #907: Add distance measurement choice for answer type
For validation this requires an image port

jmsmkn:
This PR:

- Adds distance measurement annotations
- Increases the default order value so users have more numbers to play with
- Adds a required option to questions (not implemented on the front end)
- Adds a help text for each question that the user can use for clarity
- Changes the widgets used in the forms to encourage short questions

Closes #907


app/grandchallenge/core/management/commands/init_gc_demo.py
```diff
@@ -81,6 +82,7 @@ def handle(self, *args, **options):
         self._create_demo_challenge()
         self._create_external_challenge()
         self._create_algorithm_demo()
+        self._create_workstation()
```
pkcakeout: Very helpful, thanks!

app/grandchallenge/reader_studies/models.py
```diff
@@ -347,12 +383,22 @@ def assign_permissions(self):
...
+            in [
+                self.ANSWER_TYPE_2D_BOUNDING_BOX,
+                self.ANSWER_TYPE_DISTANCE_MEASUREMENT,
+            ]
+        ) != bool(self.image_port):
+            raise ValidationError(
+                "The image port must (only) be set for annotation questions."
+            )
+
+        if self.answer_type == self.ANSWER_TYPE_BOOL and self.required:
```
pkcakeout: "Required" for text would mean "something other than whitespace", I guess? Good catch otherwise!

jmsmkn: Yes, my intention for this is the same as htmls input required attr: "specifies that an input field must be filled out before submitting the form". Will simplify the implementation if we ever generate pure html forms :-)

PR #909, opened 2019-09-02
Adds shuffling of the hanging list per user

### LINKED ISSUE #890: Reader study order randomization for images
There should be a flag on the reader-studies page that randomizes the order of scans presented to a reader.

**Additional context**
Randomized trials are standard scientific method to counteract biases in answers. We should have that feature.

jmsmkn:
Implementation done, needs a test

Closes #890


jmsmkn:
I think that this is now ready for review.

PR #912, opened 2019-09-04
Adds keep alive endpoint for workstation sessions

### LINKED ISSUE #911: A workstation session keep-alive endpoint
The workstations sessions will automatically time out after 10 minutes. We need a way to keep these alive. 

Add a patch endpoint for this, setting the session maximum duration to now + the timeout. Reduce the timeout to 5 minutes perhaps. Call this endpoint every minute from JS. The downside to this is that someone could keep a session alive forever, but these can be stopped in the admin.

jmsmkn:
This adds an `workstations/sessions/{pk}/keep_alive/` endpoint for the users workstation session that only does 1 thing - increases the expiration time of the session to now + 5 minutes, up to a maximum duration of 10000 seconds (default, environment variable). This will help us fix an issue where the workstation will shut down after 10 minutes, so long as we periodically call this function from JS. 

Closes #911

### DIFF
--- a/app/config/settings.py
+++ b/app/config/settings.py
@@ -640,8 +640,8 @@ WORKSTATIONS_MAXIMUM_SESSIONS = int(
 )
 # The name of the group whose members will be able to create workstations
 WORKSTATIONS_CREATORS_GROUP_NAME = "workstation_creators"
-WORKSTATIONS_MAXIMUM_TIMEOUT = int(
-    os.environ.get("WORKSTATIONS_MAXIMUM_TIMEOUT", "10000")
+WORKSTATIONS_SESSION_DURATION_LIMIT = int(
+    os.environ.get("WORKSTATIONS_SESSION_DURATION_LIMIT", "10000")
 )
 
 # Disallow some challenge names due to subdomain or media folder clashes
--- a/app/config/settings.py
+++ b/app/config/settings.py
@@ -640,6 +640,9 @@ WORKSTATIONS_MAXIMUM_SESSIONS = int(
 )
 # The name of the group whose members will be able to create workstations
 WORKSTATIONS_CREATORS_GROUP_NAME = "workstation_creators"
+WORKSTATIONS_MAXIMUM_TIMEOUT = int(
+    os.environ.get("WORKSTATIONS_MAXIMUM_TIMEOUT", "10000")
+)
 
 # Disallow some challenge names due to subdomain or media folder clashes
 DISALLOWED_CHALLENGE_NAMES = [
--- a/app/grandchallenge/core/permissions/rest_framework.py
+++ b/app/grandchallenge/core/permissions/rest_framework.py
@@ -1,0 +1,26 @@
+from rest_framework.permissions import DjangoObjectPermissions
+
+
+class DjangoObjectOnlyPermissions(DjangoObjectPermissions):
+    """
+    Workaround for using object permissions without setting model perms,
+    which is required by the implementation in Django Rest Framework. It
+    enforces that the user is logged in so that the permissions can be checked
+    later.
+
+    See similar issue detailed here:
+    https://stackoverflow.com/questions/34371409/
+    """
+
+    def has_permission(self, request, view):
+        # Workaround to ensure DjangoModelPermissions are not applied
+        # to the root view when using DefaultRouter.
+        if getattr(view, "_ignore_model_permissions", False):
+            return True
+
+        if not request.user or (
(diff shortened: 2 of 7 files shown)


app/config/settings.py
```diff
@@ -640,6 +640,9 @@ def strtobool(val) -> bool:
 )
 # The name of the group whose members will be able to create workstations
 WORKSTATIONS_CREATORS_GROUP_NAME = "workstation_creators"
+WORKSTATIONS_MAXIMUM_TIMEOUT = int(
+    os.environ.get("WORKSTATIONS_MAXIMUM_TIMEOUT", "10000")
```
HarmvZ: If I understand correctly, this (and the equally named environment variable) will enforce a maximum of 10.000 seconds (~2,78 hours). I think for my use case it will be better if the max is >8 hours. As the graders will probably want to use the workstation for a full working day without needing to restart the session.

app/grandchallenge/core/permissions/rest_framework.py
```diff
@@ -0,0 +1,26 @@
...
+    later.
+
+    See similar issue detailed here:
+    https://stackoverflow.com/questions/34371409/
+    """
+
+    def has_permission(self, request, view):
+        # Workaround to ensure DjangoModelPermissions are not applied
+        # to the root view when using DefaultRouter.
+        if getattr(view, "_ignore_model_permissions", False):
```
HarmvZ: **Warning**: Nitpicking ahead
In the stackoverflow thread it says that this workaround is undocumented and I can't even find any documentation on the `view._ignore_model_permissions` attribute or any usage of it (except [here](https://github.com/encode/django-rest-framework/blob/89ac0a1c7ef6153312890376ac142674b28b04f1/rest_framework/schemas/views.py#L14)). It's not really clear to me why this is needed. You are not using the DefaultRouter for the Sessions ViewSet. I guess it is better for future compatibility if someone will use this Permissions object on a ViewSet with a DefaultRouter. Although I do not see why someone would use the DefaultRouter over the SimpleRouter for grand-challenge. As far as I understand, the API root view is only useful when you have a public facing API.

app/tests/workstations_tests/test_permissions.py
```diff
@@ -262,3 +264,40 @@ def test_session_api_permissions(client, two_workstation_sets, viewname):
                 assert response.json()["count"] == 0
         else:
             assert response.status_code == test[1]
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize("method", ["put", "patch"])
+def test_session_api_patch_permissions(client, two_workstation_sets, method):
+    tests = (
```
HarmvZ: Why not add these different tests also as parametrized tests?

app/config/settings.py
jmsmkn: This is why I made it an environment variable, we can set that in production for our use case :-)

app/tests/workstations_tests/test_permissions.py
jmsmkn: Because we need to parametrise it with a fixture, as far as I can tell this is not possible yet (there is a proposal here: https://docs.pytest.org/en/latest/proposals/parametrize_with_fixtures.html)

app/grandchallenge/core/permissions/rest_framework.py
jmsmkn: No, good question! 

On stack overflow they are implying that the dirty hack would be:

```
class SessionViewSet(...): # Note, using the view set here
    _ignore_model_permissions = True
```

This would be using the undocumented attribute to bypass `has_permission()` which is implemented by `DjangoModelPermissions` through the super class of `DjangoObjectPermissions`. Note that the dirty hack would bypass the authentication check, this is not something we want.

What I have done with `DjangoObjectOnlyPermissions` is copy the `has_permission()` function from `DjangoModelPermissions` verbatim, kept the authentication check, and only replaced the last 3 lines of code that checks the model permissions, replacing them with True as object permissions will be checked by `has_object_permissions` and (if the view set is hooked up correctly) the `DjangoObjectPermissionsFilter`. 

You could argue if we need to keep `if getattr(view, "_ignore_model_permissions", False): return True` line, but, it's there for consistency as we do use this with `DefaultRouter` in `SessionViewSet`.

HarmvZ: Ah, I see the use of the `DefaultRouter` now. I missed it because I was looking in `workstations/urls.py`. I think this is indeed the right way to do it since implementing the model permissions seems redundant to me when the object permissions are set correctly (and more work).

PR #916, opened 2019-09-10
Make logic in delete method atomic

HarmvZ:
This ensures all DB queries that are executed in the delete method will rollback if something fails.

### DIFF
--- a/app/grandchallenge/archives/models.py
+++ b/app/grandchallenge/archives/models.py
@@ -1,4 +1,4 @@
-from django.db import models
+from django.db import models, transaction
 from django.db.models import Count
 
 from grandchallenge.core.models import UUIDModel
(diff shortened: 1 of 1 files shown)


HarmvZ:
> Just a question - what happens to the image files if this is aborted?

By default, Django does not delete the associated files on the filesystem when a model is deleted. From the [docs](https://docs.djangoproject.com/en/dev/ref/models/fields/#django.db.models.fields.files.FieldFile.delete): 

> Note that when a model is deleted, related files are not deleted. If you need to cleanup orphaned files, you’ll need to handle it yourself (for instance, with a custom management command that can be run manually or scheduled to run periodically via e.g. cron).

We need to decide if we do want that to happen and implement it by overriding the `ImageFile.delete()` method. I think this would be preferred since you otherwise end up with orphaned files on the filesystem that you can not link back to their models. The other option is to implement a soft delete of the `ImageFile`/`Image` models which enables us to restore them if necessary. What do you think @jmsmkn?

jmsmkn:
I think that we can add a periodic task later that cleans up the folders left behind, they're pretty organised on the storage.

PR #920, opened 2019-09-20
Added option for defining annotations for reader studies as a multipl…

miriam-groeneveld:
…e of measurement lines. Merged JSON validation schemas. Added description when reader study configuration is not valid.


app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
```diff
@@ -21,7 +21,8 @@ <h1>Reader Study {{ object.title }}</h1>
 
     {% if not object.is_valid %}
         <div class="alert alert-danger">
-        This reader study is not ready to be used. Please update the configuration.
+        This reader study is not ready to be used. Please update the configuration. All of the images should be included in the hanging list
+        exactly once.
```
jmsmkn: There are actually more requirements than this, I would remove this message as we need to determine the exact reason why the validation has failed.

Issue #942, opened 2019-10-09
Add user permissions to RawImageUploadSessionViewSet and RawImageFileSerializer


jmsmkn:
**Requirements**

* The request user must have `view_algorithm` permissions on `algorithm_image.algorithm`
* `algorithm_image.ready == True`
* The request user must have permissions to use the upload with the id (check the upload session on each RawImageFile)

PR #948, opened 2019-10-14
Add raw image file serializer

haimasree:
Closes: #942


haimasree:
@jmsmkn I am not sure what exactly is the problem here with flake8. I have adjusted the test and they pass  but not too sure if I did it correctly. Let me know.

jmsmkn:
Ok, here's how to interpret the error message:

```
/app/tests/cases_tests/test_api.py:24:16: B009 Do not call getattr with a constant attribute value, it is not any safer than normal property access.
```

In file `/app/tests/cases_tests/test_api.py` on line `24` column `16` there is a programming error - Do not call getattr with a constant attribute value, it is not any safer than normal property access.

The code in question is this (line 24, column 16 to the end of the line):

```
getattr(client, "get")(
```

In short, we shouldn't access attributes in that way (with a constant attribute value `"get"`) as it's confusing for the reader. If we know what attribute we want to get from the object we would do this instead:

```
client.get(
```

haimasree:
Yes, figure it out shortly after. Sorry about the premature message

app/grandchallenge/cases/views.py
```diff
@@ -94,6 +98,17 @@ class RawImageUploadSessionViewSet(
...
         serializer.save(creator=self.request.user)
+
+
+class RawImageFileViewSet(
+    CreateModelMixin, RetrieveModelMixin, ListModelMixin, GenericViewSet
+):
+    serializer_class = RawImageFileSerializer
+    queryset = RawImageFile.objects.all()
+    permission_classes = [DjangoObjectOnlyWithCustomPostPermissions]
+    filter_backends = [ObjectPermissionsFilter]
```
jmsmkn: We need to add validation to the Serializer to check extra permissions on the POST view (see reader study answers):
- [ ] Check that the user has view permissions for the related `RawImageUploadSession`
- [ ] Add setting the algorithm field and check that the user has permission to view that algorithm

haimasree: @jmsmkn Just double checking RawImageFile  should have an "algorithm" field?

jmsmkn:
No, the `RawImageUploadSession` has this attribute, see
`RawImageUploadSession.algorithm_image` in `cases/models.py`

app/grandchallenge/cases/serializers.py
```diff
@@ -50,3 +56,28 @@ class Meta:
...
+        algorithm_image = upload_session.algorithm_image
+        user = self.context.get("request").user
+        if upload_session.creator != user:
+            raise ValidationError(
+                f"User {user} does not have permission to see this raw image upload session "
+            )
+        if algorithm_image.creator != user:
+            raise ValidationError(
+                f"User {user} does not have permission to see the corresponding algorithm image "
+            )
```
jmsmkn: These are not the correct checks, see the permissions tests for the algorithms `execution-session-create` view. 

We should check if the user has permission to view the upload session (this was set in `cases/models.py`) using `user.has_perm()`.

The algorithm permission check should be done in the upload session serializer, and should check if the user has the permission to view the algorithm that the algorithm image belongs to.

haimasree:
@jmsmkn 
I am running a local instance of grand challenge. But I cant seem to be able to access  gc.localhost/api/v1/algorithms/create/. I am token authentication with the token that gets printed out in the console.
Also, may I verify the sequence of post requests here? 

- [ ] gc.localhost/api/v1/cases/upload-sessions/ (post request parameters: creator, and an algorithimage)
- [ ] gc.localhost/api/v1/cases/image-files/ (uplaod session and a filename)
- [ ]  Keep checking gc.localhost/api/v1/algorithms/jobs until the state reads finished
- [ ] Downlaod results from gc.localhost/api/v1/algorithms/results

jmsmkn:
> @jmsmkn
> I am running a local instance of grand challenge. But I cant seem to be able to access gc.localhost/api/v1/algorithms/create/. I am using token authentication with the token that gets printed out in the console.

That's correct - the `AlgorithmViewSet` inherits `ReadOnlyModelViewSet`, so there is no API create view. You need to create algorithms and upload algorithm container images via the template views.

> Also, may I verify the sequence of post requests please?
> 
> * [ ]  gc.localhost/api/v1/cases/upload-sessions/ (post request parameters: creator, and an algorithimage)
> * [ ]  gc.localhost/api/v1/cases/image-files/ (uplaod session and a filename)
> * [ ]   Keep checking gc.localhost/api/v1/algorithms/jobs until the state reads finished
> * [ ]  Downlaod results from gc.localhost/api/v1/algorithms/results

That's correct as implemented right now, but I think that a job will not be created as the upload session (step 1) will start processing image files immediately, before the image-files have been attached to the upload-session (step 2). If that is the case, we will need to change the implementation here.

haimasree:
I received otsu container from Harm. The compressed size is 86 MB. I am getting 413 request entitty too large. Any workaround for this?

jmsmkn:
How are you uploading it? There is no way to upload containers via the API,
so you should be using some sort of template view for this, which shouldn't
have any problems as it is chunking the image.

haimasree:
Its alright. Harm helped me to upload it. I wasnt aware of the algorithm user

haimasree:
James' prediction is correct. I will rename this branch to confirm with the name rules which will close this PR.

PR #952, opened 2019-10-22
Add endpoint for lazy loading archives, patients, studies and images

### LINKED ISSUE #951: Create API endpoints to support lazy loading of Archives/Patients/Studies and Images
In the old and new ophthalmology workstation all archive data gets loaded from one endpoint. Since more and more data is being imported in the workstation, this endpoint has become really slow. Besides that, all data from the endpoint is loaded in the Vuex store which also results in a hit on performance. 
The solution for this is to lazy load the data that is actually required. This will improve initial load time and overall performance.
This requires a new endpoint in the backend that takes in a UUID of either an archive, patient or study and returns it's child objects and images.

HarmvZ:
Closes #951

### DIFF
--- a/app/grandchallenge/challenges/serializers.py
+++ b/app/grandchallenge/challenges/serializers.py
@@ -1,0 +1,9 @@
+from rest_framework import serializers
+
+from grandchallenge.challenges.models import ImagingModality
+
+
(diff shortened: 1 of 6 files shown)


jmsmkn:
Re-running the PR build

PR #963, opened 2019-11-04
614 feature support for 4dct images

### LINKED ISSUE #614: Support for 4DCT images
Midas asked about support for 4DCT images on grand-challenge and allow processors to process them. In particular the goals should be to:

- Allow uploading DICOM files so that algorithms can be tried out by regular clinicians
- Processors must be able to read extra image attributes

# 4D-MHD file converison

The 4DCT group has written Python software that can assemble numpy ndarrays from loose DICOM images and extract the required DICOM tags. We can repurpose this code and implement it as an image assembler that creates MHD files. Extra image attributes can be directly written into the MHD files themselves.

Sil made an itk image reader/writer that supports a limited subset of features of the MetaIO file format, but can write correct 4D ITK images. The problem with this custom reader/writer is that it does not support the full set of features that an mhd file does support, most notably it does not support writing and reading extra attributes. When this module was created SimpleITK did not support 4D Images. Recently, thanks to changes in SimpleITK, this module _can_ be used to load and write 4D images:

~~~~~~~~~~
# Using SimpleITK 1.1.0 on Linux(!)

import numpy as np
import SimpleITK as sitk
import SimpleITK._SimpleITK as _sitk

a = np.ndarray([20, 20, 500, 500], dtype=np.float32)
a[1, 2, 3, 4] = 400
img = sitk.Image(a.shape[::-1], sitk.sitkFloat32, 1)
_sitk._SetImageFromArray( a.tostring(), img )

print("Equal?", (a == sitk.GetArrayViewFromImage(img)).all())

# Extra disclaimer: For some reason this does not seem to work on Windows with python2.7. It works on Linux and on repl.it: https://repl.it/@PKCakeout/FatalLatestApplet
~~~~~~~~~~

It is slightly hacky and might warrant a pull request to SimpleITK so that they add the proper "SetImageFromArray" function to their API - It is there just not exposed directly via the API.

There is not need to port Sil's itk reader for this reason, and we can rely on SimpleITK to do the saving instead. 

# Action points

- [x] Modify the [model](https://github.com/comic/grand-challenge.org/blob/master/app/grandchallenge/cases/models.py) to support 4D.
- [x] Modify existing [mhd image builder](https://github.com/comic/grand-challenge.org/blob/master/app/grandchallenge/cases/image_builders/metaio_mhd_mha.py) to support 4D images
- ~~Create an [image builder](https://github.com/comic/grand-challenge.org/tree/master/app/grandchallenge/cases/image_builders) from 4DCT-groups's dicom reader. The reader should use SimpleITK and can add required dicom tags as image attributes to the mhd files.~~ This issue can be found here #962.

silvandeleemput:
This PR closes #614


app/grandchallenge/cases/models.py
```diff
@@ -271,6 +272,8 @@ def __str__(self):
     @property
     def shape_without_color(self) -> List[int]:
         result = []
+        if self.timepoints is not None:
+            result.append(self.timepoints)
```
jmsmkn: Should timepoints `t` be returned before depth `z`? I thought that `[x, y, (z), (t)]` would be the logical order but maybe this is done differently in the libraries we use? 

Another thing, is there a reason why we're not including the `resolution_levels` (`u` in pyramid formats)?

app/grandchallenge/cases/image_builders/metaio_utils.py
```diff
@@ -0,0 +1,165 @@
...
+                        key, value = line.split("=", 1)
+                        result[key.strip()] = value.strip()
+                    else:
+                        result[line.strip()] = None
+            if "ElementDataFile" in result:
+                break  # last parsed header...
+    return result
+
+
+def load_sitk_image_with_nd_support_from_headers(
```
jmsmkn: It would be nice to simplfy this function a little, there are several places where some functions can be split out to help with readability. Same with `parse_mh_header`. If you have no time can do it.

app/grandchallenge/cases/models.py
silvandeleemput: Ah, I assumed that the shape method on the Cases model should follow numpy ordering, so: `[t, z, y, x]`, because of this [test ](https://github.com/comic/grand-challenge.org/blob/d6cef76cb85affc9ad0016c3b286bfba7cc11697/app/tests/cases_tests/test_background_tasks.py#L163L186) in test_background_tasks.py. However, I just noticed that it is based on the `image5x6x7.mhd` file which is the only mhd file in the test files that has its order in the name and the actual shape order reversed! We might want to have a better look at what order is required for grand-challenge and fix the image5x6x7 image and ordering.

Regarding `resolution_levels` I just do not know, what its purpose is and what it should represent. Is this only for pyramid formats? If so my guess is that these aren't so common for MHD and MHA files. At least there is no default tag for this in the headers.

app/grandchallenge/cases/image_builders/metaio_utils.py
silvandeleemput: I will have a look at the functions. `parse_mh_headers` was implemented entirely by Paul though, and I have only moved, not touched, his function.

app/grandchallenge/cases/models.py
silvandeleemput: I just spoke with Paul, the ordering is indeed supposed to return `numpy` ordering using the `shape` function, otherwise these things would show the wrong way on grand-challenge right now.  The test-file indeed had inverted shape ordering, so I'll fix that and I'll write a test which asserts that the shape order is the inverse of the get_simplei_itk object retrieved from the model and document the ordering: `[(t), (z), y, x, (c)]`

app/grandchallenge/cases/image_builders/metaio_utils.py
jmsmkn: I know, but this is a good opportunity to do some cleaning, always good to leave the codebase in a better state than you found it ;-)

silvandeleemput: Ah, I have no problems with cleaning it, but I was a bit curious as to why it wasn't picked up before by code climate?

jmsmkn: We weren't running it back then! I'm going to integrate all of the useful checks from code climate in there to flake8 so that they're available in development too, and we can remove the dependency on a 3rd party service. There is another function in the codebase with a complexity of :100:, so, this is not the worst...

silvandeleemput:
The last commit fixes the coverage issue by covering all the exceptions for the 4D mh loader

app/grandchallenge/cases/image_builders/metaio_utils.py
silvandeleemput: That'll be very useful, right now it is kind of a pain to work with.

PR #966, opened 2019-11-06
Replace '/' with '-' in TRAVIS_BRANCH for docker tagging

MikeOverkamp-diag:
Without this change, the travis build fails when a '/' character is included in the branch name, as it is creates an invalid docker tag.

### DIFF
--- a/cycle_docker_compose.sh
+++ b/cycle_docker_compose.sh
@@ -5,7 +5,7 @@ echo "Press Ctrl+C (once) to stop"
 sleep 1
 
 export TRAVIS_BUILD_NUMBER=$(git describe --always --dirty)
-export TRAVIS_BRANCH=$(git rev-parse --abbrev-ref HEAD | sed "s/[^[a-zA-Z0-9]]//")
+export TRAVIS_BRANCH_SAFE=$(git rev-parse --abbrev-ref HEAD | sed "s/[^[a-zA-Z0-9]]//" | sed "s/[/]/-/")
 
 make -j2 build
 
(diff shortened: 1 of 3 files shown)


cycle_docker_compose.sh
```diff
@@ -6,6 +6,7 @@ sleep 1
 
 export TRAVIS_BUILD_NUMBER=$(git describe --always --dirty)
 export TRAVIS_BRANCH=$(git rev-parse --abbrev-ref HEAD | sed "s/[^[a-zA-Z0-9]]//")
+export TRAVIS_BRANCH_SAFE=$(echo $TRAVIS_BRANCH | sed "s/[/]/-/")
```
jmsmkn: I think that this:

```
export TRAVIS_BRANCH=$(git rev-parse --abbrev-ref HEAD | sed "s/[^[a-zA-Z0-9]]//")
export TRAVIS_BRANCH_SAFE=$(echo $TRAVIS_BRANCH | sed "s/[/]/-/")
```

Can be simplified to:
```
export TRAVIS_BRANCH_SAFE=$(git rev-parse --abbrev-ref HEAD | sed "s/[^[a-zA-Z0-9]]//")
```

As this is only used in development and `TRAVIS_BRANCH_SAFE` is now set in travis.yml

MikeOverkamp-diag: I have removed the `TRAVIS_BRANCH` variable, but kept the `s/[/]/-/` replacement as I believe it is still needed, even for development.

PR #967, opened 2019-11-06
Feature: Add csv exporter for reader study questions and answers

### LINKED ISSUE #961: Add a button to download answers / questions for reader studies in csv
The button should only be available for challenge editors.

MikeOverkamp-diag:
Closes #961

This will add two buttons to the reader study detail view: 

export_buttons

### DIFF
--- a/app/grandchallenge/reader_studies/models.py
+++ b/app/grandchallenge/reader_studies/models.py
@@ -472,6 +472,15 @@ class Question(UUIDModel):
         )
 
     @property
+    def csv_values(self):
+        return [
+            self.question_text,
+            self.get_answer_type_display(),
+            self.required,
+            f"{self.get_image_port_display() + ' port,' if self.image_port else ''}",
+        ]
+
+    @property
     def api_url(self):
         return reverse(
             "api:reader-studies-question-detail", kwargs={"pk": self.pk}
@@ -542,6 +551,14 @@ class Answer(UUIDModel):
         return reverse(
             "api:reader-studies-answer-detail", kwargs={"pk": self.pk}
         )
+
+    @property
+    def csv_values(self):
+        return self.question.csv_values + [
+            self.answer,
+            "; ".join(self.images.values_list("name", flat=True)),
+            self.creator.username,
+        ]
 
     def save(self, *args, **kwargs):
         adding = self._state.adding
--- a/app/grandchallenge/reader_studies/models.py
+++ b/app/grandchallenge/reader_studies/models.py
@@ -455,6 +455,8 @@ class Question(UUIDModel):
     )
     order = models.PositiveSmallIntegerField(default=100)
 
+    csv_headers = ["Question text", "Answer type", "Required", "Image port"]
+
     class Meta:
         ordering = ("order", "created")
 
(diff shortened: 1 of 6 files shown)


app/grandchallenge/reader_studies/views.py
```diff
@@ -236,13 +240,67 @@ class ReadersUpdate(ReaderStudyUserGroupUpdateMixin):
...
+
+    @action(detail=True)
+    def export_answers(self, request, pk=None):
+        reader_study = self.get_object()
+        self._check_export_perms(request.user, reader_study)
+
+        fields = ["creator__username", "question__question_text", "answer"]
+        data = Answer.objects.filter(
+            question__reader_study=reader_study
+        ).values_list(*fields)
```
jmsmkn: The `Images` field is missing here - this is required to know what images this answer is for. The best way to serialize this for the user here is to use the images `name` as this should be unique for the reader study.

app/tests/reader_studies_tests/test_api.py
```diff
@@ -294,3 +298,157 @@ def test_answer_is_correct_type(client, answer_type, answer, expected):
...
+    processed = exporter._preprocess_data(data)
+    assert len(processed) == lines
+
+    # Unfortunately, we have to create an actual file here, as both tempfile
+    # and StringIO seem to cause issues with line endings
+    with open("/tmp/csv.csv", "w+", newline="") as f:
+        writer = csv.writer(f)
+        writer.writerows(processed)
+
+    with open("/tmp/csv.csv", "r", newline="") as f:
```
jmsmkn: `/tmp/` should not be used directly - better to use the `tmp_path` fixture from pytest http://doc.pytest.org/en/latest/tmpdir.html#the-tmp-path-fixture

app/grandchallenge/reader_studies/views.py
```diff
@@ -236,13 +240,67 @@ class ReadersUpdate(ReaderStudyUserGroupUpdateMixin):
...
+            processed.append(
+                map(lambda x: re.sub(r"[\n\r\t]", " ", str(x)), entry)
+            )
+        return processed
+
+    def _create_csv_response(self, data, filename="export.csv"):
+        response = HttpResponse(content_type="text/csv")
+        response["Content-Disposition"] = f'attachment; filename="{filename}"'
+        writer = csv.writer(response, quoting=csv.QUOTE_ALL, escapechar="\\")
+        writer.writerows(self._preprocess_data(data))
```
jmsmkn: Could we also add a header row?

app/grandchallenge/reader_studies/views.py
MikeOverkamp-diag: Added and moved all fields to the answers export

app/tests/reader_studies_tests/test_api.py
MikeOverkamp-diag: Changed

app/grandchallenge/reader_studies/views.py
MikeOverkamp-diag: Added

app/grandchallenge/reader_studies/models.py
```diff
@@ -455,6 +455,8 @@ class Question(UUIDModel):
     )
     order = models.PositiveSmallIntegerField(default=100)
 
+    csv_headers = ["Question text", "Answer type", "Required", "Image port"]
+
```
jmsmkn: Last question: would it be better to use an OrderedDict for the csv headers (keys) and corresponding values? I'm worried that these 2 lists could get out of sync.

MikeOverkamp-diag: Good point, I'll update it.

PR #970, opened 2019-11-06
Fix up the docstrings

jmsmkn:
The docstrings in the project are a bit of a mess, this adds checks to ensure that they're consistent.


jmsmkn:
Wait for #963 to land on master first as there will be conflicts.

Issue #974, opened 2019-11-07
Automatically generate the hanging_protocol_list from the set of images in a reader study


jmsmkn:
Right now a reader study admin must manually create the hanging protocol list. However, if they do not care about the ordering or the arrangement (eg, they only want 1 image in the main port in each hanging) then it would be nice to have a button that will automatically generate this hanging protocol list for them and store it in the reader study config.

PR #979, opened 2019-11-08
Add button to auto generate hanging list for reader studies

MikeOverkamp-diag:
Closes https://github.com/comic/grand-challenge.org/issues/974

### DIFF
--- a/app/grandchallenge/reader_studies/models.py
+++ b/app/grandchallenge/reader_studies/models.py
@@ -265,10 +265,8 @@ class ReaderStudy(UUIDModel, TitleSlugDescriptionModel):
         return hanging_list
 
     def generate_hanging_list(self):
-        hanging_list = []
-        for image_name in self.images.values_list("name", flat=True):
-            hanging_list.append({"main": image_name})
-        self.hanging_list = hanging_list
+        image_names = self.images.values_list("name", flat=True)
+        self.hanging_list = [{"main": name} for name in image_names]
         self.save()
 
 
--- a/app/grandchallenge/reader_studies/models.py
+++ b/app/grandchallenge/reader_studies/models.py
@@ -264,6 +264,13 @@ class ReaderStudy(UUIDModel, TitleSlugDescriptionModel):
 
         return hanging_list
 
+    def generate_hanging_list(self):
+        hanging_list = []
+        for image_name in self.images.values_list("name", flat=True):
+            hanging_list.append({"main": image_name})
+        self.hanging_list = hanging_list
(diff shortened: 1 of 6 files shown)


app/grandchallenge/reader_studies/models.py
```diff
@@ -264,6 +264,13 @@ def get_hanging_list_images_for_user(self, *, user):
 
         return hanging_list
 
+    def generate_hanging_list(self):
+        hanging_list = []
+        for image_name in self.images.values_list("name", flat=True):
+            hanging_list.append({"main": image_name})
+        self.hanging_list = hanging_list
```
jmsmkn: It's quicker to make this a list comprehension than by using append, eg:

```
image_names = self.images.values_list("name", flat=True)
self.hanging_list = [{"main": name} for name in image_names]
```

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
```diff
@@ -116,6 +116,15 @@ <h2>Images</h2>
             </a>
         </p>
 
+        <h2>Hanging list</h2>
+
+        <p class="mt-3">
+            <a class="btn btn-primary"
```
jmsmkn: I think that we should give the users a warning that this will destroy their current hanging list. Maybe we should only allow/show this if the current reader study is not valid?

app/grandchallenge/reader_studies/views.py
```diff
@@ -335,6 +337,14 @@ def export_answers(self, request, pk=None):
             filename=f"{reader_study.slug}-answers.csv",
         )
 
+    @action(detail=True)
```
jmsmkn: Should this be a POST request as the reader_study is mutated?

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
MikeOverkamp-diag: Good point, I also considered that. I think a warning would probably be sufficient. That way we give the user the freedom to regenerate the list if they would like to do so.

app/grandchallenge/reader_studies/views.py
MikeOverkamp-diag: Yes, definitely.

MikeOverkamp-diag:
Great! I had quite a lot of issues with the permissions, but this is a much cleaner solution than mine and it's good to know of its existence now.

PR #1005, opened 2019-11-25
Add option to request viewing rights for algorithms

### LINKED ISSUE #968: "Request access" button for algorithms
Accessing a public algorithm, without access, results in an uninformative forbidden page. The only option now is to email the owner of the algorithm.

1. Users should have a way for requesting access to an algorithm. For example, a "request access" button.
2. Public algorithms should have a more user-friendly error page when you don't have access (yet). This page could include a request button.

### LINKED ISSUE #1002: Add ability to revoke access to algorithm
It is currently not possible for the editors of an algorithm to remove users from the user group of the algorithm. So I can give people access, but I cannot undo that. Should be possible, in my opinion. Same for editors.

MikeOverkamp-diag:
Closes #968 closes #1002


MikeOverkamp-diag:
Closes #1002

app/grandchallenge/algorithms/emails.py
```diff
@@ -0,0 +1,70 @@
...
+            [editor.email],
+        )
+
+
+def send_permission_granted_email(obj):
+    """
+    Email the requester that their request has been approved.
+
+    obj:         AlgorithmPermissionRequest object containing info on which
+                 user requested access to which algorithm.
```
jmsmkn: Could you set your IDE to generate numpy style docstrings? Thanks!

app/grandchallenge/participants/models.py
```diff
@@ -4,7 +4,7 @@
 from grandchallenge.challenges.models import Challenge
 
 
-class RegistrationRequest(models.Model):
+class RequestBase(models.Model):
```
jmsmkn: Could this be moved to the `core` application to avoid any cross-app dependencies?

app/grandchallenge/algorithms/views.py
```diff
@@ -321,3 +356,119 @@ class ResultViewSet(ReadOnlyModelViewSet):
...
+        queryset = queryset.filter(algorithm=self.algorithm)
+        return queryset
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context.update({"algorithm": self.algorithm})
+        return context
+
+
+class AlgorithmPermissionRequestUpdate(SuccessMessageMixin, UpdateView):
```
jmsmkn: Is this missing permissions? Should only the algorithm editors be able to view this?

app/grandchallenge/algorithms/emails.py
MikeOverkamp-diag: Sure, updated :slightly_smiling_face:

app/grandchallenge/algorithms/views.py
MikeOverkamp-diag: No, this is also used if the user's request has been rejected. In that case they can request permission again, changing the status back to `pending`. If the user does not have editing rights, the value for `status` is always set to `pending` if the requesting user is not already a user or an editor for the algorithm.

PR #1008, opened 2019-11-27
537 retry uploads

### LINKED ISSUE #537: Chunked uploads not attempting retries
I forgot to implement the actual retry-portion for the jQuery file uploading. As such, jQuery file upload will only try to upload a given chunk once. See here:

https://github.com/blueimp/jQuery-File-Upload/wiki/Chunked-file-uploads

MikeOverkamp-diag:
Closes #537

### DIFF
--- a/app/grandchallenge/jqfileupload/views.py
+++ b/app/grandchallenge/jqfileupload/views.py
@@ -1,7 +1,6 @@
 import re
 from datetime import timedelta
 
-from django.db.models import Max
 from django.utils.timezone import now
 from rest_framework import mixins
 from rest_framework.decorators import action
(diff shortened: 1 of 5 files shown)


app/grandchallenge/jqfileupload/views.py
```diff
@@ -93,3 +94,11 @@ def _handle_file(self, uploaded_file):
             "total_size": total_size,
             "user_pk_str": self.user_pk_str,
         }
+
+    @action(detail=False, methods=["get"])
+    def get_current_file_size(self, request):
+        client_id = request.GET.get("file", None)
+        files = StagedFile.objects.filter(client_id=client_id).aggregate(
+            Max("end_byte")
+        )
```
jmsmkn: Is this correct? What happens if the chunks arrive out of order?

MikeOverkamp-diag: Good point, I guess we also need to check for any gaps in the byte sequence and return the `end_byte` for the largest uninterrupted sequence of chunks.

PR #1011, opened 2019-11-28
Save pixel sizes for tiff files.

### LINKED ISSUE #1010: Add voxel size of uploaded images to GC database
When tif files are uploaded to GC a dzi file is created, which is used to view the image. The headers of the tif file contain the resolution that can be used to calculate pixel size. Because a dzi can't store pixel size, and the pixel size is needed to accurately calculate distance measurements, this information needs to be stored in the database.

miriam-groeneveld:
Closes #1010


app/grandchallenge/cases/image_builders/metaio_mhd_mha.py
```diff
@@ -97,6 +97,8 @@ def convert_itk_file(
                 timepoints=timepoints,
                 resolution_levels=None,
                 color_space=color_space,
+                pixel_width=None,
+                pixel_height=None,
```
jmsmkn: Would be useful to have this information for mhd too. This should be something like:

```
                pixel_width=simple_itk_image.GetSpacing()[0],
                pixel_height=simple_itk_image.GetSpacing()[1],
```

app/grandchallenge/cases/image_builders/tiff.py
```diff
@@ -57,56 +85,54 @@ def _validate_tifffile(  # noqa: C901
...
     if Image.COLOR_SPACE_COMPONENTS[color_space] != tif_color_channels:
         raise ValidationError("Image contains invalid amount of channels.")
 
-    try:
-        image_width = tags["ImageWidth"].value
-        image_height = tags["ImageLength"].value
-    except KeyError:
-        raise ValidationError("Missing tags in tiff file")
+    pixel_width = calculate_pixel_size(tags, "XResolution")
+    pixel_height = calculate_pixel_size(tags, "YResolution")
```
jmsmkn: Do you also have ZResolution in tiff (eg, for the thickness of the slide)

app/tests/cases_tests/test_tiff.py
```diff
@@ -146,6 +159,8 @@ def test_tiff_image_entry_creation(resource, expected_error_message):
         assert image_entry.color_space == get_color_space(
             str(tiff_tags["PhotometricInterpretation"].value)
         )
+        assert image_entry.pixel_height == pixel_height
+        assert image_entry.pixel_width == pixel_width
```
jmsmkn: Probably best to use https://docs.pytest.org/en/latest/reference.html#pytest-approx here

app/grandchallenge/cases/image_builders/tiff.py
miriam-groeneveld: Not that I found, I'll check with Geert

miriam-groeneveld: Nope, defaults to 1

PR #1020, opened 2019-12-04
Sub src/dest links in challenge copy command

### LINKED ISSUE #991: Replace links in the challenge copying command
People might use full url links in the page HTML, use a regex to replace the links using the new challenge short name.

MikeOverkamp-diag:
This checks for any `<a href>` tags in the src challenge's pages' html that link to the src challenge and replaces them with links to the same pages in the dest challenge.

Closes #991

### DIFF
--- a/app/grandchallenge/challenges/management/commands/copy_challenge.py
+++ b/app/grandchallenge/challenges/management/commands/copy_challenge.py
@@ -1,3 +1,6 @@
+import re
+
+from django.contrib.sites.models import Site
 from django.core.management import BaseCommand, CommandError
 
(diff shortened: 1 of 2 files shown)


MikeOverkamp-diag:
Black fails in the pre-commit for some reason, but locally it will not refactor any files for me.

jmsmkn:
> Black fails in the pre-commit for some reason, but locally it will not refactor any files for me.

It looks like pyupgrade makes a change, then black tries to format that. Do you have all of the pre-commit hooks installed? Try running `pre-commit run --all-files`, copy_challenge.py should be changed by that and then you can commit the result.

PR #1022, opened 2019-12-05
Update dependencies

jmsmkn:
This updates all of the dependencies and fixes the docs generation. Python-dateutil version was pinned due to a tree resolution bug in poetry.


jmsmkn:
Ready for review, but do not merge until the pending migrations have been applied.

PR #1034, opened 2019-12-09
Fix cirrus in development

jmsmkn:


jmsmkn:
The CI tests are broken, workaround in #1035

PR #1036, opened 2019-12-09
Use standardized format for all emails

MikeOverkamp-diag:


MikeOverkamp-diag:
> There are errors with sending emails now, might be my fault as I was guessing the API when I sent the mail, sorry!
> 
> ```
> File "/app/grandchallenge/challenges/emails.py", line 13, in send_challenge_created_email
>     f"Dear {{}},\n\n"
> AttributeError: 'Site' object has no attribute 'url'
> ```

No, I should have just checked before committing. Should be fixed now.

PR #1037, opened 2019-12-09
Add views for workstation configs

MikeOverkamp-diag:


app/grandchallenge/workstation_configs/forms.py
```diff
@@ -0,0 +1,10 @@
+from django.forms import ModelForm
+
+from grandchallenge.core.forms import SaveFormInitMixin
+from grandchallenge.workstation_configs.models import WorkstationConfig
+
+
+class WorkstationConfigForm(SaveFormInitMixin, ModelForm):
+    class Meta:
+        model = WorkstationConfig
+        exclude = ("creator",)
```
jmsmkn: Nitpick: I prefer to use `include` in case we add another field that should stay private, then we're explicit about what fields are included in the form.

app/grandchallenge/workstation_configs/views.py
```diff
@@ -11,3 +26,52 @@ class WorkstationConfigViewSet(ReadOnlyModelViewSet):
...
+class WorkstationConfigDetail(LoginRequiredMixin, DetailView):
+    model = WorkstationConfig
+
+
+class WorkstationConfigUpdate(
+    LoginRequiredMixin, ObjectPermissionRequiredMixin, UpdateView
+):
+    model = WorkstationConfig
+    form_class = WorkstationConfigForm
+    permission_required = f"change_{WorkstationConfig._meta.model_name}"
```
jmsmkn: Question: is the app label required? eg,
```suggestion
    permission_required = f"{WorkstationConfig._meta.app_label}.change_{WorkstationConfig._meta.model_name}"
```

app/grandchallenge/workstation_configs/templates/workstation_configs/workstationconfig_confirm_delete.html
```diff
@@ -0,0 +1,21 @@
+{% extends "base.html" %}
+
```
jmsmkn: Breadcrumbs are missing

app/grandchallenge/workstation_configs/templates/workstation_configs/workstationconfig_detail.html
```diff
@@ -0,0 +1,84 @@
...
+    <div class="my-3">
+        <div class="d-flex justify-content-between">
+            <div class="font-weight-bold">Window presets</div>
+            <div>
+                {% for x in object.window_presets.all %}
+                    <div>{{ preset }}</div>
+                {% endfor %}
+            </div>
+        </div>
+        <div class="d-flex justify-content-between">
```
jmsmkn: The separation between key and value makes this hard to read:

Screenshot_20191210_145428

app/grandchallenge/workstation_configs/views.py
MikeOverkamp-diag: I think it also works without it, but it is of course cleaner to include it.

PR #1039, opened 2019-12-10
Change LandmarkAnnotationSet endpoint and add custom action

### LINKED ISSUE #1032: Move LandmarkAnnotation API endpoint to core API and remove user_id dependency
The LandmarkAnnotationSet API endpoint will only be used by the CIRRUS workstation so it can be moved to the core API (`/api/v1/`). 
Also, it will at first only be used by the graders so it is not necessary to allow `retina_admin`s to access the annotations of other graders so the dependency on `user_id` in the URL should be removed and permission class should change.

HarmvZ:
- Changes endpoint url from `/retina/api/...` to `/api/v1/retina/landmark-annotation/`
- Adds a custom action on the viewset for retrieving all annotations for a specific image provided in the querystring

Closes #1032

### DIFF
--- a/app/grandchallenge/retina_api/views.py
+++ b/app/grandchallenge/retina_api/views.py
@@ -976,7 +976,7 @@ class ImageTextAnnotationViewSet(viewsets.ModelViewSet):
 
 
 class LandmarkAnnotationSetViewSet(viewsets.ModelViewSet):
-    permission_classes = (RetinaOwnerAPIPermission,)
+    permission_classes = (RetinaAPIPermission,)
     authentication_classes = (authentication.TokenAuthentication,)
     serializer_class = LandmarkAnnotationSetSerializer
     filter_backends = (filters.ObjectPermissionsFilter,)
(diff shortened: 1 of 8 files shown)


app/grandchallenge/retina_api/views.py
```diff
@@ -976,9 +978,38 @@ class ImageTextAnnotationViewSet(viewsets.ModelViewSet):
...
+        :param request: Request
+        :return: Response
+        """
+        image_id = self.request.query_params.get("image_id")
+        image = get_object_or_404(Image.objects.all(), pk=image_id)
+        queryset = self.filter_queryset(self.get_queryset()).filter(
+            singlelandmarkannotation__image=image
+        )
+        serializer = self.get_serializer(queryset, many=True)
+        return Response(serializer.data)
```
jmsmkn: Should this be done using a filter on the list view? https://www.django-rest-framework.org/api-guide/filtering/

HarmvZ: Good point. I moved the functionality to the list view.

HarmvZ:
Refactored filtering:
 - Created filter backend for filtering annotations on owner or `retina_admin`
 - Filtered on query parameter in `get_queryset()` method
 - Added `prefetch_related` to improve performance since related `SingleLandmarkAnnotation` objects are used during serialization

PR #1041, opened 2019-12-13
Create policies app

### LINKED ISSUE #981: Terms and conditions for algorithms on upload page
Technically, if grand challenge gets a Terms & Conditions page (see #980), algorithms itself do not need separate terms and conditions if they are covered by the site-wide one.

However, I think it would be still good to make people aware of this when submitting data to an algorithm. Especially when we want to use uploaded data for future research, it has to be crystal clear to users that we can use this data if they upload it.

A minimal version of this could be a single sentence like "By uploading data to this algorithm you agree with the T&C as defined [here]". It could be even beter if algorithms could define additional terms on top, but that is a nice to have.

MikeOverkamp-diag:
Closes #981


jmsmkn:
The tests aren't broken due to changes in this PR, something else has broken: #1042

app/grandchallenge/core/templates/terms_of_service.html
```diff
@@ -0,0 +1,11 @@
...
+
+{% block title %}
+    Terms of service - {{ block.super }}
+{% endblock %}
+
+{% block content %}
+    <h1>Terms of service</h1>
+
+    <p>{{ terms.body }}</p>
+{% endblock %}
```
jmsmkn:
```suggestion
{% endblock %}

```

app/grandchallenge/profiles/urls.py
```diff
@@ -19,6 +20,7 @@
         {"signup_form": SignupFormExtra},
         name="profile_signup",
     ),
+    path("signup-social", PreSocialView.as_view(), name="pre-social"),
```
jmsmkn:
```suggestion
    path("signup-social/", PreSocialView.as_view(), name="pre-social"),
```

app/grandchallenge/core/templates/terms_of_service.html
```diff
@@ -0,0 +1,11 @@
+{% extends "base.html" %}
+
+{% block title %}
+    Terms of service - {{ block.super }}
```
jmsmkn:
```suggestion
    Terms of Service - {{ block.super }}
```

app/grandchallenge/core/templates/terms_of_service.html
```diff
@@ -0,0 +1,11 @@
...
+{% endblock %}
+
+{% block content %}
+    <h1>Terms of service</h1>
```
jmsmkn:
```suggestion
    <h1>Terms of Service</h1>
```

app/grandchallenge/core/templates/terms_of_service.html
```diff
@@ -0,0 +1,11 @@
...
+
+    <p>{{ terms.body }}</p>
```
jmsmkn:
```suggestion
    <p>{{ terms.body|safe }}</p>
```

PR #1043, opened 2019-12-13
Pin poetry to previous version

### LINKED ISSUE #1042: Builds are broken
It seems that the build is not reproducible right now, I retried master that was successful yesterday, fails today: https://travis-ci.org/comic/grand-challenge.org/builds/624157575 compare with https://travis-ci.org/comic/grand-challenge.org/builds/623723638

MikeOverkamp-diag:
Version 1.0.0 seems to break the build.

Closes #1042

### DIFF
--- a/dockerfiles/web/Dockerfile
+++ b/dockerfiles/web/Dockerfile
@@ -20,7 +20,7 @@ ENV PYTHONUNBUFFERED 1
 
 RUN mkdir -p /opt/poetry /app /static /opt/static /dbox/Dropbox/media
 RUN python -m pip install -U pip
-RUN python -m pip install -U poetry==1.0.0b9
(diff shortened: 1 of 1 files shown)


dockerfiles/web/Dockerfile
```diff
@@ -20,7 +20,7 @@ ENV PYTHONUNBUFFERED 1
 
 RUN mkdir -p /opt/poetry /app /static /opt/static /dbox/Dropbox/media
 RUN python -m pip install -U pip
-RUN python -m pip install -U poetry
+RUN python -m pip install -U poetry==0.12.7
```
jmsmkn: There is a more recent version:
```suggestion
RUN python -m pip install -U poetry==0.12.17
```

PR #1045, opened 2019-12-13
Upgrade poetry and get virtualenv path from poetry

### LINKED ISSUE #1044: Upgrade poetry to 1.0.0
Using poetry 1.0.0 broke the build, so as an intermediate solution we now use version 0.12.17. We should make the necessary changes to be able to use version 1.0.0.

MikeOverkamp-diag:
Closes #1044

### DIFF
--- a/dockerfiles/web/Dockerfile
+++ b/dockerfiles/web/Dockerfile
@@ -20,7 +20,7 @@ ENV PYTHONUNBUFFERED 1
 
 RUN mkdir -p /opt/poetry /app /static /opt/static /dbox/Dropbox/media
 RUN python -m pip install -U pip
-RUN python -m pip install -U poetry==0.12.17
+RUN python -m pip install -U poetry
 
 RUN groupadd -g 2001 -r django && useradd -m -u 2001 -r -g django django
 RUN chown django:django /opt/poetry /app /static /opt/static /dbox/Dropbox/media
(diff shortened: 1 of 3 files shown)


MikeOverkamp-diag:
> This doesn't seem to work, poetry 1.0.0 is installed but I don't think that you can use bash commands in an `ENV` block. You might be able to use `RUN export ...` instead.

Sorry, I should have updated the PR when I realized it does not work this way. The `RUN` executive actually does not work either, as the env variable is not persistent if you set it that way. For now, creating a symlink from the virtualenv's bin directory to some other directory and then adding that to the `PATH` seems to be the most promising solution.

dockerfiles/web/Dockerfile
```diff
@@ -30,8 +30,9 @@ USER django:django
 WORKDIR /opt/poetry
 COPY pyproject.toml /opt/poetry
 COPY poetry.lock /opt/poetry
+RUN poetry config --local virtualenvs.in-project true
```
jmsmkn: Does the `--local` argument modify `pyproject.toml`? Could we add the configuration there instead. Would then need to add the `.venv` directory to `.gitignore` if it is not there already.

MikeOverkamp-diag: Apparently it needs to by in a separate `poetry.toml` file: https://github.com/python-poetry/poetry/pull/1272, which I have added now.

PR #1047, opened 2019-12-16
Add search field to Token admin

HarmvZ:
There are currently 15 pages of tokens in production and no way to search through them which is inconvenient. This adds a search field that allows an admin to search for username, email, fist or last name of a user in the admin. 

**WARNING!** This requires a change in the order of the `INSTALLED_APPS` setting for correctly unregistering and registering the admin. I'm not sure if that will break other things.

### DIFF
--- a/app/config/settings.py
+++ b/app/config/settings.py
@@ -360,7 +360,7 @@ LOCAL_APPS = [
     "grandchallenge.workstation_configs",
 ]
 
-INSTALLED_APPS = DJANGO_APPS + LOCAL_APPS + THIRD_PARTY_APPS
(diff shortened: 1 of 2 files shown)


jmsmkn:
> **WARNING!** This requires a change in the order of the `INSTALLED_APPS` setting for correctly unregistering and registering the admin. I'm not sure if that will break other things.

I think that it does... There might be a better way to do it - maybe add the token as a related field to the User model admin? See the group relationship in that same `admin.py`

HarmvZ:
Ok, I just tried to do that quickly but it was not working. I will make a ticket for this and do it some other time since it has no high priority.

PR #1048, opened 2019-12-16
Made api::schema-json and api::schema-docs public for safe rest-api methods

### LINKED ISSUE #1017: Create a simple interface for fetching datatypes schemas on grand-challenge
**Problem**
The grand-challenge datatypes are currently only stored in the [gc-api](https://github.com/DIAGNijmegen/rse-gcapi/tree/master/gcapi/schemas) repository. However, the information is required by other libraries as well. Duplication of this information seems bad.

**Solution**
* [x] It would be nice to have this information in a central location like grand-challenge and provide a simple GET interface to allow the libraries /comic/evalutils and /DIAGNijmegen/rse-gcapi/ to fetch and cache this information (only a few kbs) from grand-challenge.
* [x] The answer type schemas should be added to the generated schema

silvandeleemput:
#1017

* Note several tests with method=OPTIONS and shema="schema-docs" fail and have now been added as expected failures

### DIFF
--- a/app/grandchallenge/api/urls.py
+++ b/app/grandchallenge/api/urls.py
@@ -68,9 +68,7 @@ router.register(
 
 
 class IsSuperUserOrReadOnly(permissions.BasePermission):
-    """
-    The request is authenticated as an admin, or is a read-only request.
-    """
+    """The request is authenticated as an admin, or is a read-only request."""
 
     def has_permission(self, request, view):
         return bool(
--- a/app/grandchallenge/api/urls.py
+++ b/app/grandchallenge/api/urls.py
@@ -67,17 +67,6 @@ router.register(
 )
 
 
-class IsSuperUserOrReadOnly(permissions.BasePermission):
-    """The request is authenticated as an admin, or is a read-only request."""
-
-    def has_permission(self, request, view):
-        return bool(
-            request.method in permissions.SAFE_METHODS
-            or request.user
-            and request.user.is_superuser
-        )
-
-
 # TODO: add terms_of_service and contact
 schema_view = get_schema_view(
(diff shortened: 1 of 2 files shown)


app/grandchallenge/api/urls.py
```diff
@@ -74,7 +86,7 @@
         description=f"The API for {settings.SESSION_COOKIE_DOMAIN.lstrip('.')}.",
         license=openapi.License(name="Apache License 2.0"),
     ),
-    permission_classes=(permissions.IsAuthenticated,),
+    permission_classes=(IsSuperUserOrReadOnly,),
```
jmsmkn: I'm not sure why we would ever want to POST or PATCH to this view. All of this can probably be replaced with
```suggestion
    permission_classes=(permissions.AllowAny,),
    public=True,
```

See https://drf-yasg.readthedocs.io/en/stable/readme.html#quickstart

silvandeleemput: That will also work, I just wanted to enforce read-only permissions for all users for safety if it ever changes in the future.

jmsmkn: I think it would be fairly safe, the API for this method is clear and follows a good deprecation cycle, there's no DB table for this app to even edit here.

silvandeleemput: Permissions on the view are now AllowAny

jmsmkn:
Note that this PR does not close #1017, as the answer type schemas are not added to the generated schema so needs to be extended, see https://drf-yasg.readthedocs.io/en/stable/custom_spec.html

silvandeleemput:
The changes LGTM

PR #1056, opened 2019-12-18
Feature/902 append errors

### LINKED ISSUE #902: MHD import errors are clobbered by TIFF import errors
**Describe the bug**
If a bad MHD file is uploaded the importer will report an error and then pass the file to the tiff importer, which will then fail too and overwrite the error message, confusing the user. eg:

image

These messages should be appended, rather than replaced.

**To Reproduce**
Upload a file that is not a MHD or TIFF

MikeOverkamp-diag:
Closes #902


app/grandchallenge/cases/image_builders/tiff.py
```diff
@@ -190,13 +190,13 @@ def image_builder_tiff(path: Path) -> ImageBuilderResult:
...
         except ValidationError as e:
-            invalid_file_errors[file_path.name] = str(e)
+            invalid_file_errors[file_path.name] = e.message  # noqa: B306
             continue
 
         try:
             dzi_output = create_dzi_images(tiff_file=tiff_file, pk=pk)
         except ValidationError as e:
-            invalid_file_errors[file_path.name] = str(e)
+            invalid_file_errors[file_path.name] = e.message  # noqa: B306
```
jmsmkn: Why `e.message` and not `str(e)`?

MikeOverkamp-diag: for `ValidationError`s `str(e)` creates a string from a list, e.g. `[""Image isn't a TIFF file"]`, whereas `e.message` just contains the string value.

MikeOverkamp-diag: This only seems to hold for ValidationErrors by the way, for standard python exceptions `str(e)` works fine and is preferred over e.message, which has been deprecated since python2.7 (hence the `#noqa`).

jmsmkn: Is `message` guaranteed to be set? It doesn't look like it to me: https://docs.djangoproject.com/en/3.0/_modules/django/core/exceptions/#ValidationError

Would it be better to iterate over `ValidationError.messages` with a `join`?

MikeOverkamp-diag: In this case it is actually guaranteed to be set, as all the ValidationErrors that can be caught are set within the `tiff.py` file and all are called with a string as the `message` argument, in which case the `message` property is always set.

PR #1066, opened 2020-01-07
Adding exposing enum support for rest_framework.CharField through swagger_schemas_fields

silvandeleemput:
This PR changes the default behavior of swagger_schemas_fields with update instead of overwriting #1017

* Created a swagger utility for creating char field swagger fields
  * This adds additional enum information to the generated schemas
* Added this to the serializers for:
  * reader_studies: answer_type, form_direction, image_port
  * workstation_configs: default_orientation
  * workstations: status

### DIFF
--- a/app/grandchallenge/api/swagger.py
+++ b/app/grandchallenge/api/swagger.py
@@ -1,0 +1,45 @@
+from typing import Dict, List, Tuple, Union
+
+from drf_yasg.inspectors import FieldInspector
+from drf_yasg.openapi import Schema
+
+
+def add_manual_fields(self, serializer_or_field, schema):
+    meta = getattr(serializer_or_field, "Meta", None)
+    swagger_schema_fields = getattr(meta, "swagger_schema_fields", {})
+    if swagger_schema_fields:
+        for attr, val in swagger_schema_fields.items():
+            if isinstance(val, dict) and isinstance(
+                getattr(schema, attr, None), dict
+            ):
+                to_update = dict(
+                    list(getattr(schema, attr).items()) + list(val.items())
+                )
+                setattr(schema, attr, to_update)
+            else:
+                setattr(schema, attr, val)
+
+
+# This overrides the default behavior when the swagger_schema_fields attribute is set
+# in the serializers. Now dictionary attributes are updated instead of overwritten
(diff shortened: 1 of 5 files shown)


app/grandchallenge/reader_studies/serializers.py
```diff
@@ -31,6 +32,14 @@ class Meta:
             "reader_study",
             "required",
         )
+        swagger_schema_fields = swagger_schema_fields_for_charfield(
+            ("answer_type", "form_direction", "image_port"),
+            (
+                Question.ANSWER_TYPE_CHOICES,
+                Question.DIRECTION_CHOICES,
+                Question.IMAGE_PORT_CHOICES,
+            ),
```
jmsmkn: This syntax could be improved, always better to use keyword only arguments and if your passing data together it is better to use a dictionary rather than 2 lists that can get out of step. Could you do something like:

```
swagger_schema_fields_for_charfield(fields=["answer_type", "form_direction", "image_port"], model=Question)
```

and then within that function:

```
def swagger_schema_fields_for_charfield(*, model, fields):
   ...
   fields_with_choices = {f: model._meta.field(f).choices for f in fields}
```

If you can find a way to do this without passing the model (DRY) then even better.

app/grandchallenge/api/swagger.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def add_manual_fields(self, serializer_or_field, schema):
+    meta = getattr(serializer_or_field, "Meta", None)
+    swagger_schema_fields = getattr(meta, "swagger_schema_fields", {})
+    if swagger_schema_fields:
+        update_swagger_schema_fields(schema, swagger_schema_fields)
+
+
+def update_swagger_schema_fields(schema, swagger_schema_fields):
```
jmsmkn: These should probably be protected functions, prepend a `_` to both.

silvandeleemput: OK.

app/grandchallenge/reader_studies/serializers.py
silvandeleemput: These have been changed as discussed during our meeting.

PR #1075, opened 2020-01-08
DRF-ed RawImageUploadSession and RawImageFile with custom permission

haimasree:
- RawImageUploadSession is serialized with custom DjangoObjectLevelPermission
- RawImageFile is serialized with custom DjangoObjectLevelPermission
- A custom implmenetation of Patch for RawImageUploadSession


app/grandchallenge/cases/serializers.py
```diff
@@ -53,3 +68,23 @@ class Meta:
...
+class RawImageFileSerializer(serializers.ModelSerializer):
+    upload_session = HyperlinkedRelatedField(
+        queryset=RawImageUploadSession.objects.all(),
+        view_name="api:upload-session-detail",
+    )
+
+    def validate(self, attrs):
+        upload_session = attrs["upload_session"]
+        user = self.context.get("request").user
+        if not user.has_perm("view_rawimageuploadsession", upload_session):
```
jmsmkn: This modifies the raw image upload session, so we want to check for `change_rawimageuploadsession`.

app/grandchallenge/cases/views.py
```diff
@@ -94,6 +101,26 @@ class RawImageUploadSessionViewSet(
...
+    permission_classes = [DjangoObjectOnlyWithCustomPostPermissions]
+    filter_backends = [ObjectPermissionsFilter]
 
     def perform_create(self, serializer):
         serializer.save(creator=self.request.user)
+
+    @action(detail=True, methods=["patch"])
+    def process_images(self, request, pk=None):
+        upload_session = self.get_object()
+        upload_session.process_images()
```
jmsmkn: We should only schedule this job if there are unprocessed images, and if the `RawImageUploadSession.session_state` is `UploadSessionState.stopped`, otherwise users could flood the system with jobs.

haimasree: @jmsmkn 
Isnt checking uplaod_session.session_state == UploadSessionState.stopped enough? We can add a field to RawImageFile model which will be marked processed? 
```
im.state == "processed" for all im in RawImageFile.objects.filter(upload_session=upload_session).all()
```

jmsmkn: Ok, yes, it turns out that state is never saved. We need a migration for that, but that can be another issue.

PR #1082, opened 2020-01-09
Fixes answer type annotation schema checking bug and added tests

### LINKED ISSUE #1078: It is possible to assign other annotations to annotation answers
The validation for annotation answers will accept any annotation, so it is possible to set say, a bounding box for a distance line answer. Discovered when reading code with @silvandeleemput yesterday.

silvandeleemput:
This PR closes  #1078

* Altered generic schema to include string and bool types
* Made a filter method which creates a copy of the schema but which checks for a specific answer type


jmsmkn:
Restarting the PR build

app/grandchallenge/reader_studies/models.py
```diff
@@ -445,10 +442,41 @@ def delete_reader_study_groups_hook(*_, instance: ReaderStudy, using, **__):
...
     ],
 }
 
 
+def _filter_answer_type_annotation_schema(
+    *, type_name: str
+) -> Union[bool, Dict]:
+    """Yields a schema filtered for a specific answer type annotation."""
+    if "TXT" in type_name:
+        type_name = "string"
```
jmsmkn: I think that I can see a cleaner way to do this, I will make these changes

Issue #1088, opened 2020-01-13
RawImageUploadSessionViewSet.process_images should return an error if the job is not scheduled


jmsmkn:
https://github.com/comic/grand-challenge.org/blob/master/app/grandchallenge/cases/views.py#L113-L118

This function will always return an OK status, even if the `if` statement fails. There should be a new message here and `HttpResponseBadRequest` if `upload_session.session_state != UploadSessionState.stopped`

Issue #1087, opened 2020-01-13
Add a field indicating whether a RawImageFile object has been consumed by corresponding RawImageUploadSession


haimasree:
Currently there is no way to check if there is an unconsumed RawImageFile object attached with a RawImageUploadSession. Also, update Patch method of RawImageUploadSession to check whether there are unprocessed images in addition to checking the state of the session.

PR #1090, opened 2020-01-13
Add a state field to RawImageFile and improved messaging for partial update of RawImageUploadSession API

haimasree:
Closes: #1087 
Closes: #1088


jmsmkn:
I think that the `RawImageUploadSession` stores the job state, so this information does not need to be duplicated. What we really need to know is if the `RawImageFile` has either been successfully consumed into an `Image` or not, so a boolean field with a default of `False` is what I think we need.

jmsmkn:
It might be worth combining this with #1089 into 1 PR as they're closely related.

app/tests/cases_tests/test_api.py
```diff
@@ -295,21 +295,55 @@ def test_image_file_post_permissions(client, is_active, expected_response):
...
+
+    response = get_view_for_user(
+        viewname="api:upload-session-process-images",
+        reverse_kwargs={"pk": us.pk},
+        user=user,
+        client=client,
+        method=client.patch,
+        content_type="application/json",
+    )
+    assert response.status_code == 400
```
jmsmkn: This test is just testing if the post save task on `us.save()` works, which we know it does as it is tested elsewhere. What are you intending to test here? Maybe it was restarting the job, in which case 

```
    us.refresh_from_db()
    us.session_state = "created"
    us.save()
```

needs to happen before

```
    ri.refresh_from_db()
    ri.consumed = False
    ri.save()
```

and then 

```
```

as the job will be restarted in this case.

haimasree: @jmsmkn 
This gives me a 400 response. Why should it be 200? For that, us.session_state must be "stopped". Isn't it?

jmsmkn: I don't understand what you're trying to do here then, what is the intention of this test?

haimasree: I guess the last test is superfluous. I will remove it.

PR #1094, opened 2020-01-15
Add ground truth to reader study questions

MikeOverkamp-diag:


app/grandchallenge/reader_studies/models.py
```diff
@@ -657,6 +680,13 @@ def save(self, *args, **kwargs):
 
         super().save(*args, **kwargs)
 
+        if self.is_ground_truth:
+            Answer.objects.filter(
+                question=self.question,
+                is_ground_truth=True,
+                images__in=self.images.values_list("id", flat=True),
+            ).exclude(pk=self.pk).update(is_ground_truth=False)
+
```
MikeOverkamp-diag: I'm not entirely sure if this is the best option when ground truth has already been added, another option would be to delete any existing ground truth answers for this question/images combination.

jmsmkn: No, I really don't think that the ground truth state should be modified right now, there will be a lot of edge cases that we can discuss. I guess you can't set a uniqueness constraint at the database level on `("question", "image_set", "ground truth")` as `image_set` is an M2M field. I think the best way for now is to add a clean method on the model and check for this before save (along with a test).

app/grandchallenge/reader_studies/models.py
```diff
@@ -251,6 +251,28 @@ def hanging_list_images(self):
...
 
+    @property
+    def image_groups(self):
+        return [sorted(x.values()) for x in self.hanging_list]
+
+    def add_ground_truth(self, *, data, user):
+        for gt in data:
+            images = self.images.filter(
+                name__in=gt["images"].split(";")
+            ).values_list("id", flat=True)
```
jmsmkn: Have we checked that these images belong to this study?

jmsmkn: Or is that done in L267 on answer create?

MikeOverkamp-diag: That is done in the field clean method on the form (https://github.com/comic/grand-challenge.org/pull/1094/files#diff-8f1ba48a837c1a7befad0375e2fc0740R235). This checks if the images provided in the csv match with the hanging list.

app/grandchallenge/reader_studies/models.py
MikeOverkamp-diag: So you mean to check if there is already a ground truth answer for the images/question combination? 

One of the difficulties with checking is that the images are not set on creation, as they are a M2M field. So initially, the new ground truth could pass the clean method, or it could fail even if it really should not. 

Another issue I see with only checking is that it could leave the user in a position where they really cannot do anything, as we would presumably raise a `ValidationError` if the ground truth has already been added for this question, but there is no way for the user to remove the existing ground truth.

Maybe an option would be to show a warning and then delete any existing ground truth for the current reader study whenever a new csv is uploaded?

PR #1104, opened 2020-01-21
Add reader study scoring

MikeOverkamp-diag:


app/grandchallenge/reader_studies/models.py
```diff
@@ -723,9 +795,12 @@ def validate(*, creator, question, answer, images, is_ground_truth=False):
                 f"{type(answer)} found."
             )
 
+    def calculate_score(self, ground_truth):
+        self.score = self.question.calculate_score(self.answer, ground_truth)
+        self.save()
```
jmsmkn: This could cause an infinite update loop, right? I guess the signal is filtered on `post_add` so probably ok. 

I think saving here goes beyond the responsibilities of this function. Probably better to remove the save method and use filter(...).update(...) in the signal.

app/grandchallenge/reader_studies/views.py
```diff
@@ -166,6 +166,28 @@ def delete(self, request, *args, **kwargs):
         return super().delete(request, *args, **kwargs)
 
 
+class ReaderStudyLeaderBoard(
+    LoginRequiredMixin, ObjectPermissionRequiredMixin, DetailView
+):
+    model = ReaderStudy
+    permission_required = (
+        f"{ReaderStudy._meta.app_label}.view_{ReaderStudy._meta.model_name}"
```
jmsmkn: Do we want all of the readers of the study to see this? Won't they see everyone elses scores?

MikeOverkamp-diag: I was unsure about that. Should we just make this view available to editors and only show the reader's own score somewhere on the reader study page?

app/grandchallenge/reader_studies/models.py
MikeOverkamp-diag: I think it should not cause an infinite loop, but I agree that it would be cleaner to explicitly call save if you want the score to be saved.

app/grandchallenge/reader_studies/views.py
jmsmkn: Lets go with editors only for now

app/grandchallenge/reader_studies/models.py
MikeOverkamp-diag: moved.

app/grandchallenge/reader_studies/views.py
MikeOverkamp-diag: Changed.

Issue #1107, opened 2020-01-21
CI broken/flaky tests


jmsmkn:
There are 2 tests that seem to randomly fail. I cannot reproduce either on my local machine so have been trying on travis by removing things from the travis file. The tests:

1. `test_method_validation`
2. `TestArchiveAPIView.test_caching`

seem to fail randomly. See:

- https://travis-ci.org/comic/grand-challenge.org/builds/640086532 (1)
- https://travis-ci.org/comic/grand-challenge.org/builds/640136945 (1&2)
- https://travis-ci.org/comic/grand-challenge.org/builds/640137801 (1)
- https://travis-ci.org/comic/grand-challenge.org/builds/640138832 (green)
- https://travis-ci.org/comic/grand-challenge.org/builds/640139133 (1&2)
- https://travis-ci.org/comic/grand-challenge.org/builds/640139554 (green)
- https://travis-ci.org/comic/grand-challenge.org/builds/640142477 (green)
- https://travis-ci.org/comic/grand-challenge.org/builds/640150657 (green)
- https://travis-ci.org/comic/grand-challenge.org/builds/640157524 (green)
- https://travis-ci.org/comic/grand-challenge.org/builds/640162768 (green)

Note that by re-running a test you might get a different output.

jmsmkn:
Note that the method validation error was first seen here: https://github.com/comic/grand-challenge.org/commit/68f30b43e50a

This change seems innocuous so probably not it.

jmsmkn:
It might be to do with the Docker service on travis, since removing 

```
service:
  - docker
```

everything has been ok.

jmsmkn:
Rebuilding old jobs now fails, something in travis changed?

jmsmkn:
I have added a basic github actions config to run all of the tests and this passes, so something is definitely wrong with travis and we will migrate to github actions.

jmsmkn:
TODO:

- [ ] Coverage reporting
- [ ] Uploading of the containers to docker hub
- [ ] Changing the environment variables
- [ ] Adding docker_gid to the docs to pycharm testing

jmsmkn:
I've replaced travis with github actions and everything works really nicely. Last thing:

- [ ] The git branch name doesn't work, so extract this from `${{ github.ref }}`

PR #1113, opened 2020-01-22
Add commands to clean media urls in the page html

jmsmkn:


jmsmkn:
All these test failures are I think due to #1107

PR #1121, opened 2020-01-24
Fix adding ground truth for boolean type questions

MikeOverkamp-diag:
In a later stage, when we add support for more types of answers, we might need a more proper preprocessing step for the csv data, but for now this fixes the issue that adding boolean values was not possible via the csv.


jmsmkn:
Please rebase this PR on to master to get the CI changes

PR #1122, opened 2020-01-24
Move reader study scoring to a celery task

MikeOverkamp-diag:


jmsmkn:
Please rebase this PR on to master to get the CI changes

app/grandchallenge/reader_studies/signals.py
```diff
@@ -44,23 +44,10 @@ def update_image_permissions(instance, action, reverse, model, pk_set, **_):
...
-        ground_truth = Answer.objects.filter(
-            question=instance.question,
-            is_ground_truth=True,
-            images__in=pk_set,
-        ).first()
-        if ground_truth:
-            instance.calculate_score(ground_truth.answer)
-            instance.save()
+    add_scores.apply_async(
+        kwargs={"instance_pk": instance.pk, "pk_set": list(pk_set)}
```
jmsmkn: Are you certain that this works? Even outside of automated tests? I thought that sometimes you get problems with Celery and serialisation of the kwargs, so you might need to use `str(instance.pk)` and ensure that `list(pk_set)` is a list of strings.

app/grandchallenge/reader_studies/signals.py
```diff
@@ -44,23 +44,10 @@ def update_image_permissions(instance, action, reverse, model, pk_set, **_):
 
 @receiver(m2m_changed, sender=Answer.images.through)
 def assign_score(instance, action, reverse, model, pk_set, **_):
+    from grandchallenge.reader_studies.tasks import add_scores
```
jmsmkn: Does this need to be a local import?

app/grandchallenge/reader_studies/signals.py
MikeOverkamp-diag: I did not run into any issues when testing it by hand, but I added a cast to string just to be sure.

app/grandchallenge/reader_studies/signals.py
MikeOverkamp-diag: No, changed.

PR #1128, opened 2020-01-29
Changes default internal metaimage format to MHA

### LINKED ISSUE #715: Replace mhd/zraw with mha as internal image representation?
I think that we should replace mhd/zraw with compressed mha as our internal representation. I think that this would have the following advantages:

- At the moment, we re-write all meta image files to be named out.mhd and out.zraw, which are located at `images/<image.pk>/out.<type>`. This is awkward as everywhere we need to handle two files, sort out which one is the header, and already in the codebase there are a lot of select by file extension statements.
- If we want to share a bunch of files, say, as an ImageSet for training, all of the files will have the same name, and will have to be re-written (including mhd modification) before they're given to the user. 
- We have the `ImageFile.IMAGE_TYPE_MHD` flag but this actually represents 2 types: a header and data container.

I would prefer that we use mha files. Then, we only have to deal with 1 file, we don't need to sort it everywhere, and we can rename it on the fly. This would then be consistent with the tiff representation. We could then serve images from `images/<image.pk>/<imagefile.pk>/`.

I'm fairly sure that this would still be compatible with the Python (SimpleITK), Mevislab and Javascript (ITK.js) loading libraries.

What could break:
- Everywhere that we've made an assumption about getting 2 images:
  - CIRRUS Web 
  - The `get_sitk_image` function (mhd is not used directly by the ophthalmology workstation).
  - Processors: how do the existing algorithms handle mhd?

Any thoughts? cc. @pkcakeout @HarmvZ @haimasree

jmsmkn:
Adds a `ITK_INTERNAL_FILE_FORMAT` setting so that we can switch this out if one of the frontends cannot cope with mha images. This will only effect new uploads, we will not convert existing files.

Also removes some unused test resources and shuffles the settings file.

Closes #715


pkcakeout:
Changes to CIRRUS are needed for this. Check this snippet:

~~~.py
        def make_handle(key):
            if isinstance(key, ImageHandle):
                return key  # already a handle
            if key.startswith("https"):
                # This is not a key but a url, assume it's for an image and
                # get it directly
                json_resp = self.grand_challenge_client(url=key)
            else:
                json_resp = self.grand_challenge_client.images.detail(key)
            files_list = json_resp["files"]
            if any(file["image_type"].lower() == "mhd" for file in files_list):
                return make_mhd_handle(files_list)
            else:
                return make_histopathology_handle(json_resp)
~~~

jmsmkn:
Thanks! Only CIRRUS Core is really used, so I will update that before deploying this. We need an un-broken master on there first.

jmsmkn:
@HarmvZ this is ready to go now, could you check that its ok, and in particular, does not break the retina front end?

HarmvZ:
I just quickly checked and I think somehting is broken. I haven't been able to find what exactly. I will look into it tomorrow.

HarmvZ:
False alarm! I did not notice that the changes from #1138 are not in here which made the retina front end seem empty. But I tested it now and it works correctly. I also added MHA support for the retina_importers which I needed to test it thoroughly. I think this is now good to go if the tests pass.

PR #1129, opened 2020-01-29
Upgrade to Python 3.8

jmsmkn:
Some `noqa` statements are duplicated as `pre-commit` likes to use system python, and the lines are reported differently between 3.6 (my system) and 3.8.

No changes to any app code, which is nice.


MikeOverkamp-diag:
Just to be sure: this does not require any user action right? Poetry can handle the update by itself without any issues?

jmsmkn:
Yep! Had to re-generate the lock file but it's done (you get a warning from the container builds if you forget to do that).

PR #1134, opened 2020-01-30
Load archive data as celery task and cache result

### LINKED ISSUE #1095: SystemExit popping up because of retina archives endpoint time out
See #913 
The archives endpoint is taking too long again to load all the data. I need to filter out some data that is not used currently to decrease the loading time.

HarmvZ:
This should solve the `SystemExit` errors that occur on the `/retina/archives/` endpoint. It runs an hourly task that builds the response object and caches it. When requesting the endpoint it will check if the result is cached and return that. If it is not in the cache it will trigger the task and return an error message explaining that the result will be ready after a minute or so.

Closes #1095

### DIFF
--- a/app/grandchallenge/retina_api/views.py
+++ b/app/grandchallenge/retina_api/views.py
@@ -3,9 +3,9 @@ from enum import Enum
 
 from django.conf import settings
 from django.contrib.auth import get_user_model
+from django.core.cache import cache
 from django.core.exceptions import (
     MultipleObjectsReturned,
-    ObjectDoesNotExist,
     ValidationError,
 )
 from django.http import HttpResponse
@@ -63,6 +63,7 @@ from grandchallenge.retina_api.serializers import (
     TreeImageSerializer,
     TreeObjectSerializer,
 )
+from grandchallenge.retina_api.tasks import cache_archive_data
 from grandchallenge.serving.permissions import user_can_download_image
 from grandchallenge.studies.models import Study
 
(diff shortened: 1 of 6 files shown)


app/grandchallenge/retina_api/views.py
```diff
@@ -72,142 +73,19 @@ class ArchiveView(APIView):
...
     def get(self, request):
-        return Response(self.create_response_object())
+        archive_data = cache.get(settings.RETINA_ARCHIVE_DATA_CACHE_KEY)
+        if archive_data is None:
+            cache_archive_data.delay()
+            return Response(
+                {
+                    "error": [
+                        1,
+                        "Archive data task triggered. Try again in 30 seconds.",
```
jmsmkn:
```suggestion
                        "Archive data task triggered. Try again in 2 minutes.",
```

jmsmkn: Headroom

app/tests/retina_api_tests/test_views.py
```diff
@@ -66,269 +67,51 @@ def test_archive_view_retina_auth(self, client):
...
-            },
-            "info": "level 2",
-            "name": "Archives",
-            "id": "none",
-            "images": {},
+        # check if correct error is sent
+        assert response_data == {
+            "error": [
+                1,
+                "Archive data task triggered. Try again in 30 seconds.",
```
jmsmkn:
```suggestion
                "Archive data task triggered. Try again in 2 minutes.",
```

PR #1144, opened 2020-02-04
Fix reader study leaderboard statistics

MikeOverkamp-diag:


app/grandchallenge/reader_studies/models.py
```diff
@@ -419,13 +419,25 @@ def statistics(self):
             .annotate(Sum("score"), Avg("score"))
             .order_by("-score__avg")
         )
+        # TODO: the ground truth only works correctly when there is one question
```
jmsmkn: We should make an issue for this

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_statistics.html
```diff
@@ -32,7 +33,9 @@ <h2 class="h3">Statistics per case</h2>
                     <tr>
                         <td data-order="{{ entry.images__name }}">{{ entry.images__name }}</td>
                         <td data-order="{{ entry.score__sum|stringformat:'020f' }}">{{ entry.score__sum }} / {{ object.statistics.max_score_cases }}</td>
-                        <td data-order="{{ entry.score__avg|stringformat:'020f' }}">{{ entry.score__avg }}</td>
+                        <td data-order="{{ entry.score__avg|stringformat:'020f' }}">{{ entry.score__avg|floatformat:4 }}</td>
+
+                        <td data-order="{{ ground_truth }}">{{ entry.ground_truth }}</td>
```
jmsmkn: We need to remember to filter this if we open up this page to non-editors

jmsmkn: Maybe add a TODO or warning around the ListView so the next programmer is aware?

MikeOverkamp-diag: Added a `TODO` to the view class.

app/grandchallenge/reader_studies/models.py
MikeOverkamp-diag: Issue added: #1145

Issue #1142, opened 2020-02-04
Add a user in init_gc_demo for using "Test Algorithm" in a rse-gcapi test


haimasree:
**Is your feature request related to a problem? Please describe.**
rse-gcapi now has a functionality to run an algorithm on uploaded images using rest api.  Unfortunately, I cant write a test because user and retina does not have access to Test Algorithm. 
**Describe the solution you'd like**
I would like to add a user with permission to upload and use a new algorithm in init_gc_demo.

PR #1146, opened 2020-02-04
Permits access to "Test Algorithm" to "algroithmuser" in init_gc_demo

haimasree:
Closes: #1142


haimasree:
> The "user" user should not have any special permissions as we use this to manually test other features. Create a new algorithm_user instead.
> 
> This user only need to be added to `algorithm.users_group` and not to the `algorithm_group`.

Done

PR #1148, opened 2020-02-05
Fix concurrent workstation startup

### LINKED ISSUE #1143: If many sessions are started at the same time sometime you get a "could not connect" message
Refreshing works fine, so we probably need to retry the connection with some backoff in the js. 

image (3)

jmsmkn:
When workstation sessions are started concurrently it can
take a while for traefik to route them, resulting in some
running workstations returning 404.

We use Math.Random to add jitter to the clients who are
trying to connect. A max_attempts parameter is added so
that the client will try for ~30 seconds (currently)
after the workstation is started to connect to it,
before giving up.

Closes #1143

### DIFF
--- a/app/grandchallenge/workstations/static/workstations/js/session.js
+++ b/app/grandchallenge/workstations/static/workstations/js/session.js
@@ -1,6 +1,7 @@
 "use strict";
 
 const timeout = 1000;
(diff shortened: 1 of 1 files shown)


HarmvZ:
LGTM

PR #1149, opened 2020-02-05
Fix reader message after course completion

MikeOverkamp-diag:


app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
```diff
@@ -68,7 +68,7 @@ <h1>Reader Study {{ object.title }}</h1>
         <div class="h3">
```
jmsmkn: Actually, this looks weird in an h3 block. Maybe make this change:

```suggestion
        <div class="alert alert-info" role="alert">
```

MikeOverkamp-diag: Yes, that looks a lot better.

PR #1150, opened 2020-02-05
Account for ground truth in progress, export and when adding answers

MikeOverkamp-diag:


jmsmkn:
I saw this test fail earlier today on master, but re-running the action solved it. I don't recall it being flaky before.

app/grandchallenge/reader_studies/templatetags/get_ground_truth.py
```diff
@@ -0,0 +1,10 @@
+from django import template
+
+register = template.Library()
+
+
+@register.simple_tag
+def get_ground_truth(obj, image, question):
+    """Get the auth token for the user."""
```
jmsmkn: Update the docstring

app/grandchallenge/reader_studies/models.py
```diff
@@ -339,39 +343,47 @@ def get_progress_for_user(self, user):
...
-                        readerstudies=self, answers__isnull=True
+        # Group the answers by images and filter out the images that
+        # have an inadequate amount of answers
+        unanswered_images = (
+            answers.order_by("images__name")
+            .values("images__name")
+            .annotate(answer_count=Count("images__name"))
+            .filter(answer_count__lt=self.answerable_question_count)
+        )
+        image_names = set(
```
jmsmkn: Wow! Nothing to change here but I think that we probably need a model manager at this point.

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_statistics.html
```diff
@@ -24,7 +25,9 @@ <h2 class="h3">Statistics per case</h2>
                     <th>Image name</th>
                     <th>Total score / max score</th>
                     <th>Average score</th>
-                    <th>Ground truth</th>
+                    {% for question in object.statistics.questions %}
```
jmsmkn: In `results_list.html` there is some js for hiding multiple columns at once, we will probably need that here in the future.

app/grandchallenge/reader_studies/templatetags/get_ground_truth.py
MikeOverkamp-diag: Changed it locally. I'll include it in the next PR (making a separate PR for would be a bit much I'd say).

app/grandchallenge/reader_studies/models.py
MikeOverkamp-diag: Added an issue for this: https://github.com/comic/grand-challenge.org/issues/1154

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_statistics.html
MikeOverkamp-diag: Added an issue for this as well: https://github.com/comic/grand-challenge.org/issues/1155

Issue #1151, opened 2020-02-05
Add functionality to filter ImageFile objects by RawImageUploadSession from a client


haimasree:
**Is your feature request related to a problem? Please describe.**
I would like to know which ImageFile was generated by a specific RawImageUploadSession from gcapi
**Describe the solution you'd like**
Add an additional item to the existing set of filters in ImageViewSet.get_queryset()

PR #1152, opened 2020-02-05
Filter Image by RawImageUploadSession from a client

haimasree:
Closes: #1151 
test_submission_conversion is failing on master branch as well. I will merge master after it is fixed.


app/tests/cases_tests/test_api.py
```diff
@@ -321,3 +325,38 @@ def request_processing():
...
+    alg_result.images.add(im2)
+
+    response = get_view_for_user(
+        viewname="api:image-list",
+        client=client,
+        user=user,
+        content_type="application/json",
+    )
+    assert response.status_code == 200
+    assert response.json()["count"] == 3  # [alg_result.job.image, im1, im2]
```
jmsmkn: It's important to check for content in code, rather than just leaving it in a comment. Why not `assert {r["pk"] for r in response.json()["results"]} == {i.pk for i in [alg_result.job.image, im1, im2]}`?

haimasree:
> This needs a test

Done

PR #1157, opened 2020-02-07
Add AI for radiology website

MikeOverkamp-diag:


app/config/settings.py
```diff
@@ -356,6 +356,7 @@ def strtobool(val) -> bool:
     "grandchallenge.workstation_configs",
     "grandchallenge.policies",
     "grandchallenge.favicons",
+    "grandchallenge.ai_website",
```
jmsmkn: App names stick around and I think that it's important to get this right. Can we think of a better name? Products?

app/grandchallenge/ai_website/admin.py
```diff
@@ -0,0 +1,10 @@
+from django.contrib import admin
+
+from grandchallenge.ai_website.models import (
+    CompanyEntry,
+    ProductEntry,
```
jmsmkn: I would remove the `Entry` suffix from all of the models, again, naming sticks around in django.

app/grandchallenge/ai_website/management/commands/import_data.py
```diff
@@ -0,0 +1,18 @@
+from django.core.management import BaseCommand
+
+from grandchallenge.ai_website.utils import DataImporter
+
+
+class Command(BaseCommand):
+    help = "Reads excel database and creates sqlite database"
```
jmsmkn: This is only temporary, right? `xlsrd` can be removed once the initial import is done?

app/grandchallenge/ai_website/models.py
```diff
@@ -0,0 +1,170 @@
+from django.db import models
+from django.utils import timezone
+from django.utils.text import get_valid_filename
+
+from grandchallenge.core.storage import public_s3_storage
+
+
+def get_logo_path(instance, filename):
+    return f"logos/{instance.__class__.__name__.lower()}/{instance.pk}/{get_valid_filename(filename)}"
```
jmsmkn: Import this from the other location (in challenges or core IIRC)

app/grandchallenge/ai_website/models.py
```diff
@@ -0,0 +1,170 @@
...
+        help_text="Short summary of this project, max 250 characters.",
+    )
+
+    def __str__(self):
+        return self.company_name
+
+
+class ProductImage(models.Model):
+    img = models.ImageField(
+        upload_to=get_images_path, storage=public_s3_storage
```
jmsmkn: This could use `get_logo_path` and it would be sorted into a subdirectory

jmsmkn: `public_s3_storage` is the default now, do not need to define this

app/grandchallenge/ai_website/models.py
```diff
@@ -0,0 +1,170 @@
...
+    # author = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
+    created_date = models.DateField(default=timezone.now)
+    modified_date = models.DateField(auto_now=True)
+    company_name = models.CharField(max_length=200)
+    website = models.URLField(max_length=500)
+    founded = models.IntegerField()
+    hq = models.CharField(max_length=100)
+    email = models.EmailField(max_length=500)
+    logo = models.ImageField(
+        upload_to=get_logo_path, storage=public_s3_storage, null=True,
```
jmsmkn: `public_s3_storage` is the default now, do not need to define this

app/grandchallenge/ai_website/models.py
```diff
@@ -0,0 +1,170 @@
...
+def get_logo_path(instance, filename):
+    return f"logos/{instance.__class__.__name__.lower()}/{instance.pk}/{get_valid_filename(filename)}"
+
+
+def get_images_path(instance, filename):
+    return f"product_images/{instance.__class__.__name__.lower()}/{instance.pk}/{get_valid_filename(filename)}"
+
+
+class CompanyEntry(models.Model):
+    # author = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
```
jmsmkn: Remove this line, `AUTH_USER_MODEL` should not be used directly.

app/grandchallenge/ai_website/models.py
```diff
@@ -0,0 +1,170 @@
...
+    created_date = models.DateField(default=timezone.now)
```
jmsmkn: add `auto_add_now` rather than default?

app/grandchallenge/ai_website/models.py
```diff
@@ -0,0 +1,170 @@
...
+    modified_date = models.DateField(auto_now=True)
+    company_name = models.CharField(max_length=200)
+    website = models.URLField(max_length=500)
```
jmsmkn: Probably don't need max_length on this (at least I hope not!)

app/grandchallenge/ai_website/models.py
```diff
@@ -0,0 +1,170 @@
...
+    founded = models.IntegerField()
+    hq = models.CharField(max_length=100)
+    email = models.EmailField(max_length=500)
```
jmsmkn: Probably don't need max_length

app/grandchallenge/ai_website/models.py
```diff
@@ -0,0 +1,170 @@
...
+    logo = models.ImageField(
+        upload_to=get_logo_path, storage=public_s3_storage, null=True,
+    )
+    description = models.CharField(
```
jmsmkn: Use a text field instead?

app/grandchallenge/ai_website/models.py
```diff
@@ -0,0 +1,170 @@
...
+    STATUS_CERTIFIED = "cer"
+    STATUS_YES = "yes"
+    STATUS_NO = "no"
+    STATUS_NA = "na"
+    STATUS_CLEARED = "cle"
+    STATUS_DE_NOVO_CLEARED = "dnc"
+    STATUS_PMA_APPROVED = "pma"
+    STATUS_UNKNOWN = "unk"
+
+    VERFIFIED_CHOICES = ((STATUS_YES, "Yes"), (STATUS_NO, "No"))
```
jmsmkn: We can use Enumeration types now :-) https://docs.djangoproject.com/en/3.0/ref/models/fields/#field-choices-enum-types

app/grandchallenge/ai_website/models.py
```diff
@@ -0,0 +1,170 @@
...
+        STATUS_YES: "icon_check.png",
+        STATUS_NO: "icon_no.png",
+        STATUS_NA: "icon_na.png",
+        STATUS_CLEARED: "icon_check.png",
+        STATUS_DE_NOVO_CLEARED: "icon_check.png",
+        STATUS_PMA_APPROVED: "icon_check.png",
+        STATUS_UNKNOWN: "icon_question.png",
+    }
+
+    created_date = models.DateField(default=timezone.now)
```
jmsmkn: Use `auto_now_add`

jmsmkn: Also, created, and a `DateTimeField`

app/grandchallenge/ai_website/models.py
```diff
@@ -0,0 +1,170 @@
...
+    modified_date = models.DateField(default=timezone.now)
```
jmsmkn: Change to a `DateTimeField` and use `auto_now`

app/grandchallenge/ai_website/models.py
```diff
@@ -0,0 +1,170 @@
...
+    published_date = models.DateField(blank=True, null=True)
+    product_name = models.CharField(max_length=200)
+    company = models.ForeignKey(
+        CompanyEntry, on_delete=models.CASCADE
+    )  # product.company.company_name
+    short_name = models.CharField(
```
jmsmkn: TextField?

app/grandchallenge/ai_website/models.py
```diff
@@ -0,0 +1,170 @@
...
+        max_length=500,
+        blank=False,
+        help_text=(
+            "short name used in url, specific css, files etc. No spaces allowed"
+        ),
+        unique=True,
+    )
+    description = models.CharField(
```
jmsmkn: TextField?

app/grandchallenge/ai_website/models.py
```diff
@@ -0,0 +1,170 @@
...
+        blank=True,
+        help_text="Short summary of this project, max 300 characters.",
+    )
+    description_short = models.CharField(
+        max_length=250,
+        default="",
+        blank=True,
+        help_text="Short summary of this project, max 250 characters.",
+    )
+    modality = models.CharField(max_length=100)
```
jmsmkn: We have this model in challenges already and an m2m field could be used here instead.

app/grandchallenge/ai_website/models.py
```diff
@@ -0,0 +1,170 @@
...
+    subspeciality = models.CharField(max_length=300)
+    diseases = models.CharField(max_length=200)
+    population = models.CharField(max_length=200)
```
jmsmkn: Choice fields? Or another M2M like we do with modalities, organ, task type etc.

app/grandchallenge/ai_website/models.py
```diff
@@ -0,0 +1,170 @@
...
+
+    input_data = models.CharField(max_length=150)
+    file_format_input = models.CharField(max_length=500)
+    output_data = models.CharField(max_length=150)
+    file_format_output = models.CharField(max_length=500)
+    key_features = models.CharField(max_length=200)
+    key_features_short = models.CharField(max_length=120)
+    software_usage = models.CharField(max_length=300)
```
jmsmkn: Choices maybe? Unclear to me what these are for.

app/grandchallenge/ai_website/models.py
```diff
@@ -0,0 +1,170 @@
...
+
+    verified = models.CharField(
+        choices=VERFIFIED_CHOICES, max_length=3, default=STATUS_NO
+    )
+    ce_status = models.CharField(
+        choices=CE_STATUS_CHOICES, max_length=3, default=STATUS_NO
+    )
+    ce_class = models.CharField(max_length=500, default="unknown")
```
jmsmkn: Shouldn't this be a choice of `1, 2A,` etc?

app/grandchallenge/ai_website/models.py
```diff
@@ -0,0 +1,170 @@
...
+        choices=FDA_STATUS_CHOICES, max_length=3, default=STATUS_UNKNOWN
+    )
+    fda_class = models.CharField(max_length=500, default="unknown")
+    ce_verified = models.CharField(
+        choices=VERFIFIED_CHOICES, max_length=3, default=STATUS_NO
+    )
+
+    integration = models.CharField(max_length=500, default="unknown")
+    deployment = models.CharField(max_length=500, default="unknown")
+    process_time = models.CharField(max_length=500)
```
jmsmkn: Duration field?

app/grandchallenge/ai_website/models.py
```diff
@@ -0,0 +1,170 @@
...
+    trigger = models.CharField(max_length=100)
+
+    market_since = models.CharField(
+        max_length=500, default="unknown"
+    )  # choose only month year in datefield?
```
jmsmkn: DateField?

app/grandchallenge/ai_website/models.py
```diff
@@ -0,0 +1,170 @@
...
+    countries = models.CharField(max_length=500, default="unknown")
```
jmsmkn: Use a country field from Django-countries

app/grandchallenge/ai_website/templates/ai_website/base.html
```diff
@@ -0,0 +1,91 @@
...
+      href="https://cdn.datatables.net/v/bs4/dt-1.10.18/b-1.5.4/b-colvis-1.5.4/datatables.min.css"
+      crossorigin="anonymous">
+  <link href="{% static 'vendor/fa/css/all.css' %}" rel="stylesheet">
+
+  <link rel="stylesheet" href="{% static 'ai_website/css/main.css' %}">
+  <link rel="icon" href="{% static 'ai_website/images/logo_16.ico' %}">
+  <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Raleway">
+  <style>
+    body,h1,h2,h3,h4,h5,h6 {font-family: "Raleway", sans-serif}
+  </style>
```
jmsmkn: Can we use {{ block.super }} in any way here?

app/grandchallenge/ai_website/templates/ai_website/contact.html
```diff
@@ -0,0 +1,19 @@
...
+  <li class="breadcrumb-item active" aria-current="page">Contact</li>
+{% endblock %}
+
+{% block content %}
+  <div>
+    <h1>Contact us</h1>
+    <p>Contact us if you have questions, ideas, would like to update information, or if you believe your product is missing from this overview. <br>
+    Please send an e-mail to:
+      <div class='text-black'> Kicky van Leeuwen <br>
+      <a href = "mailto: kicky.vanleeuwen@radboudumc.com" class="text-radboud">kicky.vanleeuwen@radboudumc.com</a> </div>
```
jmsmkn: Use `random_encode` :-)

app/grandchallenge/ai_website/templates/ai_website/base.html
```diff
@@ -0,0 +1,91 @@
...
+          </div>
+        </div>
+
+        <div class="col-md-3 px-4">
+          <div class="h4">Contact</div>
+          <div>Radboud University Medical Center </div>
+          <div>Geert Grooteplein Zuid 10 </div>
+          <div>6525 GA, Nijmegen </div>
+          <div>The Netherlands</div>
+          <div><a href="mailto:kicky.vanleeuwen@radboudumc.nl" class="text-radboud">kicky.vanleeuwen@radboudumc.nl</a></div>
```
jmsmkn: `random_encode`

app/grandchallenge/ai_website/views.py
```diff
@@ -0,0 +1,187 @@
...
+        kwargs = super().get_form_kwargs()
+        kwargs.update({"user": self.request.user})
+        return kwargs
+
+    def form_valid(self, *args, **kwargs):
+        response = super().form_valid(*args, **kwargs)
+        form = self.get_form()
+        if form.is_valid():
+            di = DataImporter()
+            di.import_data(
```
jmsmkn: How long does this take?

app/grandchallenge/ai_website/management/commands/import_data.py
MikeOverkamp-diag: The management command can actually be deleted, I have moved this functionality to a view. The import may have to be done a number of times, so I wouldn't say we can remove `xlsrd` straight away, but I think we should be able to move away from it fairly quickly.

app/grandchallenge/ai_website/models.py
MikeOverkamp-diag: I think these are just fields that contain extra data and can be filled with anything really. So I do not think choices are appropriate here. A `TextField` probably is though.

app/grandchallenge/ai_website/views.py
MikeOverkamp-diag: This takes a couple of seconds on my local machine with the current dataset.

app/grandchallenge/ai_website/templates/ai_website/base.html
MikeOverkamp-diag: We could add an extra block in `style.html`, but that does not make it much cleaner in my opinion. Also, when I tried that locally for some reason some of the styling broke, so I'd rather keep it like this for now and possibly debug the issue later. I did remove `datatables.js`, as it's not used at all in the products app.

app/grandchallenge/ai_website/views.py
jmsmkn: Ok, as long as it is not going to kill the view

app/grandchallenge/ai_website/templates/ai_website/base.html
jmsmkn: :+1:

PR #1163, opened 2020-02-11
Allow anonymous users to view public algorithms

jmsmkn:
This PR:

- Adds `REGISTERED_USERS_GROUP` and `REGISTERED_AND_ANON_USERS_GROUP` that keep track of all the registered and all the registered + anonymous users
  - This allows assigning permissions to this group for django guardian rather than hacking around their `get_user_permission` to allow all users to do something with an object
  - We do not do anything with the `REGISTERED_USERS_GROUP`, but this will be useful later
- Adds a new `execute_algorithm` permission, this is needed for starting jobs rather than `view_algorithm`
- Assigns the `view_algorithm` permission to the `REGISTERED_AND_ANON_USERS_GROUP` if `visible_to_public=True`
- Adds the session size and uniqueness checks to the `process_images` endpoint which was previously skipped


app/grandchallenge/cases/views.py
```diff
@@ -88,27 +90,53 @@ class RawImageUploadSessionViewSet(
...
         serializer.save(creator=self.request.user)
 
+    def validate_staged_files(self):
+        upload_session: RawImageUploadSession = self.get_object()
+
+        file_ids = [
+            f.staged_file_id for f in upload_session.rawimagefile_set.all()
+        ]
+
+        if not all(f_id is not None for f_id in file_ids):
```
MikeOverkamp-diag: It might be nicer to use
```suggestion
        if any(f_id is None for f_id in file_ids):
```
I  think removing the double negation improves readability a bit.

jmsmkn: Agreed, thanks!

PR #1169, opened 2020-02-13
Fix small products issues

MikeOverkamp-diag:


app/grandchallenge/products/templatetags/products_tags.py
```diff
@@ -45,3 +46,11 @@ def icon(obj, field):
...
     if icon:
         return static(f"products/images/{icon}")
+
+
+@register.filter
+@stringfilter
+def short(value, max_char):
+    if len(value) > max_char:
+        return value[:max_char].rsplit(" ", 1)[0] + " ..."
+    return value
```
jmsmkn: Isn't this the same as https://docs.djangoproject.com/en/3.0/ref/templates/builtins/#truncatechars?

MikeOverkamp-diag: It's a bit of a mix between truncate_chars and truncate_words. It leaves words intact and adds an ellipsis if anything is truncated.

PR #1170, opened 2020-02-14
Add public results and workstations

jmsmkn:
This PR:

- Adds `change_result` permission to the algorithm editors group
  - [ ] Requires all existing results to be saved after deployment of this PR
- Adds `public` property to results, if true, anyone can view the job image, result, and result images (including anonymous users)
- Adds `public` property to workstations. If true, all logged in users can use the workstation (not anonymous)
- Adds a `result-update` view that uses the `change_result` permission.
  - [ ] The permissions check for this in the `jobs-list` page only checks for `change_algorithm` right now, this will be changed in a future PR, but permission wise this is ok as all algorithm editors have permission to change this result.

So, to make a result visible in a workstation you need the algorithm to use a public workstation AND to make the result public, in which case only logged in users will be able to use the workstation but anyone can download the relevant images (without authentication). This would allow a user to use an alternative viewer, such as open sea dragon.


MikeOverkamp-diag:
```Requires all existing results to be saved after deployment of this PR```
Maybe you could add a simple data migration for this in which you simply save all results?

jmsmkn:
Django guardian does not work well in migrations (Harm faced this issue in the past too) so I would need to be done with a management command anyway, rather than adding that and deleting it once it's done it's easier to do it via the shell.

jmsmkn:
I don't know what is up with the coverage report, for some reason it seems to be picking up the product changes too. The coverage in algorithms/views.py actually goes up...

PR #1175, opened 2020-02-18
Add 3D dicom support

### LINKED ISSUE #1060: Add 3D DICOM data loader
Aside from the 4D loader implemented in #999 a 3D DICOM loader would benefit a lot of projects for the RTC and AI for Health.

MikeOverkamp-diag:
This PR adds support for some 3D dicom files.
Because it is very hard to create a generic dicom importer that will work for all files, this PR also introduces a new user group. Whenever a user in this group uploads a dicom file and we cannot process the file, the file is not deleted at the end of the processing step (which is what normally happens. The failed files can then be downloaded in the Django admin.

NB. Tests for 3D Dicom files are not included in the PR. The files I currently have available are too large to include in the repo / use in CI.

Closes #1060


MikeOverkamp-diag:
Closes #1060

jmsmkn:
> Closes #1060

These tags need to be in the opening comment to work

app/grandchallenge/cases/migrations/0020_add_dicom_group.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def create_dicom_group(apps, schema_editor):
+    Group.objects.create(name=settings.DICOM_DATA_CREATORS_GROUP_NAME)
+
+
+def remove_dicom_group(apps, schema_editor):
+    group = Group.objects.filter(
+        name=settings.DICOM_DATA_CREATORS_GROUP_NAME
+    ).first()
```
jmsmkn: I'd rather do a `get` here, the group names are unique.

app/grandchallenge/cases/views.py
```diff
@@ -146,3 +147,23 @@ class RawImageFileViewSet(
     queryset = RawImageFile.objects.all()
     permission_classes = [DjangoObjectOnlyWithCustomPostPermissions]
     filter_backends = [ObjectPermissionsFilter]
+
+    def get_permissions(self):
+        if self.action == "download":
+            return [IsAdminUser()]
```
jmsmkn: Wouldn't this bypass the object permissions for staff members?

app/grandchallenge/cases/views.py
```diff
@@ -146,3 +147,23 @@ class RawImageFileViewSet(
...
+        return super().get_permissions()
+
+    @action(detail=True, methods=["get"])
+    def download(self, request, pk=None):
```
jmsmkn: Do we really need to expose this? Isn't having it in the admin panel enough?

app/grandchallenge/cases/admin.py
```diff
@@ -102,7 +104,35 @@ def algorithm(self, obj):
...
+
+class RawImageFileAdmin(admin.ModelAdmin):
+    list_filter = (DownloadableFilter,)
+    list_display = ("filename", "upload_session", "download")
+    readonly_fields = ("download",)
+
+    def download(self, instance):
+        if not instance.staged_file_id:
+            return
+        return mark_safe(
```
jmsmkn: I'm always paranoid about using `mark_safe`. I know that it's technically not needed but could we use `format_html` here instead?

app/grandchallenge/cases/views.py
MikeOverkamp-diag: It would, but I felt like we might want to make the downloads available to all staff members, regardless of object permissions? But maybe we should check for superuser status instead?

jmsmkn: We shouldn't let staff members do this. I think remove the check, super users have permission anyway.

MikeOverkamp-diag: I moved all the download logic to the admin, so this is no longer an issue

app/grandchallenge/cases/admin.py
MikeOverkamp-diag: Moved this to the admin.

MikeOverkamp-diag:
> Looks good! A few questions, also consider `grandchallenge.jqfileupload.tasks.cleanup_stale_uploads` because that is going to nuke the uploaded files so they might not be available for download.

I increased the timeout for files that should be downloadable.

app/grandchallenge/cases/tasks.py
```diff
@@ -509,6 +511,9 @@ def _delete_session_files(*, session_files):
                     and Path(file.filename).suffix == ".dcm"
                     and getattr(file.creator, "username", None) in users
                 ):
+                    saf.staged_files.update(
+                        timeout=timezone.now() + timedelta(days=100)
```
jmsmkn:
```suggestion
                        timeout=timezone.now() + timedelta(days=21)
```

Less extreme, and still gives some time to pick it up after a vacation.

MikeOverkamp-diag:
Rebased on master to solve merge conflicts.

app/grandchallenge/cases/admin.py
```diff
@@ -102,7 +111,72 @@ def algorithm(self, obj):
...
+            def wrapper(*args, **kwargs):
+                return self.admin_site.admin_view(view)(*args, **kwargs)
+
+            wrapper.model_admin = self
+            return update_wrapper(wrapper, view)
+
+        urls = super().get_urls()
+
+        # Replace catchall with download endpoint, as the catchall is only there
+        # for django < 1.10 compatibility.
```
jmsmkn: I don't understand this comment..?

app/grandchallenge/cases/views.py
jmsmkn: This code can go now, right?

MikeOverkamp-diag: Yes, I thought I had removed it.

app/grandchallenge/cases/admin.py
MikeOverkamp-diag: Agreed, it is a bit unclear. Come to think of it, this probably isn't the best way to go about this anyway, as it relies on the code in the super function to not change.

PR #1177, opened 2020-02-19
Algorithm upload flow

jmsmkn:
Peek 2020-02-19 17-29

This PR adds feedback on what is happening when you upload images to be run by an algorithm. It is much less janky than in the gif :-)

- Adds fields to the serializers that allow direct lookup of images, jobs, results, and import of result images
  - @haimasree once this is merged it is better to use these directly that filtering images, jobs, etc
- Adds some JS to do this animation, no handling of connection errors (not sure what to do there)
- Re-skins the site using flatly from bootswatch with a couple of changes
  - This is for consistency with the theme used in CIRRUS
- Adds an scss compiler
  - This does not watch changes so annoying to change, I think that it's best to add a separate npm container to do that

TODO:
- [ ] Add some tests!
- [ ] Filter the job by the algorithm for this session
- [x] Change the field name `rawimageuploadsession` on `ResultSerializer`
- [x] When using nested objects in urls do `get_object_or_404`
- [x] New status char fields should be enums in the api schema

@HarmvZ would you be able to have a look at the js already?


app/grandchallenge/algorithms/static/algorithms/js/session.js
```diff
@@ -0,0 +1,154 @@
...
+            }, Math.floor(Math.random() * timeout) + 100);
+            break;
+        case "started":
+            setCardActiveMessage(cards.imageImport, status);
+            setTimeout(function () {
+                getUploadSessionStatus(statusUrl, cards)
+            }, Math.floor(Math.random() * timeout) + 100);
+            break;
+        case "succeeded":
+            let msg = "Imported " + imageUrls.length + " Image";
```
MikeOverkamp-diag: you could also use backticks here:
```suggestion
            let msg = `Imported ${imageUrls.length} Image`;
```

jmsmkn:
Also good to know - I could not get django-compressor to work with whitenoise, this is why I resorted back to npm.

app/grandchallenge/algorithms/views.py
```diff
@@ -323,10 +316,29 @@ def get_context_data(self, *args, **kwargs):
...
+    LoginRequiredMixin, ObjectPermissionRequiredMixin, DetailView
+):
+    model = RawImageUploadSession
+    template_name = "algorithms/executionsession_detail.html"
+    permission_required = "cases.view_rawimageuploadsession"
+    raise_exception = True
+
+    @property
+    def algorithm(self) -> Algorithm:
+        return Algorithm.objects.get(slug=self.kwargs["slug"])
```
MikeOverkamp-diag: maybe use `get_object_or_404` here, just in case the algorithm does not exist.

jmsmkn: Ah, of course! We use this in a lot of places... I will make the change.

app/grandchallenge/algorithms/static/algorithms/js/session.js
jmsmkn: Thanks! I was desperately searching on "how do I f-string in js" yesterday...

app/grandchallenge/algorithms/views.py
jmsmkn: Fixed in 97e8c65

jmsmkn:
Thanks! I still need to add some tests, on it now, I should have made a draft PR but forgot to click the right button when creating this.

PR #1183, opened 2020-02-24
Reader study detail refactor

jmsmkn:
This PR adds a sidebar to the reader study detail page for easier navigation. It is still a single page so likely still slow.

Screenshot_20200224_131521

### DIFF
--- a/app/grandchallenge/core/templates/grandchallenge/partials/userlinks.html
+++ b/app/grandchallenge/core/templates/grandchallenge/partials/userlinks.html
@@ -9,7 +9,7 @@
             <img class="mugshot"
                  src="{{ user.user_profile.get_mugshot_url }}"
                  alt="User Mugshot"
-                 style="height: 1.5em; vertical-align: middle;"/>
+                 style="height: 1.33em; vertical-align: -25%;"/>
         </a>
         <div class="dropdown-menu dropdown-menu-right">
             <a class="dropdown-item"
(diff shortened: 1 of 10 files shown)


app/grandchallenge/core/templates/grandchallenge/partials/userlinks.html
```diff
@@ -9,7 +9,7 @@
             <img class="mugshot"
                  src="{{ user.user_profile.get_mugshot_url }}"
                  alt="User Mugshot"
-                 style="height: 1.5em; vertical-align: middle;"/>
+                 style="height: 1.33em; vertical-align: -25%;"/>
```
MikeOverkamp-diag: Does the -25% work well in all resolutions?

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
```diff
@@ -1,24 +1,76 @@
...
+            </div>
+        </div>
+    {% endif %}
+{% endblock %}
 
-    <h1>Reader Study {{ object.title }}</h1>
+{% block content %}
+    {% get_obj_perms request.user for object as "readerstudy_perms" %}
 
     {% if not object.is_valid %}
```
MikeOverkamp-diag: Something I just noticed and might as well be part of this PR: this message also shows for readers. We should also check if the user is an editor.

app/grandchallenge/core/templates/grandchallenge/partials/userlinks.html
jmsmkn: I think so, this is to match the alignment from font-awesome.

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
jmsmkn: Updated

PR #1189, opened 2020-02-26
Updates the algorithm results page

jmsmkn:
This PR:

- Tidies up the algorithm results page
- Makes it visible to non-logged in users (but the results are still filtered)
- Fixes the pre-commit hook which is broken with the release of `virtualenv==20.0.6`
- Fixes the algorithm job cleanup

With this PR there is nowhere to see the failed jobs, I will add an upload list page tomorrow.

### DIFF
--- a/app/grandchallenge/algorithms/templates/algorithms/result_list.html
+++ b/app/grandchallenge/algorithms/templates/algorithms/result_list.html
@@ -83,26 +83,24 @@
                     </td>
                     <td>
                         <ul class="list-unstyled">
-                            {% if result.images.all %}
-                                {% for image in result.images.all %}
-                                    <li>
-                                        <a href="{% url 'workstations:workstation-session-redirect' slug=algorithm.workstation.slug %}?{% workstation_query image=result.job.image overlay=image config=algorithm.workstation_config %}">
+                            {% for image in result.images.all %}
+                                <li>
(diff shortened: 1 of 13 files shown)


app/grandchallenge/algorithms/templates/algorithms/result_list.html
```diff
@@ -129,11 +82,11 @@ <h2>Results for {{ algorithm.title }}</h2>
                         {% endif %}
                     </td>
                     <td>
-                        {% if job.result.images.all %}
-                            <ul class="list-unstyled">
-                                {% for image in job.result.images.all %}
+                        <ul class="list-unstyled">
+                            {% if result.images.all %}
+                                {% for image in result.images.all %}
```
MikeOverkamp-diag: You probably want to use `{% for .. %}...{% empty %}` here, rather than wrap it in an `if..else` block. 

I think now `job.result.images.all` gets evaluated twice, which could lead to performance issues if there are many images.

jmsmkn: Oops!

jmsmkn: We should also do a prefetch_related on the queryset to stop that...

PR #1202, opened 2020-03-06
Fix some bugs in models(fda status), product_detail(p.institutes_clin…

Kicksta:
…ical), etc


app/grandchallenge/products/views.py
```diff
@@ -118,6 +119,7 @@ def get_queryset(self):
                 Q(),
             )
             queryset = queryset.filter(q)
+        self.queryset = queryset
```
jmsmkn: This should not be necessary

app/grandchallenge/products/templates/products/product_detail.html
```diff
@@ -192,16 +192,16 @@ <h1 class="mb-1">{{product.product_name}}</h1>
             </tr>
           </thead>
           <tr>
-            <td class="text-muted">Peer reviewed papers regarding performance as commercially available</td>
-            <td>{{ product.tech_peer_papers|linebreaksbr }}</td>
+            <td class="text-muted">Peer reviewed papers on performance</td>
+            <td>{{ product.tech_peer_papers|safe|linebreaksbr }}</td>
```
jmsmkn: Using `safe` is a security risk. Is it really needed? If so, use the `clean` filter instead, this will clean any dangerous html. At the top of the html file add `{% load bleach %}` and then replace `safe` with `clean`.

app/grandchallenge/products/templates/products/partials/product_image_carousel.html
```diff
@@ -25,7 +25,9 @@
   </div>
   <div class="text-muted small"> Source: {{product.company.website}}</div>
 
-{% elif product.product_img %}
+{% elif product.product_img %} <!--TODO: elif len(product.images.all)==1, no carousel-->
```
jmsmkn: Nitpick: it's nicer to use `{# my comment here #}` rather than `<!-- my comment here -->` in Django.

app/grandchallenge/products/models.py
```diff
@@ -132,7 +132,7 @@ class FDAStatus(models.TextChoices):
     countries = models.TextField()
     distribution = models.CharField(max_length=100, blank=True)
     institutes_research = models.TextField()
-    institutes_clinic = models.TextField()
+    institutes_clinical = models.TextField()
```
jmsmkn: As you changes the model in `models.py` you need to include a migration, you can do this with `make migrations`, run black, and then add and commit the generated file.

jmsmkn: @Kicksta This will be the cause of your local error

app/grandchallenge/products/views.py
Kicksta: If I currently search in the company bar for e.g. 'Thirona', one company comes up, but it states it found 50 results instead of 1 --> "company_total": len(self.queryset). Defining self.queryset solves this.

app/grandchallenge/products/models.py
Kicksta: Yup, changed it and then saw your comment ;) this was my way around it.

app/grandchallenge/products/views.py
jmsmkn: Ah! Maybe try this instead: in `CompanyList.get_context_data` change `"company_total": len(self.queryset),` to `"company_total": len(context["object_list"]),`.

Kicksta: fixed (also for ProductList)

app/grandchallenge/products/templates/products/product_detail.html
Kicksta: fixed, clean indeed also works

app/grandchallenge/products/templates/products/partials/product_image_carousel.html
Kicksta: fixed, good to know ;)

app/grandchallenge/products/templates/products/partials/product_image_carousel.html
```diff
@@ -1,31 +1,33 @@
...
-        <li data-target="#product_img_carousel" data-slide-to={{ forloop.counter0 }} {% if forloop.counter0 == 0%}class="active"{% endif %}></li>
-      {% endfor %}
-    </ol>
-    <div class="carousel-inner">
-      {% for image in product.images.all %}
-        <div class="carousel-item {% if forloop.counter0 == 0 %}active{% endif %}">
-          <img class="d-block w-100" src={{ image.img.url }}>
-        </div>
-      {% endfor %}
+  {% if product.images.all.count == 1 %}
```
MikeOverkamp-diag: `.all` is not needed here, you can just use:
```suggestion
  {% if product.images.count == 1 %}
```
By calling `.all` you first get all the objects from the database and then perform a count on them. By calling `.count` directly you create a query to the extend of `SELECT COUNT(*) from ...` which is a lot more efficient, especially on larger datasets.

app/grandchallenge/products/templates/products/partials/product_image_carousel.html
```diff
@@ -1,31 +1,33 @@
...
+    <img src={{product.images.first.img.url}} style="width:100%;">
+  {% else %}
+    <div id="product_img_carousel" class="carousel slide" data-ride="carousel" data-interval="false">
+      <ol class="carousel-indicators">
+        {% for image in product.images.all %}
+          <li data-target="#product_img_carousel" data-slide-to={{ forloop.counter0 }} {% if forloop.counter0 == 0%}class="active"{% endif %}></li>
+        {% endfor %}
+      </ol>
+      <div class="carousel-inner">
+        {% for image in product.images.all %}
```
MikeOverkamp-diag: This is a bit nitpicky, but you could also create variable for `product.images.all`, as you currently evaluate the queryset twice. You could either do this in the views `get_context_data`, or by using `{% with images=product.images.all %}` in the template.

This is not something that I think has to be changed, but it is something to be mindful of when writing django code.

PR #1205, opened 2020-03-11
Adds image-level endpoints

### LINKED ISSUE #1203: Add endpoint to retrieve image level annotation ids for a specific image and user
I need to create an endpoint that returns the ids of the most recent image-level annotations for a specific user and image. Also, the existing image level annotation models need new viewsets that are added to the default API instead of the retina API.

HarmvZ:
- Adds a new url, serializer and viewset that returns ids of most recent image-level annotations for a specific image and user
- Adds new viewsets for image-level annotations and adds these to the core API router
- Adds tests for new endpoints

Closes #1203

### DIFF
--- a/app/grandchallenge/api/urls.py
+++ b/app/grandchallenge/api/urls.py
@@ -22,7 +22,10 @@ from grandchallenge.reader_studies.views import (
     QuestionViewSet,
     ReaderStudyViewSet,
 )
-from grandchallenge.retina_api.views import LandmarkAnnotationSetViewSet
+from grandchallenge.retina_api.views import (
+    ImageLevelAnnotationsForImageViewSet,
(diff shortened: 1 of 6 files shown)


app/grandchallenge/api/urls.py
```diff
@@ -74,6 +81,31 @@
     LandmarkAnnotationSetViewSet,
     basename="landmark-annotation",
 )
+router.register(
+    r"retina/image-level-annotation-for-image",
+    ImageLevelAnnotationsForImageViewSet,
+    basename="image-level-annotation-for-image",
```
jmsmkn: Should the base name be `retina-image-level-annotation-for-image`? Then it is more consistent with the other namespaces. But, these are so specific they're unlikely to clash...

PR #1209, opened 2020-03-16
Point and polygon annotations

miriam-groeneveld:


jmsmkn:
Could we add CSI annotations at the same time?

app/grandchallenge/reader_studies/migrations/0009_auto_20200316_1439.py
```diff
@@ -0,0 +1,31 @@
...
+            name="answer_type",
+            field=models.CharField(
+                choices=[
+                    ("STXT", "Single line text"),
+                    ("MTXT", "Multi line text"),
+                    ("BOOL", "Bool"),
+                    ("HEAD", "Heading"),
+                    ("2DBB", "2D bounding box"),
+                    ("DIST", "Distance measurement"),
+                    ("MDIS", "Multiple distance measurements"),
```
jmsmkn: We don't want individual points? I can think of some studies in the past that have needed this.

app/grandchallenge/reader_studies/migrations/0010_auto_20200317_1027.py
```diff
@@ -0,0 +1,33 @@
+# Generated by Django 3.0.2 on 2020-03-17 10:27
```
jmsmkn: Consolidate the migrations to 1 file per app per PR.

app/tests/reader_studies_tests/validation_tests.py
```diff
@@ -192,31 +227,18 @@ def test_answer_type_annotation_header_schema_fails(
...
-    )
+def test_answer_type_annotation_schema_mismatch():
+    # Answers to STXT are valid for MTXT as well, that's why STXT is excluded
+    # Other than that, each answer is only valid for a single answer type
+    unique_answer_types = [
+        key for key in ANSWER_TYPE_NAMES_AND_ANSWERS.keys() if key != "STXT"
+    ]
+    for answer_type in unique_answer_types:
+        answer = ANSWER_TYPE_NAMES_AND_ANSWERS[answer_type]
+        for answer_type_check in unique_answer_types:
```
jmsmkn: For the future, it's better to use [zip](https://docs.python.org/3/library/functions.html#zip) rather than the double for loop here, but this is a test so it doesn't really matter :-)

PR #1210, opened 2020-03-17
Add new Polygon viewsets

### LINKED ISSUE #1207: Add support for CSI polygon annotations and migrate viewsets
Changes made for image-level annotations in #1205 also need to be made for `PolygonAnnotationSet` and `SinglePolygonAnnotation` models. New viewsets need to be created to support correct lookup. Most likely, the `PolygonListView` and `PolygonAnnotationSetViewSet` can be consolidated into one viewset using filters.
Also, an attribute needs to be added to the `SinglePolygonAnnotation` to indicate if the annotation is created manually or interpolated.

HarmvZ:
Closes #1207


app/grandchallenge/retina_api/filters.py
```diff
@@ -17,3 +17,32 @@ def filter_queryset(self, request, queryset, view):
...
+    def filter_queryset(self, request, queryset, view):
+        if is_in_retina_admins_group(request.user):
+            return queryset
+        return queryset.filter(annotation_set__grader=request.user)
+
+
+class ImageAnnotationFilter(filters.BaseFilterBackend):
+    """
+    Filter backend that only returns objects that belong to the image that
+    is supplied by the image_id query parameter.
```
jmsmkn, line 41: It might be nicer to use this: https://django-filter.readthedocs.io/en/master/guide/rest_framework.html#using-the-filterset-fields-shortcut

If you rebase on master django-filter will be available

jmsmkn:
Ah, but the tests do need to be fixed ;-)

app/grandchallenge/retina_api/filters.py
HarmvZ: Done!

PR #1215, opened 2020-03-23
Feature/choice answer

MikeOverkamp-diag:


app/grandchallenge/core/templates/layout/formset.html
```diff
@@ -0,0 +1,37 @@
...
+                    {% endif %}
+                    {{ field.errors.as_ul }}
+                    {{ field|as_crispy_field }}
+                </td>
+            {% endfor %}
+        </tr>
+    {% endfor %}
+
+</table>
+<script src="//ajax.googleapis.com/ajax/libs/jquery/2.1.3/jquery.min.js">
```
jmsmkn: JQuery is already included in head

app/grandchallenge/reader_studies/models.py
```diff
@@ -1129,7 +1147,19 @@ def validate(*, creator, question, answer, images, is_ground_truth=False):
                 raise ValidationError(
                     "This user is not a reader for this study."
                 )
-
+        if (
+            question.answer_type == Question.ANSWER_TYPE_CHOICE
+            and answer not in question.options.values_list("id", flat=True)
+        ):
+            raise ValidationError(
+                "Provided option is not valid for this question"
```
jmsmkn: When the frontend provides an integer, is this the foreign key to a CategoricalOption object? eg, they're globally unique?

app/grandchallenge/reader_studies/views.py
```diff
@@ -214,8 +216,37 @@ def get_context_data(self, **kwarsg):
         return context
 
 
+class QuestionOptionMixin(object):
+    def validate_options(self, form, _super):
+        context = self.get_context_data()
+        options = context["options"]
+        if form.data["answer_type"] in ["CHOI", "MCHO"] and not any(
```
jmsmkn: Use `ANSWER_TYPE_CHOICE` and `ANSWER_TYPE_MULTIPLE_CHOICE` instead of referring to the keys directly

app/grandchallenge/reader_studies/views.py
```diff
@@ -214,8 +216,37 @@ def get_context_data(self, **kwarsg):
...
+            option.get("title") for option in options.cleaned_data
+        ):
+            error = [
+                "At least one option should be supplied for (multiple) choice questions"
+            ]
+            form.errors["answer_type"] = error
+            return self.form_invalid(form)
+        if form.data["answer_type"] == "CHOI" and not any(
```
jmsmkn: Same here

app/grandchallenge/reader_studies/models.py
MikeOverkamp-diag: Yes, it should be.

app/grandchallenge/reader_studies/views.py
MikeOverkamp-diag: Ah, yes, forgot to clean this part up, sorry!

PR #1222, opened 2020-03-23
Fix dicom loader spacing, origin and pixel data order

MikeOverkamp-diag:


app/grandchallenge/cases/image_builders/dicom.py
```diff
@@ -167,49 +168,56 @@ def _process_dicom_file(dicom_ds):  # noqa: C901
...
+    for partial in dicom_ds.headers:
+        ds = partial["data"]
+        if "ImagePositionPatient" in ds:
+            file_origin = tuple(float(i) for i in ds.ImagePositionPatient)
+            if origin:
+                diff = tuple(np.subtract(file_origin, origin))
+                origin_diff = tuple(np.add(origin_diff, diff))
+            origin = file_origin
+            n_origins += 1
+    avg_origin_diff = tuple(x / n_origins for x in origin_diff)
```
pkcakeout: since you are using numpy anyway, you can write it straightup in the numpy way - prroabaly more efficient and everything:

```
origin = None
origin_diff = np.array((0, 0, 0), dtype=float)
n_origins = 0
for partial in dicom_ds.headers:
    ds = partial["data"]
    if "ImagePositionPatient" in ds:
        file_origin = np.array(ds.ImagePositionPatient, dytpe=float)
        if origin:
            diff = file_origin - origin
            origin_diff = origin_diff + diff
        origin = file_origin
        n_origin += 1
avg_origin_diff = tuple(origin_diff / n_origins)
```

Just a little tip for the future. more readable and less conversions between datatypes this way...

app/grandchallenge/cases/image_builders/dicom.py
```diff
@@ -151,6 +151,7 @@ def _create_sitk_image(dcm_array):
 
 def _process_dicom_file(dicom_ds):  # noqa: C901
     ref_file = pydicom.dcmread(str(dicom_ds.headers[0]["file"]))
+    ref_origin = tuple(float(i) for i in ref_file.ImagePositionPatient)
```
jmsmkn: DICOM files are not guaranteed to have ImagePositionPatient

MikeOverkamp-diag: Ah true, I forgot the if statement there

jmsmkn:
I have invited @cjacobs1 to this organisation, would be good to get his review.

app/grandchallenge/cases/image_builders/dicom.py
MikeOverkamp-diag: Thank you. I definitely need to learn to make better use of numpy's capabilities.

cjacobs1:
Checking now

app/grandchallenge/cases/image_builders/dicom.py
```diff
@@ -167,49 +170,56 @@ def _process_dicom_file(dicom_ds):  # noqa: C901
...
+    for partial in dicom_ds.headers:
+        ds = partial["data"]
+        if "ImagePositionPatient" in ds:
+            file_origin = np.array(ds.ImagePositionPatient, dtype=float)
+            if origin is not None:
+                diff = file_origin - origin
+                origin_diff = origin_diff + diff
+            origin = file_origin
+            n_origins += 1
+    avg_origin_diff = tuple(origin_diff / n_origins)
```
cjacobs1: I think this should be (n_origins-1)

app/grandchallenge/cases/image_builders/dicom.py
```diff
@@ -167,49 +170,56 @@ def _process_dicom_file(dicom_ds):  # noqa: C901
...
+    z_i = avg_origin_diff[-1]
```
cjacobs1: -1 should be 2, otherwise it will fail for 4D or 2D data, right?

app/grandchallenge/cases/image_builders/dicom.py
```diff
@@ -167,49 +170,56 @@ def _process_dicom_file(dicom_ds):  # noqa: C901
...
         ds = pydicom.dcmread(str(partial["file"]))
+
         # Apply RescaleSlope and RescaleIntercept
         pixel_array = float(
             getattr(ds, "RescaleSlope", 1)
         ) * ds.pixel_array + float(getattr(ds, "RescaleIntercept", 0))
         if len(ds.pixel_array.shape) == dimensions:
             dcm_array = pixel_array
             break
+        z_index = index if z_i >= 0 else len(dicom_ds.headers) - index
```
cjacobs1: I think this should be len(dicom_ds.headers) - index - 1, because index is 0-based, right? otherwise, the last slice has index len(dicom_ds.headers) which 1 too large

cjacobs1: Also, use 0.0 instead of 0 to make clear that you are doing a float comparison

Issue #1226, opened 2020-03-24
Remove single polygon annotations


jmsmkn:
Do not work

HarmvZ:
Paul just notified me that this has been fixed in https://github.com/DIAGNijmegen/rse-cirrus-core/pull/547. Most likely by addition of this line: https://github.com/DIAGNijmegen/rse-cirrus-core/blob/2e1c55e392166131a54388522ac885c4c517233a/src/cirrus/DIAG/CIRRUSCore/Modules/Scripts/python/cirrus/workstations/core_web/reader_study/cso_serialization.py#L483

I will close all related PRs.

PR #1231, opened 2020-03-24
Remove references to single polygon

HarmvZ:
work in progress


HarmvZ:
This is not needed anymore. See https://github.com/comic/grand-challenge.org/issues/1226

PR #1234, opened 2020-03-24
Only prepend containing folder name when extracting zip

### LINKED ISSUE #1217: Zip of DICOMs fail with long filenames
In the zip unpacking the folder names are prepended to the filenames, which results in a filename being too long to store in the db.

MikeOverkamp-diag:
Closes #1217

### DIFF
--- a/app/grandchallenge/cases/tasks.py
+++ b/app/grandchallenge/cases/tasks.py
@@ -229,7 +229,7 @@ def fix_mhd_file(file, prefix):
         pass
 
 
-def extract_and_flatten(file, path, prefix=None, is_tar=False):
+def extract_and_flatten(file, path, prefix="", is_tar=False):
     """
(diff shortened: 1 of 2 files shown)


jmsmkn:
I guess that this doesn't (and never did) work for zips of mhd/(z)raw pairs in folders, right?

MikeOverkamp-diag:
> I guess that this doesn't (and never did) work for zips of mhd/(z)raw pairs in folders, right?

It does actually, the mhd files are updated here: https://github.com/comic/grand-challenge.org/pull/1234/files#diff-69480c5ccb3ebba148f554fd8cc3559cL206

Though I must admit I did not thoroughly test that part after this update, so it might be a good idea to do so once the pressure is off a bit.

jmsmkn:
I've made a note to follow this up

PR #1254, opened 2020-03-27
Updated email text for staf users when upload fails

miriam-groeneveld:
I changed the email that is sent to staf members when the upload of a container fails. 

Sorry, old commit comment

### DIFF
--- a/app/grandchallenge/container_exec/emails.py
+++ b/app/grandchallenge/container_exec/emails.py
@@ -9,9 +9,10 @@ def send_invalid_dockerfile_email(*, container_image):
     site = Site.objects.get_current()
 
     creator = container_image.creator
+    creator_username = creator.username if creator else "unknown"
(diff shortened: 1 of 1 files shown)


app/grandchallenge/container_exec/emails.py
```diff
@@ -22,17 +22,37 @@ def send_invalid_dockerfile_email(*, container_image):
         f"This is an automated service email from {site.domain}."
     )
 
-    recipients = list(get_user_model().objects.filter(is_staff=True))
+    staff_message = (
+        f"Dear {{}},\n\n"
```
jmsmkn: Using `{}` is ok with only 1 argument, but for 2 arguments to `format` could you use keywords? See "accessing arguments by name" here https://docs.python.org/3.8/library/string.html#format-examples

PR #1261, opened 2020-03-30
Adds overview pages

jmsmkn:
This PR adds overview pages that can be used to add extra items to the nav bar, linking together archives, algorithms, reader studies and challenges. It has a user customisable markdown at the top of the page, followed by cards:

Screenshot_20200330_154029


app/grandchallenge/core/templates/grandchallenge/partials/navbar.html
```diff
@@ -31,6 +31,14 @@
                         Reader Studies
                     </a>
                 </li>
+                {% for overview_page in overview_pages %}
+                    <li class="nav-item">
+                        <a class="nav-link {% if request.resolver_match.view_name == 'overview-pages:detail' %}active{% endif %}"
```
MikeOverkamp-diag: Wouldn't this make all nav items active on all overview pages?

app/grandchallenge/overview_pages/admin.py
```diff
@@ -0,0 +1,33 @@
...
+    class Meta:
+        model = OverviewPage
+        widgets = {
+            "detail_page_markdown": MarkdownEditorAdminWidget,
+        }
+        exclude = ("description",)
+
+
+class OverviewPageAdmin(MarkdownxModelAdmin):
+    can_add_related = False
```
MikeOverkamp-diag: Does this do something different from the widget keyword?

jmsmkn: Nope, left over by mistake, thanks!

app/grandchallenge/core/templates/grandchallenge/partials/navbar.html
jmsmkn: Ah yes, we don't do anything with active in the navbar anyway so I will remove it

jmsmkn: Actually kept it by looking up slug too. Can't use the url checks as we have this complex url reversal for subdomains.

PR #1263, opened 2020-03-31
Allow direct link requests

jmsmkn:


app/tests/serving_tests/test_views.py
```diff
@@ -76,12 +76,18 @@ def test_imageset_annotationset_download(
...
+@pytest.mark.parametrize("internal", (True, False))
+def test_image_response(client, internal):
     image_file = ImageFileFactory()
     user = UserFactory()
 
+    if internal:
+        data = {"internal": True}
+    else:
+        data = {}
+
```
MikeOverkamp-diag: A bit nitpicky, but you could save these lines by parametrizing with the `data` dict itself.

jmsmkn: Nope, the test will fail that way. We don't actually inspect the value of the `internal` query parameter, RFC2396 doesn't define how booleans should appear in uris, so in the view we only test for the presence of this key.

jmsmkn: This could be a bug, I guess, but I didn't want to do in to how we would define truthy values here.

MikeOverkamp-diag: Ah sorry, I was a bit too quick with my comment. I agree that this is the best way to go.

Issue #1013, opened 2019-11-28
Allow moving backwards in reader studies


MikeOverkamp-diag:
**Is your feature request related to a problem? Please describe.**
Currently it is not possible for a reader to move back to previous cases in a reader study. However, if the reader has accidentally entered a wrong answer, it would be useful to be able to change it.

**Describe the solution you'd like**
An option could be added to a reader study to indicate whether or not readers are allowed to move backwards. It should be optional, because in some reader studies answers to previous cases might become apparent once the reader advances to the next one.

jmsmkn:
We would also need to add a patch method to `Answers`. It would also be an idea to add object history to the answers when doing this. 

We should make a different issue for the work to the front end, this would be a significant change to support.

xarion:
Hello, is there any update on this? I find it quite important for my upcoming studies.

PR #1264, opened 2020-03-31
Allow patching of answer to facilitate backwards navigation

MikeOverkamp-diag:
#1013


app/grandchallenge/reader_studies/models.py
```diff
@@ -1299,3 +1311,4 @@ def assign_permissions(self):
             self,
         )
         assign_perm(f"view_{self._meta.model_name}", self.creator, self)
+        assign_perm(f"change_{self._meta.model_name}", self.creator, self)
```
jmsmkn, line 1314: Do we need to run a data migration to update the permissions of all answers?

app/grandchallenge/reader_studies/serializers.py
```diff
@@ -95,13 +95,23 @@ class AnswerSerializer(HyperlinkedModelSerializer):
...
 
     def validate(self, attrs):
-        question = attrs["question"]
-        images = attrs["images"]
-        answer = attrs["answer"]
+        question = attrs.get("question") or getattr(
+            self.instance, "question", None
+        )
+        images = attrs.get("images")
+        answer = attrs.get("answer") or getattr(self.instance, "answer", None)
```
jmsmkn, line 102: Do we limit what attributes are patchable? IMO it should only be the answer.

app/grandchallenge/reader_studies/models.py
```diff
@@ -1164,6 +1165,7 @@ class Answer(UUIDModel):
     answer = JSONField()
     is_ground_truth = models.BooleanField(default=False)
     score = models.FloatField(null=True)
+    renditions = HistoricalRecords(related_name="history")
```
jmsmkn, line 1168: Why not call this `history`, would be consistent with the other places that it's used.

app/grandchallenge/reader_studies/models.py
jmsmkn: Then if answer is the only thing that is patchable then that is the only field that should be included in `HistoricalRecord`

MikeOverkamp-diag:
Do not merge yet, the data migration is not correct

app/grandchallenge/reader_studies/migrations/0016_assign_permissions.py
```diff
@@ -0,0 +1,38 @@
...
+    except Permission.DoesNotExist:
+        # We are running tests
+        return
+    UserObjectPermission = apps.get_model(  # noqa: N806
+        "guardian", "UserObjectPermission"
+    )
+    ContentType = apps.get_model("contenttypes", "ContentType")  # noqa: N806
+    content_type_id = ContentType.objects.get_for_model(Answer).id
+
+    for answer in Answer.objects.all():
```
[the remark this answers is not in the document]
jmsmkn: See `migrate_images` for a base of how to do this

app/grandchallenge/reader_studies/serializers.py
```diff
@@ -95,15 +96,38 @@ class AnswerSerializer(HyperlinkedModelSerializer):
...
+
+        if self.instance:
+            add_scores.apply_async(
+                kwargs={
+                    "instance_pk": str(self.instance.pk),
+                    "pk_set": list(
+                        map(str, images.values_list("pk", flat=True))
+                    ),
+                }
+            )
```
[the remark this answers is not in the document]
MikeOverkamp-diag: This is what I thought at first as well, but the problem is that we cannot do that, because m2m relations get added after the save. So if we call this post save, there are no images set for the answer and we cannot calculate the score. We could do this for the update of course, but I felt this was a bit nicer than adding another signal handler. I'd be happy to change it if you disagree though.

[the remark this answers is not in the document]
jmsmkn: Thanks, wasn't clear to me. I agree, we should avoid signal handlers at all costs but it's impossible to do that for m2m relationships.

PR #1271, opened 2020-04-02
Account for different series in the same study in dicom importer

MikeOverkamp-diag:


app/grandchallenge/cases/image_builders/dicom.py
```diff
@@ -67,12 +68,16 @@ def _get_headers_by_study(path):
...
                     f, stop_when=pixel_data_reached
                 )
-                studies[ds.StudyInstanceUID] = studies.get(
-                    ds.StudyInstanceUID, {}
-                )
-                headers = studies[ds.StudyInstanceUID].get("headers", [])
+                key = f"{ds.StudyInstanceUID}-{ds.SeriesInstanceUID}"
+                studies[key] = studies.get(key, {})
+                index = series_indices.get(ds.SeriesInstanceUID)
+                if not index:
```
jmsmkn, line 74: Will this be a problem if index is zero? Don't you want to check for None instead?

jmsmkn: And can't we use this directly? https://dicom.innolitics.com/ciods/cr-image/general-series/00200011

MikeOverkamp-diag: I'm a bit worried about the last part here: `Required, Empty if Unknown (2)`. Especially with anonymization.

But yes, this would always start at 1, I'll change the check.

PR #1276, opened 2020-04-03
Make some tools configurable

miriam-groeneveld:
It concerns the flip, invert, reset and window level tools in the display plugin. These can now be configured in a workstation configuration.
The other tools are shown depending on the image shown. Slab rendering, slab thickness and orientation tools are only shown for 3D images. The overlay tool is only shown when an overlay is present. Therefor these are not configurable.


app/grandchallenge/workstation_configs/serializers.py
```diff
@@ -73,6 +73,10 @@ class Meta:
             "default_zoom_scale",
             "show_image_info_plugin",
             "show_display_plugin",
+            "show_invert_tool",
+            "show_flip_tool",
+            "show_window_level_tool",
+            "show_reset_tool",
```
jmsmkn, line 79: Actually, could you add a validator method here as these new things should only be set if `show_display_plugin` is also true.

jmsmkn: Ok I've had to merge this as is as there is a problem in CIRRUS otherwise.

PR #1277, opened 2020-04-03
Make ground truth error messages more descriptive

### LINKED ISSUE #1259: Make ground truth error message more descriptive
When uploading ground truth to a reader study the validation can fail on many points. A couple of important ones are the images not matching the hanging list, or the formatting being incorrect. When validation fails, an error message is shown, but it is not very descriptive. E.g. when the images do not match the hanging list, the error displayed is `Images provided do not match hanging protocol`. It would be a lot more useful to indicate where the file and the hanging list differ.

We should check if we can make the error more descriptive for all errors that might be raised.

MikeOverkamp-diag:
Closes #1259

### DIFF
--- a/app/grandchallenge/reader_studies/forms.py
+++ b/app/grandchallenge/reader_studies/forms.py
@@ -1,5 +1,6 @@
 import csv
 import io
+import itertools
 
 from crispy_forms.helper import FormHelper
 from crispy_forms.layout import (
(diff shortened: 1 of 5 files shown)


app/config/settings.py
```diff
@@ -588,6 +588,7 @@ def strtobool(val) -> bool:
 CELERY_BROKER_URL = os.environ.get("CELERY_BROKER_URL", "redis://redis:6379/0")
 CELERY_RESULT_BACKEND = os.environ.get("CELERY_RESULT_BACKEND", "django-db")
 CELERY_RESULT_PERSISTENT = True
+CELERY_TASK_ALWAYS_EAGER = True
```
jmsmkn, line 591: Shouldn't have been set, right?

PR #1288, opened 2020-04-08
1283 voxel spacing

### LINKED ISSUE #1283: Get spacing from mhd headers without loading image as SimpleITK object
The endpoint for getting the voxel spacing (`retina_api/views.py:ImageElementSpacingView`) opens an image as itk file to use the SimpleITK method `GetSpacing()` to retrieve the spacing. This means it has to read and decompress the whole file to identify the spacing. The performance for this endpoint can be hugely improved by getting the spacing by reading only the mhd header.

### LINKED ISSUE #1284: Spacing for retina images is not in voxel_(width|height|depth)_mm fields
At some point, the `voxel_(width|height|depth)_mm` fields have been added to the `Image` object. These are not set or used by retina images. This should be done by:
- Adding the fields to:
  - retina import scripts
  - retina image serializer
  - more?
- Creating a management command that iterates over all retina images and gets the spacing from the `ElementSpacing` or `ElementSize` mhd/mha header in and copies it into these fields.

HarmvZ:
- Adds `Image.get_mh_header()` method to the Image model that returns a key value dict of mhd/mha headers
- Adds `Image.spacing` property that returns the voxel spacing of an image by getting either the `voxel_(width|height|depth)_mm` fields, the `ElementSpacing` header or the `ElementSize` header.
- Adds `voxel_(width|height|depth)_mm` fields to `TreeImageSerializer`
- Fixes broken retina importer endpoint by adding request to context in `ArchiveSerializer`
- Sets `voxel_(width|height|depth)_mm` fields from mhd/mha headers in retina importer
- Adds a management command to copy voxel spacing data from mhd/mha headers to  `voxel_(width|height|depth)_mm` fields for all retina images
- Adds tests
- Adds a tag to identify sentry errors from legacy retina workstation

After merging this PR the command `copyspacingintoimagefields` can be run in production to set spacing fields for all retina images.

Closes #1283 
Closes #1284


app/grandchallenge/cases/models.py
```diff
@@ -358,6 +361,67 @@ def shape(self) -> List[int]:
...
         return result
 
+    @property
+    def spacing(self) -> List[float]:
+        """
+        Return the voxel spacing (or size if spacing is nonexistent) of the image.
+
+        Returns
+        -------
+            The voxel spacing in mm in SimpleITK ordering [x, y, (z)]
```
jmsmkn, line 371: Why does this return the ordering in a different ordering to shape? This will cause issues down the road, we should stick to 1 ordering for arrays. This is why we explicitally set width, height and depth.

app/grandchallenge/retina_core/management/commands/copyspacingintoimagefields.py
```diff
@@ -0,0 +1,53 @@
...
+        images = Image.objects.filter(
+            archive__title__in=[
+                "AREDS - GA selection",
+                "RS1",
+                "kappadata",
+                "Rotterdam_Study_1",
+                "Rotterdam Study 1",
+                "Australia",
+                "RS3",
+                "RS2",
```
jmsmkn, line 20: These will go soon, right? If not, move them into settings?

app/grandchallenge/cases/models.py
HarmvZ: I did this so it the order is equivalent to the output of SimpleITK's `Image.GetSpacing()` but indeed it is probably better to keep it consistent with shape and change it to NumPy ordering.

app/grandchallenge/retina_core/management/commands/copyspacingintoimagefields.py
HarmvZ: The filtering of archives on their name will be replaced with correct object level permissions in #1296. I probably won't change it in this command however since this is meant to be run once only.

jmsmkn: :+1:

PR #1296, opened 2020-04-10
Retina archive permissions

HarmvZ:
- Changes the `cache_archive_data()` task to only retrieve archives that the retina_grader users have permission for. (For legacy workstation endpoint)
- Changes the `ArchiveAPIView` to only return objects where the user has permissions to access the related archive. (For CIRRUS ophthalmology endpoint)
- Adapt/add tests to reflect and these changes and test if permissions are working correctly
- Adds a management command `setretinaarchivepermissions` which:
  - Creates the users/editors/uploads permission groups for all existing retina archives (if nonexistent)
  - Assigns view/upload/change permissions to the groups (by calling `archive.save()`
  - Adds all users from the `retina_graders` group as a user for each archive
  - Adds all users from the `retina_admins` group as an editor for each archive

The management command is meant to be executed once to set the correct permissions for all archives and users. I know all the tasks it performs can also be done manually in the django admin, but I think doing it like this is less error-prone.

~~It is best to first merge #1210. This will most likely result in merge conflicts which I will then fix before setting this PR to "Ready for review".~~


app/grandchallenge/retina_api/tasks.py
```diff
@@ -11,24 +12,20 @@
...
-    )
+    # This task creates the archive data object for the first user in the
+    # retina_graders group. The object does not have to be user specific
+    # currently because all retina_graders have access to all retina
+    # archives . This should be made user specific when different graders
+    # get different access rights.
+    user = Group.objects.get(
+        name=settings.RETINA_GRADERS_GROUP_NAME
+    ).user_set.first()
+    archives = get_objects_for_user(user, "archives.view_archive")
```
jmsmkn, line 28: Wouldn't it be better to use `get_objects_for_group` instead?

jmsmkn:
Also please resolve the merge conflicts, thanks!

app/grandchallenge/retina_api/tasks.py
HarmvZ: That wouldn't work because the `view_archive` permission is assigned to the users not the group.

jmsmkn: Hmm, I think that this is a little bit of a mess then. It's not a great practice to assign permissions direct to users as it is really hard to track. Why not assign permissions directly to the groups, either by giving the retina graders group permission to view an archive, and/or adding graders to the automatically generated archive users group?

HarmvZ: I realize now that my last comment was actually incorrect. The graders are added to the users group for each retina archive by the management command. So the `get_objects_for_user` actually finds the `view_archive` permissions that are assigned to the different retina archive users groups. I think I can use the `get_objects_for_group` but that would require me to first identify all the retina archives by explicitly finding them by name as is done in the management command. So I think the way I'm doing it now is the easiest way, do you agree? 

Sidenote: this task is only used for caching the data for the legacy workstation so this can hopefully be removed soon.

jmsmkn: I would have thought that:

```
retina_group = Group.objects.get(name=settings.RETINA_GRADERS_GROUP_NAME)
retina_archives = get_objects_for_group(retina_group, "archive.view_archive")
```

would achieve what you need. Right now, what happens if we assign the first user in the retina group permission to view another archive? I think it will be included a a retina archive the way it's done in the PR.

HarmvZ: I think that will not work since the `view_archive` permissions are set on the archive users group. So the `retina_graders` group itself does not have the permissions. 

But you are right, if the first user of the retina_graders group has permission for another archive then this will be included indeed which we do not want. I think the best way to change this is then to move this list to settings file as a constant: 
https://github.com/comic/grand-challenge.org/blob/bb3fd103822059cef2dd1ceb02384bb7038f0165/app/grandchallenge/retina_core/management/commands/setretinaarchivepermissions.py#L27-L34
And use it both in the management command and this task to specifically select those archives.

jmsmkn: Ah, I see, clear now, sorry! Yes, I agree with this approach then, fine to keep maintaining the list.

PR #1302, opened 2020-04-14
Allow ground truth explanation for educational reader studies

MikeOverkamp-diag:


jmsmkn:
I don't really understand what this functionality is for, is there an issue? I don't understand where the `foo_explanation` column comes from or is used for later.

MikeOverkamp-diag:
> I don't really understand what this functionality is for, is there an issue? I don't understand where the `foo_explanation` column comes from or is used for later.

Ah yes sorry, this is something that one of the radiologists I talked to about the education platform suggested. It allows editors to add some extra background to the ground truth provided, so for instance if a certain pathology is present, but it is not very apparent from the images, they could add some pointers to indicate why it is so.

The explanation will be added to the `verify answers` pop up in cirrus, I'll create a PR for that once this is merged.

jmsmkn:
Ok, and this is something that cannot be done with `case_text`? I guess that is always displayed.

MikeOverkamp-diag:
> Ok, and this is something that cannot be done with `case_text`? I guess that is always displayed.

It could, but this is more specific to a question and is only shown when answers are verified. Maybe it is a bit much to have multiple options to add some specific explanations, but I felt this was easy enough to add and it gives editors a bit more freedom to explain certain things.

app/tests/reader_studies_tests/resources/ground_truth.csv
```diff
@@ -1,4 +1,4 @@
-images,foo,bool,choice,mchoice
-im1,'"yes"','true','"option1"','["option1", "option2"]'
-im2,'"no"','false','"option1"','["option1", "option2"]'
-im3;im4,'"yes"','true','"option1"','["option1", "option2"]'
+images,foo,bool,choice,mchoice,foo_explanation
```
jmsmkn, line 1: Does this work if there is a space in the key? Should we use 2 underscores to be consistent with django lookups?

app/tests/reader_studies_tests/resources/ground_truth_new.csv
```diff
@@ -1,4 +1,4 @@
-images,foo,bool,choice,mchoice
-im1,'"no"','false','"option2"','["option0"]'
-im2,'"no"','false','"option1"','["option1", "option2"]'
-im3;im4,'"yes"','true','"option1"','["option1", "option2"]'
+images,foo,bool,choice,mchoice,foo_explanation
+im1,'"no"','false','"option2"','["option0"]',new explanation
```
jmsmkn, line 2: What happens when there is a comma in the explanation?

app/tests/reader_studies_tests/resources/ground_truth.csv
MikeOverkamp-diag: Spaces are fine. The double underscore is a great suggestion!

app/tests/reader_studies_tests/resources/ground_truth_new.csv
MikeOverkamp-diag: Good point, I'll add the same quotation scheme used in other fields here as well.

MikeOverkamp-diag:
This has become relevant again, as this option has also been requested for the other demo case for educational studies.

PR #1303, opened 2020-04-14
Fix prefixes in zip extraction

MikeOverkamp-diag:


jmsmkn:
Would you be able to add a test case that replicates the bug?

MikeOverkamp-diag:
> Would you be able to add a test case that replicates the bug?

I'll try!

PR #1316, opened 2020-04-21
Added openslide to support non-standard tiff files

### LINKED ISSUE #1309: Upload of Phillips tiff files fails
When uploading tiff files from a Philps scanner, there are a number of errors that occur while uploading. The files are also not shown properly by MeVisLab

miriam-groeneveld:
Closes #1309 

By adding openslide the creation of dzi for other tiff formats (like Phillips) is also supported.


app/grandchallenge/cases/image_builders/tiff.py
```diff
@@ -248,37 +244,48 @@ def image_builder_tiff(path: Path, session_id=None) -> ImageBuilderResult:
...
+        voxel_width_mm=tiff_file.voxel_width_mm,
+        voxel_height_mm=tiff_file.voxel_height_mm,
+        voxel_depth_mm=tiff_file.voxel_depth_mm,
     )
 
 
-def create_dzi_images(*, tiff_file: GrandChallengeTiffFile, pk) -> str:
+def create_dzi_images(
+    gc_file: GrandChallengeTiffFile, pk: UUID
+) -> (str, GrandChallengeTiffFile):
```
jmsmkn, line 270: Please go back to the keyword only function, I have a strong preference for this as it makes the code much more maintainable and grok-able, happy to discuss this decision.

miriam-groeneveld: Ok. For all functions or just the public ones?

jmsmkn: For both changed in this PR at least, may be more work to get the others to be keyword only and doesn't have priority right now

miriam-groeneveld: Not sure why the tests fail now, do I need to do something more than adding the libopenslide-dev to the dockerfile?

jmsmkn:
I would really keep `set -e`, this does not affect the tests (it would error out before the django tests started if this were a problem)

PR #1320, opened 2020-04-21
Allow null answers for annotation types

### LINKED ISSUE #1313: Allow null in annotation answers
For string answers we allow empty strings to represent that the question was skipped.

For annotations and choices the front end skips submitting an answer if skipped, causing this issue: https://github.com/DIAGNijmegen/rse-cirrus-core/issues/635

So, we need to represent a question being intentionally left blank, and the way to do this is to add optional `null` (JSON/in the db)/`None` (python) to the schema for the non-boolean & string answer types.

jmsmkn:
It seems that multiple choice answers are already represented with empty arrays, and users have added their own empty choice fields for single choice answers, so this had only been done for annotation objects.

`gcapi` will need to be updated for both GET (I think) and POST (certainly) to be able to create null answers. I will leave that as an exercise for @pkcakeout for when the frontend is updated.

We could have made heading objects answerable as null here, but this would probably need other changes both in grand challenge and CIRRUS to calculate scores and progress.

We might also need to do a data migration for Erdi's study which is affected by this issue.

Closes #1313


jmsmkn:
As long as you agree with the decisions re: headings, gcapi, and the data migration.

PR #1341, opened 2020-04-30
Add entrypoint for algorithm result in Cirrus Core

### LINKED ISSUE #1340: Provide entrypoint for algorithm results endpoint in Cirrus Core
An algorithm result can produce multiple overlays. At the moment, these can only be viewed one by one from Grand Challenge. It would be nice if there's a single entry to an algorithm result.

Related to https://github.com/DIAGNijmegen/rse-cirrus-core/issues/654

miriam-groeneveld:
Closes #1340


jmsmkn:
Are CIRRUS Core and the CORADS workstations ready to use this? We will need to deploy those first before adding this to master and deploying.

app/grandchallenge/algorithms/templates/algorithms/result_list.html
```diff
@@ -83,14 +83,16 @@ <h2>Results for {{ algorithm.title }}</h2>
...
-                                                <i class="fa fa-eye"></i> View Input &amp; Output
-                                            </span>
+                                    <a href="{% url 'workstations:workstation-session-redirect' slug=algorithm.workstation.slug %}?{% workstation_query algorithm_result=result config=algorithm.workstation_config %}">
+                                        <span class="badge badge-primary">
+                                            <i class="fa fa-eye"></i> View Result
+                                        </span>
                                     </a>
                                 </li>
+                            {% endif %}
+                            {% for image in result.images.all %}
```
jmsmkn, line 95: On line 113, currently: 
```
<a href="{% url 'workstations:workstation-session-redirect' slug=algorithm.workstation.slug %}?{% workstation_query image=result.job.image config=algorithm.workstation_config %}
```

This should be updated to use the result API too right?

jmsmkn: (even if there a no output images produced, maybe the metrics json is still interesting to view next to the image)

miriam-groeneveld:
Not yet. I could bring the "View input and output" back as a fallback.

jmsmkn:
Lets wait until CIRRUS is ready and deployed, then we can merge and deploy this relatively quickly

app/grandchallenge/algorithms/templates/algorithms/result_list.html
miriam-groeneveld: It's a bit strange as there's always at least one image expected. But can change that so CirrusCore can handle it.

pkcakeout:
I will give a :+1: for readding but also deprecating support for the `image=...&overlay=...` style. That will leave us more room to incrementally update 3 workstations where one contains a hack (CORADS)
.

jmsmkn:
> I will give a +1 for readding but also deprecating support for the `image=...&overlay=...` style. That will leave us more room to incrementally update 3 workstations where one contains a hack (CORADS)
> .

We absolutely have to keep supporting `image` and `overlay` in the frontend. It's used in a bunch of other places.

PR #1355, opened 2020-05-12
Rebuild folder structure for Mirax files.

### LINKED ISSUE #1332: Add support for zipped tiff files.
I would like to be able to upload all image formats supported by OpenSlide.
OpenSlide supports a number of image formats that consist of a collection of files like for example Hamamatsu-vms and Mirax (http://openslide.cs.cmu.edu/download/openslide-testdata/)

These are not yet supported on Grand Challenge.

miriam-groeneveld:
Gather related files for Hamamatsu-vms files.
Convert formats to tif (Aperio, Mirax, Hamamatsu, Hamamatsu-vms, Leica, Trestle, Ventana)

Not sure how to fix the code coverage without large testfiles.

Also not sure if large filenames will give issues, see the change in tasks.py. I now append all foldernames to the filename. Otherwise there's no certain way to link a file with a folder. Example (where Image1 folder contains data files for Image1.mrxs etc.):

Original structure:
```
Images
    Image1
        Image.mrxs
        Image
            ImageData1.dat
            ImageData2.dat
    Image2
        Image.mrxs
        Image
            ImageData1.dat
            ImageData2.dat
```
Would be unzipped to:
```
1-Image1-Image.mrxs
2-Image-ImageData1.dat
2-Image-ImageData2.dat
3-Image2-Image.mrxs
4-Image-ImageData1.dat
4-Image-ImageData1.dat

```
making it very hard to reconstruct the original structure.
Closes #1332


jmsmkn:
> Also not sure if large filenames will give issues, see the change in tasks.py. I now append all foldernames to the filename. Otherwise there's no certain way to link a file with a folder. Example (where Image1 folder contains data files for Image1.mrxs etc.)

Good that you point this out, indeed, this PR will cause a bunch of problems with DICOM imports as they have very deep folders, all with incredibly long names. We need to find a better way to do this. Why not preserve the folder structure? My guess is that it would break something else. I think that we need to spend some time improving the image importing module a lot, I don't like the design and it's proving really difficult to work with.

jmsmkn:
Could you add tests for this but skip them in CI unless the external test data folder is present (see the `skipif` decorator from pytest), then build up a folder of cases on one of the network shares for now. We can then expand it with DICOM data too.

miriam-groeneveld:
> > Also not sure if large filenames will give issues, see the change in tasks.py. I now append all foldernames to the filename. Otherwise there's no certain way to link a file with a folder. Example (where Image1 folder contains data files for Image1.mrxs etc.)
> 
> Good that you point this out, indeed, this PR will cause a bunch of problems with DICOM imports as they have very deep folders, all with incredibly long names. We need to find a better way to do this. Why not preserve the folder structure? My guess is that it would break something else. I think that we need to spend some time improving the image importing module a lot, I don't like the design and it's proving really difficult to work with.

For dicom the files should be in one folder (right, @MikeOverkamp-diag?). For Mirax, I rebuild the folder structure and then convert to tif, which works fine.

MikeOverkamp-diag:
> > > Also not sure if large filenames will give issues, see the change in tasks.py. I now append all foldernames to the filename. Otherwise there's no certain way to link a file with a folder. Example (where Image1 folder contains data files for Image1.mrxs etc.)
> > 
> > 
> > Good that you point this out, indeed, this PR will cause a bunch of problems with DICOM imports as they have very deep folders, all with incredibly long names. We need to find a better way to do this. Why not preserve the folder structure? My guess is that it would break something else. I think that we need to spend some time improving the image importing module a lot, I don't like the design and it's proving really difficult to work with.
> 
> For dicom the files should be in one folder (right, @MikeOverkamp-diag?). For Mirax, I rebuild the folder structure and then convert to tif, which works fine.

No sorry, I forgot to mention that :(. The way you do it now is what I did initially as well, but that created filenames that were way too long for dicom images, as they are generally three folders deep and all folders generally have very long names themselves. So while with this the dicom files are still extracted to a single folder, their individual filenames would become too long.

miriam-groeneveld:
I suggest I'll rewrite it so we extract to the original files names as much as possible, completely drop the prepending of folder names and in stead preserving the original file structure in a dict? That way we retain all information and original filenames (which is nice for the user when something goes wrong) and avoid long  names

jmsmkn:
Sounds fine, but be careful if you have the same file name in 2 different folders, I think there are some assumptions about the uniqueness of input filenames.

Why does the structure need to be flattened anyway? Why not extract each input zip file to its own sub directory, and then walk over all of the files, and then carry results over via the fullpath to the file rather than only the filename?

miriam-groeneveld:
I'm nor sure, I thought for dicom it had to be in a flattened structure. Maybe because of the very long folder/file names?

jmsmkn:
On disk it would not be a problem on Linux, we don't need to worry about windows support. IIRC the problem is in how the case builder keeps track of, and iterates over, input files.

miriam-groeneveld:
Refactored extraction method so original folder structure is restored. Each algorithms handle files in folders and algorithms only handle files that were not handled by other algorithms already. 
This allowed simplification of the handling of complex pathology files.

app/grandchallenge/cases/image_builders/tiff.py
```diff
@@ -214,52 +250,196 @@ def _add_folder_uploads(*, dzi_output, image, new_folder_upload):
...
+                f
+                for f in ini_file.parent.iterdir()
+                if f.name.lower() == ini_file.name.lower()
+            ][0]
+        with open(ini_file, "r") as f:
+            lines = [
+                line for line in f.readlines() if re.match(mirax_pattern, line)
+            ]
+            for line in lines:
+                original_name = line.split("=")[1].strip()
```
MikeOverkamp-diag, line 274: Could there be lines without `=` in them? E.g. an empty line at the end of the file? In that case it might be safe to wrap this in a `try.. except IndexError`.

app/grandchallenge/cases/image_builders/tiff.py
```diff
@@ -214,52 +250,196 @@ def _add_folder_uploads(*, dzi_output, image, new_folder_upload):
...
+
+def _compile_vms(files: List[Path], converter) -> List[GrandChallengeTiffFile]:
+    def get_filenames_from_vms(vms_file: Path):
+        file_matched = []
+        with open(str(vms_file.absolute()), "r") as f:
+            lines = [
+                line for line in f.readlines() if re.match(vms_pattern, line)
+            ]
+            for line in lines:
+                original_name = line.split("=")[1].strip()
```
MikeOverkamp-diag, line 316: same as above

app/grandchallenge/cases/image_builders/tiff.py
```diff
@@ -214,52 +250,196 @@ def _add_folder_uploads(*, dzi_output, image, new_folder_upload):
...
+    for file in files:
+        try:
+            gc_file = GrandChallengeTiffFile(file)
+            vms_files = get_filenames_from_vms(gc_file.path)
+            tiff_file = _convert_to_tiff(
+                path=file, pk=gc_file.pk, converter=converter
+            )
+        except Exception:
+            continue
+        else:
```
MikeOverkamp-diag, line 331: You could possibly create a separate function for this, as you repeat this bit of code 3 times (with some very minor variations). This is of course a bit nitpickey.

app/tests/cases_tests/test_background_tasks.py
```diff
@@ -312,133 +307,98 @@ def test_mhd_file_annotation_creation(settings, format):
...
...345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345/123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345/123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345/123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345/123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890/t.png",
```
MikeOverkamp-diag, line 366: Wow! Nice little python trick: to repeat a string you can actually use `*`, so for instance `"a" * 4` would become `"aaaa"`.

jmsmkn: Probably not the longest line in our codebase, please don't look at `legacy_luts.py` :see_no_evil:

app/grandchallenge/cases/image_builders/tiff.py
miriam-groeneveld: No, not possible actually, the regex makes sure of that.

app/grandchallenge/cases/image_builders/tiff.py
miriam-groeneveld: same :)

app/tests/cases_tests/test_background_tasks.py
miriam-groeneveld: ah, that would make it a bit easier to read :)

app/grandchallenge/cases/image_builders/tiff.py
miriam-groeneveld: good point

MikeOverkamp-diag:
LGTM, I don't know if @jmsmkn still wants to have a look, otherwise we can merge this.

app/grandchallenge/cases/image_builders/tiff.py
```diff
@@ -18,27 +20,38 @@
...
     def validate(self):
         if not self.image_width:
-            raise ValidationError("ImageWidth could not be determined")
+            raise ValidationError(
+                "Not a valid tif: ImageWidth could not be determined"
+            )
         if not self.image_height:
-            raise ValidationError("ImageHeigth could not be determined")
+            raise ValidationError(
+                "Not a valid tif: ImageHeigth could not be determined"
```
jmsmkn, line 41:
```suggestion
                "Not a valid tif: ImageHeight could not be determined"
```

app/grandchallenge/cases/image_builders/tiff.py
```diff
@@ -18,27 +20,38 @@
...
     voxel_height_mm: float = 0
     voxel_depth_mm: Optional[float] = None
+    source_files: List = field(default_factory=list)
+    associated_files: List = field(default_factory=list)
 
     def validate(self):
         if not self.image_width:
-            raise ValidationError("ImageWidth could not be determined")
+            raise ValidationError(
+                "Not a valid tif: ImageWidth could not be determined"
```
jmsmkn, line 37:
```suggestion
                "Not a valid tif: Image width could not be determined"
```

app/grandchallenge/cases/tasks.py
```diff
@@ -229,9 +230,24 @@ def fix_mhd_file(file, prefix):
...
+        if info.issym() or info.islnk():
+            raise ValidationError("(Symbolic) links are not allowed.")
+    # Also check for max path length, drop folders when path is too long
+    filename_attr = "name" if is_tar else "filename"
+    while len(os.path.join(path, getattr(info, filename_attr))) > 4095:
+        filename = getattr(info, filename_attr)
+        filename = os.path.join(
+            *Path(filename).parts[1 : len(Path(filename).parts)]
+        )
+        setattr(info, filename_attr, filename)
```
jmsmkn, line 245: Style wise: split these out into 2 more functions rather than using comments. Fine for now.

app/grandchallenge/cases/image_builders/tiff.py
```diff
@@ -174,38 +208,40 @@ def _create_image_file(*, path: str, image: Image):
...
+    )
+    if gc_file.source_files:
+        for s in gc_file.source_files:
+            new_image_files.append(_create_image_file(path=s, image=image))
     return new_image_files
 
 
-def _add_folder_uploads(*, dzi_output, image, new_folder_upload):
+def _add_folder_uploads(
+    *, dzi_output: str, image: Image, new_folder_upload: List
```
jmsmkn, line 243: Aside: When you add a `List` annotation add the `List[<type>]`. No real benefit right now as we are not running mypy on the codebase, I'm unsure if we should even be adding type hints right now without static checking (https://mypy.readthedocs.io/en/stable/existing_code.html#start-small)

> If your codebase is large, pick a subset of your codebase (say, 5,000 to 50,000 lines) and run mypy only on this subset at first, _without any annotations_.

Emphasis theirs.

app/grandchallenge/cases/tasks.py
```diff
@@ -287,44 +279,43 @@ def check_compressed_and_extract(file_path, target_path, index, prefix=None):
...
-        file_path.unlink()
-    for file in new_files:
-        if file["path"].name.endswith(".mhd") and file["prefix"]:
-            fix_mhd_file(file["path"], file["prefix"])
-        index = check_compressed_and_extract(
-            file["path"], target_path, index, prefix=file["prefix"]
-        )
-    return index
+
+    def do_extraction(file_path):
```
jmsmkn, line 286: Nitpick on style: `do_extration` => `extract`

app/grandchallenge/cases/tasks.py
```diff
@@ -387,8 +378,8 @@ def build_images(upload_session_uuid: UUID):
             upload_session.status = upload_session.FAILURE
             upload_session.save()
             return
-        except Exception:
-            upload_session.error_message = "An unknown error occurred"
+        except Exception as e:
+            upload_session.error_message = f"An unknown error occurred {e}"
```
jmsmkn, line 382:
```suggestion
        except Exception:
            upload_session.error_message = f"An unknown error occurred"
```

Undo this, the exception could contain private data that is then emailed to the user. The raise below will mean that we get the traceback in sentry.

app/grandchallenge/cases/tasks.py
```diff
@@ -428,16 +425,20 @@ def _handle_raw_image_files(tmp_dir, upload_session):
         )
 
         for filename in algorithm_result.consumed_files:
-            if filename in unconsumed_filenames:
-                unconsumed_filenames.remove(filename)
-                raw_image = filename_lookup[filename]  # type: RawImageFile
+            if filename in unconsumed_files:
```
jmsmkn, line 428: Is this still a filename we're iterating over or a file path?

app/grandchallenge/cases/image_builders/tiff.py
```diff
@@ -214,52 +250,155 @@ def _add_folder_uploads(*, dzi_output, image, new_folder_upload):
...
+    for file in files:
+        try:
+            gc_file = GrandChallengeTiffFile(file)
+            if associated_files_getter:
+                associated_files = associated_files_getter(gc_file.path)
+            tiff_file = _convert_to_tiff(
+                path=file, pk=gc_file.pk, converter=converter
+            )
+        except Exception:
+            continue
```
jmsmkn, line 307: Should we really silence all errors here? We should be explicit about the errors we expect and silencing those is ok, but 
I think that we should have some way of knowing when something unexpected is happening.

miriam-groeneveld:
Found an issue with large filenames. Because the folder structure gets restored, the RawImageFile's filename is no longer correct. It's the relative path now, and could be longer than the max_length of 255 now specified in the model. Can I increase the max_length?

jmsmkn:
> Found an issue with large filenames. Because the folder structure gets restored, the RawImageFile's filename is no longer correct. It's the relative path now, and could be longer than the max_length of 255 now specified in the model. Can I increase the max_length?

I _think_ so, we did this before and didn't hit any side effects. For most filesystems the maximum filename length is 255 chars. Some older POSIX APIs limit the `PATH_MAX` to 4096 bytes. So I would set it to the latter as we don't ever use the filename on disk (we use the pk). I know that we display the filename in several tables, so maybe we need to handle that. 

https://en.wikipedia.org/wiki/Comparison_of_file_systems#Limits

PR #1365, opened 2020-05-20
Extract intended window level from dicom tags

MikeOverkamp-diag:


app/grandchallenge/cases/image_builders/dicom.py
```diff
@@ -238,6 +238,11 @@ def _process_dicom_file(dicom_ds, session_id):  # noqa: C901
     img.SetSpacing(sitk_spacing)
     img.SetOrigin(sitk_origin)
 
+    if "WindowCenter" in ref_file:
+        img.SetMetaData("WindowCenter", str(ref_file.WindowCenter))
+    if "WindowWidth" in ref_file:
+        img.SetMetaData("WindowWidth", str(ref_file.WindowWidth))
+
```
jmsmkn, line 245: Extend `OPTIONAL_METADATA_FIELDS` with `"WindowCenter"` and `"WindowWidth"` rather than adding specific handlers.

MikeOverkamp-diag:
Rebased on master and regenerated migrations.

PR #1369, opened 2020-05-25
Use floats for window level values

### LINKED ISSUE #1367: Cannot import DICOM with non-integer window levels.
I think you can have non int window and level, please also check for the MHD loader
Dicom image builder: invalid literal for int() with base 10: '1906.5'

MikeOverkamp-diag:
Closes #1367 

No need for `try.. except ValueError`, as any non-numerical values get caught by this regex: https://github.com/comic/grand-challenge.org/blob/master/app/grandchallenge/cases/image_builders/metaio_utils.py#L71


jmsmkn:
Could you update the tests to include a window center with a 'something.5' value so that we have a test to replicate what happened in prod?

MikeOverkamp-diag:
Not sure what happened with codecov, the diff seems fine.

jmsmkn:
> Not sure what happened with codecov, the diff seems fine.

Yeah, I take those with a large grain of salt, it's why they're not required.

PR #1383, opened 2020-06-04
Workstation region assignments

### LINKED ISSUE #888: workstations:session-create post fails when the container image is not ready
See https://sentry.io/organizations/grand-challenge/issues/1163103558/?project=303639&query=is%3Aunresolved&statsPeriod=14d

### LINKED ISSUE #1382: Workstation Session create view always uses default region
Add a check on `Session.save()` if adding and region is None to set the region to one that is in use (or closest to the user).

jmsmkn:
This PR:

- Removes the workstation session redirect view, instead uses the create view which is doing a get or create for a running session for that user in the requested region
- Adds region assignment to the create view
- Adds a ping endpoint with CORS for the main domain to nginx
- Automatically determine the region

This PR does not let you override the region, but on failure will launch in the default region. Overrides can follow later if needed (I will use a vpn for now).

Fixes #888 
Closes #1382

### DIFF
--- a/app/grandchallenge/workstations/static/workstations/js/session-region.js
+++ b/app/grandchallenge/workstations/static/workstations/js/session-region.js
@@ -1,7 +1,14 @@
+function setModalLoadingMessage(msg) {
+    document.getElementById("sessionStateMsg").innerHTML = msg;
+}
+
 $(document).ready(function () {
+    const session_modal = $("#sessionModal");
+
+    session_modal.modal({show: true});
+
     const ping_endpoint = JSON.parse(document.getElementById("ping-endpoint-data").textContent);
     const region_selection = document.getElementById("id_region");
-
     const urls = [...region_selection.options].map(function (o) {
         return `https://${o.value}.${ping_endpoint}`
     });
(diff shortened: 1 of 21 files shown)


app/grandchallenge/workstations/static/workstations/js/session-region.js
```diff
@@ -0,0 +1,46 @@
...
+    session_modal.modal({show: true});
+
+    const ping_endpoint = JSON.parse(document.getElementById("ping-endpoint-data").textContent);
+    const region_selection = document.getElementById("id_region");
+    const urls = [...region_selection.options].map(function (o) {
+        return `https://${o.value}.${ping_endpoint}`
+    });
+
+    Promise.race(
+        urls.map(
```
HarmvZ, line 17: This map could be combined with the `map()` on line 12

app/grandchallenge/workstations/static/workstations/js/session-region.js
```diff
@@ -0,0 +1,46 @@
...
+                )
+                .catch(
+                    // Nothing resolved
+                    error => console.log(error)
+                )
+        )
+    )
+        .then(
+            response => {
+                if (response !== undefined) {
```
HarmvZ, line 32: Unfortunately it is still better to use `typeof response !== 'undefined'` to support <ES2015 browsers.

app/grandchallenge/workstations/static/workstations/js/session-region.js
```diff
@@ -0,0 +1,46 @@
...
+                    // If we have a match, find out what server this is for
+                    let regexp = /^https:\/\/([\w-]+).*/;
+                    let m = response.url.match(regexp);
+
+                    if (m !== undefined) {
```
HarmvZ, line 37: `typeof m !== 'undefined'`

app/grandchallenge/workstations/static/workstations/js/session-region.js
```diff
@@ -0,0 +1,46 @@
+function setModalLoadingMessage(msg) {
+    document.getElementById("sessionStateMsg").innerHTML = msg;
```
HarmvZ, line 2: It is generally better to avoid `innerHTML` to prevent XSS attacks unless you explicitly want to add child elements by text. An XSS attack is not possible here, but still, it is good practice to use `textContent` or `innnerText` here since you are only setting text.

app/grandchallenge/workstations/static/workstations/js/session-region.js
```diff
@@ -0,0 +1,46 @@
...
+                    error => console.log(error)
+                )
+        )
+    )
+        .then(
+            response => {
+                if (response !== undefined) {
+                    // If we have a match, find out what server this is for
+                    let regexp = /^https:\/\/([\w-]+).*/;
+                    let m = response.url.match(regexp);
```
HarmvZ, line 35: Can be `const`.

(This comment dit not get submitted during the review somehow)

PR #1386, opened 2020-06-04
Move algorithm result pagination and search to backend

MikeOverkamp-diag:


app/grandchallenge/algorithms/templates/algorithms/data_tables/result_list.html
```diff
@@ -0,0 +1,90 @@
+{% load url %}
+{% load profiles %}
+{% load workstations %}
+{% load guardian_tags %}
+{% load pathlib %}
+{% load humanize %}
+{% load evaluation_extras %}
+
+{{ result.job.created|naturaltime }}
+<td>
```
jmsmkn, line 10: Shouldn't the `<td>` tags be closed?

app/grandchallenge/algorithms/templates/algorithms/data_tables/result_list.html
```diff
@@ -0,0 +1,90 @@
+{% load url %}
+{% load profiles %}
+{% load workstations %}
+{% load guardian_tags %}
+{% load pathlib %}
+{% load humanize %}
+{% load evaluation_extras %}
+
+{{ result.job.created|naturaltime }}
```
jmsmkn, line 9: Maybe this is answered later but how does ordering work on these data cells now? Aren't things like `data-order="{{ result.job.created|date:"U" }}` lost now?

app/grandchallenge/algorithms/views.py
```diff
@@ -386,43 +389,54 @@ def get_context_data(self, *args, **kwargs):
...
+        "job__creator__username",
+        "job__image__name",
+        "visibility",
+        "images__files__file",
+        "comment",
+    ]
+    order_by = "job__created"
+
+    def get_row_context(self, result, *args, **kwargs):
+        checker = ObjectPermissionChecker(self.request.user)
```
jmsmkn, line 415: How many queries does this result in? One for each row? Where has `prefetch_perms` gone?

app/grandchallenge/algorithms/templates/algorithms/result_list.html
```diff
@@ -54,97 +54,6 @@ <h2>Results for {{ algorithm.title }}</h2>
...
-                               data-toggle="modal"
-                               data-target="#resultInfoModal"
-                               data-title="Result Comment"
-                               data-output="{{ result.comment }}"
-                               data-pk="{{ result.pk }}"
-                               title="Result Comment">
-                                <i class="fa fa-file"></i>
-                            </a>
-                        {% endif %}
-                        {% if change_result|get_jsonpath:result.pk %}
```
jmsmkn, line 138: `|get_jsonpath:result.pk` has been lost from the refactoring, are you sure that the permissions check is correct? It could be that a user only has permission to change one of the results.

app/grandchallenge/algorithms/views.py
```diff
@@ -386,43 +389,54 @@ def get_context_data(self, *args, **kwargs):
...
+    ]
+    columns = [
+        "job__created",
+        "job__creator__username",
+        "job__image__name",
+        "visibility",
+        "images__files__file",
+        "comment",
+    ]
+    order_by = "job__created"
```
jmsmkn, line 412: Will this add a sort on unixtime?

app/grandchallenge/algorithms/templates/algorithms/data_tables/result_list.html
```diff
@@ -0,0 +1,90 @@
...
+       class="badge badge-primary"
+       data-toggle="modal"
+       data-target="#resultInfoModal"
+       data-title="Result Info"
+       data-output="{% firstof result.output 'No output produced.' %}"
+       data-pk="{{ result.pk }}"
+       title="Result Info">
+        <i class="fa fa-file"></i> {{ result.job.image.name }}
+    </a>
+<td>
```
jmsmkn, line 23: How would we add `data-order` to this column? At the moment, sorting by visibility (in production) does not work.

app/grandchallenge/algorithms/templates/algorithms/data_tables/result_list.html
jmsmkn: Oh, I see, there is a split later. Should some other identifier be used instead? This confuses PyCharm/VSCode.

MikeOverkamp-diag: I agree, this is a bit hacky. I'll think of a cleaner way to do this.

app/grandchallenge/algorithms/templates/algorithms/result_list.html
MikeOverkamp-diag: The check is now added on a row by row basis. The rows are formatted one by one, so `change_result` now only contains the value for this particular entry

app/grandchallenge/algorithms/views.py
MikeOverkamp-diag: That's a good question, I'll look into it.

app/grandchallenge/algorithms/views.py
MikeOverkamp-diag: Good point, I did not check the queries, I'll have a look with the django debug toolbar.

app/grandchallenge/algorithms/templates/algorithms/data_tables/result_list.html
MikeOverkamp-diag: The ordering is handled in the view. Datatables just sends an index for the column to sort on, so the columns need to coincide with what is defined in the view. Visibility sorting is working I think, but I'll check to be sure.

jmsmkn: Yes, but `data-order` tells data tables how to sort, so it's important to set. Without it, it will sort on the contents of the cell which are human readable, so goes completely wrong for dates and doesn't work for visibility right now as data tables does not know how to sort icons. This is why we had `data-order="{{ result.job.created|date:"U" }}"` so that it can sort on a unixtime, but display `{{ result.job.created|naturaltime }}` a natural time for the user.

MikeOverkamp-diag: The sorting is now done on the database using the values defined in the `columns` attribute in the view. So the `data-order` is not used anymore.

jmsmkn: Oh, I see! Even better.

app/grandchallenge/algorithms/views.py
MikeOverkamp-diag: It just sorts on the datetime field in postgres, but that effectively achieves the same.

jmsmkn: Nice, yes, this is much nicer!

app/grandchallenge/algorithms/views.py
MikeOverkamp-diag: Added it back in,

app/grandchallenge/algorithms/templates/algorithms/data_tables/result_list.html
MikeOverkamp-diag: I replaced it with a custom `<split/>` tag, it's self closing, so it shouldn't cause any issues with IDEs I'd say.

app/grandchallenge/algorithms/views.py
```diff
@@ -411,14 +411,22 @@ class AlgorithmResultsList(PermissionListMixin, PaginatedTableListView):
     ]
     order_by = "job__created"
 
-    def get_row_context(self, result, *args, **kwargs):
-        checker = ObjectPermissionChecker(self.request.user)
+    def get_row_context(self, result, *args, checker=None, **kwargs):
+        if not checker:
+            checker = ObjectPermissionChecker(self.request.user)
```
jmsmkn, line 416: Just to double check, is this line ever actually hit? If so, could there be a better way to persist the checker that has had `prefetch_perms` applied?

MikeOverkamp-diag: It's not, I just included it because the default for the kwarg is None.

MikeOverkamp-diag: I actually tested it before adding the if statement and it worked fine

jmsmkn: Ok, great! Then maybe make it a required kwarg? `checker=None,` -> `checker,`, then you can remove the if block.

PR #1387, opened 2020-06-05
Store the measured ping times

jmsmkn:
This data will be useful to determine where to run new instances.


app/grandchallenge/workstations/static/workstations/js/session-region.js
```diff
@@ -3,15 +3,30 @@ function setModalLoadingMessage(msg) {
...
-    let end = null;
+    let errored = false;
+
+    // First try establishes TLS
+    $.ajax({
+        url: url,
+        async: false,
+        cache: false,
+        success: ()=>{},
+        error: function () {
```
MikeOverkamp-diag, line 14: I'd say either use an arrow function here, or make the success function a normal `function()`

PR #1400, opened 2020-06-12
Adds GDCM

### LINKED ISSUE #1293: DICOM importer should return a nicer error if the transfer syntax is not supported
Would be nice to get gdcm in here, otherwise, raise a proper error if the syntax is not supported: https://pydicom.github.io/pydicom/stable/old/image_data_handlers.html

jmsmkn:
This PR adds GDCM for importing DICOM with compressed image data (JPEG 2000, etc, see https://pydicom.github.io/pydicom/stable/old/image_data_handlers.html?highlight=transfer#supported-transfer-syntaxes).

This will likely add quite some time to the build processed, we should separate out the dependency installation in another PR.

Closes #1293


jmsmkn:
Don't know why the build is failing right now, worked locally but might be related to this https://github.com/pypa/setuptools/issues/2017

dockerfiles/web/Dockerfile
```diff
@@ -3,37 +3,69 @@
...
+        libfreetype6-dev \
+        libxft-dev \
+        libffi-dev \
+        wget \
+        gettext \
+        libopenslide-dev \
+        libvips-dev \
+    && mkdir -p /opt/gdcm/build \
+    && cd /opt/gdcm \
+    && git clone --branch v3.0.6 --depth 1 git://git.code.sf.net/p/gdcm/gdcm \
```
MikeOverkamp-diag, line 24: Should we add an issue to switch to master when possible?

jmsmkn: I don't think so - it's best to pin our library dependencies. I could not work out if GDCMs master was always installable, but this is how they do releases. We could move it to a build arg, but, this is also where we update python so I don't know, maybe tidy it up when splitting the container builds from the Django tests as these both take ~15 minutes now.

MikeOverkamp-diag: Ah yes of course, if this is their official release branch, then it would actually be better than using master.

Issue #1354, opened 2020-05-11
Algorithm Interfaces


jmsmkn:
At the moment the interface to algorithms is rigid:

Input:
- 1 image

Output:
- A results json (arbitrary schema)
- 0...n images

We need to better define the inputs and outputs, and make this flexible per algorithm, whilst still supporting the original interface. This would allow us to set up pipelines and opens up visualization options.

jmsmkn:
Algorithms are now using the default ComponentInterfaces. The next step is to get the evaluations using this too.

jmsmkn:
The interfaces are implemented and ready to be used, will move any requests to new issues.

PR #1410, opened 2020-06-22
Adds interface definition to algorithms

jmsmkn:
This PR allows us to flexibly define inputs and outputs to algorithms. It adds:

- Component interfaces: these define the interface to a component (a container batch job). 
- Component interface values: these encapsulate the value of the interface at a certain point in the pipeline.

*NOTE that this does not allow for flexible definition of inputs and outputs yet*, so the method (predictions in, metrics json out) and algorithms container (image in, json and image(s) out) API has not changed. That will follow later. However, it does change the internal model for Algorithms:

- The `Algorithm.Result` model will be removed. This is one-to-one related to the `Algorithm.Job` anyway so saves a join. `comment` and `public` have been migrated to `Algorithm.Job`.
- The biggest change is that that public API of the Algorithm Results now needs to change, this is only used in CIRRUS and I will update separately:
  - `job.image` (fk to an image) -> `job.inputs` (m2m to component interface values, for now, this inputs will only contain one component interface value object: `generic-medical-image`)
  - `result.images` (m2m to images) -> `job.outputs` (m2m to component interface values, with the images stored in a `generic-overlay` instance)
  - `result.output` (json field) -> `job.outputs` (m2m to component interface value, one of these will be an instance of `results-json-file` where `job.outputs[idx].value` is the json dump).
- For CIRRUS: `algorithmResult` has been replaced with `algorithmJob`

Susequently - any results views have been replace by job views.

What is not covered here:
- Removing the evaluation Results model
- Loading images in the same celery process
- Interface value validation
- Changes to the API to work with multiple inputs and outputs
- Defining interfaces that are different from the default
- Using multiple inputs and outputs

#1354


pkcakeout:
>     * For CIRRUS: `algorithmResult` has been replaced with `algorithmJob`

There is a tiny issue that I do not know how to solve immediately. There are a few workstations right now taht we do not (directly) control anymore: The CORADs workstation and the 3D-brain-thing of Silvan and Ajay come to mind immediately. We need a proper migration strategy for those projects. Is it maybe possible to have "deprecated" support for the `algorithmResults` endpoint? A time where we serve both? Then everybody can update their code to migrate to the new endpoint. Less work for us. I reckon this issue will get bigger and bigger in the future as we amass more legacy workstations from students.

jmsmkn:
> > ```
> > * For CIRRUS: `algorithmResult` has been replaced with `algorithmJob`
> > ```
> 
> There is a tiny issue that I do not know how to solve immediately. There are a few workstations right now that we do not (directly) control anymore: The CORADs workstation and the 3D-brain-thing of Silvan and Ajay come to mind immediately. We need a proper migration strategy for those projects. Is it maybe possible to have "deprecated" support for the `algorithmResults` endpoint? A time where we serve both? Then everybody can update their code to migrate to the new endpoint. Less work for us. I reckon this issue will get bigger and bigger in the future as we amass more legacy workstations from students.

No, it is something that I tried to do but in the end it is _way_ less work to update those workstations. I plan on making PRs for that. Django REST Framework is a massive pain to work with once it diverges from your database schema, same with Django Guardian that manages permissions around Jobs and Results, there just wasn't a good way to do it.

app/grandchallenge/components/models.py
```diff
@@ -2,19 +2,142 @@
...
+    default_value = JSONField(
+        null=True,
+        default=None,
+        help_text="Default value for this field, only valid for inputs.",
+    )
+    kind = models.CharField(
+        blank=False,
+        max_length=4,
+        choices=Kind.choices,
+        help_text="What kind of field is this interface?",
```
miriam-groeneveld, line 81: I find the word interface in the help text a bit confusing, isn't it the entity's kind?

app/grandchallenge/components/serializers.py
```diff
@@ -0,0 +1,40 @@
...
+    kind = serializers.CharField(source="get_kind_display", read_only=True)
+
+    class Meta:
+        model = ComponentInterface
+        fields = [
+            "title",
+            "description",
+            "slug",
+            "kind",
+            "pk",
```
miriam-groeneveld, line 21: default_value not needed?

app/tests/algorithms_tests/test_models.py
```diff
@@ -81,3 +82,33 @@ def test_algorithm_job_update_status():
...
+def test_default_interfaces_created():
+    a = AlgorithmFactory()
+
+    assert {i.kind for i in a.inputs.all()} == {InterfaceKindChoices.IMAGE}
+    assert {o.kind for o in a.outputs.all()} == {InterfaceKindChoices.JSON}
+
+
+@pytest.mark.django_db
+def test_outputs_are_set():
+    j = AlgorithmJobFactory()
```
miriam-groeneveld, line 97: not used I think?

app/tests/cases_tests/test_permissions.py
```diff
@@ -54,36 +55,38 @@ def test_permissions(self, user, access):
...
-    assert "view_image" in get_perms(g_reg_anon, result.job.image)
+    job.public = True
+    job.save()
 
-    result.public = False
-    result.save()
+    assert "view_image" not in get_perms(g_reg, output_image)
+    assert "view_image" in get_perms(g_reg_anon, output_image)
+    assert "view_image" not in get_perms(g_reg, job.inputs.first().image)
+    assert "view_image" in get_perms(g_reg_anon, job.inputs.first().image)
```
miriam-groeneveld, line 81: just for my own understanding, making a job public doesn't mean everyone can view the images? Only the users in the registered and anonymous group?

app/grandchallenge/algorithms/management/commands/migrate_algorithm_results.py
```diff
@@ -0,0 +1,60 @@
...
+                            for im in result.images.all()
+                        ]
+                    )
+                    job.create_result(result=result.output)
+                    job.image = None
+                    job.comment = result.comment
+                    job.public = result.public
+                    job.save()
+                else:
+                    print(f"Skipping job {job.pk} as there is no input image.")
```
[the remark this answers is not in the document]
jmsmkn: This command is for migrating the old jobs that have an input image - and once the values are copied over to the new inputs and outputs we set the legacy input image to None. This helps us track which Jobs have been migrated. The Null input image is new with this PR, so all existing Jobs currently have an input image. Once this command has been run manually I will make another PR to remove all of the Results objects, but all data will have been preserved on the Job component interface values.

app/grandchallenge/algorithms/templates/algorithms/data_tables/job_list.html
```diff
@@ -0,0 +1,95 @@
...
+<ul class="list-unstyled">
+    <li>
+        <a href="{% url 'workstations:workstation-session-create' slug=algorithm.workstation.slug %}?{% workstation_query algorithm_job=job config=algorithm.workstation_config %}">
+            <span class="badge badge-primary"
+                  title="View algorithm result">
+                <i class="fa fa-eye"></i> View Result
+            </span>
+        </a>
+    </li>
+    {% for output in job.outputs.all %}
```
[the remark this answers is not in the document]
jmsmkn: Maybe! Users have usually uploaded the images themselves so already have these, it's the outputs that they're interested in first, but maybe lets swap things around in another PR.

app/grandchallenge/algorithms/templates/algorithms/executionsession_list.html
```diff
@@ -59,10 +59,11 @@ <h2>Experiments for {{ algorithm.title }}</h2>
                     <td>
                         <ul class="list-unstyled">
                             {% for image in object.image_set.all %}
-                                {% for job in image.job_set.all %}
-                                    <li>
-                                        <a href="#jobInfoModal"
-                                           class="badge
+                                {% for civ in image.componentinterfacevalue_set.all %}
+                                    {% for job in civ.algorithms_job_inputs.all %}
```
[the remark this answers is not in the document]
jmsmkn: It returns jobs, maybe the name of this relation can be improved. It is "the set of jobs that this component interface value is used as an input for". It is auto-generated with `"%(app_label)s_%(class)s_inputs"` on the `ComponentInterfaceValue` m2m relationship. Default the name would be `algorithms_job_set`, but we need to distinguish between inputs and outputs.

app/grandchallenge/components/models.py
jmsmkn: Agreed, can you think of a better way to phrase this? I wanted to use `type` rather than `kind`, but this is a reserved word in python.

app/grandchallenge/components/serializers.py
jmsmkn: Not yet! Will add it when we start to use them.

app/tests/algorithms_tests/test_models.py
jmsmkn: It's there because you want to test that you're picking up the correct interface (in the first implementation I was updating the same CIV for each job, oops!). I will add a comment.

app/tests/cases_tests/test_permissions.py
jmsmkn: Yes - reg_anon is everyone who is registered plus anonymous users. This was requested as people want to show results without having to sign up.

app/grandchallenge/algorithms/templates/algorithms/executionsession_list.html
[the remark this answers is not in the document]
miriam-groeneveld: Ok, so maybe `"%(app_label)s_%(class)s_as_inputs"` would be clearer, at least in what you can expect to be returned?

app/grandchallenge/algorithms/signals.py
```diff
@@ -2,56 +2,79 @@
...
     else:
-        algorithm_results = [instance]
+        jobs = [instance]
         if pk_set is None:
             # When using a _clear action, pk_set is None
             # https://docs.djangoproject.com/en/2.2/ref/signals/#m2m-changed
-            images = instance.images.all()
+            component_interface_values = getattr(
+                instance, forward_lookup
+            ).all()
```
MikeOverkamp-diag, line 53: couldn't you filter these for `image__isnull=False`? That way you'd have a smaller queryset and wouldn't need the if statement on line 69.

jmsmkn: Great, thanks!

app/grandchallenge/components/models.py
miriam-groeneveld: I guess for the help text we could use type right? so maybe "What kind (or type) of field does this interface describe?"

PR #1414, opened 2020-06-25
Aiforrad text changes

Kicksta:


app/grandchallenge/products/static/products/css/main.css
```diff
@@ -33,3 +33,14 @@ a:hover {
   max-width: 100%;
   max-height: 80px;
 }
+
+ul {
+    padding: 0;
+    font-size: 0;
+}
+ul li {
+    list-style-type: disc !important;
```
jmsmkn, line 42: Just a note on css: try and find ways to do things without using `!important`, it becomes an arms race once you introduce it. A better approach is usually to use more specific selectors.

PR #1415, opened 2020-06-26
Remove dependency on upload session from the image builders

### LINKED ISSUE #1406: Remove dependency on upload session from image import
For #1354 I would like to import a bunch of files in a folder that have not originated from an upload session. We should separate out the celery task and remove the dependency on the upload session so that this will work, and can be called at the end of a component job. The API would be a path to a folder in, returning a set of `Image` objects.

jmsmkn:
This PR removes the dependency on the upload session from the image builders, tidies up the API to use sets, and created a `import_images` method that can be used to create images from a set of files that already exist, outside of the upload session infrastructure. 

I do not have the pathology test data so it would be super if someone could run the test suite with that.

Closes #1406


app/grandchallenge/cases/image_builders/tiff.py
```diff
@@ -409,26 +414,21 @@ def format_error(message):
...
 
-        new_folder_upload = _add_folder_uploads(
-            dzi_output=dzi_output,
-            image=image,
-            new_folder_upload=new_folder_upload,
-        )
-        new_images.append(image)
-        consumed_files.append(gc_file.path)
         if gc_file.associated_files:
             consumed_files += list(f for f in gc_file.associated_files)
```
miriam-groeneveld, line 424: This needs to be converted to append set as well

jmsmkn: Thanks, done!

app/grandchallenge/cases/tasks.py
```diff
@@ -487,17 +493,31 @@ def _handle_image_relations(*, collected_images, upload_session):
...
 ):
+    unconsumed_files = input_files - consumed_files
+
     errors = []
-    for unconsumed_filepath in unconsumed_files:
-        raw_file = filepath_lookup[str(unconsumed_filepath)]
-        error = raw_file.error or ""
+
+    for filepath in consumed_files:
+        raw_image = filepath_lookup[str(filepath)]
```
miriam-groeneveld, line 509: When files are converted to tif, the newly created file is added to consumed files but of course do not exist in the filepath_lookup. I'm afraid I didn't add a test to check for that, sorry about that.

jmsmkn: Hmm - should they really be added to consumed files then? The images are still in the new_images and the associated files will be in new_image_files, right?

jmsmkn: Rather, are the two lines here (L423, L424) actually needed at all? https://github.com/comic/grand-challenge.org/pull/1415/commits/4af11423a99b02af501d4e1bae8ca02df4387d71

miriam-groeneveld: yes, the associated files hold the source files that were used to create the tiff file. So file.mrxs and a bunch of .data files get converted to a single tif. It's the line above that adds the tiff to the consumed files. At that point, there's no check if the gc_file was a converted tif (in that case it doesn't need to be added) or original tif (then it should be added or it will be regarded as failed). I think this change should fix that: 
```
        if gc_file.associated_files:
            consumed_files |= set(f for f in gc_file.associated_files)
        else:
            consumed_files.add(gc_file.path)
```

jmsmkn: Great! That did the trick, we should add a test using the existing VMS and Mirax files in the resources directory.

jmsmkn: I ran 7809668 with individual zips of the complex formats and with convert_to_tiff.zip and everything worked well.

PR #1416, opened 2020-06-29
Remove results table

### LINKED ISSUE #1397: Indicate that an image import is pending for an algorithm result
If the image import queue is full, algorithm results can show up in the results list with the note "No output images were found" even if the algorithm produced an output image because the import is delayed. This changes once the result image has finally been imported, but until then the user has the impression that the algorithm failed. It would be nicer if the results page would indicate that the import of the results is not finished.

jmsmkn:
This PR:

- Removes the algorithm results model as everything is now stored on the job
- Moves any algorithm  image importing to the component job rather that through indirection via an upload session
- Runs assigning images to related objects to a chained celery task (`linked_task`)

The relations from an upload session to archives/reader studys/algorithm images is still there, but could be removed in the future if we can think of ways to improve the UX with listing upload sessions for the related objects.

Fixes #1397
#1354


jmsmkn:
See #1417 regarding read the docs

app/grandchallenge/algorithms/models.py
```diff
@@ -355,48 +334,37 @@ def _copy_output_files(self, *, container, base_dir: Path):
...
-                client_id=self._job_id,
-                client_filename=file.name,
-                file_id=new_uuid,
-                timeout=timezone.now() + timedelta(hours=24),
-                start_byte=0,
-                end_byte=django_file.size - 1,
-                total_size=django_file.size,
+            importer_result = import_images(
+                files=input_files,
+                builders=[image_builder_mhd, image_builder_tiff],
```
MikeOverkamp-diag, line 355: Why just these two image builders?

app/grandchallenge/submission_conversion/models.py
```diff
@@ -92,49 +87,33 @@ def _copy_output_files(self, *, container, base_dir: Path):
...
-
-            images = []
-
-            for file in output_files:
-                new_uuid = uuid.uuid4()
+                for file in output_files:
+                    tmpfile = Path(
+                        safe_join(tmpdir, file.relative_to(base_dir))
+                    )
+                    tmpfile.parent.mkdir(parents=True, exist_ok=True)
```
MikeOverkamp-diag, line 98: This is the exact same code that is used in algorithm/models.py, correct? If so, could it not be consolidated into a single utility function?

app/grandchallenge/algorithms/models.py
jmsmkn: For the algorithm image output we only want to support mha/mhd and tiff files.

app/grandchallenge/submission_conversion/models.py
jmsmkn: This will eventually be consolidated once the Component Interfaces are used for real, rather than the loose API we have now.

PR #1428, opened 2020-07-08
Aiforrad text class changes

Kicksta:
because of windows-docker issues haven't been able to properly review the looks of the CE/FDA class changes on the products page.


app/grandchallenge/products/templates/products/partials/product_list.html
```diff
@@ -34,10 +34,14 @@
       <div class="col-sm-3 bg-light pt-2">
         <div class="d-flex justify-content-between flex-column h-100">
           <div>
-            <div class="h6 text-muted d-flex justify-content-between">CE:
+            <div class="d-flex justify-content-between">
+              <div class="h6 text-muted mr-auto pb-1">CE: </div>
+              <div class="h6 pr-2">{% if product.ce_class %}{{product.ce_class}}{% endif %}</div>
```
MikeOverkamp-diag, line 39: Do you still want the html element here, even if it's empty? It adds padding and margin I think from the `h6` class, which could be either desirable or not and I'm not sure which is the case here.

Kicksta: Dont think it matters. Because the 'CE:' and icon stay in the same place whatever the situation is. And otherwise there is just blanc padding in between...?

MikeOverkamp-diag: Sure, then all looks good to me!

PR #1437, opened 2020-07-10
Cache the test container image, separate docs build

### LINKED ISSUE #1436: Speed up the tests
At the start of this week the tests were taking 30+ minutes, which was far too long with efficient development. We need to find a way to speed them up. With parallelisation and caching the base image we're now down to ~16 minutes. Find where we can gain some wins.

jmsmkn:
This PR adds caching of the test image, separates the docs build and uses simple staticfile serving in tests, taking another 200s off each build.

Closes #1436 as we're getting in to the territory of diminishing returns.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,9 +45,6 @@ jobs:
       - name: Build the containers
         run: |
           make -j2 build
-      - name: Build the docs
-        run: |
-          make docs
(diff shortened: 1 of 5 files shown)


jmsmkn:
Note that the PR build will be slow right now as it's having to build the base container again. Will be <3 minutes after this has been deployed to master.

Issue #1327, opened 2020-04-23
Replace lesion type item list with new lesion types


HarmvZ:
A new lesion item list has been supplied. I need to:
- Replace the old list, both in CIRRUS Ophthalmology and the legacy workstation
- Split up the lists for 2d (enface) images and 3d (OCT) images
- Create a data migration to change all the old values to the new ones according to the list I received
Only support the following features in CIRRUS Ophthalmology:
  - Freetext/ICD9 lesion annotation option
  - Checkbox (boolean) lesion specific annotations
  - OCT Retinal layer segmentation/annotation

PR #1441, opened 2020-07-14
Adds BooleanClassificationAnnotation viewset, url and tests

HarmvZ:
The model was already existing but the viewset was not created yet. This is needed to add checkbox lesion annotations as described in #1327


HarmvZ:
I forgot to define `filterset_fields` in the viewset to make use of the `DjangoFilterBackend`. Added it now!

PR #1443, opened 2020-07-31
Image builder for NifTI image file format

### LINKED ISSUE #1344: Add support for .nii and .nii.gz images
This would be a useful feature since this format can be easily read and written with SimpleITK, has similar features to mha/mhd and is quite popular.

nlessmann:
Here my attempt to add an image builder for images with .nii or nii.gz file extension (NifTI format).

Closes #1344


app/grandchallenge/cases/image_builders/nifti.py
```diff
@@ -0,0 +1,61 @@
...
+    new_images = set()
+    new_image_files = set()
+    consumed_files = set()
+    for file in files:
+        if not (file.name.endswith(".nii") or file.name.endswith(".nii.gz")):
+            continue
+
+        try:
+            img: SimpleITK.Image = SimpleITK.ReadImage(str(file.absolute()))
+        except RuntimeError:
```
jmsmkn, line 41: All of the exceptions that `ReadImage` can raise should be handled here. Is `RuntimeError` the only one that is possible?

app/tests/cases_tests/test_nifti.py
```diff
@@ -0,0 +1,47 @@
...
+
+import pytest
+
+from grandchallenge.cases.image_builders.nifti import image_builder_nifti
+from grandchallenge.cases.models import Image
+from tests.cases_tests import RESOURCE_PATH
+
+
+def float_close(f1: float, f2: float) -> bool:
+    return abs(f1 - f2) < 0.0001
```
jmsmkn, line 13: Please use `pytest.approx` https://docs.pytest.org/en/5.4.1/reference.html#pytest-approx

.gitignore
```diff
@@ -15,6 +15,7 @@
 *.7z
 *.dmg
 *.gz
+!*.nii.gz
```
jmsmkn, line 18: This line can be removed as you have explicitly added the test file.

nlessmann:
> * If the nifti file contains metadata (patient id, study date, etc) is this also preserved? There is a list `ADDITIONAL_HEADERS` that we expect to keep for meta images.

The nifti format does not support additional metadata.

.gitignore
nlessmann: Done

app/tests/cases_tests/test_nifti.py
nlessmann: Neat, didn't know about this!

app/grandchallenge/cases/image_builders/nifti.py
nlessmann: As far as I know that's the only type of exception raised by SimpleITK. The SimpleITK documentation is not very clear about this unfortunately.

nlessmann:
Something that I am not sure about is how vulnerable SimpleITK is for decompression attacks, so somebody uploading a very large image that is all zeros and that therefore is a small file when saved as compressed .nii.gz.

app/tests/cases_tests/test_nifti.py
jmsmkn: No problem, this is what code review is for!

app/grandchallenge/cases/image_builders/nifti.py
jmsmkn: Lets see how it goes, the downside is that one bad nifti file could mean that the whole import is cancelled in which case the user will get an "unexpected error".

jmsmkn:
> Something that I am not sure about is how vulnerable SimpleITK is for decompression attacks, so somebody uploading a very large image that is all zeros and that therefore is a small file when saved as compressed .nii.gz.

This is ok - this code is only ever executed asynchronously and the workers run in a container with memory limits, so jobs will be killed if this limit is exceeded. The killed worker then recovers automatically.

PR #1457, opened 2020-08-11
Add annotation migration command

HarmvZ:
- Adds `migratelesionnames` management command which migrates polygon annotation sets from old lesion names to new names
- Adds tests for the command
- Improves `PolygonAnnotationSet` admin

After merging this PR we can run the command which will change the annotation name of all matching annotations. The command will output the ids of all unmatched annotations and I will discuss what to do with these annotations with the graders.

Related issue #1327


jmsmkn:
Could you rebase this on master?

PR #1465, opened 2020-08-14
Add PhotometricInterpretation, ImagerPixelSpacing, and PixelSpacing field

### LINKED ISSUE #1464: Feature request: adding Pixel Spacing and Photometric Interpretation as optional DICOM/MHA headers
Some time ago, I created an issue for preserving the (Imager) Pixel Spacing attribute in a DICOM file (x-ray) uploaded to Grand Challenge (https://github.com/comic/grand-challenge.org/issues/1216). Pixel spacing is the distance between adjacent pixels in the patient, whereas imager pixel spacing is the distance between adjacent pixels on the sensor. I use the pixel spacing information to resize images that are put into my deep learning algorithm. After realizing that the pixel spacing could still be accessed in the output MHA file using the GetSpacing function from the SimpleITK Python library, I closed the issue. However, I found that in quite some cases the pixel spacing does not match the imager pixel spacing, and GetSpacing only returns the imager pixel spacing. Since the Pixel Spacing attribute accounts for corrections or calibrations, I prefer this attribute over the Imager Pixel Spacing attribute (if available).

Besides the Pixel Spacing attribute, I use the Photometric Interpretation attribute to modify pixel intensity values. If the photometric interpretation is MONOCHROME1 (minimum pixel value should be displayed as white), then the pixel intensities need to be inverted. I saw that this attribute is not in the list of additional DICOM headers that are preserved by Grand Challenge, so I assume that this attribute will be lost after uploading the DICOM file.  

If possible, I would like that the Pixel Spacing attribute (tag: 0028,0030) and Photometric Interpretation attribute (tag: 0028,0004) are added as additional headers to the output MHA file.

I have considered using the Imager Pixel Spacing attribute only and instructing users to upload MONOCHROME2 DICOM files.

njmhendrix:
Closes #1464


jmsmkn:
There are several files that you need to add the headers too. Also, the test values that you have put for `ImagerPixelSpacing` and `PixelSpacing` will not pass the `FLOAT_LIST_MATCH_REGEXP`, are these really the format of the values that you would get in the DICOM header?

njmhendrix:
Okay, I assume that I have to add the headers to all "extra-stuff" mhd files? ImagerPixelSpacing and PixelSpacing are lists of floats when I access them using the pydicom library, but perhaps SimpleITK reads them as tuples. I will try another regular expression.

app/grandchallenge/cases/image_builders/metaio_utils.py
```diff
@@ -55,7 +55,12 @@
 
 STUDYDATE_MATCH_REGEXP: Pattern = re.compile(r"^\d{4}\d{1,2}\d{1,2}$")
 
+PIXELSPACING_MATCH_REGEXP: Pattern = re.compile(r"^(\d*\.\d+|\d+)\\(\d*\.\d+|\d+)$")
```
jmsmkn, line 58: This regex does not match the definition of DS: http://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_6.2.html

`[-+]?\d*\.?\d+([eE][-+]?\d+)?\\[-+]?\d*\.?\d+([eE][-+]?\d+)?` is probably what you want.

Also, you need to run `black` on this file to format the code correctly.

njmhendrix: Okay, I just pushed a new change before reading your reply, I will format the code correctly next time.

njmhendrix: The test just failed, I will change the pattern to the one you suggested. It seems that I looked at the wrong error log before writing my new pattern.

njmhendrix: It finally works, all checks have passed. Thanks for the help!

bramvanginneken:
I strongly doubt if we should store photometric interpretation as a tag in mha (is this what this ticket is about?). If so it would be logical to also require our software (ie CIRRUS to handle this correctly have displaying. For pixel spacing, this is already available in the header. Anyway, I was not involved in this discussion and it seems to have been added now. Within DIAG, staff and data team, were discussing which dicom headers to store outside of mha header (or all in it, not decided yet) because they may be needed in certain cases and we want to move away from the DICOM format in general as it tends to give issues with privacy (We are now instructed to delete all COPDGene DICOM files because of some privacy issues in private headers. It is allowed to convert to other file formats, it is not allowed to remove the private headers and keep using the DICOM data, I expect such issues to arise more often).

jmsmkn:
> I strongly doubt if we should store photometric interpretation as a tag in mha (is this what this ticket is about?). If so it would be logical to also require our software (ie CIRRUS to handle this correctly have displaying. For pixel spacing, this is already available in the header. 

See the discussion in #1464 for why pixel spacing was not sufficient for Niels needs.

> Within DIAG, staff and data team, were discussing which dicom headers to store outside of mha header (or all in it, not decided yet) because they may be needed in certain cases and we want to move away from the DICOM format in general as it tends to give issues with privacy.

This is exactly why we only pass through a subset of the headers. The three added here are not going to contain any PII data.

PR #1496, opened 2020-08-28
Carousel color visibility

Kicksta:


app/grandchallenge/products/static/products/css/main.css
```diff
@@ -44,3 +44,16 @@ ul li {
...
     margin-bottom: 0.5em;
 }
+
+.carousel-indicators li {
+    background-color: #00afdc;
+    opacity: 0.2;
+  }
+
+.carousel-control-next-icon {
+    background-image: url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' fill='%2300afdc' width='8' height='8' viewBox='0 0 8 8'%3e%3cpath d='M2.75 0l-1.5 1.5L3.75 4l-2.5 2.5L2.75 8l4-4-4-4z'/%3e%3c/svg%3e");
```
jmsmkn, line 54: Where did you get these SVGs from? We need to check the licensing before we can add it

app/grandchallenge/products/static/products/css/main.css
```diff
@@ -44,3 +44,16 @@ ul li {
...
+}
+
+.carousel-control-prev-icon {
+    background-image: url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' fill='%2300afdc' width='8' height='8' viewBox='0 0 8 8'%3e%3cpath d='M5.25 0l-4 4 4 4 1.5-1.5L4.25 4l2.5-2.5L5.25 0z'/%3e%3c/svg%3e");
+}
```
jmsmkn, line 59: Please add a newline to the end of the file

app/grandchallenge/products/static/products/css/main.css
Kicksta: It is an overwrite of the bootstrap carousel component. It was not trivial to be able to change the color of the icon as it is an image. Fora told me do it by changing the fill color in the url pointing towards the image. https://getbootstrap.com/docs/4.0/components/carousel/#with-controls If you inspect the icons here, you see back the same url. Would that be fine?

jmsmkn: Ok - in this case we need to distribute the license with the modifications. 

- Make a new folder `app/grandchallenge/products/static/products/css/bootstrap/`
- Move `carousel-control-next-icon` and `carousel-control-prev-icon` to a new css file in there
- Include that css file in the templates where it is needed (search for html files that use `products/css/main.css` and add it after that)
- Add the bootstrap license to `app/grandchallenge/products/static/products/css/bootstrap/LICENSE`. You can find it here: https://github.com/twbs/bootstrap/blob/main/LICENSE

jmsmkn (approved):

PR #1504, opened 2020-09-02
Add configuration to allow reader study reader to show all annotation…

miriam-groeneveld:
…s of a case

Closes #1504


jmsmkn (approved):
LGTM - we should tidy up the reader study create fields as this is getting to be a lot. In cases like this the create form should show the minimum needed to create the object, and any extra things can be added to the update form later.

Issue #682, opened 2019-01-24
make results hidden


bramvanginneken:
Every page has permission, so you can set it to only registered users. But the Results page is an automatically created page, see e.g. https://lola19.grand-challenge.org/evaluation/results/

Is it possible to also make this page visible only to registered users. Would also be nice if you can make one setting to set all pages to visible to registered users only.

prash-p:
I second this request :)

jmsmkn:
This has now been deployed - if a challenge is set to `hidden` then the published evaluations will only be visible to the challenges participants group.

PR #1509, opened 2020-09-03
Adds guardian permissions to all evaluation models

jmsmkn:
This PR updates the evaluations app to set guardian permissions. For each model:

- The challenge admins group can view all `Phase`s, `Method`s, `Submission`s, `AlgorithmEvaluation`s and `Evaluation`s for their challenge
- The challenge admins group can change all `Phase`s and `Evaluation`s for their challenge
- The submission creator can view their `Submission`s
- If an `Evaluation` is published, and the `Challenge` is hidden, then only the challenge participants group can view it
- If an `Evaluation` is published, and the `Challenge` is not hidden, then anyone can view it

The permissions will need to be added to the existing objects before the views can be updated to use these.

#682

### DIFF
--- a/app/grandchallenge/challenges/models.py
+++ b/app/grandchallenge/challenges/models.py
@@ -23,6 +23,7 @@ from grandchallenge.challenges.emails import (
     send_external_challenge_created_email,
 )
 from grandchallenge.core.storage import public_s3_storage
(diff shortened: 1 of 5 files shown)


MikeOverkamp-diag:
Late review, but LGTM!

jmsmkn:
Thanks! Sorry to barge through with these things, deadlines looming...

PR #1520, opened 2020-09-10
Adds challenge forums

### LINKED ISSUE #678: Add challenge forums
I would be nice if each challenge could have its own forum where participants can post messages and receive updates, and admins can post announcements.

jmsmkn:
This PR adds a forum per challenge using django-machina. Features:

- Automatically creates a forum for each challenge
- Challenge admins can moderate and create announcements in their challenge forums
- Only participants and admins of that challenge can create new posts or reply in a forum
- If a challenge is hidden, the forum is only visible to that challenge's participants and admins
- If a challenge is not hidden, the forum and posts are visible to everyone
- Available at `https://gc.localhost/forums/`

`django-machina` creates a hierarchical structure of forums using `mptt`, so we could group forums by challenge series, etc. It is really nice to keep this business logic up to date using django, but, customising django-machina is quite complex as they use a lot of magic imports. I have changed 2 things: redirecting any member detail views to our existing profile pages, and disabling setting a signature in a forum. Both required creating new apps with specific names and starred imports, the templates are not namespaced correctly so the `APPS_DIRS: True` setting no longer worked (you need to load a template directory first). Now we have to maintain the migrations for those apps in our codebase. It also includes django haystack for searching, this package is not maintained and only a beta version is available for django 3, which we have to install and then silence. This should be replaced with postgres full-text search. I think that we're ultimately going to need to fork this package for our own needs, which is fine, this is OSS.

What is not included but would be nice:

- Email alerts
- Subscriptions to entire forums
- The challenge forum available on the challenge subdomain
- More powerful moderation, waiting time before posting, sending users to Coventry, etc.

Closes #678


app/grandchallenge/forums/templates/board_base.html
```diff
@@ -0,0 +1,28 @@
...
+{% endblock %}
+
+{% block js %}
+<script src="{% static 'js/machina.min.js' %}" type="text/javascript" charset="utf-8"></script>
+<script type="text/javascript">
+  $(function() {
+    machina.init();
+    {% block onbodyload %}{% endblock onbodyload %}
+  });
+  console.log("js");
```
MikeOverkamp-diag, line 26: I think this could be removed.

MikeOverkamp-diag (approved):
This doesn't look nearly as hacky as I would have expected based on what you described yesterday. Just a lot of extra migrations.

On very minor comment.

app/grandchallenge/forums/templates/board_base.html
jmsmkn: Haha, oops

PR #1521, opened 2020-09-10
Adds key bindings configuration

### LINKED ISSUE #1497: Keybindings store for workstations

We have a key bindings editor in cirrus, but no way of persisting modified keybindings. We need some sort of API end point to store this information

We have two ideas to be considered:
- Store presets as part of workstation configurations
- Store workstation presets for users

Ultimately, I assume, that we want both, but we should pick one as a first solution. My preference would be to allow users to store custom keybinding presets, because we have a ready-to-be-used keybindings editor in cirrus that we can leverage without too much work.

Internally, cirrus uses three parameters to describe an action - an action is something that a shortcut can be bound to:
- action-name: A unique symbolic name for the action
- description: A user-readable string describing the action the keybinding triggers
- shortcuts: A list of strings describing the bound key combinations. Example: 
    - `["Ctrl+A", "B,A"]` would mean two shortcuts are assigned to an action. One shortcut is Ctrl+A, the other is a sequence of two key presses, namely "B" followed by an "A".

The available actions and keybindings differ per loaded application, however, I think we can consider action names to be unique because they are typically namespaced with the plugin name. So my suggestion would be to create a "keybindings" JSON field on users that CIRRUS can use to store alternative keybindings for workstations. The keybindings then could be edited right within CIRRUS and stored in grand-challenge. On loading, the keybindings list will be retrieved from grand-challenge and applied.

Because action names are generally shared across our applications, this would allow users to set personal shortcuts that apply to many workstations. I think this is a nice feature.

I do not know if the user is known to CIRRUS at this point however. Is the GRAND_CHALLENGE_AUTHORIZATION-key a per-user key?

Alternatively, an easier but less convenient solution might be to add another JSON structure to a workstation configuration. This will enable shortcut configuration on an algorithm or reader study level. This is little different from the feature I envisioned but it also allows some modifications of shortcuts. I think this is less nice, because keyboard shortcuts cannot be shared among CIRRUS applications, cannot be customized by a user, and therefore we cannot leverage the builtin keybindings editor in CIRRUS which was designed around the idea of shortcuts being assignable by _users_.

Maybe this is easier to implement however?

------------------

Another way would be to store the keybindings in a browser-local storage: https://www.w3schools.com/html/html5_webstorage.asp This would be by-far- the easiest way to persist keybindings, but this would only work on a browser+user basis. If a users moves to a different device or computer, keybindings will not transfer as well. This could be a nice feature in itself however. I am not super sure but I think that I would expect keybindings to carry over with my website profile.

**Additional context**

The CIRRUS side of this feature is ready to go for months now. This was repeatedly requested by Bram, so this might be worth picking up....

Also: I asked about a 1-2 day project in a different domain in chat. Maybe this would be a good candidate? After we decided _what_ the solution should be?

jmsmkn:
Adds keybinding configuration with the same schema as used in VSCode - a list of objects with a mandatory `key` and `command` and optional `when`, all strings. Serialises the json to the workstation config endpoint.

Closes #1497


app/grandchallenge/workstation_configs/models.py
```diff
@@ -73,6 +73,56 @@
...
+    "$schema": "http://json-schema.org/draft-06/schema",
+    "$id": "http://example.com/example.json",
+    "type": "array",
+    "title": "The Key Bindings Schema",
+    "description": "Define the key bindings for the workstation.",
+    "items": {
+        "$id": "#/items",
+        "type": "object",
+        "title": "The Key Binding Schema",
+        "description": "Defines a key binding for a command.",
```
pkcakeout, line 86: Copy-pasta mistake? Repeated above.

app/grandchallenge/workstation_configs/models.py
```diff
@@ -73,6 +73,56 @@
...
+                "default": "",
+                "examples": ["ctrl+shift+k"],
+            },
+            "command": {
+                "$id": "#/items/properties/command",
+                "type": "string",
+                "title": "The Command Schema",
+                "description": "The command called by this binding.",
+                "default": "",
+                "examples": ["editor.action.deleteLines"],
```
pkcakeout, line 112: We use hyphons as separators now but well... not really important. Example:
https://github.com/DIAGNijmegen/rse-cirrus-core/blob/e89c0542b3e9b79bb64c0843547b9d415cc21585/src/cirrus/DIAG/CIRRUSCore/Modules/Macros/ToolboxPlugin/ToolboxPlugin.py#L157

pkcakeout (approved):
Alright, LGTM!

app/grandchallenge/workstation_configs/models.py
jmsmkn: That's ok, the values are not checked, example was from the VSCode documentation.

Issue #1073, opened 2020-01-08
Add a publications app


jmsmkn:
Users should be able to create publications and attach them to algorithms and challenges (many to many). See the existing challenge models and CITATION.cff for what fields should be included. List all of the publications in the top navbar.

Also remove the old publication fields #1482

jmsmkn:
Some issues with this
- [ ] arxiv links do not work as they do not provide a DOI, probably we need a separate model for pre-prints
- [ ] Parsing of consortium names does not work as they do not provide a family name

jmsmkn:
All existing publications have been migrated

PR #1533, opened 2020-09-17
Adds a publications model

### LINKED ISSUE #1482: error updating google scholar id of challenge publication
image

This is a strange error, but my proposal is to remove the google scholar id and citations editboxes altogether as we are not using them.

jmsmkn:
This PR adds a publications app where the user will input a DOI and we automatically fetch the citeproc JSON information from doi.org, which allows us to automatically generate an ama citation and citation count. We index the title, year and citation count (provided by cross ref) for later use. Challenges and now m2m related to the publication model, this data will need to be migrated. Views will follow later.

#1073 
Closes #1482


bramvanginneken:
Is there also the option to add publications without DOI, eg arXiv?

jmsmkn:
> Is there also the option to add publications without DOI, eg arXiv?

Not yet, will add it though, keep following #1073

PR #1548, opened 2020-09-25
Implement draggable question for reordering

pkcakeout:
- Add [SortableJS](https://github.com/SortableJS/Sortable) to easily implement draggable list items. Development seems pretty active - better than "dragula", for example.
- Messily adapted the REST-API to allow patching of the order-field
- Rewriting the order of each question to force absolute ordering

I am pretty sure this is not well implemented, but maybe a nice feature! I thought: "let's try something that might be useful for grand-challenge, just to get up to speed with that project again..."

out

### DIFF
--- a/app/grandchallenge/reader_studies/models.py
+++ b/app/grandchallenge/reader_studies/models.py
@@ -1091,15 +1091,11 @@ class Question(UUIDModel):
     class Meta:
         ordering = ("order", "created")
 
-    @property
-    def answer_type_display(self):
-        return self.get_answer_type_display()
-
     def __str__(self):
         return (
             f"{self.question_text} "
             "("
-            f"{self.answer_type_display}, "
+            f"{self.get_answer_type_display()}, "
             f"{self.get_image_port_display() + ' port,' if self.image_port else ''}"
             f"{'' if self.required else 'not'} required, "
             f"order {self.order}"
@@ -1111,7 +1107,7 @@ class Question(UUIDModel):
         """Values that are included in this ``Question``'s csv export."""
         return [
             self.question_text,
-            self.answer_type_display,
+            self.get_answer_type_display(),
             self.required,
             f"{self.get_image_port_display() + ' port,' if self.image_port else ''}",
         ]
--- a/app/grandchallenge/reader_studies/models.py
+++ b/app/grandchallenge/reader_studies/models.py
@@ -1091,11 +1091,15 @@ class Question(UUIDModel):
     class Meta:
         ordering = ("order", "created")
 
+    @property
+    def answer_type_display(self):
+        return self.get_answer_type_display()
+
     def __str__(self):
         return (
             f"{self.question_text} "
             "("
-            f"{self.get_answer_type_display()}, "
+            f"{self.answer_type_display}, "
             f"{self.get_image_port_display() + ' port,' if self.image_port else ''}"
             f"{'' if self.required else 'not'} required, "
             f"order {self.order}"
@@ -1107,7 +1111,7 @@ class Question(UUIDModel):
         """Values that are included in this ``Question``'s csv export."""
         return [
             self.question_text,
-            self.get_answer_type_display(),
+            self.answer_type_display,
             self.required,
             f"{self.get_image_port_display() + ' port,' if self.image_port else ''}",
         ]
--- a/app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
+++ b/app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
@@ -295,11 +295,20 @@
                  aria-labelledby="v-pills-questions-tab">
                 <h2>Questions</h2>
 
-                <ul class="list-group list-group-flush">
+                <script src="https://cdn.jsdelivr.net/npm/sortablejs@latest/Sortable.min.js"></script>
+
+                <ul class="list-group list-group-flush" id="full-questions-list">
                     {% for question in object.questions.all %}
-                        <li class="list-group-item">
+                        <li class="list-group-item" gc-update-link="{% url 'api:reader-studies-question-detail' pk=question.pk %}">
                             <div class="w-100 d-flex justify-content-between">
-                                <div>{{ question }}</div>
+                                <div>
+                                    <i class="fa fa-grip-vertical grip-handle" style="cursor: grab; color: gray;"></i>
+                                    {{ question.question_text }}
+                                </div>
+                                <div>
+                                    {{ question.answer_type_display }}
+                                </div>
+
                                 <a class="btn btn-primary btn-sm"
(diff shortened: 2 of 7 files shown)


app/grandchallenge/reader_studies/models.py
```diff
@@ -1091,11 +1091,15 @@ class Question(UUIDModel):
     class Meta:
         ordering = ("order", "created")
 
+    @property
+    def answer_type_display(self):
+        return self.get_answer_type_display()
+
```
pkcakeout, line 1097: Added to allow use in template

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
```diff
@@ -295,11 +295,20 @@ <h2>Upload Sessions</h2>
                  aria-labelledby="v-pills-questions-tab">
                 <h2>Questions</h2>
 
-                <ul class="list-group list-group-flush">
+                <script src="https://cdn.jsdelivr.net/npm/sortablejs@latest/Sortable.min.js"></script>
+
+                <ul class="list-group list-group-flush" id="full-questions-list">
                     {% for question in object.questions.all %}
-                        <li class="list-group-item">
+                        <li class="list-group-item" gc-update-link="{% url 'api:reader-studies-question-detail' pk=question.pk %}">
```
pkcakeout, line 302: Should I use the REST API like this? Or should I rather submit to a local API endpoint in the view?

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
```diff
@@ -295,11 +295,20 @@ <h2>Upload Sessions</h2>
                  aria-labelledby="v-pills-questions-tab">
                 <h2>Questions</h2>
 
-                <ul class="list-group list-group-flush">
+                <script src="https://cdn.jsdelivr.net/npm/sortablejs@latest/Sortable.min.js"></script>
```
pkcakeout, line 298: Messily added this in-line. This is of course not acceptable? What is the way to do this? I saw something like this:

~~~~~.py
class Media:
    JS = ("something.js", )
    CSS = ("else.css",)
~~~~~

Is that the answer?

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
```diff
@@ -308,6 +317,36 @@ <h2>Questions</h2>
                         </li>
                     {% endfor %}
                 </ul>
+                <script>
```
pkcakeout, line 320: Do we have a place for local JavaScript? Same pattern as jqfileuploader maybe?

app/grandchallenge/reader_studies/views.py
```diff
@@ -896,7 +896,7 @@ def ground_truth(self, request, pk=None, case_pk=None):
         )
 
 
-class QuestionViewSet(ReadOnlyModelViewSet):
+class QuestionViewSet(ModelViewSet):
```
pkcakeout, line 899: This seems like a brutal security issue to me. Is there a better way to allow _patching_ of the Question-model?

I only want to rewrite the "order" field of the Question-model.

app/grandchallenge/reader_studies/models.py
jmsmkn: I'd rather keep to Django's API, you can call `{{ question.get_answer_type_display }}` from a django template, see their documentation about how template variables are resolved: https://docs.djangoproject.com/en/3.0/ref/templates/language/#variables

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
jmsmkn: Yes, you're hand-rolling a form here, it's better to let django render it. Create the form in the view and add it to the context.

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
jmsmkn: No, use django forms.

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
jmsmkn: Also, it's best to serve it ourselves, these free CDNs are not reliable.

jmsmkn: You can add it to packages.json.

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
```diff
@@ -308,6 +317,36 @@ <h2>Questions</h2>
...
+                    (function () {
+                        const el = document.getElementById("full-questions-list");
+
+                        function processUpdate() {
+                            let i = 1;
+                            for (const c of el.children) {
+                                const url = c.attributes.getNamedItem("gc-update-link").value;
+                                jQuery.ajax({
```
jmsmkn, line 328: Don't use jquery ajax, use `fetch`.

app/grandchallenge/reader_studies/views.py
jmsmkn: Yes, this change bypasses all security controls that means anyone can do anything to a question.

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
jmsmkn: Add it as a file in the `js` subdirectory of the reader study templates, then add it to the `script` template block.

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
jmsmkn: In general, don't use custom attributes, if you need to pass extra things use data attributes: https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

jmsmkn (changes requested):
Please see my comments, I think that another approach should be taken that does not use the REST API endpoints. IMO it would be better to use a formset with the existing question forms. See https://docs.djangoproject.com/en/3.1/topics/forms/formsets/

jmsmkn:
Looks good though! Would be nice to have this on pages too.

pkcakeout:
Roger, I think I get it! Strategy: New sake a widget for FormSets that allows one to sort the entries instead, correct?

I will rework this to match those specs. I will convert the PR to a draft until then, because it is so far off with regards to being "mergable". Thank you for the detailed comments. It is great to learn a bit more about grand-challenge/django again!

jmsmkn:
I'm going to close this to get it out of the list of open PRs, please feel free to re-open when you have a solution.

PR #1553, opened 2020-09-28
Added result_template to Algorithm model

### LINKED ISSUE #1070: More informative "results" column in algorithm job list
The job list currently shows the entire "results.json" file that a algorithm job produced in the "Results" column. This can get messy, for example, the calcium scoring algorithm computes all kinds of scores:

image

Would be nicer if this was hidden by default. Instead, "results.json" could support something like a "results_message" key next to the "metrics" key, so that the algorithm can output a (short) string that is displayed in the job list. For example, in case of the calcium scoring algorithm, I would simply output the overall score, and the vertebra segmentation algorithm could output the number of detected vertebrae, etc.

miriam-groeneveld:
Added rendered result text to Algorithm Job

Closes #1070


app/grandchallenge/algorithms/templates/algorithms/data_tables/job_list.html
```diff
@@ -33,6 +33,7 @@
         {% firstof job.output|user_error job.get_status_display %}
    {% endif %}"
    data-pk="{{ job.pk }}"
+   data-rendered-output="{{ job.rendered_result_text }}"
```
miriam-groeneveld, line 36: This doesn't work when the rendered_result_text contains quotes, and I'm not sure what would be the right way to handle it. I think I can't set it as a property this way?

MikeOverkamp-diag: I think that `job.rendered_result_text|escape`  might suffice here to fix that issue.

app/grandchallenge/algorithms/models.py
```diff
@@ -419,6 +431,21 @@ def create_result(self, *, result: dict):
...
+        try:
+            output = self.outputs.get(interface=interface)
+            if not self.algorithm_image.algorithm.result_template:
+                return output.value
+            template_output = JINJA_ENGINE.from_string(
+                self.algorithm_image.algorithm.result_template
+            ).render(dict=output.value)
+
+            return md2html(template_output)
+        except (ObjectDoesNotExist, SecurityError):
```
jmsmkn, line 446: Could you split this into two try statements? That way, we know what exceptions we expect from what methods. So, one to get the output and catch `ObjectDoesNotExist`, and the second to catch `SecurityError` from the template rendering.

app/grandchallenge/algorithms/models.py
```diff
@@ -39,6 +42,8 @@
 DEFAULT_INPUT_INTERFACE_SLUG = "generic-medical-image"
 DEFAULT_OUTPUT_INTERFACE_SLUG = "generic-overlay"
 
+JINJA_ENGINE = sandbox.SandboxedEnvironment()
```
jmsmkn, line 45: Shouldn't we use `ImmutableSandboxedEnvironment` to prevent modification of the objects? I think so as we send `output.value` directly and we do not want to allow accidental modification of that.

app/grandchallenge/algorithms/models.py
```diff
@@ -419,6 +431,21 @@ def create_result(self, *, result: dict):
...
+    @property
+    def rendered_result_text(self):
+        interface = ComponentInterface.objects.get(slug="results-json-file")
+        try:
+            output = self.outputs.get(interface=interface)
+            if not self.algorithm_image.algorithm.result_template:
+                return output.value
+            template_output = JINJA_ENGINE.from_string(
+                self.algorithm_image.algorithm.result_template
+            ).render(dict=output.value)
```
jmsmkn, line 443: This creates a context environment called dict, which is a reserved built in. Could we rename it? `result` or `result_dict` maybe?

app/grandchallenge/algorithms/serializers.py
```diff
@@ -28,6 +28,7 @@ class Meta:
             "latest_ready_image",
             "pk",
             "title",
+            "result_template",
```
jmsmkn, line 31: Do we need to serialize this? I thought that we rely on Django to interpolate it?

app/grandchallenge/algorithms/templates/algorithms/data_tables/job_list.html
```diff
@@ -26,7 +26,7 @@
    data-title="Result Info"
    data-output="
    {% if job.status == job.SUCCESS %}
-        {% for output in job.outputs.all %}{% if output.value %}{{ output.value }}{% endif %}{% empty %}No output produced.{% endfor %}
+        {% if job.rendered_result_text %}{{ job.rendered_result_text|force_escape }}{% else %}No output produced.{% endif %}
```
jmsmkn, line 29: For my understanding: why `force_escape` over `escape`?

app/grandchallenge/algorithms/templates/algorithms/job_list.html
```diff
@@ -97,8 +115,19 @@ <h5 class="modal-title"></h5>
             let button = $(event.relatedTarget);
             let modal = $(this);
             modal.find('.modal-title').text(button.data('title'));
-            modal.find('.modal-job-output').text(button.data('output'));
+            modal.find('.modal-job-output').html(unescape(button.data('output')));
```
jmsmkn, line 118: Is this safe? I think escaping here might introduce an XSS vulnerability. My thinking: user template -> unsafe code is escaped by `md2html`, which calls `clean` and marked safe as dangerous stuff is escaped -> on L29 _everything_ is escaped including the naughty stuff -> this line unescapes everything including the bad stuff.

Is there another way to pass this information to the modal rather than escaping and unescaping? A javascript string literal maybe? Or by using djangos `json_script` filter https://docs.djangoproject.com/en/3.0/ref/templates/builtins/#json-script (create a unique id for each row and then pass that to the model data insead).

app/grandchallenge/core/management/commands/init_gc_demo.py
```diff
@@ -302,7 +306,11 @@ def _create_algorithm_demo(self):
             self.users["algorithm"], self.users["demo"]
         )
         algorithm.users_group.user_set.add(self.users["algorithmuser"])
-
+        algorithm.result_template = (
+            "{% for key, value in dict.metrics.items() -%}"
```
jmsmkn, line 310: If we do change the context variable name this needs to be updated here.

jmsmkn (changes requested):
Cool feature! I have a few comments that we should address as providing user templates is risky and I want to be sure that we get things right.

app/grandchallenge/algorithms/models.py
```diff
@@ -419,6 +431,21 @@ def create_result(self, *, result: dict):
             )
             self.outputs.add(output_civ)
 
+    @property
+    def rendered_result_text(self):
+        interface = ComponentInterface.objects.get(slug="results-json-file")
+        try:
+            output = self.outputs.get(interface=interface)
+            if not self.algorithm_image.algorithm.result_template:
```
jmsmkn, line 439: We might need to add `"algorithm_image__algorithm"` to the `select_related` in the `AlgorithmJobsList` view to avoid an n+1 queries. If you turn on the debug toolbar are the number of SQL queries stable as the number of successful results with an algorithm template increase?

app/grandchallenge/algorithms/templates/algorithms/job_list.html
miriam-groeneveld: The rendered_result_text is text that has been escaped by md2html. It is therefor marked as safe, which is why I need to force_escape (I think), as escape doesn't seem to have any effect. The force_escape/unescape is only needed to escape quotes that are added by the markdown, and mess up the html.

jmsmkn: Ok - but it also escapes stuff that shouldn't be escapes, introducing the XSS vunrulbility. To see this, add `%3Cscript%3Ealert%28%27uh-oh%27%29%3C/script%3E` to the algorithm template, you should see a pop up when the modal is activated.

miriam-groeneveld: ah, I see. I'll find another way to do this then.

app/grandchallenge/algorithms/models.py
miriam-groeneveld: Yes, the number of queries remains at 11, regardless of number of results, or whether I have a template defined.

app/grandchallenge/algorithms/models.py
miriam-groeneveld: Split the errors and added UndefinedError as well.

app/grandchallenge/algorithms/templates/algorithms/data_tables/job_list.html
miriam-groeneveld: Changed to json_script

app/grandchallenge/algorithms/templates/algorithms/job_list.html
miriam-groeneveld: Changed to json_script

app/grandchallenge/algorithms/models.py
```diff
@@ -419,6 +436,23 @@ def create_result(self, *, result: dict):
...
+                self.algorithm_image.algorithm.result_template
+            ).render(result_dict=output.value)
+
+            return md2html(template_output)
+        except UndefinedError as e:
+            return f"Jinja template is incorrect: {e}"
+        except SecurityError:
+            return "Jinja template is not allowed"
+        except ObjectDoesNotExist:
+            return ""
```
jmsmkn, line 454:
```suggestion
        try:
            output = self.outputs.get(interface=interface)
        except ObjectDoesNotExist:
            return ""

        try:
            template_output = JINJA_ENGINE.from_string(
                self.algorithm_image.algorithm.result_template
            ).render(result_dict=output.value)
        except (UndefinedError, SecurityError):
            return "Jinja template is invalid"

        return md2html(template_output)
```

Sorry - this is what I meant, this way we can see where we expect the exceptions to be raised.

app/grandchallenge/algorithms/models.py
```diff
@@ -84,6 +89,18 @@ class Algorithm(UUIDModel, TitleSlugDescriptionModel):
             "terms of usage, define them here."
         ),
     )
+    result_template = models.TextField(
+        blank=True,
```
jmsmkn, line 93: Should we add a default to help people get going? `default="<pre>{{ result_dict }}</pre>"`?

app/grandchallenge/algorithms/templates/algorithms/job_list.html
```diff
@@ -94,11 +112,27 @@ <h5 class="modal-title"></h5>
 
     <script type="text/javascript">
         $('#resultInfoModal').on('show.bs.modal', function (event) {
+            let button = $(event.relatedTarget);
+            let modal = $(this);
+            const json_element = document.getElementById(button.data('pk'));
+            if (json_element){
+                modal.find('.modal-job-output').html(JSON.parse(json_element.textContent));
+            }else {
```
jmsmkn, line 120: I think this means that we can remove the if statement on L28 of jobs_list.html?

```
   {% if job.status == job.SUCCESS %}
        {% for output in job.outputs.all %}{% if output.value %}{{ output.value }}{% endif %}{% empty %}No output produced.{% endfor %}
   {% elif change_job %}
```

would become:

```
   {% if change_job %}
```

app/grandchallenge/algorithms/models.py
jmsmkn: Also, we should avoid giving raw exceptions back to users, this has caused issues in the past, so, I've changed both messages to `"Jinja template is invalid"`

app/tests/algorithms_tests/test_models.py
```diff
@@ -89,3 +89,21 @@ def test_outputs_are_set():
...
+    assert job.rendered_result_text == "<p>foo score: 13.37</p>"
+
+    job.algorithm_image.algorithm.result_template = "{% for key, value in dict.metrics.items() -%}{{ key }}  {{ value }}{% endfor %}"
+    assert (
+        job.rendered_result_text
+        == "Jinja template is incorrect: 'type object' has no attribute 'metrics'"
+    )
+
+    job.algorithm_image.algorithm.result_template = "{{ str.__add__('test')}}"
+    assert job.rendered_result_text == "Jinja template is not allowed"
```
jmsmkn, line 109:
```suggestion
    assert job.rendered_result_text == "Jinja template is invalid"
```
If you accept the changes above.

app/tests/algorithms_tests/test_models.py
```diff
@@ -89,3 +89,21 @@ def test_outputs_are_set():
...
+
+    assert job.rendered_result_text == ""
+    job.create_result(result={"foo": 13.37})
+    assert job.rendered_result_text == "<p>foo score: 13.37</p>"
+
+    job.algorithm_image.algorithm.result_template = "{% for key, value in dict.metrics.items() -%}{{ key }}  {{ value }}{% endfor %}"
+    assert (
+        job.rendered_result_text
+        == "Jinja template is incorrect: 'type object' has no attribute 'metrics'"
+    )
```
jmsmkn, line 106:
```suggestion
    assert job.rendered_result_text == "Jinja template is invalid"
```
If you accept the changes abvoe

jmsmkn (commented):
Looks great! Last round of small comments.

app/grandchallenge/algorithms/templates/algorithms/job_list.html
miriam-groeneveld: Well, not completely, if there's no template, the json_element won't be added. The raw result.json is the fallback option.

jmsmkn: OK!

app/grandchallenge/algorithms/models.py
miriam-groeneveld: Good idea

PR #1560, opened 2020-10-06
Improve results list loading time

### LINKED ISSUE #1559: Algorithm results list AJAX view is slow
Following #1553 generating a page of results is slow, 5s for 100 results. Likely an n+1 queries resulting from `interface = ComponentInterface.objects.get(slug="results-json-file")`. Might need a similar solution as in #1477. My guess is that we need to run with the HistoryPanel of Django Debug Toolbar.

jmsmkn:
Adds loading spinner
Reduces number of rows
Adds related queries to API view

Co-authored-by: miriam-groeneveld <51329283+miriam-groeneveld@users.noreply.github.com>

Closes #1559

### DIFF
--- a/app/grandchallenge/algorithms/templates/algorithms/job_list.html
+++ b/app/grandchallenge/algorithms/templates/algorithms/job_list.html
@@ -49,6 +49,13 @@
             </tr>
             </thead>
             <tbody>
+            <tr>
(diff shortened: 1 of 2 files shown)


miriam-groeneveld (approved):
LGTM

PR #1566, opened 2020-10-14
Add viewset for UserProfile

HarmvZ:


app/grandchallenge/profiles/serializers.py
```diff
@@ -0,0 +1,21 @@
...
+
+    class Meta:
+        model = UserProfile
+        fields = (
+            "user",
+            "mugshot",
+            "privacy",
+            "institution",
+            "department",
+            "country",
```
jmsmkn, line 19: Unfortunately, we need to expose this as `location`, not `country`.

app/grandchallenge/profiles/views.py
```diff
@@ -209,3 +218,18 @@ class PreSocialView(FormView):
...
+
+
+class UserProfileViewSet(viewsets.ReadOnlyModelViewSet):
+    serializer_class = UserProfileSerializer
+    permission_classes = (DjangoObjectOnlyPermissions,)
+    filter_backends = (UserProfileObjectPermissionsFilter,)
+    queryset = UserProfile.objects.all()
+
+    @action(detail=False, methods=["get"])
+    def self(self, request):
```
jmsmkn, line 230: Do we really need the `self` view? The `view_profile` permission is only assigned to the user that owns the profile, so I think that this does the same thing as the normal list view.

jmsmkn (changes requested):

app/grandchallenge/profiles/views.py
HarmvZ: Currently the permission is indeed only assigned to the user that owns it, but I thought this might change in the future? Since right now it is possible to see profiles of other users because the permission check is not implemented in those views. Also for superusers it would not work as expected.

jmsmkn (approved):

PR #1570, opened 2020-10-15
Prevent upload retries for empty files

MikeOverkamp-diag:
When uploading an empty file, drf raises a validation error `The submitted file is empty.` and returns a status code 400. This would trigger jqfileupload to retry the upload numerous times. This PR adds a check for empty files in the retry handling.


jmsmkn (approved):

MikeOverkamp-diag:
I don't think the test failure is caused by these changes and the test does not fail locally for me. I think it might just be a matter of list ordering, but I'm not 100% sure about that.

jmsmkn:
That's right, nothing to worry about, that test should be updated.

PR #1572, opened 2020-10-16
Adds json pretty printing to default algorithm template

jmsmkn:


pkcakeout (approved):
Cool! :+1:

bramvanginneken:
Very nice! I think it would be great if there would be an Info plugin to also show this contents in a nice manner. What would be great if there is a way to include locations in the scans in this results json, for example the 5 possible lung cancer locations of grt123, and this would be formatted in such a way in the Result plugin so you could click on the text and you would jump to that location in the scan. I do this now manually with scrolling.

jmsmkn:
> What would be great if there is a way to include locations in the scans in this results json, for example the 5 possible lung cancer locations of grt123, and this would be formatted in such a way in the Result plugin so you could click on the text and you would jump to that location in the scan. I do this now manually with scrolling.

Paul is meeting with Cheryl on Monday to discuss how to implement that.

PR #1573, opened 2020-10-20
Allows reader studies to be viewed without login

### LINKED ISSUE #1546: Make reader studies visible without login
Currently when signed out, when you click on a public reader study you have to login right away  before seeing any info about it. For algorithms this is not the case, you can read basic information while logged-out. Feature request from Bram and me as we would like to use the initial reader study page as an information page about our upcoming study to gather data from other hospitals. 

Would like the same flow as for the algorithms, where without an account, basic info for the reader studies is visible.

jmsmkn:
This PR:

- adds a `read_readerstudy` permission
- gives the reader studies editors and readers group permission to read the study
- gives the registered and anon users group permission to view the reader study if the reader study is public

Note: post deployment save needs to be run on each reader study object.

Closes #1546


MikeOverkamp-diag:
LGTM, but the `test_answer_creator_is_reader` needs to be fixed I think. Do we also need a data migration / management command to give existing readers/editors the new permission?

jmsmkn:
Ah! I thought I updated that test. This is because I also made a change so that permission checks are done using permissions objects, rather than on group membership. Should have been that way from the start. This fixed a weirdness that reader study editors also needed to be added to the readers group in order to read the study, not needed anymore now that we have explicit permissions.

MikeOverkamp-diag:
Only question I have left is whether we need some dat migration or a management command to give existing editors / readers the new permission?

jmsmkn:
Yes, post deployment save needs to be run on each reader study object, I will do that manually.

MikeOverkamp-diag (approved):
Ok great!

PR #1575, opened 2020-10-21
1564 calculate resolutions

### LINKED ISSUE #1564: Images converted to tiff can get wrong image spacing.
Images that are converted to tif files can get incorrect image spacing during conversion. Problem was found with certain svs files, where openslide found certain mpp propreties (micron per pixel) but these were not used to set the tiff tags ResolutionX and ResolutionY, resulting in these properties having a value of 1.

When used in algorithms that expect a certain image spacing, these images are then rejected.

miriam-groeneveld:
Temporary solution for #1564. This problem is resolved in VIPS 8.10, which is not available yet in the Debian buster that is currently used in GC. 

Closes #1564

### DIFF
--- a/app/grandchallenge/cases/image_builders/tiff.py
+++ b/app/grandchallenge/cases/image_builders/tiff.py
@@ -328,6 +328,7 @@ def _convert_to_tiff(*, path: Path, pk: UUID, converter) -> Path:
     )
 
     # correct xres and yres if they have default value of 1
+    # can be removed once updated to VIPS 8.10
     if image.get("xres") == 1 and "openslide.mpp-x" in image.get_fields():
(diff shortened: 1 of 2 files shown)


jmsmkn (approved):
LGTM - can we close 1564 with this?

miriam-groeneveld:
Well, I thought we'd keep it open, so we are reminded to remove this fix again when updated to 8.10

jmsmkn:
Is there a way that we could write a test that would fail once we have updated libvips? I'm worried that we will forget about the issue and the workaround.

jmsmkn (approved):
Super! I didn't know about assert messages either. I will close the ticket and this will give us a good reminder to remove the workaround when upgrading.

PR #1577, opened 2020-10-28
Feature/observable notebook cell render

pushpanjalip:
This PR adds dynamic rendering of selected Observable notebook cells.
Its not compatible with the latest Observable notebook from Maarten & Berend as Metrics is changed.
I have temporarily replaced Observable notebook url in evaluation_results.py


jmsmkn (approved):
LGTM! We will also need to handle the cases when all cells are selected (`cells === ["*"]`)

PR #1582, opened 2020-11-03
remove g-c footer from aiforradiology

Kicksta:


jmsmkn (approved):

bramvanginneken:
Should these particular requests for AI for Radiology be part of the general grand-challenge.org repository? Should we not plit it off? I would say that in general functionality to hide or replace header and footer of pages on a grand-challenge.org server would be nice to have. Als the option that when many topics are shown (136 now for the products page) to have a way to make the page not so long by default is nice. Like the result pages of grand-challenge, but also to have only a more button, and then it loads more items. This way you can see the fotter without having to scroll a lot.

jmsmkn:
I want to keep products in grand challenge so there is that link from archive to reader study to challenge to algorithm to product, and makes it easier to represent and link objects across each. Splitting the products code out would make that a lot of extra work both for development and ops.

jmsmkn:
And for the long list pagination needs to be added in the backend, that's not difficult. A more button (or infinite scroll) is a lot of work as it's not a feature built in to django.

PR #1585, opened 2020-11-04
Adds algorithm job credits

### LINKED ISSUE #1085: Credit system for algorithms
I think either Bram or James mentioned something similar a while back, but I couldn't find a feature request ticket for this: I'm getting now requests every day from people who want to try out my algorithms. It would be fantastic if I could grant them access with a limited "job budget" so that they can upload one or two of their scans but not more.

miriam-groeneveld:
Added job_limit to algorithm model. This limit will be checked in the upload view and when starting tasks. Failed jobs will not be counted.

Closes #1085


jmsmkn (changes requested):
Looks great! A couple of small things that would be good to tidy up before deployment.

jmsmkn (approved):
Nice work, thanks!

Issue #1053, opened 2019-12-17
Sharing of jobs with algorithm users


nlessmann:
A low priority feature request:

If I want to demo my algorithm to somebody, I can add them as algorithm users, but they currently start with an empty job list and would need to upload scans themselves to test the algorithm.

It would be a nice additional feature if I could also upload a few scans myself (as algorithm editor) and share the results with algorithm users, i.e., select for each job additional users that also see the job in their account and can inspect the results in a workstation.

jmsmkn:
We have a few requests for this. Requirements:

- Sharing the results with other users. We then need to ensure that those users have permission to read the result, result images, input images, algorithm (but not algorithm execution) and workstation for that algorithm.
- Making a result available to any user (except Anonymous User). Same as above.

jmsmkn:
You can now mark results as public (working on the UI)

PR #1593, opened 2020-11-13
Manage algorithm viewing permissions with a group of groups

jmsmkn:
This PR adds a viewers group to each algorithm job which is used to assign view permissions for each algorithm job and associated input and output image.

- The job creator is added to the viewers group by default
- The algorithm editors are added as viewers
- If the job is public, then the registered and anon group is added to the viewers

This will allow for a lot more flexibility in the future. 

Still TODO:

- Migration of existing permissions:
  - Group creation, adding the creators to this group and removal of direct permission assignments
  - Adding editors and viewer groups to the m2m
- Setting the viewers group to non-nullable
- Allow algorithm editors to add viewers to the viewers group

Future work will allow assigning archive groups, and using the same models for algorithm evaluation.

#1053


app/grandchallenge/challenges/models.py
```diff
@@ -452,7 +452,7 @@ def save(self, *args, **kwargs):
             self.create_default_phases()
             send_challenge_created_email(self)
 
-        if self.hidden != self._hidden_orig:
+        if adding or self.hidden != self._hidden_orig:
```
HarmvZ, line 455: For my understanding, can you explain why this is added? Nothing else is changed with the Challenge model, right?

HarmvZ (approved):
Very nice job!

app/grandchallenge/challenges/models.py
jmsmkn: This would fix a bug if you create a challenge with arguments of `hidden=False`, we don't actually do that anywhere but maybe one day we would. I hit that problem in the tests with public Jobs so figured to fix it here too.

PR #1596, opened 2020-11-17
Update image perms when used across apps

jmsmkn:
This PR fixes an issue where permissions would be dropped for archive or reader study groups when that image is used in an algorithm.

#1053


app/grandchallenge/reader_studies/signals.py
```diff
@@ -40,6 +42,7 @@ def update_image_permissions(instance, action, reverse, model, pk_set, **_):
     op = assign_perm if "add" in action else remove_perm
 
     for rs in reader_studies:
+        op("view_image", rs.editors_group, images)
```
HarmvZ, line 45: Did the editors of a reader study not yet have permission to view the images before this PR?

HarmvZ (approved):
LGTM!

app/grandchallenge/reader_studies/signals.py
jmsmkn: No - this was a new thing we added in #1573 but I missed the image permissions. No one complained though so maybe they were still using the same workflow?

PR #1597, opened 2020-11-17
Allow adding users to job with change_job permission

### LINKED ISSUE #1053: Sharing of jobs with algorithm users

jmsmkn:
This PR allows people with permissions to change the job (currently algorithm editors) to add other users to the jobs viewers group so that they can view this job. The public eye is set to a warning if the job is shared with more than 1 person. If you hover over the eye you can see which users have access to view this job.

The users who have view permissions will have to manually navigate to the jobs list page, as they might not have permission to view or execute the algorithm. The link for this is displayed in the success message.

Demo: https://youtu.be/J1C45nTF0uc

Closes #1053

### DIFF
--- a/app/grandchallenge/algorithms/forms.py
+++ b/app/grandchallenge/algorithms/forms.py
@@ -123,13 +123,11 @@ class UserGroupForm(SaveFormInitMixin, Form):
             raise ValidationError("You cannot add this user!")
         return user
 
-    def add_or_remove_user(self, *, algorithm):
+    def add_or_remove_user(self, *, obj):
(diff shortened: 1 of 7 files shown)


HarmvZ (approved):
Very nice, especially the orange (warning) eye to visualize a shared job

PR #1602, opened 2020-11-20
Adds filtering back to the challenge list page

### LINKED ISSUE #1591: Add filtering to the Challenge List view
Filtering on Modality, Host, etc was removed when we had to implement serverside pagination, add this back.

jmsmkn:
This uses django filters on the challenge list page. This is kindof a complex example as we have 2 querysets of different models with a common abstract base class, but I think that we should replace any other searching or filtering of `ListViews` with a `FilterSet` so that we can use it in DRF. Should only be a few loc for other views. An `update_search_params` tag is added so that filtering and pagination work well together. 

We used to also filter by Host and Year, but those are not easily accessible as they are calculated in python right now.

Demo: https://youtu.be/BZMLGCgq6v4

Closes #1591

### DIFF
--- a/app/grandchallenge/challenges/filters.py
+++ b/app/grandchallenge/challenges/filters.py
@@ -21,7 +21,7 @@ from grandchallenge.challenges.models import (
 
 
 class ChallengeFilter(FilterSet):
-    search = CharFilter(method="search_filter", label="Search")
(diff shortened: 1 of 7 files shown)


MikeOverkamp-diag (approved):
Looks great!

PR #1606, opened 2020-11-23
Adds a Job Detail Page

### LINKED ISSUE #1599: Make color of algorithm job status configurable
The color in the algorithm results table indicates only whether the docker image was successfully run or not. But many algorithms can actually run without any error and still produce a negative result. For example, the vertebra segmentation algorithm can run without problems but also without finding a single vertebra in the entire image. I've used the jinja template to show some text to the user in that case because this usually means that something is wrong with the input data. It would be very nice if it was possible to turn the status from green to orange based on the data in the results.json file (maybe by allowing the editors to configure a jinja-style if/else statement for this?). That way, we could highlight jobs with problems and distinguish them from really successful and really failed jobs.

jmsmkn:
This PR adds a job detail page, which allows the job editor to manage the users group, and the algorithm editors to see the logs. As this is a new page, I have removed some of the clutter from the job list view, such as downloading images and editing the job. Job editors are able to remove users from the viewers group as long as they are not the request user. If the job is successful but there is output on stderr the job badge will have a warning.

Demo: https://youtu.be/v7rP35j2xks

Closes #1599

### DIFF
--- a/app/grandchallenge/algorithms/templates/algorithms/job_detail.html
+++ b/app/grandchallenge/algorithms/templates/algorithms/job_detail.html
@@ -138,9 +138,10 @@
                             {% for file in input.image.files.all %}
                                 <li>
                                     <a class="badge badge-primary"
-                                       title="Download Input {{ input.image.name|stem }} ({{ file.file|suffix }})"
                                        href="{{ file.file.url }}">
-                                        <i class="fa fa-download"></i> Input ({{ file.file|suffix }})
+                                        <i class="fa fa-download"></i>
+                                        Input {{ forloop.parentloop.counter }} {{ input.image.name|stem }}
+                                        ({{ file.file|suffix }})
                                     </a>
                                 </li>
                             {% endfor %}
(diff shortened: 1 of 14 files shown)


app/grandchallenge/algorithms/templates/algorithms/job_detail.html
```diff
@@ -0,0 +1,258 @@
...
+                    <dd class="col-sm-9">{{ object.comment }}</dd>
+                {% endif %}
+
+                <dt class="col-sm-3">Visibility</dt>
+                <dd class="col-sm-9">
+                    {% if object.public %}
+                        <i class="fa fa-eye text-success"></i> Result and images are public
+                    {% else %}
+                        {% if object.viewers.user_set.all|length > 1 %}
+                            {# TODO: Hack, we need to exclude the creator rather than checking the length is > 1 #}
```
miriam-groeneveld, line 128: Can't the algorithm's editors view all results as well?

jmsmkn: This status is about members of the viewers group, on the Users tab there is a list of all the other groups that have access via the `viewer_groups` many to many, which could be empty, the editors, and/or archive groups (PR incoming from Harm).

miriam-groeneveld (approved):
LGTM

app/grandchallenge/algorithms/templates/algorithms/job_detail.html
miriam-groeneveld: ok

PR #1608, opened 2020-11-24
Moves BodyRegion to Anatomy app

jmsmkn:
The BodyRegion, BodyStructure, and Modality models are in
the wrong app, so this is the first of the moves. Followed
the guide from
https://realpython.com/move-django-model/#the-django-way-rename-the-table


jmsmkn:
The PR errors were spurious.

PR #1624, opened 2020-11-26
Allow attaching of algorithms to archives to run for every image of an archive

HarmvZ:
This PR adds a many-to-many field in the `Archive` model for `Algorithm`s. This will allow an archive editor to add algorithms to the archive:
image
Which will run the algorithm on all images in the archive. The result will be visible to all archive editors, users and uploaders. When a new image is uploaded to the archive the algorithm will automatically be run with that image as input. The results will be visible in the archive  -> cases list view. 
FAxXtyDQXm


jmsmkn (approved):
Looks great! I will make a few small changes before merging.

app/grandchallenge/archives/signals.py
```diff
@@ -2,43 +2,71 @@
...
 
     for archive in archives:
         op("view_image", archive.editors_group, images)
         op("view_image", archive.uploaders_group, images)
         op("view_image", archive.users_group, images)
+
+    if "add" in action:
+        create_algorithm_jobs_for_archive_images.apply_async(
+            args=(list(archive_pks), list(image_pks))
+        )
```
jmsmkn, line 52: We should not use args in celery tasks

PR #1635, opened 2020-12-04
Download the files using the original names

### LINKED ISSUE #1634: add original input name to the output names
Currently the input files appear with random strings as names in the algorithm-input directory and the output files therefore don't contain the original input name (currently shown in details). Thus a user has to rename the images one by one manually and if they when on a batch of images it would be impossible or difficult to know which output(s) belong to which input. A simple solution would be to keep the original input name (shortening the random name-part would be nice too). Or to automatically append the input name to all output files.

jmsmkn:
Closes #1634


jmsmkn:
This does not work, I will add an explanation in #1635

PR #1639, opened 2020-12-08
Replaces AlgorithmEvaluation with Job

### LINKED ISSUE #1632: Replace AlgorithmEvaluation with Job
In #1472 I added an `AlgorithmEvaluation` model to the `evaluation` app, which executes the selected algorithm on 1 image from the challenge test set archive. This was done as there is a hard requirement that the algorithm editors should not be able to view the job inputs or logs, which are granted by `view_job` and `change_job` respectively. `AlgorithmEvaluation` only really differs from `algorithms.models.Job` due to the permissions: on an `AlgorithmEvaluation` only the challenge's admins group has the `view_algorithmevaluation` permission. At the time, the project pressure was too high to refactor the `Job` permissions business logic, so the model was more or less duplicated.

In #1593 we added an m2m field to maintain the `view_job` permission for `algorithms.models.Job`, and in #1624 Harm added the ability to set these groups at creation time. 

Everything is now in place to replace `AlgorithmEvaluation` with `Job`. What needs to change:

- The algorithm editors group should not be added to `job.viewer_groups` by default and should be added as an extra group in the algorithm tasks instead
- `AlgorithmEvaluation` init should be replaced with a `Job` init, the creator should be `None` so that the jobs viewers group is empty, and the challenge editors group added to the `viewer_groups`

jmsmkn:
This PR consolidates functionality. The only functional change is that algorithm creators now have permission to change the job, rather than the algorithm editors as the creator owns the data.

Closes #1632


HarmvZ (approved):
LGTM!

PR #1640, opened 2020-12-09
Fix repeated context fetching

jmsmkn:
Fixes an n+1 problem on the leaderboard page, introduced in #1636.


MikeOverkamp-diag (approved):
Oh wow, nice catch!

jmsmkn:
Ah it was my fault, too eager with code tidying. I remember implementing it like this the first time around for this very reason only to forget about it when cleaning up the patch for the PR.

PR #1647, opened 2020-12-10
Add tooltip ce fda

Kicksta:


jmsmkn (commented):
Where does the icon come from? Couldn't you use https://fontawesome.com/icons/info-circle?style=solid which is already included?

Kicksta:
@jmsmkn didn't know about it so I made my own in the same style as the verification icons, but this one works better also with the scaling, so I will change them. Thanks.

app/grandchallenge/products/templates/products/partials/product_list.html
```diff
@@ -31,16 +31,30 @@
...
           <div>
             <div class="d-flex justify-content-between">
-              <div class="h6 text-muted mr-auto pb-1">CE: </div>
-              <div class="h6 pr-2">{% if product.ce_class %}{{product.ce_class}}{% endif %}</div>
+                <a href="#" data-toggle="tooltip" data-placement="top" class="h6 text-muted mr-auto pb-1"
+                   title= {% include "products/partials/tooltip_CE.html" %}>
+                   CE:
+                </a>
+              <div class="h6 pr-2">
+                  {% if product.ce_class %}
```
MikeOverkamp-diag, line 44: I think it's better to put the `if` statement around the `<div class="h6 pr-2">`, otherwise you'll add some extra whitespace for items with no CE class. Unless that's what you want of course.

app/grandchallenge/products/templates/products/partials/tooltip_CE.html
```diff
@@ -0,0 +1,2 @@
+{% load static %}
```
MikeOverkamp-diag, line 1: I don't think this is needed. Nor is it in the other two tooltip templates.

MikeOverkamp-diag (commented):
Looks good! Just a couple of minor things.

app/grandchallenge/products/templates/products/partials/product_list.html
Kicksta: fixed

app/grandchallenge/products/templates/products/partials/tooltip_CE.html
Kicksta: fixed

jmsmkn:
Could you remove the unused icon? Then I think that this is good to go.

jmsmkn (approved):

PR #1651, opened 2020-12-11
Adds filtering to Archives, Algorithms and Reader Studies

### LINKED ISSUE #1649: Add Modality and Structure labels and filters to Archive, Reader Studies and Algorithms
Now that `structures` and `modalities` are in the correct apps we can add m2m relationships to the from Archives, Reader Studies and Algorithms. Add this, and the filters along with filtering on title and description to the list pages. Required for #1191.

`task_type` and `challenge_series` I don't think are required here so they should not be added.

jmsmkn:
This PR adds links from Archives, Algorithms and Reader Studies to Publications, Structures and Modalities, and adds common filter methods for these models and Challenges.

Closes #1649


app/grandchallenge/reader_studies/forms.py
```diff
@@ -96,6 +99,8 @@ class Meta:
         help_texts = READER_STUDY_HELP_TEXTS
         widgets = {
             "description": TextInput,
+            "modalities": Select2MultipleWidget,
+            "structures": Select2MultipleWidget,
         }
```
HarmvZ, line 104: Did you leave out the `publications` field here on purpose?

HarmvZ (approved):
LGTM. One question

app/grandchallenge/reader_studies/forms.py
jmsmkn: No, nice find!

jmsmkn: We really should use the same form for create and update...

PR #1654, opened 2020-12-15
Adds use_archive permission

### LINKED ISSUE #1653: Public archives never filtered in list view
As the `view_archive` permission is used to allow downloading of the archive the queryset for the list view is not constructed correctly. Add a new `download_archive` permission, and assign the `view_archive` permission to the reg and anon group as we do in algorithms and reader studies. Would also be good to add similar permissions to challenges.

See #1573

jmsmkn:
This PR fixes an issue with public Archives never being filtered by having a consistent way of using the `view_object` permission.

Closes #1653


miriam-groeneveld (approved):
lgtm

PR #1655, opened 2020-12-16
Check if image exists before pulling

jmsmkn:
This PR fixes an issue with docker pull limits as even empty pulls count towards your 200 layer limit. Ultimately we should not use any docker hub images but this is a project for next year. 

Default GID changed in development as this is the default in WSL2, which I want to encourage new developers to use.


miriam-groeneveld (approved):
LGTM

Issue #1628, opened 2020-12-02
Add support for Image answer types in Reader Studies


jmsmkn:
Some annotations are too large to be serialised to JSON and set directly on the answer, and the users are converting the cso's to masks anyway, so it would be better if we can store these directly. In the future, it would open up storing of secondary images, etc, that SATORI was hacking around by sending them to an archive.

Discussed the approach with @MikeOverkamp-diag today:

- We will implement this only for the Polygon annotations right now, and add splitting this back up later to #1357
- A new answer type needs to be added. The answer should store the following JSON as the answer: `{"upload_session_pk": <string, a UUID>}`
- A nullable foreign key to `Image` should be added. We need to validate that this is only going to be set for the correct new answer type.
  - Add this to the object history
  - Ensure the view permissions for the image are set correctly, these should match the groups for the answer view permissions
- The client will more or less need to implement this: https://github.com/DIAGNijmegen/rse-gcapi/blob/master/gcapi/gcapi.py#L593-L611
  - Stretch goal: upload the images asynchronously
- Once the answer has been created, the client will be able to call `self.raw_image_upload_sessions.process_images` with `json={"answer": <the answer pk>}`
- James will squash the migrations and can handle the API updates via Celery linked tasks once the models are in place

michelkok:
Outputting masks directly sounds like a great solution for me.

I assume that if a reader wants to answer in two session the mask is stored in between, loaded into the Reader Study and can then be updated with the CSO tools. As such, you are implementing a mapping between CSOList and masks, is that right? 
If that is correct, is it easy to extend this functionality a bit to allow us to upload partial masks even before a reader has started? It would solve [this cirrus ticket](https://github.com/DIAGNijmegen/rse-cirrus-core/issues/887) immediately.

jmsmkn:
Glad you mentioned this - this means that we would need to add the Image to the history as those are immutable. What we'd implement here would allow the workflow mentioned in the ticket, but the conversion from image to CSO would still need to be added on the CIRRUS side. 

@MikeOverkamp-diag you should integrate @michelkok's tools for converting from CSO to mask into CIRRUS Core.

MikeOverkamp-diag:
Just to be clear: are we talking about this bit of code: https://github.com/DIAGNijmegen/msk-mevislab-modules/pull/1/files

michelkok:
I am slightly confused now as well. So, we need a mapping between CSO and masks.

1. CSO -> Mask
[...]

PR #1657, opened 2020-12-17
Add image answer type

MikeOverkamp-diag:
This adds gc support for #1628. The implementation for the workstation is in progress.


MikeOverkamp-diag:
I'm going to update this one with the added migration from #1658

app/grandchallenge/cases/models.py
```diff
@@ -503,6 +503,12 @@ def update_viewer_groups_permissions(self, *, exclude_jobs=None):
         for g in groups_with_extra_perms:
             remove_perm("view_image", g, self)
 
+    def update_answer_image_viewer_permissions(self):
+        for answer in self.answer_set.all():
+            assign_perm("view_image", answer.creator, self)
+            editors = answer.question.reader_study.editors_group
+            assign_perm("view_image", editors, self)
```
jmsmkn, line 510: I think that there is an issue with this - if we add an image from an answer as an input to an algorithm (we don't do that right now but I can see it happening), then the editors permission will be removed by L504.

app/grandchallenge/cases/serializers.py
```diff
@@ -99,22 +103,29 @@ class RawImageUploadSessionPatchSerializer(RawImageUploadSessionSerializer):
...
-            sum(f in attrs for f in ["algorithm", "archive", "reader_study"])
+            sum(
+                f in attrs
+                for f in ["algorithm", "archive", "reader_study", "answer"]
+            )
             != 1
         ):
             raise ValidationError(
-                "1 of algorithm, archive or reader study must be set"
+                "1 of algorithm, archive, answer or reader study must be set"
```
jmsmkn, line 128: We also need a validator to ensure that the user has permission to add the image to the answer, see `validate_reader_study` below.

app/grandchallenge/reader_studies/models.py
```diff
@@ -999,6 +999,11 @@ def delete_reader_study_groups_hook(*_, instance: ReaderStudy, using, **__):
                 "version",
             ],
         },
+        "PIMG": {
+            "type": "object",
+            "properties": {"upload_session_pk": {"type": "string"}},
```
jmsmkn, line 1004: Should we include format here? `{"type": "string", "format": "uuid"}`.

app/grandchallenge/reader_studies/tasks.py
```diff
@@ -11,6 +11,13 @@ def add_score(obj, answer):
     obj.save()
 
 
+@transaction.atomic
+def add_image(obj, image):
+    obj.answer_image = image
+    obj.save()
+    image.update_answer_image_viewer_permissions()
```
jmsmkn, line 18: Are the `"view_image"` permissions updated if an image is replaced?

app/grandchallenge/reader_studies/tasks.py
```diff
@@ -37,3 +44,11 @@ def add_images_to_reader_study(*, upload_session_pk, reader_study_pk):
     reader_study = ReaderStudy.objects.get(pk=reader_study_pk)
 
     reader_study.images.add(*images.all())
+
+
+@shared_task
+def add_image_to_answer(*, upload_session_pk, answer_pk):
+    image = Image.objects.filter(origin_id=upload_session_pk).first()
```
jmsmkn, line 51: I think it would be good to error out if more than 1 image is uploaded, that would be something for the workstation/client developers to fix. What do you think?

jmsmkn:
Looks good! A few comments.

app/grandchallenge/reader_studies/tasks.py
MikeOverkamp-diag: With this I'm just wondering if we should remove these permissions for the creator of the answer if the image is changed? I think it would be ok for them to be able to view the image in any case. That would make assigning the permissions a bit more straight forward as well, as otherwise we'd have to check for users that have view permissions and are not in any of the expected groups.

jmsmkn: I'm not sure that I understand, could you clarify this?

MikeOverkamp-diag: Sorry, I did not realize I hadn't pushed the changes I made yet, hopefully that will make things a bit more clear.

What I meant was that if the image for an answer is changed, I think we should still leave the view permission intact for the user that uploaded the old image. I do not see any issue with doing so. The last part of what I wrote refers to the checks performed in this function: https://github.com/comic/grand-challenge.org/pull/1657/files#diff-4776f6f9ffb86c949abe083c2c7613f3991cfdf9387fe11202575e29cbcbba9fR456 (which I have just pushed the changes for). These checks would become needlessly complicated if we do want to remove the permission I think. Because we would then also have to check for users that have viewing permissions but are not in any of the groups that should have them.

Does that make things a bit more clear?

jmsmkn:
@MikeOverkamp-diag would you be able to fix the tests?

MikeOverkamp-diag:
Ah yes, of course

jmsmkn (approved):
LGTM

jmsmkn:
Sorry - reviewed too early. The approach looks good but obviously, the tests need to be fixed first!

PR #1661, opened 2020-12-18
change ordering, include distributions in search, include totalcount

Kicksta:


Kicksta:
Since moving to WSL2, I am using VS code and I haven't got black and flake8 configured. Will try to get that back in place and recommit.

app/grandchallenge/products/views.py
```diff
@@ -26,6 +27,7 @@ def get_queryset(self):
         ce_class_query = self.request.GET.get("ce_class")
         fda_class_query = self.request.GET.get("fda_class")
         search_query = self.request.GET.get("search")
+        product_total_all = queryset.count()
```
jmsmkn, line 30:
```suggestion
        self.product_total_all = queryset.count()
```

app/grandchallenge/products/views.py
```diff
@@ -120,8 +123,10 @@ def get_context_data(self, *args, **kwargs):
                 "selected_fda_class": fda_class_query,
                 "products_selected_page": True,
                 "product_total": context["object_list"].count(),
+                "product_total_all": Product.objects.filter(ce_status=Status.CERTIFIED).count(),
```
jmsmkn, line 126:
```suggestion
                "product_total_all": self.product_total_all,
```

app/grandchallenge/products/views.py
```diff
@@ -120,8 +123,10 @@ def get_context_data(self, *args, **kwargs):
...
             }
         )
+        print(context)
```
jmsmkn, line 129:
```suggestion
```

app/grandchallenge/products/views.py
```diff
@@ -16,8 +16,9 @@ class ProductList(ListView):
     model = Product
     context_object_name = "products"
     queryset = Product.objects.filter(ce_status=Status.CERTIFIED).order_by(
-        "-verified", "-ce_verified", "company__company_name"
+        "-verified", "-ce_verified", "product_name"
     )
+    product_total_all = queryset.count()
```
jmsmkn, line 21:
```suggestion
```

jmsmkn (changes requested):
I've made a few suggestions that should reduce the number of database calls for this as every time you call `queryset.count()` the query is executed. Please check if it works as intended when they're all applied.

What I've changed it to - we store the count of the unfiltered queryset on the view object in L30, then filter it in the rest of `get_queryset`, then lookup the number again in L126.

jmsmkn:
Did this work? I was making the suggestions blind. Also, did you get pre-commit set back up?

Kicksta:
The code worked, checked it before the holidays, but I am having trouble with pre-commit. If I understand correctly [it isn't available for Ubuntu 20.04](https://packages.ubuntu.com/search?keywords=pre-commit&searchon=names&suite=focal&section=all). So after some googling figured out I could use [snap](https://snapcraft.io/install/pre-commit/ubuntu) for it. But using `snap install pre-commit` gives the following: `error: cannot communicate with server: Post http://localhost/v2/snaps/pre-commit: dial unix /run/snapd.socket: connect: no such file or directory` Now I have found two options of which both I am not sure what they do and I am somewhat hesitant to give them a try, however it seems to help people with similar issues: [option 1](https://github.com/microsoft/WSL/issues/2374#issuecomment-699110721), [option 2](https://github.com/DamionGans/ubuntu-wsl2-systemd-script). Any advice on which one to try while not breaking things...?

jmsmkn:
The best way is via pip, they provide a single command to set everything up

```
curl https://pre-commit.com/install-local.py | python -
```

See https://pre-commit.com/#installation

jmsmkn (approved):

Issue #1659, opened 2020-12-17
Show how many Algorithms are found, just like for Challenges


bramvanginneken:
![image](https://user-images.githubusercontent.com/1818749/102520857-7360e700-4094-11eb-81dc-549e86756aa2.png)

Below the filter option it is shown how many Challenges are found, this is not the case for Algorithms, but would be nice there too.

GeorgeJoLo:
Hello! I'm very new in github (meaning this is my first repository visiting). I want to contribute to something. I think I can handle this one. What should I do? 
Is it still open right?

jmsmkn:
Hi @GeorgeJoLo! Great that your interested in contributing to open source, and this project in particular. You don't seem to have any activity in your GitHub yet so I'm unsure how much guidance you would need. This is a web app built on Django, do you have any experience with Django, or web development? If not, I would start with the Django tutorial (https://www.djangoproject.com/start/ and/or https://learndjango.com/tutorials/how-learn-django) and build a couple of apps for yourself first. Once you're familiar with Django, then follow the developer documentation for this project to get up and running (https://comic.github.io/grand-challenge.org/development.html).

jmsmkn:
Closed by #1681

PR #1681, opened 2021-01-18
Unify backend-code for counters in list views

chrisvanrun:
Counts of found/filtered results for Archives/Reader Studies/Challenges/Algorithms are now displayed using the same backend logic.

For instance:
- "_No algorithms found._"
- "_1 algorithm found_"
- "_10 algorithms found_"

Or when a filter was applied:
- "_Filters included 1 of 10 total algorithms_"

Addresses #1659


MikeOverkamp-diag (approved):
LGTM!

app/grandchallenge/algorithms/views.py
```diff
@@ -85,7 +85,7 @@ def form_valid(self, form):
         return response
 
 
-class AlgorithmList(PermissionListMixin, FilterMixin, ListView):
+class AlgorithmList(FilterMixin, PermissionListMixin, ListView):
```
jmsmkn, line 88: Why have the mixins changed order here? Is this to get the number of objects pre-permissions filtering?

chrisvanrun: The other way around. It's to get the correct number of objects on the context generation in the `FilterMixin`; post-permissions. The method resolution order had the permissions exclude objects after the `FilterMixin` making the mixin count objects that would not have been visible to the user anyway.

jmsmkn: You're right, my mistake!

PR #1664, opened 2020-12-22
Use ComponentInterfaces to define output of algorithm and challenges

miriam-groeneveld:
For deployment `set_default_output_interfaces` needs to be run for all existing algorithms and phases.

Closes #1662

PR #1683, opened 2021-01-19
Consolidates output loading

jmsmkn:
I noticed when reviewing #1664 that there was still some duplicated interface loading methods, so this consolidates them. It also allows us to remove the `AlgorithmExecutor` class, and I think shows a way forward for using interface inputs (but, that's for later). I have left a couple of TODO notes, not sure where the CIV loading code should live but it does not matter when docker is the only backend.


miriam-groeneveld (approved):
Much cleaner now by using output_interfaces as property, thanks

PR #1684, opened 2021-01-19
Use common user group management forms, views and templates

jmsmkn:
Refactoring to reduce duplicated functionality.


MikeOverkamp-diag (approved):
LGTM

PR #1686, opened 2021-01-20
Use PaginatedTableListView for Images List in a reader study

### LINKED ISSUE #1668: Use Paginated DataTables for the Image List view in a Reader Study
**Problem**
In order to view the cases in a reader study, I need to sequentially scroll through pages one by one. When a reader study contains hundreds of images this is not doable anymore. 

**Proposed solution**
Archives allow the option of searching through the list of cases/ images and also to jump ahead several pages when browsing through the images. It would be very helpful if this feature would also be available for reader studies.

**Additional context**
This improved case navigation would be most useful when there is also an option to select the cases and perform an action e.g. delete them all at once. There is an existing feature request for selecting and deleting multiple cases  #1600 that could perhaps be combined with this one.

jmsmkn:
The only tricky thing in refactoring this was the JS interaction for image remove, would be worth double checking that, unfortunately the diff is messed up as a file was renamed but it's not that much of a change.

Closes #1668


MikeOverkamp-diag (approved):
Js changes look fine to me.

PR #1688, opened 2021-01-21
1682 add number answer type for reader studies

amickan:
- added number answer type and applied migrations
- added tests for the new answer type


jmsmkn (approved):
LGTM!

PR #1690, opened 2021-01-25
add meta property

Kicksta:
I wouldn't know how to check this locally, so in case the code itself looks okay I will make sure to check the functioning of it when it is live.


jmsmkn (changes requested):
You're not including `{{ block.super }}` so you are now missing some important tags that are included with https://github.com/comic/grand-challenge.org/blob/master/app/grandchallenge/core/templates/grandchallenge/partials/meta.html:

```
<meta charset="UTF-8">
<meta name="viewport" content="width=device-width, initial-scale=1">
```

These are required for bootstrap to work. Maybe the partial need to be split so that you can override only the social metadata.

The tests need to be fixed too.

Kicksta:
What do you prefer? The `{{block.super}}` in the aiforradiology code or shall I change `core/templates/base.html` and `meta.html`? I can imagine the splitting of the meta leaves a bit more freedom for the tags, either still using `{{block.super}}` to partly overwrite the defaults or just defining the tags needed.

For the latter, just to check if I understand correctly, in the base file I would do...
```
{% block meta %}
      {% include "grandchallenge/partials/meta.html" %}
{% endblock %}

{% block meta_social %}
      {% include "grandchallenge/partials/meta_social.html" %}
{% endblock %}
```

... creating a `meta_social.html` file, splitting the first few lines (meta) from all meta name and meta property tags (meta_social)?

jmsmkn:
Yes, `meta.html` and `meta_social.html` looks great! You will need to change both `core/base.html` and `products/base.html` to use these new partials.

app/grandchallenge/products/templates/products/base.html
```diff
@@ -19,6 +19,15 @@
     </style>
 {% endblock %}
 
+{% block meta_social %}
+    <meta name="title" content="AI for Radiology">
+    <meta property="og:title" content="AI for Radiology" />
+    <meta property="og:url" content="https://www.aiforradiology.com" />
+    <meta property="og:image" content= "{% static 'products/images/logo-collage-aiforradiology.JPG' %}" />
+    <meta property="og:description" content="A platform to help you find the right AI software for radiology" />
+    {{ block.super }}
```
jmsmkn, line 28: I don't think that you need `{{ block.super }}` in any of these 3 places now that you have split out the other meta params. If you keep it then the properties defined in the super block will override the ones you define here.

jmsmkn: Sorry if I caused any confusion from my first review!

Kicksta: Oh, my understanding was it would complement the items that I have not defined. For example, I do not define the twitter card which is defined in `core/base`, I thought to use it as some sort of fallback mechanism, that if the item is not defined it takes it from `core/base` because of the `{{  block.super }}`. But if it overwrites it I should indeed leave it out. 
https://docs.djangoproject.com/en/3.1/howto/overriding-templates/#extending-an-overridden-template

jmsmkn: The key point in the docs is the 3rd there:

> The template replaces just the branding block, adding a custom logo, and using block.super to retain the prior content.

I think what is missing from the docs is that the prior content doesn't contain an image already. If it did, you would be adding 2nd image before the 1st included in block.super.

Unfortunately I don't know a way of doing property merging in templates like you'd like, but it would be possible with a python dictionary in a context variable.

jmsmkn (approved):
LGTM

app/grandchallenge/products/templates/products/base.html
Kicksta: Thanks for explaining!

PR #1691, opened 2021-01-26
Improve the user search and selection form

### LINKED ISSUE #1685: Improve the user search and selection form
In #1684 the user selection for groups was consolidated, so now it would be good to improve it. The form uses a ModelSelect2 widget (https://github.com/comic/grand-challenge.org/blob/master/app/grandchallenge/groups/forms.py#L19) that calls an autocomplete view that only allows filtering by username (https://github.com/comic/grand-challenge.org/blob/master/app/grandchallenge/groups/views.py#L67). 

Top priority is to extend the filter:
- Switch to filtering with `icontains` rather than `istartswith`
- Allow filtering on the users username, email address, full name, OR verification email address. To do this use Django `Q` objects (https://docs.djangoproject.com/en/3.1/topics/db/queries/#complex-lookups-with-q-objects).

Next, is to improve the look of the widget, at the moment you just get a string of the username:
image

On github it's much nicer
image

Extend this to
- Include the users full name in the string
- Show whether the user is verified or not in the string, and if so, what is the domain of their verification address (see https://github.com/comic/grand-challenge.org/blob/master/app/grandchallenge/profiles/templatetags/profiles.py#L16)
- Stretch goal: Show the users mugshot

amickan:
Closes #1685

### DIFF
--- a/app/grandchallenge/groups/views.py
+++ b/app/grandchallenge/groups/views.py
@@ -3,15 +3,16 @@ from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.contrib.auth.mixins import UserPassesTestMixin
 from django.contrib.messages.views import SuccessMessageMixin
+from django.db.models import CharField, Q, Value
+from django.db.models.functions import Concat
+from django.utils.html import format_html
 from django.views.generic import FormView
 from guardian.mixins import (
     LoginRequiredMixin,
     PermissionRequiredMixin as ObjectPermissionRequiredMixin,
 )
 from guardian.shortcuts import get_objects_for_user
-from django.db.models import Q, Value, CharField
-from django.db.models.functions import Concat
-from django.utils.html import format_html
+
 from grandchallenge.verifications.models import Verification
 
 
@@ -94,7 +95,7 @@ class UserAutocomplete(
         if is_verified:
             return format_html(
                 '<img src="{}" width ="20" height ="20" style="vertical-align:top"> '
-                '&nbsp; <b>{}</b> &nbsp; {} &nbsp; Verified at {} &nbsp; '
+                "&nbsp; <b>{}</b> &nbsp; {} &nbsp; Verified at {} &nbsp; "
                 '<i class="fas fa-user-check text-success">',
                 result.user_profile.get_mugshot_url(),
                 result.get_username(),
@@ -104,7 +105,7 @@ class UserAutocomplete(
         else:
             return format_html(
                 '<img src="{}" width ="20" height ="20" style="vertical-align:top"> '
-                '&nbsp; <b>{}</b> &nbsp; {} &nbsp; not verified',
+                "&nbsp; <b>{}</b> &nbsp; {} &nbsp; not verified",
                 result.user_profile.get_mugshot_url(),
                 result.get_username(),
                 result.get_full_name().title(),
--- a/app/grandchallenge/groups/views.py
+++ b/app/grandchallenge/groups/views.py
@@ -6,7 +6,6 @@ from django.contrib.messages.views import SuccessMessageMixin
 from django.db.models import CharField, Q, Value
 from django.db.models.functions import Concat
 from django.utils.html import format_html
-from django.utils.safestring import mark_safe
 from django.views.generic import FormView
 from guardian.mixins import (
     LoginRequiredMixin,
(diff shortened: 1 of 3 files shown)


app/grandchallenge/groups/views.py
```diff
@@ -61,9 +68,50 @@ def get_queryset(self):
...
+            .select_related("verification", "user_profile")
         )
 
         if self.q:
-            qs = qs.filter(username__istartswith=self.q)
+            qs = qs.filter(
+                Q(username__icontains=self.q)
+                | Q(email__icontains=self.q)
+                | Q(full_name__icontains=self.q)
+                | Q(verification__email=self.q)
```
jmsmkn, line 87:
```suggestion
                | Q(verification__email__icontains=self.q)
```

app/grandchallenge/groups/views.py
```diff
@@ -61,9 +68,50 @@ def get_queryset(self):
...
+        except Verification.DoesNotExist:
+            is_verified = False
+
+        if is_verified:
+            return format_html(
+                '<img src="{}" width ="20" height ="20" style="vertical-align:top"> '
+                "&nbsp; <b>{}</b> &nbsp; {} &nbsp;"
+                '<i class="fas fa-user-check text-success" '
+                'title="Verified email address at {}">',
+                mark_safe(result.user_profile.get_mugshot_url()),
```
jmsmkn, line 105: Do you really need `mark_safe` here? I think that it is not needed eg, https://github.com/comic/grand-challenge.org/blob/master/app/grandchallenge/profiles/templatetags/profiles.py#L24

app/tests/groups_tests/test_views.py
```diff
@@ -81,3 +89,106 @@ def get_user_autocomplete():
...
+    def test_autocomplete_filter_options(self, client, filter, is_verified):
+        archive = ArchiveFactory()
+
+        admin = UserFactory()
+        archive.add_editor(admin)
+        first_name = "Jane"
+        last_name = "Doe"
+
+        if is_verified:
+            call_command("check_permissions")
```
jmsmkn, line 111: `check_permissions` should only be called once at test setup (outside of the individual tests), why is it being called again here?

jmsmkn (changes requested):
Looks great! A couple of small comments. Would also be great to include a small video to demonstrate the feature, I use OBS to record those, save them in mp4 and then upload them to the PR comment. It makes communicating UI features to the users much easier.

app/grandchallenge/groups/views.py
HarmvZ: It's not strictly necessary because the mugshot urls on production do not contain html unsafe characters. But we noticed during testing that the mugshot url that it uses then contains a `&` which will be converted to `&amp;` by `format_html` which would make the url invalid. So I think it should also be added on the line you refer to. 
Although we should be careful with this because it could potentially introduce an XSS vulnerability (I guess the mugshot url is user input in essence?), so maybe since it is not necessary for production we should remove it to be safe.

jmsmkn: I'm pretty sure that the user file names are cleaned and normalized so that they do not contain special symbols, and yes, better to remove it than risk an XSS vulnerability.

amickan: ok, removing it sounds good, but then we again have the issue that the test no longer works, so is it ok then to not test whether the mugshot url is presented?

app/tests/groups_tests/test_views.py
amickan: I'm not sure I understand what you mean by calling it at test setup. We put it here because it seemed necessary to create a verified user. Where should it go instead?

app/grandchallenge/groups/views.py
jmsmkn: Yes, use `html_escape` in the tests. This is automatically applied in Django templates, but sometimes we need to be explicit about it.

app/tests/groups_tests/test_views.py
jmsmkn: Ah yes, I took a look at the verification tests and it's called there too (https://github.com/comic/grand-challenge.org/blob/master/app/tests/verification_tests/test_forms.py#L85), it _shouldn't_ be necessary, but looks like it is. We don't need to solve that here. OK to leave this as it.

jmsmkn (approved):
LGTM

PR #1694, opened 2021-01-28
Moves verification domain out of title

jmsmkn:
Small change - when adding users to groups I think that it's important people see which domain the user has verified with, so this saves users having to hover over the icon.


amickan (approved):
Looks good to me!

PR #1696, opened 2021-02-01
1498 social media pictures

### LINKED ISSUE #1498: Pictures for messages on social media
Challenges, algorithms, Reader studies and archive have a logo that is used in the cards. Would be great if when you'd post a link for these things on social media (eg linkedinb) this logo image is shown. For challenges the banner image is now shown but that image has a Recommended resolution 2200x440 px, and is too wide to fit nicely. For algorithms the default gc image is shown, whioch does not fit at all. For algorithms it would also be nice if the logo is shown on the page itself, on top.

amickan:
Closes #1498


jmsmkn (changes requested):
Looks great! Could you make a couple of changes:

- Please update the resolution in the help text to 640×320px (1280×640px for best display).
- Could you remove and regenerate the new migrations that you have generated? For PRs there should only be 1 new migration per application (eg, Algorithms right now is introducing two: 0004 and 0005). It takes time to run these so it's best practise to keep the migration history clean.

marcopus:
@amickan @jmsmkn I was playing with a local instance of GC when I noticed this `help_text` under New Archive:
> square with a resolution of 640x320 px

Sounds a bit awkward to me.. Maybe leave out the "square" part?

PR #1698, opened 2021-02-01
1693 algorithm output as file

### LINKED ISSUE #1693: Add option to save algorithm output as file
ComponentInterfaces are used to define the output of an algorithm. As the output of an algorithm might be a large json file, we need to be able to save these outputs as files instead of database fields. ComponentInterface needs to be expanded with a boolean to make this distinction.

miriam-groeneveld:
Added Boolean field to determine how output value should be saved. 
Added file serving for output files.

Closes #1693


app/config/settings.py
```diff
@@ -140,6 +140,7 @@ def strtobool(val) -> bool:
 JQFILEUPLOAD_UPLOAD_SUBIDRECTORY = "jqfileupload"
 IMAGE_FILES_SUBDIRECTORY = "images"
 EVALUATION_FILES_SUBDIRECTORY = "evaluation"
+COMPONENTS_FILES_SUBDIRECTORY = "components"
```
jmsmkn, line 143: Is this used?

jmsmkn: Yes, it is, my search doesn't work and this should have been part of a review not a single comment...

app/grandchallenge/serving/urls.py
```diff
@@ -41,4 +42,15 @@ def to_url(self, value):
         ),
         serve_submissions,
     ),
+    path(
+        (
+            f"{settings.COMPONENTS_FILES_SUBDIRECTORY}/"
+            f"componentinterfacevalue/"
+            f"<uuidprefix:pa>/"
```
jmsmkn, line 49: Are we really looking for `uuidprefix` types here? I don't think that this matches the definition in `component_interface_value_path`

app/grandchallenge/serving/views.py
```diff
@@ -74,3 +75,33 @@ def serve_submissions(request, *, submission_pk, **_):
         )
 
     raise PermissionDenied
+
+
+def serve_component_interface_value(
```
jmsmkn, line 80: Could we get some tests on this function? It's really annoying to test but it's important when dealing with permissions. See https://github.com/comic/grand-challenge.org/blob/24ea0061be90a9fca3ca093d847e206052e9e7b0/app/tests/serving_tests/test_views.py#L16

app/grandchallenge/serving/views.py
```diff
@@ -74,3 +75,33 @@ def serve_submissions(request, *, submission_pk, **_):
...
+            .first()
+        )
+
+    except ComponentInterfaceValue.DoesNotExist:
+        raise Http404("No ComponentInterfaceValue found.")
+
+    if request.user.has_perm(
+        "view_job", civ.algorithms_jobs_as_output.first()
+    ):
+        create_download.apply_async(
```
jmsmkn, line 99: This will always fail as create_download does not take `component_interface_value_id` as a kwarg. We do not need to track downloads of these files IMO.

app/grandchallenge/components/models.py
```diff
@@ -182,29 +190,34 @@ def _create_images_result(self, *, reader, job):
...
+                )
+
+            civ = ComponentInterfaceValue.objects.create(
+                interface=self, value=result
             )
-        except JSONDecodeError as e:
-            raise ComponentException(f"Could not decode results file: {e.msg}")
+        else:
+            civ = ComponentInterfaceValue.objects.create(interface=self,)
+            civ.file.save(str(output_file.name), File(file))
```
jmsmkn, line 219: Didn't we want to do validation of these files here too?

jmsmkn (commented):
Looks good! A few questions

app/grandchallenge/serving/views.py
miriam-groeneveld: Yes, hadn't committed the test yet.

app/grandchallenge/serving/views.py
miriam-groeneveld: ok, I'll remove the create_download. I did wonder about the serving of these files. When I access one of these files, I don't get a download option but rather the browser shows the json content. This exposes the minio url, not sure if that's ok

jmsmkn: Yes, that's perfect, in production this will be CloudFront. If you would like to download the file instead use the download attribute in the template https://www.w3schools.com/tags/att_a_download.asp

app/grandchallenge/serving/views.py
jmsmkn: Ah! Sorry, didn't notice this was a draft. My fault.

miriam-groeneveld: I changed it to a draft, should have done that right away. Got distracted...

app/grandchallenge/serving/urls.py
miriam-groeneveld: Should I add a converter here for the `pk_as_padded_hex` that is used in the model?

miriam-groeneveld: Renamed it to more generic prefix converter. If you'd still rather just want comments, I'll change it back

miriam-groeneveld:
I think this is ready to go now.

app/grandchallenge/serving/views.py
```diff
@@ -74,3 +75,30 @@ def serve_submissions(request, *, submission_pk, **_):
...
+
+    try:
+        # output should only be connected to a single job; throw error if not?
+        civ = (
+            ComponentInterfaceValue.objects.filter(
+                pk=component_interface_value_pk
+            )
+            .prefetch_related("algorithms_jobs_as_output")
+            .first()
+        )
```
jmsmkn, line 96: Rather than doing a filter and then `first()`, it is better to do a `get(pk=component_interface_value_pk)`, catching the `MultipleObjectsReturned` exception if necessary. You can still add the `prefetch_related` before the `get`.

app/grandchallenge/serving/views.py
```diff
@@ -74,3 +75,30 @@ def serve_submissions(request, *, submission_pk, **_):
...
+
+    except ComponentInterfaceValue.DoesNotExist:
+        raise Http404("No ComponentInterfaceValue found.")
+
+    if user.has_perm("view_job", civ.algorithms_jobs_as_output.first()):
```
jmsmkn, line 101: We can squash the multiple jobs problem by re-writing this a little. Something like:

```
get_objects_for_user(user=user, perms="algorithms.view_job").filter(outputs__pk=component_interface_value_pk).exists()
```

I've not tested this so we would need to make sure that really works, but don't spend too much time on that if it doesn't.

app/grandchallenge/components/models.py
```diff
@@ -227,15 +243,34 @@ class ComponentInterfaceValue(models.Model):
...
+    image = models.ForeignKey(
+        to=Image, null=True, blank=True, on_delete=models.CASCADE
     )
-    image = models.ForeignKey(to=Image, null=True, on_delete=models.CASCADE)
 
     def __str__(self):
         return f"Component Interface Value {self.pk} for {self.interface}"
 
+    def save(self, *args, **kwargs):
+        self.full_clean()
```
jmsmkn, line 271: I'm not in favour of adding `full_clean` to save as it means that any callers of `save` now need to handle a `ValidationError`, and I'm not confident that this is handled everywhere within django itself. Could you move this to where it is needed?

jmsmkn (commented):
Looks good! A couple of questions, and still the one from last time about validation.

app/grandchallenge/serving/views.py
jmsmkn: But, I think if you implement my comment on L98 then the prefetch related is unnecessary.

app/grandchallenge/components/models.py
jmsmkn: I think this is to handle the `CIV.create()`, right? Without doing `full_clean` in `save`, `CIV.create(...)` would become:

```
CIV(...)
CIV.full_clean()  # This line would then go in a try: ... except ValidationError: ... block for handling
CIV.save()
```

miriam-groeneveld:
> Looks good! A couple of questions, and still the one from last time about validation.

Not sure what kind of validation should be added besides the mimetype and extension validation that have been added already?

jmsmkn:
> > Looks good! A couple of questions, and still the one from last time about validation.
> 
> Not sure what kind of validation should be added besides the mimetype and extension validation that have been added already?

Sorry! I was thinking of reader studies where we do validate the json values - I was thinking "does this file really contain a json structure of the form `MPOI`", etc. But I realise this feature is not implemented at all. It would be great to have a schema field on the `ComponentInterface`, or to use the ones defined in the reader studies, but that is obviously a feature request that is orthogonal to this PR. Another argument for unifying `Answer`s and `ComponentInterfaceValue`s. So, ignore this :-)

Is the rest finished?

miriam-groeneveld:
Yes, the rest is finished.

jmsmkn (approved):
LGTM!

PR #1704, opened 2021-02-03
Clean code of reader-study detail and readers progress pages

chrisvanrun:
Removes dead JS and, in style, add a trailing `/` to a path route definition.


jmsmkn (approved):
LGTM, thanks!

PR #1705, opened 2021-02-03
change url, add type, adapt image width

Kicksta:


app/grandchallenge/products/templates/products/base.html
```diff
@@ -22,9 +22,10 @@
 {% block meta_social %}
     <meta name="title" content="AI for Radiology">
     <meta property="og:title" content="AI for Radiology" />
-    <meta property="og:url" content="https://www.aiforradiology.com" />
+    <meta property="og:url" content="//www.AIforRadiology.com/" />
```
jmsmkn, line 25: Question - is this right without the `https:`?

jmsmkn (approved):
LGTM, just a question for my understanding

app/grandchallenge/products/templates/products/base.html
Kicksta: not sure, currently it doesn't show the url of www.aiforradiology.com but the www.grand-challenge.org. https://www.linkedin.com/post-inspector/inspect/aiforradiology.com
I saw in some documentation they left out the `https`, so I thought to give that a try, but I guess little chance that that would improve it. Any suggestions? I think now with grand-challenge as url it looks a bit inconsistent.

jmsmkn: Unsure, unfortunately, I cannot access the LinkedIn validator, no idea why that doesn't work at all for me. I tend to use the twitter one but the aiforradiology cards do not work at all there (https://cards-dev.twitter.com/validator). It might not be possible, imagine if you owned `stealyourcreditcarddetails.com` and then posted cards that said it was content `amazon.com` or something. Let try this and see.

We also need a name as grand-challenge is too specific... Contributions welcome there!

Kicksta: that makes sense... didn't think of that. I don't really use twitter so didn't check that one, noticed they use their own tags (not og) I should look into it though in case anyone else likes to promote the page ;)

You mean a substitute name for grand-challenge.org all together? Hmm.. thinking out loud here.
What connects everything: medical imaging, algorithms, the whole process from development to product.
- medimp (based on MEDical IMaging Processing, not to complex as a name, and seems available)
- medrithms (medical algorithms, but from the sound you don't know how to spell it - could be rhythm - one of my least favorite words to spell in English btw)
- umed-imaging
- mimagine (medical, imagine, imaging)
- grandc (staying close to home but making it more of a brandname)
- ....

For further inspiration: https://namelix.com/app/?keywords=medical+imaging+algorithms+artificial+intelligence ;)

PR #1703, opened 2021-02-02
Reduce load-time of the reader-study detail page

chrisvanrun:
# Closes #1158 
* decoupling the readers progress from the reader-study detail page (for editors)
* split reader list+progress into 
  * reader list (in page)
  * reader progress (separate page)

# Misc.:
* Uses user group management templates
* Readers now have a badge that shows the number of readers

# Performance
The current load time for the editor page for 110 readers is *12.7s*. 
The new load time is *2.6s* for the reader-study page (for editors) and *6.3s* for the reader progress overview page.

The reader progress overview page is not yet as optimized as I'd like but I am calling it "good enough" .

## Current

https://user-images.githubusercontent.com/7871310/106597360-98cb8680-6556-11eb-8a18-e929a777b2c9.mp4

## This PR

https://user-images.githubusercontent.com/7871310/106597374-9ec16780-6556-11eb-828a-b224ced6639a.mp4

PR #1706, opened 2021-02-03
Add cases nav item back to reader study page

MikeOverkamp-diag:
The menu item was accidentally removed in https://github.com/comic/grand-challenge.org/pull/1703/files


chrisvanrun (approved):
LGTM! Must have missed that when having edited the nav for the 4th time or such in a PR =P

PR #1707, opened 2021-02-05
Catch celery soft time limit for image imports

### LINKED ISSUE #1695: Image import job does not get marked as failed when the time limit is exceeded
Image import jobs should catch the celery soft time limit errors, mark the import job as failed and clean up before the hard time limit is reached. This is handled correctly for component jobs.

amickan:
Image import jobs now get marked as failed when they exceed the celery soft time limit and are cleaned up before the hard time limit is reached. 

Closes #1695


jmsmkn (approved):
LGTM

PR #1710, opened 2021-02-09
Add remove button for questions in reader studies

### LINKED ISSUE #1544: Remove questions in reader study
A question in a Reader Study cannot be removed, even if it has no answers yet. 

I think this should at least be possible when there are no answers yet. There are even use cases when I would want to remove a question after it had a few answers (of course, I would want to be warned, but then it should be possible).

An alternative solution, but probably a bit harder to implement, is to mark the Reader Study as 'Draft' or 'In Development', so that editors can play around with questions and dummy answers. Once the questions are confirmed and the reader study is actually sent to readers, it can be disallowed to remove questions.

amickan:
Editors of reader studies can now delete questions if the question has no answers associated to it yet. If a question has answers, deleting is not possible and editors first have to manually remove the answers.

https://user-images.githubusercontent.com/30069334/107365067-93cd8080-6adc-11eb-981f-83953f48f948.mp4

Closes #1544

### DIFF
--- a/app/grandchallenge/reader_studies/templates/reader_studies/question_confirm_delete.html
+++ b/app/grandchallenge/reader_studies/templates/reader_studies/question_confirm_delete.html
@@ -1,28 +1,18 @@
 {% extends "base.html" %}
 
 {% block content %}
-
-    {% if not object.is_fully_editable %}
-        <h2 class="text-danger">Not possible to delete this question</h2>
+    <h2>Confirm Deletion</h2>
+    <form action="" method="post">
         {% csrf_token %}
-        <p>The question "{{ object }}" already has answers associated to it. <br>You first need to delete all answers for this question before you can delete it. <br></p>
+        <p>Are you sure that you want to delete the question "{{ object }}"? <br>
+            <b class="text-danger">WARNING:
+                You are not able to undo this action. Once the question is deleted
+                it is deleted forever.</b></p>
         <a href="{% url 'reader-studies:detail' slug=object.reader_study.slug %}"
            type="button"
            class="btn btn-secondary">Cancel</a>
-    {% else %}
-        <h2>Confirm Deletion</h2>
-        <form action="" method="post">
-            {% csrf_token %}
-            <p>Are you sure that you want to delete the question "{{ object }}"? <br>
-                <b class="text-danger">WARNING:
-                    You are not able to undo this action. Once the question is deleted
-                    it is deleted forever.</b></p>
-            <a href="{% url 'reader-studies:detail' slug=object.reader_study.slug %}"
-               type="button"
-               class="btn btn-secondary">Cancel</a>
-            <input type="submit"
-                   value="I understand, delete the question '{{ object }}'"
-                   class="btn btn-danger"/>
-        </form>
-    {% endif %}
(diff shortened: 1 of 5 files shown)


app/grandchallenge/reader_studies/templates/reader_studies/question_confirm_delete.html
```diff
@@ -0,0 +1,28 @@
+{% extends "base.html" %}
+
+{% block content %}
+
+    {% if not object.is_fully_editable %}
+        <h2 class="text-danger">Not possible to delete this question</h2>
+        {% csrf_token %}
```
MikeOverkamp-diag, line 7: You only need to add the csrf_token to forms, it's not needed here.

app/grandchallenge/reader_studies/templates/reader_studies/question_confirm_delete.html
```diff
@@ -0,0 +1,28 @@
...
+        <p>The question "{{ object }}" already has answers associated to it. <br>You first need to delete all answers for this question before you can delete it. <br></p>
```
MikeOverkamp-diag, line 8: This is nitpicking, but I'd prefer `associated with`

app/tests/reader_studies_tests/test_views.py
```diff
@@ -128,3 +128,92 @@ def test_answer_remove(client):
...
+    assert Answer.objects.count() == 0
+
+    response = get_view_for_user(
+        viewname="reader-studies:question-delete",
+        client=client,
+        method=client.get,
+        reverse_kwargs={"slug": rs.slug, "pk": q.pk},
+        user=editor,
+    )
+
```
MikeOverkamp-diag, line 218: It would also be nice to assert here that the question has actually been removed. I know you already tested that in the other test case, but I still think it would be a good assertion here as well, just in case something weird happens if a question has previously had answers associated with it.

MikeOverkamp-diag (changes requested):
LGTM! Just a couple of very minor comments.

app/tests/reader_studies_tests/test_views.py
amickan: The get method that I'm calling doesn't actually delete the question. I would need to use the post method instead for the deletion to succeed. but the post method succeeds in all cases (even if the question has answers), so that didn't make all that much sense. What do you suggest I should do?

MikeOverkamp-diag: Ah that's a good point and something I missed. What you could probably do is add a check here: https://github.com/comic/grand-challenge.org/pull/1710/files#diff-d1601f07f94a69bc5526b8feb2501e91db4d928e1bfad9062514905e6a316d65R978

You could just add an if-statement before you add the success message and raise a forbidden if there are any answers for the question. Let me know if you need any help with that.

jmsmkn:
A request for the list view - could you disable the buttons if the question is not deletable, and then use something along the lines of `title="This question cannot be deleted as..."`? eg https://getbootstrap.com/docs/4.0/components/buttons/#disabled-state

MikeOverkamp-diag (approved):

PR #1712, opened 2021-02-10
Add RGB support for dicom importer

MikeOverkamp-diag:
This also fixes issues with importing dicom files that contain multiple slices as well as improves the importer's memory usage.


MikeOverkamp-diag:
Unfortunately I do not have good files to add tests for this, so this has only been tested manually.

jmsmkn (approved):

PR #1716, opened 2021-02-15
Show all users reader study progress

jmsmkn:
Add the breadcrumbs to the question delete view

The page now includes all users who have answered questions, plus users in the editors group
who can also read the study. Removes the "Remove Reader" button as this is available elsewhere.


chrisvanrun:
> Removes the "Remove Reader" button as this is available elsewhere.
I left this in initially because I envisioned that an editor might want to remove readers after a while that had poor progression but it looks cleaner when left out.

Am I correct in assuming that 'Users' > 'Readers'?

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
```diff
@@ -32,12 +32,12 @@
...
                    href="#v-pills-readers" role="tab" aria-controls="v-pills-readers"
-                   aria-selected="false"><i class="fas fa-users fa-fw"></i> Readers &nbsp;<span
+                   aria-selected="false"><i class="fas fa-users fa-fw"></i>&nbsp;Readers&nbsp;<span
                         class="badge badge-pill badge-secondary align-middle">{{ num_readers }}</span>
                 </a>
                 <a class="nav-link" id="v-pills-readers-tab"
-                   href="{% url 'reader-studies:readers-progress' slug=object.slug %}"
-                ><i class="fas fa-tasks fa-fw"></i> Readers Progress
+                   href="{% url 'reader-studies:users-progress' slug=object.slug %}"
+                ><i class="fas fa-tasks fa-fw"></i>&nbsp;Users Progress
```
chrisvanrun, line 40: Just curious: is the logic behind replacing spaces with `&nbsp;` for outlining?

jmsmkn: Non-breaking space, we never want the browser to insert a line break there.

jmsmkn:
Editors can also answer questions, so we need to show their progress too. Also, a user can be in the readers group, create some answers, then be removed from the readers group. In that case, there is no way for their questions to be removed without adding them back to the readers group (which user?) and then removing their answers, and then removing the user from the readers group again.

PR #1719, opened 2021-02-16
Add pre-commit linting of template files

jmsmkn:
With many developers now working on the project I've noticed that the template files are starting to get some errors. This PR adds `curlylint` as a pre-commit hook, which doesn't fix template errors but at least highlights them. Also fixes a few missing closing tags, indentation fixes will follow probably in another PR.

### DIFF
--- a/app/grandchallenge/algorithms/templates/algorithms/job_list_row.html
+++ b/app/grandchallenge/algorithms/templates/algorithms/job_list_row.html
@@ -11,13 +11,13 @@
    title="View result details">
     <i class="fa fa-info-circle"></i>
 </a>
-<split/>
+<split></split>
 
 {{ object.created|naturaltime }}
-<split/>
+<split></split>
 
 {{ object.creator|user_profile_link }}
-<split/>
+<split></split>
 
 <a href="#resultInfoModal"
    class="badge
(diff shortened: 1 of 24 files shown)


app/grandchallenge/algorithms/templates/algorithms/job_list_row.html
```diff
@@ -11,13 +11,13 @@
    title="View result details">
     <i class="fa fa-info-circle"></i>
 </a>
-<split/>
+<split></split>
```
MikeOverkamp-diag, line 14: I think we need to revert these changes and all similar ones in the other templates. Or we need to change the processing here https://github.com/comic/grand-challenge.org/blob/master/app/grandchallenge/datatables/views.py#L30, which may be a better option.

MikeOverkamp-diag (changes requested):
Just one comment.

app/grandchallenge/algorithms/templates/algorithms/job_list_row.html
jmsmkn: The change to `datatables/views.py` is included in this PR.

MikeOverkamp-diag: Wow, sorry, I somehow completely missed that.

MikeOverkamp-diag (approved):

app/grandchallenge/algorithms/templates/algorithms/job_list_row.html
jmsmkn: That's ok! The line is also covered well by tests.

PR #1723, opened 2021-02-17
Replace result_dict with results

jmsmkn:
Rename this variable as it often is not a dictionary, keeps the old variable around in the template to give us some time to migrate.

### DIFF
--- a/app/grandchallenge/algorithms/migrations/0001_initial.py
+++ b/app/grandchallenge/algorithms/migrations/0001_initial.py
@@ -86,8 +86,8 @@ class Migration(migrations.Migration):
                     "result_template",
                     models.TextField(
                         blank=True,
-                        default="<pre>{{ result_dict|tojson(indent=2) }}</pre>",
-                        help_text="Define the jinja template to render the content of the result.json to html. For example, the following template will print out all the keys and values of the result.json. Use result-dict to accessthe json root.{% for key, value in result_dict.metrics.items() -%}{{ key }}  {{ value }}{% endfor %}",
+                        default="<pre>{{ results|tojson(indent=2) }}</pre>",
+                        help_text="Define the jinja template to render the content of the results.json to html. For example, the following template will print out all the keys and values of the result.json. Use result-dict to accessthe json root.{% for key, value in results.metrics.items() -%}{{ key }}  {{ value }}{% endfor %}",
                     ),
                 ),
                 (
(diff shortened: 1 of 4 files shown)


app/grandchallenge/algorithms/models.py
```diff
@@ -95,13 +95,13 @@ class Algorithm(UUIDModel, TitleSlugDescriptionModel):
     )
     result_template = models.TextField(
         blank=True,
-        default="<pre>{{ result_dict|tojson(indent=2) }}</pre>",
+        default="<pre>{{ results|tojson(indent=2) }}</pre>",
         help_text=(
             "Define the jinja template to render the content of the "
-            "result.json to html. For example, the following template will print "
+            "results.json to html. For example, the following template will print "
             "out all the keys and values of the result.json. Use result-dict to access"
```
miriam-groeneveld, line 102: result-dict should also be updated

app/grandchallenge/algorithms/migrations/0001_initial.py
```diff
@@ -86,8 +86,8 @@ class Migration(migrations.Migration):
                     "result_template",
                     models.TextField(
                         blank=True,
-                        default="<pre>{{ result_dict|tojson(indent=2) }}</pre>",
-                        help_text="Define the jinja template to render the content of the result.json to html. For example, the following template will print out all the keys and values of the result.json. Use result-dict to accessthe json root.{% for key, value in result_dict.metrics.items() -%}{{ key }}  {{ value }}{% endfor %}",
+                        default="<pre>{{ results|tojson(indent=2) }}</pre>",
+                        help_text="Define the jinja template to render the content of the results.json to html. For example, the following template will print out all the keys and values of the result.json. Use result-dict to accessthe json root.{% for key, value in results.metrics.items() -%}{{ key }}  {{ value }}{% endfor %}",
```
miriam-groeneveld, line 90: Replace result-dict and fix my typo please 
```suggestion
                        help_text="Define the jinja template to render the content of the results.json to html. For example, the following template will print out all the keys and values of the result.json. Use result to access the json root.{% for key, value in results.metrics.items() -%}{{ key }}  {{ value }}{% endfor %}",
```

app/grandchallenge/algorithms/models.py
miriam-groeneveld, line 102:
```suggestion
            "out all the keys and values of the result.json. Use result to access "
```

app/grandchallenge/algorithms/migrations/0001_initial.py
jmsmkn: :-) will do, I will make a new commit with the help text changed in both places

PR #1728, opened 2021-02-19
Fix CSV export of annotations

### LINKED ISSUE #1726: Annotation answers get parsed incorrectly in csv export
For annotation type answers, the csv export looks like this currently:
Screenshot from 2021-02-18 16-26-04

It appears the annotation json gets part as part of the export. We should probably add some escaping.

jmsmkn:
This PR flattens any nested structures to strings via JSON serialization, loads in excel out of the box. This has the nice property that the number of columns is now constant.

image

Closes #1726

### DIFF
--- a/app/grandchallenge/core/renderers.py
+++ b/app/grandchallenge/core/renderers.py
@@ -1,3 +1,5 @@
+import json
+
 from rest_framework_csv.renderers import CSVRenderer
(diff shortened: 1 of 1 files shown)


MikeOverkamp-diag (approved):
LGTM

PR #1731, opened 2021-02-24
Mailchimp subscribe

Kicksta:
I have added a mailchimp subscribe option in the footer and a popup that opens once each browser session. Mailchimp has an integration for popups that uses cookies to track whether users have seen the popup somewhere in the year before to not overload people with it. However this requires the whole circus around cookie approval (which mailchimp doesn't handle for you). We don't have anything like that yet in grand-challenge right? For now (wanting to have it live before ECR next week) I fixed it in this manner. For the long term arranging the cookie thing might be better though..?

### DIFF
--- a/app/grandchallenge/products/templates/products/base.html
+++ b/app/grandchallenge/products/templates/products/base.html
@@ -64,18 +64,18 @@
         <div class="container py-5 px-4">
             <div class="row">
                 <div class="col-md-3">
-                    <img src="{% static 'products/images/banner2.png' %}" alt="Logo" style="width:100%">
-                </div>
-
-                <div class="col-md-6 px-4">
                     <div class="h4">About</div>
                     <div>
-                        This is an initiative of the <a href="http://diagnijmegen.nl/" title="DIAG" target="_blank"
+                        AI for Radiology is a non-commercial initiative of the <a href="http://diagnijmegen.nl/" title="DIAG" target="_blank"
                                                         class="text-radboud">Diagnostic Image Analysis Group</a>,
                         part of the Department of Radiology and Nuclear Medicine at the <a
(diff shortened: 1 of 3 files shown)


jmsmkn:
Hi Kicky, maybe I missed the discussion and issue for this first? We do not have anything on the website that requires cookie consent, and I do not want to add it for privacy and usability reasons. Even for this we would need to update the TOS and Privacy Policy. Please can you just link to the signup form hosted on mailchimp instead?

Kicksta:
Hi James, sorry, discussed this verbally with Mike last week. Is it better if I create issues before starting to work on specific features?

I am not sure if I understand where the hitch exactly is in this. Is it the GDPR? Would that be any different if there is a redirection step in between? The alternative you suggest would be a button popping up/in the footer redirecting to this [url ](http://eepurl.com/hrqfdr) right?

And just to be clear about the current implementation: there is no use of cookies, information is only shared with mailchimp when clicked at 'subscribe' after which a confirmatory email is being sent, and there is GDPR statement included about sharing conditions as advised by mailchimp.

jmsmkn:
> Is it better if I create issues before starting to work on specific features?

Yes please! With remote working it's difficult to stay in the loop so we try to do everything through issues.

> Is it the GDPR? 

This is the ePrivacy Directive and GDPR. I think that in this case, if you are 100% sure that the script you include from MailChimp does not set any cookies that could be used for tracking then this is ok. But, you might want to use localStorage rather than sessionStorage as otherwise the popup will re-appear after you close the browser and come back. I'm fairly sure tracking this state is compatible with ePrivacy and GDPR, but again, not a lawyer, it's including any js from mailchimp that I am concerned about.

> Would that be any different if there is a redirection step in between? The alternative you suggest would be a button popping up/in the footer redirecting to this url right?

I would be more comfortable with that.

PR #1732, opened 2021-02-24
Increase the allowed number of image ports in an hanging list

chrisvanrun:
The reader-study configuration will now allow the setting of up to denary (10) image ports in the hanging list.

Capture

Note: when saving and restoring the reader-study configuration, these port->image definitions will be shown in an arbitrary order. For example: `main`, `tertiary`, `secondary`.  As far as I can tell, this cannot be changed since JSON dictionaries are unordered by definition. I would be interested in any suggestions to circumvent this.


app/grandchallenge/reader_studies/models.py
```diff
@@ -123,6 +135,38 @@
...
+        properties[port] = {
+            "$id": f"#/items/properties/{port}",
+            "type": "string",
+            "title": f"The {port[0].upper() + port[1:]} Schema",
+            "default": "",
+            "examples": [f"im_{port}.mhd"],
+            "pattern": "^(.*)$",
+        }
+    return properties
+
```
chrisvanrun, line 167: Not sure what the cleanest way of generating this is. Having them in hardcoded (20 times) seemed a bit much.

app/grandchallenge/reader_studies/models.py
```diff
@@ -123,6 +135,38 @@
...
+def generate_image_port_properties(image_ports):
+    properties = {}
+    image_port_overlays = [f"{port}-overlay" for port in image_ports]
+    for port in image_ports + image_port_overlays:
+        properties[port] = {
+            "$id": f"#/items/properties/{port}",
+            "type": "string",
+            "title": f"The {port[0].upper() + port[1:]} Schema",
+            "default": "",
+            "examples": [f"im_{port}.mhd"],
```
jmsmkn, line 163: Small nitpick - you can use port.title() rather than constructing that yourself.

app/grandchallenge/reader_studies/models.py
jmsmkn: You may have been able to do it in a single generator but it's a good approach.

jmsmkn (approved):

jmsmkn:
Will the changes to the port selection for annotation questions follow in another PR? That would also require a schema migration so might be good to roll both into the same change set.

jmsmkn:
> As far as I can tell, this cannot be changed since JSON dictionaries are unordered by definition. I would be interested in any suggestions to circumvent this

This is correct, a list of objects is probably the way to go but can be addressed when we change the API for labels and multiple overlays etc, not now.

app/grandchallenge/reader_studies/forms.py
```diff
@@ -161,8 +161,8 @@ class Meta(ReaderStudyCreateForm.Meta):
             "hanging_list": (
                 "A list of hangings. "
                 "The hanging defines which image (the hanging value) "
-                "should be assigned to which image port. "
-                'e.g., [{"main":"im1.mhd","secondary":"im2.mhd"}]'
+                "should be assigned to which image port or overlay.  "
+                'e.g., [{"main":"im1.mhd","secondary":"im2.mhd", "secondary-overlay": "im2_overlay.mhd"}]'
```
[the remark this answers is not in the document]
chrisvanrun: Excellent point.

chrisvanrun:
> 
> 
> Will the changes to the port selection for annotation questions follow in another PR? That would also require a schema migration so might be good to roll both into the same change set.

Did not (yet) realize those also will need to be changed. That would have likely have come up when testing the tools on the multiple ports but I will first thing tomorrow!

app/grandchallenge/reader_studies/models.py
chrisvanrun: Sorry @jmsmkn, not sure that I follow. So just to check, do mean an in-line generator instead of a rogue function?

chrisvanrun:
> 
> 
> Will the changes to the port selection for annotation questions follow in another PR? That would also require a schema migration so might be good to roll both into the same change set.

Rolled the port selection for annotation questions into this PR. Hence, I am re-requested your review @jmsmkn.

app/grandchallenge/reader_studies/forms.py
```diff
@@ -161,8 +161,9 @@ class Meta(ReaderStudyCreateForm.Meta):
             "hanging_list": (
                 "A list of hangings. "
                 "The hanging defines which image (the hanging value) "
-                "should be assigned to which image port or overlay.  "
-                'e.g., [{"main":"im1.mhd","secondary":"im2.mhd", "secondary-overlay": "im2_overlay.mhd"}]'
+                "should be assigned to which image port or overlay. Options are: "
+                "main, secondary, tertiary, quaternary, quinary, senary, septenary, octonary, nonary, denary. "
+                'For instance: [{"main":"im1.mhd","main":"im1-overlay.mhd"}]'
```
jmsmkn, line 166: `[{"main":"im1.mhd","main":"im1-overlay.mhd"}]` would not be a valid hanging protocol. Also check you line length, limit is 80 chars (not enforced right now, but will be).

app/grandchallenge/reader_studies/migrations/0008_auto_20210226_0949.py
```diff
@@ -1,17 +1,60 @@
...
 
 from django.db import migrations, models
 
 import grandchallenge.core.validators
 
 
+IMAGE_PORT_EMPTY_PLACEHOLDER = 9
+
+
+def remap_image_ports(apps, _):
```
jmsmkn, line 11: You cannot run a data migration in the same migration as schema migration in postgres. I strongly urge you not to change the field type here, feel free to increase the length of the field.

app/grandchallenge/reader_studies/migrations/0008_auto_20210226_0949.py
```diff
@@ -1,17 +1,60 @@
...
+
+
 class Migration(migrations.Migration):
 
     dependencies = [
         ("reader_studies", "0007_auto_20210201_2220"),
     ]
 
     operations = [
+        migrations.RunPython(remap_image_ports),
```
jmsmkn, line 36: If you do use `RunPython` you are missing `elidable=True` as nothing needs to be done with an empty table.

jmsmkn (changes requested):
It's been caught by the tests but this is not the best approach. Changing a field type is a complex operation and this is not the way to do it, I can show you how at a later date. Please stick with a char field, do not change the representations of main and secondary, increase the maximum length of the char field if needs be.

app/grandchallenge/reader_studies/migrations/0008_auto_20210226_0949.py
chrisvanrun: > migration in postgres.

@MikeOverkamp-diag , fyi.

app/grandchallenge/reader_studies/migrations/0008_auto_20210226_0949.py
chrisvanrun: Going to revert to a minimal migration, but the `elidable=True` is a valuable item I will remember. Thanks!

app/grandchallenge/reader_studies/migrations/0008_auto_20210226_0949.py
MikeOverkamp-diag: I was afraid you would need a separate migration for this, but it somehow did seem to work when you tried it this way.

MikeOverkamp-diag:
Be careful with force pushing to a PR that has already been reviewed, as it makes re-reviewing a bit harder sometimes. I understand that you want a clean commit history (I'm the same), but we squash squash PRs anyway, so the commit history on master will still be clean.

jmsmkn (approved):
LGTM

jmsmkn:
I was too quick to review, sorry, overlooked the draft status, please re-request when ready.

chrisvanrun:
> 
> 
> I was too quick to review, sorry, overlooked the draft status, please re-request when ready.

Not to worry, since I made plenty of minor mistakes in my previous versions I wanted to double-check this time around.

> 
> 
> Be careful with force pushing to a PR that has already been reviewed, as it makes re-reviewing a bit harder sometimes. I understand that you want a clean commit history (I'm the same), but we squash squash PRs anyway, so the commit history on master will still be clean.

Will refrain from doing that in the future!

chrisvanrun:
I took the opportunity to re-do the Question model members to `TextChoice` style. All it took were some replace-all calls, but ANSWER_TYPE is being used all over the place. So the PR got a bit bloated, sorry!

jmsmkn:
> I took the opportunity to re-do the Question model members to TextChoice style. All it took were some replace-all calls, but ANSWER_TYPE is being used all over the place. So the PR got a bit bloated, sorry!

Yes, please separate out these into a separate PR in the future, putting it together makes reviewing a lot more complex.

jmsmkn:
LGTM - made a small change so will wait for the tests.

jmsmkn:
Force pushed to get actions running again.

jmsmkn (approved):

Issue #1110, opened 2020-01-21
Replace drf-yasg with in built drf OpenAPI generator


jmsmkn:
It seems that drf-yasg depends on some deprecated projects that are producing warnings in our build system (coreapi). We should try and move back to drf if it has all the features we need (it did not when we first wanted to generate the schema).

CC @silvandeleemput

silvandeleemput:
There at least is some support for this in django-rest-frameork 3.9: https://www.django-rest-framework.org/community/3.9-announcement/#built-in-openapi-schema-support
Let's see how far I can get.

jmsmkn:
YASG currently blocks upgrading of django rest framework 3.12, we really should remove it as it now supports the common json field.

jmsmkn:
There is a branch for this here: https://github.com/comic/grand-challenge.org/tree/1110_remove_yasg

Issues
- Two retina endpoints use `BaseSerialiser` rather than `Serialiser` so do not provide the `serializer.fields` attribute. Changing them to use `Serialiser` breaks the retina tests
- The schema includes _everything_, we only want the main API
- Cannot automatically authenticate on swagger
- Have not chacked enum fields yet, the tests fail

PR #1735, opened 2021-02-26
Refactor retina serializers

jmsmkn:
This PR fixes a couple of issues needed for the schema generation of #1110. 

We change the use of `BaseSerializer` to `Serializer` with read only fields, and consolidates the thumbnail endpoints so that they can be rendered in any renderer, rather than the specialist `Base64Renderer` which is also removed.

The endpoint for `ImageLevelAnnotationsForImageSerializer` remains unchanged. For `BytesImageSerializer` the consumer will now find the image content in `response.json()["content"]` rather than `response.content`.


HarmvZ (approved):
LGTM!

PR #1736, opened 2021-02-26
Mailchimp subscribe external

Kicksta:
I created a new branch for an alternative implementation of the mailchimp signup. Now it's a button linking to the external page showing as a ribbon below the navigation bar once per browser session + in the footer. Will this work for you, @jmsmkn?


jmsmkn (approved):
LGTM

PR #1737, opened 2021-02-26
Add a valid OpenAPI3.0 schema

### LINKED ISSUE #1110: Replace drf-yasg with in built drf OpenAPI generator

### LINKED ISSUE #1132: The API schema is invalid
Go to https://grand-challenge.org/api/ and at the very far bottom right there is a validate button. Click on it, it returns:

```
{"schemaValidationMessages":[{"level":"error","domain":"validation","keyword":"additionalProperties","message":"object instance has properties which are not allowed by the schema: [\"$schema\",\"anyOf\",\"definitions\"]","schema":{"loadingURI":"http://swagger.io/v2/schema.json#","pointer":"/definitions/schema"},"instance":{"pointer":"/definitions/Answer/properties/answer"}}]}
```

jmsmkn:
This PR replaces drf-yasg with drf-spectacular so that we can upgrade to Python 3.9, and add an OpenAPI3.0 schema. A test is now added to check the validity of the schema.

However, a new issue will need to be opened (#1737) as the schema is incomplete:
* The `CharField(source="get_*_display",...)` should be replaced with custom `ChoiceFields` that use the display values for their representations, so that we get enums for these back in the schema
* The Schemas for JSONFields with validators should be included in the top level schema
* There are warnings where return types have been inferred, these are mostly correct anyway but need to be fixed, see the TODO in `app/tests/api_tests/test_schema.py`

Closes #1110 
Closes #1132


jmsmkn:
I think that many people were assigned as codeowners, but this really is only to do with the API so it's best if @HarmvZ can review this one.

MikeOverkamp-diag (commented):
Wow, this looks so much cleaner than `drf_yasg`.

Only just saw that you specifically requested @HarmvZ  to review, so I'll just leave this as a comment. But LGTM.

jmsmkn:
Merge conflict resolved so this is ready for review.

app/grandchallenge/api/urls.py
```diff
@@ -158,34 +155,22 @@
...
-        r"^swagger(?P<format>\.json|\.yaml)$",
-        schema_view.without_ui(),
-        name="schema-json",
-    ),
+    path("schema/", SchemaView.as_view(), name="schema"),
     # Do not namespace the router.urls without updating the view names in
     # the serializers
     path("v1/", include(router.urls)),
     path("v1/metrics/", MetricsAPIView.as_view(), name="metrics"),
-    path("auth/", include("rest_framework.urls", namespace="rest_framework")),
```
HarmvZ, line 189: Did you intentionally remove these `auth` endpoints? I don't know if it was used anywhere?

HarmvZ (approved):
Nice cleanup indeed! One question:

app/grandchallenge/api/urls.py
jmsmkn: Yes, I'm pretty sure they're only used in the HTML renderer which is only sometimes enabled by developers. We have our own login views that you can use instead.

PR #1740, opened 2021-02-28
Add API token management

jmsmkn:
Adds django rest knox for API token management. This has the nicer feature that a user can create many tokens so that they can logout of different applications independently, and that the keys are not stored in plain text in the database.

We use this to:
* Create a token per workstation session, keeping the token expiry matched with the session expiry and deleting once the session has stopped
* Add user token management so that users no longer need to ask admins for tokens, these are passed back to the user at creation time via the messages framework
  * Note that a token created through the UI are set to never expire, we could change this policy at a later date and add the `AUTO_REFRESH` functionality
  * In the message it might be nice to have a "reveal token" button or direct copy to clipboard in case people do this when screen sharing

https://user-images.githubusercontent.com/12661555/109688949-206fd980-7b85-11eb-9148-f3be426755c1.mp4

**DEPRECATION NOTICE**: The old auth tokens will be deleted and users will need to regenerate their own. The `TOKEN ` keyword is deprecated and replaces with `BEARER ` (OpenAPI Spec), this is handled automatically in gcapi 0.3.5. The user fixtures tokens created in `gcapi` have not changed.


HarmvZ (approved):
LGTM.

HarmvZ (approved):

PR #1744, opened 2021-03-01
Fix markdown editor preview

### LINKED ISSUE #1743: Markdown preview fails CSRF validation checks
Caused by the name change of the CSRF cookie.

jmsmkn:
Due to the name change of the csrf cookie in #1739 this was failing, the name change was necessary as old csrf cookies had been set on subdomains so were out of date. Copied the js sources from https://raw.githubusercontent.com/neutronX/django-markdownx/v3.0.1/markdownx/static/markdownx/js/markdownx.js and included this in the widgets.

The only functional change was https://github.com/comic/grand-challenge.org/commit/14975551ff4cf177e77f2fb3e6c03cf7d44b8035#diff-82dff9ae163bd3e682bcac475a187dc06110ba297fc157bb65c158387825da62R391

If this library does not respect `CSRF_COOKIE_NAME` (see https://github.com/neutronX/django-markdownx/issues/93) this PR and the name change can be reverted to "csrftoken" on 1 March 2022 when all the existing cookies will have expired. 

Closes #1743

### DIFF
--- a/app/grandchallenge/core/static/js/markdownx.js
+++ b/app/grandchallenge/core/static/js/markdownx.js
@@ -1,0 +1,507 @@
+(function e(t, n, r) {
+    function s(o, u) {
+        if (!n[o]) {
+            if (!t[o]) {
+                var a = typeof require == "function" && require;
+                if (!u && a) return a(o, !0);
+                if (i) return i(o, !0);
+                var f = new Error("Cannot find module '" + o + "'");
+                throw f.code = "MODULE_NOT_FOUND", f;
+            }
+            var l = n[o] = {
+                exports: {}
+            };
+            t[o][0].call(l.exports, function(e) {
+                var n = t[o][1][e];
+                return s(n ? n : e);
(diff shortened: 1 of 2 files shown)


HarmvZ (approved):
LGTM. Unfortunate that the plugin does use the setting's value forcing us to clutter the codebase.

jmsmkn:
I guess it's an edge case, I can't imagine many cases where `CSRF_COOKIE_NAME` would need to change, we just hit one of them with having to change `CSRF_COOKIE_DOMAIN` as this was set incorrectly for us from when subdomains were added in 2013. 

It looks like the markdownx project is largely unmaintained right now, it's a shame as apparently it's still quite widely used (https://djangopackages.org/grids/g/wysiwyg/). Perhaps we can fork it, I've had an open PR there for years. Maybe more people are using wagtail and their beautiful editing interface now (https://djangopackages.org/grids/g/cms/).

PR #1745, opened 2021-03-01
Add jinja2 to main dependencies

jmsmkn:
`jinja2` was previously included in the main dependencies via `drf-yasg`, but after removal of that in #1737 this was demoted only to development, which is why the deployments failed on master. We only test the development container as this includes the test libraries that we do not ship to prod.


jmsmkn:
We already know that the tests will pass so force merging this.

PR #1746, opened 2021-03-02
Added contrast enhancement plugin to workstation configuration

### LINKED ISSUE #1729: Add image preprocessing plugin workstation config values
The workstation config needs to be extended to support new values implemented in https://github.com/DIAGNijmegen/rse-cirrus-core/pull/1001.

Fields to be added:
- `show_image_preprocessing_plugin`: `BooleanField` with default value `False`
- `enabled_preprocessors`: `JSONField` containing an object with (for now) a single key `enabled_preprocessors` with a boolean value, e.g. `{ "contrast_enhanced": true }`

amickan:
Users can now enable contrast enhancement preprocessing for fundus images. 

Closes #1729


jmsmkn:
As a new commit for the workstation configuration application has been made on master we now have a problem with multiple leaf nodes as you're also trying to add a new migration in that app, they need to be linear. You will need to:

1. Rebase/merge master into your branch (looks like you've done that)
2. Delete the new migration: the one you made, never delete a migration that is on master!
3. Run `make migrations` - your migration should be regenerated, but this time dependent on the one that is coming from master
4. Commit and push

All the other changes look great!

app/grandchallenge/workstation_configs/serializers.py
```diff
@@ -84,4 +86,12 @@ class Meta:
             "show_flip_tool",
             "show_window_level_tool",
             "show_reset_tool",
+            "enable_contrast_enhancement",
         ]
+
+        def get_enabled_preprocessors(self):
+            if self.enable_contrast_enhancement:
```
jmsmkn, line 93: The signature here is not correct, it should be `def get_enabled_preprocessors(self, obj):` (see https://www.django-rest-framework.org/api-guide/fields/#serializermethodfield)

`self` here refers to the serializer, not the instance of the model, so the lookup you will need is then `obj.enable_contrast_enhancement`

jmsmkn (changes requested):
Actually noticed a bug

app/grandchallenge/workstation_configs/serializers.py
amickan: oh yes, thank you! Fixed it!

app/grandchallenge/workstation_configs/serializers.py
```diff
@@ -84,4 +86,13 @@ class Meta:
             "show_flip_tool",
             "show_window_level_tool",
             "show_reset_tool",
+            "enable_contrast_enhancement",
```
jmsmkn, line 89: One last change - you don't need to include `enable_contrast_enhancement` now that `enabled_preprocessors` is there, this saves duplicate information in the API.

PR #1752, opened 2021-03-02
Added organizations display to profile page, made it optional

### LINKED ISSUE #1751: Show organizations in user profile
Currently, the organizations a member is part of are not visible on their profile pages. It would be nice to add the logos of each of the organizations that a member is part of to their profile with links to the organizations' pages. 

It should look something like on Github: 
image
 
Once this is implemented, make this an optional feature that users can enable or disable in their profile settings.

amickan:
If users are part of organizations, these can now be shown on the user's profile page. 
The user needs to enable this in their profile settings. 

Closes #1751


app/grandchallenge/profiles/models.py
```diff
@@ -44,6 +44,7 @@ class UserProfile(models.Model):
     department = models.CharField(max_length=100)
     country = CountryField()
     website = models.CharField(max_length=150, blank=True)
+    organization_display = models.BooleanField(default=False, blank=True)
```
jmsmkn, line 47: Could we call it "display_organizations" instead? And add some help text: "Display the organizations that you are a member of in your profile". Default should be `True`. Blank should be removed (either display it or do not).

app/grandchallenge/profiles/forms.py
```diff
@@ -75,4 +82,7 @@ def signup(self, request, user):
         user_profile.department = self.cleaned_data["department"]
         user_profile.country = self.cleaned_data["country"]
         user_profile.website = self.cleaned_data["website"]
+        user_profile.organization_display = self.cleaned_data[
+            "organization_display"
+        ]
```
jmsmkn, line 87: We shouldn't show this when the user signs up, only have it as an option when they are editing their profile later. Otherwise, this form gets too messy, and when they are new they do not need to know about organisations.

jmsmkn: Same behavior as `mugshot`, you will also need to pop it from the fields in L34/L38.

app/grandchallenge/profiles/views.py
```diff
@@ -63,6 +64,10 @@ def get_context_data(self, **kwargs):
         profile_user = context["object"].user
         profile_groups = profile_user.groups.all()
 
+        organizations = Organization.objects.filter(
+            members_group__in=profile_groups
```
jmsmkn, line 68: You could also add the admins group here.

app/grandchallenge/profiles/templates/profiles/userprofile_detail.html
```diff
@@ -85,6 +85,20 @@ <h2 class="mb-3">{{ profile.user.username }}</h2>
                                     href="{{ profile.website }}">Website</a></li>
                         {% endif %}
                     </ul>
+                    {% if profile.organization_display %}
+                        {% if organizations %}
```
jmsmkn, line 89: Can be combined into one with an `and`

app/grandchallenge/profiles/templates/profiles/userprofile_detail.html
```diff
@@ -85,6 +85,20 @@ <h2 class="mb-3">{{ profile.user.username }}</h2>
...
+                            <h5>Organizations</h5>
+                            <ul class="list-unstyled">
+                                 {% for organization in organizations %}
+                                   <a href="{{ organization.get_absolute_url }}">
+                                     <img width ="40" height ="40"
+                                                 loading="lazy"
+                                                 src="{{ organization.logo.url }}"
+                                                 alt="{{ organization }} Logo"></a>
```
jmsmkn, line 97: Could we add a `title={{organization.title}}` too so that we get this on hover?

app/grandchallenge/profiles/views.py
amickan: do you mean the editors group? admin group is not an option here, it tells me.

app/grandchallenge/profiles/forms.py
amickan: ok, now that you meantion the sign-up, I get an error message when I try to sign-up `IntegrityError: null value in column "organizations" violates not-null constraint`. I can't figure out how to allow organizations to be empty...

app/grandchallenge/profiles/views.py
jmsmkn: Yes, sorry, inconsistent naming

app/grandchallenge/profiles/forms.py
jmsmkn: I don't really understand, I cannot see where an `organizations` column would be coming from with anything in this changeset.

jmsmkn: Is the default set correctly?

amickan: Yeah, I didn't add an organizations field anywhere and so also didn't set a default, that's why I'm a bit confused... the only place where I define organizations is in views.py `organizations = Organization.objects.filter(members_group__in=profile_groups`

amickan: never mind, it works

jmsmkn (approved):

Issue #1476, opened 2020-08-20
retina_api.tasks.cache_archive_data performs too many queries


jmsmkn:
There is a periodic very high read load on the database, I've tracked this down to the task `retina_api.tasks.cache_archive_data`, which creates a lot of queries, and almost exhausts our provisioned IOPS when running:

https://sentry.io/organizations/grand-challenge/performance/summary/?project=303639&query=&statsPeriod=30d&transaction=grandchallenge.retina_api.tasks.cache_archive_data

Please use `select_related` or `prefetch_related`, in pytest you have access to the assert (max) num queries fixtures https://pytest-django.readthedocs.io/en/latest/helpers.html#django_assert_num_queries that can help with reducing this.

Issue #677, opened 2019-01-16
Clean up (ophthalmology) tests


HarmvZ:
I added a few batch tests that should be converted to pytest parametrized tests. I also added some helper functions that use factories to create a several (connected) models for use in testing. These helper functions should be converted to fixtures.

This is a low priority refactoring issue.

- [ ] Convert batch tests to parametrized tests
- [ ] Convert helper functions to fixtures

PR #1753, opened 2021-03-03
Legacy retina code cleanup

### LINKED ISSUE #1734: Clean up legacy code
The legacy retina workstation has been phased out a while ago. Now there is still quite a bit of unused code remaining in the codebase. This can all be removed.

HarmvZ:
This PR cleans up a lot of unused stuff from the legacy retina workstation. 
- Removes retina_static submodule
- Removes retina specific templates and core views
- Removes unused models, serializers, views, viewsets, mixins, tasks, admins

Closes #1734 
Closes https://github.com/comic/grand-challenge.org/issues/677
Closes https://github.com/comic/grand-challenge.org/issues/1476

It adds a bit of functionality useful for (ophthalmology-specific) development:
- ~Creates a superuser in `init_gc_demo`. I'm not sure if this was intenionally not included before, if so I can remove it again. I use it quite often to inspect objects in the admin so that's why I added it.~
- Creates a token for `retina_import_user` in `init_gc_demo`. I need that to run my custom import scripts after starting the dev environment.
- ~Adds `post_init_retina_demo` command that adds the `retina` user as a user of all existing archives and algorithms. Saves me time to do this manually every time. I could also make a custom .sh script to do this via the django shell if you do not want this cluttering the codebase.~
- Added bash script `/scripts/create_superuser.sh` that creates a verified Django superuser. Pass an argument to change the name/password, default is 'superuser'.
- Added bash script `/scripts/add_user_to_archives_algorithms.sh` that adds a user as user to all existing algorithms and archives. Passed argument is used as username lookup for the user.


jmsmkn:
I would rather not create a superuser in `init_gc_demo`, just in case it somehow gets executed in (someone elses) production. `manage.py createsuperuser` does not work any more right as they need a valid email? Maybe move that to a command in the makefile `make superuser` that will do this via the django shell? If you need a place to put small scripts like this please feel free to make a top level `scripts` directory which won't get included in the container images. Same with `post_init_retina_demo`, would be good to have that outside the distribution, `init_gc_demo` would be harder to move so don't bother with that for this PR.

Super cleanup though - shame it reduces the test coverage! I will review later.

**NOTE** When deploying we need to remember to remove the celery beat task from the database.

app/grandchallenge/core/management/commands/init_gc_demo.py
```diff
@@ -458,18 +470,25 @@ def _create_user_tokens(self):
...
         out = f"{'*' * 80}\n"
         for user, token in user_tokens.items():
             salt = crypto.create_salt_string()
             digest = crypto.hash_token(token, salt)
+            if user == "retina_import_user":
+                user = get_user_model().objects.get(
+                    username=settings.RETINA_IMPORT_USER_NAME
+                )
+            else:
+                user = self.users.get(user)
```
jmsmkn, line 485: I don't understand why this doesn't work for the retina import user?

HarmvZ: The retina import user is created elsewhere: https://github.com/comic/grand-challenge.org/blob/ba63026bc1024eac7233b6a677ba1da780387670/app/grandchallenge/retina_importers/apps.py#L6-L11

app/tests/retina_api_tests/helpers.py
```diff
@@ -82,404 +65,3 @@ def create_datastructures_data(archive_pars=None):
...
-                ds["patient"].name,
-                ds["study"].name,
-                ds["image_cf"].name,
-                "default",
-            ],
-        )
-
-        user, _ = get_retina_user_with_token()
-        client.force_login(user=user)
-        ds["image_cf"].permit_viewing_by_retina_users()
```
jmsmkn, line 162: Is the function `permit_viewing_by_retina_users()` required anymore https://github.com/comic/grand-challenge.org/blob/ba63026bc1024eac7233b6a677ba1da780387670/app/grandchallenge/cases/models.py#L447

HarmvZ:
Ok, I'll move the superuser and post_init stuff into separate scripts.

app/tests/retina_api_tests/helpers.py
jmsmkn: Actually still seems to be used here https://github.com/comic/grand-challenge.org/blob/8ef178ef1b0f135ef3eb85818c84cfc4e6538f53/app/grandchallenge/retina_importers/views.py#L162

app/grandchallenge/core/management/commands/init_gc_demo.py
jmsmkn: Isn't it added to the users dictionary here? https://github.com/comic/grand-challenge.org/pull/1753/files/8ef178ef1b0f135ef3eb85818c84cfc4e6538f53#diff-2aa3920180b8235fb825e66d37bc3a291ec820b8d7111384a42d6d2d81eea9deR100

jmsmkn: Otherwise I'm not sure why R100 is there

HarmvZ: Ah you're right! I'll fix that

jmsmkn:
The rest looks good to me.

jmsmkn (approved):
Nice cleanup

PR #1755, opened 2021-03-04
Fix json serialization

### LINKED ISSUE #1754: CSV export fails on converting uuid to json
Problematic JSON structure in https://grand-challenge.org/api/v1/cases/images/redacted_uuid/?format=csv

```
{"pk":"redacted_uuid","name":"redacted.png","study":null,"files":[{"pk":"redacted_uuid","image":"redacted_uuid","file":"https://grand-challenge.org/media/images/...mhd","image_type":"MHD"},{"pk":"redacted_uuid","image":"09b3b3d6-0994-43d2-b6a9-eaff634b8805","file":"https://grand-challenge.org/media/images/...zraw","image_type":"MHD"}],"reader_study_set":["https://grand-challenge.org/api/v1/reader-studies/redacted_uuid/","https://grand-challenge.org/api/v1/reader-studies/redacted_uuid/"],"archive_set":[],"job_set":[],"width":596,"height":596,"depth":null,"color_space":"RGB","modality":null,"eye_choice":"NA","stereoscopic_choice":null,"field_of_view":null,"shape_without_color":[596,596],"shape":[596,596,3],"voxel_width_mm":null,"voxel_height_mm":null,"voxel_depth_mm":null,"api_url":"https://grand-challenge.org/api/v1/cases/images/redacted_uuid/"}
```

Probably due to trying to serialize the list of files to json in https://github.com/comic/grand-challenge.org/blob/14bc3dd4002756e9cf4a32bb0f238859a9175252/app/grandchallenge/core/renderers.py#L26-L27

jmsmkn:
Use the same arguments to json.dumps as JSONRenderer
https://github.com/encode/django-rest-framework/blob/ec29ff8a8013dd3383344bb78eea479d025e2a87/rest_framework/renderers.py#L100-L104

Closes #1754

### DIFF
--- a/app/grandchallenge/core/renderers.py
+++ b/app/grandchallenge/core/renderers.py
@@ -1,5 +1,7 @@
 import json
 
+from rest_framework.settings import api_settings
+from rest_framework.utils.encoders import JSONEncoder
 from rest_framework_csv.renderers import CSVRenderer
 
 
@@ -24,6 +26,11 @@ class PaginatedCSVRenderer(CSVRenderer):
     @staticmethod
     def _flatten_value(value):
         if isinstance(value, (dict, list)):
-            return json.dumps(value)
+            return json.dumps(
+                value,
+                cls=JSONEncoder,
+                ensure_ascii=not api_settings.UNICODE_JSON,
+                allow_nan=not api_settings.STRICT_JSON,
+            )
         else:
             return value


HarmvZ (commented):
Does it not need the `separators` argument?
I used the something similar in with separators here: 
https://github.com/comic/grand-challenge.org/blob/af191efa3f70fb7894552a3dd06642aff56a81ed/app/tests/retina_api_tests/test_views.py#L213-L221

Anyway, should be fine either way if it works.

jmsmkn:
I left that out as that is only for whitespace changes, it defaults to true but given the CSVs are likely to be looked at by humans I'd figure that we can ignore it and keep the items spaced out which is more or less the same as `LONG_SEPARATORS` (see https://docs.python.org/3/library/json.html)

PR #1758, opened 2021-03-05
Extract image conversion module

jmsmkn:
This is the first step for extracting an image conversion module from grand challenge for use elsewhere. Moves all related files into a new module. Note that there are still internal dependencies on the grand challenge application itself.


jmsmkn:
Nothing but moves so far up until 5a867d7. Now for refactoring.

HarmvZ (approved):

PR #1759, opened 2021-03-05
Remove all django and grand challenge dependencies from panimg

jmsmkn:
This PR allows us to split out the image loaders from grand challenge. There may still be some interdependencies with the tests. The module itself can do with a good clean up, but I suggest we get this out first and double check for any issues. We can keep the integration tests in grand challenge for a while.

A migration is needed as we now call `full_clean` on the created models. The `PanImg` models are tightly coupled to what we have on grand challenge. We should add some internal validation.


miriam-groeneveld (dismissed):
LGTM

jmsmkn:
Awaiting #1760

PR #1760, opened 2021-03-05
Adds pydantic to panimg

jmsmkn:
Adds pydantic to do runtime validation of the panimg models, removes some more dependencies on grand challenge in the tests, and switches to pathlib for the tiff locations. Adds an extra argument to `panimg.convert` to determine where to put the output files, this is controlled by the user who manages the lifecycle. Changes the input to use directories, we then recursively visit each subdirectory and load all the files within, assuming that all the files for one image are located in the same directory - not an issue on grand challenge as the input directory is flat.

This PR is merging into #1759, so review that one first, then this, then merge this into #1759, and then #1759 into master.

### DIFF
--- a/app/grandchallenge/cases/tasks.py
+++ b/app/grandchallenge/cases/tasks.py
@@ -341,7 +341,9 @@ def _handle_raw_image_files(tmp_dir, upload_session):
         for raw_image_file in session_files
     }
 
-    importer_result = import_images(files=input_files, origin=upload_session,)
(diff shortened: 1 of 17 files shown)


jmsmkn:
Miriam - would you be able to run the tests on the full Pathology test data?

miriam-groeneveld:
Will test this morning

miriam-groeneveld (approved):

PR #1766, opened 2021-03-09
Use pypi version of panimg

jmsmkn:
Version 0.1.0 of panimg uses the same code that was extracted here, the changes are only due to now moving from an internal to external dependency.


MikeOverkamp-diag (approved):
LGTM

PR #1769, opened 2021-03-16
Remove unused API attributes and endpoints

jmsmkn:
This is no longer necessary as the image to use is decided internally. I have searched all DIAG code and this is not used.


MikeOverkamp-diag (approved):
LGTM

PR #1770, opened 2021-03-16
Add non-image inputs to algorithms

### LINKED ISSUE #1742: Flexible Algorithm Inputs
See pitch https://github.com/DIAGNijmegen/rse-roadmap/issues/9
Allow for more inputs to an algorithm than just a single image

miriam-groeneveld:
image

Closes #1742


MikeOverkamp-diag:
@jmsmkn I added you as a reviewer already to identify any issues early. That we we can fix them in time for this to be merged tomorrow.

app/grandchallenge/algorithms/forms.py
```diff
@@ -13,18 +19,109 @@
...
+        if user:
+            self._field.widget.user = user
+
+    @property
+    def field(self):
+        return self._field
+
+
+def field_for_interface(i: InterfaceKind.InterfaceKindChoices):
+    if i == InterfaceKind.InterfaceKindChoices.BOOL:
```
jmsmkn, line 93: `InterfaceKind.InterfaceKindChoices.BOOL` is repeated 3 times here, style wise, wouldn't a dictionary make more sense?

MikeOverkamp-diag: I agree. I'll change it.

cycle_docker_compose.sh
```diff
@@ -3,6 +3,9 @@
 echo "Starting docker-compose, press Ctrl+D to cycle docker compose using a down-up cycle"
 echo "Press Ctrl+C (once) to stop"
 sleep 1
+docker build -t algorithm_io app/tests/resources/gc_demo_algorithm/
+docker save algorithm_io -o app/tests/resources/gc_demo_algorithm/algorithm_io.tar
+chmod a+r app/tests/resources/gc_demo_algorithm/algorithm_io.tar
```
jmsmkn, line 8: Could this be moved to the makefile?

`app/tests/resources/gc_demo_algorithm/algorithm_io.tar` needs to be added to gitignore.

The tar file original stored has been squashed out, right?

app/grandchallenge/algorithms/forms.py
```diff
@@ -13,18 +19,109 @@
...
+    "The following file formats can be uploaded and will be converted to "
+    "tif: Aperio(.svs), Hamamatsu(.vms, .vmu, .ndpi), Leica(.scn), MIRAX"
+    "(.mrxs) and Ventana(.bif)."
+)
+
+file_upload_text = (
+    "The total size of all files uploaded in a single session "
+    "cannot exceed 10 GB.<br>"
+    "The following file formats are supported: "
+)
```
jmsmkn, line 51: Would be good if we can remove this duplication somehow as we're always editing this text, but low priority.

app/grandchallenge/algorithms/forms.py
```diff
@@ -13,18 +19,109 @@
...
+    def __init__(
+        self, kind: InterfaceKind.InterfaceKindChoices, initial=None, user=None
+    ):
+        field_type = field_for_interface(kind)
+
+        # bool can't be required
+        kwargs = {
+            "required": (kind != InterfaceKind.InterfaceKindChoices.BOOL)
+        }
+        if initial:
```
jmsmkn, line 64:
```suggestion
        if initial is not None:
```
Thinking aloud in case the initial value is falsey.

app/grandchallenge/algorithms/static/algorithms/js/job_session.js
```diff
@@ -0,0 +1,95 @@
+"use strict";
```
jmsmkn, line 1: Completely unrelated, but I wonder if we should only create js modules from now on? Maybe on for @HarmvZ

app/grandchallenge/components/backends/docker.py
```diff
@@ -252,11 +254,19 @@ def _provision_input_volume(self):
...
+            if isinstance(input_file, tuple):
+                name, input_file = input_file
+                if not hasattr(input_file, "name"):
+                    file = NamedTemporaryFile(delete=True)
+                    file.write(bytes(str(input_file), "utf-8"))
+                    file.flush()
+                    input_file = File(file, name=name)
+            else:
+                name = input_file.name
+            self._create_subdirs(os.path.join("/input", *name.split("/")[:-1]))
```
jmsmkn, line 267:
```suggestion
            subdirs = os.path.join("/input", *name.split("/")[:-1])
            writer.exec_run(f"mkdir -p {subdirs}")
```

Then `_create_subdirs` can be removed, saves orchestrating another container instance for each file

app/grandchallenge/algorithms/tasks.py
```diff
@@ -21,6 +23,100 @@
...
+def execute_algorithm_job_for_inputs(*, job_pk):
+    job = Job.objects.get(pk=job_pk)
+
+    # Send an email to the algorithm editors and creator on job failure
+    linked_task = send_failed_jobs_email.signature(
+        kwargs={"job_pks": [job.pk]}, immutable=True
+    )
+
+    # check if all ComponentInterfaceValue's have a value.
+    # Todo: move this check to execute() code when using inputs is done
```
jmsmkn, line 102: TODO

app/grandchallenge/algorithms/templates/algorithms/algorithm_detail.html
```diff
@@ -66,6 +66,13 @@
                 </a>
             {% endif %}
 
+            {% if request.user.is_staff and "execute_algorithm" in algorithm_perms and object.latest_ready_image %}
```
jmsmkn, line 69: Rather than limiting this to staff, should we have a boolean on the `Algorithm` that switches between the two modes? Default use the legacy, but then we can turn this feature on for specific algorithms. Only make the boolean available in the Django admin.

app/tests/components_tests/factories.py
```diff
@@ -21,4 +21,6 @@ class ComponentInterfaceValueFactory(factory.django.DjangoModelFactory):
     class Meta:
         model = ComponentInterfaceValue
 
+    pk = factory.Sequence(lambda n: n + 9999999)
```
jmsmkn, line 24: Why do we need a lambda here?

app/grandchallenge/algorithms/tasks.py
```diff
@@ -21,6 +23,100 @@
...
+            for civ_pk, upload_pk in upload_pks.items()
+        )
+        start_jobs = chord(image_tasks, start_jobs)
+
+    start_jobs.apply_async()
+    return job
+
+
+@shared_task
+def create_component_interface_value_for_image(
```
jmsmkn, line 72: The name of this function is confusing as the component interface value has already been created, this is building the images and then attaching it to the CIV.

app/grandchallenge/algorithms/tasks.py
```diff
@@ -21,6 +23,100 @@
...
+    *, component_interface_value_pk, upload_pk
+):
+    session = RawImageUploadSession.objects.get(pk=upload_pk)
+    session.status = RawImageUploadSession.REQUEUED
```
jmsmkn, line 76: The session has definitely not been run at this point?

app/grandchallenge/algorithms/tasks.py
```diff
@@ -21,6 +23,100 @@
...
 
 
+@shared_task
+def create_algorithm_job_for_inputs(
+    *, algorithm_image_pk, civ_pks, upload_pks, creator_pk
+):
+    try:
+        algorithm_image = AlgorithmImage.objects.get(pk=algorithm_image_pk)
+    except AlgorithmImage.DoesNotExist:
+        return
```
jmsmkn, line 33: Should this pass silently?

app/grandchallenge/algorithms/views.py
```diff
@@ -342,41 +393,130 @@ def get_success_url(self):
...
+class AlgorithmExperimentCreate(
+    UserFormKwargsMixin,
+    LoginRequiredMixin,
+    ObjectPermissionRequiredMixin,
+    FormView,
+    RemainingJobsMixin,
+):
+    form_class = AlgorithmInputsForm
+    template_name = "algorithms/algorithm_inputs_form.html"
+    permission_required = (
```
jmsmkn, line 406: Have these views been added to the existing permission tests?

app/grandchallenge/algorithms/views.py
```diff
@@ -342,41 +393,130 @@ def get_success_url(self):
...
+        f"{Algorithm._meta.app_label}.execute_{Algorithm._meta.model_name}"
+    )
+    job = None
```
jmsmkn, line 409: Is this needed?

app/grandchallenge/components/apps.py
```diff
@@ -41,6 +41,66 @@ def init_default_interfaces(*_, **__):
...
+        {
+            "title": "Polygon",
+            "kind": ComponentInterface.Kind.POLYGON,
+            "relative_path": "polygon",
+        },
+        {
+            "title": "Multiple polygons",
+            "kind": ComponentInterface.Kind.MULTIPLE_POLYGONS,
+            "relative_path": "multiple_polygons",
+        },
```
jmsmkn, line 103: These will end up in the production database. Do we really want those with these default values? Isn't it better to add these as fixtures in `init_gc_demo` or as factories in the tests?

app/grandchallenge/components/models.py
```diff
@@ -264,6 +305,10 @@ class ComponentInterfaceValue(models.Model):
         to=Image, null=True, blank=True, on_delete=models.CASCADE
     )
 
+    @property
+    def has_value(self):
+        return self.value is not None or self.image or self.file
```
jmsmkn, line 310: Could you add a docstring re: the intention of this propetry? Sometimes a value of `None` _is_ valid.

jmsmkn:
Looks great! Can't wait to try it out.

app/tests/components_tests/factories.py
MikeOverkamp-diag: the civ is expected to have a pk here: https://github.com/comic/grand-challenge.org/blob/master/app/grandchallenge/components/models.py#L227. Apparently the factory hasn't set a pk at that point if it's not explicitly included here. I agree this isn't the nicest solution, but I couldn't think of a better option. I'll think it over again though.

cycle_docker_compose.sh
jmsmkn: Also, do we really need it on every development startup?

app/grandchallenge/algorithms/templates/algorithms/algorithm_detail.html
MikeOverkamp-diag: I like this idea.

app/grandchallenge/algorithms/forms.py
```diff
@@ -39,6 +134,7 @@ class Meta:
             "logo",
             "social_image",
             "public",
+            "inputs",
```
jmsmkn, line 137: Could we use `Select2MultipleWidget` for this?

jmsmkn: Or, maybe hide input selection until it is generally deployed?

app/grandchallenge/algorithms/tasks.py
MikeOverkamp-diag: I assumed so, given the existing `test_no_algorithm_image_does_nothing`. We could also raise an exception of course.

jmsmkn:
Notes with 675fe52:

* The UI for the experiment is a little off (new left, old right), note the alignment and colours when the job is running:
image

app/grandchallenge/components/backends/docker.py
```diff
@@ -252,11 +254,20 @@ def _provision_input_volume(self):
             self._copy_input_files(writer=writer)
 
     def _copy_input_files(self, writer):
-        for file in self._input_files:
+        for input_file in self._input_files:
+            if isinstance(input_file, tuple):
+                name, input_file = input_file
+                if not hasattr(input_file, "name"):
+                    file = NamedTemporaryFile(delete=True)
+                    file.write(bytes(str(input_file), "utf-8"))
```
jmsmkn, line 262: This serializes all of the values to text. I don't think this is what we want, I think we want to serialize as JSON, so that it can be loaded in the algorithm with something like (in L14 `copyio.py`):

```
    files = {x for x in Path("/input").rglob("*") if x.is_file()}

    for file in files:
        try:
            with open(file) as f:
                val = json.loads(f.read())
        except Exception as e:
            warn(f"Could not load {file} as json, {e}")
            continue

        res[str(file.absolute())] = val
```

Otherwise, converting from stringified Python to types is going to be a minefield for the user.

app/grandchallenge/algorithms/templates/algorithms/job_detail.html
```diff
@@ -148,6 +148,21 @@ <h2>Algorithm Result</h2>
...
+                                <li>
+                                    <a class="badge badge-primary"
+                                       href="{{ input.file.url }}" download="{{ input.file.name|name }}">
+                                        <i class="fa fa-download"></i>
+                                        Input {{ forloop.parentloop.counter }} {{ input.file.name|stem }}
+                                        ({{ input.file|suffix }})
+                                    </a>
+                                </li>
+                            {% endif %}
+                            {% if input.value %}
```
jmsmkn, line 161: Falsey values do not show up (eg, boolean is False here)

image

app/grandchallenge/algorithms/forms.py
MikeOverkamp-diag: In that case we only allow it in the admin for now I assume?

app/grandchallenge/algorithms/static/algorithms/js/job_session.js
```diff
@@ -0,0 +1,95 @@
...
+moment.relativeTimeThreshold('ss', 1);
+
+function getJobStatus(jobUrl) {
+    // Checks on the status of the Job (queued, running, started, etc)
+    fetch(jobUrl)
+        .then(response => response.json())
+        .then(job => handleJobStatus(job));
+}
+
+function handleJobStatus(job) {
```
jmsmkn, line 23: If you upload two images to one image interface you end up with this screen forever, no error message:

image

app/grandchallenge/algorithms/forms.py
jmsmkn: Yes, now that you have added the toggle.

app/grandchallenge/algorithms/static/algorithms/js/job_session.js
jmsmkn: The job will also stay in the Queued state forever
image

app/tests/components_tests/factories.py
jmsmkn: Weird, is that a problem in the codebase itself? Would be strange if we added files or values to a None folder.

MikeOverkamp-diag: No I think it has to do with the way the factories are invoked. It errored hard, so we would have seen this happen in production for sure.

app/grandchallenge/algorithms/static/algorithms/js/job_session.js
MikeOverkamp-diag: The backend actually does error: `ValueError: Image imports should result in a single image`. I'm looking into this now.

app/grandchallenge/components/models.py
MikeOverkamp-diag: We currently use this to check if all civs have a value/file/image before actually running the algorithm. In the form all inputs are required, so there should not be any None values I'd say. In what situation would a None value be valid?

app/grandchallenge/algorithms/static/algorithms/js/job_session.js
MikeOverkamp-diag: The job now gets marked as failed when this happens. This also happens in the ui. The image import still gets displayed as if it is still running though. The reason is that we currently use the job to determine if the images have been imported. We simply check if all image civs have an image. In order to properly display the image import status we need to add calls to each of the upload sessions.

app/grandchallenge/algorithms/tasks.py
jmsmkn: I think it should error out here so fine to remove the try except, it would be really strange if an algorithm image was used and then it is gone at this point, we should be aware of that. I will take a look at the existing test once this PR is done.

app/grandchallenge/algorithms/templates/algorithms/algorithm_detail.html
```diff
@@ -60,10 +60,21 @@
...
-                </a>
+                       href="{% url 'algorithms:execution-session-create' slug=object.slug %}">
+                        <i class="fas fa-file-import fa-fw"></i>&nbsp;Try-out Algorithm
+                    </a>
+                {% endif %}
+            {% endif %}
+
+            {% if request.user.is_staff and "execute_algorithm" in algorithm_perms and object.latest_ready_image %}
+
             {% endif %}
```
jmsmkn, line 78:
```suggestion
```

app/grandchallenge/algorithms/templates/algorithms/job_detail.html
jmsmkn: I think that this is still not working

app/grandchallenge/components/models.py
jmsmkn: There was some discussion about it for (multi) choice, but yes, I _think_ that all of the values are non-nullable.

app/grandchallenge/algorithms/templates/algorithms/algorithm_detail.html
MikeOverkamp-diag: Nice catch, I think something went wrong when automerging Miriam's changes that I didn't have locally yet.

app/grandchallenge/algorithms/tasks.py
miriam-groeneveld: No, it will be pending at this point

app/grandchallenge/components/models.py
MikeOverkamp-diag: Wouldn't it make more sense for the value to be `[]` in case of a multiple choice with no selected values?

cycle_docker_compose.sh
MikeOverkamp-diag: That's true, we could just make it a separate Make target that you'd have to specifically run if you want the actual container to be there.

app/grandchallenge/algorithms/views.py
miriam-groeneveld: Yes, otherwise we get warnings "instance attribute job defined outside __init__". The view doesn't have a model, it's jsust a FormView

app/grandchallenge/algorithms/templates/algorithms/algorithm_detail.html
miriam-groeneveld: weird, the is_staff has been removed already?

app/grandchallenge/algorithms/tasks.py
jmsmkn: Do we need to check to confirm that?

app/grandchallenge/algorithms/tasks.py
miriam-groeneveld: Added it here so the old way keeps working, even when another input has been added to the algorithm. So other inputs are only required when using the new form. Do you want me to change that?

app/grandchallenge/algorithms/views.py
```diff
@@ -342,41 +393,130 @@ def get_success_url(self):
...
+            },
+            immutable=True,
+        )
 
-        return {
-            "remaining_jobs": int(total_jobs / max(credits_per_job, 1)),
-            "next_job_at": next_job_at,
-            "user_credits": total_jobs,
-        }
+        self.job = create_job()
```
jmsmkn, line 482: We want to run this in place and not async? Is so, why is create_algorithm_job_for_inputs a shared celery task that needs to make database queries again for all the civs and algorithm image, and not just a method?

app/grandchallenge/components/models.py
jmsmkn: Yes, discussion here https://github.com/DIAGNijmegen/rse-roadmap/issues/9#issuecomment-799579824

app/grandchallenge/algorithms/views.py
miriam-groeneveld: Ah, it was made in place to get the job, so we can use it in the get_success_url. But you are right, might as well make it a method.

jmsmkn: Just check that this is possible, as what that method does might be slow...

jmsmkn: Also, you do not need to use get_success_url, you can always return your own HTTP redirect after calling `super().form_valid` (check the super method to see what that does).

miriam-groeneveld: you mean in stead of calling `super()form_valid` right?

app/grandchallenge/algorithms/static/algorithms/js/job_session.js
MikeOverkamp-diag: Ok, the image imports now also display as errored in case not all images have been properly imported.

miriam-groeneveld:
Ready for review!

jmsmkn:
Usability:

- Fix later: When uploading an image sometimes both boxes are in the waiting state:
image

- Fix later: If you upload 2 images to a widget the error messages are not user friendly. The middle box here is not clickable:
image
  The filed job takes you to the job list page, maybe it should go to the job detail? The error message is then an array with a raw exception, should be tidied up so the user only has a string to read:
image

- Link can be simplified, should be the same text as the normal try out algorithm link
image

- The try out algorithm on the job list page goes to the wrong submission page

jmsmkn:
Sorry, don't work on these yet, clicked the wrong button. If it's all minor things I will fix.

jmsmkn:
Questions:
- Are default values used?

Issues:
- Cannot create CIVs in the admin interface

Minor:
- Show the component interface description in the form?

app/grandchallenge/algorithms/views.py
```diff
@@ -342,41 +393,131 @@ def get_success_url(self):
...
+
+
+class AlgorithmExperiment(
+    LoginRequiredMixin, ObjectPermissionRequiredMixin, TemplateView
+):
+    template_name = "algorithms/experiment.html"
+    permission_required = (
+        f"{Algorithm._meta.app_label}.execute_{Algorithm._meta.model_name}"
+    )
+    raise_exception = True
```
jmsmkn, line 500: The permissions don't seem correct here, shouldn't this be an algorithm job detail view as the job has already been created?

miriam-groeneveld:
> Questions:

> * Are default values used?
Yes, the civ will be created with the default value if one is provided and the user doesn't change it. 
> Issues:
> 
> * Cannot create CIVs in the admin interface
Not sure why you would want to create civ's in the admin. We only need to be able to edit the ComponentInterface right?
> Minor:
> 
> * Show the component interface description in the form?
We could i guess, yes

jmsmkn:
I have:

- Fixed creation of CVs in the admin (c2ee34a)
- Updated the permissions and view locations (2ed0d5d)
- Updated the execution session links (8016893)
- Added the descriptions to the help text (with clean, as this is not auto-escaped!) (3cb9620)

Should the tests pass I think this is ready to 🚢 

The things left to wrap up later are the error messages and the UI.

PR #1771, opened 2021-03-16
Added user profile information to verification admin

### LINKED ISSUE #1761: Add profile information to the verification admin
When manually reviewing verification requests it would be helpful to have more information in the admin such as the users full name, location, department and website.

amickan:
Closes #1761

### DIFF
--- a/app/grandchallenge/verifications/admin.py
+++ b/app/grandchallenge/verifications/admin.py
@@ -31,7 +31,7 @@ class VerificationAdmin(admin.ModelAdmin):
         "is_verified",
         "verified_at",
     )
-    list_select_related = ("user",)
+    list_select_related = ("user__user_profile",)
     list_filter = ("email_is_verified", "is_verified")
     readonly_fields = (
         "created",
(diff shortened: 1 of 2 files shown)


app/grandchallenge/verifications/admin.py
```diff
@@ -23,14 +23,15 @@ def mark_not_verified(modeladmin, request, queryset):
...
+        "user_info",
         "created",
         "signup_email",
-        "signup_email_activated",
         "email",
         "email_is_verified",
         "is_verified",
         "verified_at",
     )
+    list_select_related = ("user",)
```
jmsmkn, line 34: In `user_info` we access the profile a lot, so this should be `user__user_profile`

app/grandchallenge/verifications/models.py
```diff
@@ -69,6 +70,16 @@ def save(self, *args, **kwargs):
 
         super().save(*args, **kwargs)
 
+    def user_info(self):
```
jmsmkn, line 73: This is a property of the model, so it's best to add `@property` to this method

amickan: Ok, why is this preferred? I understand that adding `@property` makes `user_info` a property of the model and so we can then call `verification.user_info` rather than `verification.user_info()`, but I don't reallyunderstand why the former is preferred over the latter. Is it faster, more efficient?

jmsmkn (approved):

PR #1773, opened 2021-03-18
Added extra validation for image type answers

### LINKED ISSUE #1700: Add extra validation for image type answers
We should add two extra checks when adding image type answers:

* Check that the upload session-id exists and is owned by the answer creator in https://github.com/comic/grand-challenge.org/blob/24ea0061be90a9fca3ca093d847e206052e9e7b0/app/grandchallenge/reader_studies/models.py#L1446
* Check that the answer accepts image type answers https://github.com/comic/grand-challenge.org/blob/24ea0061be90a9fca3ca093d847e206052e9e7b0/app/grandchallenge/cases/serializers.py#L165
  * On the `Question` we probably need something like `image_type` like we have for `annotation_type` eg https://github.com/comic/grand-challenge.org/blob/24ea0061be90a9fca3ca093d847e206052e9e7b0/app/grandchallenge/reader_studies/models.py#L1316

These cases should be covered by tests first.

amickan:
- This PR adds a check for whether the question accepts image type answers.
- The check for whether the upload session exists and is owned by the creator of the answer was already implemented. Added a test to cover this check though.

Closes #1700

### DIFF
--- a/app/grandchallenge/cases/serializers.py
+++ b/app/grandchallenge/cases/serializers.py
@@ -168,7 +168,7 @@ class RawImageUploadSessionPatchSerializer(RawImageUploadSessionSerializer):
 
         if not value.question.is_image_type:
             raise ValidationError(
-                "This answer does not accept image type answers."
+                "This question does not accept image type answers."
             )
 
         return value
--- a/app/grandchallenge/cases/serializers.py
+++ b/app/grandchallenge/cases/serializers.py
@@ -163,7 +163,12 @@ class RawImageUploadSessionPatchSerializer(RawImageUploadSessionSerializer):
 
         if not user.has_perm("change_answer", value):
             raise ValidationError(
-                "User does not have permission add an image to this answer"
+                "User does not have permission to add an image to this answer"
+            )
+
+        if not value.question.is_image_type:
+            raise ValidationError(
+                "This answer does not accept image type answers."
             )
 
         return value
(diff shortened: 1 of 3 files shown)


app/grandchallenge/cases/serializers.py
```diff
@@ -163,7 +163,12 @@ def validate_answer(self, value):
 
         if not user.has_perm("change_answer", value):
             raise ValidationError(
-                "User does not have permission add an image to this answer"
+                "User does not have permission to add an image to this answer"
```
MikeOverkamp-diag, line 166: Nice catch

app/grandchallenge/cases/serializers.py
```diff
@@ -163,7 +163,12 @@ def validate_answer(self, value):
...
+            )
+
+        if not value.question.is_image_type:
+            raise ValidationError(
+                "This answer does not accept image type answers."
```
MikeOverkamp-diag, line 171: Maybe change this to `This question does not..` I think that's a bit clearer.

MikeOverkamp-diag (changes requested):
Looks very good. Just one small suggestion.

app/grandchallenge/cases/serializers.py
amickan: ah yes, of course! thanks for catching that

MikeOverkamp-diag (approved):

PR #1775, opened 2021-03-19
Added citation field to publications and period celery task to automatically update pub metadata

### LINKED ISSUE #1774: Add citation field to publication model
Using publications is very slow. The citation of a publication is calulcated every time the `ama_html` is called, which takes time. Better to store the rendered citation text in a new field in the publication model and pass that to the __str__ method.

amickan:
- Publication now have a citation field that contains the cleaned citation. 
- Publication metadata is now also updated once a day.

Closes #1774


app/grandchallenge/publications/tasks.py
```diff
@@ -0,0 +1,15 @@
+from celery import shared_task
+
+from grandchallenge.publications.models import Publication
+from grandchallenge.publications.utils import get_identifier_csl
+
+
+@shared_task
+def update_publication_metadata():
+    for publication in Publication.objects.all():
+        csl, new_identifier = get_identifier_csl(
```
jmsmkn, line 10: `get_identifier_csl` can raise a `ValueError`, and if that happens this loop will stop and the later publications will never be updated. Would you be able to wrap this call in a `try:...except...` block, and if there is a value error add a warning and continue?

amickan: ah yes, by adding a warning, do you mean through `logger.warning('some warning')`?

jmsmkn: Yes that would be the preferred way to do it. Setting up logging is a bit awkward, at the top of the file you need to do something like:

```
import logging

```

Then you can use the logger in the exception handler.

app/grandchallenge/publications/tasks.py
```diff
@@ -0,0 +1,25 @@
...
+def update_publication_metadata():
+    for publication in Publication.objects.all():
+        try:
+            csl, new_identifier = get_identifier_csl(
+                doi_or_arxiv=publication.identifier
+            )
+        except ValueError:
+            logger.warning(
+                f"Identifier {publication.identifier} not recognised"
+            )
```
jmsmkn, line 21:
```suggestion
            logger.warning(
                f"Identifier {publication.identifier} not recognised"
            )
            continue
```
Otherwise the loop will continue, and use the values of csl and identifier from the previous iteration.

jmsmkn (approved):

PR #1780, opened 2021-03-26
Allow auth token to last longer than the session

jmsmkn:
Sometimes it takes a few minutes for the system to kill a session,
which means that the auth token will be invalid whilst the session
is still usable. A minute grace period allows the user to finish
up during which time the container will be stopped by the
"grandchallenge.components.tasks.stop_expired_services" periodic task.

### DIFF
--- a/app/config/settings.py
+++ b/app/config/settings.py
@@ -824,6 +824,8 @@ WORKSTATIONS_RENDERING_SUBDOMAINS = {
     "eu-nl-1",
     "eu-nl-2",
 }
+# Number of minutes grace period before the container is stopped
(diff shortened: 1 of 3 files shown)


MikeOverkamp-diag (changes requested):
I think it would be nicer to also create a setting for the 2. Now it's just a random number that is used multiple times. I think making it a setting will make it more clear what its function is, as well as make it easier to change it without having to change multiple files.

jmsmkn:
I've remove it - it was there to get around a race condition but doesn't actually matter that much.

MikeOverkamp-diag (approved):

PR #1781, opened 2021-03-26
Add an interface list view

jmsmkn:
Updates the algorithm documentation with a current list of the available interfaces and fixes the component interface loading to accept all interface types.

image


miriam-groeneveld (approved):
LGTM!

PR #1783, opened 2021-03-28
Fix slow query

jmsmkn:
The join of the 8 Q objects is really slow, I cannot explain why
right now, but going to the db each time speeds this query up
by 3 orders of magnitude.


MikeOverkamp-diag (approved):
Nice find, I'd be interested to hear why this is faster at some point.

PR #1788, opened 2021-03-30
Implemented topbar

amickan:
- this implements the topbar in base.html and challenge.html
- superordinate `Info`, `Submit`, and `Leaderboard` tabs currently only link to the `detail view` for the first `page` / `phase` respectively, no subordinate tabs/pills yet
- moves `join` and `admin` buttons to the right (they are buttons now, no longer tabs)

### DIFF
--- a/app/grandchallenge/core/context_processors.py
+++ b/app/grandchallenge/core/context_processors.py
@@ -5,6 +5,7 @@ from guardian.shortcuts import get_perms
 from guardian.utils import get_anonymous_user
 
 from grandchallenge.blogs.models import Post
+from grandchallenge.participants.models import RegistrationRequest
 from grandchallenge.policies.models import Policy
 
 logger = logging.getLogger(__name__)
@@ -31,6 +32,11 @@ def challenge(request):
         "challenge_perms": get_perms(user, challenge),
         "user_is_participant": challenge.is_participant(user),
         "pages": challenge.page_set.all(),
+        "num_admins": challenge.get_admins().count(),
+        "num_participants": challenge.get_participants().count(),
+        "num_requests": challenge.registrationrequest_set.filter(
+            status=RegistrationRequest.PENDING
+        ).count(),
     }
 
 
--- a/app/grandchallenge/core/context_processors.py
+++ b/app/grandchallenge/core/context_processors.py
@@ -32,11 +32,9 @@ def challenge(request):
         "challenge_perms": get_perms(user, challenge),
         "user_is_participant": challenge.is_participant(user),
         "pages": challenge.page_set.all(),
-        "num_admins": challenge.get_admins().count(),
-        "num_participants": challenge.get_participants().count(),
-        "num_requests": challenge.registrationrequest_set.filter(
+        "pending_requests": challenge.registrationrequest_set.filter(
             status=RegistrationRequest.PENDING
-        ).count(),
+        ),
     }
 
 
--- a/app/grandchallenge/core/templates/challenge.html
+++ b/app/grandchallenge/core/templates/challenge.html
@@ -39,8 +39,9 @@
             <div class="col-md-10">
                 <div class="nav nav-tabs" role="tablist">
                     <li class="nav-item">
-                        <a class="nav-link {% if request.resolver_match.view_name == 'pages:detail' %} show active{% endif %}"
-                           href="{% url 'pages:detail' challenge_short_name=challenge.short_name page_title=challenge.page_set.first.title %}">
+                    <! -- Todo: Adjust the view_name before merging with master (include the page_title slug)-->
+                        <a class="nav-link {% if request.resolver_match.view_name == 'pages:home' or request.resolver_match.view_name == 'pages:detail' %} show active{% endif %}"
+                           href="{% url 'pages:home' challenge_short_name=challenge.short_name %}">
                             Info
                         </a>
                     </li>
@@ -73,13 +74,13 @@
                             <li class="nav-item">
                                 <a class="nav-link {% if request.resolver_match.view_name == 'evaluation:leaderboard' %}active{% endif %}"
                                    href="{% url 'evaluation:leaderboard' challenge_short_name=challenge.short_name slug=challenge.phase_set.first.slug %}">
-                                    Leaderboard(s)</a>
+                                    Leaderboard{{ challenge.phase_set.all|pluralize }}</a>
                             </li>
                         {% elif not challenge.hidden %}
                             <li class="nav-item">
                                 <a class="nav-link {% if request.resolver_match.view_name == 'evaluation:leaderboard' %}active{% endif %}"
                                    href="{% url 'evaluation:leaderboard' challenge_short_name=challenge.short_name slug=challenge.phase_set.first.slug %}">
-                                    Leaderboard(s)</a>
+                                    Leaderboard{{ challenge.phase_set.all|pluralize }}</a>
                             </li>
                         {% endif %}
                     {% endif %}
--- a/app/grandchallenge/core/templates/challenge.html
+++ b/app/grandchallenge/core/templates/challenge.html
@@ -33,79 +33,78 @@
     {{ block.super }}
 {% endblock %}
 
-{% block sidebar %}
+{% block topbar %}
     {% if challenge %}
-        <div class="col-12 col-md-3 col-lg-2 mb-3">
-            <ul class="nav nav-pills flex-column">
-                {% for page in pages %}
-                    {% if not page.hidden %}
+        <div class="row">
+            <div class="col-md-10">
+                <div class="nav nav-tabs" role="tablist">
+                    <li class="nav-item">
+                        <a class="nav-link {% if request.resolver_match.view_name == 'pages:detail' %} show active{% endif %}"
+                           href="{% url 'pages:detail' challenge_short_name=challenge.short_name page_title=challenge.page_set.first.title %}">
+                            Info
+                        </a>
+                    </li>
+
+                    {% if challenge.display_forum_link %}
                         <li class="nav-item">
-                            <a class="nav-link {% if page == currentpage %}active{% endif %}"
-                               href="{{ page.get_absolute_url }}">
-                                {% filter title %}
-                                    {% firstof page.display_title page.title %}
-                                {% endfilter %}
+                            <a class="nav-link"
+                               href="{% url 'forum:forum' slug=challenge.forum.slug pk=challenge.forum.pk %}">
+                                Forum
                             </a>
                         </li>
                     {% endif %}
-                {% endfor %}
 
-                {% if challenge.display_forum_link %}
-                    <li class="nav-item">
-                        <a class="nav-link"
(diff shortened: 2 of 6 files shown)


app/grandchallenge/core/templates/challenge.html
```diff
@@ -33,79 +33,78 @@
...
-
-                {% if challenge.display_forum_link %}
+        <div class="row">
+            <div class="col-md-10">
+                <div class="nav nav-tabs" role="tablist">
                     <li class="nav-item">
-                        <a class="nav-link"
-                           href="{% url 'forum:forum' slug=challenge.forum.slug pk=challenge.forum.pk %}">
-                            Forum
+                        <a class="nav-link {% if request.resolver_match.view_name == 'pages:detail' %} show active{% endif %}"
```
jmsmkn, line 42: Does this view exist? I think not, maybe add a TODO to fix before it goes on master.

app/grandchallenge/core/templates/challenge.html
```diff
@@ -33,79 +33,78 @@
...
+                                   href="{% url 'evaluation:submission-create' challenge_short_name=challenge.short_name slug=challenge.phase_set.first.slug %}">
+                                    Submit</a>
                             </li>
                             <li class="nav-item">
-                                <a class="nav-link {% if request.resolver_match.view_name == 'evaluation:leaderboard' and request.resolver_match.kwargs.slug == phase.slug %}active{% endif %}"
-                                   href="{% url 'evaluation:leaderboard' challenge_short_name=challenge.short_name slug=phase.slug %}">{{ phase.title }}
-                                    Leaderboard</a>
+                                <a class="nav-link {% if request.resolver_match.view_name == 'evaluation:leaderboard' %}active{% endif %}"
+                                   href="{% url 'evaluation:leaderboard' challenge_short_name=challenge.short_name slug=challenge.phase_set.first.slug %}">
+                                    Leaderboard(s)</a>
```
jmsmkn, line 76: I think the plural can be removed here. If you really want to pluraise it then it is cleaner to use the built in template tag https://docs.djangoproject.com/en/3.1/ref/templates/builtins/#pluralize

app/grandchallenge/core/templates/challenge.html
```diff
@@ -33,79 +33,78 @@
...
-                {% if challenge.display_forum_link %}
+        <div class="row">
+            <div class="col-md-10">
+                <div class="nav nav-tabs" role="tablist">
                     <li class="nav-item">
-                        <a class="nav-link"
-                           href="{% url 'forum:forum' slug=challenge.forum.slug pk=challenge.forum.pk %}">
-                            Forum
+                        <a class="nav-link {% if request.resolver_match.view_name == 'pages:detail' %} show active{% endif %}"
+                           href="{% url 'pages:detail' challenge_short_name=challenge.short_name page_title=challenge.page_set.first.title %}">
```
jmsmkn, line 43: We need to work out which is the first page as some might be hidden or limited view. In fact, it might be better to use `{% url "pages:home" challenge_short_name=challenge.short_name %}` here, which does this.

app/grandchallenge/core/templates/challenge.html
```diff
@@ -33,79 +33,78 @@
...
-                    {% endif %}
-                    {% for phase in challenge.phase_set.all %}
+
                         {% if "change_challenge" in challenge_perms or user_is_participant %}
                             <li class="nav-item">
-                                <a class="nav-link {% if request.resolver_match.view_name == 'evaluation:submission-create' and request.resolver_match.kwargs.slug == phase.slug %}active{% endif %}"
-                                   href="{% url 'evaluation:submission-create' challenge_short_name=challenge.short_name slug=phase.slug %}">Create {{ phase.title }}
-                                    Submission</a>
+                                <a class="nav-link {% if request.resolver_match.view_name == 'evaluation:submission-create' %}active{% endif %}"
+                                   href="{% url 'evaluation:submission-create' challenge_short_name=challenge.short_name slug=challenge.phase_set.first.slug %}">
```
jmsmkn, line 70: Using `challenge.phase_set.first.slug` is good! The phases are always visible, but maybe we need to think about that later.

jmsmkn (approved):
Looks great! I think the only thing to think about is how the pages nav behaves but that can be done later.

app/grandchallenge/core/templates/challenge.html
amickan: Ah, you're right, but just for good practice for me. If I were to pluralize, it would be `Leaderboard{{ challenge.phase_set.all|pluralize }}` - correct?

app/grandchallenge/core/templates/challenge.html
amickan: well, normally the pages detail view would require `page_title` as slug to work, but it seems to work just fine without the slug here (no resolver error or anything), not sure why.

app/grandchallenge/core/templates/challenge.html
jmsmkn: Spot on

amickan: It seems to work and I kinda like that it changes, so I might keep it in there - but of course I can always take it out if you'd rather not have this

app/grandchallenge/core/templates/challenge.html
amickan: Yes, I see that `pages:home` returns the detail view with page set as `self.request.challenge.page_set.first()`. But that's no different from what I'm doing up here in the code, right? The home view does not solve the issue if this first page is hidden, or am I missing something?

app/grandchallenge/core/templates/challenge.html
jmsmkn: Check that it does not make another db query, I think that it does. See this section for how to optimise that https://docs.djangoproject.com/en/3.1/topics/db/optimization/#don-t-overuse-count-and-exists

app/grandchallenge/core/templates/challenge.html
amickan: it does catch the case when there is no page at all, which I guess is good, so in any case, it's better to use `pages:home`

amickan: oh wait, is it the case that because it's querying from the `request` that `self.request.challenge.page_set.first()` will get the first of the set of pages you as the requester / viewer has access to? then, I get how it solves the problem

app/grandchallenge/core/templates/challenge.html
jmsmkn: This was a comment, not a request for change. Phases have a start and closing date which are currently unused, I know it's a feature that we want, but not now.

amickan: gottcha!

app/grandchallenge/evaluation/templates/evaluation/leaderboard_detail.html
```diff
@@ -13,101 +13,114 @@
...
+                <a class="nav-link {% if request.resolver_match.view_name == 'evaluation:leaderboard' and request.resolver_match.kwargs.slug == phase.slug %}active{% endif %}"
+                   href="{% url 'evaluation:leaderboard' challenge_short_name=challenge.short_name slug=phase.slug %}">{{ phase.title }}</a>
+            </li>
+        {% endfor %}
+    </div>
+{% endblock %}
+
 {% block content %}
-    <h2>{{ phase.title }} Leaderboard {{ request.GET.date }}</h2>
+    <div class="mt-3">
```
jmsmkn, line 28: Why has this whole block been wrapped in a new div?

amickan: So that I could add a custom margin to this block to make it align nicely with the nav-pills / nav-tabs. The alternative, as far as I understand, would have been to add the margin to `base.html` but I thought it would be better not to mess with `base.html` since other templates inherit from there. Not sure though.

jmsmkn: Why not change `<div class="nav nav-pills col-12 pl-3 mt-3">` to `<div class="nav nav-pills col-12 pl-3 my-3">`? Where `my-3` is shorthand for `mt-3 mb-3`.

amickan: yes, but that would not affect the content block, right? and it's the content block that was not nicely aligned and that I added the additional `<div class="mt-3">` to

amickan: oh, sorry, I get what you mean now, sorry! will make the change! ;-)

app/grandchallenge/pages/templates/pages/page_detail.html
```diff
@@ -4,18 +4,9 @@
 {% load bleach %}
 {% load static %}
 
-{% block title %}
```
jmsmkn, line 7: Why was this block deleted?

app/grandchallenge/pages/templates/pages/page_detail.html
```diff
@@ -27,46 +18,45 @@
...
+                    </a>
+                </li>
             {% endif %}
-        {% endif %}
-    {% endif %}
+        {% endfor %}
+    </div>
 {% endblock %}
 
-{% block script %}
```
jmsmkn, line 54: Also this block seems to have gone missing?

app/grandchallenge/pages/templates/pages/page_detail.html
amickan: oh, good question! It's because I accidentally deleted everything but content in this file, noticed only later, but then apparently did not put everything back. Fixing this now (also the script block below).

app/grandchallenge/core/context_processors.py
```diff
@@ -31,6 +32,11 @@ def challenge(request):
         "challenge_perms": get_perms(user, challenge),
         "user_is_participant": challenge.is_participant(user),
         "pages": challenge.page_set.all(),
+        "num_admins": challenge.get_admins().count(),
+        "num_participants": challenge.get_participants().count(),
+        "num_requests": challenge.registrationrequest_set.filter(
+            status=RegistrationRequest.PENDING
+        ).count(),
```
jmsmkn, line 39: These are really expensive queries that as it's in a context processor will be performed on every request. To me, this information it is not worth 3 extra db queries on every challenge page. If needed, this information should be either:

- stored in fields on the challenge objects (big change, needs tracking)
- only evaluated in the templates so that only admins get the 3 extra queries (slow, but doesn't effect everyone else)

amickan: Ok, makes sense. I can move these queries to the challenge,html template so that they only get evaluated for admins, but it's not necessary to have these badges at all, so we can totally leave them out. The only one that is maybe a useful addition from the admin perspective is the count of open requests, then admins only need to open the request approval page if there are pending approvals, but this is only relevant for challenges where participants are not automatically accepted, so again probably just a tiny gain in terms of user friendliness.

jmsmkn (approved):

PR #1792, opened 2021-04-01
Moved challenge submission list view to tab

amickan:
Users can now access a list of their submissions to a challenge through a seperate tab next to the phase-specific submission tabs. 

image

### DIFF
--- a/app/grandchallenge/evaluation/templates/evaluation/submission_form.html
+++ b/app/grandchallenge/evaluation/templates/evaluation/submission_form.html
@@ -25,10 +25,15 @@
         {% if "change_challenge" in challenge_perms or user_is_participant %}
             {% for phase in challenge.phase_set.all %}
                 <li class="nav-item">
-                    <a class="nav-link {% if request.resolver_match.view_name == 'evaluation:submission-create' and request.resolver_match.kwargs.slug == phase.slug %}active{% endif %}"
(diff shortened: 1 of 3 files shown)


app/grandchallenge/evaluation/templates/evaluation/submission_form.html
```diff
@@ -25,10 +25,15 @@
         {% if "change_challenge" in challenge_perms or user_is_participant %}
             {% for phase in challenge.phase_set.all %}
                 <li class="nav-item">
-                    <a class="nav-link {% if request.resolver_match.view_name == 'evaluation:submission-create' and request.resolver_match.kwargs.slug == phase.slug %}active{% endif %}"
+                    <a class="nav-link {% if request.resolver_match.view_name == 'evaluation:submission-create' and request.resolver_match.kwargs.slug == phase.slug or request.resolver_match.view_name == 'evaluation:submission-create-legacy' and request.resolver_match.kwargs.slug == phase.slug%}active{% endif %}"
```
jmsmkn, line 28: Is is right to include `evaluation:submission-create-legacy` here?

jmsmkn (approved):

app/grandchallenge/evaluation/templates/evaluation/submission_form.html
amickan: I included it so that the respective phase tab would be highlighted when creating a submission for someone else. Looked nicer to me, but it's not strictly necessary, of course. 
image
If we don't include it, it looks like this:
image

jmsmkn: I agree, this is great! Just wasn't sure where it was being linked from.

jmsmkn:
Feel free to squash this PR onto the feature branch when you're ready.

PR #1798, opened 2021-04-07
Nav-pill and nav-tab layout improvements

amickan:
- In the challenge navigation, nav-tabs no longer have a top/left/right border, only a bottom border, which highlights the currently active tab (in dark blue). Hovering over other tabs makes their bottom border visible in grey.

- Nav-pills are now thinner and rounded, making them visibly different from other nav-pills on the website and from the buttons used on the page etc. The active pill has a dark blue background, hovering over other pills highlights them in light grey. 

- All text is now black and no longer blue. 

- The bottom border of the nav-bars now extends also to underneath the join/admin buttons.

- The Join button is now green and the Admin button is dark grey. 

- Top nav-bar and nav-pills for phases now collapse into a dropdown on small screens. The info pages nav-pills display horizontally (rather than vertically) on small screens now (i.e., mobile devices). 

https://user-images.githubusercontent.com/30069334/114397663-34a0f080-9b9f-11eb-8d52-a36517f9428b.mp4

I also tried changing the background color of the nav-tab-bar (topbar) to a light grey, but I'm not sure how much I like it. 
 
image

The above changes also affect the nav-bars and nav-pills in the admin window (not the ones in readerstudies and algorithms and so on though). Do we want this, or do we want the nav-tabs to display as they used to for those pages? 

Before:

image

Now: 

image

### DIFF
--- a/app/grandchallenge/evaluation/templates/evaluation/leaderboard_detail.html
+++ b/app/grandchallenge/evaluation/templates/evaluation/leaderboard_detail.html
@@ -14,14 +14,33 @@
 {% endblock %}
 
 {% block sidebar %}
-    <div class="nav nav-pills col-12 pl-3 mb-3">
+    <div class="nav nav-pills d-none {% if challenge.phase_set.count > 4 %}d-lg-inline-flex{% elif challenge.phase_set.count == 4 %}d-md-inline-flex{% else %}d-sm-inline-flex{% endif %} col-12 pl-3 mb-3">
         {% for phase in challenge.phase_set.all %}
             <li class="nav-item">
-                <a class="nav-link {% if request.resolver_match.view_name == 'evaluation:leaderboard' and request.resolver_match.kwargs.slug == phase.slug %}active{% endif %}"
+                <a class="nav-link mr-1 px-4 py-1 {% if request.resolver_match.view_name == 'evaluation:leaderboard' and request.resolver_match.kwargs.slug == phase.slug %}active{% endif %}"
                    href="{% url 'evaluation:leaderboard' challenge_short_name=challenge.short_name slug=phase.slug %}">{{ phase.title }}</a>
             </li>
         {% endfor %}
     </div>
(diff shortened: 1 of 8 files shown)


miriam-groeneveld (approved):
LGTM!

app/grandchallenge/evaluation/templates/evaluation/leaderboard_detail.html
```diff
@@ -14,14 +14,33 @@
...
+    <div class="nav nav-pills d-none {% if challenge.phase_set.count > 4 %}d-lg-inline-flex{% elif challenge.phase_set.count == 4 %}d-md-inline-flex{% else %}d-sm-inline-flex{% endif %} col-12 pl-3 mb-3">
         {% for phase in challenge.phase_set.all %}
             <li class="nav-item">
-                <a class="nav-link {% if request.resolver_match.view_name == 'evaluation:leaderboard' and request.resolver_match.kwargs.slug == phase.slug %}active{% endif %}"
+                <a class="nav-link mr-1 px-4 py-1 {% if request.resolver_match.view_name == 'evaluation:leaderboard' and request.resolver_match.kwargs.slug == phase.slug %}active{% endif %}"
                    href="{% url 'evaluation:leaderboard' challenge_short_name=challenge.short_name slug=phase.slug %}">{{ phase.title }}</a>
             </li>
         {% endfor %}
     </div>
+    <div class="nav nav-pills d-block {% if challenge.phase_set.count > 4 %}d-lg-none{% elif challenge.phase_set.count == 4 %}d-md-none{% else %}d-sm-none{% endif %} col-12 pl-3 mb-3">
```
jmsmkn, line 25: Have you checked for duplicate queries using the debug toolbar? You might be going to the database quite a few times here with a lot of instances of `phase_set.all` and `phase_set.count`. Check, and if so you can probably reduce the number of calls by using the challenge context processor.

amickan: As far as I can tell, there are no duplicate database hits. Same number of queries with and without the if statement in there. I also thought that the count method does not actually evaluate the queryset and hence would not hit the database at all? Looking through the list of queries with the debug toolbar, I also can't find this query anywhere. I hope I'm not just missing it.

jmsmkn (approved):

PR #1802, opened 2021-04-09
Allow window-preset definitions with percentiles

chrisvanrun:
Allows the definition (by admins) of window presets that contain lower and upper percentile values.
percentile_window_lut

The viewer (CIRRUS), should then generate and apply a window LUT that covers all intensities that have a frequency between the lower and upper percentiles.

Implementation status on CIRRUS: https://github.com/DIAGNijmegen/rse-cirrus-core/issues/1095
Part of pitch: https://github.com/DIAGNijmegen/rse-roadmap/issues/30


app/grandchallenge/workstation_configs/models.py
```diff
@@ -264,15 +264,45 @@ def save(self, *args, **kwargs):
 
 class WindowPreset(TitleSlugDescriptionModel):
     width = models.PositiveIntegerField(
-        validators=[MinValueValidator(limit_value=1)]
+        blank=True, null=True, validators=[MinValueValidator(limit_value=1)]
+    )
+    center = models.IntegerField(blank=True, null=True)
+
+    lower_percentile = models.PositiveSmallIntegerField(
+        blank=True, null=True, validators=[MaxValueValidator(limit_value=100)]
```
jmsmkn, line 272: Could you add a `MinValueValidator` to both the lower and upper percentiles, and test these?

app/grandchallenge/workstation_configs/models.py
```diff
@@ -264,15 +264,45 @@ def save(self, *args, **kwargs):
...
+                        width__isnull=False,
+                        lower_percentile__isnull=True,
+                        upper_percentile__isnull=True,
+                    )
+                    | models.Q(
+                        center__isnull=True,
+                        width__isnull=True,
+                        lower_percentile__isnull=False,
+                        upper_percentile__isnull=False,
+                    )
```
jmsmkn, line 296: Could you add a test for this constraint?

app/grandchallenge/workstation_configs/models.py
```diff
@@ -264,15 +264,45 @@ def save(self, *args, **kwargs):
...
-        validators=[MinValueValidator(limit_value=1)]
+        blank=True, null=True, validators=[MinValueValidator(limit_value=1)]
+    )
+    center = models.IntegerField(blank=True, null=True)
+
+    lower_percentile = models.PositiveSmallIntegerField(
+        blank=True, null=True, validators=[MaxValueValidator(limit_value=100)]
+    )
+
+    upper_percentile = models.PositiveSmallIntegerField(
```
jmsmkn, line 275: Could you add a check and test that `upper_percentile > lower_percentile`?

jmsmkn (commented):
Looks good, just could do with a bit more validation and tests. Would be nice if these things didn't just raise a database integrity error and you got some nice feedback in the forms, but as this is admin only it doesn't matter right now.

app/grandchallenge/workstation_configs/models.py
jmsmkn: Thinking about it this is not necessary as it is implicitly included with the `PositiveSmallIntegerField`.

app/grandchallenge/workstation_configs/models.py
chrisvanrun: Slightly outside the scope, but added form validation also.

jmsmkn: It's not "out of scope". When you're taking input from other sources it needs to be validated.

app/grandchallenge/workstation_configs/models.py
```diff
@@ -264,15 +264,81 @@ def save(self, *args, **kwargs):
...
+        if self.upper_percentile < self.lower_percentile:
+            raise ValidationError(
+                f"Upper percentile ({self.upper_percentile}%) should be below the "
+                f"lower percentile ({self.lower_percentile}%)"
+            )
+
+    def _validate_fixed(self):
+        pass
+
+    def clean(self):
```
jmsmkn, line 289: This is missing a call to `super().clean()`.

app/grandchallenge/workstation_configs/models.py
```diff
@@ -264,15 +264,81 @@ def save(self, *args, **kwargs):
...
+                        width__isnull=True,
+                        lower_percentile__isnull=False,
+                        upper_percentile__isnull=False,
+                    )
+                ),
+            ),
+            models.CheckConstraint(
+                name="%(app_label)s_%(class)s_upper_gte_lower_percentile",
+                check=models.Q(
+                    upper_percentile__gte=models.F("lower_percentile")
```
jmsmkn, line 332: `_gte` is not what we need, otherwise you definitely end up with a zero width window.

app/grandchallenge/workstation_configs/models.py
chrisvanrun: I'll add it, but what is the reasoning behind it? Currently, there is no super to call upon. Is this to make it future proof for when something is added upstream of the inheritance?

jmsmkn: > Currently, there is no super to call upon.

There is.

app/grandchallenge/workstation_configs/models.py
chrisvanrun: I missed that the current `width` has a minimum of `1`! On the CIRRUS side, restrictions needed to be a lot more relaxed in order to account for the `inverse` option so I was in a permissive mindset. Thank you for the spot!

app/grandchallenge/workstation_configs/models.py
chrisvanrun: Apart from the `pass` base class one?

jmsmkn: How do you know that it will always return pass? How about in the next major revision? Doesn't matter what it returns, it is there, call it.

chrisvanrun: So what I surmise it's about the future proving of call. Would it then also make sense to pass the cleaned data? As in:
```
def clean(self):
  cleaned_data = super.clean()
  [...]
  return cleaned_data
```

jmsmkn: No, models do not have a concept of cleaned data (forms do, which is where you would do this). The requirement of this method is only to raise a `ValidationError` if the set of values are not valid which is then handled by `full_clean`.

jmsmkn (approved):

PR #1808, opened 2021-04-19
Removed duplicate code in challenge layout

amickan:
Nav-pill to dropdown logic now in js script.

### DIFF
--- a/app/grandchallenge/challenges/static/js/challenges/dropdown.js
+++ b/app/grandchallenge/challenges/static/js/challenges/dropdown.js
@@ -10,12 +10,14 @@ function handle_nav_tab_dropdown() {
 
 function navs_to_dropdown(reverse = false) {
     // Get all elements that need to change class
-    const containers = document.getElementsByClassName("nav-tab-dropdown-container");
+    // const tabContainers = document.getElementsByClassName("nav-tab-dropdown-container");
+    const containers = document.querySelectorAll('.nav-tab-dropdown-container,.nav-pill-dropdown-container');
 
     const ulDropdownClasses = ["dropdown-menu", "dropdown-menu-left"];
     const ulTabClasses = ["nav", "nav-tabs", "border-0"];
     const itemDropdownClasses = ["dropdown-item", "challengeDropdown"];
-    const itemTabClasses = ["nav-item"];
+    const itemTabPillClasses = ["nav-item"];
+    const ulPillClasses = ["nav", "nav-pills", "col-12", "pl-3", "mb-3"];
 
     // Change the classes
     for (let container of containers) {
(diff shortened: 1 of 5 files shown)


jmsmkn (approved):

app/grandchallenge/challenges/static/js/challenges/dropdown.js
```diff
@@ -1,51 +1,79 @@
...
+    if (window.innerWidth < 576) {
+        // 576 px is the bootstrap breakpoint for sm
+        navs_to_dropdown(false, true);
+    } else if (window.innerWidth < 992) {
         // 992 px is the bootstrap breakpoint for lg
         navs_to_dropdown();
     } else {
         navs_to_dropdown(true);
-    }
+        }
```
HarmvZ, line 11: Fix whitespace
```suggestion
    }
```

app/grandchallenge/challenges/static/js/challenges/dropdown.js
```diff
@@ -1,51 +1,79 @@
...
-            ul.classList.add(...ulTabClasses);
             ul.classList.remove(...ulDropdownClasses);
+
+            if (container.classList.contains("nav-tab-dropdown-container")) {
+                ul.classList.add(...ulTabClasses);
+            } else if (container.classList.contains("nav-pill-dropdown-container")) {
+                ul.classList.add(...ulPillClasses);
+            }
         }
         else {
```
HarmvZ, line 53:
```suggestion
        } else {
```

app/grandchallenge/challenges/static/js/challenges/dropdown.js
```diff
@@ -1,51 +1,79 @@
...
+                item.classList.add(...itemTabPillClasses);
                 item.classList.remove(...itemDropdownClasses);
             } else {
                 item.classList.add(...itemDropdownClasses);
-                item.classList.remove(...itemTabClasses);
+                item.classList.remove(...itemTabPillClasses);
             }
         }
     }
+
```
HarmvZ, line 76:
```suggestion
```

app/grandchallenge/challenges/static/js/challenges/dropdown.js
```diff
@@ -1,51 +1,79 @@
...
     const ulTabClasses = ["nav", "nav-tabs", "border-0"];
     const itemDropdownClasses = ["dropdown-item", "challengeDropdown"];
-    const itemTabClasses = ["nav-item"];
+    const itemTabPillClasses = ["nav-item"];
+    const ulPillClasses = ["nav", "nav-pills", "col-12", "mb-3"];
+    const ulHorizPagePillClasses = ["d-flex", "align-items-stretch", "mb-3"];
+    const ulVertPagePillClasses = ["flex-column"];
+
+    for (let page_container of page_containers) {
+        let ul_pages  = page_container.querySelector("ul");
```
HarmvZ, line 28: It's a bit counterintuitive, but both of these variables can be `const` instead of `let` since they are both not reassigned inside the loop. They are changed every iteration, but when using const there you redefine (!=reassign) them. Using `const` wherever possible is preferred because this provides some protection by explicitly not allowing the variables to be reassigned.

HarmvZ (changes requested):
Functionally this is good!

One code styling thing: some of the added variables are in `snake_case`. As opposed to Python, in javascript it's convention to use `camelCase` for both variables and functions. Could you change this? And then while you're at it you could also change the function names to `camelCase`.

As a general comment: This is all page styling/layout changes based on window width, I _think_ it should be possible to do it all in CSS with media queries and custom classes. If it is possible in a clean way it is also preferred since that is where this type of styling/responsiveness belongs. However, I can not see an easy way of doing this so the current method is also good!

PR #1809, opened 2021-04-19
Apply celery tasks only on commit

jmsmkn:
Only apply celery tasks on commit, adds `capture_on_commit_callbacks` backport to execute the tasks in tests rather than using transaction test cases. Removes unused return types. See https://adamj.eu/tech/2020/05/20/the-fast-way-to-test-django-transaction-on-commit-callbacks/ and https://adamj.eu/tech/2020/02/03/common-celery-issues-on-django-projects/.

### DIFF
--- a/app/grandchallenge/algorithms/admin.py
+++ b/app/grandchallenge/algorithms/admin.py
@@ -1,4 +1,5 @@
 from django.contrib import admin
+from django.db.transaction import on_commit
 from guardian.admin import GuardedModelAdmin
 
 from grandchallenge.algorithms.models import (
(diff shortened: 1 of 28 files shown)


jmsmkn:
Ok, I think we will need to refactor to not use the transactions, as this messes up with parallel tests (see 4d19bd42d650cf53819350d78f280205b869f01b which was later removed) along with being slow.

jmsmkn:
I'm happy with this approach so am going to convert all apps.

MikeOverkamp-diag (approved):
LGTM

PR #1810, opened 2021-04-20
List all output interfaces on job detail page

### LINKED ISSUE #1720: Named overlays as algorithm job outputs
For an algorithm with multiple image outputs, the outputs are shown as overlays in the workstation but they are named as "Overlay n", "Overlay n+1", etc. see this algorithm for example: https://grand-challenge.org/algorithms/segmentation-of-advanced-amd-biomarkers-in-oct/. It would be nice if the overlays are named so that the OverlayPlugin can display those names in the dropdown box.
According to [the processor documentation](https://comic.github.io/grand-challenge.org/processor.html#processor-outputs), the images should be mentioned in the result.json. We could extract the image name and store it in the `description` field of the `ComponentInterfaceValue`. The overlayplugin can then use these values as labels for the overlays.

MikeOverkamp-diag:
Closes #1720

### DIFF
--- a/app/grandchallenge/algorithms/templates/algorithms/job_detail.html
+++ b/app/grandchallenge/algorithms/templates/algorithms/job_detail.html
@@ -172,25 +172,22 @@
                     <dd class="col-sm-9">
                         <ul class="list-unstyled mb-0">
                             {% for output in object.outputs.all %}
-                                {% for file in output.image.files.all %}
-                                    <li>
-                                        <a class="badge badge-primary"
-                                           href="{{ file.file.url }}">
-                                            <i class="fa fa-download"></i>
-                                            Output {{ forloop.parentloop.counter }} {{ file.image.name|stem }}
-                                            ({{ file.file|suffix }})
-                                        </a>
-                                    </li>
-                                {% endfor %}
                                 {% if output.file %}
-                                    <li>
-                                        <a class="badge badge-primary"
-                                           href="{{ output.file.url }}" download="{{ output.file.name|name }}">
-                                            <i class="fa fa-download"></i>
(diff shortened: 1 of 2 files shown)


app/grandchallenge/algorithms/templates/algorithms/job_detail.html
```diff
@@ -172,25 +172,22 @@ <h2>Algorithm Result</h2>
...
-                                            ({{ output.file|suffix }})
-                                        </a>
-                                    </li>
+                                  <li>
+                                    <a class="badge badge-primary"
+                                       href="{{ output.file.url }}">
+                                        <i class="fa fa-download"></i> {{ output.interface.title }}
+                                    </a>
+                                </li>
+                                {% elif output.image %}
```
jmsmkn, line 182: I think we're missing the last case of `output.value`

MikeOverkamp-diag: I wasn't sure if we wanted that in here. Should it always be fully visible here? Or should it be truncated and fully shown in a popup in case the value is very long?

MikeOverkamp-diag: Come to think of it, it might be nice to always show it in a popup, so we can use the same type of links we use for files and images.

jmsmkn: It's up to you, was only thinking of consistency with the inputs. Actually this is probably fine, we need to really think how to do this given the result templating is only for results json. Maybe just truncate the values for now, revisit it later.

jmsmkn (approved):

PR #1812, opened 2021-04-21
Add horizontal navigation bars to challenges

### LINKED ISSUE #1786: Make challenge page layout more user-friendly
See pitch DIAGNijmegen/rse-roadmap#24

Make the challenge page layout more user friendly

### LINKED ISSUE #1789: Increase width of algorithm result table
The table on the algorithm results page can become wider than the page container if the name of the scan is very long. The user then has to scroll to the right to see the "Open Result in Viewer" button, which is quite confusing.

image

amickan:
We redesigned the layout of challenge pages. Previously, navigation was possible through a sidebar that contained links to all sub-pages (info pages, submission forms, etc.). With multiple phases, this list could get very long and confusing. We now implemented a horizontal navigation bar which groups related pages together: all info pages can be accessed through the Info tab, the submission forms and leaderboards for all phases can be accessed through the Submit and Leaderboard tabs respectively. You can then navigate to the phase-specific submission form / leaderboard through the subtabs. The subtabs for the phases are also displayed horizontally now, leaving more space for the leaderboard tables. Finally, the Join and Admin buttons are now separated from the rest of the navigation and hence easier to spot. We hope this design will make navigation through the challenge page more intuitive for users. Let us know what you think! 

Screenshot to demonstrate will be added here soon

Part of pitch: https://github.com/DIAGNijmegen/rse-roadmap/issues/24

Closes #1786 
Closes #1789

### DIFF
--- a/app/grandchallenge/core/templates/challenge.html
+++ b/app/grandchallenge/core/templates/challenge.html
@@ -35,11 +35,11 @@
 
 {% block topbar %}
     {% if challenge %}
-        <div class="row mb-3">
-            <div class="col-md-10 col-sm-8 col-6">
-                <div class="nav nav-tabs" role="tablist">
+        <div class="row mb-3 mx-0 border-bottom">
+            <div class="d-none d-lg-block col-md-10 col-sm-8 col-6 px-0">
+                <div class="nav nav-tabs border-0" role="tablist">
                     <li class="nav-item">
-                    <! -- Todo: Adjust the view_name before merging with master (include the page_title slug)-->
+                        <! -- Todo: Adjust the view_name before merging with master (include the page_title slug)-->
                         <a class="nav-link {% if request.resolver_match.view_name == 'pages:home' or request.resolver_match.view_name == 'pages:detail' %} show active{% endif %}"
                            href="{% url 'pages:home' challenge_short_name=challenge.short_name %}">
                             Info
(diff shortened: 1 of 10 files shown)


app/grandchallenge/pages/templates/pages/page_detail.html
```diff
@@ -27,46 +25,65 @@
...
             src="{% static "js/render_geocharts.js" %}"></script>
-
     {# make the tables sortable #}
     <script type="text/javascript"
             src="{% static "js/sort_tables.js" %}"></script>
     {# render embeddable google forums #}
     <script type="text/javascript"
             src="{% static "js/embed_forums.js" %}"></script>
-{% endblock %}
+{% endblock %}
```
HarmvZ, line 89: Missing newline
```suggestion
{% endblock %}

```

app/grandchallenge/core/templates/challenge.html
```diff
@@ -33,111 +34,125 @@
...
+                       role="button"
+                       aria-expanded="false"><i class="fas fa-bars"></i></a>
+                    <ul class="nav nav-tabs border-0">
                         <li class="nav-item">
-                            <a class="nav-link {% if page == currentpage %}active{% endif %}"
-                               href="{{ page.get_absolute_url }}">
-                                {% filter title %}
-                                    {% firstof page.display_title page.title %}
-                                {% endfilter %}
+                            <! -- Todo: Adjust the view_name before merging with master (include the page_title slug)-->
```
HarmvZ, line 49: Is this left in here intentionally?

HarmvZ (commented):
LGTM! This is only a partial review of the JS, I will leave the rest to be properly reviewed by James.

app/grandchallenge/core/templates/challenge.html
amickan: Thanks for catching this! I'm actually not fully sure what the correct view_name would be. It works as it is now, even though the detail view for pages normally requires the page slug to work. Not sure why it works without that, and also not sure how I would say something along the lines of "any page slug" because this is the superordinate Info tab which should remain active no matter which subpage the user is on.

jmsmkn (approved):

PR #1817, opened 2021-04-23
Use .html for result list page, as contents are properly escaped

MikeOverkamp-diag:


app/grandchallenge/algorithms/templates/algorithms/job_list.html
```diff
@@ -70,7 +70,10 @@ <h5 class="modal-title"></h5>
             let modal = $(this);
             const json_element = document.getElementById(button.data('pk'));
             if (json_element) {
-                modal.find('.modal-job-output').text(json_element.textContent);
+                {% comment %}
+                    json_element is created with the json_script filter and should be safe to set with the .html function
```
jmsmkn, line 74:
```suggestion
                    json_element is created with the json_script filter using data from job.rendered_result_text,
                    which has been passed through bleach and so safe to use with .html here.
```

jmsmkn (approved):

PR #1819, opened 2021-04-23
Move challenge admin tab into nav-bar

amickan:
- Moved admin button into top nav bar
- Refactored `UserIsChallengeAdminMixin`, now using only `ObjectPermissionRequiredMixin`
- Changed redirect success_url of challenge update view


app/grandchallenge/pages/urls.py
```diff
@@ -15,6 +16,7 @@
     path("pages/", PageList.as_view(), name="list"),
     path("pages/create/", PageCreate.as_view(), name="create"),
     path("", ChallengeHome.as_view(), name="home"),
+    path("/admin", ChallengeSettingsBase.as_view(), name="admin"),
```
jmsmkn, line 19: Paths should not start with a `/`, but end with a `/`
```suggestion
    path("admin/", ChallengeSettingsBase.as_view(), name="admin"),
```

app/grandchallenge/pages/views.py
```diff
@@ -78,8 +98,21 @@ def get_object(self, queryset=None):
...
+class ChallengeSettingsBase(
+    LoginRequiredMixin, ObjectPermissionRequiredMixin, TemplateView
+):
+    template_name = "pages/challenge_settings_base.html"
+    permission_required = "change_challenge"
+    raise_exception = True
+    login_url = reverse_lazy("account_login")
+
+    def get_permission_object(self):
+        return self.request.challenge
```
jmsmkn, line 110: Do we need this view and URL? It seems to only show the settings sidebar. 

If we do need it, then I think we need a test for the permissions at least.

app/tests/common_tests/test_link_visibility.py
```diff
@@ -11,20 +11,11 @@
 )
 def test_admins_see_links(view, client, two_challenge_sets):
     url = reverse(
-        "pages:home",
+        "pages:admin",
```
jmsmkn, line 14: Hmm, is it only to support this test? Then maybe what we need to do is re-write the tests to test each of the challenge admin views for permissions. I can show you some good examples that we have.

app/grandchallenge/pages/views.py
amickan: Ah yes, you're right, we do not need this view anymore. I will remove it and also adjust the tests accordingly

jmsmkn (approved):

jmsmkn:
Great consolidation, thanks!

PR #1822, opened 2021-04-26
Stay on current menu item in reader studies after page refresh

### LINKED ISSUE #1578: Stay on current menu item in reader studies
Currently when you are on a particular menu item in a reader study, you return to the overview when refreshing the page, or after performing an action that results in a refresh. It would be nice to stay on your current menu item instead.

The easiest way to achieve this would be to add hashes for the menu items to the urls.

amickan:
When refreshing a readerstudy page, the selected tab/menu item remains active and is displayed. The user no longer returns to the overview (i.e., the first tab) after refreshing.

Closes #1578


MikeOverkamp-diag (approved):

jmsmkn:
This looks great! Could you do the same for Archives and Algorithms?

PR #1829, opened 2021-05-04
Add probability to annotation types

jmsmkn:
This allows the user to add probabilities to their annotation objects. Values are between zero and one.

Also fixes some issues with the validation:
- Allows name to be optional on polygon annotation
- Rejects additional items in objects

https://github.com/DIAGNijmegen/rse-cirrus-core/issues/1193


app/grandchallenge/components/models.py
```diff
@@ -268,6 +268,7 @@ def interface_type_annotation():
                     ],
                     "sub_type": "brush",
                     "groups": []
+                }
```
MikeOverkamp-diag, line 271: Nitpicky, but should we add extra indentation for the objects in the list? Having the `}` and `]` with the same indentation level threw me off a little.

MikeOverkamp-diag (dismissed):
LGTM, one small comment, but I'd be fine with merging this as is.

app/grandchallenge/components/models.py
jmsmkn: Easy fix, it did take me a while to find what was wrong with the JSON.

miriam-groeneveld:
I do wonder if this change will break stuff with the currently deployed Cirrus. At the moment, the json that is being sent by Cirrus will still contain property "type" on the sub types, which is not allowed anymore.

jmsmkn:
> I do wonder if this change will break stuff with the currently deployed Cirrus. At the moment, the json that is being sent by Cirrus will still contain property "type" on the sub types, which is not allowed anymore.

This is why we code review :-)

Added back, would be good to remove the superfluous info from CIRRUS.

miriam-groeneveld:
> > I do wonder if this change will break stuff with the currently deployed Cirrus. At the moment, the json that is being sent by Cirrus will still contain property "type" on the sub types, which is not allowed anymore.
> 
> This is why we code review :-)
> 
> Added back, would be good to remove the superfluous info from CIRRUS.

They are removed with the bug fix for showing annotations, but that's not deployed yet. And I'm not 100% sure if all superfluous properties have been removed.

PR #1833, opened 2021-05-10
Fix issues from changing to archive items

MikeOverkamp-diag:


jmsmkn (approved):
Looks good, we need to re-write the test for the archive.items m2m signal which is currently marked as xfail.

jmsmkn (approved):
LGTM

PR #1836, opened 2021-05-11
Add ArchiveItem model

jmsmkn:
At the moment Archives can only contain a set of images. This is rather limiting as usually you want to add metadata or annotations about those images to the archive. Additionally, if a user wants to link an algorithm to an archive it only supports one interface (generic medical image), and so algorithms that use other interfaces will not work.

This PR:
- Adds an `ArchiveItem` model, that is foreign keyed to an `Archive` and m2m related to `ComponentInterfaceValues`
- Adds migration commands to migrate existing images to archive items containing generic medical images
- Updates the API to use these new relations

Algorithms and adding images still work, adding metadata to an archive item will follow in another PR.

https://github.com/DIAGNijmegen/rse-roadmap/issues/45


app/grandchallenge/algorithms/tasks.py
```diff
@@ -288,6 +303,30 @@ def create_algorithm_jobs(
...
+            archive_items = archive_items[
+                : remaining_jobs(
+                    creator=creator, algorithm_image=algorithm_image
+                )
+            ]
+
+    for archive_item in archive_items:
+        job_inputs = archive_item.values.filter(interface__in=input_interfaces)
+
+        # TODO: check that this filter works
```
jmsmkn, line 315: @MikeOverkamp-diag Could you check this?

jmsmkn:
I cannot review the PR as I opened it but I'm happy with this. My only thoughts are on re-use of CIVs as discussed before but lets kick that can down the road.

MikeOverkamp-diag (dismissed):

PR #1841, opened 2021-05-17
Notifications model

amickan:
- Users can now delete notifications and mark them as read/unread. This can be done for individual notifications as well as for multiple notifications at the same time. There is a `select all` checkbox to easily select all notifications in the list. 

- From the notification list view, users can navigate to their subscription overview, which lists all forums and topics they are subscribed to and offers the option to unsubscribe from each of them. Topics are ordered by forum. 

- Manually subscribing to forums and topics is now also possible through the respective forum / topic pages. 

- Users can opt in or out of notification emails through their profile settings. If they opt in (currently the default), they receive emails about new (max. 1 day old), unread notifications. 

Demo video

https://user-images.githubusercontent.com/30069334/121366910-2589b500-c93a-11eb-9760-9b6717e3bf92.mp4

To-do:
- after subscribing to a forum/topic, the user is redirected to their subscription overview, rather than the respective forum/topic page - this might be annoying, fix redirect?

Possible nice-to-haves for notification list view:
- filtering by forum or topic 
- filtering to only see unread notifications


app/grandchallenge/notifications/utils.py
```diff
@@ -0,0 +1,141 @@
...
+def _queryset_gfk_content_generator(queryset, gfks):
+    for model in queryset:
+        for field_name, field in gfks.items():
+            (
+                content_type,
+                object_pk,
+            ) = _get_related_content_type_and_related_object_pk(
+                model, field.model, field
+            )
+            yield (model, field_name, content_type, object_pk)
```
jmsmkn, line 88: Maybe use a named tuple here and unpack by name, not item. https://docs.quantifiedcode.com/python-anti-patterns/readability/not_using_named_tuples_when_returning_more_than_one_value.html

app/grandchallenge/notifications/tasks.py
```diff
@@ -0,0 +1,33 @@
...
+        ):
+            recipients[user] = len(user.unread_notifications)
+            user.notification_email_last_sent_at = now()
+            user.save()
+        elif user.notification_email_last_sent_at is not None:
+            unread_notifications = Notification.objects.filter(
+                Q(user=user.user)
+                & Q(read=False)
+                & Q(action__timestamp__gt=user.notification_email_last_sent_at)
+            )
```
jmsmkn, line 27:
```suggestion
            unread_notifications = Notification.objects.filter(
                user=user.user,
                read=False,
                action__timestamp__gt=user.notification_email_last_sent_at
            )

```

app/grandchallenge/notifications/tasks.py
```diff
@@ -0,0 +1,33 @@
...
+from django.utils.timezone import now
+
+from grandchallenge.notifications.emails import send_unread_notifications_email
+from grandchallenge.notifications.models import Notification
+from grandchallenge.profiles.models import UserProfile
+
+
+@shared_task
+def send_unread_notification_emails():
+    users = UserProfile.objects.filter(receive_notification_emails=True)
```
jmsmkn, line 12: Rename `users` to `profiles`.

Paginate the objects here.

app/grandchallenge/notifications/tasks.py
```diff
@@ -0,0 +1,33 @@
...
+    for user in users:
+        if (
+            user.notification_email_last_sent_at is None
+            and user.has_unread_notifications
+        ):
+            recipients[user] = len(user.unread_notifications)
+            user.notification_email_last_sent_at = now()
+            user.save()
+        elif user.notification_email_last_sent_at is not None:
+            unread_notifications = Notification.objects.filter(
```
jmsmkn, line 23: Add `count()` to unread_notifications and replace the len below.

app/grandchallenge/notifications/tasks.py
```diff
@@ -0,0 +1,33 @@
...
+@shared_task
+def send_unread_notification_emails():
+    users = UserProfile.objects.filter(receive_notification_emails=True)
+    recipients = {}
+    for user in users:
+        if (
+            user.notification_email_last_sent_at is None
+            and user.has_unread_notifications
+        ):
+            recipients[user] = len(user.unread_notifications)
```
jmsmkn, line 19: User `count()` rather than `len`.

app/grandchallenge/notifications/emails.py
```diff
@@ -0,0 +1,21 @@
...
+
+def send_unread_notifications_email(recipients):
+    subject = "You have unread notifications"
+
+    for user, n_notifications in recipients.items():
+        msg = (
+            f"Dear {user.user.username},\n\n"
+            f"You have {n_notifications} new notification{pluralize(n_notifications)}."
+            "To read and manage your notifications, visit: https://grand-challenge.org/notifications/.\n\n"
+            "If you no longer wish to receive emails about notifications, you can disable them in your profile settings."
```
jmsmkn, line 14: Say this is an automated message and provide link to unsubscribe (profile edit view).

app/grandchallenge/notifications/emails.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def send_unread_notifications_email(recipients):
+    subject = "You have unread notifications"
+
+    for user, n_notifications in recipients.items():
+        msg = (
+            f"Dear {user.user.username},\n\n"
+            f"You have {n_notifications} new notification{pluralize(n_notifications)}."
+            "To read and manage your notifications, visit: https://grand-challenge.org/notifications/.\n\n"
```
jmsmkn, line 13: Hardcoded grand-challenge.org, get this from the `Site` model.

jmsmkn: `Site.objects.get_current().domain.lower()`

app/grandchallenge/notifications/admin.py
```diff
@@ -18,3 +20,4 @@ class FollowAdmin(ModelAdmin):
 
 admin.site.unregister(Follow)
 admin.site.register(Follow, FollowAdmin)
+admin.site.register(Notification)
```
jmsmkn, line 23: Custom model with `user` being read only.

app/grandchallenge/forum_conversation/templatetags/forum_extras.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+@register.simple_tag
+def get_follow_object_pk(user, follow_object):
+    if isinstance(follow_object, Topic):
+        object_follows_for_user = Follow.objects.filter(
+            Q(user=user)
+            & Q(
+                content_type=ContentType.objects.get(
+                    app_label="forum_conversation", model="topic"
```
jmsmkn, line 18: `app_label` can be found in `Model._meta.app_label`, same for `model_name`?

app/grandchallenge/notifications/views.py
```diff
@@ -1,27 +1,180 @@
...
         )
+        # TODO side-effect, not nice
+        profile.update_notifications_last_read_timestep()
+        return context
 
-        UserProfile.objects.filter(pk=profile.pk).update(
-            notifications_last_read_at=now()
-        )
+    def post(self, request, *args, **kwargs):
+        if "delete" in request.POST:
```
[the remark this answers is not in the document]
jmsmkn: Postponed for moving to DRF.

app/grandchallenge/notifications/views.py
```diff
@@ -1,27 +1,113 @@
...
+            action = "delete"
+        elif "mark_read" in request.POST:
+            action = "mark_read"
+        elif "mark_unread" in request.POST:
+            action = "mark_unread"
+
+        selected_notifications = request.POST.getlist("checkbox")
+        notifications = get_objects_for_user(
+            request.user,
+            ["delete_notification", "change_notification"],
```
jmsmkn, line 46: Split these permissions out so that you're only looking for one, eg, where you're setting `action = "mark_read"` also set `required_permission = "change_notification"`, and then only look for the required permission. This will help when refactoring.

app/grandchallenge/notifications/views.py
```diff
@@ -1,27 +1,113 @@
...
+
+    def get_success_url(self):
+        return reverse("notifications:follow-list")
+
+
+class FollowCreate(
+    LoginRequiredMixin, SuccessMessageMixin, CreateView,
+):
+    model = Follow
+    fields = ["user", "content_type", "object_id", "actor_only"]
```
jmsmkn, line 109: I think that this view might be too permissive. I think that this means that any user (including anonymous users) can create a follow for any other user, as there is no validation that the `user` provided is the current user from the request. We should also check that the user has permission to follow that object. Some things might be private and information could be leaked if this is not checked carefully, it might need separate methods as this is different for Forums, Topics, and any other objects we add in the future (eg, evaluations).

app/grandchallenge/notifications/templates/actstream/follow_form.html
```diff
@@ -0,0 +1,19 @@
...
+        <li class="breadcrumb-item active" aria-current="page">Subscriptions</li>
+    </ol>
+{% endblock %}
+
+{% block content %}
+    <form method="post">
+        {% csrf_token %}
+        {{ form.as_p }}
+        <input type="submit" value="Update">
+    </form>
```
jmsmkn, line 18: We should use django crispy forms, but this is linked to my comments about the FollowCreate view.

jmsmkn (commented):
I have another meeting now so will continue the review later, looks great! Only problem I see is with the `FollowCreate` view which is too permissive. I think we need a form that takes the user and the topic or forum and checks that they have permission to first create a follow for the user, and that the user has permission to view the topic or forum.

app/grandchallenge/notifications/signals.py
```diff
@@ -45,3 +53,27 @@ def create_post_action(sender, *, instance, created, **_):
...
+
+
+@receiver(post_save, sender=Action)
+def create_notification(*, instance, **_):
+    if instance.target:
+        follower_group = followers(instance.target)
+        for follower in follower_group:
+            # only send notifications to followers other than the poster
+            if follower != instance.actor:
+                Notification(user=follower, action=instance).save()
```
jmsmkn, line 65: Just FYI - `Notification.objects.create(...)` is preferred if you're doing this in one step.

app/grandchallenge/notifications/tasks.py
```diff
@@ -0,0 +1,40 @@
...
+        current_time = now()
+        recipients = {}
+        for profile in current_page_profiles:
+            if (
+                profile.notification_email_last_sent_at is None
+                and profile.has_unread_notifications
+            ):
+                recipients[profile] = profile.unread_notifications.count()
+                profile.notification_email_last_sent_at = current_time
+                profile.save()
```
jmsmkn, line 28: Rather than calling `profile.save()` each time, at the end of the for loop you can do `Notification.objects.bulk_update(current_page_profiles, ["notification_email_last_sent_at"])` and that gets it done in 1 database call.

app/grandchallenge/notifications/tasks.py
```diff
@@ -0,0 +1,40 @@
...
+from grandchallenge.notifications.emails import send_unread_notifications_email
+from grandchallenge.notifications.models import Notification
+from grandchallenge.profiles.models import UserProfile
+
+
+@shared_task
+def send_unread_notification_emails():
+    profiles = UserProfile.objects.filter(
+        receive_notification_emails=True
+    ).order_by("user")
```
jmsmkn, line 14: Maybe we can add a filter here for users with unread notifications to save the lookups below.

amickan: I think that is not possible because `has_unread_notifications` is a property and not stored in the database, right?

jmsmkn: You should be able to add something like `"user__notification_set__read=False"` to the filter, then you might also need to add a `.distinct()` before the `order_by`.

app/grandchallenge/forums/templates/forum/forum_detail.html
```diff
@@ -0,0 +1,98 @@
...
+            {% csrf_token %}
+            <input name="content_type" type="hidden" value="{{ forum|get_content_type }}">
+            <input name="actor_only" type="hidden" value="False">
+            <input name="object_id" type="hidden" value="{{ forum.id }}">
+            <input name="user" type="hidden" value="{{ request.user.id }}">
+            <button type="submit" class="btn btn-info btn-sm btn-subscription"><i class="fas fa-check"></i>
+                {% trans "Subscribe" %}
+            </button>
+        </form>
+    {% endif %}
```
jmsmkn, line 59: General comment - do not construct forms yourself in templates, Django Forms are designed to do this and are much easier to change, re-use and test.

app/grandchallenge/forums/templates/forum_conversation/partials/topic_detail_actions.html
```diff
@@ -16,13 +14,26 @@
...
+    <form style="display: inline;" method="post" action="{% url 'notifications:follow-create' %}">
+        {% csrf_token %}
+        <input name="content_type" type="hidden" value="{{ topic|get_content_type }}">
+        <input name="actor_only" type="hidden" value="False">
+        <input name="object_id" type="hidden" value="{{ topic.id }}">
+        <input name="user" type="hidden" value="{{ request.user.id }}">
+        <button type="submit" class="btn btn-info btn-sm btn-subscription"><i class="fas fa-check"></i>
+            {% trans "Subscribe" %}
+        </button>
+    </form>
```
jmsmkn, line 35: Replace with Django forms

jmsmkn (approved):
LGTM 🚀

PR #1845, opened 2021-05-18
Fix silent fail when missing `USER` directive in Dockerfile of processor

### LINKED ISSUE #1844: Not setting `USER` directive in processor Dockerfile fails silently
Building a processor image without setting `USER` directive in Dockerfile will result in the container never being marked as ready and triggers a (silent) error in sentry: 
https://sentry.io/organizations/grand-challenge/issues/2396054397/?project=303639&query=is%3Aunresolved

It should fail properly returning a validation error to the user indicating that the `USER` directive should be set in the Dockerfile according to docker best practices.

HarmvZ:
Fixes #1844

### DIFF
--- a/app/grandchallenge/components/tasks.py
+++ b/app/grandchallenge/components/tasks.py
@@ -66,7 +66,9 @@ def _validate_docker_image_manifest(*, model, instance) -> str:
     )
     config = json.loads(config)
 
-    if str(config["config"]["User"].lower()) in ["", "root", "0"]:
+    if "User" not in config["config"] or str(
+        config["config"]["User"].lower()
+    ) in ["", "root", "0"]:
         model.objects.filter(pk=instance.pk).update(
             status=(
                 "The container runs as root. Please add a user, group and "


jmsmkn (approved):

Borda:
hello,
 I started receiving recently the following email as an organizer:

> Unfortunately we were unable to validate your docker image at https://anhir.grand-challenge.org/evaluation/methods/... The error message was:
> The container runs as root. Please add a user, group and USER instruction to your Dockerfile, rebuild, test and upload the container again, see https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user

This seems something new which was not required back when we created the challenge so it would be handy to read more about the validation process to be able to debug it locally if needed.

jmsmkn:
This has always been a requirement. The link to the docker documentation sent in that email (https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user) tells you exactly what to do. If you use `evalutils` to generate your container images then this is all done for you (https://pypi.org/project/evalutils/).

PR #1846, opened 2021-05-18
Add logo and description fields for cards

jmsmkn:
Ensures that Algorithms, Archives and Reader Studies have a "title", "description", "logo", and "url" fields, and that the list views are filterable by slug. Allows anonymous users to list the Archives and Reader Studies, but the objects are still filtered with the object permissions.

Whilst here I cleaned up some inconsistencies in the REST API, and removed some hacks around our permissions model and Django Rest Framework. `DjangoObjectPermissions` can now be used everywhere now that the correct global permissions are set for adding or changing objects too. Where those are used, the additional checks for posting new related objects are doing in validation (reader studies), and patching is covered by `ObjectPermissionsFilter` (sessions, reader studies, answers). Some status codes have consequently changed.

I also opened up the UserAutoComplete for all users, I think this is ok now that all users will get permission to add algorithms so should be able to manage the groups. We should add in some anti-abuse tools here. Also we allow the anonymous user to get their profile rather than add special handling in the permissions, I don't think that there is a real downside to this.

https://github.com/DIAGNijmegen/website-content/issues/363


HarmvZ:
>Whilst here I cleaned up some inconsistencies in the REST API, and removed some hacks around our permissions model and Django Rest Framework. DjangoObjectPermissions can now be used everywhere now that the correct global permissions are set for adding or changing objects too. Where those are used, the additional checks for posting new related objects are doing in validation (reader studies), and patching is covered by ObjectPermissionsFilter (sessions, reader studies, answers). Some status codes have consequently changed.

I'm not sure if I understand exactly what is going on. Model-level add and change permissions are now given to all registered users so that the workaround can be removed and only object-level permissions are responsible for checking if a specific user has access to a specific object. Do I understand it correctly?
Does this then also mean that for all new models where we want to use object-level permissions we need to set the model-level add and change permissions? Could we refactor the post_migrate signal into something general we can reuse for this?

app/tests/groups_tests/test_views.py
```diff
@@ -52,9 +50,6 @@ def get_user_autocomplete():
                 data={"q": u.username.lower()},
             )
 
-        response = get_user_autocomplete()
-        assert response.status_code == 403
-
```
HarmvZ, line 57: Does this now return HTTP 200 and an empty list? If so, it would be better to check for that explicitly I think in case something is changed later and it accidentally returns non empty.

app/tests/workstations_tests/test_api.py
```diff
@@ -98,6 +98,25 @@ def test_session_update_extends_timeout(client):
     assert s.maximum_duration != timedelta(minutes=10)
 
 
+@pytest.mark.django_db
+def test_session_only_patchable_by_creator(client):
```
HarmvZ, line 102:
```suggestion
def test_session_not_patchable_by_other_user(client):
```

jmsmkn:
It's not just for code removal, this is for working with the contracts that DRF and Django provide. You can see in here that developers had no idea which permission classes to use, or why, or what the differences were. This also makes the responses from the API completely inconsistent, returning in some cases 400, 401, 403, or 405, and it really depended on the endpoint.

It comes from this: https://github.com/encode/django-rest-framework/blob/3875d3284e73ed4d8e36c07d9b70c1b22c9d5998/rest_framework/permissions.py#L254-L262

In DRF non-safe operations require the model permissions, so if you want user to be able to make the non-safe operations you need to assign the model permission to the user. This makes sense, right - you cannot have an add permission for an object, it must be on the model level. For put/patch/delete the model permissions are required to authorize the unsafe operation, and the object permissions are required to modify the specific object. This is the contract from DRF.

Safe operations are never needed the model level permissions, and still don't. You only need to add them if the methods are unsafe and you are using model or object permissions on the model and viewset. 

I think the functions are general enough, the permissions are then managed in each app and they're always importing specific models, I don't think that belongs centrally.

miriam-groeneveld (approved):
LGTM

HarmvZ (commented):
Good new feature and nice cleanup!

>I also opened up the UserAutoComplete for all users, I think this is ok now that all users will get permission to add algorithms so should be able to manage the groups. We should add in some anti-abuse tools here. Also we allow the anonymous user to get their profile rather than add special handling in the permissions, I don't think that there is a real downside to this.

Indeed anti-abuse tools would be good. I think it's quite easy now create an account and start scraping userdata (no email) But I guess this is possible on a lot of sites and not really useful. Possibly we can change _adding_ of users/editors into _inviting_ users, requiring confirmation of the user. This will catch possible permission logic errors which we discussed a bit when working on the archive-algorithm relation.

PR #1847, opened 2021-05-18
Add job create API endpoint

miriam-groeneveld:


app/grandchallenge/algorithms/serializers.py
```diff
@@ -7,19 +7,22 @@
...
+    ComponentInterfaceSerializer,
+    ComponentInterfaceValuePostSerializer,
     ComponentInterfaceValueSerializer,
     HyperlinkedComponentInterfaceValueSerializer,
 )
 
 
 class AlgorithmSerializer(serializers.ModelSerializer):
     average_duration = SerializerMethodField()
+    inputs = ComponentInterfaceSerializer(many=True)
```
jmsmkn, line 25: It would be useful to have the outputs here too, make sure both are prefetched in the views.

app/grandchallenge/algorithms/serializers.py
```diff
@@ -90,3 +94,85 @@ class HyperlinkedJobSerializer(JobSerializer):
 
     class Meta(JobSerializer.Meta):
         pass
+
+
+class JobPostSerializer(JobSerializer):
+    algorithm_title = CharField(write_only=True)
```
jmsmkn, line 100: You shouldn't select algorithms by title as these are not unique, use slug instead, and I think that there is a SlugField which also does the validation. We also need to ensure that the permissions are checked correctly.

app/grandchallenge/algorithms/serializers.py
```diff
@@ -90,3 +94,85 @@ class HyperlinkedJobSerializer(JobSerializer):
...
+    inputs = ComponentInterfaceValuePostSerializer(many=True)
+
+    class Meta:
+        model = Job
+        fields = ["pk", "algorithm_title", "inputs", "status"]
```
jmsmkn, line 105: Are status and pk read only?

app/grandchallenge/algorithms/serializers.py
```diff
@@ -90,3 +94,85 @@ class HyperlinkedJobSerializer(JobSerializer):
...
+
+    def validate(self, data):
+        alg = Algorithm.objects.get(title=data.pop("algorithm_title"))
```
jmsmkn, line 108: Thinking about it, it might be better to have this functionality as a POST action on the algorithm itself? That's how we structure it in the Django template views. Unsure.

app/grandchallenge/components/serializers.py
```diff
@@ -1,57 +1,152 @@
...
             "pk",
+            "upload_pk",
         ]
 
+    def validate(self, attrs):  # noqa: C901
+        interface = ComponentInterface.objects.get(
+            title=attrs["interface_title"]
+        )
+        attrs.pop("interface_title")
+        attrs["interface_id"] = interface.id
```
jmsmkn, line 63: `_pk` and `.pk` is preferred, just in case you somehow end up dealing with a non-ORM object (it happens...)

app/grandchallenge/algorithms/serializers.py
```diff
@@ -90,3 +94,85 @@ class HyperlinkedJobSerializer(JobSerializer):
...
+        upload_pks = {}
+        algorithm_input_ids = set(
+            a.id for a in job.algorithm_image.algorithm.inputs.all()
+        )
+        input_ids = set(i["interface_id"] for i in inputs_data)
+
+        for input_data in inputs_data:
+            # check for upload_pk in input
+            upload_pk = input_data.pop("upload_pk", None)
+            civ = ComponentInterfaceValue.objects.create(**input_data)
```
jmsmkn, line 157: Do we check that the data are valid?

jmsmkn (commented):
I think this approach looks good! I think there are a few things around using slugs rather than titles, we're doing a lot of id passing which would be better with objects, and we need to do more validation of the input data and permission checking. There are lots of places where CIVs are created (both here and in the codebase generally), I'm not sure if this is what we want but it would be good to see the changes from the API client too so I can understand better.

app/grandchallenge/algorithms/serializers.py
miriam-groeneveld: yes

app/grandchallenge/algorithms/serializers.py
miriam-groeneveld: Yes, validation of the ComponentInterfaceValues is done in components/serializers.py

miriam-groeneveld:
I've created a PR for the changes in gcapi as well: https://github.com/DIAGNijmegen/rse-gcapi/pull/76
It allows starting a job with either files or a hyperlink to an image.

miriam-groeneveld:
> a lot of id passing which would be better with objects
I can pass the algorithm object rather than slug. But ComponentInterface can't be passed as an object as it will try and create it, and it already exists. 

> more validation of the input data and permission checking.
I've added permission checking for change_algorithm, that was missing. Permission checking for using images and upload sessions are there already. Not sure what validation should be added still?

app/grandchallenge/algorithms/serializers.py
miriam-groeneveld: It's a bit strange, as you post a job, and the response is also a job. So for the api I think this makes more sense actually.

miriam-groeneveld:
I think this PR is ready for review now

jmsmkn:
Thanks Miriam!

@HarmvZ would you be able to take a look at the DRF stuff here?

app/grandchallenge/components/serializers.py
```diff
@@ -26,32 +30,130 @@ class SimpleImageSerializer(serializers.ModelSerializer):
...
             "file",
             "image",
             "pk",
+            "upload_pk",
         ]
 
+    def validate(self, attrs):  # noqa: C901
+        interface = ComponentInterface.objects.get(
+            slug=attrs["interface_slug"]
+        )
```
HarmvZ, line 68: I think this will throw ModelDoesNotExist if no ComponentInterface exists with this `interface_slug`. Probably better to raise a ValidationError if that is the case.

app/grandchallenge/components/serializers.py
```diff
@@ -26,32 +30,130 @@ class SimpleImageSerializer(serializers.ModelSerializer):
...
+            validate_simple()
+        if interface.kind in InterfaceKind.interface_type_annotation():
+            validate_annotations()
+        if interface.kind in InterfaceKind.interface_type_image():
+            validate_image()
+
+        return attrs
+
+    def validate_upload_pk(self, value):
+        session = RawImageUploadSession.objects.get(pk=value)
```
HarmvZ, line 118: Also raise ValidationError if it does not exist, might not be necessary if using PrimaryKeyRelatedField.

app/grandchallenge/components/serializers.py
```diff
@@ -26,32 +30,130 @@ class SimpleImageSerializer(serializers.ModelSerializer):
...
+
+    image = serializers.HyperlinkedRelatedField(
+        queryset=Image.objects.all(),
+        view_name="api:image-detail",
+        required=False,
+    )
+
+    interface_slug = serializers.CharField(write_only=True)
+    interface = ComponentInterfaceSerializer(read_only=True)
+    upload_pk = serializers.CharField(write_only=True, required=False)
```
HarmvZ, line 51: Possibly [PrimaryKeyRelatedField](https://www.django-rest-framework.org/api-guide/relations/#primarykeyrelatedfield) can be used here to make it more strict. This would also allow you to pass a queryset and filter on session.status.

app/grandchallenge/components/serializers.py
```diff
@@ -26,32 +30,130 @@ class SimpleImageSerializer(serializers.ModelSerializer):
...
+
+        if not user.has_perm("view_rawimageuploadsession", session):
+            raise serializers.ValidationError(
+                f"User does not have permission to use raw image upload session {value}"
+            )
+
+        if session.status is not RawImageUploadSession.PENDING:
+            raise serializers.ValidationError(
+                f"Upload session {value} is not ready to be used"
+            )
```
HarmvZ, line 130: What would happen if the session has another status like failed, cancelled, requeued?

HarmvZ (dismissed):
Cool feature! I have a few small comments about edge cases in validation which might throw non ValidationErrors. It's probably also a good idea to add tests for these cases.

app/grandchallenge/components/serializers.py
miriam-groeneveld: There's a check on session status later in the process, so the job would fail later on.

app/grandchallenge/components/serializers.py
miriam-groeneveld: I opted for the HyperLinkedRelatedField to be more in line with the image (also HyperLinkedRelatedField)

app/grandchallenge/components/serializers.py
```diff
@@ -26,32 +31,139 @@ class SimpleImageSerializer(serializers.ModelSerializer):
...
+    """
+
+    image = serializers.HyperlinkedRelatedField(
+        queryset=Image.objects.all(),
+        view_name="api:image-detail",
+        required=False,
+    )
+
+    interface_slug = serializers.CharField(write_only=True)
+    interface = ComponentInterfaceSerializer(read_only=True)
```
jmsmkn, line 51: Rather than having the object change between post and read we could replace both `interface_slug` and `interface` with:

```
interface = SlugRelatedField(queryset=...,slug_field="slug")
```

What do you think? Would this be too inconsistent with the pks and hyperlinks used elsewhere? I think it would save some of the things happening in L71-L82 in this file.

app/grandchallenge/components/serializers.py
```diff
@@ -26,32 +31,139 @@ class SimpleImageSerializer(serializers.ModelSerializer):
...
+        if value.status is not RawImageUploadSession.PENDING:
+            raise serializers.ValidationError(
+                f"{value} is not ready to be used"
+            )
+        return value
+
+    def validate_image(self, value):
+        user = self.context.get("user")
+
+        if not user.has_perm("view_image", value):
```
jmsmkn, line 146: I wonder if this is too broad, but right now it's probably all we have. Users can get permission to view an image in a variety of ways, and that may not necessarily mean that they can use the image in any way they like. I guess right now they could always download the image and then upload it again as part of an algorithm experiment, so this is adding something new. So, no need to change anything, but I wanted to document it.

app/grandchallenge/algorithms/views.py
```diff
@@ -647,17 +644,30 @@ class AlgorithmImageViewSet(ReadOnlyModelViewSet):
...
     filter_backends = [DjangoFilterBackend, ObjectPermissionsFilter]
     filterset_class = JobViewsetFilter
 
+    def get_serializer_class(self):
+        if self.action == "create":
+            return JobPostSerializer
+        else:
+            return HyperlinkedJobSerializer
+
+    def get_serializer_context(self):
```
jmsmkn, line 665: Good to know that this works!

app/grandchallenge/algorithms/views.py
```diff
@@ -647,17 +644,30 @@ class AlgorithmImageViewSet(ReadOnlyModelViewSet):
     filterset_fields = ["algorithm"]
 
 
-class JobViewSet(ReadOnlyModelViewSet):
+class JobViewSet(
+    CreateModelMixin, RetrieveModelMixin, ListModelMixin, GenericViewSet
```
jmsmkn, line 648: Do you test this from the views? I would have thought that the users would need the global `add_job` permission otherwise it would be rejected by `DjangoObjectPermissions`?

jmsmkn:
Looks great! Couple of comments and questions.

app/grandchallenge/algorithms/views.py
jmsmkn: Ah! That would be it. You can add the model level permissions, then no need to change anything in the ViewSet.

Any authenticated user can create a job as you do the checks on algorithm use later. So, like this but for jobs, not answers: https://github.com/comic/grand-challenge.org/blob/master/app/grandchallenge/reader_studies/apps.py#L27-L35, remember to register the function on post migrate: https://github.com/comic/grand-challenge.org/blob/master/app/grandchallenge/reader_studies/apps.py#L46

app/grandchallenge/components/serializers.py
miriam-groeneveld: Surprised that this works, thought I would get an "interface already exists" warning. Much better!

jmsmkn (approved):
LGTM

jmsmkn (approved):

PR #1850, opened 2021-05-20
Remove retina importer app

jmsmkn:
Removes the retina importer app as this should be done over the normal channels.


HarmvZ (approved):
LGTM

PR #1851, opened 2021-05-20
Extract fixture creation

jmsmkn:
This PR extracts fixture creation from the main application, moves temporary image creation to a fixture, removes an unused command. The `development_fixtures.py` script will work with `runscript` from django extensions which is already included. 

The migration step is moved from the command of the web container to the makefile, instead of using CTRL+D to apply migrations then use `make migrate` instead.

We could now start to split things out so that only the fixtures that the developer needs are created, especially if fixture creation takes a while. If we do that, we should add an all fixtures option.


miriam-groeneveld (approved):
LGTM!

PR #832, opened 2019-06-06
Adds viewsets for image quality and pathology

HarmvZ:
Adds models, serializers, viewsets, urls and tests for image quality, pathology and retinapathology. (part of #734)
Also contains refactoring of some dynamically added batch tests into parametrized tests (part of #677).

PR #1852, opened 2021-05-20
Remove unused test code

jmsmkn:
I noticed on codecov that there were some test code that was no longer used, this PR removes it. At the same time, increase the test speed by not rendering all of the logo variations. Removed an in memory storage provider that wasn't really used (if we need in memory storage use a tempfs for the minio data volume, API consistency is more important). Settings that were only used once in a test factory or not at all were also removed. Removed tests around factory instantiation as these are covered elsewhere, and even if not, they're not really that informative for tests as they tell us nothing about the models behaviour.

The only thing I was unsure about was the classes with the comment "test functions are added dynamically to this class", they actually appeared to do nothing.


HarmvZ (approved):
Nice cleanup!

Indeed the tests with _"test functions are added dynamically to this class"_ did not do anything anymore. It was a leftover from tests I wrote when I was not aware of parametrized tests in pytest yet and tried to implement it myself. Those tests were converted to parametrized tests in https://github.com/comic/grand-challenge.org/pull/832 but I forgot to remove the comments.

jmsmkn:
> 
> 
> Nice cleanup!
> 
> Indeed the tests with _"test functions are added dynamically to this class"_ did not do anything anymore. It was a leftover from tests I wrote when I was not aware of parametrized tests in pytest yet and tried to implement it myself. Those tests were converted to parametrized tests in #832 but I forgot to remove the comments.

Does that mean the `try: ... except Exception: pytest.fail(...)` blocks can also go? eg in https://github.com/comic/grand-challenge.org/blob/2a995006b899c9a0f17e0f4ba862b796423c20a1/app/tests/annotations_tests/test_serializers.py - I guess that was special handling for those?

HarmvZ:
I don't think those had something to do with the custom parametrized tests, but those blocks seem to add no value indeed.

PR #1853, opened 2021-05-21
Fix pytest cache sharing between container runs and add check for breakpoints

jmsmkn:
Fixes the pytest cache to allow re-running of the last failed tests with `--lf`, or running with the failing test first with `--ff``.


miriam-groeneveld (approved):
LGTM

PR #1854, opened 2021-05-22
1843 mdd mdr

Kicksta:
Issue 1843. Add MDD/MDR to product information and add as filter. Change order products


Kicksta:
@MikeOverkamp-diag, @jmsmkn. I notice two checks failed. It says: "Added line #L%%% was not covered by tests"
Do you have hints on how I can fix this? I was hoping to have this live by tomorrow as the new rules (MDR) comes into effect then and I will send out a relating newsletter on Thursday. Thanks for your help!

jmsmkn:
You don't need to worry about those two, they are not required.

@MikeOverkamp-diag could you review this? Thanks!

app/grandchallenge/products/models.py
```diff
@@ -113,6 +113,7 @@ class FDAStatus(models.TextChoices):
     ce_status = models.CharField(
         choices=CEStatus.choices, max_length=3, default=CEStatus.NO
     )
+    ce_under = models.CharField(max_length=10, default="")
```
MikeOverkamp-diag, line 116: You could also use `blank=True` instead of `default=""`. That is a bit more common.

MikeOverkamp-diag (commented):
One minor comment.

MikeOverkamp-diag (approved):

PR #1863, opened 2021-06-01
Add flexible inputs for archives

MikeOverkamp-diag:
This PR adds support for `ComponentInterfaces` other than generic medical images to Archives.

Images can still be added to an Archive the same way as before. The only difference being that the desired `ComponentInterface` for the images now needs to be selected in the form:

Screenshot from 2021-06-01 12-12-50

After adding cases users can now view the uploaded cases in the same manner as before, but there is also a new menu item `View Items`:

Screenshot from 2021-06-01 12-12-14

This takes the user to an overview of all the archive's `Archive Items`:

Screenshot from 2021-06-01 12-13-37

The user then has the option to edit an `ArchiveItem` using a form that lists all available `ComponentInterfaces`:

Screenshot from 2021-06-01 12-14-15

Any fields that are filled in the form will overwrite existing values for the `ComponentInterface` or create a new `ComponentInterfaceValue` in the `ArchiveItem` in case there are none yet for the chosen interface.

Some caveats in the form:

- It is currently not possible to completely remove a `ComponentInterfaceValue` from an `ArchiveItem`
- The current value for a `ComponentInterfaceValue` is only shown for non file or image interfaces.

### DIFF
--- a/app/grandchallenge/archives/forms.py
+++ b/app/grandchallenge/archives/forms.py
@@ -173,7 +173,7 @@ class ArchiveCasesToReaderStudyForm(SaveFormInitMixin, Form):
 class AddCasesForm(UploadRawImagesForm):
     interface = ModelChoiceField(
         queryset=ComponentInterface.objects.filter(
-            kind=InterfaceKind.InterfaceKindChoices.IMAGE
+            kind__in=InterfaceKind.interface_type_image()
         )
     )
 
--- a/app/grandchallenge/archives/forms.py
+++ b/app/grandchallenge/archives/forms.py
@@ -1,3 +1,4 @@
+from crispy_forms.helper import FormHelper
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.forms import (
     Form,
@@ -12,12 +13,20 @@ from guardian.shortcuts import get_objects_for_user
 
 from grandchallenge.algorithms.models import Algorithm
 from grandchallenge.archives.models import Archive, ArchivePermissionRequest
+from grandchallenge.cases.forms import UploadRawImagesForm
 from grandchallenge.cases.models import Image
+from grandchallenge.components.form_fields import InterfaceFormField
+from grandchallenge.components.models import (
+    ComponentInterface,
+    ComponentInterfaceValue,
+    InterfaceKind,
+)
 from grandchallenge.core.forms import (
     PermissionRequestUpdateForm,
     SaveFormInitMixin,
     WorkstationUserFilterMixin,
 )
+from grandchallenge.core.templatetags.bleach import clean
 from grandchallenge.core.widgets import MarkdownEditorWidget
 from grandchallenge.groups.forms import UserGroupForm
 from grandchallenge.reader_studies.models import ReaderStudy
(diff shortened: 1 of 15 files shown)


MikeOverkamp-diag:
@jmsmkn @miriam-groeneveld I'll still add some tests to this, but other than that this can be reviewed.

jmsmkn:
Style wise:

- It is strange to have both the items view and cases view, can we move all of the functionality from cases to items?
- We already have ~50 interfaces so the edit page in production is going to be very busy, is there a better way to do this? Using some JS or getting the slug in the url and filtering the create view by that?

Probably not crucial for now as we just need something that works for internal users, but I think needs some polish very soon.

app/grandchallenge/archives/tasks.py
```diff
@@ -29,3 +33,10 @@ def add_images_to_archive(*, upload_session_pk, archive_pk):
             continue
         item = ArchiveItem.objects.create(archive=archive)
         item.values.set([civ])
+
+
+@shared_task
+def add_values_to_archive_item(*, archive_item_pk, civ_pks):
+    archive_item = ArchiveItem.objects.get(pk=archive_item_pk)
+    archive_item.values.clear()
```
jmsmkn, line 41: Do you really want to clear here? If so, we must change the function name as this is not what I would expect it to do. Must this always be called with the pks that you want to keep? I think there would be less room for future programming mistakes if we separated this out into add and remove operations.

app/grandchallenge/jqfileupload/widgets/uploader.py
```diff
@@ -309,8 +309,8 @@ def delete(self):
 
 class UploadedAjaxFileList(forms.Field):
     def to_python(self, value):
-        if value is None:
-            value = ""
+        if value is None or value == "None":
```
jmsmkn, line 312: Seems strange, in what situations does this occur?

app/grandchallenge/archives/views.py
```diff
@@ -334,6 +358,153 @@ def get_context_data(self, **kwargs):
...
+                with value[0].open() as f:
+                    civ.file = File(f, name=name)
+                    civ.save()
+                civ_pks.add(civ.pk)
+            else:
+                civ.value = value
+                civ.save()
+                civ_pks.add(civ.pk)
+
+        tasks = add_values_to_archive_item.signature(
```
jmsmkn, line 443: Is there a need to do this in a celery task? Downside is that these changes are not instant.

app/grandchallenge/archives/views.py
```diff
@@ -334,6 +358,153 @@ def get_context_data(self, **kwargs):
...
+        return self.archive
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context.update({"archive": self.archive})
+        return context
+
+    def get_queryset(self):
+        qs = super().get_queryset()
+        return qs.filter(archive=self.archive)
```
jmsmkn, line 505: I think that we will need some prefetching for the values titles, did you check the debug toolbar?

app/grandchallenge/archives/tasks.py
MikeOverkamp-diag: What I really wanted to do here is use `.set()`, but unfortunately that did not trigger the `m2mchanged` signal.

app/grandchallenge/jqfileupload/widgets/uploader.py
MikeOverkamp-diag: I haven't figured out the exact situation, but I think for empty file uploads (so when multiple is st to False) an empty field would get a `"None"` value rather than `None`.

app/grandchallenge/archives/views.py
MikeOverkamp-diag: Yes, because this can only be done after the image imports have finished. It could theoretically be done immediately if no image type interfaces are used.

app/grandchallenge/archives/views.py
MikeOverkamp-diag: Did not, but will do!

MikeOverkamp-diag:
As for moving everything to the items view: that was my initial plan, but I couldn't think of a clean way of including all the option that are in the cases view in the items view. I think it would get messy quite quickly and I just don't have the eye for it to make that look halfway decent tbh.

app/grandchallenge/archives/tasks.py
jmsmkn: Ooh wasn't aware that didn't trigger, we should check the codebase for usages of that. Here I would still prefer the explicit remove/add operations, should mean less work with what we do in the signal handler(s) too.

app/grandchallenge/archives/views.py
jmsmkn: Clear!

miriam-groeneveld (commented):
Looks nice! I added one comment.

app/grandchallenge/archives/forms.py
```diff
@@ -159,3 +168,43 @@ def clean(self):
             )
 
         return cleaned_data
+
+
+class AddCasesForm(UploadRawImagesForm):
+    interface = ModelChoiceField(
+        queryset=ComponentInterface.objects.filter(
+            kind=InterfaceKind.InterfaceKindChoices.IMAGE
```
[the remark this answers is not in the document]
MikeOverkamp-diag: Nice catch, you are absolutely right!

jmsmkn (approved):
LGTM - there are a few things that we need to clean up in the next cycle as discussed but this does the job.

PR #1864, opened 2021-06-03
Split add and remove tasks

jmsmkn:
Sorry if my review was unclear - this is what I was intending. What do you think about this approach?

### DIFF
--- a/app/grandchallenge/archives/views.py
+++ b/app/grandchallenge/archives/views.py
@@ -55,7 +55,6 @@ from grandchallenge.archives.serializers import ArchiveSerializer
 from grandchallenge.archives.tasks import (
     add_images_to_archive,
     add_values_to_archive_item,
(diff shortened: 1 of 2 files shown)


app/grandchallenge/archives/views.py
```diff
@@ -413,47 +412,49 @@ def create_upload(image_files):
             return upload_session.pk
 
         upload_pks = {}
-        civ_pks = set(self.archive_item.values.values_list("pk", flat=True))
+        civ_pks_to_remove = set()
+        civ_pks_to_add = set()
 
         for slug, value in form.cleaned_data.items():
+
             if value is None:
```
jmsmkn, line 420: Note - what if the value the user wants to add really is None?

MikeOverkamp-diag (approved):

PR #1872, opened 2021-06-11
Added pagination to algorithms, archives and readerstudies list view

amickan:


app/grandchallenge/algorithms/views.py
```diff
@@ -111,12 +112,49 @@ class AlgorithmList(FilterMixin, PermissionListMixin, ListView):
...
+
+        num_pages = int_paginator.num_pages
+        num_results = int_paginator.count
+
+        try:
+            int_page = int_paginator.page(self._current_page)
+        except EmptyPage:
+            int_page = []
+
+        return int_page, num_pages, num_results, total_count
```
jmsmkn, line 141: This seems to be more or less the same as the code added to the reader study list. Could this be consolidated into either an existing or a new mixin?

app/grandchallenge/algorithms/views.py
```diff
@@ -111,12 +112,49 @@ class AlgorithmList(FilterMixin, PermissionListMixin, ListView):
...
 
     def get_context_data(self, *args, **kwargs):
         context = super().get_context_data(*args, **kwargs)
+        page_obj, num_pages, num_results, total_count = self._get_page()
 
         context.update(
             {
+                "filter": self.int_filter,
+                "filters_applied": self._filters_applied,
```
jmsmkn, line 150: Isn't some of this functionality already in `FilterMixin`?

app/grandchallenge/algorithms/views.py
amickan: Yes, true that. I just added a new mixin. Will change the reader study code once this and the other PR are merged.

jmsmkn: And if _that's_ true, maybe go one step further and change the html to a partial that you include in each page that uses it.

jmsmkn (approved):
LGTM 🚀

PR #1873, opened 2021-06-11
Tags for blog posts

amickan:


app/grandchallenge/blogs/models.py
```diff
@@ -28,6 +36,8 @@ class Post(models.Model):
         variations=settings.STDIMAGE_SOCIAL_VARIATIONS,
     )
 
+    tags = models.ManyToManyField(to=Tag, related_name="blog_tags")
```
jmsmkn, line 39: The related name is how you get these objects (Posts) from the Tags. So, with what you have you would end up with

```
tag = Tag.objects.first()
posts_for_tag = tag.blog_tags.all()
```

Instead, I would suggest `related_name="posts"`, so that `posts_for_tag = tag.posts.all()`.

app/grandchallenge/blogs/templates/blogs/post_detail.html
```diff
@@ -14,20 +14,36 @@
 
 {% block content %}
     <div class="row mb-3">
-        <div class="col-lg-8">
+        <div class="col-lg-9">
```
jmsmkn, line 17: Be careful as it becomes difficult for users to read blocks of text that are really wide.

app/grandchallenge/blogs/filters.py
```diff
@@ -0,0 +1,16 @@
+from django_filters import FilterSet, ModelMultipleChoiceFilter
+from django_select2.forms import Select2MultipleWidget
+
+from grandchallenge.blogs.models import Post, Tag
+from grandchallenge.core.filters import FilterForm
+
+
+class BlogFilter(FilterSet):
```
jmsmkn, line 8: I like this implementation! It also means that we could add author filtering (but be really careful about that! filter the queryset by the users that are authors first), or simple text search of the content.

app/grandchallenge/blogs/templates/blogs/post_detail.html
amickan: ok, good point, I can change it back to what it was before ;-)

app/grandchallenge/blogs/filters.py
amickan: Yeah, filtering by authors is not a bad idea. Should I go ahead and add this or would you rather leave it at the tag filter for now?

app/grandchallenge/blogs/filters.py
```diff
@@ -9,8 +10,15 @@ class BlogFilter(FilterSet):
     tags = ModelMultipleChoiceFilter(
         queryset=Tag.objects.all(), widget=Select2MultipleWidget, label="Tags",
     )
+    authors = ModelMultipleChoiceFilter(
+        queryset=User.objects.filter(
+            blog_authors__in=Post.objects.all()
+        ).all(),
```
jmsmkn, line 21: In Django you should never use `User` directly. Use `get_user_model()` instead. See https://docs.djangoproject.com/en/3.2/topics/auth/customizing/#django.contrib.auth.get_user_model. 

You can also save a query by doing `get_user_model().objects.exclude(blog_authors__isnull=True).order_by("username")`

The `order_by` is there as I think the default ordering for a django user is by signup date.

app/grandchallenge/blogs/templates/blogs/post_detail.html
```diff
@@ -20,14 +20,30 @@ <h2>{{ object.title }}</h2>
...
+                <ul class="list-group list-group-flush">
+                    {% for author in object.authors.all %}
+                        <li class="list-group-item border-0">
+                            {{ author|user_profile_link }}
+                        </li>
+                    {% endfor %}
+                </ul>
+            </div>
+            <div class="row mb-1">
+                <h3>Tags</h3><br>
```
jmsmkn, line 36: Why is the `<br>` element here? If it is just for spacing then use margins instead for consistency.

We should only show the Tags header if there are tags, right?

amickan: Good that you mention it, at the moment, tags are required. I should add `blank=True` to the tags field in the post model. 
And the line break can go as well.

jmsmkn: Blank is only to do with validation, so for any existing objects the m2m through table will be empty. I think that it is fine to keep it as is then, but after we deploy we will need to go back through the old posts and add some tags, I will make a note to do that.

amickan: oh I was too quick, sorry. I will change it back to what it was, if we really want tags to be required.

jmsmkn: No it's ok! Let's leave it as is, I don't have strong opinions about it.

jmsmkn (approved):

PR #1882, opened 2021-06-17
Hides modal after regenerating hanging protocol

### LINKED ISSUE #1877: Modal is not dismissed after hanging list is regenerated
When a user clicks on regenerate hanging list for a reader study nothing seems to happen, but the endpoint is called. They should get the message (missing redirect?)

amickan:
Closes #1877


jmsmkn:
Do the users still see the success message? I would have thought that required a redirect?

amickan:
> Do the users still see the success message? I would have thought that required a redirect?

Yes, they do see the success message. Would a redirect still be better?

PR #1883, opened 2021-06-20
Add GitHub / webhooks app

MikeOverkamp-diag:
This adds a webhooks app that allows gc to communicate with a [GitHub app](https://docs.github.com/en/developers/apps/building-github-apps).

In order to work properly, a GitHub app needs to be created. The app needs to have Read only persmissions to `Contents` only and needs to subscribe to the `Create` event.

The following settings need to be added:

Screenshot from 2021-06-20 16-11-00

A client secret, webhook secret and private key should also be generated.

The following environment variables need to then be set:

- `GITHUB_APP_INSTALL_URL`: `https://github.com/apps/{app_name}/installations/new`
- `GITHUB_APP_ID`: the app's App ID
- `GITHUB_CLIENT_ID`: the app's Client ID
- `GITHUB_CLIENT_SECRET`: the generated client secret
- `GITHUB_PRIVATE_KEY_BASE64`: base64 encoded private key dowloaded from the app
- `GITHUB_WEBHOOK_SECRET`: the generated webhook secret


app/config/settings.py
```diff
@@ -477,6 +477,7 @@ def strtobool(val) -> bool:
     "grandchallenge.datatables",
     "grandchallenge.organizations",
     "grandchallenge.groups",
+    "grandchallenge.webhooks",
```
jmsmkn, line 480: What do you think about calling the app `github`? Webhooks might be a bit too generic and the settings are already namespaced with `GITHUB_`.

app/config/settings.py
```diff
@@ -1041,6 +1042,14 @@ def strtobool(val) -> bool:
     *blocklist,
 }
 
+# GitHub App
+GITHUB_APP_INSTALL_URL = os.environ.get("GITHUB_APP_INSTALL_URL", "")
+GITHUB_APP_ID = os.environ.get("GITHUB_APP_ID", "")
+GITHUB_CLIENT_ID = os.environ.get("GITHUB_CLIENT_ID", "")
+GITHUB_CLIENT_SECRET = os.environ.get("GITHUB_CLIENT_SECRET", "")
+GITHUB_PRIVATE_KEY_PATH = os.environ.get("GITHUB_PRIVATE_KEY_PATH", "")
```
jmsmkn, line 1050: Rather than using a path could we include the key as an environment variable? Easier to deploy without data volumes then.

app/config/urls/root.py
```diff
@@ -45,6 +46,7 @@ def handler500(request):
 
 urlpatterns = [
     path("", HomeTemplate.as_view(), name="home"),
+    path("github-install-complete", post_install_redirect),
```
jmsmkn, line 49: Missing a trailing slash and would be nicer to include the urls from the app rather than importing a view from another app here.

app/grandchallenge/algorithms/views.py
```diff
@@ -753,3 +759,50 @@ def get_context_data(self, **kwargs):
...
+    def get_form_kwargs(self):
+        """Return the keyword arguments for instantiating the form."""
+        kwargs = super().get_form_kwargs()
+
+        code = self.request.GET.get("code")
+        repos = []
+        if code is not None:
+            headers = {"Accept": "application/vnd.github.v3+json"}
+
+            resp = requests.post(
```
jmsmkn, line 784: The post request should have a timeout, should be short-ish as it's in a view, 5s?

app/grandchallenge/algorithms/views.py
```diff
@@ -753,3 +759,50 @@ def get_context_data(self, **kwargs):
...
+                "https://github.com/login/oauth/access_token",
+                data={
+                    "code": code,
```
jmsmkn, line 787: Do we need to clean `code` at all?

app/grandchallenge/algorithms/views.py
```diff
@@ -753,3 +759,50 @@ def get_context_data(self, **kwargs):
...
+                    "client_id": settings.GITHUB_CLIENT_ID,
+                    "client_secret": settings.GITHUB_CLIENT_SECRET,
+                },
+                headers=headers,
+            )
+
+            payload = resp.json()
+            headers["Authorization"] = f"token {payload['access_token']}"
+            installations = requests.get(
```
jmsmkn, line 796: Needs a timeout here too.

app/grandchallenge/algorithms/views.py
```diff
@@ -753,3 +759,50 @@ def get_context_data(self, **kwargs):
...
+                "https://api.github.com/user/installations", headers=headers
+            ).json()
+
+            response = requests.get(
```
jmsmkn, line 800: And here. Are 3 external API calls going to be too much for within a request?

app/grandchallenge/webhooks/tasks.py
```diff
@@ -0,0 +1,61 @@
...
+        "iat": int(now.timestamp()) - 60,
+        "exp": int(now.timestamp()) + 60 * 10,
+        "iss": settings.GITHUB_APP_ID,
+    }
+    token = jwt.encode(msg, private_key, algorithm="RS256")
+    headers = {
+        "Authorization": f"Bearer {token}",
+        "Accept": "application/vnd.github.v3+json",
+    }
+    resp = requests.post(
```
jmsmkn, line 38: Timeout.

app/grandchallenge/webhooks/tasks.py
```diff
@@ -0,0 +1,61 @@
...
+        private_key = serialization.load_pem_private_key(
+            key_file.read(), password=None, backend=default_backend()
+        )
+    now = datetime.now()
+    msg = {
+        "iat": int(now.timestamp()) - 60,
+        "exp": int(now.timestamp()) + 60 * 10,
+        "iss": settings.GITHUB_APP_ID,
+    }
+    token = jwt.encode(msg, private_key, algorithm="RS256")
```
jmsmkn, line 33: Has this been double checked? JWTs have a bad rep for being hard to get right, I think with the algorithm explicitally set we're ok but I'm unsure.

app/grandchallenge/webhooks/tasks.py
```diff
@@ -0,0 +1,61 @@
...
+        f"https://api.github.com/app/installations/{installation_id}/access_tokens",
+        headers=headers,
+    )
+    access_token = json.loads(resp.content)["token"]
+    full_name = payload["repository"]["full_name"]
+    headers["Authorization"] = f"token {access_token}"
+    tarball_url = (
+        f"https://api.github.com/repos/{full_name}/tarball/{payload['ref']}"
+    )
+    with requests.get(tarball_url, headers=headers, stream=True) as file:
```
jmsmkn, line 48: Does getting the tarball like this include any submodules and LFS objects? Is it definitely the branch and tag we need? Again, timeout needed.

app/grandchallenge/webhooks/views.py
```diff
@@ -0,0 +1,57 @@
...
+@require_POST
+@non_atomic_requests
+def github_webhook(request):
+    signature = hmac.new(
+        bytes(settings.GITHUB_WEBHOOK_SECRET, encoding="utf8"),
+        msg=request.body,
+        digestmod=hashlib.sha256,
+    ).hexdigest()
+    signature = f"sha256={signature}"
+    if not compare_digest(
```
jmsmkn, line 29: This is constant time, right?

app/grandchallenge/webhooks/views.py
```diff
@@ -0,0 +1,57 @@
...
+from django.views.decorators.csrf import csrf_exempt
+from django.views.decorators.http import require_POST
+
+from grandchallenge.webhooks.models import GitHubWebhookMessage
+
+
+@csrf_exempt
+@require_POST
+@non_atomic_requests
+def github_webhook(request):
```
jmsmkn, line 22: Would prefer CBV but not that important.

jmsmkn (commented):
Looks great! Looking forward to getting this. I will see what integration we can do with terraform. Few questions before we merge.

app/config/settings.py
MikeOverkamp-diag: I'm not sure how well that would work with the length of the key and the multiline nature?

jmsmkn: It's possible (use single quotes in your `.env` file), but base64 encoding might be more robust if you want to go down that route.

app/grandchallenge/webhooks/views.py
MikeOverkamp-diag: It is.

app/grandchallenge/webhooks/tasks.py
MikeOverkamp-diag: I'm not entire sure on that, as I didn't test with any repos that involved either. Is there a repo we can use to test with?

app/grandchallenge/webhooks/tasks.py
MikeOverkamp-diag: It took me forever to get right, so that rep is well earned I'd say. This seems to work properly though.

app/grandchallenge/algorithms/views.py
MikeOverkamp-diag: I think this should be ok as it is a query param and should be url encoded? Either way, it only gets posted to github, so worst case scenario for us is that we don't get the proper response I think.

app/config/settings.py
MikeOverkamp-diag: That's probably better, as there's also a redirect view in there now. Not just a webhook receiver.

app/grandchallenge/algorithms/views.py
MikeOverkamp-diag: It's a bit much. But there's really no way around it, as we cannot do these async. We need the repos before we can return the response. And unfortunately with he way the github API is set up, we need all of these calls to get the repos.

app/config/settings.py
jmsmkn: Did you try doing this?

MikeOverkamp-diag: Ah sorry, I missed that comment. I'll try it now.

app/grandchallenge/github/views.py
```diff
@@ -0,0 +1,55 @@
...
+    if not compare_digest(
+        signature, request.headers.get("X-Hub-Signature-256", "")
+    ):
+        return HttpResponseForbidden(
+            "Signatures do not match", content_type="text/plain",
+        )
+
+    GitHubWebhookMessage.objects.filter(
+        created__lte=timezone.now() - timedelta(days=7)
+    ).delete()
```
jmsmkn, line 38: Why are all the old messages being deleted in the POST view? Won't this delete all of the tarballs too? If this is needed, I would have thought that it should be done in a periodic task.

PR #1885, opened 2021-06-24
Adds api endpoint for notifications

amickan:
Bulk update and delete on notification list view is now done through the rest api.


app/grandchallenge/notifications/serializers.py
```diff
@@ -0,0 +1,9 @@
+from rest_framework import serializers
+
+from grandchallenge.notifications.models import Notification
+
+
+class NotificationSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Notification
+        fields = "__all__"
```
jmsmkn, line 9: It's better to be explicit about what fields you're serializing. When using all `"__all__"`, if a developer adds a really private field on the model then it will be delivered to the user over the rest API! The same goes for Django forms, or django create/update views. Always set `fields = ("this", "that", etc ...)`.

jmsmkn: We should find if there is a way to automate a check like that...

amickan: Ah yes, that makes sense. Now I'm wondering whether the user and action fields of the notification are necessary to include in the serializer. All we really need to update is the 'read' field, right?

amickan: > We should find if there is a way to automate a check like that...

By that you mean checking that model serializers don't use `__all__` in their fields specification?

jmsmkn: Yes to both of those questions.

app/grandchallenge/notifications/templates/notifications/notification_list.html
```diff
@@ -59,25 +62,18 @@ <h2>Notifications</h2>
...
-             document.getElementById('LabelSelectAll').innerHTML = n_checked + ' selected'
-          }
-        }
+        window.drf = {
+            csrfHeaderName: "{{ csrf_header_name|default:'X-CSRFToken' }}",
+            csrfToken: "{% if request %}{{ csrf_token }}{% endif %}"
+        };
     </script>
+    <script src="{% static "rest_framework/js/csrf.js" %}"></script>
+    <script src="{% static "js/notifications/bulk_delete_update.js" %}"></script>
```
jmsmkn, line 78:
```suggestion
    <script src="{% static "js/notifications/bulk_delete_update.js" %}" type="module"></script>
```

app/grandchallenge/notifications/views.py
```diff
@@ -1,28 +1,61 @@
...
 from grandchallenge.notifications.models import Notification
+from grandchallenge.notifications.serializers import NotificationSerializer
 from grandchallenge.notifications.utils import (
     prefetch_generic_foreign_key_objects,
     prefetch_nested_generic_foreign_key_objects,
 )
 from grandchallenge.subdomains.utils import reverse
 
 
+class NotificationViewSet(viewsets.ModelViewSet):
```
jmsmkn, line 25: `ModelViewSet` includes a bunch of methods we do not use, or test. I would prefer it if we were explicit about the allowed methods, just in case the user somehow is assigned permission. If you look at https://github.com/encode/django-rest-framework/blob/e92016ac2e926483e05e296558fc3d1ea3279625/rest_framework/viewsets.py#L235-L240 it includes:

```
class ModelViewSet(mixins.CreateModelMixin,
                   mixins.RetrieveModelMixin,
                   mixins.UpdateModelMixin,
                   mixins.DestroyModelMixin,
                   mixins.ListModelMixin,
                   GenericViewSet):
```

I think that what we would need everything except the create, and we can use read only as the base as this is safer:

```
from rest_framework import mixins
...
NotificationViewSet(mixins.UpdateModelMixin,
                   mixins.DestroyModelMixin,
                   viewsets.ReadOnlyModelViewSet):
```

jmsmkn (approved):
LGTM

PR #1886, opened 2021-06-29
Auto-jump to center of gravity

miriam-groeneveld:
Add setting for auto-jump to center of gravity when viewing algorithm results


miriam-groeneveld:
Not sure why the tests fail on the migration file...

jmsmkn:
Does makemigrations generate a new file locally? If so, what are the contents of that?

jmsmkn (approved):

PR #1889, opened 2021-06-30
Adds api view for subscription list

amickan:
- bulk delete for subscriptions
- pagination for subscription list
- filters for subscriptions 

- also added domain name to notification email subject

### DIFF
--- a/app/grandchallenge/core/templates/grandchallenge/partials/filters.html
+++ b/app/grandchallenge/core/templates/grandchallenge/partials/filters.html
@@ -1,6 +1,12 @@
 {% load crispy_forms_tags %}
 {% load meta_attr %}
 
+{% comment %}
+    The model_name variable was added to adjust the displayed filter name for the actstream Follow model.
+    It would be nicer to instead add a proxy model for the actstream Follow model on which we set the verbose name accordingly.
+    Trying this out, however, we could not get the permission checks to work for the proxy model.
+    It might be worth debugging why the permissions for the proxy model were not set correctly and to get rid of this work-around eventually.
+{% endcomment %}
 <p>
     <a class="btn btn-primary"
        role="button"
--- a/app/grandchallenge/core/templates/grandchallenge/partials/filters.html
+++ b/app/grandchallenge/core/templates/grandchallenge/partials/filters.html
@@ -8,7 +8,7 @@
        data-target="#filterForm"
        aria-expanded="false"
        aria-controls="filterForm">
-        <i class="fas fa-filter"></i>&nbsp;Filter {{ filter.Meta.model|meta_attr:'verbose_name_plural'|title }}
+        <i class="fas fa-filter"></i>&nbsp;Filter  {% if not model_name %}{{ filter.Meta.model|meta_attr:'verbose_name_plural'|title }}{% else %} {{ model_name|title }}s {% endif %}
     </a>
     {% if filters_applied %}
         <a class="btn btn-danger" href="?">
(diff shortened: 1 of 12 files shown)


app/grandchallenge/core/templates/grandchallenge/partials/filters.html
```diff
@@ -8,7 +8,7 @@
        data-target="#filterForm"
        aria-expanded="false"
        aria-controls="filterForm">
-        <i class="fas fa-filter"></i>&nbsp;Filter {{ filter.Meta.model|meta_attr:'verbose_name_plural'|title }}
+        <i class="fas fa-filter"></i>&nbsp;Filter  {% if not model_name %}{{ filter.Meta.model|meta_attr:'verbose_name_plural'|title }}{% else %} {{ model_name|title }}s {% endif %}
```
amickan, line 11: I made these changes here, because otherwise it would say `Filter Follows`, which I thought might be confusing for the user. But maybe it's a bit of a convoluted code change for such little benefit...

app/grandchallenge/notifications/filters.py
```diff
@@ -48,3 +50,73 @@ def search_filter(self, queryset, name, value):
                 or_, [Q(**{f"{f}__in": name_qs}) for f in search_fields], Q(),
             )
         )
+
+
+FOLLOW_CHOICES = (
+    ("110", "Forum subscriptions"),
+    ("121", "Topic subscriptions"),
```
jmsmkn, line 57: I think that you have hardcoded the content type primary key here, right? I don't think that you can do that as those IDs are unlikely to be the same in production. Another approach is needed.

app/grandchallenge/notifications/filters.py
```diff
@@ -48,3 +50,73 @@ def search_filter(self, queryset, name, value):
...
+                for x in Topic.objects.filter(subject__icontains=value).all()
+            ]
+            return queryset.filter(
+                **{"object_id__in": name_qs},
+                **{
+                    "content_type__exact": ContentType.objects.filter(
+                        model="topic", app_label="forum_conversation"
+                    ).get()
+                },
+            )
```
jmsmkn, line 106: Nit-pick: you could tidy this up a little by setting `model`, `model_name`, and `app_label` variables, then applying the constructions rather than duplicating them.

app/grandchallenge/notifications/views.py
```diff
@@ -48,6 +54,22 @@ def update(self, request, *args, **kwargs):
         return response
 
 
+class FollowViewSet(
+    mixins.DestroyModelMixin, viewsets.ReadOnlyModelViewSet,
+):
+    queryset = Follow.objects.all()
+    serializer_class = FollowSerializer
+    permission_classes = (DjangoObjectPermissions,)
```
jmsmkn, line 62: Are there tests for the permissions and functionality?

jmsmkn (changes requested):
Looks good! Only one change really required.

app/grandchallenge/core/templates/grandchallenge/partials/filters.html
amickan: Ah, yeah, that makes sense. But since the Follow model is from actstream, this would mean overriding the actstream model, but if I get it to work that should be fine.

app/grandchallenge/notifications/views.py
amickan: There are now! Sorry, I was under the impression that I had added them, but I apparently didn't.

jmsmkn (approved):

PR #1892, opened 2021-06-30
Add CodeBuild app

MikeOverkamp-diag:


app/grandchallenge/codebuild/tasks.py
```diff
@@ -0,0 +1,28 @@
+from celery import shared_task
+from django.apps import apps
+from django.conf import settings
+
+from grandchallenge.algorithms.models import Algorithm
+from grandchallenge.codebuild.client import CodeBuildClient
+
+
+@shared_task()
+def create_algorithm_image(*, pk):
```
jmsmkn, line 10: All celery tasks need to be idempotent and (ideally) short lived. I think if this is called twice with the same arguments now then it will create a new job every time. Then, rather than blocking for completion add a new celery task that checks for completion using the countdown argument (eg https://github.com/comic/grand-challenge.org/blob/master/app/grandchallenge/workspaces/tasks.py#L84 and how that is used in a chain, note that MaxRetriesExceeded should be handled). We also need to handle setting a time limit on the build task.

app/grandchallenge/codebuild/client.py
```diff
@@ -0,0 +1,85 @@
...
+    build_id = None
+
+    def __init__(
+        self, *, project_name,
+    ):
+        self.client = boto3.client(
+            "codebuild",
+            aws_access_key_id=settings.CODEBUILD_ACCESS_KEY,
+            aws_secret_access_key=settings.CODEBUILD_SECRET_KEY,
+            region_name="eu-central-1",
```
jmsmkn, line 21: Region should be configurable

app/grandchallenge/codebuild/client.py
```diff
@@ -0,0 +1,85 @@
...
+        )
+        self.s3_client = boto3.client(
```
jmsmkn, line 23: You shouldn't need to set up the separate s3 client, can't you use `grandchallenge.core.storage.private_s3_storage` which is already configured?

app/grandchallenge/codebuild/client.py
```diff
@@ -0,0 +1,85 @@
...
+                "image": "docker:dind",
+                "computeType": "BUILD_GENERAL1_SMALL",
+                "privilegedMode": True,
+                "registryCredential": {
+                    "credential": settings.DOCKER_SECRET_ARN,
+                    "credentialProvider": "SECRETS_MANAGER",
+                },
+                "imagePullCredentialsType": "SERVICE_ROLE",
+            },
+            "serviceRole": settings.CODEBUILD_SERVICE_ROLE,
```
jmsmkn, line 48: What permissions does the service role need?

app/grandchallenge/codebuild/templates/codebuild/buildspec.yml
```diff
@@ -0,0 +1,16 @@
+version: 0.2
+
+phases:
+  install:
+    commands:
+      - nohup /usr/local/bin/dockerd --host=unix:///var/run/docker.sock --host=tcp://127.0.0.1:2375 --storage-driver=overlay2 &
+      - timeout 15 sh -c "until docker info; do echo .; sleep 1; done"
+  pre_build:
+    commands:
+      - docker build -t {{ algorithm_name }} {{ folder_name }}
```
jmsmkn, line 10: We should think about the repo and labels we use. `algorithm_name` might not be docker safe. I want to migrate from storing the images on S3 to private ECR, so will look up the format to use for the repo and see if we can pass this in. For the label we can generate a new UUID before building, and then use this UUID when the `AlgorithmImage` is created.

app/grandchallenge/codebuild/tasks.py
MikeOverkamp-diag: I'll look into that. I'm currently adding the `Build` model which should make this a bit easier as we don't need to save the state in the client anymore. We need to think of an acceptable time limit for the build to finish though, as the build might take a while.

app/grandchallenge/codebuild/client.py
MikeOverkamp-diag: Of course!

app/grandchallenge/codebuild/client.py
MikeOverkamp-diag: I'll look into this.

app/grandchallenge/codebuild/client.py
MikeOverkamp-diag: I'll narrow down the permissions to a minimum after I finish this PR and let you know.

app/grandchallenge/codebuild/templates/codebuild/buildspec.yml
MikeOverkamp-diag: `algorithm_name` should be docker save as it's lower case and all non alphanumerical characters are replaced by dashes. It really does not matter much what we use here though, so a UUID would work as well.

app/grandchallenge/codebuild/tasks.py
jmsmkn: It's just a build right? The users are probably not compiling a huge amount and not running tests so I think that a 15 minute limit would be a good starting point.

MikeOverkamp-diag: Sounds good to me. I think the average build time for my (very simple) algorithm was under a minute.

app/grandchallenge/codebuild/client.py
```diff
@@ -0,0 +1,99 @@
...
+        self.config = {
+            "name": project_name,
+            "source": {"type": "S3", "location": ""},
+            "artifacts": {
+                "type": "S3",
+                "location": settings.PRIVATE_S3_STORAGE_KWARGS["bucket_name"],
+            },
+            "environment": {
+                "type": "LINUX_CONTAINER",
+                "image": "docker:dind",
```
jmsmkn, line 42: Will we need a docker pro account to pull this? Should we instead have our own build images on ECR and pull from there?

app/grandchallenge/codebuild/client.py
```diff
@@ -0,0 +1,99 @@
...
+                },
+                "imagePullCredentialsType": "SERVICE_ROLE",
+            },
+            "serviceRole": settings.CODEBUILD_SERVICE_ROLE,
+            "encryptionKey": settings.CODEBUILD_ENCRYPTION_KEY,
+        }
+
+    def create_build_project(self, *, source):
+        self.config["source"]["location"] = source
+        self.client.create_project(**self.config)
```
jmsmkn, line 57: This creates a new project for each build, right? Would it not make more sense for use to create the project in Terraform, and then we use it from grand challenge with some parametrisation eg with the input and output variables? Is that even possible?

app/grandchallenge/codebuild/client.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def get_logs(self):
+        return self.log_client.get_log_events(
+            logGroupName=self.build["logs"]["groupName"],
+            logStreamName=self.build["logs"]["streamName"],
+        )
+
+    def add_image_to_algorithm(self):
+        with private_s3_storage.open(
+            f"{self.project_name}/{self.project_name}.tar"
```
jmsmkn, line 85: This is pushing to the root of the private storage, we need to add some subfolders.

app/grandchallenge/codebuild/models.py
```diff
@@ -0,0 +1,23 @@
+from django.db import models
+
+from grandchallenge.core.models import UUIDModel
+
+
+class Build(UUIDModel):
+    created = models.DateTimeField(auto_now_add=True)
```
jmsmkn, line 7: Created is already in UUIDModel

app/grandchallenge/codebuild/tasks.py
```diff
@@ -0,0 +1,66 @@
...
+
+@shared_task(bind=True, max_retries=100)
+def wait_for_build_completion(self, *, build_pk):
+    build = Build.objects.get(pk=build_pk)
+
+    client = CodeBuildClient(build_id=build.build_id)
+
+    status = client.get_build_status()
+    if status == "IN_PROGRESS":
+        self.retry(countdown=30)
```
jmsmkn, line 45: What happens when the time limit is exceeded? Can we limit that in code build?

app/grandchallenge/codebuild/tasks.py
```diff
@@ -0,0 +1,66 @@
...
+                )
+            )
+
+
+@shared_task
+def add_image_to_algorithm(*, build_pk):
+    build = Build.objects.get(pk=build_pk)
+
+    client = CodeBuildClient(
+        project_name="mikeoverkamp-diag-private-3-v0-0-103",
```
jmsmkn, line 63: Hardcoded project name

app/grandchallenge/codebuild/client.py
MikeOverkamp-diag: The number of pulls you can do as a free user is limited to (I think) 200 every 6 hours, but it adds up quickly. It would be nice if we can host our own for sure. For now we can at least use a docker hub account and feed the credentials into a secret in AWS Secrets Manager. The ARN for the secret can then be added under the `DOCKER_SECRET_ARN` setting.

app/grandchallenge/codebuild/client.py
MikeOverkamp-diag: No unfortunately not. The source file location is set on build project creation, so we need a new project for each new zip file. I could look into a workaround for this in the cooldown week, but I think this is at least the official way to do things.

app/grandchallenge/codebuild/client.py
MikeOverkamp-diag: Ah yes, that's probably a good idea.

app/grandchallenge/codebuild/tasks.py
MikeOverkamp-diag: I think CodeBuild itself has a 60 minute timeout. I can either raise our own timeout to that, or I can check if that is configurable somewhere.

app/grandchallenge/codebuild/tasks.py
MikeOverkamp-diag: Oops. Blame failing builds.

app/grandchallenge/codebuild/client.py
jmsmkn: From https://docs.aws.amazon.com/codebuild/latest/userguide/sample-docker.html you should be able to use one of the standard images to do this

```
  "environment": {
    "type": "LINUX_CONTAINER",
    "image": "aws/codebuild/standard:4.0",
    "computeType": "BUILD_GENERAL1_SMALL",
    "environmentVariables": [
      {
        "name": "AWS_DEFAULT_REGION",
        "value": "region-ID"
      },
      {
        "name": "AWS_ACCOUNT_ID",
        "value": "account-ID"
      },
      {
        "name": "IMAGE_REPO_NAME",
        "value": "Amazon-ECR-repo-name"
      },
      {
        "name": "IMAGE_TAG",
        "value": "latest"
      }
    ],
    "privilegedMode": true
  },
```

MikeOverkamp-diag: That would be nice. The the only bottleneck could be images pulled by the dockerfiles uploaded by users. I cannot add this change anymore though, since the config has been moved out of gc.

app/grandchallenge/codebuild/client.py
```diff
@@ -0,0 +1,77 @@
...
+            aws_access_key_id=settings.CODEBUILD_ACCESS_KEY,
+            aws_secret_access_key=settings.CODEBUILD_SECRET_KEY,
+            region_name=settings.CODEBUILD_REGION,
+        )
+        self.log_client = boto3.client(
+            "logs",
+            aws_access_key_id=settings.CODEBUILD_ACCESS_KEY,
+            aws_secret_access_key=settings.CODEBUILD_SECRET_KEY,
+            region_name=settings.CODEBUILD_REGION,
+        )
```
jmsmkn, line 28: No longer needed, fetch logs from `PRIVATE_S3_STORAGE_BUCKET_NAME/codebuild/logs/<long_build_number>.gz`

app/grandchallenge/codebuild/client.py
```diff
@@ -0,0 +1,100 @@
...
+        self.msg = msg
+        self.algorithm = algorithm
+        self.build_id = build_id
+        if build_id is not None:
+            self.build_number = build_id.split(":")[-1]
+
+    def start_build(self, source, folder_name):
+        data = self.client.start_build(
+            projectName=settings.CODEBUILD_PROJECT_NAME,
+            sourceLocationOverride=source,
```
jmsmkn, line 33:
```suggestion
            sourceLocationOverride=source,
            sourceTypeOverride="S3",
```

app/grandchallenge/codebuild/models.py
```diff
@@ -0,0 +1,36 @@
...
+    )
+
+
+class Build(UUIDModel):
+    build_id = models.CharField(max_length=1024)
+    project_name = models.CharField(max_length=1024)
+    status = models.CharField(blank=True, max_length=256)
+    build_log = models.FileField(
+        null=True, upload_to=log_path, storage=private_s3_storage
+    )
```
jmsmkn, line 24: You should be able to read the log files with the `gzip` module and then save them to a TextField.

jmsmkn:
I have done some refactoring to:

- Keep the github sources in the github folder on S3 for app name consistency, updated the codebuild permissions to read from there
- Refactored the client as I found tracing through the ids very confusing, this is now all on the `Build` model
- `Build` is now foreign keyed to the algorithm image that it is going to fill, will be useful for tracking state later
- Saving the entire job config and using this later to get the logs and artefacts in case the projects change
- Removed having to set the folder name, in the buildspec file we assume that only 1 directory will be present, ok for now
- Added admin views
- Added a timeout

We need to check that the workflow is ok - I couldn't test end to end adding a repo and building from that, and I didn't test private repos either. The user should be able to see what repo their algorithm is linked to, and the status of the jobs. Also weird that if a branch is tagged then it will still be used, the user should be aware of that. We also need tests!

jmsmkn (approved):

PR #1897, opened 2021-07-02
Notifications for archive, reader study and algorithm access requests

### LINKED ISSUE #1891: Send notification rather than email for new algorithm/readerstudy/archive access requests
Currently, when a user requests access to an algorithm (or archive or reader study), editors of that algorithm receive an email about the request. After acceptance or rejection of the request, the user in turn receives an email with the status update. Instead of sending emails, it would be nice to send notifications about access requests and request status updates.

To-do:
- create follow object: `editors of an algorithm` --> `algorithm` (add this on algorithm creation and when an editor is added, remove when editor is removed)
- create follow object: `requester` --> `AlgorithmPermissionRequest` (post-save of the request object)
- post-save of `AlgorithmPermissionRequest` send an action with `algorithm` as target and `requester` as actor 
- when `AlgorithmPermissionRequest` status is updated, send an action with `AlgorithmPermissionRequest` as target (and editor who grants/denies access as actor, if necessary, need to check) 

The actions will then automatically  create the desired notitications.

amickan:
- adds notifications for permission requests for archives, reader studies and algorithms 
- gets rid of emails that used to be sent for those

Closes #1891


amickan:
Once this is added, we might also need to adjust the notification and subscription filters. They currently only work for forums and topics, so searching for access request related notifications / subscriptions is not possible. I'm not sure if this type of filter is necessary though, just thinking out loud. The more types of notifications we add, the longer the filter form, also something to keep in mind.

app/grandchallenge/algorithms/models.py
```diff
@@ -293,9 +294,11 @@ def is_editor(self, user):
         return user.groups.filter(pk=self.editors_group.pk).exists()
 
     def add_editor(self, user):
+        follow(user=user, obj=self, actor_only=False, send_action=False)
         return user.groups.add(self.editors_group)
 
     def remove_editor(self, user):
+        unfollow(user=user, obj=self, send_action=False)
```
jmsmkn, line 301: I don't think adding the events here is the best way to go as there are many ways that an editor can be added and removed from the group and we cannot force all codepaths down this route. I think what is needed is to add a signal handler on `m2m_changed` of the groups members, those are quite tricky to set up but there are a few examples in the codebase, let me know and I can point you to them.

app/grandchallenge/core/signals.py
```diff
@@ -54,11 +58,50 @@ def process_permission_request(sender, instance, *_, **__):
...
+
+
+@receiver(pre_delete, sender=Algorithm)
+@receiver(pre_delete, sender=Archive)
+@receiver(pre_delete, sender=ReaderStudy)
+@receiver(pre_delete, sender=AlgorithmPermissionRequest)
+@receiver(pre_delete, sender=ArchivePermissionRequest)
+@receiver(pre_delete, sender=ReaderStudyPermissionRequest)
+def delete_editor_follows(sender, instance, *_, **__):
+    Follow.objects.filter(object_id=instance.pk).delete()
```
jmsmkn, line 107: Signals really should only be used as a last resort as it makes code very difficult to reason about. We are in control of these models, and will not have a problem with circular imports, which means that we can extend their `save` and `delete` methods rather than using signals.

app/grandchallenge/notifications/templates/notifications/notification_list.html
```diff
@@ -45,7 +46,12 @@ <h2>Notifications</h2>
                                 <span class="text-primary align-middle ml-2"><i class="fa fa-circle pl-2"></i></span>
                             {% endif %}
                             &nbsp;
-                            {{ object.action.actor|user_profile_link }}{{ object.action.verb }}
+                            {% if object.action.actor|get_content_type == 3 %}
+                                {{ object.action.actor|user_profile_link }}
+                            {% elif object.action.actor|get_content_type == 35 or object.action.actor|get_content_type == 70 or object.action.actor|get_content_type == 39 %}
```
jmsmkn, line 51: Remove hardcoded content types.

app/grandchallenge/archives/models.py
```diff
@@ -174,9 +175,11 @@ def is_editor(self, user):
         return user.groups.filter(pk=self.editors_group.pk).exists()
 
     def add_editor(self, user):
+        follow(user=user, obj=self, actor_only=False, send_action=False)
         return user.groups.add(self.editors_group)
 
     def remove_editor(self, user):
+        unfollow(user=user, obj=self, send_action=False)
```
jmsmkn, line 182: Use a `m2m_changed` signal instead.

app/grandchallenge/reader_studies/models.py
```diff
@@ -451,10 +452,12 @@ def is_editor(self, user):
 
     def add_editor(self, user):
         """Adds ``user`` as an editor for this ``ReaderStudy``."""
+        follow(user=user, obj=self, actor_only=False, send_action=False)
         return user.groups.add(self.editors_group)
 
     def remove_editor(self, user):
         """Removes ``user`` as an editor for this ``ReaderStudy``."""
+        unfollow(user=user, obj=self, send_action=False)
```
jmsmkn, line 460: Use a `m2m_changed` signal instead.

jmsmkn (changes requested):
Looks good! Needs to change a little as we cannot guarantee that a particular code path will be used (I think that you found that problem in the tests!), and the same thing that we cannot hardcode a content type id.

app/grandchallenge/algorithms/models.py
```diff
@@ -182,10 +185,16 @@ def save(self, *args, **kwargs):
 
         if adding:
             self.set_default_interfaces()
+            for user in self.editors_group.user_set.all():
+                follow(user=user, obj=self, send_action=False)
```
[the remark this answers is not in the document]
amickan: oh, you're right, this is handled by the m2m_changed signal and can go.

jmsmkn (approved):

PR #1910, opened 2021-07-07
Adds list and create views for publications

### LINKED ISSUE #1909: Publication views
Users currently cannot add publications themselves and there is also no publication overview. It would be nice to add publication list and create views and some general publications stats to the list view (# of publications and # of citations).

amickan:
- Users can view a list of all publications
- Publications can be filtered by author names and publication year 
- If a publication is associated with a reader study, algorithm, archive etc., these related objects are linked underneath the publication, but only if the viewing user has permission to view these objects
- Only registered users can add new publications 

Closes #1909 

image

### DIFF
--- a/app/grandchallenge/publications/apps.py
+++ b/app/grandchallenge/publications/apps.py
@@ -14,9 +14,6 @@ def init_publication_permissions(*_, **__):
     assign_perm(
         f"{Publication._meta.app_label}.add_{Publication._meta.model_name}", g
     )
-    assign_perm(
-        f"{Publication._meta.app_label}.view_{Publication._meta.model_name}", g
-    )
 
 
 class PublicationsConfig(AppConfig):
--- a/app/grandchallenge/publications/apps.py
+++ b/app/grandchallenge/publications/apps.py
@@ -1,0 +1,26 @@
+from django.apps import AppConfig
+from django.conf import settings
+from django.db.models.signals import post_migrate
+
+
+def init_publication_permissions(*_, **__):
+    from django.contrib.auth.models import Group
+    from guardian.shortcuts import assign_perm
+    from grandchallenge.publications.models import Publication
+
+    g, _ = Group.objects.get_or_create(
+        name=settings.REGISTERED_USERS_GROUP_NAME
+    )
+    assign_perm(
+        f"{Publication._meta.app_label}.add_{Publication._meta.model_name}", g
+    )
+    assign_perm(
+        f"{Publication._meta.app_label}.view_{Publication._meta.model_name}", g
+    )
+
+
+class PublicationsConfig(AppConfig):
+    name = "grandchallenge.publications"
+
+    def ready(self):
+        post_migrate.connect(init_publication_permissions, sender=self)
--- a/app/grandchallenge/publications/models.py
+++ b/app/grandchallenge/publications/models.py
@@ -150,8 +150,4 @@ class Publication(models.Model):
 
     @property
     def authors(self):
-        authors = []
-        match = re.search(r".+?\.", self.citation)
-        for author in re.split(",", match.group()):
-            authors.append(re.split(r"\s", re.sub(r"^\s", "", author))[0])
-        return authors
+        return [a["family"] for a in self.csl.get("author", [])]
--- a/app/grandchallenge/publications/models.py
+++ b/app/grandchallenge/publications/models.py
@@ -147,3 +147,11 @@ class Publication(models.Model):
         citation = re.sub(r"^1\. ", "", citation)
 
         return clean(citation)
+
+    @property
+    def authors(self):
+        authors = []
+        match = re.search(r".+?\.", self.citation)
+        for author in re.split(",", match.group()):
+            authors.append(re.split(r"\s", re.sub(r"^\s", "", author))[0])
+        return authors
--- a/app/grandchallenge/publications/templates/publications/publication_list.html
+++ b/app/grandchallenge/publications/templates/publications/publication_list.html
@@ -1,0 +1,47 @@
+{% extends "base.html" %}
+{% load static %}
+{% load publication_extras %}
+{% load guardian_tags %}
+{% load url %}
+
+
+{% block title %}Publications - {{ block.super }}{% endblock %}
+
+{% block breadcrumbs %}
+    <ol class="breadcrumb">
+        <li class="breadcrumb-item active" aria-current="page">Publications</li>
+    </ol>
+{% endblock %}
+
+{% block content %}
+    <div class="d-flex justify-content-between">
+        <h2>Publications</h2>
+        <a class="btn btn-primary my-3" href="{% url 'publications:create' %}" role="button">Add a publication</a>
+    </div>
+        <ul class="list-group">
+            {% for publication in object_list %}
+                <li class="list-group-item p-0">
+                    <div class="py-2 pl-3">
+                        {{ publication.citation|safe }}
+                    </div>
+                    <div class="pt-1 pb-2 pl-3">
+                        <a class="btn btn-outline-primary btn-sm mb-1 mr-1" href="{{ publication.url }}"><i class="fa fa-link"></i></a>
+                        {% with publication|get_associated_objects as publication_objects %}
+                            {% if publication_objects %}
+                                {% for obj, obj_model in publication_objects.items %}
+                                    {% get_obj_perms request.user for obj as "object_perms" %}
+                                    {% if "view_"|add:obj_model in object_perms %}
+                                        <a class="btn btn-outline-primary btn-sm mb-1 mr-1" href="{{ obj.get_absolute_url }}">{{ obj }}</a>
+                                    {% endif %}
+                                {% endfor %}
+                            {% endif %}
+                        {% endwith %}
+                    </div>
+                </li>
+                {% empty %}
+                <li class="list-group-item">No publications found.</li>
+            {% endfor %}
+        </ul>
+    <br>
+    {% include "grandchallenge/partials/pagination.html" %}
+{% endblock %}
--- a/app/grandchallenge/publications/templates/publications/publication_list.html
+++ b/app/grandchallenge/publications/templates/publications/publication_list.html
@@ -31,16 +31,12 @@
                     </div>
                     <div class="pt-1 pb-2 pl-3">
                         <a class="btn btn-outline-primary btn-sm mb-1 mr-1" href="{{ publication.url }}"><i class="fa fa-link"></i></a>
-                        {% with publication|get_associated_objects as publication_objects %}
-                            {% if publication_objects %}
-                                {% for obj, obj_model in publication_objects.items %}
-                                    {% get_obj_perms request.user for obj as "object_perms" %}
-                                    {% if "view_"|add:obj_model in object_perms %}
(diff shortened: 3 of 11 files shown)


app/grandchallenge/publications/templates/publications/publication_list.html
```diff
@@ -0,0 +1,51 @@
...
+    </div>
+    <div class="d-flex justify-content-between my-3">
+        <h6 class="ml-1 mt-2">{{ num_publications }} publications | {{ num_citations }} citations </h6>
+        <a class="btn btn-primary" href="{% url 'publications:create' %}" role="button">Add a publication</a>
+    </div>
+        <ul class="list-group">
+            {% for publication in object_list %}
+                <li class="list-group-item p-0">
+                    <div class="py-2 pl-3">
+                        {{ publication.citation|safe }}
```
amickan, line 29: Since the citation is created in the backend, I thought that marking html characters as safe would be ok here, no chance for XSS vulnerability. Right? Alternatively, we could use `striptags` here and miss out on the italics for the journal name, which would also be ok, I guess.

app/grandchallenge/publications/apps.py
```diff
@@ -0,0 +1,26 @@
...
+    from grandchallenge.publications.models import Publication
+
+    g, _ = Group.objects.get_or_create(
+        name=settings.REGISTERED_USERS_GROUP_NAME
+    )
+    assign_perm(
+        f"{Publication._meta.app_label}.add_{Publication._meta.model_name}", g
+    )
+    assign_perm(
+        f"{Publication._meta.app_label}.view_{Publication._meta.model_name}", g
```
jmsmkn, line 18: The `view_publication` permission doesn't seem to be used, so I do not think that this assignment is necessary.

app/grandchallenge/publications/models.py
```diff
@@ -147,3 +147,11 @@ def ama_html(self):
         citation = re.sub(r"^1\. ", "", citation)
 
         return clean(citation)
+
+    @property
+    def authors(self):
```
jmsmkn, line 152: This might be quite slow, especially with how it is used in the filter. What we could do instead is extract this information out on import and then save it on the model. We can deploy it and see.

app/grandchallenge/publications/models.py
```diff
@@ -147,3 +147,11 @@ def ama_html(self):
...
+        authors = []
+        match = re.search(r".+?\.", self.citation)
+        for author in re.split(",", match.group()):
+            authors.append(re.split(r"\s", re.sub(r"^\s", "", author))[0])
```
jmsmkn, line 156: Rather than parsing the citation which could change it will be quicker and more robust to extract this from the `csl`. Something like `authors = [a["family"] for a in self.csl.get("author", [])]`

app/grandchallenge/publications/templates/publications/publication_list.html
jmsmkn: Good thinking, and great that you highlight this. We already have a template filter for that, `clean` that you can get from `{% load bleach %}` https://github.com/comic/grand-challenge.org/blob/2a995006b899c9a0f17e0f4ba862b796423c20a1/app/grandchallenge/core/templatetags/bleach.py#L15

app/grandchallenge/publications/templatetags/publication_extras.py
```diff
@@ -0,0 +1,43 @@
...
+from grandchallenge.algorithms.models import Algorithm
+from grandchallenge.archives.models import Archive
+from grandchallenge.challenges.models import Challenge, ExternalChallenge
+from grandchallenge.reader_studies.models import ReaderStudy
+
+register = template.Library()
+
+
+@register.filter
+def get_associated_objects(publication):
```
jmsmkn, line 12: Have you checked the number of requests that are being made for these views? My feeling is that this will produce a lot.

app/grandchallenge/publications/views.py
```diff
@@ -0,0 +1,50 @@
...
+from grandchallenge.subdomains.utils import reverse
+
+
+class PublicationList(LoginRequiredMixin, FilterMixin, ListView):
+    model = Publication
+    paginate_by = 50
+    filter_class = PublicationFilter
+
+    def get_queryset(self):
+        return super().get_queryset().order_by("-created")
```
jmsmkn, line 19: Probably missing a select related.

app/grandchallenge/publications/templates/publications/publication_list.html
```diff
@@ -0,0 +1,51 @@
...
+                    </div>
+                    <div class="pt-1 pb-2 pl-3">
+                        <a class="btn btn-outline-primary btn-sm mb-1 mr-1" href="{{ publication.url }}"><i class="fa fa-link"></i></a>
+                        {% with publication|get_associated_objects as publication_objects %}
+                            {% if publication_objects %}
+                                {% for obj, obj_model in publication_objects.items %}
+                                    {% get_obj_perms request.user for obj as "object_perms" %}
+                                    {% if "view_"|add:obj_model in object_perms %}
```
jmsmkn, line 37: This is probably going to make a lot of requests. See the Organization Detail view about how to use an ObjectPermissionsChecker to reduce these.

jmsmkn (changes requested):
Looks good! I'm worried about the number of queries that the list view will be making, and the author extraction can be made more robust.

app/grandchallenge/publications/models.py
amickan: Yes, this was how I initially tried it, but I believe that the csl does not only contain the authors of the paper but also authors of papers cited in the paper, and they are all under `authors`, as far as I could tell, so that's why I resorted to extracting the authors from the citation. I will take another look though.

amickan: Ah you were completely right, your query above works like a charm. Thank you!

app/grandchallenge/publications/views.py
amickan: There is nothing to prefetch from the publication model, unfrotauntely, but I will look into the template tag and how we can minimize queries there ;-)

app/grandchallenge/publications/templates/publications/publication_list.html
amickan: I believe I'm following the Organization Detail view as closely as possible already. The only difference is that in this case here, we need to construct the permission string based on the object we're trying to display, but this `"view_"|add:obj_model` doesn't do any extra requests, so it should be fine, I think/

app/grandchallenge/publications/templatetags/publication_extras.py
amickan: Yes, this produces 5 queries per publication, so that explodes quickly. I'm not immediately sure how to reduce the queries, but I will look into it.

app/grandchallenge/publications/templates/publications/publication_list.html
jmsmkn: I think the `get_object_perms` is the line which will do the extra calls right now.

app/grandchallenge/publications/views.py
jmsmkn: Ah ok I see, I wonder then if a better way is to fetch the publications, then prefetch the related archives/algorithms/reader studies/challenges, add an object permissions checker to those?

app/grandchallenge/publications/templates/publications/publication_list.html
jmsmkn: I see what is happening now, you're right! I suggested another approach in a different comment.

app/grandchallenge/publications/views.py
amickan: Actually, I realized that you can prefetch the `algorithm_set`, `archive_set` etc. on the list view. This plus adding a non-database-calling check for existence to the template filter before looping over possible related objects does the trick.

amickan: > Ah ok I see, I wonder then if a better way is to fetch the publications, then prefetch the related archives/algorithms/reader studies/challenges, add an object permissions checker to those?

I'm not sure I folllow. Isn't that what we do already? We're fetching the publication, then in the template filter function we fetch all related objects (now with as few queries as possible) and then we check the current users permission for each of those objects.

app/grandchallenge/publications/views.py
```diff
@@ -0,0 +1,61 @@
...
+from guardian.mixins import LoginRequiredMixin
+
+from grandchallenge.core.filters import FilterMixin
+from grandchallenge.publications.filters import PublicationFilter
+from grandchallenge.publications.forms import PublicationForm
+from grandchallenge.publications.models import Publication
+from grandchallenge.subdomains.utils import reverse
+
+
+class PublicationList(LoginRequiredMixin, FilterMixin, ListView):
```
jmsmkn, line 13: `LoginRequiredMixin` can be removed.

app/grandchallenge/publications/templates/publications/publication_list.html
```diff
@@ -0,0 +1,52 @@
...
+{% endblock %}
+
+{% block content %}
+        {% include "grandchallenge/partials/filters.html" with filter=filter filters_applied=filters_applied %}
+    <div>
+        <h2>Publications</h2>
+    </div>
+    <div class="d-flex justify-content-between my-3">
+        <h6 class="ml-1 mt-2">{{ num_publications }} publications | {{ num_citations }} citations </h6>
+        <a class="btn btn-primary" href="{% url 'publications:create' %}" role="button">Add a publication</a>
```
jmsmkn, line 24: Only show the button if the user has the `add_publication` permission (note this is the global permission, so checked differently to the object permissions).

app/grandchallenge/publications/views.py
```diff
@@ -0,0 +1,61 @@
...
+        )
+
+    def get_context_data(self, *, object_list=None, **kwargs):
+        context = super().get_context_data()
+        num_citations = 0
+        for pub in context["object_list"]:
+            try:
+                num_citations += pub.referenced_by_count
+            except TypeError:
+                continue
```
jmsmkn, line 39: There is a better way to do this in the database with an aggregate. Something like `self.get_queryset().exclude(referenced_by_count__isnull=True).aggregate(Sum("referenced_by_count"))["referenced_by_count__sum"]`

app/grandchallenge/publications/templates/publications/publication_list.html
```diff
@@ -0,0 +1,52 @@
...
+                <li class="list-group-item p-0">
+                    <div class="py-2 pl-3">
+                        {{ publication.citation|clean }}
+                    </div>
+                    <div class="pt-1 pb-2 pl-3">
+                        <a class="btn btn-outline-primary btn-sm mb-1 mr-1" href="{{ publication.url }}"><i class="fa fa-link"></i></a>
+                        {% with publication|get_associated_objects as publication_objects %}
+                            {% if publication_objects %}
+                                {% for obj, obj_model in publication_objects.items %}
+                                    {% get_obj_perms request.user for obj as "object_perms" %}
```
jmsmkn, line 37: This is going to hit the database twice for each publication, for each related object, regardless of whether the user has the permission. You can check by adding two publications, then adding those two publications to one object and viewing with a non-superuser (who skips all permissions checks). 

image

jmsmkn (approved):

PR #1912, opened 2021-07-08
Run build images in correct queue and fix error handling

### LINKED ISSUE #1911: Image import errors
Two errors have arisen recently:

- `build_images` is called directly when being used from flex algorithm inputs, a celery canvas should be used
- In `build_images`, if the upload fills the filesystem the temporary directory is not cleaned up, future jobs will be eaten until the worker is killed by the autoscaling rules

jmsmkn:
Celery tasks should not be called directly from within other tasks for routing reasons, this changes the setup to use a workflow.

Closes #1911


miriam-groeneveld (approved):
LGTM

PR #1914, opened 2021-07-08
Automatically set the users timezone

### LINKED ISSUE #1545: Show times in users local time, rather than UTC
**Describe the bug**
From the RTC Workshop Demo, Bram reported:

grafik

It reports the time in UTC (it was 10:40 CEST when he received the machine). I am not sure if this is actually correct, but I would expect that it the UTC time stamps are rendered to the local user timezone setting before showing it to the user. I find it unlikely that Bram uses GMT or UTC on his machine, but maybe in his account settings?

**To Reproduce**

Generate the join-challenge-rejected-message. 

**Expected behavior**
Us local time instead of UTC.

Don't know, I am reporting this on behalf of Bram. Likely does not matter.

jmsmkn:
This PR adds some JS to automatically set the users timezone based on `Intl.DateTimeFormat().resolvedOptions().timeZone`, which is then stored on the users session. The POST request is only made if the sessions timezone has changed. All datetimes will then be presented in the sessions timezone.

The core of the implementation follows https://docs.djangoproject.com/en/3.2/topics/i18n/timezones/#selecting-the-current-time-zone, but here we added validation and automated setting with JS.

Closes #1545


HarmvZ (approved):
Nice and compact feature! 
Reading the MDN docs on `Intl.DateTimeFormat().resolvedOptions().timeZone` it's possible that it returns `undefined` to indicate the default. Right now this would result in a HTTP 400 error which is maybe not what we want.

jmsmkn:
Yes, I was a little confused by the documentation for that. I think that only happens for IE11, which I'm ok with not supporting (2.6% of our users). I will keep an eye on it.

PR #1915, opened 2021-07-09
Fix all error messages being sent to user

jmsmkn:
All exceptions were being reported to the user, this limits them to ones we have reviewed.


miriam-groeneveld (approved):
LGTM!

PR #1923, opened 2021-07-14
Removes follows when forums, topics and posts are deleted

### LINKED ISSUE #1922: `get_follow_object_pk` errors out if `obj.follow_object` is `None`
Occurs when the follow object has been deleted and the follow is not cleaned up. See https://sentry.io/organizations/grand-challenge/issues/2511041483/?project=303639&query=is%3Aunresolved

amickan:
Closes #1922


jmsmkn:
You should also check where `follow_object.pk` is used and work out what to do when `follow_object` is `None`.

jmsmkn (approved):

PR #1925, opened 2021-07-16
Fixes author filter for publications

amickan:
- publications can now be filtered by authors 
- link to publication create view is now in the forms
- optimizes database calls for publication list view


jmsmkn (changes requested):
Ah, hang on a second, we need to update the migrations with the new help text.

app/grandchallenge/challenges/forms.py
```diff
@@ -23,6 +25,16 @@ class Meta:
...
+        help_texts = {
+            "publications": format_lazy(
+                (
+                    "The publications associated with this archive. "
+                    'If your publication is missing click <a href="{}">here</a> to add it '
+                    "and then refresh this page."
+                ),
+                reverse_lazy("publications:create"),
+            )
+        }
```
jmsmkn, line 37: Not necessary here as the field is not used

```suggestion
```

jmsmkn (approved):

PR #1932, opened 2021-07-23
Reduces db queries for challenge update celery task

amickan:


app/grandchallenge/challenges/tasks.py
```diff
@@ -1,37 +1,55 @@
 from celery import shared_task
 from django.core.mail import mail_managers
+from django.db.models import Count, Max
 from requests import exceptions, get
 
 from grandchallenge.challenges.models import Challenge, ExternalChallenge
 from grandchallenge.evaluation.models import Evaluation
+from grandchallenge.profiles.admin import User
```
jmsmkn, line 8: See https://docs.djangoproject.com/en/3.2/topics/auth/customizing/#referencing-the-user-model - instead import `get_user_model` and then replace instances of `User` with `get_user_model()`.

app/grandchallenge/challenges/tasks.py
```diff
@@ -1,37 +1,55 @@
...
-        kwargs = {
-            "cached_num_participants": c.participants_group.user_set.all().count()
-        }
-
-        challenge_results = Evaluation.objects.filter(
-            submission__phase__challenge=c, published=True
-        ).order_by("-created")
+    challenges = Challenge.objects.all()
+    evaluation_info = (
+        Evaluation.objects.filter(rank__gt=0, published=True)
```
jmsmkn, line 16:
```suggestion
        Evaluation.objects.filter(published=True)
```

Realised that rank is not required as we want all of the published results.

app/grandchallenge/challenges/tasks.py
```diff
@@ -1,37 +1,55 @@
...
+            cached_num_results=Count("submission__phase__challenge_id"),
+            cached_latest_result=Max("created"),
+        )
+    )
+    evaluation_info_by_challenge = {
+        str(v["submission__phase__challenge_id"]): v for v in evaluation_info
+    }
+    participant_counts = User.objects.values(
+        "groups__participants_of_challenge"
+    ).annotate(cached_num_participants=Count("pk"))
```
jmsmkn, line 28: Just for my understanding, does this work? It doesn't need to be `"groups__participants_of_challenge_id"` and `Count("groups__participants_of_challenge_id")`?

app/tests/challenges_tests/test_tasks.py
```diff
@@ -0,0 +1,32 @@
...
+    c2 = two_challenge_sets.challenge_set_2.challenge
+
+    _ = EvaluationFactory(
+        submission__phase__challenge=c1, method__phase__challenge=c1
+    )
+    _ = EvaluationFactory(
+        submission__phase__challenge=c2, method__phase__challenge=c2
+    )
+
+    with django_assert_num_queries(4) as _:
```
jmsmkn, line 22: 🎉 super!

app/grandchallenge/challenges/tasks.py
amickan: No, it doesn't. Actually in the other query, it would also suffice to call `submission__phase__challenge`. Is it clearer though to retrieve the id?

jmsmkn: The "Zen of Python" (`import this`) would tell us that "Explicit is better than implicit.", so I tend to err on the side of adding the id, but, if it works it works, I trust the Django developers to keep the API stable and if they do change the behaviour we would definitely hear about it before we hit issues.

jmsmkn (approved):

PR #1936, opened 2021-07-27
Challenge creation notifications

amickan:
- Adds follows for staff users to all site objects
- Adds notifications for staff users about newly created challenges

Closes #1936


app/tests/challenges_tests/test_models.py
```diff
@@ -85,3 +88,18 @@ def test_non_posters_notified(group):
 
     assert u.user_profile.has_unread_notifications is True
     assert p.user_profile.has_unread_notifications is False
+
+
+@pytest.mark.django_db
+def test_staff_users_notified_of_challenge_creation():
+    staff_user = UserFactory(is_staff=True)
+    call_command("create_staff_follows")
```
jmsmkn, line 96: I don't think that we can run a command every time a staff user is added. I really think that this is a case where an email is fine, maybe even better would be sending it to MANAGERS (https://docs.djangoproject.com/en/3.2/ref/settings/#managers).

If you want to go the notifications route could you keep the follows in sync this via signals instead? Also, what if a staff member is removed (both deleted or if they loose privileges)?

amickan: Right, I agree that an email is probably best after all. It also leave the notification template cleaner.

amickan: As for who to send the emails to: I assume you are the only manager currently, so then you would be the only one to receive those emails - is that ok with you?

jmsmkn: I think that's fine! We use a mailing list for this.

jmsmkn (approved):

PR #1937, opened 2021-07-27
Challenge participation request notifications

### LINKED ISSUE #1935: Challenge participation request notifications
Currently, admins of challenges receive emails about participation requests and requesters receive emails about acceptance and rejection to a challenge. It would be nice to turn these emails into notifications instead. 

- create follow: `admins` --> `challenge`
- create follow: `requester` --> `permission request`
- post-save of permission request: send `action` with `request creator` as `actor` and the `challenge` as `target`
- _Notification template_: print as is 
- when permission request is accepted / rejected: send `action` with `request object` as `actor` 
- _Notification template_: do not print actor, instead print "You are now a participant of `[request_object.challenge]`"

amickan:
- Admins of challenges now follow challenges and get notifications about new participation requests
- Users follow their participation request and get notified when its status changes 
- Challenge deletion and request deletion, deletes the associated follows
- When users are deleted, their follows get deleted as well 
- Follow can now be filtered on challenges as well 

Closes #1935


app/grandchallenge/participants/views.py
```diff
@@ -125,12 +114,4 @@ def get_success_url(self):
     def form_valid(self, form):
```
jmsmkn, line 114: This method can be deleted now as it's only passing through the super method.

app/grandchallenge/core/signals.py
```diff
@@ -93,6 +100,8 @@ def update_editor_follows(  # noqa: C901
             follow_objects.append(group.editors_of_archive)
         elif hasattr(group, "editors_of_readerstudy"):
             follow_objects.append(group.editors_of_readerstudy)
+        elif hasattr(group, "admins_of_challenge"):
+            follow_objects.append(group.admins_of_challenge)
```
jmsmkn, line 104: Just a note: we have to be really careful now with what objects the user follows as the challenge now has things that only the admin can see. Could you add a comment about this?

app/grandchallenge/participants/signals.py
```diff
@@ -15,7 +18,31 @@ def process_registration(
         RegistrationRequest.objects.filter(pk=instance.pk).update(
             status=instance.status
         )
+    elif created and instance.challenge.require_participant_review:
+        action.send(
+            sender=instance.user,
+            verb="requested access to",
+            target=instance.challenge,
+        )
+    if not is_following(instance.user, instance):
```
jmsmkn, line 27: Does the user who made the request get notified of acceptance and rejection? 

An aside: do they get notified for Archive/Algorithm/ReaderStudy acceptance and rejection?

jmsmkn: If I'm reading your tests correctly I think that they do, which is what we want :-)

amickan: Yes, the requester gets notified of acceptance / rejection of the request, both for challenges as well as for archives/algorithms/reader studies ;-) That's because they follow the request object, and whenever the status changes, we send an action with the request object as sender and without a target / action object. Actions with only a sender in turn trigger notifications for all followers of the sender.

app/grandchallenge/core/signals.py
amickan: So you mean a comment along the lines of "only admins of challenges should follow a challenge" to make sure we don't accidentally create follows for participants at some later point in time?

amickan:
I added some tests to increase coverage and found a bug in how we used to delete related Follows objects when archives, challenges, algorithms etc. get deleted. Identifying Follow objects for deletion based only on the follow_object pk is not enough, content type needs to be taken into account as well. Fixed this in all places where that was necessary.

jmsmkn (approved):

PR #1946, opened 2021-08-02
Display correct input and output paths when uploading algorithm images

nlessmann:
Input and output paths are not correctly displayed at the moment:
image


jmsmkn (approved):
Good catch, thanks!

PR #1951, opened 2021-08-06
Creator filter for reader study answers

### LINKED ISSUE #1950: Filter reader study answers by username
It would be handy if reader study answers could be filtered by username ("creator")

amickan:
- adds the option to filter reader study answers by creator username

Closes #1950

### DIFF
--- a/app/grandchallenge/reader_studies/filters.py
+++ b/app/grandchallenge/reader_studies/filters.py
@@ -11,11 +11,8 @@ class ReaderStudyFilter(TitleDescriptionModalityStructureFilter):
 
 
 class AnswerFilter(FilterSet):
-    reader_study = CharFilter(
-        field_name="question__reader_study__title", label="reader study title"
-    )
     creator = CharFilter(field_name="creator__username", label="username")
 
     class Meta:
         model = Answer
-        fields = ("creator", "reader_study")
+        fields = ("creator", "question__reader_study")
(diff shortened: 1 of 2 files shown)


app/grandchallenge/reader_studies/filters.py
```diff
@@ -1,7 +1,21 @@
...
 
 
 class ReaderStudyFilter(TitleDescriptionModalityStructureFilter):
     class Meta(TitleDescriptionModalityStructureFilter.Meta):
         model = ReaderStudy
+
+
+class AnswerFilter(FilterSet):
+    reader_study = CharFilter(
+        field_name="question__reader_study__title", label="reader study title"
```
jmsmkn, line 15: This breaks the existing API, we should only do that when we must. Please keep the key as `question__reader_study` and use the reader studies primary key as the value, I think that there should be a Related Model Filter rather than a `CharFilter`.

amickan: There is actually a UUIDFilter that should do the trick, right? And so we want to keep the name of the field as "question__reader_study" as well, or is it fine to simplify to reader study?

amickan: ah, I see now that we should keep the name at question__reader_study

jmsmkn (approved):

PR #1953, opened 2021-08-09
Markdown editor with preview and editor tabs

HarmvZ:
https://user-images.githubusercontent.com/2001094/128737744-7466182a-a223-42da-a281-0b5f5dc9f468.mp4

### DIFF
--- a/app/grandchallenge/core/templates/markdownx/widget.html
+++ b/app/grandchallenge/core/templates/markdownx/widget.html
@@ -1,6 +1,15 @@
-<div class="markdownx row">
-
-    <div class="col-md-6">
+<div>
+    <ul class="nav nav-tabs" id="myTab" role="tablist">
+        <li class="nav-item">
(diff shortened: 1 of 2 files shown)


jmsmkn (changes requested):
I don't think that this will work with multiple widgets in the same form. Can you update all ids, hrefs, and arias to include `{{ widget.attrs.id }}`?

jmsmkn:
Also can you check how this is in the admin?

HarmvZ:
> I don't think that this will work with multiple widgets in the same form. Can you update all ids, hrefs, and arias to include {{ widget.attrs.id }}?

Good catch, that did not work indeed. Fixed it now.

The admin also did not work as it should because it was missing the vendor bootstrap js files. Added those also. The spacing/indentation is a bit strange, but I think that's not really important. It looks like this:

image

jmsmkn (approved):

PR #1954, opened 2021-08-10
Challenge evaluation notifications

### LINKED ISSUE #1908: Challenge evaluation-related notifications
Currently, users and challenge admins receive emails about failed and successful challenge submissions. It would be nice to turn those into notifications. 

**Challenge evaluation success and failure**

- create follow: `submission creator` --> `submission`
- create follow: `admins of challenge` --> `submission`
- on success or failure: send `action` with `evaluation` as `actor` and `submission` as `target`
- _Notification template:_
  - on success, for admins: There is a new result `[evaluation]` for `[submission.phase.challenge]` from `[submission.creator]`
  - on success, for creator: You successfully submitted a new result `[evaluation]` for `[submission.phase.challenge]`
  - on failure, for admins: The submission `[evaluation]` to `[submission.phase.challenge]` from `[submission.creator]` failed.
  - on failure, for creator: Your submission `[evaluation]` to `[submission.phase.challenge]` failed.

**Missing method email**

- send action with `submission creator` as `actor` and `submission` as `target` --> this only notifies admins (followers of the target other than the actor)
- _Notification template:_ `[submission]` could not be evaluated because of missing evaluation container.

amickan:
- replaces evaluation failure / success emails with notifications
- replaces missing method email with notification

Closes #1908


app/grandchallenge/evaluation/models.py
```diff
@@ -583,6 +581,12 @@ def save(self, *args, **kwargs):
 
         if adding:
             self.assign_permissions()
+            followers = list(self.phase.challenge.get_admins())
+            followers.append(self.creator)
+            for user in followers:
+                follow(
+                    user=user, obj=self, actor_only=False, send_action=False,
```
jmsmkn, line 588: Does this create 1 follow per submission? Some challenges have tens of thousands of submissions, to unsubscribe from these notifications would the user have to unsubscribe from each one?

amickan: Oh, hm, yes this creates 1 follow per submission and users would have to unsubscribe from each one (although selecting multiple follows and bulk delete is possible). I'm not sure if unsubscribing and showing these follows to users makes sense though. Maybe it's better to hide them from the subscription list view. Ultimately they are just used to track the status of the submission and to inform users of status changes.

amickan: If I see it correctly, the status only gets updated to failed or succesful once per submission, correct? In that case, we could also delete the follow after the status change (and after the action/notification has been sent).

jmsmkn: Yes - the state actually comes from the evaluation that is foreign keyed to the submission, you can have many evaluations per submission but this is rare. The use case that must be implemented here is: as a challenge administrator, I want to opt in and opt out to all evaluation notifications about a particular phase that I am an administrator for. So, if there is a new evaluation, this admin should not be notified if they have opted out. I would have thought that following a phase could do this.

amickan: For admins, that's actually a really good idea, but that wouldn't work for the submission creator. The submission creator should only be notified about status updates for their own submissions. So then for the users we'd still need the submission follows, or am I missing something?

amickan: A solution would be to adjust the notification creation signal and have a special rule for when the action target is a phase (send to admins of challenge and creator of submission). It's not as elegant though and would make the notification creation signal a bit less intuitive yet. But we'd have way less follow objects.

amickan: I pushed the alternative implemention with Phase follows. This is still missing a management command to make all existing challenge admins followers of existing challenge phases, but otherwise, this is what it could look like if we use phase rather than submission follows.

app/grandchallenge/core/signals.py
```diff
@@ -104,7 +104,10 @@ def update_editor_follows(  # noqa: C901
             follow_objects.append(group.editors_of_readerstudy)
         elif hasattr(group, "admins_of_challenge"):
             # NOTE: only admins of a challenge should follow a challenge
+            # and it's phases
```
MikeOverkamp-diag, line 107: `its phases`, no apostrophe :)

MikeOverkamp-diag (dismissed):
LGTM, I just have a very minor nitpicky remark.

app/grandchallenge/core/signals.py
amickan: good catch! ;-)

MikeOverkamp-diag (approved):

PR #1957, opened 2021-08-11
Allow blog editors to add authors

### LINKED ISSUE #1955: Add option to add authors to blog posts
Currently, adding extra authors to blog posts can only be done through the admin interface. It would be nice to add the option to add authors in the edit form of the blog post.

amickan:
This allows blog post authors to add other authors. It would make collaborating on a blog post easier for users, although, of course, simultanously working on the blog post does not work. 

This only adds the possibility to add authors, not to remove authors. 

Closes #1955

### DIFF
--- a/app/grandchallenge/blogs/forms.py
+++ b/app/grandchallenge/blogs/forms.py
@@ -1,8 +1,11 @@
 from django import forms
+from django.contrib.auth import get_user_model
+from django.contrib.auth.models import Group
 from django.core.exceptions import ValidationError
 from django_select2.forms import Select2MultipleWidget
 from guardian.utils import get_anonymous_user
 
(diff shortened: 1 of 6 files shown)


jmsmkn (changes requested):
Unfortunately this will not work in production - this will add all users as options to the field which only works up to ~1000 users, we have a lot more than that. See the approach taken when adding users to archive/reader studies/algorithm/organisation groups for how to tackle this. I remember there being some problems when adding that so that is why the option is not there.

jmsmkn (commented):
Sorry forgot to press save on review. Would be good to have a test for the permissions just as a sanity check that this is all hooked up but this LGTM.

PR #1959, opened 2021-08-13
Fix algorithm job status check

nlessmann:
When running a job, they currently always end up as "Errored":

image

From checking the API requests that this page makes, it seems that the job status message that is returned via the API becomes "Parsing Outputs" when the job has finished. Since the Javascript expects "Parsing", it displays the "Errored" message.


jmsmkn (approved):
Ah, this was bugging me, thanks! I think the CI failures are spurious so running those again.

PR #1961, opened 2021-08-17
1878 add reader metrics

Kicksta:


Kicksta:
@MikeOverkamp-diag, I continued working on it but could use your help here… I just pushed the latest version.
-	In `conftest.py`, `_reader_study_with_gt` I defined questions and answers which represent (if I understand correctly) the ground truth values. They could not all be set to True, because `sklearn.roc_auc_score` needs two classes (otherwise aucroc is always 1). So I used a formula here (`(i + j) % 2`) resulting in `[0,1,1,0,0,1]`.
-	In `test_models.py` I define again answers which are float (`1 / (i + j + 1)`) resulting in `[1, 1/2, 1/2, 1/3, 1/3, 1/4]`. 
-	These two should result in a roc_auc_score of 0.556. Currently it outputs `score__sum` and `score__avg` of 0.0.
-	I wonder if `score__sum `and `score__avg `make sense as you can not calculate an auc for an individual case. And that makes me wonder if it will work with the whole setup as is, as it is based on case level scores, right?
Could you have a look and see how we should continue?

app/grandchallenge/reader_studies/models.py
```diff
@@ -993,9 +1001,7 @@ def calculate_score(self, answer, ground_truth):
         else:
             ans = [answer]
             gt = [ground_truth]
-        return self.SCORING_FUNCTIONS[self.scoring_function](
-            ans, gt, normalize=True
```
jmsmkn, line 997: Removing `normalize=True` is going to be a problem with the existing method. To keep that use a partial:

``` 
ScoringFunction.ACCURACY: partial(accuracy_score, normalize=True)
```

where it is defined above (see https://docs.python.org/3/library/functools.html#functools.partial)

Flipping `ans` and `gt` here was the right thing to do, that was a bug :-)

Kicksta: Normalize: default of accuracy score has normalize is True. That's why I thought it could be left out. sklearn.metrics.accuracy_score(y_true, y_pred, *, normalize=True, sample_weight=None). Is it then still necessary to add it as a partial?
Bug: yeah I thought so. Of course for accuracy it doesn't effect the result but for auc it does.

jmsmkn: Ah great, didn't see that and thought that because it was set that it wasn't the default, then indeed it can be left out and this block is fine, thanks!

MikeOverkamp-diag:
@Kicksta You are right, the scoring is currently based on a case by case basis, which means the current setup does not make much sense for AUC scoring. I think some larger changes to the entire setup are needed to make this sensible, as for a simple accuracy score it makes sense to have a score per case and having those scores can give reader study editors some useful statistics on e.g. difficult or ambiguous cases or questions. So I don't think omitting these individual scores would be a viable solution. Do you have any thoughts on this?

Kicksta:
I have tried to write out the scenarios.
- We have a set of readers, cases and questions.
- Questions may partly have a ground truth or not and can have different metrics. 
- I try three types of metrics: accuracy, auc roc, and error (if it's absolute, squared whatever doesn't matter that much for now.)

What does a reader want:
At the end of the reader study the reader want to know it's score for:
- all cases combined
- for each question separately (if different metrics they can't be combined to a single score)
- this is possible to do for acc, auc, and error by calculating the score on the all cases

What does a reader study editor want:
At the end of the reader study done by multiple readers he/she wants to know the score for:
- same as above: per reader, per question, all cases
+
- all readers combined
- for each case and question separately 
- this is possible for acc and error. for auc this is not possible as with a single case there is only one ground truth value. auc needs both 0s and 1s in the truth to be calculated.

Are there more scenarios? A score per case while doing the (teaching) reader study, or is then the right answer just given to draw your own conclusion?

Writing it up like this, it seems like two pathways are necessary to describe both, which includes some if-statements on what is returned with what metric. This problem is not as easy as it seemed...

jmsmkn:
I think this can only happen with a refactoring of the reader study models to support this, it's simply not possible with the current data model.

PR #1962, opened 2021-08-17
1960 products blog

Kicksta:
It all seems to work!

Two notes why this is still a draft:
1)  The only thing that changed for g-c bloggers is that they have to choose the location for publication of the blog --> either grand challenge or AI for radiology. Maybe it is better to automatically set this field in the view? Depending from where you create a blog it will be set to g-c or air. I tried the following, in views.PostCreate but returns all Posts so that doesn't work.
```
blog = Post.objects.get()
blog.publishing_location= 'gc' OR 'air'
blog.save()
```
Do you have an alternative idea?

2)
What I couldn't test and what needs a decision is how to deal with the tags: where are the tag-choices defined? How should we differentiate those between grand-challenge and air? @MikeOverkamp-diag will discuss this with @jmsmkn


jmsmkn:
Rather than doing this with a choice field why not using the existing tags? AIR can then be a filtered view of any Post that has the `slug=air` tag. We can then choose to exclude those from the grand challenge side, but I do not see the harm of putting them there too.

> where are the tag-choices defined?

They are defined as they are needed in the Django admin, these are the current ones:

image

Kicksta:
That sounds like another option.

Would it interfere with adding more tags specific for AIR? For now I would like to introduce three other tags: News, Research, Blog (and then we could just use the Products tag that already exists to differentiate G-C from AIR blogs). Is it okay to have them added to the current list of Tags or would that interfere with the g-c tags? 

Otherwise we would have to create some sort of hierarcy. If AIR/Products --> certain tags. But if AIR /Products is a tag itself that becomes a bit difficult, so then I guess we are back to the choices field.

PS: Someday, it would be cool to have a tag for each product/company with manytomany fields and automatically add news items on certain products to the company_detail/product_detail pages. But maybe better to leave that for future ambitions.

jmsmkn:
No totally fine to add those, I have done it now. Fine to use products, let me know if you decide to use that. No need for a hierarchy.

For your future idea you could add a companies m2m on the Post model to link those things. The same could be done for products, challenges, reader studies etc etc.

MikeOverkamp-diag:
Let's go with James` suggestion to use the tag to determine what blog posts go on the air blog page. Besides that, this looks good to me.

app/grandchallenge/products/sitemaps.py
```diff
@@ -22,3 +23,14 @@ def items(self):
 
     def lastmod(self, item: Product):
         return item.modified
+
+
+class PostsSitemap(SubdomainSitemap):
```
jmsmkn, line 28: This class doesn't do anything as it's not registered, and even if it was it just duplicates what is in `grandchallenge.blogs.sitemaps.PostsSitemap`. It can be removed.

app/grandchallenge/products/forms.py
```diff
@@ -26,3 +29,13 @@ class ProjectAirFilesForm(SaveFormInitMixin, forms.ModelForm):
     class Meta:
         model = ProjectAirFiles
         fields = ["title", "study_file"]
+
+
+class ProductsPostUpdateForm(PostUpdateForm):
+    class Meta(PostForm.Meta):
```
jmsmkn, line 35: The inheritance of the Meta class here is probably not you want as it's skipping `PostUpdateForm.Meta` and going directly to the super class of that. Probably what you want to do here is replace all insances of `PostForm.Meta` with `PostUpdateForm.Meta` and then the field and widget inheritance becomes much simpler.

app/grandchallenge/products/views.py
```diff
@@ -241,10 +260,28 @@ def get_success_url(self):
         return reverse("products:product-list")
 
 
-class ProjectAirForm(PermissionRequiredMixin, CreateView):
-    template_name = "products/projectairfiles_form.html"
-    form_class = ProjectAirFilesForm
-    permission_required = f"{ProjectAirFiles._meta.app_label}.add_{ProjectAirFiles._meta.model_name}"
+class ProductsPostCreate(PostCreate):
```
jmsmkn, line 263: As is, these 3 new views need tests for permissions at the very least as you can edit any post with these views, and we need to ensure that a change made in the future here will not open up editing posts to users that do not have the correct permissions.

app/grandchallenge/products/views.py
```diff
@@ -241,10 +260,28 @@ def get_success_url(self):
...
+    )
+
+
+class ProductsPostDetail(PostDetail):
+    template_name = "products/post_detail.html"
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        obj = context["object"]
+        is_news = obj.tags.all().filter(name="News").exists()
```
jmsmkn, line 280: Filtering on `name` is fragile, use `slug` instead with the appropriate value.

app/grandchallenge/blogs/models.py
```diff
@@ -60,7 +65,10 @@ def save(self, *args, **kwargs):
         super().save(*args, **kwargs)
 
     def get_absolute_url(self):
-        return reverse("blogs:detail", kwargs={"slug": self.slug})
+        if self.tags.filter(name="Products").exists():
```
jmsmkn, line 68: Use `slug` to filter, not `name`.

Kicksta: fixed

app/grandchallenge/products/views.py
Kicksta: fixed

app/grandchallenge/products/views.py
Kicksta: I have added tests similar to the blogs_tests. I haven't really figured out completely yet how tests work, please have a look if it looks all right to you. And probably the CI checks will provide me a hint soon too.

app/grandchallenge/products/forms.py
Kicksta: fixed

app/grandchallenge/products/sitemaps.py
Kicksta: removed

Kicksta:
After merging master into the branch it seems like a lot of tests are failing. Can't imagine that's because of the 1 test I included. Any hints on what may be going on?

jmsmkn:
Check `ProductsPostUpdateForm` - it is failing as the `fields` attribute is defined twice. This is checked by Django on startup so fails every test.

jmsmkn (approved):

PR #1963, opened 2021-08-18
Add icon to show page visibility

jmsmkn:
This PR adds a lock icon to show what pages are visible in the menu with tooltips. Uses danger for admin only views, and has a blank placeholder for alignment (maybe we can add a page icon option later). 

image


amickan (approved):

amickan:
LGTM! I like the placeholder and don't think a page icon is necessary. This way at least, the icons pop out and are clear. :-)

PR #1964, opened 2021-08-18
Handle missing follow_obj on subscription creation

amickan:


app/grandchallenge/notifications/forms.py
```diff
@@ -33,7 +33,10 @@ def clean_user(self):
     def clean(self):
         cleaned_data = super().clean()
 
-        obj_id = cleaned_data["object_id"]
+        try:
+            obj_id = cleaned_data["object_id"]
+        except KeyError:
+            raise ValidationError("You cannot create this subscription!")
```
jmsmkn, line 39: It might be nicer to set `self.fields["object_id"].required = True` in `__init__`. Also for the other fields, I think that `actor_only` is the only optional one, right?

amickan: Yes,I can do that, but only doing that won't result in a nice error message, as far as I can tell. It will error out as it did before with the KeyError, right?

jmsmkn: Did you try it? What usually happens is that the errors should get added to the field and the browser will not even let the user submit the form.

amickan: Yes, I tried, and it let's me submit and then throws the KeyError. So the automatic validation doesn't seem to work, maybe because all the fields are hidden?

amickan: Looks like the error are added to the field, but this does not stop the clean() method from running, which then throws the KeyError.

jmsmkn: Ok, good to know, thanks!

jmsmkn (approved):

PR #1973, opened 2021-08-23
Allow NUMBER and CHOICE answer types to be null when not required

HarmvZ:
Currently questions with `AnswerType` `NUMBER` and `CHOICE`  are not allowed to be null. This means that setting the `required` field to `False` on the question will not have the intended effect. 

This PR enables users to "skip" these non-required questions and make a reader study work like described in the table in https://github.com/DIAGNijmegen/rse-cirrus-core/pull/1432. By allowing the answer to be `null` (JSON schema)/ `None` (python) when the question is not required.

### DIFF
--- a/app/grandchallenge/reader_studies/models.py
+++ b/app/grandchallenge/reader_studies/models.py
@@ -1279,7 +1279,7 @@ class Answer(UUIDModel):
         valid_options = question.options.values_list("id", flat=True)
         if question.answer_type == Question.AnswerType.CHOICE:
             if not question.required:
-                valid_options.append(None)
+                valid_options = (*valid_options, None)
             if answer not in valid_options:
                 raise ValidationError(
                     "Provided option is not valid for this question"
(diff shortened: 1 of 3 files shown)


jmsmkn:
Question and a comment

- How are choices currently stored? Are they integers or strings? If strings, we should not allow null and represent this with a blank string instead.
- I think that the `ANSWER_TYPE_SCHEMA` will need to be updated to reflect that we're allowing null values

HarmvZ:
Choices are stored as integers (pk of `CategoricalOption` model).

The `ANSWER_TYPE_SCHEMA` does not have to be updated because it is dynamically changed to allow null here:
https://github.com/comic/grand-challenge.org/blob/fd9b891d2a240567760eab03208b0f2b54f8f707/app/grandchallenge/reader_studies/models.py#L1091-L1114

MikeOverkamp-diag (approved):

PR #1974, opened 2021-08-24
Failed job notifications

### LINKED ISSUE #1907: Failed algorithm job notifications
Currently, users receive emails about failed algorithm jobs. It would be nice to turn those into notifications instead.

Workflow: 
- create follow: `job creator` and `admins of algorithm` --> `job`
- replace message code in `send_failed_jobs_email` with action creation
- send action with `job` as actor 
- this means sending seperate notifications for failed jobs and excluded images, both in one is too long for a notification

amickan:
- turns failed job emails into notifications
- users can seperately unsubscribe from job-related notifications and other algorithm-related notifications 

Closes #1907

### DIFF
--- a/app/grandchallenge/algorithms/models.py
+++ b/app/grandchallenge/algorithms/models.py
@@ -2,7 +2,7 @@ import logging
 from datetime import timedelta
 
 from actstream import action
-from actstream.actions import follow
+from actstream.actions import follow, is_following
 from actstream.models import Follow
 from django.conf import settings
 from django.contrib.auth.models import Group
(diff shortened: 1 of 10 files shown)


MikeOverkamp-diag (approved):
I think this is ok for now, but it really highlights the importance of using something custom for the notifications.

jmsmkn:
@amickan I ended up having to split the notification method up into two in #1977 due to an unrelated issue with SQS. Noticed that the verbs for the notifications are very long and more like an error message, is that really what we want? I don't think that they will look good on the notifications page.

PR #1979, opened 2021-08-24
Verified algorithm creation

jmsmkn:
Only allow verified users to create algorithms or upload containers.


MikeOverkamp-diag:
Just reviewed this, LGTM. Only thing I would have liked is an extra test to ensure that unverified users are not allowed to create algorithms or upload containers.

jmsmkn:
Agreed! Was implemented in an emergency, it's on my TODO list to revisit. Given the number of complaints we're getting today it is working though :-)

MikeOverkamp-diag:
Figured as much. Code looks fine and as long as it's working the extra tests are just a nice bonus for now.

PR #1980, opened 2021-08-25
Reduces db calls from notification and follow list views

amickan:
Prefetching of generic foreign keys was broken, fixed now, which reduces the number of db queries by a lot.
Also refactored our custom prefetch function a bit while at it. Once the new notification model is out, this can be further refactored.


jmsmkn (dismissed):
LGTM, I thought that you added a test last time to catch this if it broke?

MikeOverkamp-diag:
@amickan can you have a look at the test James mentioned?

amickan:
> @amickan can you have a look at the test James mentioned?

Yes, I'm on it.

PR #1984, opened 2021-08-27
Small improvements to retina API enpoints

HarmvZ:
- Only loads MetaIO images for ArchiveAPI endpoints so that failing pathology images are not loaded accidentally.
- Adds support for 16bit images for thumbnail endpoint.
- Adds a script for setting up an Archive structure containing Patients, studies and ophthalmology images for quick setup of dev environment for CIRRUS ophthalmology. I used to do this with a script for the retina_importers app but this got removed a while back.


app/tests/retina_api_tests/test_views.py
```diff
@@ -143,6 +145,49 @@ def test_with_data(
...
+        client,
+        archive_patient_study_image_set,
+        django_assert_num_queries,
+    ):
+        cache.clear()
+        user = get_user_from_str("retina_user")
+        archive_patient_study_image_set.archive1.add_user(user)
+        pk = archive_patient_study_image_set.study113.pk
+
+        with django_assert_num_queries(22):
```
jmsmkn, line 188: Would prefer to use `max_num_queries` instead as having the exact number can be quite fragile.

jmsmkn (approved):

PR #1986, opened 2021-08-30
1985 air add archive files

Kicksta:
issue 1985, add option to archive protocol files, and display separately.


app/grandchallenge/products/models.py
```diff
@@ -158,3 +158,4 @@ def get_absolute_url(self):
 class ProjectAirFiles(models.Model):
     title = models.CharField(max_length=150)
     study_file = models.FileField(upload_to=get_pdf_path)
+    archive = models.BooleanField(default=False)
```
jmsmkn, line 161: Could you add some help text to explain what this field is intended to help as a future developer? We also have an Archive model which it can be confused with.

Kicksta: done

Kicksta: oops forgot to migrate. will do!

jmsmkn (approved):

PR #1987, opened 2021-08-30
Add view_logs permission

jmsmkn:
Adds an explicit view_logs permission so that non-algorithm editors can view the logs for a job.


miriam-groeneveld (approved):
LGTM

PR #1988, opened 2021-09-01
New notification model

amickan:
This is the first step towards implementing a new notification model, as described [here](https://github.com/DIAGNijmegen/rse-roadmap/issues/81).

This intermediate step keeps the `action` field of the notification model as an optional field. After all old notification have been transferred to the new format via the `transfer_notification` management command, the `action` field will be deprecated and ultimately deleted from the model.


app/grandchallenge/notifications/models.py
```diff
@@ -1,27 +1,120 @@
...
     )
 
     read = models.BooleanField(default=False, db_index=True)
 
+    context_class = models.CharField(
+        max_length=10,
+        blank=True,
+        null=True,
+        help_text="Bootstrap contextual class to style notification list items.",
+    )
```
amickan, line 69: This is currently only used for forum posts of type 'accouncement'. They get the Bootstrap contextual class `info` and will have a blue background color in the notification list. This actually did not work previously. I'm not sure how important we think this is, but a field like this doesn't hurt I guess, even if we decide to drop the blue background color for the announcements. Can be handy for alerts at some point.

app/grandchallenge/notifications/templates/notifications/notification_list.html
```diff
@@ -45,60 +46,65 @@ <h2>Notifications</h2>
...
-                                    {% if object.action.verb == "failed" %}
-                                       The <a href="{{ object.action.actor.submission.get_absolute_url }}">submission</a> from {{ object.action.actor.submission.creator|user_profile_link }} to <a href="{{ object.action.target.challenge.get_absolute_url }}">{{ object.action.target.challenge.short_name }}</a> failed. <span class="text-truncate font-italic text-muted align-middle mx-2 ">{{ object.action.actor.error_message }}</span>
-                                    {% else %}
-                                        There is a new <a href="{{ object.action.actor.submission.get_absolute_url }}">result</a> for <a href="{{ object.action.target.challenge.get_absolute_url }}">{{ object.action.target.challenge.short_name }}</a> from {{ object.action.actor.submission.creator|user_profile_link }}.
+                            {% if object.type != "GENERIC" %}
+                                {% print_notification notification=object user=request.user as current_notification %}
+                                {{ current_notification }}
+                            {% else %}
```
amickan, line 52: Everything within the else statement here can go after we are done transferring the old notificiations to the new format. The template will be much cleaner!

app/grandchallenge/notifications/views.py
```diff
@@ -89,13 +88,14 @@ class NotificationList(
...
             .get_queryset()
             .select_related(
-                "action__actor_content_type",
-                "action__target_content_type",
-                "action__action_object_content_type",
+                "actor_content_type",
+                "target_content_type",
+                "action_object_content_type",
                 "user",
             )
```
amickan, line 100: This means that for the time that we still display notifications through the action field (i.e. the old way), there will again be lots of queries. Only until we transfer all old notifications to the new format though, so this shouldn't be too bad. I had to change this already because for the new notifications, we no longer have nested generic foreign keys, and displaying them would fail (i.e. the prefetch function would throw an error).

app/grandchallenge/notifications/models.py
```diff
@@ -37,3 +130,282 @@ def _assign_permissions(self):
...
+                context_class=context_class,
+            )
+
+    def print_notification(self, user):  # noqa: C901
+        if self.type == NotificationType.NotificationTypeChoices.FORUM_POST:
+            return format_html(
+                "{} {} {} in {} {}.",
+                user_profile_link(self.actor),
+                self.verb,
+                mark_safe(
```
jmsmkn, line 207: You shouldn't need `mark_safe` in here. Add any safe markup to the string in the first argument to `format_html`.

amickan: Actually, I think mark_safe is called by format_html by default, so I don't need to add mark_safe anywhere, I guess, right? But I can just replace all the mark_safe instances with format_html then.

app/grandchallenge/notifications/models.py
```diff
@@ -1,27 +1,119 @@
...
+        related_name="notification_actor",
+        on_delete=models.CASCADE,
+        db_index=True,
+    )
+    actor_object_id = models.CharField(
+        max_length=255, db_index=True, blank=True, null=True
+    )
+    actor = GenericForeignKey("actor_content_type", "actor_object_id")
+
+    verb = models.CharField(
```
MikeOverkamp-diag, line 84: I'm being a bit nitpicky here, but `verb` seems like it does not really cover the purpose of this field. If we're recreating the model anyway, I'd prefer to rename this field to something like `message`.

app/grandchallenge/core/management/commands/transfer_notifications.py
```diff
@@ -0,0 +1,201 @@
...
+                    notification.action.actor_content_type
+                )
+                notification.action_object_object_id = (
+                    notification.action.actor_object_id
+                )
+                notification.verb = notification.action.verb
+                notification.save()
+                num_notifications += 1
+        self.stdout.write(
+            self.style.SUCCESS(f"{num_notifications} notifications processed")
```
MikeOverkamp-diag, line 200: Maybe it would be nice to ad the count for all `Notification` objects here. That way we can spot if any notifications were somehow missed in the conversion step.

app/grandchallenge/notifications/models.py
```diff
@@ -1,27 +1,119 @@
...
+from grandchallenge.profiles.templatetags.profiles import user_profile_link
+from grandchallenge.subdomains.utils import reverse
+
+
+class NotificationTypeChoices(models.TextChoices):
+    """Notification type choices."""
+
+    GENERIC = "GENERIC", _("Generic")
+    FORUM_POST = "FRM-POST", _("Forum post")
+    FORUM_POST_REPLY = "FRM-RPL", _("Forum post reply")
```
MikeOverkamp-diag, line 21: Any reason to use abbreviations here, but full names later on (e.g. in `MISSING-METHOD`)?

app/grandchallenge/notifications/models.py
MikeOverkamp-diag: I don't see any harm in this. It might be nice to have the bootstrap class as part of the actual logic, but this works as well and might make things a bit cleaner and more flexible in case we get more notification types that require their own classes.

app/grandchallenge/notifications/models.py
```diff
@@ -37,3 +129,274 @@ def _assign_permissions(self):
         assign_perm("view_notification", self.user, self)
         assign_perm("delete_notification", self.user, self)
         assign_perm("change_notification", self.user, self)
+
+    @staticmethod
+    def send(type, **kwargs):  # noqa: C901
+        verb = kwargs.pop("verb")
+        try:
+            actor = kwargs.pop("actor")
```
MikeOverkamp-diag, line 137: You can get rid of all these `try.. except` statements by providing a default for `pop`, e.g. `actor = kwargs.pop("actor", None)`.

app/grandchallenge/notifications/templates/notifications/notification_list.html
MikeOverkamp-diag: That's really great!

MikeOverkamp-diag (changes requested):
Looks good! A couple of small remarks.

app/grandchallenge/notifications/models.py
amickan: ah that's a lot better, thanks!

app/grandchallenge/notifications/models.py
amickan: No good reason, I'll change them all to full names.

app/grandchallenge/notifications/models.py
amickan: Ok, that makes sense actually.

MikeOverkamp-diag (changes requested):
Still a coupe of `verb`s left in the management command. Other than that LGTM.

app/grandchallenge/core/management/commands/transfer_notifications.py
```diff
@@ -0,0 +1,204 @@
...
+                    notification.actor_object_id = (
+                        notification.action.actor_object_id
+                    )
+                    notification.target_object_id = (
+                        notification.action.target_object_id
+                    )
+                    notification.action_object_object_id = (
+                        notification.action.action_object_object_id
+                    )
+                    notification.verb = notification.action.verb
```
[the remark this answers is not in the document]
amickan: oh wow, I totally overlooked those, thanks for catching that!

MikeOverkamp-diag (approved):

jmsmkn:
LGTM! Can we hold off on merging this until tomorrow so that 6359f3bd74fdd245ebd552202b9427e39d49e040 has been deployed for ~a day? I also want to roll in #1991 to that as this is causing problems for Kits21.

Issue #1943, opened 2021-07-29
CSL tests are flaky


jmsmkn:
The CSL tests call out to the network (arxiv and citeproc), which often times out. Mock those requests.

jmsmkn:
We're still seeing this with `test_arxiv_to_citeproc` https://github.com/comic/grand-challenge.org/runs/3542959545?check_suite_focus=true

jmsmkn:
And again in `test_publication_creation` 🙂 https://github.com/comic/grand-challenge.org/runs/3554040120?check_suite_focus=true

jmsmkn:
It might be worth turning your wifi off before running `docker-compose run --rm web pytest -n 2 -k publications_tests` to catch any others.

PR #1991, opened 2021-09-07
Add editor permissions when using the job create API

jmsmkn:
This matches what is done in `AlgorithmExperimentCreate`


jmsmkn:
Failed test is due to #1943

PR #1993, opened 2021-09-07
Fix 404 for 5XX errors

jmsmkn:
Would be good to have nice error pages (including the one for sentry), but this is fine for now, more important to get the correct error codes.

Closes #1993


jmsmkn:
Skipping the tests as the nginx config is not covered by CI.

PR #1998, opened 2021-09-08
Fixes archive item list search function

### LINKED ISSUE #1995: Search is broken for Archive Items list view
To reproduce, navigate to `/archives/<slug>/items/`, type something in the search box.

amickan:
Closes #1995

### DIFF
--- a/app/grandchallenge/archives/views.py
+++ b/app/grandchallenge/archives/views.py
@@ -492,10 +492,13 @@ class ArchiveItemsList(
     row_template = "archives/archive_items_row.html"
     search_fields = [
         "pk",
-        "name",
+        "values__interface__title",
+        "values__value",
+        "values__image__name",
+        "values__file",
     ]
     columns = [
-        Column(title="Values", sort_field="pk"),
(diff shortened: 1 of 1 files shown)


app/grandchallenge/archives/views.py
```diff
@@ -492,10 +492,13 @@ class ArchiveItemsList(
     row_template = "archives/archive_items_row.html"
     search_fields = [
         "pk",
-        "name",
+        "values__interface__title",
+        "values__value",
+        "values__image__name",
+        "values__file",
```
amickan, line 498: For some reason it did not allow me to query `values__file__name`. Since the `file` returns the name as well thouh, I guess querying like this is fine, though I'd be interested to know why `name` is not an allowed look-up for a FileField.

jmsmkn: `__image` is a model, which has the Field `name`, so you can look that up in the db. `__file` is a Field on the values, so you cannot do a through lookup there.

jmsmkn (approved):
Thanks!

PR #2002, opened 2021-09-09
Fixes algorithm request notifications and notification clean up

amickan:
Algorithm access request notifications were sent not only to editors of the algorithm, but also to users who had submitted jobs to the algorithm. This is fixed now. Requires a management command to update existing algorithm follows for editors.

I also added a signal that takes care of deleting notifications when an object that is currently used for notifications (as actor, target or action object) is deleted.

### DIFF
--- a/app/grandchallenge/notifications/models.py
+++ b/app/grandchallenge/notifications/models.py
@@ -141,12 +141,12 @@ class Notification(UUIDModel):
         message = kwargs.pop("message", None)
         description = kwargs.pop("description", None)
         context_class = kwargs.pop("context_class", None)
-
         if (
             type == NotificationType.NotificationTypeChoices.FORUM_POST
             or type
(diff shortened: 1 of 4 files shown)


app/grandchallenge/notifications/models.py
```diff
@@ -158,6 +158,11 @@ def send(type, **kwargs):  # noqa: C901
                 ]
             else:
                 receivers = followers(target)
+        elif (
+            type == NotificationType.NotificationTypeChoices.ACCESS_REQUEST
+            and target._meta.model_name == "algorithm"
+        ):
+            receivers = followers(target, flag="access_request")
```
amickan, line 165: This is unfortunately necessary because `followers(target, flag='')` unexpectedly returns not only follows with a flag of '' but also follows with a filled flag (i.e. 'job-active' or 'job-inactive')....

jmsmkn (approved):

PR #2004, opened 2021-09-09
Mocks get_doi_csl for publication creation test

### LINKED ISSUE #1943: CSL tests are flaky

amickan:
This is the last test that needs mocking. Sorry I didn't check properly before. 

Closes #1943


jmsmkn (approved):
No problem :-)

PR #2005, opened 2021-09-09
Only add image to answer where the sessions match

jmsmkn:
As answer image types are editable it could be the case that there are multiple pending upload sessions for an answer at the same time. The value of the latest upload session pk is always stored in the answer. As the upload sessions may be processed out of order (no FIFO guarantees) this PR only allows the image to be assigned to the answer if it originates from the session expected by the answer. Older versions of the `answer_image` are still saved and tracable through the answer history.


miriam-groeneveld (approved):
Looks good to me. Didn't know about casefold()

jmsmkn:
Yes, this is the way to do caseless string comparison, nasty things can happen otherwise:

https://eng.getwisdom.io/hacking-github-with-unicode-dotless-i/
https://www.djangoproject.com/weblog/2019/dec/18/security-releases/

PR #2008, opened 2021-09-10
Remove action field from notification model

amickan:
Last step to finalize the new notification model.


amickan:
Let's wait with merging this until after the weekend.

jmsmkn (approved):

jmsmkn:
> 
> 
> Let's wait with merging this until after the weekend.

It's ok, we've not had any warnings about the field since deploying the last change so this should be safe.

PR #2009, opened 2021-09-10
Improve CodeBuild and GitHub apps

MikeOverkamp-diag:
This fixes all issues in https://github.com/DIAGNijmegen/rse-roadmap/issues/77. The only thing that needs to be added is developer documentation, which I will add next week.


MikeOverkamp-diag:
@jmsmkn the tests are failing because the licensee dependency is not installed. Do I need to push a new web-base image?

jmsmkn:
The containers are only rebuilt when the python version is updated, or when the sha256 of the poetry.lock file changes. You can force this by adding a newline in poetry.lock (eg https://github.com/comic/grand-challenge.org/commit/e67bf2d47f902ce9e991068a8e1992e6e0823c0e#diff-f53a023eedfa3fbf2925ec7dc76eecdc954ea94b7e47065393dbad519613dc89R9)

app/grandchallenge/algorithms/migrations/0016_auto_20210910_1308.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 3.1.13 on 2021-09-10 13:08
```
jmsmkn, line 1: Please only add 1 migration per app per PR unless necessary.

app/grandchallenge/algorithms/models.py
```diff
@@ -156,10 +157,23 @@ class Algorithm(UUIDModel, TitleSlugDescriptionModel):
     )
     use_flexible_inputs = models.BooleanField(default=True)
     repo_name = models.CharField(blank=True, max_length=512)
+    image_requires_gpu = models.BooleanField(default=False)
+    image_requires_gpu_memory_gb = models.PositiveIntegerField(default=4)
+    image_requires_memory_gb = models.PositiveIntegerField(default=4)
+    image_requires_cpu_cores = models.DecimalField(
+        default=Decimal("1.0"), max_digits=4, decimal_places=2
+    )
```
jmsmkn, line 165: `image_requires_gpu_memory_gb` and `image_requires_cpu_cores` are never used on the image and do not need to be set, so can be removed. We can add them if we ever need it. No need to remove the fields on the `ComponentImage` for now.

Would be nice if `image_requires_gpu` and `image_requires_memory_gb` also provide the initial values to the form when uploading a new container via the web UI.

app/grandchallenge/algorithms/templates/algorithms/algorithm_no_slug.html
```diff
@@ -0,0 +1,23 @@
...
+        <li class="breadcrumb-item active"
+            aria-current="page">No algorithm found</li>
+    </ol>
+{% endblock %}
+
+{% block content %}
+
+<div class="mb-2">Unfortunately something went wrong while trying to find the requested algorithm.</div>
+
+<div>If you were trying to link a github repository to an algorithm, please do so manually in the algorithm's settings.</div>
```
jmsmkn, line 21: Do we need a separate view for this or can this be done by sending context to a 404 page?

app/grandchallenge/algorithms/views.py
```diff
@@ -208,6 +210,9 @@ def get_context_data(self, **kwargs):
             {
                 "pending_permission_requests": pending_permission_requests,
                 "github_app_install_url": f"{settings.GITHUB_APP_INSTALL_URL}?state={self.object.slug}",
+                "builds": Build.objects.filter(
```
jmsmkn, line 213: Is it not better to go the other way around and use `object.build_set.all()`? Then it can be used in a prefetch related too.

app/grandchallenge/github/models.py
```diff
@@ -38,7 +38,7 @@ def __str__(self):
 
     @property
     def access_token_is_expired(self):
-        return self.access_token_expires > timezone.now()
+        return self.access_token_expires < timezone.now()
```
jmsmkn, line 41: Good catch!

app/grandchallenge/algorithms/views.py
MikeOverkamp-diag: Of course! That makes a lot more sense.

app/grandchallenge/algorithms/templates/algorithms/algorithm_no_slug.html
MikeOverkamp-diag: Hmm, I hadn't considered that. That should probably work.

app/grandchallenge/github/tasks.py
```diff
@@ -47,27 +46,73 @@ def get_zipfile(*, pk):
...
+    payload = ghwm.payload
+    repo_url = get_repo_url(payload)
+    zip_name = f"{ghwm.repo_name}-{ghwm.tag}.zip"
+    tmp_zip = tempfile.NamedTemporaryFile()
+    has_open_source_license = False
+    license = "No license file found"
+    with tempfile.TemporaryDirectory() as tmpdirname:
+        proces = subprocess.Popen(
+            [
+                "git",
```
jmsmkn, line 69: Do we need to authorise these calls in some way to use private reops and/or lfs?

app/grandchallenge/github/tasks.py
```diff
@@ -47,27 +46,73 @@ def get_zipfile(*, pk):
...
+                "clone",
+                "--recurse-submodules",
```
jmsmkn, line 71: Would be nice to make recursing submodules an option as sometimes this can lead to issues (eg, for a repo we can't access but which is unimportant for building the inference container).

app/grandchallenge/github/tasks.py
MikeOverkamp-diag: This is handled by the `get_repo_url` function, which adds a authentication token to the repo url.

app/config/settings.py
```diff
@@ -1089,6 +1089,21 @@ def strtobool(val) -> bool:
 CODEBUILD_ACCESS_KEY = os.environ.get("CODEBUILD_ACCESS_KEY", "")
 CODEBUILD_SECRET_KEY = os.environ.get("CODEBUILD_SECRET_KEY", "")
 CODEBUILD_REGION = os.environ.get("CODEBUILD_REGION", "")
+CODEBUILD_ONLY_ALLOW_VERIFIED = strtobool(
```
jmsmkn, line 1092: This setting can be removed, we only want verified users to do this. Maybe add the verified checks to the views too?

app/grandchallenge/github/tasks.py
```diff
@@ -47,27 +46,73 @@ def get_zipfile(*, pk):
...
+    )
+    ghwm = GitHubWebhookMessage.objects.get(pk=pk)
+    payload = ghwm.payload
+    repo_url = get_repo_url(payload)
+    zip_name = f"{ghwm.repo_name}-{ghwm.tag}.zip"
+    tmp_zip = tempfile.NamedTemporaryFile()
+    has_open_source_license = False
+    license = "No license file found"
+    with tempfile.TemporaryDirectory() as tmpdirname:
+        proces = subprocess.Popen(
```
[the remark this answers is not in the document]
MikeOverkamp-diag: Agreed, that would be better.

app/grandchallenge/github/views.py
```diff
@@ -35,8 +35,11 @@ def github_webhook(request):
         )
 
     payload = json.loads(request.body)
-
-    GitHubWebhookMessage.objects.create(payload=payload)
+    if (
+        not settings.CODEBUILD_ONLY_ALLOW_VERIFIED
+        or request.user.verification.is_verified
```
jmsmkn, line 40: Only allow verified users to do this.

jmsmkn (commented):
Looks good! Few minor things

app/grandchallenge/github/tasks.py
MikeOverkamp-diag: I guess that would then be an option on the algorithm object? I agree that would be a smart thing to do, as it could potentially lower the zip size by a substantial amount and like you said, we might run into authentication issues with the submodules.

app/grandchallenge/algorithms/views.py
MikeOverkamp-diag: Hmm upon further inspection that wouldn't work, as the build is linked to `AlgorithmImages` rather than the algorithm itself. So we'd have to go over the builds for each image linked to the algorithm. In which case I think the current query is more efficient.

app/grandchallenge/algorithms/forms.py
```diff
@@ -111,6 +111,7 @@ class Meta:
             "social_image",
             "public",
             "use_flexible_inputs",
+            "repo_name",
```
jmsmkn, line 114: Just to double check - do we want to make this user editable?

app/grandchallenge/algorithms/models.py
```diff
@@ -156,10 +156,23 @@ class Algorithm(UUIDModel, TitleSlugDescriptionModel):
     )
     use_flexible_inputs = models.BooleanField(default=True)
     repo_name = models.CharField(blank=True, max_length=512)
+    image_requires_gpu = models.BooleanField(default=False)
+    image_requires_memory_gb = models.PositiveIntegerField(default=4)
```
jmsmkn, line 160: Sorry, missed these before - can we default to `True` and `15`? Users always forget to set these which is why we changed the defaults in the form, which are not being populated from here.

app/grandchallenge/algorithms/templates/algorithms/algorithm_detail.html
```diff
@@ -206,6 +212,20 @@ <h2>Permission Requests</h2>
...
             </div>
 
+            <div class="tab-pane fade" id="builds" role="tabpanel"
+                 aria-labelledby="v-pills-builds-tab">
+
+                <h2>Builds</h2>
+
+                {% for build in builds %}
+                    <div class="mb-2">
+                        <a class="btn btn-{% if build.status == 'SUCCEEDED' %}success{% elif build.status == 'IN_PROGRESS' %}warning{% else %}danger{% endif %}">
```
jmsmkn, line 222: Don't need to change this but I quite like the pattern of adding `animate` and `status_class` on models to do this, saves some repetition and UI bugs https://github.com/comic/grand-challenge.org/blob/be58bf10a93baa4344f783535e3badd8ce7c9aa5/app/grandchallenge/components/models.py#L844

app/grandchallenge/codebuild/tasks.py
```diff
@@ -29,7 +29,13 @@ def create_codebuild_build(*, pk):
         # Repository is not linked to algorithm
         return
 
-    algorithm_image = AlgorithmImage.objects.create(algorithm=algorithm)
+    algorithm_image = AlgorithmImage.objects.create(
+        algorithm=algorithm,
+        requires_gpu=algorithm.image_requires_gpu,
+        requires_gpu_memory_gb=algorithm.image_requires_gpu_memory_gb,
+        requires_memory_gb=algorithm.image_requires_memory_gb,
+        requires_cpu_cores=algorithm.image_requires_cpu_cores,
```
jmsmkn, line 37: Two of these have now been deleted, check for other instances.

app/grandchallenge/algorithms/urls.py
```diff
@@ -35,6 +36,7 @@
         ComponentInterfaceList.as_view(),
         name="component-interface-list",
     ),
+    path("not-found/", AlgorithmNotFound.as_view(), name="no-slug"),
```
jmsmkn, line 39: Lost the comment but saw the 404 template was updated, is this route still needed? If not, remove the view too,

app/grandchallenge/algorithms/forms.py
MikeOverkamp-diag: I think that's the way to go. It's not foolproof, but it allows users to change the repo if they have already setup the installation and it is part of solving the issue if an admin user needs to approve the installation before linking it on gc.

app/grandchallenge/algorithms/templates/algorithms/algorithm_detail.html
MikeOverkamp-diag: I'll use that pattern in the future, thanks for pointing it out.

jmsmkn (approved):

PR #2018, opened 2021-09-16
Integrate challenge forums into challenge layout

### LINKED ISSUE #2017: Integrate Forums into challenges
Navigating to the forum of a challenge currently takes the participant outside of the challenge environment. Navigating back to the challenge is not possible through the breadcrumbs on the forum page and instead requires going via the Challenge tab and searching for the respective Challenge again. It would be nicer if the forums were visually integrated into the challenge page layout and if the breadcrumbs reflected their nesting in the challenge rather than their nesting under all forum on GC. 

See here: https://github.com/DIAGNijmegen/rse-roadmap/issues/83#issuecomment-919250835

amickan:
This integrates challenge forums into the challenge layout. The challenge navbar remains visible at all times and the breadcrumbs reflect the nesting of the forum under the challenge and allow easy backwards navigation to the other challenge pages. 

image

The forum-related pages that are common to all forums (i.e. not challenge specific) stay with their previous layout:
image

Closes #2017

### DIFF
--- a/app/grandchallenge/core/templates/challenge.html
+++ b/app/grandchallenge/core/templates/challenge.html
@@ -17,115 +17,12 @@
 {% endblock %}
 
 {% block messages %}
-    {% if challenge.banner %}
-        <div class="row mb-3">
-            <div class="col-12">
-                <a style="width: 100%;"
-                   href="{% url 'pages:home' challenge_short_name=challenge.short_name %}">
-                    <img alt="{% firstof challenge.title challenge.short_name %} Banner"
-                         class="rounded w-100"
-                         src="{{ challenge.banner.x20.url }}"
-                         srcset="{{ challenge.banner.x10.url }} 1x,
-                                 {{ challenge.banner.x15.url }} 1.5x,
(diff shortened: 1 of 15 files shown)


amickan:
I did not change the urls for the forum related pages. All I changed are the templates and the breadcrumbs. This means that the url for a challenge forum still reads `forums/forum/challenge-name` and does not match the breadcrumbs (which now read `Challenges/challenge-name/Forum`. That's not so nice, but maybe not that big of a deal? 

Related to this is a small layout issue which I can't figure out how to solve: the layout of all subdomain pages (all challenge pages basically) appears to be slightly different from the rest of grand challenge. The outer container is somewhat bigger so that all the content appears relatively bigger. The forum pages, however, follow the layout of the rest of grand challenge and this is noticeable when navigating from say the `Info` tab of the challenge page to the `forum` tab... the navbar shrinks a little... I couldn't find any extra css or html that would explain why the subdomain pages have a different layout though, so I'm not sure how to fix this...

amickan:
closed in favor of this [PR](https://github.com/comic/grand-challenge.org/pull/2027) instead

PR #2022, opened 2021-09-17
Refactor challenge.html

amickan:


app/grandchallenge/core/templates/base.html
```diff
@@ -66,11 +66,14 @@
...
 
             {% block messages %}
+                {% include 'challenges/challenge_banner.html' %}
                 {% include "grandchallenge/partials/messages.html" %}
             {% endblock %}
 
             {% block outer_content %}
-                {% block topbar %}{% endblock %}
+                {% block topbar %}
+                    {% include 'challenges/challenge_topbar.html' with challenge=challenge %}
```
jmsmkn, line 75: `with challenge=challenge` should be unnecessary, right? Why is it needed here and not for the banner partial?

app/grandchallenge/core/templates/base.html
```diff
@@ -101,6 +104,9 @@
 {% block script %}
     {% include "grandchallenge/partials/script.html" %}
     {% include "timezones/partials/detect_timezone.html" %}
+    {% if challenge %}
+        <script src="{% static 'js/challenges/dropdown.js' %}" type="module"></script>
```
jmsmkn, line 108: This script line can move to app/grandchallenge/challenges/templates/challenges/challenge_topbar.html, as long as it's deferred with `<script src="{% static 'js/challenges/dropdown.js' %}" type="module" defer></script>`

app/grandchallenge/core/templates/base.html
amickan: oh yes, that can go indeed!

app/grandchallenge/challenges/templates/challenges/challenge_topbar.html
```diff
@@ -0,0 +1,93 @@
...
+                       role="button">
+                            Join
+                    </a>
+                </div>
+            {% endif %}
+        </div>
+    </div>
+{% endif %}
+
+<script src="{% static 'js/challenges/dropdown.js' %}" type="module" defer></script>
```
jmsmkn, line 93:
```suggestion
    <script src="{% static 'js/challenges/dropdown.js' %}" type="module" defer></script>
{% endif %}

```

jmsmkn (approved):
This looks good to me. I think there will be a couple of changes but these are not vital:
- The title block is not handled so some page titles will change, this needs to be revisited as our strategy around page titles is all over the place
- The breadcrumbs will be missing for temppage, but I'm unsure what that thing is for, needs to be investigated

PR #2025, opened 2021-09-20
Cache the product list for 1 hour

jmsmkn:
Assumes that the contents of products/partials/product_list.html
only varies with the query string

### DIFF
--- a/app/grandchallenge/products/templates/products/product_list.html
+++ b/app/grandchallenge/products/templates/products/product_list.html
@@ -1,4 +1,5 @@
 {% extends 'products/base.html' %}
+{% load cache %}
 
 {% block title %}
     AI for radiology
@@ -115,7 +116,8 @@
       </p>
     </div>
 
-
-  {% include "products/partials/product_list.html" %}
+  {% cache 3600 product_list request.META.QUERY_STRING %}
+    {% include "products/partials/product_list.html" %}
+  {% endcache %}
 
 {% endblock %}


MikeOverkamp-diag (approved):

Kicksta:
Search and filtering works.
Logo's don't show anymore (Shows image icon + alt string).
The main page: https://gc.localhost/aiforradiology/, shows no products but states 19/19 results. I have to filter or search to get to see the products. I guess this discrepancy will last for an hour then because it cached the 'empty state'?

jmsmkn:
Did you load the page before populating the database?

Logos not loading would be very strange.

Kicksta:
Yes, I loaded the page before populating the database. I now removed cache from my browser, but problem persists (is this the right way to check?). I switched to master and I get to see the database again while being on the homepage. Switched back to branch, and no products are shown.

Logo problem is also on master with cycle-docker-compose on branch.
When on master with cycle-docker-compose on master, logo's are visible again.

jmsmkn:
What happens on the branch if you change `{% cache 3600 product_list request.META.QUERY_STRING %}` to `{% cache 3600 product_list_test request.META.QUERY_STRING %}` and refresh the page?

Kicksta:
homepage list is not empty

jmsmkn:
So it looks ok? Do the logos work?

Kicksta:
Okay i am again intrigued by the mystery of computers. 
On branch, i did again cycledockercompose.
Then went directly to aiforradiology/import-data. Uploaded database. Went to homepage: logo's work, database list shows. With the original line of code `{% cache 3600 product_list request.META.QUERY_STRING %}`.

Conclusion: original review request approved. Sorry for the hassle 🥴

jmsmkn:
It's ok! Yes, that is what I expected, caches are difficult and probably easier to explain what is going on synchronously, and we should try to find an easier way to test this in development so that I can do it too. When you're working on this template in development it will be easiest to drop the cache period to `1`s, rather than `3600`s.

PR #2027, opened 2021-09-20
Integrate forums into challenge layout

### LINKED ISSUE #2017: Integrate Forums into challenges

amickan:
Closes #2017

### DIFF
--- a/app/grandchallenge/forums/templates/board_base.html
+++ b/app/grandchallenge/forums/templates/board_base.html
@@ -2,6 +2,8 @@
 {% load static %}
 {% load i18n %}
 {% load forum_permission_tags %}
+{% load forum_extras %}
+{% load guardian_tags %}
 
 {% block title %}{% block sub_title %}{% endblock sub_title %} - {{ request.site.name }} Forums{% endblock title %}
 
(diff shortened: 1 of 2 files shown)


app/grandchallenge/forums/templates/board_base.html
```diff
@@ -12,31 +14,63 @@
...
-            <div class="float-right controls-link-wrapper">
-                {% if not request.user.is_anonymous %}
-                    <a href="{% url 'notifications:follow-list' %}" class="d-inline-block ml-3"><i class="fas fa-bookmark">&nbsp;</i>{% trans "Subscriptions" %}</a>
-                    <a href="{% url 'forum_member:user_posts' request.user.id %}" class="d-inline-block ml-3"><i class="fas fa-comments">&nbsp;</i>{% trans "View my posts" %}</a>
-                {% endif %}
-                {% get_permission 'can_access_moderation_queue' request.user as can_access_moderation_queue %}
-                {% if can_access_moderation_queue %}
-                    <a href="{% url 'forum_moderation:queue' %}" class="d-inline-block ml-3"><i class="fas fa-gavel">&nbsp;</i>{% trans "Moderation queue" %}</a>
-                {% endif %}
+    {% if forum %}
```
jmsmkn, line 17: Shouldn't this be `{% if forum.challenge %}`?

amickan: For consistency, I guess that's better, but it works like this as well because the general forum views don't have a specific `forum` in their context, so checking for `forum` is enough to tell challenge-specific forum pages from general forum pages.

jmsmkn: Can't say that I understand why that is, but ok!

jmsmkn (approved):

PR #2030, opened 2021-09-20
Fix verification check before building from github message

MikeOverkamp-diag:


app/grandchallenge/github/views.py
```diff
@@ -36,8 +37,17 @@ def github_webhook(request):
...
 
     payload = json.loads(request.body)
-    if request.user.verification.is_verified:
-        GitHubWebhookMessage.objects.create(payload=payload)
+    try:
+        user = GitHubUserToken.objects.get(
+            github_user_id=payload["sender"]["id"]
+        ).user
+    except GitHubUserToken.DoesNotExist:
+        return HttpResponse("ok", content_type="text/plain")
```
jmsmkn, line 45: Is that really what we want? Shouldn't it be a 4xx?

MikeOverkamp-diag: I'm not sure. My reasoning behind always returning a 200 is that this endpoint is pubicly available. By returning different status codes, we are giving away some information on what's going on. Either way, GitHub does not do anything with the response.

jmsmkn: Ok - is there any risk that the anonymous user can get associated with a GitHubUserToken? Do we need to guard for that here?

MikeOverkamp-diag: The tokens only get created in views with the `login_required` decorator, so the query here should return an actual user. There is a chance that the GitHub user that creates a tag does not have a token. This can happen for repos that get linked to an algorithm by another user than the one creating the tag. I do not see a better way to do this though, as the webhook only provides us info about the GitHub user and has no knowledge of anything that happens on gc.

jmsmkn: Hmm, so does every user who wants to tag repos for an algorithm need to install the app somehow?

MikeOverkamp-diag: We need to somehow link a user's github id to their gc account. Adding it on install was the only way I could really think of to do that in the current setup. I'm open to any other suggestions of course.

jmsmkn: I think this will end up with weird errors and people will not understand why their algorithms are not building. The user who does the installation first time must be verified (we can check this) and be an admin for the repo (as they install the app). Do we keep track of this user, algorithm and repo? As they are an admin on the repo they should trust any other user that can tag it, they must be a collaborator who may not necessarily have a gc account. But, as the verified user trusts them, then I think that this is ok to build. What we would then need is that if the verified user looses their verification status, then builds for that repo and algorithm are no longer built.

MikeOverkamp-diag: That sounds like a very good solution. I'll make the verification check part of the installation procedure, rather than the build procedure. The only pitfall I see here is the case where the algorithm user on gc is not an admin for the linked repo and the app installation needs to be approved. In that case, we lose the link between the algorithm and the installer and the repo_name needs to be filled manually. But I think there should be a way around that by tracking some more details wrt the github installations. 

I will do a bit of refactoring to support this tomorrow.

jmsmkn: We can also make the algorithm update view verified only.

MikeOverkamp-diag:
Rather than over-complicating this, I made all views that allow adding a repo to an algorithm verification only. I think that should be sufficient.

jmsmkn (approved):

PR #2031, opened 2021-09-20
Extended config options to hide for plugins and tools

miriam-groeneveld:
Related issue: https://github.com/DIAGNijmegen/rse-cirrus-core/issues/1430


app/grandchallenge/workstation_configs/migrations/0008_auto_20210920_1439.py
```diff
@@ -0,0 +1,47 @@
...
+
+    dependencies = [
+        (
+            "workstation_configs",
+            "0007_workstationconfig_auto_jump_center_of_gravity",
+        )
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
jmsmkn, line 16: We can no longer remove fields straight up, instead, we need to go through a process of deprecation (https://www.caktusgroup.com/blog/2021/05/25/django-migrations-and-deployment/). However, these views are rarely called, so a little downtime here is acceptable in this instance.

jmsmkn (approved):

jmsmkn:
Test failures are fine, the parent branch for this PR was stale.

PR #2035, opened 2021-09-22
Adds documentation app and docpage model

amickan:
This adds a documentation app to grand challenge. This is a first step towards adding a documentation to GC and is part of https://github.com/DIAGNijmegen/rse-roadmap/issues/92


app/grandchallenge/documentation/admin.py
```diff
@@ -0,0 +1,25 @@
...
+from django.forms import ModelForm
+from markdownx.admin import MarkdownxModelAdmin
+
+from grandchallenge.documentation.models import DocPage
+
+
+class AdminDocPageForm(ModelForm):
+    class Meta:
+        model = DocPage
+        exclude = ()
```
jmsmkn, line 11: Isn't this another way of doing `__all__`? If so, I would prefer if we are explicit about what fields to include, even for admins.

app/grandchallenge/documentation/models.py
```diff
@@ -0,0 +1,136 @@
...
+from django.utils.translation import gettext_lazy as _
+from django_extensions.db.fields import AutoSlugField
+from simple_history.models import HistoricalRecords
+
+from grandchallenge.core.templatetags.bleach import clean
+from grandchallenge.core.utils.query import index
+from grandchallenge.subdomains.utils import reverse
+
+
+class LevelChoices(models.TextChoices):
```
jmsmkn, line 13: Nitpick but an `IntegerChoice` and a `SmallPositiveIntegerField` would probably make more sense here.

app/grandchallenge/documentation/models.py
```diff
@@ -0,0 +1,136 @@
...
+        ),
+    )
+
+    order = models.IntegerField(
+        editable=False,
+        default=1,
+        help_text="Determines order in which pages appear in side menu",
+    )
+
+    history = HistoricalRecords()
```
jmsmkn, line 68: You probably only want to keep a few fields in the history. For instance, only the content.

app/grandchallenge/documentation/models.py
```diff
@@ -0,0 +1,136 @@
...
+        if not self.id:
+            # get max value of order for current pages.
+            try:
+                max_order = DocPage.objects.aggregate(Max("order"))
+            except ObjectDoesNotExist:
+                max_order = None
+            try:
+                self.order = max_order["order__max"] + 1
+            except TypeError:
+                self.order = 1
```
jmsmkn, line 84:
```suggestion
            try:
                self.order = DocPage.objects.aggregate(Max("order")) + 1
            except ObjectDoesNotExist:
                # Use the default
                pass
```

app/grandchallenge/documentation/models.py
```diff
@@ -0,0 +1,136 @@
...
+            self.normalize_page_order(pages)
+
+    @staticmethod
+    def normalize_page_order(pages):
+        """Make sure order in pages Queryset starts at 1 and increments 1 at
+        every page. Saves all pages
+        """
+        for idx, page in enumerate(pages):
+            page.order = idx + 1
+            page.save()
```
jmsmkn, line 123: This would be better done with a bulk update as this will be very slow once you start to have a few documentation pages, just FYI, doesn't need to be fixed for now.

app/grandchallenge/documentation/models.py
```diff
@@ -0,0 +1,136 @@
...
+        blank=True,
+        help_text=(
+            "On pages and in menu items, use this text. Spaces and special "
+            "chars allowed here. Optional field. If emtpy, title is used"
+        ),
+    )
+
+    content = models.TextField()
+
+    level = models.CharField(
```
jmsmkn, line 53: Are you sure this approach works for what you need? Would rather get the model right on the first migration.

amickan: I've played around with a possible detail view already and we can definitely make it work with this model. For the detail view, I will need to add a few more property methods (apart from getting the next page, I will need a method to get a page's parent and its children), but that's not a problem. 

Are there any particular issues you are worried about?

amickan: Or would it be better to add `parent` and `children` fields to the model rather than making these two things property methods?

amickan: I added the parent and children methods together with the other changes so you get a better idea ;-)

app/grandchallenge/documentation/models.py
```diff
@@ -0,0 +1,185 @@
...
+            try:
+                self.order = (
+                    DocPage.objects.aggregate(Max("order"))["order__max"] + 1
+                )
+            except ObjectDoesNotExist:
+                # Use the default
+                pass
+            except TypeError:
+                # Use the default
+                pass
```
jmsmkn, line 85: You can group exceptions if you're going to take the same action:

```suggestion
            except (ObjectDoesNotExist, TypeError):
                # Use the default
                pass

```

app/grandchallenge/documentation/models.py
```diff
@@ -0,0 +1,185 @@
...
+    def previous(self):
+        try:
+            previous_page = DocPage.objects.filter(order=self.order - 1).get()
+        except ObjectDoesNotExist:
+            previous_page = None
+        return previous_page
+
+    @property
+    def parent(self):
+        if self.level == 1:
```
jmsmkn, line 148: You should compare with `self.Level.PRIMARY` rather than using the values directly. Same for when you use them for selecting a particular level.

app/grandchallenge/documentation/models.py
```diff
@@ -0,0 +1,185 @@
...
+                    order__lt=self.order, level=1
+                ).last()
+            elif self.level == 3:
+                parent = DocPage.objects.filter(
+                    order__lt=self.order, level=2
+                ).last()
+        return parent
+
+    @property
+    def children(self):
```
jmsmkn, line 162: This is going to be really inefficient as it goes to the database a lot. I feel now that using a `parent = models.ForeignKey("self", null=True, blank=True, on_delete=???, related_name="children")` would be a better approach, then getting the children would be a case of doing `page.children.all()`.

app/grandchallenge/documentation/models.py
```diff
@@ -0,0 +1,149 @@
...
+        blank=True,
+        help_text=(
+            "On pages and in menu items, use this text. Spaces and special "
+            "chars allowed here. Optional field. If emtpy, title is used"
+        ),
+    )
+
+    content = models.TextField()
+
+    level = models.PositiveSmallIntegerField(
```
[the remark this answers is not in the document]
amickan: If we only had top and second level pages, we wouldn't need the levels anymore, but if we also want to have a third level, I think the levels field will be necessary to tell second from third level pages. Both second and third level pages have a parent and while only second level pages can have children, they don't necessarily need to have children, so I think it would be difficult to retrieve third level pages seperately unless we have an explicit way of marking them as third level pages.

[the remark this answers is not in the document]
jmsmkn: It's hard to say without seeing how this would be used in practice. There are approaches that would work without having to maintain the levels/parents but it depends on how the model is going to be used. What problem are you trying to solve here? It is really hard to judge what approach we should take just from these tests. If it is for a sidebar then we need to think of a data structure that can deliver all of the information we need in a single call, so I wonder if getting all the pages by level is necessary.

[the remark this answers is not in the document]
amickan: Hm, yes, this is only for the sidebar, so for creating a visual hierarchy of pages. I've now pushed what I had in mind for the detail view. I'm not sure how we would combine the level and the parent/children information into just one field on the model.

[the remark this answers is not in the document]
amickan: I think I found a way to not use the levels field after all. I should have seen this earlier...

app/grandchallenge/documentation/templatetags/docpage_extras.py
```diff
@@ -0,0 +1,12 @@
+from django import template
+
+register = template.Library()
+
+
+@register.simple_tag
+def get_grandchildren(page):
```
jmsmkn, line 7: Can the number of queries here be reduced with a prefetch related?

app/grandchallenge/documentation/views.py
```diff
@@ -0,0 +1,41 @@
...
+            }
+        )
+
+        return context
+
+
+class DocumentationHome(DocPageDetail):
+    def get_object(self, queryset=None):
+        page = DocPage.objects.first()
+        return page
```
amickan, line 41: This will result in an error when there are no pages yet. We could add an error page as it is done for pages, or instead (since this is just an issue at the very beginning and never again after that), just create a default first page, like so: 
 ```
  if not page:
            page = DocPage.objects.create(title='main')
```

jmsmkn: Django has something built-in just for that :-) https://docs.djangoproject.com/en/3.2/topics/http/shortcuts/#get-object-or-404

This PR LGTM, only thing I'm wondering is if `display_title` is actually used? This is an artefact from the ancient `Pages` model that uses `title` and `display_title` rather than `slug` and `title`.

amickan: True, it's not necessary. I thought that the title would otherwise not allow special characters and icons, but it does I see now. ;-)

jmsmkn (approved):

PR #2036, opened 2021-09-22
Add event driven component execution

jmsmkn:
This PR adds support for event-driven backends for component execution, and implements this with the `AmazonECSExecutor` that uses EventBridge + SQS to send messages to the Celery workers. The success and failure tasks now need to be added to the model as there is a break in the chain of tasks for the event-driven backends between `execute_job` and `parse_outputs`, if a Celery chain is used then it would be executed after `provisioning`. These are serialized to json on the model automatically by Celery, so cannot contain non-json serializable things (eg UUIDs, cast these to strings first).


app/grandchallenge/algorithms/tasks.py
```diff
@@ -139,10 +140,10 @@ def create_algorithm_jobs_for_session(
     algorithm_editors = [algorithm_image.algorithm.editors_group]
 
     # Send an email to the algorithm editors and creator on job failure
-    linked_task = send_failed_session_jobs_notifications.signature(
+    task_on_success = send_failed_session_jobs_notifications.signature(
```
miriam-groeneveld, line 143: shouldn't this be task_on_failure?

app/grandchallenge/algorithms/tasks.py
```diff
@@ -107,8 +107,8 @@ def execute_algorithm_job_for_inputs(*, job_pk):
     job = Job.objects.get(pk=job_pk)
 
     # Send an email to the algorithm editors and creator on job failure
-    linked_task = send_failed_job_notification.signature(
-        kwargs={"job_pk": job.pk}, immutable=True
+    task_on_success = send_failed_job_notification.signature(
```
miriam-groeneveld, line 110: shouldn't this be task_on_failure?

app/grandchallenge/algorithms/tasks.py
jmsmkn: You would think so, but that is not how the `send_failed_session_jobs_notifications` task was implemented. This was a straight search and replace, linked tasks were only ever run when the chain was successful.

app/grandchallenge/algorithms/tasks.py
jmsmkn: See above.

PR #2044, opened 2021-09-27
Conditionally determine visibility of columns in datatables

HarmvZ:
This PR allows us to conditionally show / hide columns in a datatable. 

It works by defining an `optional_condition` callable on a `Column` definition for datatables. The `optional_condition` will be called with the object for that row as parameter for each visible row on the page. If non of the calls evaluate to `True`, the column will be hidden. The visibility is recalculated each time the table data changes: pagination changes, search changes, ordering changes, etc.

This will be useful when the fields in https://github.com/DIAGNijmegen/rse-panimg/pull/51 are implemented here and we want to show these columns in the archive cases list view, but hide the columns that contain empty values. 

As an example I made the "Algorithm Results" column in the cases list view only visible if it actually contains algorithm results.

### DIFF
--- a/app/grandchallenge/archives/views.py
+++ b/app/grandchallenge/archives/views.py
@@ -538,7 +538,14 @@ class ArchiveCasesList(
         Column(title="Created", sort_field="created"),
         Column(title="Creator", sort_field="origin__creator__username"),
         Column(title="View", sort_field="pk"),
-        Column(title="Algorithm Results", sort_field="pk"),
+        Column(
+            title="Algorithm Results",
+            sort_field="pk",
+            optional_condition=lambda o: any(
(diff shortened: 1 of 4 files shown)


jmsmkn:
At a quick glance, I think the way this is implemented has an N+1 problem with 2 extra calls per object. Have you checked by using the debug toolbar, including the history tab to monitor the AJAX calls?

HarmvZ:
By the `optional_condition` for algorithm results? The data for that is prefetched in `get_queryset` so it should not need extra calls to get it. The toolbar indicates 14 queries total independent of the page length / nr of objects.

jmsmkn:
Great! Then my mistake :-)

jmsmkn (approved):

PR #2051, opened 2021-09-28
Add extra metadata fields to image model and cases list view

HarmvZ:
Adds fields described in https://github.com/DIAGNijmegen/rse-panimg/pull/51 to the `Image` model. This needs to be merged before the PR on panimg is merged and released, otherwise the task converting from `PanImg` to `Image` object will fail.

Note: The enum below is copied from panimg, it needs to be imported from there when we upgraded to the correct version:
https://github.com/comic/grand-challenge.org/blob/328a6cfa37534472f465a8dcf9d3f041a3b489cd/app/grandchallenge/cases/models.py#L269-L272


app/grandchallenge/cases/models.py
```diff
@@ -292,6 +309,24 @@ class Image(UUIDModel):
     color_space = models.CharField(
         max_length=5, blank=False, choices=COLOR_SPACES
     )
+    patient_id = models.CharField(max_length=64, null=True, blank=True)
```
jmsmkn, line 312: You should avoid `null` on char fields (https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.Field.null), remove that and set a default to an empty string for all of the CharFields introduced here.

app/grandchallenge/cases/models.py
```diff
@@ -292,6 +309,24 @@ class Image(UUIDModel):
...
+    patient_name = models.CharField(max_length=324, null=True, blank=True)
+    patient_birth_date = models.DateField(null=True, blank=True)
+    patient_age = models.CharField(max_length=4, null=True, blank=True)
```
jmsmkn, line 315: CharField for age? Wouldn't a small positive integer be more appropiate?

app/grandchallenge/cases/models.py
```diff
@@ -292,6 +309,24 @@ class Image(UUIDModel):
...
+        choices=PATIENT_SEX_CHOICES,
+        default=PATIENT_SEX_EMTPY,
+    )
+    study_date = models.DateField(null=True, blank=True)
+    study_instance_uid = models.CharField(max_length=64, null=True, blank=True)
+    series_instance_uid = models.CharField(
+        max_length=64, null=True, blank=True
+    )
+    study_description = models.CharField(max_length=64, null=True, blank=True)
+    series_description = models.CharField(max_length=64, null=True, blank=True)
```
jmsmkn, line 329: Ensure that these field limits are validated/truncated in PanImg

app/grandchallenge/cases/models.py
```diff
@@ -292,6 +309,24 @@ class Image(UUIDModel):
     color_space = models.CharField(
         max_length=5, blank=False, choices=COLOR_SPACES
     )
+    patient_id = models.CharField(max_length=64, null=True, blank=True)
+    patient_name = models.CharField(max_length=324, null=True, blank=True)
```
jmsmkn, line 313: Why 324?

app/grandchallenge/cases/models.py
HarmvZ: No the value in DICOM standard is represented with 3 digits and a single character (either "D", "W", "M", "Y"). So a zeropadded 3 digit number and a character to indicate days/weeks/months/years. We can convert it to a small positive integer but I think this will result in some dataloss/uncertainty because the length of month and year is variable

app/grandchallenge/cases/models.py
```diff
@@ -292,6 +309,24 @@ class Image(UUIDModel):
...
+    patient_birth_date = models.DateField(null=True, blank=True)
+    patient_age = models.CharField(max_length=4, null=True, blank=True)
+    patient_sex = models.CharField(
+        max_length=1,
+        null=True,
+        blank=True,
+        choices=PATIENT_SEX_CHOICES,
+        default=PATIENT_SEX_EMTPY,
```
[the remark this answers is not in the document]
HarmvZ: The value for `PATIENT_SEX_EMPTY` is `None`. I implemented this like the `stereoscopic_choice` field which I think is correct (unlike the `eye_choice` field).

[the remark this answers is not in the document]
jmsmkn: Then the default should be an empty string and `null` removed. This is handled automatically by Django, so really, the `PATIENT_SEX_EMTPY` is not necessary.

app/grandchallenge/cases/models.py
HarmvZ: They are validated using regex, but only in the dicom/mhd -> SimpleITKImage conversion step. Do you want me to add an extra validation step in the `SimpleITKImage` -> `PanImg` conversion step? This should not make a difference currently, but if we add the extra fields to other image builders later and forget to add the validation, an extra validation step will catch that developer error.

app/grandchallenge/cases/models.py
jmsmkn: Urgh, that really limits the kind of queries that we can do. But, ok, if we need that then we can add a secondary field.

app/grandchallenge/cases/models.py
HarmvZ: If I understand correctly this value will consist of 5 components of each 64 characters max, delimited by the `^` character. So that's 5*64 + 4 = 324.

It's described here (PN value representation): http://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_6.2.html

app/grandchallenge/cases/models.py
jmsmkn: Can you try it? I think it will be caught by the `full_clean` here https://github.com/comic/grand-challenge.org/blob/122d60a8be7637df7f5b16b3520a7401b6f19e23/app/grandchallenge/cases/tasks.py#L477. However, the ValidationError that is raised there is not handled, so the entire session will fail with "An unexpected error occurred" and no images from the session will be imported. If it is handled in PanImg validation (either by truncation + warning or by rejection of that image) then the rest of the upload session is ok.

app/grandchallenge/cases/models.py
jmsmkn: Super! Please add a comment for a future developer there, and again I think this should be checked in the PanImg model so that we can send a nice warning to the user.

app/grandchallenge/cases/models.py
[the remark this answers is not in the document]
HarmvZ: Ah right. But the value does need to be defined in `choices` then, right? So then for clarity I would just keep it in here as well but change it to be `""` instead of `None`.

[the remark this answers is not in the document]
jmsmkn: It does not, see https://github.com/comic/grand-challenge.org/blob/122d60a8be7637df7f5b16b3520a7401b6f19e23/app/grandchallenge/workstation_configs/models.py#L164-L166 (although I think leaving out `default=""` only works as it was part of an initial migration, could be wrong).

app/grandchallenge/cases/models.py
HarmvZ: I added it to the PR as explained in this comment: https://github.com/DIAGNijmegen/rse-panimg/pull/51#issuecomment-930114858.

PR #2052, opened 2021-09-29
Adds rounded egdes for challenge, algorithm, archive and rs cards

amickan:
This is a first step towards redesigning the home page. 

Next to adding a css class for the cards, this also changes all absolute units in our css to relative units.


app/grandchallenge/core/static/css/core.css
```diff
@@ -168,3 +171,20 @@ ul.socialaccount_providers > li {
 .phaseButton:focus, .phaseButton:hover {
     box-shadow:none !important;
 }
+
+.card, .card-header {
+    border-radius: 20px;
+}
+
+.card:hover {
+    box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.2), 0 6px 20px 0 rgba(0, 0, 0, 0.19);
```
HarmvZ, line 180: Just curious, why did you use a double shadow here?

Styling nitpick: I like box-shadow without x/y offset making the shadow drop evenly around the element. Example:
Current: 
image

Even:
image

(Personal preference, do with it what you want)

HarmvZ (changes requested):
This is something I also struggle with a lot, but as a general rule it is better for accessibility to use relative units than absolute units. So in this case use `rem` instead of `px`. At least that was the case a few years ago, it's probably good to do some research about the current best practice as this seems to change a lot.

Also, for consistency it is better to also add rounded corners to the pagination buttons.

HarmvZ (approved):
Nice job on the `rem` stuff and on keeping the custom CSS to a minimum!

PR #2053, opened 2021-09-29
Replace reader study images with component interface values

MikeOverkamp-diag:
This replaces all direct links between reader studies / answers and images with links to `ComponentInterfaceValues`. This is an intermediary step towards implementing `DisplaySets`. On the outside, nothing is changed. For both users and clients it's still as if we use images directly.

**IMPORTANT: please do not merge / deploy this until the magement command can be run in production! Otherwise, all exisiting reader studies will not work properly.**


app/grandchallenge/reader_studies/serializers.py
```diff
@@ -109,20 +125,29 @@ def validate(self, attrs):
...
+            ):
                 raise ValidationError("Only the answer field can be modified.")
             question = self.instance.question
-            images = self.instance.images.all()
+            civs = self.instance.civs.all()
             creator = self.instance.creator
         else:
             question = attrs.get("question")
-            images = attrs.get("images")
+            civs = ComponentInterfaceValue.objects.filter(
```
jmsmkn, line 138: Could there be a problem here if there are multiple CIVs for an image? Maybe we need to filter by reader study?

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
```diff
@@ -379,7 +379,7 @@ <h2>Export</h2>
                     <a class="btn btn-primary"
                        href="{% url 'api:reader-study-detail' pk=object.pk %}?format=csv"
                        download="reader_study_{{ object.slug }}_{{ now }}.csv">
-                        <i class="fas fa-file-csv"></i> Reader Study
+                        <i class="fas fa-file-csv"></i> Reader StudyImages
```
jmsmkn, line 382:
```suggestion
                        <i class="fas fa-file-csv"></i> Reader Study
```

app/grandchallenge/reader_studies/views.py
```diff
@@ -890,7 +893,7 @@ class AnswerViewSet(
     queryset = (
         Answer.objects.all()
         .select_related("creator", "question__reader_study")
-        .prefetch_related("images")
+        .prefetch_related("civs")
```
jmsmkn, line 896: Did you check the Answer and ReaderStudy API views for the number of queries?

app/grandchallenge/reader_studies/serializers.py
MikeOverkamp-diag: Ah yes, possibly. Nice catch.

MikeOverkamp-diag:
One thing I'm unsure of now, mainly because of your comment: when an answer is created for an image, should we create a new civ for that image or reuse the civ from the reader study. I think there is some ambiguity there in the way I currently implemented this. But now I'm leaning towards always creating a new civ.

jmsmkn:
I think that they should be reused. Ideally, for each image, there is at most only one interface of each type, but I know that is not the case right now.

MikeOverkamp-diag:
I'll have a look at that first then. I think it makes sense that there is only one civ for an image.

jmsmkn:
I mean one Interface type for each image. An interface is only a label, right? So it should be able to have a label of a Generic and a CT Image, but not a Generic and a Generic. But, I know the latter happens.

MikeOverkamp-diag:
Yes, that's what I meant as well. Sorry, my answer was a bit short in that regard. Ideally we want the combination of interface/image or file (I'd say value is debatable, since I could see identical strings representing actual different values) to be unique.

MikeOverkamp-diag:
PR should be ready before the weekend.

jmsmkn:
Closing as we decided to take a different approach.

PR #2057, opened 2021-09-30
Adds 'about us' template and reorganizes footer

amickan:
Another small step towards redesigning the home page. This creates an About Us template page with the contents of the current 'contributors' flatpage. Once this is merged, the contributors flatpage should be removed. 

This also moves some of the links from the developer section to the new About Us page. 

The blog posts section will eventually be removed from the footer as well, but only once we add the link to the navigation bar so that the blogs remain accessible.

### DIFF
--- a/app/grandchallenge/core/templates/grandchallenge/partials/footer.html
+++ b/app/grandchallenge/core/templates/grandchallenge/partials/footer.html
@@ -9,8 +9,14 @@
             <hr class="m-0">
             <div class="row small">
                 <div class="col-sm-6 col-md-3 p-3">
-                    <h6>About</h6>
+                    <h6>Grand Challenge</h6>
                     <ul class="list-unstyled">
+                        <li class="my-1">
+                            <a class="text-muted"
+                               href="{% url 'about' %}">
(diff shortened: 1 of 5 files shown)


jmsmkn (approved):

app/grandchallenge/core/templates/grandchallenge/partials/footer.html
```diff
@@ -74,20 +80,6 @@ <h6>Developers</h6>
...
-                               href="https://rse.diagnijmegen.nl"
-                               title="Research Software Engineering Team, Diagnostic Image Analysis Group, Radboud University Medical Center">
-                                rse.diagnijmegen.nl
-                            </a>
-                        </li>
-                        <li class="my-1">
-                            <a class="text-muted"
-                               href="http://diagnijmegen.nl/"
-                               title="Diagnostic Image Analysis Group, Radboud University Medical Center">
-                                diagnijmegen.nl
```
jmsmkn, line 88: Wait, we remove RSE and DIAG but MeVis stays?

PR #2062, opened 2021-10-04
Home page redesign

amickan:
This is the last step towards redesigning the home page. 

- This adds a news carousel to the top of the page advertising the latest blog posts.
- It also adds sections advertising the latest challenges and algorithms.
- The feature overview has a different layout now.


app/config/settings.py
```diff
@@ -121,6 +121,9 @@ def strtobool(val) -> bool:
     "DOCUMENTATION_HELP_FORUM_SLUG", "general"
 )
 
+# Home page news blog tag
+HOMEPAGE_NEWS_BLOG_TAG = os.environ.get("HOMEPAGE_NEWS_BLOG_TAG", "homenews")
+
```
amickan, line 126: Ideally, we would use the 'news' or 'announcement' tag rather than 'homenews', but the former two are already in use on GC, hence the slightly strange tag here. Open for different suggestions as well.

app/grandchallenge/core/templates/home.html
```diff
@@ -2,68 +2,120 @@
 {% load url %}
 {% load static %}
 {% load humanize %}
-{% load cache %}
-{% load round_to %}
 
 {% block content %}
```
amickan, line 6: We could cache this entire section to reduce the number of times the database is hit for the latest challenges, algorithms and blog posts. I'm not sure how "expensive" these queries are. For the blog posts maybe it's nice if the content gets updated instantly though?

app/config/settings.py
jmsmkn: How about highlight?

app/grandchallenge/core/templates/home.html
jmsmkn: Yes this absolutely should be cached. If the data are delayed by 10 minutes (or 1 hour) it doesn't really matter IMO.

app/config/settings.py
jmsmkn: You could also use a model attribute for this, see my comment about Algorithms and Challenges too

app/grandchallenge/core/views.py
```diff
@@ -130,6 +155,14 @@ def get_context_data(self, **kwargs):
                     "A platform for end-to-end development of machine "
                     "learning solutions in biomedical imaging."
                 ),
+                "latest_challenges": Challenge.objects.filter(hidden=False)
```
jmsmkn, line 158: Rather than showing the latest challenges and algorithms maybe instead add another model attribute to display them as a highlight, then we have control over what appears there rather than users.

amickan: That's a nice idea. Who should be able to set this attribute though? If we don't want users to determine this, than it should be a hidden field that can only be set in the admin (i.e., by you), or is there a better option?

jmsmkn: Yes, add it to the model and `admin.py`, just don't set it in the object create and update views the editors get. We can add the permission bit to edit those to the staff members we want to do this, not just me.

app/config/settings.py
amickan: Yes, good idea, that's better, less chance for errors / confusion.

jmsmkn (approved):

PR #2064, opened 2021-10-04
Make help text for auto jump setting in viewer configuration scope reader studies

chrisvanrun:
Configuration has a new effect when applied to reader studies. This extends the help text to signify this.


miriam-groeneveld (commented):
I think you're missing a migration file?

chrisvanrun:
> 
> 
> I think you're missing a migration file?

Ups, an oversight on my side that it was part of the model!

miriam-groeneveld:
> > I think you're missing a migration file?
> 
> Ups, an oversight on my side that it was part of the model!

been there, done that :)

miriam-groeneveld (approved):

PR #2066, opened 2021-10-05
Upgrade to panimg 0.5.1

chrisvanrun:
This PR updates panimg from 0.4.1 -> 0.5.2. This covers these changes:
* Fix default values for extra metadata fields
* Added extra metadata fields to PanImg object
* Allow array-like window level values in MHA files
* Simplified 4D ITK loading
* Added support for nrrd files

In addition, selective tests have been copied/changed from panimg to cover breaking changes in the API.


HarmvZ:
Are the tests in here duplicates of the ones in PanImg? In principle it should not be necessary to test the inner workings of PanImg in this repo, right?

nlessmann:
With this upgrade, .nrrd files would be automatically supported, right? Then it should probably also be added here: https://github.com/comic/grand-challenge.org/blob/2fd7d64567d8a0f9a1c266535899e2ff809d0fe8/app/grandchallenge/cases/forms.py#L17-L26

jmsmkn (approved):

chrisvanrun:
> 
> 
> Are the tests in here duplicates of the ones in PanImg? In principle it should not be necessary to test the inner workings of PanImg in this repo, right?

Right! I pondered the same but in the words of James: 'yes, can be removed once the API has stabilised'.

PR #2067, opened 2021-10-05
add meta_social block to blogpost

Kicksta:


app/grandchallenge/products/templates/products/post_detail.html
```diff
@@ -13,6 +13,15 @@
     <li class="breadcrumb-item active">{{ object.title }}</li>
 {% endblock %}
 
+{% block meta_social %}
+  <meta name="title" content="{{ object.title }}">
+  <meta property="og:title" content="{{ object.title }}" />
+  <meta property= "og:type" content="website"/>
+  <meta property="og:url" content="//www.AIforRadiology.com/" />
```
MikeOverkamp-diag, line 20: Do you not want this to point to the actual post?

Kicksta: You mean the url? I will change it. The funny thing is that for the product_detail the og:url is also the homepage, but if i create a post on a specific product it does redirect correctly to the product and not to the homepage.

Kicksta: I pushed the change

PR #2079, opened 2021-10-08
Add object deletion and size checks

jmsmkn:
This PR cleans up the storage when the objects on the grand challenge side are deleted, adds checks for the total size of the users upload to protect against users uploading a bunch of data (verified users get more), and handles forbidden errors on the client-side.

I'm unsure if going to S3 each time to calculate the size is the best approach, we could always make `size` a field and update it when necessary, then do the calculations within the Django database, which would remove the `get_creators_completed_uploads` call. As uppy uses put requests we cannot limit the size of the parts, nor can we trust any of the metadata handed to us by the client. This can be changed later if needed.

### DIFF
--- a/app/grandchallenge/uploads/static/js/user_upload.js
+++ b/app/grandchallenge/uploads/static/js/user_upload.js
@@ -122,6 +122,8 @@
         ).then(upload => upload.parts)
     }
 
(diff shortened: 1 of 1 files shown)


jmsmkn:
Thanks!

PR #2118, opened 2021-10-15
Adds animations to homepage

HarmvZ:
### Description

As a friday afternoon project I played around with adding some animations to the new homepage. This is what I ended up with:
- Counting animation for the stats in the jumbotron. 
- Sliding in animation for the cards and features elements on scroll into viewport.

Another idea was to add a [parallax effect](https://codepen.io/silvandiepen/pen/NOboze) on the background image of the jumbotron. This requires a lot of changes to the page structure however so I skipped it for now.

I think these kind of subtle animations make the homepage look more modern and dynamic. However, if this is not what we want we can simply close the PR again. 

### Performance

I tried to keep performance in mind so for the counting animation I wrote a custom vanilla JS function. For the sliding into viewport animation stuff I used [scrollReveal.js](https://scrollrevealjs.org/), a small (~6kB) library. It should not have a huge effect on load time. The animations are not computationally heavy, on my pc they run smoothly at 60fps. 

### Demo
See below for a demo. In the videos it seems to stutter but that is cause by the screen recording software, on my pc it runs smoothly (~60fps). 

**Desktop**

https://user-images.githubusercontent.com/2001094/137505590-34ce04d1-2cb4-4479-ab39-5e56d4fb644e.mp4

**Mobile**

https://user-images.githubusercontent.com/2001094/137505612-faf047aa-6c18-4d28-b368-e48771083c6d.mp4


amickan:
I consulted both Bram and Alex about adding these animations to the homepage. They both felt that less is more in this case, so I'm closing this PR for now.

PR #2124, opened 2021-10-21
Add contact email to challenge

### LINKED ISSUE #2123: Challenge contacts
At the moment, a challenge's contact email address is automatically the personal email address of the creator of the challenge. This email address is shown on the contact page of a challenge, which however is by default only visible to registered participants of the challenge.

It would be better to have an official and public contact-email, which can be seen by everyone (not just participants) and which the admins of a challenge can choose freely. To faciliate the contact, it would then be nice to add an email button to the nav-tabs in the info section that triggers an email to this contact email address of choice. 

This email button could be added after all page tabs in the Info section and could be labeled "Email organizers" or just contain an email icon. 

Inked2021-10-20 17_24_33-Mitosis Domain Generalization Challenge 2021 - Grand Challenge_LI

To achieve this, we will need to add a `contact_email` field to the `challenge `model. The email button will only be shown if this field is filled. 

- On the database level, this field would initially be blank by default. 
- In the challenge update view form, this field would be required though so that newly created challenges and updated challenges are guaranteed to have the field filled. 
- In the helptext for this field, we need to explain that this email will be shown publically and will be the primary contact email. 

For challenges that are no longer updated, this field will need to filled manually. Once all challenges have this field filled, we can make the field required on the database level as well.

amickan:
- This adds a contact_email field to internal challenges
- This field is required in the create and update form, but not yet on the database level 
- If the field is filled, an email is displayed under the info tab of the challenge:

image

Closes #2123


amickan:
I find the fact that there is both a `Contact` page and an `Email organizers` now confusing. Maybe we could at least hide the `Contact` page if the user does not have permission to see it? We could then do that for all restricted access pages (hiding them rather than showing the locked symbol).

PR #2151, opened 2021-11-01
Remove retina thumbnail endpoints

HarmvZ:


jmsmkn (approved):
LGTM - I've made a few changes so that the Image model never loads from object storage (removing a few properties that are only used by the tests, the settings of those values will be covered in panimg), and removed the retina core application completely.

jmsmkn:
@HarmvZ please check my extra changes here.

HarmvZ (commented):
Retrospective LGTM

PR #2152, opened 2021-11-01
Adds chart output to results list view

amickan:
- This adds a chart column to the results list view if the algorithm has a chart output interface. 
- All chart outputs are then represented by chart icons within this one column. 
- Hovering over them shows the interface title and clicking on them opens a new window with the job's detail view. This means that all icons for a given job lead to the same page. _We could add hashes to the url to direct to the specific section of the detail page where the respective chart is located if we want more specificity._ 

image

No N+1 queries :-)


jmsmkn (approved):
Cool!

PR #2159, opened 2021-11-02
Turn off showProgressDetails

### LINKED ISSUE #2158: Upload progress gets confused when doing multiple parallel uploads
**Describe the bug**

When uploading multiple images on the "Try out this algorithm page" in parallel, the upload widgets gets confused and seem to show the other's progress intermittently. See video:

https://user-images.githubusercontent.com/11632379/139873202-1caeadcc-53ef-451e-bb19-6d36ace5f452.mp4

**Expected behavior**

Progress increases and does not decrease

 - OS: Linux
 - Browser Firefox
 - Version 93.0

**Additional context**

Does not seem to be a serious issue. The uploads ultimately finish. Still, a bit confusing.

jmsmkn:
The file size and remaining time details are all buggy, so remove them.
Add ids for the UI Elements as this was missing.

Closes #2158

### DIFF
--- a/app/grandchallenge/uploads/static/js/user_upload.js
+++ b/app/grandchallenge/uploads/static/js/user_upload.js
@@ -25,12 +25,14 @@
         })
 
         uppy.use(Uppy.DragDrop, {
+            id: `${inputId}-DragDrop`,
             target: `#${inputId}-drag-drop`,
         });
 
         uppy.use(Uppy.StatusBar, {
+            id: `${inputId}-StatusBar`,
             target: `#${inputId}-progress`,
-            showProgressDetails: true,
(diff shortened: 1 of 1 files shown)


HarmvZ:
[According to the docs](https://uppy.io/docs/status-bar/#id-39-StatusBar-39) this issue should be fixed already just by adding the id so then it does not need to be hidden. It seems likely that it will work since the progressbar is also not buggy. Did you try this?

jmsmkn:
I tried just with the ID and that didn't help, and we're on the latest version of Uppy still.

HarmvZ (approved):
Ok then it's probably a bug on their side indeed.

PR #2160, opened 2021-11-03
Add thumbnails to results list, make table responsive

amickan:
This adds thumbnails, floats, integers, strings and booleans to the results list and download buttons for all outputs of type file. 
image

The table is responsive in that columns that don't fit the viewport, are hidden and can be made visible through the toggle button in the first column. 

This also adds floats, integers, strings and booleans to the detail view inthe form of a table. 
image

Video demo:

https://user-images.githubusercontent.com/30069334/140719127-87c9c24c-919a-4462-ac9b-ea99f76f8e84.mp4


app/grandchallenge/algorithms/models.py
```diff
@@ -581,19 +581,46 @@ def run_job(self, upload_pks=None):
...
+        return outputs
+
+    @cached_property
+    def output_interface_dictionary(self):
+        outputs = {
+            "CHART": {},
+            "PDF": {},
+            "TIMG": {},
+        }
+        for kind, title in self.output_interfaces.values_list("kind", "title"):
```
amickan, line 616: This results in duplicate queries (one for each job), but I cannot think of a way around this right now. First creating a dictionary with all output interfaces that the algorithm has (and only then filling them with the interface values) is necessary (I think) because not all jobs for an algorithm might have values for all configured interfaces, and if we just loop over all existing interface values per job, a given value might end up in the wrong column...

app/grandchallenge/algorithms/models.py
```diff
@@ -581,19 +581,46 @@ def run_job(self, upload_pks=None):
...
-                outputs["CHART"].append(output)
+                for title in outputs["CHART"]:
+                    if title == output.interface.title:
+                        outputs["CHART"][f"{title}"] = output
             if output.interface.kind == InterfaceKind.InterfaceKindChoices.PDF:
-                outputs["PDF"].append(output)
+                for title in outputs["PDF"]:
+                    if title == output.interface.title:
+                        outputs["PDF"][f"{title}"] = output
+            if output.interface.kind in (
```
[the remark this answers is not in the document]
amickan: True!

app/grandchallenge/algorithms/models.py
jmsmkn: Yes, we need to avoid that. Why not fill out the data for each job and then filter when you fill out the columns in the view instead? Also, do not use titles for indexes, always use slugs.

app/grandchallenge/algorithms/views.py
```diff
@@ -555,6 +556,19 @@ def algorithm(self):
 
     def get_queryset(self):
         queryset = super().get_queryset()
+
+        def get_annotation(key):
```
jmsmkn, line 560: Is there a reason for defining a function here rather than doing it inline?

app/grandchallenge/algorithms/views.py
```diff
@@ -593,21 +607,63 @@ def columns(self):
...
-                Column(title="Chart", sort_field="", classes=("nonSortable",)),
-            )
+        for grouped_interfaces in self.outputs_list_display.values():
+            for interface in grouped_interfaces:
+                if interface.kind in (
+                    InterfaceKind.InterfaceKindChoices.STRING,
+                    InterfaceKind.InterfaceKindChoices.INTEGER,
+                    InterfaceKind.InterfaceKindChoices.FLOAT,
+                    InterfaceKind.InterfaceKindChoices.BOOL,
+                ):
```
jmsmkn, line 617: Nitpick: Rather than having the same list of things here what you could do is take the key in L610 by looping over the items (`for key, grouped_interfaces in self.outputs_list_display.items():`) and then checking `if key == "JSON"`. Makes it easier for a future developer to change in the future.

jmsmkn (approved):
You made sorting these look straightforward, great work!

app/grandchallenge/algorithms/views.py
amickan: For some reason, doing it inline did not work. I always got a TypeError:

image


`TypeError: QuerySet.annotate() received non-expression(s): {'probability-covid-19': <django.db.models.expressions.Subquery object at 0x7fc70ce41e80>}.`

jmsmkn: You;re not unpacking the dictionary there, how about `.annotate(interface.slug=Subquery(...))`?

jmsmkn: I think that you're missing a `*` before the `{` to unpack the dictionary. I'm also thinking that you could do this in one annotate call by building up the dictionary (eg with `{"slug1": ..., "slug2": ...}`) and then calling `queryset.annotate` only once after the for loop. However, thinking about it I'm wondering how performant it is as it will be doing 1 subquery per interface, we can always try it and see.

amickan: Ah right, unpacking does the trick ;-) Annotating in one call also work, but it does not make a performance difference (same number of db calls). It's a bit easier to read though, so II pushed that change. In terms of performance, it actually only does the subqueries when the outer queryset is evaluated, so it does not add any extra db calls, as far as I can tell.

jmsmkn (approved):
Looks great! I don't know if in the detail table we should call it interface and value, would metric and value be better? Might be confusing. Also would be good to lose the quotes around string values. Anyway, nothing blocking, lets see how this performs.

PR #2164, opened 2021-11-09
Add newsletter field to profile, show modal for newsletter sign-up

### LINKED ISSUE #2163: Ask users whether they want to be put on mailing list
Currently, under the GDPR rules, we are not able to send email updates to our users on Grand Challenge, because we do not have their explicit permission to do so. In order to make it possible to reach out to the user base in the future, it would be good to ask for their consent in retrospect. 

To do so, add a Boolean `receive_newsletter` field to the profile model.

For existing users, show a pop-up upon next log-in that asks for whether they want to be put on our newsletter mailing list and store their choice in the database. 

For new users, make this field obligatory upon sign-up.

amickan:
- Adds `receive_newsletter` to profile model
- Adds the field to the profile form and the sign-up form
- Adds a modal that is shown to the user when the `receive_newsletter` field is not filled yet.
- For the updating of the field to work through the API, users need not only the `change_userprofile` object permission (which they already have), but also the `view_userprofile` object permission and the global permission `profiles.change_userprofile`. To grant the object level permissions for existing users, I added a management command. 

image

Closes #2163


amickan:
Locally all tests pass. Not sure why some are failing here. Also, in the last run a different test failed than in the run before.

jmsmkn:
I think it's just a flaky test, I have re-run the CI jobs.

jmsmkn:
Closing in favor of #2179

PR #2171, opened 2021-11-18
Add model facts fields to algorithm

amickan:
- This adds the `model facts` fields to algorithms (based on this [paper](https://www.nature.com/articles/s41746-020-0253-3)) as optional, free text fields. 
- This also adds a required `contact email` field as well as a required `display_editors` field allowing the user to indicate whether editors should be displayed on the algorithm detail page or not (I believe we cannot do this by default because some algorithms need to remain anonymous).

These new fields (together with a list of inputs and outputs and the last updated date) are then printed on the algorithm Information tab. 
 
When the two new required fields are not filled yet, users will get an alert to update their algorithm:
image

Once filled in, it looks like this: 

image


app/grandchallenge/algorithms/forms.py
```diff
@@ -40,6 +44,10 @@
 from grandchallenge.subdomains.utils import reverse_lazy
 
 
+class ModelFactsTextField(Field):
+    template = "algorithms/model_facts_field.html"
```
amickan, line 48: The help texts are bit longer for the model facts fields and we did not want them to get lost underneath the markdown helptext that is printed for each field already. Hence the custom template.

app/grandchallenge/algorithms/forms.py
```diff
@@ -205,11 +230,109 @@ class Meta:
...
+            "You might want to consider listing them in a table like this:<br><br>"
+            "| Error message | Solution | <br>"
+            "| --------- | ----------- | <br>"
+            "| error 1 | solution 1| <br>",
         }
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.helper = FormHelper(self)
+        self.helper.attrs = {"novalidate": "novalidate"}
```
amickan, line 256: This is necessary to make validation work on the tabbed form. Required fields cannot be validated when a user submits the form from another form tab. This disables the standard validation. Instead we now validate the fields server-side (see clean method). I think the alternative here would have been to make the required fields optional, but that didn't seem like a better idea. A third option would be to split the form up into 2 seperate forms (then we'd no longer have the tabs) or put everything on one page (but that makes the form very long).

app/grandchallenge/algorithms/templates/algorithms/algorithm_form.html
```diff
@@ -25,3 +25,17 @@ <h2>{{ object|yesno:"Update,Create" }} Algorithm</h2>
...
+    <script>
+        if (document.getElementById("id_display_editors").value == ''){
+            document.getElementById("id_display_editors").classList.add('is-invalid');
+            const alert = document.createElement('span');
+            alert.id = "error_1_id_display_editors";
+            alert.className = "invalid-feedback"
+            alert.innerHTML = "<strong>This field is required.</strong>";
+            document.getElementById("div_id_display_editors").children[1].insertBefore(alert, document.getElementById("hint_id_display_editors"));
+        }
+    </script>
```
amickan, line 40: Unlike the other required fields, the `display_editors` field was not marked red after validation (only the alert was shown). This is a work around to have the same markup as for the other required fields. This also makes finding the missing input easier for the user.

app/grandchallenge/algorithms/forms.py
jmsmkn: I think the best thing to do here is to split this up into two forms, one for the algorithm with only the very basic data, and another for the algorithm metadata. Otherwise, you are having to duplicate a lot of work that Django should give you out of the box with form validation.

app/grandchallenge/algorithms/forms.py
jmsmkn: You can create a docstring for this new class to let future developers know this :-)

jmsmkn (approved):

PR #2175, opened 2021-11-22
Add slug to documentation admin

jmsmkn:


jmsmkn:
Ok, it's not an editable field, that is protection I'd like to keep, so, forget it...

app/grandchallenge/documentation/admin.py
```diff
@@ -11,6 +11,7 @@ class Meta:
         model = DocPage
         fields = (
             "title",
+            "slug",
```
amickan, line 14: I think this does not work because `AutoSlugField` is not editable. To still show it, you could add it to `readonly_fields` in the `DocPageAdmin`. Alternatively, we might need to change the slug field in the DocPage model, I'm afraid.

PR #2177, opened 2021-11-22
Add website url validation to user profile

amickan:


app/grandchallenge/profiles/forms.py
```diff
@@ -50,6 +51,13 @@ def save(self, *args, **kwargs):
 
         return instance
 
+    def clean(self):
+        cleaned_data = super().clean()
+        url = cleaned_data.get("website")
+
+        if not url.startswith("https"):
```
jmsmkn, line 58:
```suggestion
        if not url.startswith("https://"):
```

If we don't include the `://` here then the same linking problem will still exist. Also add this to the command on L13.

app/grandchallenge/profiles/forms.py
```diff
@@ -50,6 +51,13 @@ def save(self, *args, **kwargs):
 
         return instance
 
+    def clean(self):
+        cleaned_data = super().clean()
+        url = cleaned_data.get("website")
+
+        if not url.startswith("https://"):
+            raise ValidationError("Your url needs to start with https://")
```
jmsmkn, line 59: Missed this the first time around, but it's not crucial - as you're only dependent on the website field you do not need to use clean and can define `clean_website` instead, this way the errors appear with the field too, see https://docs.djangoproject.com/en/3.2/ref/forms/validation/#cleaning-a-specific-field-attribute. Note that the API is slightly different and you need to return a value.

jmsmkn (approved):

app/grandchallenge/profiles/forms.py
amickan: Ah yes, that's much nicer, I changed it to that ;-) Thanks!

jmsmkn (approved):

PR #2178, opened 2021-11-23
Trigger resize event when toggling extra results columns

amickan:


app/grandchallenge/evaluation/static/js/evaluation/leaderboard.js
```diff
@@ -214,6 +214,7 @@ function getDataTablesButtons() {
                         dt.columns('.toggleable').visible(true);
                         $(node).addClass('metrics-hidden');
                         $(node).text('Hide additional metrics');
+                        setTimeout($('#ajaxDataTable').DataTable().columns.adjust, 300);
```
jmsmkn, line 217: Not too sure that I understand the solution here. Why 300ms? Is this to catch the response from the server? If so, maybe there is another signal we can hook into instead? https://datatables.net/reference/event/

amickan: Not really. We can use the `button-action` event, which is fired when a user clicks on `Show all metrics`. However, even if we issue the `columns.adjust` call there, nothing happens, unless I also add a timeout. I don't have an explanation for why that is unfortunately. I'm not sure if hooking into this event is better than what I had before. The event is triggered after the button's action, but it's not clear to me if this means that all data has been loaded when this event is fired as well. If so, then using this event would be better than what I did before.

jmsmkn: What about using the 'column-visibility' which will be fired when `dt.columns('.toggleable').visible(true);` completes (I think)? https://datatables.net/reference/event/column-visibility

amickan: Actually turns out that the problem was the `scrollX` property on the table ;-)

jmsmkn: Ah much easier!

jmsmkn (approved):

PR #2179, opened 2021-11-23
Add newsletter sign-up form

amickan:
Same functionality as in https://github.com/comic/grand-challenge.org/pull/2164 , but implemented with a Django form rather than through the Rest API. 

Regarding the tests: I spent a lot of time trying to change the `UserFactory()` so that the newsletter field on the `user_profile `is always automatically True, but did not get it to work unfortunately, hence the repetitive code in the tests.


jmsmkn:
> Regarding the tests: I spent a lot of time trying to change the UserFactory() so that the newsletter field on the user_profile is always automatically True, but did not get it to work unfortunately, hence the repetitive code in the tests.

The `UserFactory` was a bit of a red herring - the tests were catching a real bug in the implementation: `{% load url %}` was missing from app/grandchallenge/core/templates/grandchallenge/partials/newsletter_signup_modal.html which caused a URL resolver error on views on subdomains for logged in users. I think that I have resolved it.

jmsmkn (dismissed):
Looks good! Small suggestions for improvements, up to you whether to include them.

jmsmkn (approved):

PR #2182, opened 2021-11-24
make the filter 'other' work for modalities and subspecialities

Kicksta:


MikeOverkamp-diag (approved):

jmsmkn (approved):
LGTM! A note on style and apologies if you're already aware of this and I'm missing something - you do not need to use Q objects for simple filters: you could do `queryset.exclude(subspeciality__in=self.subspecialities)` for a not filter or `queryset.filter(subspeciality__icontains=subspeciality_query)` for a normal filter.

PR #2185, opened 2021-11-25
Update navbar

Kicksta:
Split up about page in several pages and create dropdown for about


app/grandchallenge/products/templatetags/products_tags.py
```diff
@@ -22,23 +22,33 @@ def navbar(context):
...
-                "title": "Project AIR",
-            },
             {
                 "url": "blogs-list",
-                "active": url == "blogs-list",
+                "active": url in ["blogs-list", "blogs-detail"],
                 "title": "Blogs",
             },
+        ],
+        "contact": {
```
MikeOverkamp-diag, line 31: Why not add these tho items as well, as they're also items. For About you could then add a `subitems` key and simply add the html for the subitems in an `{% if %}` block in the template.

MikeOverkamp-diag (changes requested):

app/grandchallenge/products/templatetags/products_tags.py
Kicksta: Let's have a look at it together tomorrow!

Kicksta: fixed

MikeOverkamp-diag (approved):

PR #2190, opened 2021-11-26
Adds publication citation and optional image to info page

amickan:
- Adds publication citation and optional image to info page
- Moves the original detail page markdown text down below About


app/grandchallenge/algorithms/templates/algorithms/algorithm_detail.html
```diff
@@ -128,10 +128,12 @@ <h2 class="col-9 p-0">{{ object.title }}</h2>
...
             </div>
             <hr>
-
-            {{ object.detail_page_markdown|md2html }}
-
-            <h3 class="mb-3">About</h3>
+            {% if object.image_on_information_page == 'LOGO' %}
+                <img class="w-50" loading="lazy" src="{{ object.logo.url }}" alt="Logo for {{ object.title }}">
+            {% elif object.image_on_information_page == 'SOCIAL_IMG' %}
+                <img class="w-100" loading="lazy" src="{{ object.social_image.url }}" alt="Social image for {{ object.title }}">
```
jmsmkn, line 134: These will break if the `logo` or `social_image` does not exist, and this is not something that can be caught by validation as the images can be removed independently of the image choice. IMO I wouldn't add an option on the model for this, I missed the discussion on whether to add it but I would prefer consistency between algorithms, and therefore show the logo every time. User can always add more images by including them in their information page markdown.

jmsmkn: I see. As a general approach: if you add too many options it gets difficult for the users to use and for us to maintain, so I would always think about how you can do things without adding another parameter. This also forces consistency which I value a lot (otherwise it feels a bit like MySpace 2004). Have a think about how you would like the layout of the page to be, if we are going to make it look like the model facts label then a logo is a nice addition to that.

kiranvaidhya: I agree that we should reduce the number of options to maintain consistency. I am fine with GC automatically rendering the logo on the main page, above the About section.

amickan: Alright, I agree as well. I removed image choice field and changed the template so that we always print the logo. I played around with showing the logo to the right of the about section, but since we have some information that can span multiple lines in the about section (input and output descriptions, publication citation), it looks better to just display the logo on top.

jmsmkn (approved):

PR #2193, opened 2021-11-27
Automatically remove access request notifications

nlessmann:
I'd like to improve the access request workflow for algorithm editors, reader study editors etc. a little bit. The current workflow is cumbersome as it requires many clicks: (1) view the notification (2) go to the algorithm (3) go to the list of access requests (4) go to the request update form (5) accept or reject the request (6) go back to the notifications list (7) delete the notification or mark it as read.

This PR removes the need for the last two clicks by automatically deleting notifications that editors or challenge admins receive when there is a new access request. These notifications are removed when the access request has been taken care of, i.e., when it no longer has status "pending".

If this is not how notifications should behave, please just close the PR.

I've added this only in one place (`core/signals.py`) while most of the logic for challenge participation requests seems to be separate from the logic for other kinds of access requests. Therefore, I was doubting whether this should be separately added for `RegistrationRequest` objects to `participants/signals.py`. Instead, I've now added a `base_object` property to that class to be able to handle them all in one place.

### DIFF
--- a/app/grandchallenge/core/signals.py
+++ b/app/grandchallenge/core/signals.py
@@ -87,6 +87,32 @@ def process_permission_request_update(sender, instance, *_, **__):
             )
 
 
+@receiver(post_save, sender=AlgorithmPermissionRequest)
+@receiver(post_save, sender=ReaderStudyPermissionRequest)
+@receiver(post_save, sender=ArchivePermissionRequest)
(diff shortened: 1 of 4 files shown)


amickan (approved):
LGTM! And I agree that handling all 4 types of access requests in one place makes sense!

PR #2194, opened 2021-11-29
Changes sorting of viewer and viewer config to sort by title.

miriam-groeneveld:
Especially the list of viewer configs is getting really long, making it hard to find the correct config


app/grandchallenge/algorithms/views.py
```diff
@@ -617,7 +617,7 @@ def columns(self):
             Column(title="Result", sort_field="inputs__image__name"),
             Column(title="Comment", sort_field="comment"),
             Column(title="Visibility", sort_field="public"),
-            Column(title="Viewer", sort_field="inputs__image__files__file"),
+            Column(title="Viewer", sort_field="title"),
```
jmsmkn, line 620: The sort field here is the one that is used on the model, in this case an Algorithm `Job`. A job does not have a `title` field, so in this case this will break the sorting of the table when the user clicks on Viewer. IMO this change should be reverted.

app/grandchallenge/algorithms/models.py
```diff
@@ -97,7 +97,7 @@ class Algorithm(UUIDModel, TitleSlugDescriptionModel):
     additional_terms_markdown = models.TextField(
         blank=True,
         help_text=(
-            "By using this algortihm, users agree to the site wide "
+            "By using this algorithm, users agree to the site wide "
```
jmsmkn, line 100: If you change anything on a model field you also need to regenerate the migrations. In the case of a help text change this is one of the few that do not change the DB schema, if you're confident of that (test it) you can also correct the typo in the last migration where this field was updated. But, adding a new migration is the easiest!

app/grandchallenge/workstation_configs/models.py
```diff
@@ -257,7 +257,7 @@ class ImageInterpolationType(models.TextChoices):
     )
 
     class Meta(TitleSlugDescriptionModel.Meta, UUIDModel.Meta):
-        ordering = ("created", "creator")
+        ordering = ("title",)
```
jmsmkn, line 260: This requires a migration file to be generated and committed.

jmsmkn: But actually, IMO it's best not to change the default ordering and better to do this in the view instead.

app/grandchallenge/algorithms/views.py
miriam-groeneveld: ah, ok. But a strange sorting then?

app/grandchallenge/algorithms/models.py
miriam-groeneveld: Ah, just saw a typo. Will create a migration.

app/grandchallenge/algorithms/views.py
jmsmkn: I guess! I don't know what the user would expect when sorting by Viewer. Maybe `pk` makes more sense, or `inputs__image__name`?

miriam-groeneveld: > 

Well, I thought I would try that out, and it looks like something really strange happens when sorting by viewer, see video. The list shows all results (14 out of 14), but sorting on viewer seems to change the results (just look at the number of results that have failed).

https://user-images.githubusercontent.com/51329283/143855916-53f5a53e-8f49-4671-90e3-2a35ab00ba5a.mp4

jmsmkn: Weird, maybe you start to get duplicates when sorting by `inputs__image__files__file` as there are many files per image, so a `distinct` is missing somewhere. Maybe `status` is the best thing to go with?

jmsmkn (approved):

PR #2196, opened 2021-11-29
Add new default internal challenge view

### LINKED ISSUE #2112: Seperate external challenges from internal challenges
It would be nice to seperate internal from external challenges a little more in the list view. 

The simplest (?) way would be to only include internal challenges in the list view and add a seperate list view for external challenges. This seperate list view could be linked to through a button on the normal challenge list view.

amickan:
This replaces the `/challenges` page with a listview of internal challenges only. The page links to the old challenge list that includes both internal and external challenges. 

I'm not restricting the view with permission because I assume we want everyone to be able to see this pages (also not-registered users). 

Closes #2112

### DIFF
--- a/app/grandchallenge/challenges/urls.py
+++ b/app/grandchallenge/challenges/urls.py
@@ -3,6 +3,7 @@ from django.urls import path
 from grandchallenge.challenges.views import (
     ChallengeCreate,
     ChallengeList,
+    CombinedChallengeList,
     ExternalChallengeCreate,
     ExternalChallengeList,
     ExternalChallengeUpdate,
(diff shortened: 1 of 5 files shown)


app/grandchallenge/challenges/urls.py
```diff
@@ -13,6 +14,9 @@
 
 urlpatterns = [
     path("", ChallengeList.as_view(), name="list"),
+    path(
+        "all-challenges", CombinedChallengeList.as_view(), name="combined-list"
```
jmsmkn, line 18: Missing trailing slash. Rest looks great!

jmsmkn (approved):

PR #2198, opened 2021-11-30
Adds status bagde to challenge cards

### LINKED ISSUE #2111: Challenge status markers
For visitors of Grand Challenge it would be nice to know which challenges are still ongoing and which have already ended. 

As far as I can tell we currently have no way of knowing whether a challenge is ongoing or not. One way of implementing this would be to add `start` and `end` date fields to the Phase model, which need to be filled in by the challenge organizers for each of their phases. 

We could then mark challenges as either closed or ongoing in the top corner of the challenge card, and add information on how many days there are left for submitting solutions to a specific phase in the card body. 

Once this is in place, we should also enable filtering challenges by their status and order challenges according to whether they are ongoing (and secondly by creation date).

amickan:
This adds a status badge to all challenge cards, indicating whether they are currently open to submissions (in any phase) or not. 
This also enables filtering internal challenges on their status and it fixes the count of the challenges. 

I got feedback on the looks of the badge from Bram and the 10% team, rather than superimposing the badge on the card-img (as I proposed in the milestone pitch), we agreed to just add it as the first badge in the card-body. 

RE implementation: I also played around with annotating the queryset instead of adding `accepting_submissions` as a property. Both work fine and are no different in terms of db queries as far as I could tell, but I opted for implementing this as a property to be able to add it to the filter easily (much more complicated to enable filtering on an annotated queryset it seems) and also in light of the possible changes we might make in the future. 

Closes #2111

### DIFF
--- a/app/grandchallenge/challenges/filters.py
+++ b/app/grandchallenge/challenges/filters.py
@@ -1,4 +1,4 @@
-from django_filters import ModelMultipleChoiceFilter
+from django_filters import ChoiceFilter, ModelMultipleChoiceFilter
 from django_select2.forms import Select2MultipleWidget
 
 from grandchallenge.challenges.models import (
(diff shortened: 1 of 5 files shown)


jmsmkn (approved):
This looks good!

Indeed this is a trade-off: it's quite simple to write the rules in python, but I think that in `filters.py` you will need to retrieve potentially all challenges from the database. Performance-wise it is better to do the work there, or to create a property on the challenge model that is updated whenever the phases change, but, we don't have that many public challenges and the filter may be applied last which will help, lets see if we need that extra complexity.

jmsmkn:
Rolled this out and the performance is fine, onwards!

PR #2201, opened 2021-12-03
Use phase start and end dates for challenge status badge

amickan:


app/grandchallenge/challenges/models.py
```diff
@@ -558,9 +559,67 @@ def remove_admin(self, user):
 
     @property
     def accepting_submissions(self):
```
jmsmkn, line 561: A general comment: this method is hard to follow with a lot of nested if statements. Can you think of a way to simplify it? Maybe split it out into some smaller methods?

app/grandchallenge/challenges/models.py
```diff
@@ -558,9 +559,67 @@ def remove_admin(self, user):
...
+                    for phase, bool in phase_details.items()
+                    if bool
+                ]
+        else:
+            status_strings = [
+                phase.submission_status_string
+                for phase in phase_details.keys()
+            ]
+            completed_phases = [
+                string for string in status_strings if "completed" in string
```
jmsmkn, line 584: Checking for substrings in strings is going to be fragile. I think a better way would be to add a property to the phase that is an enum to handle all of these cases, eg all the ones found in `submission_status_string`:

- Closed
- Opening Soon
- Open
- Completed

app/grandchallenge/evaluation/forms.py
```diff
@@ -124,8 +126,28 @@ class Meta:
             "extra_results_columns": JSONEditorWidget(
                 schema=EXTRA_RESULT_COLUMNS_SCHEMA
             ),
+            "submissions_open": forms.DateInput(
```
jmsmkn, line 129: Does it make sense to use a DateInput here? I think we use a DateTimeField for both of the deadlines, so that the challenge admin can set a time of say 5pm on this date.

app/grandchallenge/evaluation/forms.py
```diff
@@ -124,8 +126,28 @@ class Meta:
...
         }
 
+    def clean(self):
+        submissions_open = self.cleaned_data["submissions_open"]
+        submissions_close = self.cleaned_data["submissions_close"]
+
+        if (
+            submissions_open
+            and submissions_close
+            and not submissions_close > submissions_open
```
jmsmkn, line 144: Nitpick: easier to read if the bool is not negated: `and submissions_close < submissions_open`

app/grandchallenge/evaluation/models.py
```diff
@@ -582,9 +591,49 @@ def has_pending_evaluations(self, *, user):
...
+            submission_period_open = (
+                now > self.submissions_open and now < self.submissions_close
+            )
+        elif self.submissions_open:
+            submission_period_open = now > self.submissions_open
+        elif self.submissions_close:
+            submission_period_open = now < self.submissions_close
+        else:
+            submission_period_open = True
+        return submission_period_open
```
jmsmkn, line 607: Again I find this hard to follow, can you simplify it? I would suggest something like:

```
now = timezone.now()
upper_bound = self.submissions_close or now + timedelta(days=1)
lower_bound = self.submissions_open or now - timedelta(days=1)
return lower_bound < now < upper_bound
```

And is `submission_period_open` the best name? I don't know if it clashes with `submissions_open`. Maybe we should change `submissions_open` to `submissions_open_at` or this method to `is_in_submission_period`? Thoughts welcome :-)

app/grandchallenge/evaluation/models.py
```diff
@@ -582,9 +591,49 @@ def has_pending_evaluations(self, *, user):
...
+
     @property
     def open_for_submissions(self):
-        return self.submission_limit != 0
+        if self.submission_limit == 0:
+            return False
+        elif not self.submission_period_open:
+            return False
+        else:
+            return True
```
jmsmkn, line 616: I think this can all simplify to `return self.submission_period_open and self.submission_limit > 0`

jmsmkn (commented):
This looks good! It's a lot of complex business logic that I found hard to follow so I suggested some simplifications for future developers.

app/grandchallenge/challenges/models.py
```diff
@@ -557,10 +558,51 @@ def remove_admin(self, user):
...
+            ]
+            if len(detail) > 1:
+                # if there are multiple open phases it is unclear which
+                # status to print, so stay vague
+                detail = ["Accepting submissions"]
+        elif self.status == StatusChoices.COMPLETED:
+            detail = ["Challenge completed"]
+        elif self.status == StatusChoices.CLOSED:
+            detail = ["Not accepting submissions"]
+        else:
```
jmsmkn, line 589: This method looks a lot better! Really easy to follow now. Last thing, we should be defensive in case a future programmer extends `StatusChoices`, as this method should handle all cases. To do that, change the `else:` statement to `elif self.status == StatusChoices.OPENING_SOON`, and add a new `else:` clause that `raise NotImplementedError(f"{self.status} not handled")`. Then if the Enum is extended the future programmer knows where else in the code they need to make a change.

app/grandchallenge/challenges/filters.py
```diff
@@ -53,6 +56,6 @@ def filter_by_status(self, queryset, name, value):
         ids = [
             challenge.id
             for challenge in queryset
-            if challenge.accepting_submissions == eval(value)
+            if challenge.status.__str__() == value
```
jmsmkn, line 59: Nitpick: use `str(challenge.status)` rather than its dunder method.

app/grandchallenge/challenges/models.py
amickan: Ah right, good point. I also added this to the `submission_status_string` property on the `Phase` model.

jmsmkn: Nice!

jmsmkn (approved):

PR #2203, opened 2021-12-06
Adds support for file type outputs

amickan:


jmsmkn (commented):
I think this needs a bit more work:

- The files are not necessarily UTF-8 encoded, and we only want to keep the bytes anyway so they do not need to be decoded
- The files can be large, and so should not be loaded into memory completely
- create_instance assume that value is JSON serializable
- Both of these should have been caught by tests
- Support for these files also needs to be added to the docker backend

I would maybe split the handlers, so have one for images, one for files and one for JSON.

app/grandchallenge/components/models.py
```diff
@@ -570,7 +570,8 @@ def create_instance(self, *, image=None, value=None, fileobj=None):
         if image:
             civ.image = image
         elif fileobj:
-            civ.file = fileobj
+            container = File(fileobj)
+            civ.file.save(self.relative_path, container)
```
jmsmkn, line 574: Only keep the filename otherwise this will go into subfolders for nested interfaces:

```suggestion
            civ.file.save(Path(self.relative_path).name, container)
```

pkcakeout:
Woohoo! When will it be in production!? Thanks for the PR, I will try it out now! :)

pkcakeout:
:rocket:

jmsmkn:
It's deployed and now rolling out to the various sub systems, give it 10 mins or so and everything should be on the new version.

pkcakeout:
Thanks for the heads-up! That actually helps a lot. I will try the production interface in a bit then! :+1:

PR #2206, opened 2021-12-07
Add the algorithm version to the evaluation detail

jmsmkn:
Also be more strict about algorithm version checking to save wasting compute.


amickan (approved):
LGTM! How about we also add the version number to the algorithm detail page for completeness.

jmsmkn:
Yes sounds good to me

PR #2208, opened 2021-12-09
Challenge creation for verified users only

amickan:
- Challenge creation for verified users only
- Adds challenge create button to challenge page
- Removes `use_evaluation` from Challenge create and update forms

### DIFF
--- a/app/grandchallenge/challenges/forms.py
+++ b/app/grandchallenge/challenges/forms.py
@@ -18,6 +18,7 @@ class ChallengeCreateForm(forms.ModelForm):
         self.helper.layout.append(Submit("save", "Save"))
         self.fields["contact_email"].required = True
         self.fields["contact_email"].initial = creator.email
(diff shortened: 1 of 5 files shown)


app/grandchallenge/challenges/forms.py
```diff
@@ -28,6 +29,7 @@ class Meta:
             "use_evaluation",
             "contact_email",
         ]
+        widgets = {"use_evaluation": forms.HiddenInput()}
```
jmsmkn, line 32: Rather than using a hidden input I think that it's best to change the default on the model, create a migration, and then remove it from the form.

jmsmkn: Rest looks great!

amickan: The default is already True actually, so I can just go ahead and delete the field. :-)

jmsmkn (approved):

PR #2210, opened 2021-12-09
Automatically verify demo and algorithm dummy users

amickan:


jmsmkn:
I think that this can be done for all of the created users. Whilst we're here could we also get rid of the newsletter pop-up for these users by setting the accept/reject bool?

jmsmkn (approved):

PR #2212, opened 2021-12-10
Update challenge phase settings

amickan:
- Deprecate use_evaluation field
- Change submision limit default to 10


app/grandchallenge/evaluation/forms.py
```diff
@@ -148,6 +148,15 @@ def clean(self):
                 "the submissions open date."
             )
 
+    def clean_submission_limit(self):
+        submission_limit = self.cleaned_data["submission_limit"]
+        if submission_limit > 0 and not self.instance.method_set.last().ready:
```
jmsmkn, line 153: There is a shortcut for this already, it should be used instead:

```suggestion
        if submission_limit > 0 and not self.instance.latest_ready_method:
```

jmsmkn: Rest looks great, thanks!

amickan: The `latest_ready_method` is defined on the `Submission` model, not the `Phase` model. Should I add a similar method there?

amickan: I just went ahead and added this property to the `Phase` model ;-)

jmsmkn: Hmm ok, now we have two ways of doing this, we should only have one. Indeed this should be a property of `Phase`. Could you unify the implementations, and add explicit ordering for the queryset so that `last` always returns the correct value?

jmsmkn (approved):

PR #2215, opened 2021-12-13
Add chart, pdf and thumbnail outputs to evaluation detail

amickan:


app/grandchallenge/serving/views.py
```diff
@@ -111,5 +111,18 @@ def serve_component_interface_value(
         .exists()
     ):
         return protected_storage_redirect(name=civ.file.name)
+    elif (
```
jmsmkn, line 114: Instead of adding an extra block couldn't you set `perms=["algorithms.view_job", "evaluation.view_evaluation"]`?

amickan: No, that doesn't work. The content type of all permissions must be the same: https://django-guardian.readthedocs.io/en/stable/api/guardian.shortcuts.html#get-objects-for-user

jmsmkn: Ah, OK then!

jmsmkn (approved):

PR #2219, opened 2021-12-14
Remove flex uploads option

jmsmkn:
This setting is confusing and can be represented as a property instead. Also sets the default workstation in the dropdown and consolidates some duplication.


app/grandchallenge/algorithms/models.py
```diff
@@ -222,6 +223,11 @@ def get_absolute_url(self):
     def api_url(self):
         return reverse("api:algorithm-detail", kwargs={"pk": self.pk})
 
+    @property
+    def supports_batch_upload(self):
+        inputs = {inpt.slug for inpt in self.inputs.all()}
+        return inputs == {"generic-medical-image"}
```
amickan, line 229: I'm a bit confused. If this property replaces `use_flexible_inputs`, shouldn't we return `inputs != {"generic-medical-image"}` ?

jmsmkn: The logic has been reversed, hence the change in the property name. Batch uploads mean that many jobs result from one upload session, which can only happen if the only input to the algorithm is generic.

amickan: Ah ok, I misread how you're using this property in the template but I see now that indeed it's reversed. Thanks for clarifying!

amickan (approved):

PR #2221, opened 2021-12-15
Adds automatic access request handling to reader studies, algos and archives

### LINKED ISSUE #1764: Add automated acceptance for archives, algorithms and reader studies
Challenges have an option to automatically accept a participation request, add this for the other apps. Have the options of manual review all users, automatically accept verified users, or automatically accept all.

amickan:
- This adds the option to automatically accept access requests for archives, reader studies and algorithms
- This PR also refactors the participation request code so that all types of requests are now handled in the same way and place. 

As far as refactoring is concerned, we could go a step further and get rid of the `Participants` app altogether, moving all models, views etc. to the `Challenges` app, but I thought maybe that's a lot of refactoring so shortly before the vacation?

Closes #1764


app/grandchallenge/algorithms/models.py
```diff
@@ -702,6 +710,10 @@ def save(self, *args, **kwargs):
                 actor=self.user,
                 target=self.base_object,
             )
+        elif adding and not self.algorithm.require_user_review:
+            # immediately allow access, no need for a notification
+            self.status = self.ACCEPTED
+            self.save()
```
jmsmkn, line 716: Setting it up this way you call save recursively, it is better to avoid that and move conditionally setting `self.status` to before the first call to `super().save(*args, **kwargs)` on L702.

jmsmkn (commented):
I think this is good, indeed it is too risky to deploy just before the holidays - I would want to do more manual testing. Automated tests for this would be a good thing to add.

app/grandchallenge/algorithms/models.py
jmsmkn: This happens in a few other places so change those too.

amickan: Ah right, will fix that and I will also add tests. Thanks!

jmsmkn:
I'm wondering here if it would be better to add some levels, rather than using a boolean field have a choice:

- Manually review all requests (default)
- Automatically accept requests from verified users
- Automatically accept requests from all users

What do you think?

app/grandchallenge/challenges/management/commands/set_challenge_request_handling_preference.py
```diff
@@ -0,0 +1,19 @@
...
+from grandchallenge.core.utils.access_request_utils import (
+    AccessRequestHandlingOptions,
+)
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        challenges = Challenge.objects.all()
+        for challenge in challenges:
+            if not challenge.require_participant_review:
```
amickan, line 13: FYI subsetting immediately to `Challenge.object.filter(require_participant_review=False)` does not work with the deprecated field, so had to do it this way.

jmsmkn: I think the data migration needs to be done before the deprecation. From the docs:

> Any lingering references to it in your code will return None

So, here challenge.require_participant_review will always be `None` and no challenges will be migrated.

amickan: Oh yes, you're right!

app/grandchallenge/challenges/management/commands/set_challenge_request_handling_preference.py
```diff
@@ -0,0 +1,18 @@
...
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        challenges = Challenge.objects.filter(require_participant_review=False)
+        for challenge in challenges:
+            challenge.access_request_handling = (
+                AccessRequestHandlingOptions.ACCEPT_ALL
+            )
+
+        Challenge.objects.bulk_update(challenges, ["access_request_handling"])
```
jmsmkn, line 17: This doesn't matter as this is a throwaway command but just FYI you can do this in one swoop:

```
Challenge.objects.filter(require_participant_review=False).update(access_request_handling=AccessRequestHandlingOptions.ACCEPT_ALL)
```

jmsmkn (approved):
Great work! When you deprecate and remove the challenge field could you also move `app/grandchallenge/core/utils/access_request_utils.py` to `app/grandchallenge/core/utils/access_requests.py`?

PR #2224, opened 2021-12-21
Return 404 for leaderboard redirect if challenge has no phase

amickan:
Addresses the following sentry error: https://sentry.io/organizations/grand-challenge/issues/2874226578/?project=303639&query=is%3Aunresolved+is%3Afor_review+assigned_or_suggested%3A%5Bme%2C+none%5D&sort=inbox


MikeOverkamp-diag (changes requested):
I think I'd prefer to actually check for the existence of the Phase object. I don't really see anything else that could trigger an AttributeError, so functionally this works just as well, but checking for the object is a bit more clear and correct I'd say.

MikeOverkamp-diag (approved):

PR #2233, opened 2022-01-11
Add ALGORITHMS_JOB_BATCH_LIMIT

### LINKED ISSUE #2231: Task for scheduling many jobs for an algorithm times out
Too many jobs are scheduled for algorithm evaluation, change the strategy to do this in sensibly sized batches.

jmsmkn:
Sets limits on how many algorithm jobs can be scheduled at once.

Not super happy with this implementation but I think that it is the best we can do without refactoring everything. We know that Celery chords are not reliable so we cannot wait for jobs to be done in batches, so this still schedules everything in advance.

Closes #2231


amickan (approved):
LGTM! Just out of curiosity, what is the batch limit based on, or is it arbitrary?

jmsmkn:
It's arbitrary, just picked something that should run in a couple of minutes.

PR #2234, opened 2022-01-11
Add configurable inference time limits per phase

### LINKED ISSUE #2232: Configurable time limits for inference
Phases should be able to set the inference time limit.

jmsmkn:
This PR allows setting the time limit on each of the jobs, using a default of 2 hours. It also adds configuring the time limit for each inference job on a phase-by-phase basis, with a default of 20 minutes. This number can be changed for each phase in the django admin.

The time limit itself is only implemented for the ECS backend as with the existing docker backend this would take some time to refactor, so it's ok to have a difference in features between backends.

Closes #2232


amickan (approved):
Looks good! I remember Luuk asking for this for the STOIC challenge. We can set a limit of 20 minutes for his qualification phase once this is out.

PR #2236, opened 2022-01-12
Add email model

amickan:
- This adds an email model to grand challenge, which allows admins of the site to send emails to the mailing list, staff users, and admins of challenges, reader studies and algorithms.
- Sending an email sparks a celery task that batch sends emails to recipients, with a set greeting and email footer
- At the moment, the email object is marked as sent irrespective of whether all emails have successfully gone through.


app/grandchallenge/emails/tasks.py
```diff
@@ -0,0 +1,30 @@
...
+from django.contrib.sites.models import Site
+from django.core.mail import send_mass_mail
+from django.core.paginator import Paginator
+
+from config import settings
+from grandchallenge.subdomains.utils import reverse
+
+
+@shared_task
+def send_bulk_email(users, subject, body):
```
jmsmkn, line 11: For designing Celery tasks you should be aware that there are some restrictions. The main one is that they're idempotent, so that if they're called twice with the same arguments the work only happens once, but you know that already. Once you are sure that the task is idempotent you can use one of the acks-late queues, which makes the task more robust to failure (it is best to avoid the standard queue).

Another, is that the arguments to the task are passed through a JSON serializer, so sending Python objects (in this case, users) is really tricky, you need to stick to basic types (numbers, strings, etc) and then within your task fetch all the information that is needed. Due to the SQS broker we use, there is also a limit on how large the message containing all of the arguments can be (2MB), so it is best to avoid sending lists of things, instead, send enough information so that the list can be generated in the task instead (for instance, using an Enum choice).

Finally, for all Celery tasks, I prefer to use keyword-only arguments as it is easier to maintain (the same goes for almost all functions).

app/grandchallenge/emails/admin.py
```diff
@@ -0,0 +1,98 @@
...
+from grandchallenge.emails.models import Email
+from grandchallenge.emails.tasks import send_bulk_email
+from grandchallenge.reader_studies.models import ReaderStudy
+
+
+def schedule_emails(modeladmin, queryset, request, receivers, action_string):
+    emails = queryset.filter(sent=False)
+    if emails:
+        for email in emails:
+            send_bulk_email(receivers, email.subject, email.body)
```
jmsmkn, line 18: This isn't actually running the task through celery, you're calling the task synchronously here rather than using `apply_async`.

app/grandchallenge/emails/admin.py
```diff
@@ -0,0 +1,96 @@
...
+
+def schedule_emails(modeladmin, queryset, request, action):
+    emails = queryset.filter(sent=False)
+    if emails:
+        for email in emails:
+            send_admin_emails = send_bulk_email.signature(
+                kwargs={
+                    "action": action,
+                    "subject": email.subject,
+                    "body": email.body,
```
jmsmkn, line 20: The task is still taking the email subject and body, which could be a problem with the limited message size of SQS. It would be better if the task takes the primary key to the email, and then fetches the subject and body from the database.

app/grandchallenge/emails/admin.py
```diff
@@ -0,0 +1,96 @@
...
+                },
+                immutable=True,
+            )
+            on_commit(send_admin_emails.apply_async)
+
+        updated = emails.update(sent=True, sent_at=now())
```
jmsmkn, line 26: The guard and update should be in the Celery task, not the admin action as Celery tasks can be run twice (eg, if they are interrupted they will be placed again on another worker).

app/grandchallenge/emails/admin.py
```diff
@@ -0,0 +1,96 @@
...
+            "The emails you selected have already been sent.",
+            messages.WARNING,
+        )
+
+
+class EmailAdmin(MarkdownxModelAdmin):
+    list_display = ("subject", "sent", "sent_at")
+    actions = [*SendActionChoices]
+
+    @admin.action(description="Send to mailing list")
```
jmsmkn, line 50: These actions should only be accessible to staff members with change permissions.

app/grandchallenge/emails/tasks.py
```diff
@@ -0,0 +1,65 @@
...
+            user
+            for rs in ReaderStudy.objects.all()
+            for user in rs.editors_group.user_set.all()
+        ]
+    elif action == SendActionChoices.ALGORITHM_EDITORS:
+        receivers = [
+            user
+            for algorithm in Algorithm.objects.all()
+            for user in algorithm.editors_group.user_set.all()
+        ]
```
jmsmkn, line 45: Maybe split this out to another function and add a test that it does the right stuff?

jmsmkn: Also, this will fetch all of the users in memory at once, for this to work in production you will need to use querysets that return this data that are then fetched by the paginator.

app/grandchallenge/emails/tasks.py
```diff
@@ -0,0 +1,65 @@
...
+        for recipient in paginator.page(page_nr).object_list:
+            messages.append(
+                (
+                    f"[{site.domain.lower()}] {subject}",
+                    f"Dear {recipient.username},"
+                    f"\r\n\r\n{body}\r\n\r\n"
+                    f"Kind regards, \r\nGrand Challenge team \r\n\r\n"
+                    f"To unsubscribe from this mailing list, "
+                    f"uncheck 'Receive newsletter' in your profile settings:"
+                    f" {reverse('profile-update', kwargs={'username': recipient.username})}",
```
jmsmkn, line 60: Please add a test that the unsubscribe link is rendered correctly, important that we have that.

jmsmkn (commented):
Looks good! A few minor changes and some tests are needed then this will be good to go.

app/grandchallenge/emails/tasks.py
jmsmkn: It's only really about timing, if you think that it's going to take more than 300s then use the large queue.

PR #2238, opened 2022-01-13
Count displayed subscriptions only

### LINKED ISSUE #2237: Subscription count does not correspond with displayed subscriptions
**Problem**
We do not show all subscriptions to a user - only those that we allow them to unsubscribe from. Subscriptions that are not shown to the user include permission request follows and upload-session follows, as well as "inactive" algorithm job follows (a necessary work around to allow unsubscribing from failed job notifications). The paginator on the listview, however, counts all subscriptions of the user, which leads to empty pages and a discrepancy in the shown vs. indicated number of subscriptions. 

**Solution**
We need to make sure that the paginator counts only the subscriptions that are actually shown to the user. 
Simply showing all subscriptions sounds like a good alternative, but is not possible, because the inactive algorithm job follows should not be exposed to the user.

amickan:
Closes #2237

### DIFF
--- a/app/grandchallenge/notifications/views.py
+++ b/app/grandchallenge/notifications/views.py
@@ -1,6 +1,8 @@
 from actstream.models import Follow
 from django.contrib import messages
+from django.contrib.contenttypes.models import ContentType
 from django.contrib.messages.views import SuccessMessageMixin
+from django.core.paginator import EmptyPage, Paginator
 from django.views.generic import CreateView, DeleteView, ListView
 from guardian.mixins import (
     LoginRequiredMixin,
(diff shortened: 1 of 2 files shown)


jmsmkn:
I would have thought that filtering the queryset would be all that is needed. Could you explain the solution here? It seems like a lot of duplicated code,

amickan:
It was a method resolution order problem related to the FilterMixin again. I couldn't figure out how to resolve it by changing the order the classes and mixins are inherited and so went with the complicated solution, but I now remembered that I can define the queryset on the View, which avoids the call to super() in the get_queryset() method,

app/grandchallenge/notifications/views.py
```diff
@@ -119,14 +120,29 @@ class FollowList(
...
+        ).get(),
+        ContentType.objects.filter(
+            app_label="reader_studies", model="readerstudypermissionrequest",
+        ).get(),
+        ContentType.objects.filter(
+            app_label="participants", model="registrationrequest"
+        ).get(),
+        ContentType.objects.filter(
+            app_label="cases", model="rawimageuploadsession"
+        ).get(),
```
jmsmkn, line 138: This does several DB calls, you should be able to get the same information out using `Q` queries, something like 

```
ContentType.objects.filter(
  Q(app_label="archives", model="archivepermissionrequest")
  | Q(app_label="algorithms", model="algorithmpermissionrequest")
  | ...
)
```

I think you could even use that directly in the exclude statement on L141 by doing something like 

```
Follow.objects.filter(
  Q(content_type__app_label="archives", content_type__model="archivepermissionrequest")
  | Q(content_type__app_label="algorithms", content_type__model="algorithmpermissionrequest")
  | ...
)
```

Don't know if that would be efficient though, would need to profile it.

amickan:
> Hmm, doesn't this then skip the call to the `PermissionListMixin` where the queryset is filtered by the users objects? 

No, the call to PermissionListMixin is not skipped. The FilterMixin calls the get_queryset() method from the PermissionListMixin, which in turn calls the get_queryset method from the ListView, which takes the queryset we defined on the view.

> I think that we need to find a fix in the mixin rather than working around the problem here.

But I do agree that it would be best to restructure the FilterMixin to prevent this issue in the future. I will think about it.

jmsmkn:
> No, the call to PermissionListMixin is not skipped. The FilterMixin calls the get_queryset() method from the PermissionListMixin, which in turn calls the get_queryset method from the ListView, which takes the queryset we defined on the view.

Yes you're right, I was reading your changes wrong sorry! Thought that you were doing this in `get_queryset` still.

There are still issues as you need to do a db query to ContentType in order to instantiate the class, if you can find a way around that this approach will work.

jmsmkn (approved):

PR #2242, opened 2022-01-14
Basic implementation of a Cornerstone viewer

HarmvZ:
Implements a basic Cornerstone.js based viewer. The viewer supports only 2D greyscale, rgb and rgba images (no YCbCr, no 3D or 4D). It allows basic image manipulation like panning, zooming, adjusting window width/center, invert, flip horizontal/vertical and rotate.

Currently, the image loader endpoint returns a [Cornerstone image object](https://docs.cornerstonejs.org/api.html#image). This requires conversion of the MetaIO image to (numpy) array on the backend. We have to see if the performance of this endpoint holds up in production, it will probably perform similar or even worse than the retina thumbnail endpoints. To reduce the strain on our server we could move this conversion to the frontend by using something like itk-wasm to load the file directly from s3. I'll try that out and see if I can get it to work next week.

**Demo**
Not sure why my cursor is not showing up. I'm doing the panning/zooming/window leveling in the beginning. The tooltips for the buttons show up on hover.

https://user-images.githubusercontent.com/2001094/149543050-fab8e01c-d237-4e82-a129-b473d9a7992e.mp4


app/grandchallenge/cases/urls.py
```diff
@@ -30,4 +32,14 @@
         VTKImageDetail.as_view(),
         name="vtk-image-detail",
     ),
+    path(
+        "images/<uuid:pk>/cs/",
+        CSImageDetail.as_view(),
+        name="cs-image-detail",
+    ),
+    path(
+        "images/<uuid:pk>/cs/loader",
```
jmsmkn, line 41:
```suggestion
        "images/<uuid:pk>/cs/loader/",
```

Missing trailing slash

app/grandchallenge/cases/views.py
```diff
@@ -139,3 +142,39 @@ def get_context_data(self, **kwargs):
...
+            # >2D volumes and YCrBr not supported in cornerstone
+            raise Http404
+
+        context.update({"image_pk": self.object.pk})
+        return context
+
+
+class CSImageLoader(RetrieveAPIView):
+    queryset = Image.objects.all()
+    permission_classes = (DjangoObjectPermissions,)
```
jmsmkn, line 177: I think that it is best that we limit access to this endpoint given how heavyweight it is. Is it possible to add a global permission to use this feature alongside the object-based permissions?

app/grandchallenge/cases/serializers.py
```diff
@@ -208,3 +213,74 @@ def _get_linked_task(*, targets):
         )
     else:
         raise RuntimeError(f"Unknown target {targets=}")
+
+
+class CSImageSerializer(serializers.BaseSerializer):
+    """
+    Sserializes a cases.Image object into a Cornerstone image object as
+    defined in https://docs.cornerstonejs.org/api.html#image.
+    """
```
jmsmkn, line 222: If this goes to production we shouldn't serve this in this way, the images should be pre-processed by some background task and stored for serving directly from the CDN. Could you add a bit TODO explaining this and that this function is behind a feature flag?

app/config/settings.py
```diff
@@ -1160,9 +1161,38 @@ def strtobool(val) -> bool:
     "CASES_POST_PROCESSORS", "panimg.post_processors.tiff_to_dzi"
 ).split(",")
 
-# Maximum file size in bytes to be opened by SimpleITK.ReadImage in cases.models.Image.get_sitk_image()
+# Maximum file size in bytes to be opened by SimpleITK.ReadImage in cases.utils.get_sitk_image()
 MAX_SITK_FILE_SIZE = 268_435_456  # 256 mb
 
+SITK_PIXEL_TYPE_TO_BIT_DEPTH = {
```
jmsmkn, line 1167: Isn't this already defined in panimg? I'd rather you take this out of settings and just put it in the one place where it is used.

app/grandchallenge/cases/serializers.py
```diff
@@ -208,3 +213,74 @@ def _get_linked_task(*, targets):
...
+
+    def to_representation(self, instance):
+        try:
+            image_itk = get_sitk_image(image=instance)
+        except Exception as e:
+            raise Http404 from e
+
+        if instance.depth and instance.depth > 1:
```
jmsmkn, line 230: You can check for this before fetching the pixel data, right?

app/tests/cases_tests/test_models.py
```diff
@@ -7,14 +7,14 @@
 from django.core.exceptions import MultipleObjectsReturned
 from django.core.files import File
 
+from grandchallenge.cases.utils import get_sitk_image
```
jmsmkn, line 10: This method shouldn't be used in production, could you add a comment in there to say that this is temporary to evaluate this functionality?

app/grandchallenge/cases/serializers.py
jmsmkn: The more I think about it I really think this should be moved client-side, could you do this in JS instead?

HarmvZ:
Thanks for the review! The comments are addressed in #2247 which is based on this branch.

app/grandchallenge/cases/views.py
```diff
@@ -139,3 +139,33 @@ def get_context_data(self, **kwargs):
...
+
+        if self.object.color_space not in (
+            Image.COLOR_SPACE_GRAY,
+            Image.COLOR_SPACE_RGBA,
+            Image.COLOR_SPACE_RGB,
+        ):
+            # YCrBr not supported in cornerstone
+            raise Http404
+
+        context.update({"image_pk": self.object.pk})
```
jmsmkn, line 170: I don't think that you need to set this, you can use use `object.pk` in the template directly.

jmsmkn:
Much better approach, just a couple of small changes.

jmsmkn (approved):

PR #2243, opened 2022-01-17
Sort options - attempt to fix bug

Kicksta:
There is no error locally, so difficult to debug. Could you check if this fixed it?

### DIFF
--- a/app/grandchallenge/products/views.py
+++ b/app/grandchallenge/products/views.py
@@ -69,15 +69,16 @@ class ProductList(ListView):
 
     fda_classes = ["All", "Class I", "Class II", "Class III", "No FDA"]
 
+    sort_bys = {
+        "CE certification": "-ce_under",
+        "Last modified": "-modified",
+        "A-Z product": "product_name",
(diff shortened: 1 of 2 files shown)


app/grandchallenge/products/views.py
```diff
@@ -69,6 +69,17 @@ class ProductList(ListView):
...
+    sort_bys = {
+        "CE certification": "-ce_under",
+        "Last modified": "-modified",
+        "A-Z product": "product_name",
+        "A-Z company": "company__company_name",
+    }
+
+    def get_sort_by(self):
+        sort_query = self.request.GET.get("sort_by")
+        return self.sort_bys[sort_query]
```
jmsmkn, line 81: The best thing to do here is to catch and handle the `KeyError`.

jmsmkn:
We were getting too many alerts so I have tried to fix this direclt https://github.com/comic/grand-challenge.org/commit/c6f121c9fc9b75db9de7710c08c128b20daf3abc

PR #2244, opened 2022-01-17
Don't raise a 404 if user does not have a GitHub token

MikeOverkamp-diag:
Display an empty repo list instead.


amickan (approved):
LGTM! Thanks for fixing this so fast!

PR #2249, opened 2022-01-19
Update UploadSession serializer and add ArchiveItem serializer

amickan:
This reflects the changes necessary on the GC side to enable users to define an interface when uploading cases through the API.

For the tests, to check whether the correct archive items were created, I added an ArchiveItemSerializer.

Addresses first part of the [pitch](https://github.com/DIAGNijmegen/rse-roadmap/issues/114). 
 
Linked gc-api PR: https://github.com/DIAGNijmegen/rse-gcapi/pull/100


app/grandchallenge/archives/serializers.py
```diff
@@ -13,6 +31,7 @@ class ArchiveSerializer(serializers.ModelSerializer):
     url = URLField(source="get_absolute_url", read_only=True)
     # Include the read only name for legacy clients
     name = ReadOnlyField()
+    items = ArchiveItemSerializer(many=True, read_only=True)
```
jmsmkn, line 34: It's better not to add items at all to this serializer, if the user needs this then they should be able to get the information from the archive item list endpoint.

app/grandchallenge/cases/serializers.py
```diff
@@ -166,7 +173,7 @@ def validate(self, attrs):
             raise ValidationError("Filenames must be unique")
 
         num_targets = sum(f in attrs for f in self.targets)
-        if num_targets > 1:
+        if num_targets > 2 and "interface" not in self.targets:
```
jmsmkn, line 176: The logic here needs to be reworked - interface can only be set if archive it set, right?

app/grandchallenge/archives/serializers.py
jmsmkn: You will need to add an archive item list endpoint with permissions, filters and tests to do this!

app/grandchallenge/archives/permissions.py
```diff
@@ -0,0 +1,27 @@
+from django.http import Http404
+from rest_framework import exceptions, permissions
+
+
+class ArchiveItemPermission(permissions.BasePermission):
+    """
+    Checks whether current user has permission to view or update the archive
+    corresponding to the current archive item.
```
jmsmkn, line 8: Do you need a separate class for this? Doesn't DRF have a concept of `get_permission_object` like we use in the Django views? We may need instead to add the permissions directly to the archive items.

app/grandchallenge/archives/views.py
```diff
@@ -732,3 +737,20 @@ def studies(self, request, pk=None):
...
+
+
+class ArchiveItemViewSet(ReadOnlyModelViewSet):
+    serializer_class = ArchiveItemSerializer
+    permission_classes = [ArchiveItemPermission]
+    filter_backends = [DjangoFilterBackend]
+    filterset_fields = ["archive"]
+
+    def get_queryset(self):
+        # restrict queryset to items from archives the user has view permission for
```
jmsmkn, line 749: See the comment above, it seems the permissions are being done in two places that need maintaining.

app/grandchallenge/cases/serializers.py
```diff
@@ -171,6 +184,13 @@ def validate(self, attrs):
                 "Only one of archive, answer or reader study can be set"
             )
 
+        if "interface" in attrs and (
+            "reader_study" in attrs or "answer" in attrs
+        ):
```
jmsmkn, line 189: This would be more robust to changes:

```suggestion
        if "interface" in attrs and "archive" not in attrs:
```

app/grandchallenge/cases/serializers.py
```diff
@@ -144,6 +151,10 @@ def create(self, validated_data):
             for t in self.targets
             if t in validated_data
         }
+        try:
+            interface = validated_data.pop("interface")
+        except KeyError:
+            interface = None
```
jmsmkn, line 157:
```suggestion
        interface = validated_data.pop("interface", None)
```

app/grandchallenge/archives/permissions.py
amickan: You can overwrite the `get_object()` and pass the corresponing `archive` to the permission tests, but these will still fail, because we're not using the standard permissions (view and update) and because it would check for archive_item level permissions, not archive level permisisons (even if we override the get_object() method). Without rewriting the permission logic for archive items, this seemed like the cleanest solution. Would adding permissions directly to archive items be better?

jmsmkn: I think that it's best to work within the framework. How do we handle this for other nested items (eg, algorithm images)? I think we set permissions directly, right?

app/grandchallenge/archives/views.py
amickan: Yes, there are two checks, one for the object itself (to retrieve a single archive item, normally done with `DjangoObjectPermissions`) and one for retrieving a list of archive items (which is normally done through a filter `ObjectPermissionsFilter` which filters the list to those objects the user has view permission for). Again, since we need to check them for the archive and not the item, that filter does not work for us. All the filter does is filter the queryset, so I thought it would be easiest to just override the get_queryset method, but I can also create a custom filter class.

app/grandchallenge/archives/views.py
```diff
@@ -368,12 +371,12 @@ class ArchiveEditArchiveItem(
...
     template_name = "archives/archive_item_form.html"
     permission_required = (
-        f"{Archive._meta.app_label}.upload_{Archive._meta.model_name}"
+        f"{ArchiveItem._meta.app_label}.change_{ArchiveItem._meta.model_name}"
     )
     raise_exception = True
 
     def get_permission_object(self):
-        return self.archive
+        return get_object_or_404(ArchiveItem, pk=self.kwargs["pk"])
```
jmsmkn, line 379: This method can be removed now, right?

app/tests/cases_tests/test_api.py
```diff
@@ -243,20 +274,45 @@ def test_session_with_user_upload(client, obj, factory):
...
+            viewname="api:upload-session-list",
+            user=user,
+            client=client,
+            method=client.post,
+            content_type="application/json",
+            data=data,
+            HTTP_X_FORWARDED_PROTO="https",
+        )
+
+    if obj == "reader_study" and interface:
```
jmsmkn, line 297: Just a general comment (and this may be hypocritical as I know we do this in the tests already, but, I want to try and improve the tests as we go). I think that it is best to not have any logic in the tests, if you need it, write a new test (use the assemble, act, assert pattern). For test code I think that it is better to duplicate as if tests get too complex they are hard to maintain, and it can be difficult to understand the coders intent.

jmsmkn (commented):
Looks good!

app/grandchallenge/archives/views.py
amickan: For this view, we still need it. It's a FormView and it does not have a get_object() method defined, so either we define `object` on the view, or we add a `get_object()` method to the view, or (in this is how I've seen it done in other places in the code base) we just overwrite `get_permission_object()`. I think the latter is the easiest to understand, right?

jmsmkn: OK I see, then it's best to use `self.archive_item` as the handling is already done there.

app/grandchallenge/core/management/commands/archive_item_permissions.py
```diff
@@ -0,0 +1,18 @@
...
+from django.core.paginator import Paginator
+
+from grandchallenge.archives.models import ArchiveItem
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        items = (
+            ArchiveItem.objects.select_related("archive")
+            .order_by("-created")
```
jmsmkn, line 11: Needs to be oldest first otherwise we get problems with insertions when running this:

```suggestion
            .order_by("created")
```

app/grandchallenge/archives/views.py
amickan: Ah right, that makes sense! Thanks!

jmsmkn (approved):

PR #2254, opened 2022-01-25
Correct DateTimeInput format

amickan:
To use an input element of type `datetime-local`, the validation format needs to be changed to `YYYY-MM-DDThh:mm`.

### DIFF
--- a/app/grandchallenge/core/models.py
+++ b/app/grandchallenge/core/models.py
@@ -2,7 +2,7 @@ import uuid
 
 from django.conf import settings
 from django.db import models
-from django.utils.timezone import get_current_timezone
+from django.utils.timezone import localtime
 from django.utils.translation import gettext_lazy as _
 from django_extensions.db.models import (
     TitleSlugDescriptionModel as BaseTitleSlugDescriptionModel,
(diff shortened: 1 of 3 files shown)


jmsmkn:
Do we need to take into account the users timezone?

amickan:
> Do we need to take into account the users timezone?

I'm not entirely sure what would be best to be honest. As it is now, the date and time the user enters is assumed to be in UTC (despite the fact that the form sends back an aware datatime object with the user's timezone info). So we could just add a helptext explaining that. Alternatively, we could explicitly convert the datetime to utc before storing it in the database. But I think that might cause more confusion. 

Also, the start date is printed in UTC on the challenge card. We could alternatively print it according to the user's locale. If we leave it at UTC, we should add that to the string to avoid confusion.

jmsmkn:
OK - as the form is returning objects that are timezone aware then we can store that straight in the database. We should assume that the user is using their localtime and stick to that.

When presented all times should use the users localtime (it's done automatically by the template rendering engine), can we do the same for those strings?

jmsmkn (approved):

PR #2255, opened 2022-01-25
Add reader-study setting for answer roll-over

chrisvanrun:
This PR adds a setting to control the reader-study answer roll-over for CIRRUS as was implemented in:
- https://github.com/DIAGNijmegen/rse-cirrus-core/pull/1733


app/grandchallenge/reader_studies/models.py
```diff
@@ -290,6 +290,17 @@ class ReaderStudy(UUIDModel, TitleSlugDescriptionModel):
             "for a case."
         ),
     )
+    roll_over_answers_for_n_cases = models.PositiveSmallIntegerField(
+        default=False,
+        help_text=(
+            "The number of cases for which answers should roll over. "
+            "For instance, if set to 1. Case 2 will start with the answers from case 1; "
+            "whereas case 3 starts anew but its answers will rollover to case 4."
```
jmsmkn, line 298:
```suggestion
            "whereas case 3 starts anew but its answers will rollover to case 4. "
```

app/grandchallenge/reader_studies/migrations/0014_readerstudy_roll_over_answers_for_n_cases.py
```diff
@@ -0,0 +1,21 @@
...
+        ("reader_studies", "0013_auto_20211206_0925"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="readerstudy",
+            name="roll_over_answers_for_n_cases",
+            field=models.PositiveSmallIntegerField(
+                default=False,
+                help_text="The number of cases for which answers should roll over. For instance, if set to 1. Case 2 will start with the answers from case 1; whereas case 3 starts anew but its answers will rollover to case 4.It can be used for repeated readings with slightly different hangings, but is counterproductive when used in combination with case navigation or shuffling hangings.",
```
jmsmkn, line 18:
```suggestion
                help_text="The number of cases for which answers should roll over. For instance, if set to 1. Case 2 will start with the answers from case 1; whereas case 3 starts anew but its answers will rollover to case 4. It can be used for repeated readings with slightly different hangings, but is counterproductive when used in combination with case navigation or shuffling hangings.",
```

app/grandchallenge/reader_studies/models.py
```diff
@@ -290,6 +290,17 @@ class ReaderStudy(UUIDModel, TitleSlugDescriptionModel):
             "for a case."
         ),
     )
+    roll_over_answers_for_n_cases = models.PositiveSmallIntegerField(
+        default=False,
+        help_text=(
+            "The number of cases for which answers should roll over. "
+            "For instance, if set to 1. Case 2 will start with the answers from case 1; "
```
miriam-groeneveld, line 297:
```suggestion
            "For instance, if set to 1, case 2 will start with the answers from case 1; "
```

app/grandchallenge/reader_studies/models.py
miriam-groeneveld, line 298:
```suggestion
            "whereas case 3 starts anew but its answers will rollover to case 4. "
```

app/grandchallenge/reader_studies/models.py
```diff
@@ -290,6 +290,17 @@ class ReaderStudy(UUIDModel, TitleSlugDescriptionModel):
...
+            "whereas case 3 starts anew but its answers will rollover to case 4."
+            "It can be used for repeated readings with slightly different hangings, "
+            "but is counterproductive when used in combination with case navigation or "
```
miriam-groeneveld, line 300:
```suggestion
            "but should not be used in combination with case navigation or "
```

miriam-groeneveld (dismissed):
Some textual suggestions only. If the setting should not be used in certain circumstances (shuffle, case navigation), maybe add a check to prevent that? If you accept the suggestions you will need to update the migration file as well.

miriam-groeneveld:
Double review :)

chrisvanrun:
> Some textual suggestions only. If the setting should not be used in certain circumstances (shuffle, case navigation), maybe add a check to prevent that? If you accept the suggestions you will need to update the migration file as well.

Bit in the middle about that. I am edging toward a warning rather than a full blown error. Since there are some corner cases where you'd still want readers to re-view / edit earlier answers. I'll see if I can add a warning, maybe then the help text can also be shortened.

chrisvanrun:
> > Some textual suggestions only. If the setting should not be used in certain circumstances (shuffle, case navigation), maybe add a check to prevent that? If you accept the suggestions you will need to update the migration file as well.
> 
> Bit in the middle about that. I am edging toward a warning rather than a full blown error. Since there are some corner cases where you'd still want readers to re-view / edit earlier answers. I'll see if I can add a warning, maybe then the help text can also be shortened.

Ended up implementing a straight forward error and included these in the tests. Ready for review/merging.

jmsmkn (approved):
LGTM. FYI - a better way to test forms is to do it directly rather than instantiating a whole client and request (and, rich from me considering I almost certainly added that structure but I want to try and improve our tests as they're getting far too long to run). I can see why you added the test code there but that test is doing far too much. For this PR it's fine but this is something I'd like us to work on.

PR #2256, opened 2022-01-25
Improve GitHub clone error handling

jmsmkn:
Here we preserve stdout and stderr and create user facing errors
from them. There is no way to get these messages to the user
right now, but helps with administration.


app/grandchallenge/github/models.py
```diff
@@ -109,15 +110,33 @@ def tag(self):
     def tag_url(self):
         return f"https://github.com/{self.payload['repository']['full_name']}/releases/tag/{self.payload['ref']}"
 
+    @property
+    def user_error(self):
+        if "This repository is over its data quota" in self.stdout:
```
jmsmkn, line 115: This is the only problem that has occurred so far.

app/grandchallenge/github/models.py
```diff
@@ -109,15 +110,33 @@ def tag(self):
...
+            if self.payload.get("ref_type") == "tag":
+                on_commit(
+                    lambda: get_zipfile.apply_async(kwargs={"pk": self.pk})
+                )
+            elif self.payload.get("action") == "deleted":
+                on_commit(
+                    lambda: unlink_algorithm.apply_async(
+                        kwargs={"pk": self.pk}
+                    )
+                )
```
jmsmkn, line 135: I checked the database and there is no message with the combination of these options in the payload, so it is safe to make this if statement an elif IMO.

amickan (approved):
LGTM, but why can we not present the error to the user in the containers tab on the algorithm detail page?

jmsmkn:
As there is no link between algorithms and webhook messages. Those links are only created by the `Build` objects, but they are only created on success of `get_zipfile` when `create_codebuild_build` is called. Of course, it's possible, but didn't want to redesign this right now. Probably it's best to send a notification but that can be separate.

PR #2259, opened 2022-01-31
Add archive item to UploadSession serializer

amickan:
- This adds the option to target an archive item (next to an archive or a reader study) while uploading cases through the api.
- Doing this requires that an interface is specified.

### DIFF
--- a/app/grandchallenge/archives/tasks.py
+++ b/app/grandchallenge/archives/tasks.py
@@ -4,7 +4,7 @@ from django.db import transaction
 from django.db.transaction import on_commit
 
 from grandchallenge.archives.models import Archive, ArchiveItem
-from grandchallenge.cases.models import Image
+from grandchallenge.cases.models import Image, RawImageUploadSession
 from grandchallenge.components.models import (
     ComponentInterface,
     ComponentInterfaceValue,
@@ -48,6 +48,12 @@ def add_images_to_archive_item(
 ):
     archive_item = ArchiveItem.objects.get(pk=archive_item_pk)
     interface = ComponentInterface.objects.get(pk=interface_pk)
+    session = RawImageUploadSession.objects.get(pk=upload_session_pk)
+
+    if archive_item.values.filter(
+        interface=interface, image__in=session.image_set.all()
+    ).exists():
+        return
 
     with transaction.atomic():
         archive_item.values.remove(
--- a/app/grandchallenge/archives/tasks.py
+++ b/app/grandchallenge/archives/tasks.py
@@ -1,5 +1,8 @@
 from celery import shared_task
 
+from grandchallenge.algorithms.tasks import (
+    add_images_to_component_interface_value,
+)
 from grandchallenge.archives.models import Archive, ArchiveItem
 from grandchallenge.cases.models import Image
 from grandchallenge.components.models import (
@@ -36,27 +39,26 @@ def add_images_to_archive(*, upload_session_pk, archive_pk, interface_pk=None):
 
 
 @shared_task
-def add_image_to_archive_item(
+def add_images_to_archive_item(
     *, upload_session_pk, archive_item_pk, interface_pk=None
 ):
-    image = Image.objects.filter(origin_id=upload_session_pk).get()
     archive_item = ArchiveItem.objects.get(pk=archive_item_pk)
     interface = ComponentInterface.objects.get(pk=interface_pk)
 
     civ_pks_to_remove = set()
     civ_pks_to_add = set()
 
-    try:
-        old_civ = archive_item.values.filter(interface=interface).get()
-        civ_pks_to_remove.add(old_civ.pk)
-    except ComponentInterfaceValue.DoesNotExist:
-        pass
+    civ = archive_item.values.filter(interface=interface).first()
+    if civ:
+        civ_pks_to_remove.add(civ.pk)
 
-    civ = ComponentInterfaceValue.objects.create(
-        interface=interface, image=image
-    )
+    civ = ComponentInterfaceValue.objects.create(interface=interface)
     civ_pks_to_add.add(civ)
 
+    add_images_to_component_interface_value(
+        component_interface_value_pk=civ.pk,
+        upload_session_pk=upload_session_pk,
+    )
     update_archive_item_values(
         archive_item_pk=archive_item_pk,
         civ_pks_to_remove=civ_pks_to_remove,
--- a/app/grandchallenge/archives/tasks.py
+++ b/app/grandchallenge/archives/tasks.py
@@ -49,19 +49,12 @@ def add_images_to_archive_item(
     archive_item = ArchiveItem.objects.get(pk=archive_item_pk)
     interface = ComponentInterface.objects.get(pk=interface_pk)
 
-    civ_pks_to_remove = set()
-    civ_pks_to_add = set()
-
-    civ = archive_item.values.filter(interface=interface).first()
-    if civ:
-        civ_pks_to_remove.add(civ.pk)
-
     with transaction.atomic():
+        archive_item.values.remove(
+            *archive_item.values.filter(interface=interface)
+        )
         new_civ = ComponentInterfaceValue.objects.create(interface=interface)
-        civ_pks_to_add.add(new_civ)
-
-        archive_item.values.remove(*civ_pks_to_remove)
-        archive_item.values.add(*civ_pks_to_add)
+        archive_item.values.add(new_civ)
 
         on_commit(
             add_images_to_component_interface_value.signature(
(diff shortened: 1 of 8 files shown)


app/grandchallenge/archives/tasks.py
```diff
@@ -35,6 +38,34 @@ def add_images_to_archive(*, upload_session_pk, archive_pk, interface_pk=None):
         item.values.set([civ])
 
 
+@shared_task
```
jmsmkn, line 41: I would prefer if this was on a different queue as using the main one too much is causing problems like we saw over the holidays. Probably the acks-late is best, I think all that would be needed is to wrap this in an atomic transaction as there are no other side effects.

app/grandchallenge/archives/tasks.py
```diff
@@ -35,6 +38,34 @@ def add_images_to_archive(*, upload_session_pk, archive_pk, interface_pk=None):
...
+def add_images_to_archive_item(
+    *, upload_session_pk, archive_item_pk, interface_pk=None
+):
+    archive_item = ArchiveItem.objects.get(pk=archive_item_pk)
+    interface = ComponentInterface.objects.get(pk=interface_pk)
+
+    civ_pks_to_remove = set()
+    civ_pks_to_add = set()
+
+    civ = archive_item.values.filter(interface=interface).first()
```
jmsmkn, line 51: Only get the first? Archive items can only contain one of each interface type, so it should be safe to remove them all.

app/grandchallenge/archives/tasks.py
```diff
@@ -35,6 +38,34 @@ def add_images_to_archive(*, upload_session_pk, archive_pk, interface_pk=None):
...
+    if civ:
+        civ_pks_to_remove.add(civ.pk)
+
+    civ = ComponentInterfaceValue.objects.create(interface=interface)
```
jmsmkn, line 55: Maybe call this `new_civ` to differentiate it from the one(s) being removed above?

app/grandchallenge/archives/tasks.py
```diff
@@ -1,5 +1,8 @@
 from celery import shared_task
 
+from grandchallenge.algorithms.tasks import (
+    add_images_to_component_interface_value,
```
jmsmkn, line 4: This task should probably move out of `algorithms` and into `components` now.

app/grandchallenge/archives/tasks.py
```diff
@@ -35,6 +38,34 @@ def add_images_to_archive(*, upload_session_pk, archive_pk, interface_pk=None):
         item.values.set([civ])
 
 
+@shared_task
+def add_images_to_archive_item(
+    *, upload_session_pk, archive_item_pk, interface_pk=None
```
jmsmkn, line 43: `interface_pk` needs to be set, right? Otherwise, L46 will error out.

app/grandchallenge/archives/tasks.py
```diff
@@ -35,6 +38,34 @@ def add_images_to_archive(*, upload_session_pk, archive_pk, interface_pk=None):
...
+
+    add_images_to_component_interface_value(
+        component_interface_value_pk=civ.pk,
+        upload_session_pk=upload_session_pk,
+    )
+    update_archive_item_values(
+        archive_item_pk=archive_item_pk,
+        civ_pks_to_remove=list(civ_pks_to_remove),
+        civ_pks_to_add=list(civ_pks_to_add),
+    )
```
jmsmkn, line 66: It is better not to use the celery tasks directly, but instead to call them with `on_commit(<task>.apply_async(kwags={...}))`.

app/grandchallenge/cases/serializers.py
```diff
@@ -178,14 +189,24 @@ def validate(self, attrs):
...
+            "archive" in attrs or "archive_item" in attrs
+        ):
             raise ValidationError(
-                "An interface can only be defined for archive uploads."
+                "An interface can only be defined for archive "
+                "or archive item uploads."
             )
 
+        if "archive_item" in attrs:
+            if "interface" not in attrs:
```
jmsmkn, line 204: This can be simplified to:

```suggestion
        if "archive_item" in attrs and "interface" not in attrs:
```

app/grandchallenge/cases/serializers.py
```diff
@@ -219,6 +240,15 @@ def _get_linked_task(*, targets, interface):
         if interface:
             kwargs["interface_pk"] = interface.pk
         return add_images_to_archive.signature(kwargs=kwargs, immutable=True,)
+    elif "archive_item" in targets:
+        kwargs = {
+            "archive_item_pk": targets["archive_item"].pk,
+        }
+        if interface:
+            kwargs["interface_pk"] = interface.pk
```
jmsmkn, line 248: If the `interface_pk` is not set I think that `add_images_to_archive_item` will fail as-is (maybe I'm missing something). Is doing this covered by a test?

app/tests/cases_tests/test_api.py
```diff
@@ -391,3 +395,101 @@ def test_session_with_user_duplicate_upload(client):
...
+            HTTP_X_FORWARDED_PROTO="https",
+        )
+    assert response.status_code == 201
+    item.refresh_from_db()
+    # check that there is only one civ with the generic-overlay interface
+    assert item.values.filter(interface__slug="generic-overlay").count() == 1
+    # and that the previously added one is not longer associated with the item
+    assert generic_overlay_civ_1 not in item.values.all()
+
+    # try upload without interface
```
jmsmkn, line 475: Would be nicer to split this up into 3 different tests.

app/grandchallenge/archives/tasks.py
jmsmkn: Whilst we're here we can do the same with `update_archive_item_values` to improve the existing codebase :-)

app/grandchallenge/cases/serializers.py
jmsmkn: Hmm, it does seem like it is covered in `try upload without interface`, but, I'm confused about why `ComponentInterface.objects.get(pk=interface_pk)` when `interface_pk = None` works.

jmsmkn (commented):
This looks good! I have a few questions as I don't understand why one thing works, and a few suggestions about improving the Celery task.

app/grandchallenge/cases/serializers.py
amickan: Correct, the interface does need to be set. If it is not set, a validation error is thrown during validation of the serlizalier, so `add_images_to_archive_item` never gets executed unless an interface is defined. That's why the test works. But I have adjusted the function definition accordingly now to reflect that the interface needs to be set, my bad for overlooking this!

app/grandchallenge/archives/tasks.py
amickan: Hm, maybe I'm missing something, but here is my reasoning for using `first()`: there can either be exactly 1 matching object, or None. So retieving `all()` is no differet than gettiing the first, right? The only diffference is that if I call `all()` I get a queryset back, which I would then need to loop over in order to add the civ.pk to `civ_pks_to_remove`. With `first()` I get the object directly, so it seemed easier this way. In any case, `first()` seemed better than `get()` because the latter would throw an error if there is no matching interface, `first()` in turn returns None in that case.

app/grandchallenge/archives/tasks.py
amickan: Ok, understood, I went with a hybrid solution, calling `add_images_to_component_interface_value` asynchronously, but doing the achive item updating synchronously in the function. The archive item updating cannot be done asynchronously, because then calling the task a second time would (or might) result in two civs being created.

jmsmkn: Not sure that I quite follow - you're now using atomic transactions so there will be no changes until the entire task completes. If the method is called again then the existing interface will be removed, right?

app/grandchallenge/archives/tasks.py
jmsmkn: There are no guarantees that it is one or `None`, and it is best to use this point to be really sure that there are no others to save errors from propagating. You should be able to do `archive_item.values.remove(*archive_item.values.filter(interface=interface))`, and `archive_item.values.add(new_civ)`.

app/grandchallenge/cases/serializers.py
jmsmkn: This is what code review is for, don't worry!

app/grandchallenge/archives/tasks.py
jmsmkn: Or maybe we need to check at the start of `add_images_to_archive_item` that the image from the upload session does not already belong to that archive item with that interface?

amickan: The way it is now it works. The second time the task is called the existing interface will be removed. When I tried to call `update_archive_item_values` aysnchronously, there was no new interface yet the second time the task ran, and so there was no interface to remove. Somehow the updating was running in parallel to the second run of the task...

jmsmkn (approved):

jmsmkn:
Great work!

PR #2262, opened 2022-02-01
Make license detection dynamic

jmsmkn:
This changes things around so that open source determination
is done at run time, rather than at import time. This allows
us to modify the list of licenses considered open source.

Licensee is upgraded to v9.15.1 so that json output is available.
This standardises the results and opens up more options such
as dynamically determining fine-grained restrictions such as
commercial use.

Also adds support for BSD licenses.


app/grandchallenge/github/models.py
```diff
@@ -121,14 +120,29 @@ def user_error(self):
         else:
             return ""
 
+    @property
+    def license(self):
+        # Return the first license found in the results
+        return self.license_check_result.get("licenses", [{}])[0]
```
jmsmkn, line 126: I don't know if this is super robust, we need to run it on a bunch of repos and find out.

jmsmkn: Or, maybe check _all_ of the licenses found.

app/grandchallenge/github/migrations/0008_auto_20220201_1203.py
```diff
@@ -0,0 +1,39 @@
...
+                    ("FAILURE", "Failure"),
+                    ("INVALID", "Invalid"),
+                    ("NOT_APPLICABLE", "Not Applicable"),
+                ],
+                default="NOT_APPLICABLE",
+                max_length=14,
+            ),
+        ),
+        migrations.RemoveField(
+            model_name="githubwebhookmessage", name="license_check_result",
```
jmsmkn, line 32: Note that I am removing the old field here and adding the JSON one back, will re-run the check on the existing messages.

app/grandchallenge/github/admin.py
```diff
@@ -25,15 +25,12 @@ class GitHubWebhookMessageAdmin(admin.ModelAdmin):
...
-        "license_check_result",
         "user_error",
     )
-    list_filter = (
-        "clone_status",
-        "has_open_source_license",
-    )
-    search_fields = ("payload",)
+    list_filter = ("clone_status",)
+    search_fields = ("payload", "license_check")
```
MikeOverkamp-diag, line 33: Maybe I'm just overlooking something, but I cannot find the `license_check` field on the model?

app/grandchallenge/github/models.py
MikeOverkamp-diag: Maybe check all licenses and as long as there's an open source license, the check passes?

MikeOverkamp-diag: Though I'd say multi-licensing is an edge case that might not be worth the added complexity.

jmsmkn: Exactly, as this is dynamic we can always change it later. Lets see what data we get.

app/grandchallenge/github/admin.py
jmsmkn: Good catch!

PR #2267, opened 2022-02-07
Deprecate _IMAGE answer types

### LINKED ISSUE #2266: Deprecate POLYGON_IMAGE and MULTIPLE_POLYGONS_IMAGE question types
Use the mask question types as alternatives.

jmsmkn:
Disables the link for launching reader studies with deprecated answer types, showing a warning telling the user to remove them. Disallows adding new questions with deprecated answer types.

Closes #2266

### DIFF
--- a/app/grandchallenge/algorithms/tasks.py
+++ b/app/grandchallenge/algorithms/tasks.py
@@ -84,7 +84,7 @@ def on_job_creation_error(self, task_id, *args, **kwargs):
         error_message += str(res)
 
     job.update_status(
-        status=job.FAILURE, error_message=error_message,
+        status=job.CANCELLED, error_message=error_message,
     )
 
     on_commit(linked_task.apply_async)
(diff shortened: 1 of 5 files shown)


app/grandchallenge/algorithms/tasks.py
```diff
@@ -84,7 +84,7 @@ def on_job_creation_error(self, task_id, *args, **kwargs):
         error_message += str(res)
 
     job.update_status(
-        status=job.FAILURE, error_message=error_message,
```
HarmvZ, line 87: Is this commit a leftover from a different changeset?

HarmvZ (approved):
Approved with one sidenote: A user could still start a readerstudy with deprecated questions by accessing the url directly, right? I'm not sure if we want to also disable this.

MikeOverkamp-diag (approved):

jmsmkn:
> Approved with one sidenote: A user could still start a readerstudy with deprecated questions by accessing the url directly, right? I'm not sure if we want to also disable this.

Yes, they could, but once we have migrated the existing questions there will not be any reader studies with these question types.

PR #2268, opened 2022-02-08
Enable archive item updating through api

amickan:
This enables adding and updating non-image files and values for an existing archive item. For interface values of the super kind image, it also allows changing the civ kind (e.g., from generic medical image to ct-image). 

- Adds an update api endpoint for archive items and the necessary serializer.
- Refactores archive item update code. 
- Also adds a read-only api endpoint for component interfaces. This is necessary for the update function in gcapi: https://github.com/DIAGNijmegen/rse-gcapi/pull/102

This is the last step for this pitch: https://github.com/DIAGNijmegen/rse-roadmap/issues/114


app/grandchallenge/api/urls.py
```diff
@@ -69,6 +70,9 @@
     basename="upload-session",
 )
 
+# Component Interfaces
+router.register(r"interfaces", ComponentInterfaceViewSet, basename="interface")
```
jmsmkn, line 74: I would prefer to have the URL as `"components/interfaces"` with `basename="components-interface"`, just in case we want to add an all-components endpoint later (eg, would be useful for pipelines).

app/grandchallenge/archives/serializers.py
```diff
@@ -44,3 +51,67 @@ class Meta:
...
+
+
+class ArchiveItemPostSerializer(ArchiveItemSerializer):
+    archive = HyperlinkedRelatedField(
+        queryset=Archive.objects.none(),
+        view_name="api:archive-detail",
+        write_only=True,
+    )
+
+    class Meta:
```
jmsmkn, line 63: I think you can use inheritence here to save duplicating the config.

app/grandchallenge/archives/tasks.py
```diff
@@ -81,3 +83,90 @@ def update_archive_item_values(
         archive_item = ArchiveItem.objects.get(pk=archive_item_pk)
         archive_item.values.remove(*civ_pks_to_remove)
         archive_item.values.add(*civ_pks_to_add)
+
+
+def update_archive_item_update_kwargs(
```
jmsmkn, line 88: Could you add a docstring to this method? It's not really clear to me what this should do from the name

app/grandchallenge/archives/tasks.py
```diff
@@ -81,3 +83,90 @@ def update_archive_item_values(
...
+                    kwargs={
+                        "component_interface_value_pk": civ_pk,
+                        "upload_session_pk": upload_pk,
+                    },
+                    immutable=True,
+                ),
+            )
+            for civ_pk, upload_pk in upload_pks.items()
+        )
+        tasks = chord(image_tasks, tasks)
```
jmsmkn, line 169: I see that this comes from elsewhere but is there any way to avoid using a chord here? They're often broken in Celery.

jmsmkn (commented):
Looks great! Couple of small questions

app/grandchallenge/archives/tasks.py
amickan: Oh yes, the image tasks and the archive item updating can happen in parallel, so we can just use `group()`.

jmsmkn (approved):

PR #2271, opened 2022-02-09
Use ComponentInterfaceValues in reader studies through Display Sets

MikeOverkamp-diag:
This PR introduces an intermediary `DisplaySet` models, which allows the use of ComponentInterfaceValues rather than images in reader studies. The hanging list is now determined via the display sets. A mapping from ComponentInterface to image port can be provided in the reader study's settings.

The new functionality is hidden behind a feature flag, which can be toggled in the admin. No changes have been made to CIRRUS to support these changes yet.

Note that only image type interface values can be added via the web interface. For more complex structures, the api needs to be used. The reason for this is that the view would get very sloppy if other interface types are to be included. This can of course be  implemented, but for now I feel it would overcomplicate things for a use case that will probably not be very common.


app/grandchallenge/reader_studies/views.py
```diff
@@ -577,6 +622,29 @@ def get_success_url(self):
...
+                "linked_task": create_display_sets_for_upload_session.signature(
+                    kwargs={"reader_study_pk": self.reader_study.pk},
+                    immutable=True,
+                ),
+            }
+        )
+        return kwargs
+
+    def get_success_url(self):
+        return self.reader_study.get_absolute_url()
```
jmsmkn, line 645: Isn't it better to go to the upload session detail page?

app/grandchallenge/reader_studies/views.py
```diff
@@ -867,6 +935,91 @@ def ground_truth(self, request, pk=None, case_pk=None):
...
+    def get_serializer_class(self):
+        if self.action in ["partial_update", "update", "create"]:
+            return DisplaySetPostSerializer
+        return DisplaySetSerializer
+
+    def partial_update(self, request, pk=None):
+        instance = self.get_object()
+        if request.data.get("value"):
+            civ = ComponentInterfaceValue.objects.get(
+                id=request.data.get("value")
```
jmsmkn, line 969: Have the permissions been checked on this?

app/grandchallenge/reader_studies/views.py
```diff
@@ -867,6 +935,91 @@ def ground_truth(self, request, pk=None, case_pk=None):
...
+            )
+            civ.displays_set.clear()
```
jmsmkn, line 971: I'm having a hard time following this. Would be really handy to split this up into some methods so that the higher-level logic is clear. What is happening with clearing all display sets here? Could it be that the civ belongs to multiple display sets (eg from other reader studies, see the archive -> reader study link).

app/grandchallenge/reader_studies/views.py
```diff
@@ -867,6 +935,91 @@ def ground_truth(self, request, pk=None, case_pk=None):
...
+            assigned_civs = instance.values.filter(interface=civ.interface)
+            for assigned in assigned_civs:
+                ds = DisplaySet.objects.create(
+                    reader_study=instance.reader_study
+                )
+                ds.values.add(assigned)
+                instance.values.remove(assigned)
+            instance.values.add(civ)
+            instance.refresh_from_db()
```
jmsmkn, line 980: Why the refresh?

jmsmkn (commented):
I've not looked at the tests but the rest looks good, a few things to work on in the meantime. @amickan would be good if you can have a look at the display sets updating tasks and views as it's similar to what you just worked on for archive items.

app/grandchallenge/archives/views.py
```diff
@@ -658,7 +659,13 @@ def form_valid(self, form):
         reader_study: ReaderStudy = form.cleaned_data["reader_study"]
         images = form.cleaned_data["images"]
 
-        reader_study.images.add(*images)
+        if reader_study.use_display_sets:
+            civs = ComponentInterfaceValue.objects.filter(image__in=images)
+            for civ in civs:
+                ds = DisplaySet.objects.create(reader_study=reader_study)
```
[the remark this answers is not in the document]
MikeOverkamp-diag: Wouldn't that complicate things with the api endpoint for instance? As we'd then have to both collect display sets as well as archive items?

app/grandchallenge/reader_studies/models.py
```diff
@@ -321,6 +326,8 @@ class ReaderStudy(UUIDModel, TitleSlugDescriptionModel):
         help_text="The organizations associated with this reader study",
         related_name="readerstudies",
     )
+    image_port_mapping = models.JSONField(null=True)
```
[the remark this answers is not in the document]
MikeOverkamp-diag: That's part of https://github.com/DIAGNijmegen/rse-roadmap/issues/122. I think it makes sense to let the CIRRUS team decide on the best format for the field.

app/grandchallenge/reader_studies/views.py
MikeOverkamp-diag: It is, I talked about this with Chris yesterday.

app/grandchallenge/archives/views.py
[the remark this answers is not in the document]
jmsmkn: My feeling is that it makes sense to only work with archive items and create display sets from those as this is how we do archives -> algorithms, but I could be missing something. Can you go into more details on your concerns there?

app/grandchallenge/reader_studies/models.py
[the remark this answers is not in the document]
jmsmkn: Maybe it's best to remove the field for now as I think it overlaps with https://github.com/DIAGNijmegen/rse-roadmap/issues/119, maybe that can tackle the data model for Archives, Algorithms and Reader Studies at the same time.

app/grandchallenge/archives/views.py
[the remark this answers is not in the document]
MikeOverkamp-diag: Ah sorry, I think I misunderstood what you meant. That does make more sense.

app/grandchallenge/reader_studies/models.py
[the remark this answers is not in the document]
MikeOverkamp-diag: It does. I'll remove the field as it is not used anywhere at this point.

app/grandchallenge/reader_studies/tasks.py
```diff
@@ -39,6 +48,52 @@ def add_scores(*, instance_pk, pk_set):
...
+    interface = ComponentInterface.objects.get(pk=interface_pk)
+    for image in images:
+        with transaction.atomic():
+            civ, _ = ComponentInterfaceValue.objects.get_or_create(
+                interface=interface, image=image
+            )
+            if DisplaySet.objects.filter(
+                reader_study=reader_study, values=civ
+            ).exists():
+                return
```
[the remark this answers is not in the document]
MikeOverkamp-diag: Nice catch, should be a `continue`

app/grandchallenge/cases/serializers.py
```diff
@@ -193,11 +214,13 @@ def validate(self, attrs):
...
+            or "display_set" in attrs
         ):
             raise ValidationError(
-                "An interface can only be defined for archive "
-                "or archive item uploads."
+                "An interface can only be defined for archive, "
+                "archive item or display set uploads."
             )
 
         if "archive_item" in attrs and "interface" not in attrs:
```
amickan, line 226: If a `display_set` is defined, an `interface` also needs to be defined, right? Or does it make sense to have default (`generic-medical-image`)? A test for this would also be good.

app/grandchallenge/reader_studies/views.py
amickan: I'm also having a hard time following this. Maybe a test would help? Maybe I missed it, but I only see a test that updates the `values` of a display_set, but not one that provides the id of a CIV as update argument?

amickan (commented):
Looks good! Just two tiny comments regarding missing tests.

app/grandchallenge/reader_studies/tasks.py
```diff
@@ -39,6 +48,52 @@ def add_scores(*, instance_pk, pk_set):
...
+            ds.values.add(civ)
+            assign_perm(
+                f"view_{ds._meta.model_name}", reader_study.editors_group, ds
+            )
+            assign_perm(
+                f"view_{ds._meta.model_name}", reader_study.readers_group, ds
+            )
+            assign_perm(
+                f"change_{ds._meta.model_name}", reader_study.editors_group, ds
+            )
```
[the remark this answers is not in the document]
MikeOverkamp-diag: I actually think this bit of code is redundant, as it's handled in the DisplaySet save function. These permissions should only be assigned if the reader study is changed, which can only happen on the initial save. There is no option to change the reader study on an existing displays set.

app/grandchallenge/reader_studies/views.py
MikeOverkamp-diag: I added some comments and a test. Splitting it up into smaller functions made things a bit messy imo.

app/grandchallenge/reader_studies/forms.py
```diff
@@ -174,6 +179,7 @@ class Meta(ReaderStudyCreateForm.Meta):
         widgets = {
             "hanging_list": JSONEditorWidget(schema=HANGING_LIST_SCHEMA),
             "case_text": JSONEditorWidget(schema=CASE_TEXT_SCHEMA),
+            "image_port_mapping": JSONEditorWidget(),
```
jmsmkn, line 182: `image_port_mapping` no longer exists, right?

app/grandchallenge/reader_studies/views.py
```diff
@@ -867,6 +932,107 @@ def ground_truth(self, request, pk=None, case_pk=None):
...
+            civs = instance.reader_study.display_sets.values_list(
+                "values", flat=True
+            )
+            civ = ComponentInterfaceValue.objects.filter(id__in=civs).get(
+                id=request.data.get("value")
+            )
+
+            # Because the civ is added to a display set, it can no longer be
+            # a value for another display set.
+            civ.display_sets.clear()
```
jmsmkn, line 975: Why not, what about it belonging to a display set of another reader study?

MikeOverkamp-diag: Good point. It shouldn't be a part of any other display set in this reader study though right?

app/grandchallenge/reader_studies/views.py
```diff
@@ -867,6 +932,107 @@ def ground_truth(self, request, pk=None, case_pk=None):
...
+                many=True, data=request.data.get("values")
+            )
+            if serialized_data.is_valid():
+                civs = serialized_data.create(serialized_data.validated_data)
+                instance.values.remove(
+                    *instance.values.filter(
+                        interface__in=[civ.interface for civ in civs]
+                    )
+                )
+                instance.values.add(*civs)
```
jmsmkn, line 1009: Again, should the values be updated when there are already answers?

app/grandchallenge/reader_studies/views.py
```diff
@@ -867,6 +932,107 @@ def ground_truth(self, request, pk=None, case_pk=None):
...
+        serializer = self.get_serializer(instance)
+        return Response(serializer.data)
+
+    def get_queryset(self):
+        queryset = DisplaySet.objects.all().select_related("reader_study")
+        unanswered_by_user = self.request.query_params.get(
+            "unanswered_by_user"
+        )
+        if unanswered_by_user == "True":
+            queryset = queryset.exclude(answers__creator=self.request.user)
```
jmsmkn, line 1019: Does this cover the case of partial answers? Eg, an editor adds a new question after the user has gone through the study.

MikeOverkamp-diag: It does not. For the implementation: can you think of an instance where you'd want to filter by unanswered but not by reader study? The query will be a lot more efficient if we can just use a single reader study.

jmsmkn:
I still think that some thought is needed around when a display set can and cannot be changed. We have that already for Answers where they are m2m related to images. I guess that when a display set has answers that the `values` cannot be changed at all. I don't know the best way to handle that, maybe with an m2m signal?

MikeOverkamp-diag:
I think that makes most sense as well. I'll think about a solution.

app/grandchallenge/reader_studies/views.py
jmsmkn: The use case I'm thinking of (which does happen now I think) is:

- Editor creates a reader study with x questions
- Reader answers y display sets, creating x * y answers 
- Editor realises they missed something, creates a new question
- Reader opens the study, for the first y display sets they answer just the 1 new question with their existing answers in play, for display set > y they answer all x + 1 questions as they go

All the same reader study.

app/grandchallenge/reader_studies/views.py
jmsmkn: I think that users want to re-use images within the same study, see the usage of "disable hanging list validation".

MikeOverkamp-diag: Is that something we want to allow by default or should there be a flag for it like there is now?

jmsmkn:
What is the ETA on this PR? I would like to do the black upgrade as this is still waiting from last week, it touches 331 files so it would be easier if this is on master. If this is going to take a while then maybe split things up. The models look fine but the problems seem contained only to the new serializers and views.

MikeOverkamp-diag:
Should be done before the end of the day.

jmsmkn:
Actually, it's not bad at all, only 1 conflict, I will go ahead.

MikeOverkamp-diag:
That's not bad. I'll fix the conflict and rerun black when I've implemented the final changes.

jmsmkn:
It's a good idea to merge in the current master though as there are changes from #2285 that need accommodating.

MikeOverkamp-diag:
Yes, I was planning on doing that, since I expect the pre-commit to fail for the PR anyway if I wouldn't.

app/grandchallenge/reader_studies/tasks.py
```diff
@@ -37,6 +46,43 @@ def add_scores(*, instance_pk, pk_set):
...
+    interface = ComponentInterface.objects.get(pk=interface_pk)
+    with transaction.atomic():
+        for image in images:
+            civ, _ = ComponentInterfaceValue.objects.get_or_create(
+                interface=interface, image=image
+            )
+            if DisplaySet.objects.filter(
+                reader_study=reader_study, values=civ
+            ).exists():
+                continue
```
jmsmkn, line 81: I thought that we wanted to allow re-use of CIVs in a study?

app/grandchallenge/reader_studies/models.py
```diff
@@ -879,6 +977,27 @@ def statistics(self):
             "questions": questions,
         }
 
+    @cached_property
+    def values_for_interfaces(self):
+        cache_key = f"{self.slug}-{self.modified.timestamp()}"
```
jmsmkn, line 982: There probably won't be a clash but ideally the key should include the app and model name

app/grandchallenge/reader_studies/models.py
```diff
@@ -899,6 +1018,63 @@ def delete_reader_study_groups_hook(*_, instance: ReaderStudy, using, **__):
...
+        )
+        assign_perm(
+            f"view_{self._meta.model_name}",
+            self.reader_study.readers_group,
+            self,
+        )
+
+    @cached_property
+    def empty_interfaces(self):
+        cache_key = f"{self.pk}-{self.modified.timestamp()}"
```
jmsmkn, line 1055: Ideally the key should include the app and model name

app/grandchallenge/reader_studies/tasks.py
MikeOverkamp-diag: We do, but in this case the only way a display set for this civ can already exist is if this task gets called twice. This task assigns a newly created upload to a civ, so this is the proper behavior here imo.

app/grandchallenge/reader_studies/models.py
```diff
@@ -879,6 +977,27 @@ def statistics(self):
...
+        interfaces = self.display_sets.values_list(
+            "values__interface__slug", flat=True
+        )
+        values_for_interfaces = {}
+        for interface in interfaces:
+            values = ComponentInterfaceValue.objects.none()
+            for ds in self.display_sets.all():
+                values |= ds.values.filter(interface__slug=interface)
+            values_for_interfaces[interface] = values
+        cache.set(cache_key, values_for_interfaces)
```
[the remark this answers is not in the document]
MikeOverkamp-diag: We could probably use a timeout of 0 here actually, since the cache key includes the modification timestamp. I'd say we can cache this until the timestamp changes and then maybe we should actively remove any existing cache keys for this reader study?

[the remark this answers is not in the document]
jmsmkn: Does the modified time change when any of the values of the display sets change?

app/grandchallenge/reader_studies/signals.py
```diff
@@ -60,3 +66,45 @@ def assign_score(instance, action, reverse, model, pk_set, **_):
...
+
+@receiver(m2m_changed, sender=DisplaySet.values.through)
+def assert_modification_allowed(instance, action, reverse, model, pk_set, **_):
+    if "pre" not in action:
+        return
+
+    if reverse:
+        not_editable = []
+        for ds in DisplaySet.objects.filter(pk__in=pk_set):
+            if not ds.is_editable:
```
jmsmkn, line 79: N+1 problem here.

jmsmkn (approved):

PR #2273, opened 2022-02-11
Viewer config tooltips

chrisvanrun:
This PR adds several help texts to viewer configuration options: users got confused by some of the settings.

In addition, the detail view for a configuration now has a tooltip on hover that shows these help text.

I ended up not editing the tooltip looks because the SCSS construct to do so eluded me and it wasn't worth the time to further investigate.

https://user-images.githubusercontent.com/7871310/153599456-36569019-80c6-4d50-b2ce-ba72e60e9154.mp4


app/grandchallenge/workstation_configs/templates/workstation_configs/workstationconfig_detail.html
```diff
@@ -194,5 +202,13 @@ <h1>Viewer Configuration {{ object.title }}</h1>
...
     {% endif %}
+{% endblock %}
 
+{% block script %}
+    {{ block.super }}
+    <script type="text/javascript">
+      $(function() {
+        $('[data-toggle="tooltip"]').tooltip();
+      });
+    </script>
```
jmsmkn, line 213: I would like to add a CSP to grand challenge, so I don't want to add any more inline JS. Please move this to its own file, there might be one that does this already.

app/grandchallenge/workstation_configs/templatetags/get_help_text.py
```diff
@@ -0,0 +1,8 @@
+from django import template
+
+register = template.Library()
+
+
+@register.simple_tag
+def get_help_text(obj, field):
+    return obj.get_help_text(field)
```
jmsmkn, line 8: Nitpick but it would be easier for re-use if this tag did `obj._meta.get_field(field).help_text` rather than having to define `get_help_text` on the model to use it.

jmsmkn (commented):
Looks good, check for that tooltip snippet as I think that it already exists.

app/grandchallenge/workstation_configs/templatetags/get_help_text.py
chrisvanrun: That would mean it accesses a 'protected' member of the object. Would that still be okay?

chrisvanrun:
> Looks good, check for that tooltip snippet as I think that it already exists.

Only reference in the codebase seems to be in the cornerstone viewer. I'll inquire what a 'CSP' does later ;-).

jmsmkn (approved):

PR #2275, opened 2022-02-16
Handle request errors in CSL update task

jmsmkn:
Demotes a few sentry alerts to warnings and continues. Uses the 2xlarge queue as this takes a while. I think this is fine without an atomic block as it is a single update, and doesn't matter if this task runs many times.


amickan (approved):
LGTM! I agree that it doesn't matter if this tasks runs multiple times.

PR #2278, opened 2022-02-17
Add AWS_S3_DEFAULT_STORAGE_CLASS option

jmsmkn:
AWS made some changes to intelligent tiering (https://aws.amazon.com/blogs/aws/amazon-s3-intelligent-tiering-further-automating-cost-savings-for-short-lived-and-small-objects/) which means that:

-    S3 Intelligent-Tiering now has no minimum storage duration period for all objects.
-    Monitoring and automation charges are no longer collected for objects smaller than 128 KB.

I don't see any good reason not to use this now for all of the long-term storage buckets. The Uploads bucket does not use django storages so does not pick up this configuration, but the objects there are shorted-lived and deleted within 24 hours so is anyway low cost.


amickan (dismissed):
Very nice!

PR #2281, opened 2022-02-18
Adds default interpolation for base image option to workstation config

### LINKED ISSUE #2280: Incorrect mapping of image interpolation values
There seems to be a mismatch between image interpolation values on gc and cirrus. 

The types defined for image interpolation on grand challenge:

https://github.com/comic/grand-challenge.org/blob/16066abeed6315549bc815e69a9f0ee567447c9c/app/grandchallenge/workstation_configs/models.py#L142-L144

The values for the Enum in mevislab are: `FILTER_NEAREST`, `FILTER_LINEAR`, `FILTER_LINEAR_POSTCLASS`, `FILTER_CUBIC_POSTCLASS`.
In CIRRUS currently we use `FILTER_NEAREST` if the value from GC is `NearestNeighbor` otherwise it is set to `FILTER_CUBIC_POSTCLASS`. That means that `Trilinear` maps to `FILTER_CUBIC_POSTCLASS` which is incorrect.

HarmvZ:
**Based on #2273 since that PR is not yet ready to merge and basing these changes on master would then introduce unnecessary conflicts. This can be reviewed an merged into that PR.**

Does **not** fix #2280 because that probably needs an extra data migration if we decide to fix it like that.


app/grandchallenge/workstation_configs/models.py
```diff
@@ -221,6 +222,14 @@ class ImageInterpolationType(models.TextChoices):
         help_text="The method used to interpolate multiple voxels of overlay images and project them to screen pixels",
     )
 
+    default_interpolation = models.CharField(
```
jmsmkn, line 225: This won't be applied to Overlays, right? Could we call it `default_image_interpolation`?

HarmvZ: Done

jmsmkn (approved):

PR #2282, opened 2022-02-21
Adds option to hide a challenge phase

amickan:
- This PR adds the option to hide a phase and its leaderboard from a challenge. 
- Participants of the challenge can not see the submit and leaderboard tabs of a hidden phase and cannot view their previous submissions / evluations to that phase (should they exist). 
- Admins keep access to the hidden phase, its leaderboard and submissions and evaluations.
- Hiding a phase is only possible if it is closed for submissions.


app/grandchallenge/evaluation/models.py
```diff
@@ -493,6 +529,15 @@ def clean(self):
                 "the submissions open date."
             )
 
+        if self.hidden and (
+            self.submission_limit != 0 or self.open_for_submissions
```
jmsmkn, line 533: Don't we only need to check if the submission limit is zero?

jmsmkn: `open_for_submissions` already includes the submission limit check.

app/grandchallenge/evaluation/models.py
```diff
@@ -863,25 +912,38 @@ def assign_permissions(self):
...
+            if viewer_group:
+                assign_perm("view_evaluation", viewer_group, self)
         else:
-            remove_perm("view_evaluation", viewer_group, self)
+            if viewer_group:
+                remove_perm("view_evaluation", viewer_group, self)
 
-        remove_perm("view_evaluation", non_viewer_group, self)
+        for non_viewers in non_viewer_group:
+            remove_perm("view_evaluation", non_viewers, self)
```
jmsmkn, line 946: This is quite difficult to reason about now. Could you simplify it?

Something like:
```
all_users_can_view = self.published and not self.submission.phase.hidden and not self.submission.phase.challenge.hidden

if all_users_can_view:
  assign_perm("view_evaluation", all_user_group, self)
else:
  remove_perm("view_evaluation", all_user_group, self)

participants_can_view = self.published and not self.submission.phase.hidden

if participants_can_view:
  assign_perm("view_evaluation", participants_group, self)
else:
  remove_perm("view_evaluation", participants_group, self)
```

app/grandchallenge/evaluation/tasks.py
```diff
@@ -468,13 +468,31 @@ def _update_evaluations(*, evaluations, final_positions):
     )
 
 
-@shared_task
-def assign_evaluation_permissions(*, challenge_pk: uuid.UUID):
+@shared_task(**settings.CELERY_TASK_DECORATOR_KWARGS["acks-late-micro-short"])
+def assign_evaluation_permissions(*, challenge_pk: uuid.UUID, phase_pk=None):
```
jmsmkn, line 472: Maybe it's better to only send `phase_pk` (or a list of them) to this method?

jmsmkn: Same again, this method could take a while.

app/grandchallenge/evaluation/models.py
```diff
@@ -429,6 +435,17 @@ class SubmissionKind(models.IntegerChoices):
             MaxValueValidator(limit_value=4 * 60 * 60),
         ],
     )
+    hidden = models.BooleanField(
```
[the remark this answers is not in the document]
jmsmkn: I see why you used hidden as this is what Challenges use, but they are the odd one out and I would like to also move them to public at some point.

jmsmkn (approved):
LGTM, I remove the use of celery `group` as there is currently a weird bug in them (see #2297), so we should avoid them.

PR #2287, opened 2022-02-24
Use stable version of black

jmsmkn:


MikeOverkamp-diag (approved):
Not sure how I feel about the changes black made wrt trailing commas, but the PR looks good.

PR #2288, opened 2022-02-24
Add isort

jmsmkn:
At the moment we rely on developers to configure and run the "optimize import" option in Pycharm before committing. This increases the burden for contributing. This is a proposal to replace the flake8 import order inspection for `isort`, which automatically sorts the imports on commit.

I have found a configuration that makes minimal changes to what we do now.

```
[tool.isort]
profile = "black"
known_first_party = ["config", "grandchallenge", "tests"]
line_length = 79
case_sensitive = true
order_by_type = false
```

There are a couple of options that we could change beyond the black default:

## case_sensitive = True

Current situation:

```python
from django.forms import (
    BooleanField,
    CharField,
    FileField,
    Form,
    ModelChoiceField,
    ModelForm,
    TextInput,
    Textarea,
)
```

Removing `case_sensitive = True` changes the ordering of `TextInput` and `TextArea`:

```python
from django.forms import (
    BooleanField,
    CharField,
    FileField,
    Form,
    ModelChoiceField,
    ModelForm,
    Textarea, # Here
    TextInput,
)
```

## order_by_type = false

Current situation:

```python
from grandchallenge.reader_studies.models import (
    Answer,
    CASE_TEXT_SCHEMA,
    CategoricalOption,
    HANGING_LIST_SCHEMA,
    Question,
    ReaderStudy,
    ReaderStudyPermissionRequest,
)
```

Removing `order_by_type = false` moves the constants to the top:

```python
from grandchallenge.reader_studies.models import (
    CASE_TEXT_SCHEMA, # Here
    HANGING_LIST_SCHEMA, # and here
    Answer,
    CategoricalOption,
    Question,
    ReaderStudy,
    ReaderStudyPermissionRequest,
)
```

I propose we remove these two options but it will mean a bigger PR.


miriam-groeneveld:
I'm not liking the bike shedding, you can never have too many bikes :).
I prefer the case_sensitive off, (so Textarea before TextInput). 
The order_by_type, mwah, I'm neutral on that, either way is fine

pkcakeout:
IMO, I do not care much about the specific ordering of the import statements in the first place. Of course, they are important, of course they are must be there. But as long as I can find the relevant imports with CTRL+Click or CTRL+F I am please. So anything that does not require me to think how these are ordered or whatever is very much welcome. And I do not mind what flags are enabled or not - so looks good from my side! :+1:

jmsmkn:
> Just aesthetically, I like grouping by
> 
>     1. system libs
> 
>     2. 3rd party libs
> 
>     3. "local" imports

Yes me too, that ordering is maintained. Indeed, the only aim of this is to get standardization between many developers, without it things just get a bit of a mess.

> I prefer the case_sensitive off, (so Textarea before TextInput).
> The order_by_type, mwah, I'm neutral on that, either way is fine

Me too, I'm inclinded to remove both cutomisations unless there are any strong objections.

jmsmkn:
Then it is decided, the bike sheds shall be dark green.

PR #2306, opened 2022-03-04
Add display_set to answer serializer

chrisvanrun:
Small addition for working with display sets in CIRRUS

### DIFF
--- a/app/grandchallenge/reader_studies/serializers.py
+++ b/app/grandchallenge/reader_studies/serializers.py
@@ -143,6 +143,10 @@ class AnswerSerializer(HyperlinkedModelSerializer):
     answer_image = HyperlinkedRelatedField(
         read_only=True, view_name="api:image-detail"
     )
+    display_set = HyperlinkedRelatedField(
+        queryset=DisplaySet.objects.all(),
+        view_name="api:reader-studies-display-set-detail",
+    )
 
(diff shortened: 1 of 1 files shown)


app/grandchallenge/reader_studies/serializers.py
```diff
@@ -143,6 +143,10 @@ class AnswerSerializer(HyperlinkedModelSerializer):
     answer_image = HyperlinkedRelatedField(
         read_only=True, view_name="api:image-detail"
     )
+    display_set = HyperlinkedRelatedField(
+        queryset=DisplaySet.objects.all(),
```
jmsmkn, line 147: We need to check that the user has permission to create answers for this `display_set` if this field is not read only. This `queryset` (and the one for the images and questions) need to be filtered.

chrisvanrun: Definitely needs to be read_only! I'll add some tests to also cover the other members.

Although I see that those tests are rendering entire views. So those are ready to be optimized first.

jmsmkn: This is quite tricky, so maybe @MikeOverkamp-diag can help.

jmsmkn:
I'm going to close this as I think it needs a lot of work, please feel free to re-open when ready.

PR #2311, opened 2022-03-08
Validate the algorithm inputs and outputs are disjoint

jmsmkn:
Required for https://github.com/DIAGNijmegen/rse-roadmap/issues/119


miriam-groeneveld (approved):
Looks good. How about existing algorithms? Or are there none that have the same interface in both input and output?

PR #2312, opened 2022-03-08
Challenge request form

amickan:
This adds a challenge request form and a reviewer group with permission to view and update the challenge requests. 
Emails are sent when a request is created and when it is updated. 
When a request is accepted, a new challenge is created automatically.

Addresses https://github.com/DIAGNijmegen/rse-roadmap/issues/117

image

image

image


app/config/settings.py
```diff
@@ -87,6 +87,7 @@ def strtobool(val) -> bool:
 ANONYMOUS_USER_NAME = "AnonymousUser"
 REGISTERED_USERS_GROUP_NAME = "__registered_users_group__"
 REGISTERED_AND_ANON_USERS_GROUP_NAME = "__registered_and_anonymous_users__"
+CHALLENGE_REVIEWERS_GROUP_NAME = "challenge_reviewers"
```
jmsmkn, line 90: I would prefer not to add any new special groups. If you need this group in the tests then create a pytest fixture for it. We should then leave it up to the site admins to create groups that are people centric, and attach the permissions to those.

app/config/settings.py
```diff
@@ -128,6 +129,11 @@ def strtobool(val) -> bool:
 # About Flatpage
 FLATPAGE_ABOUT_URL = os.environ.get("FLATPAGE_ABOUT_URL", "/about/")
 
+# Costs
+AWS_FILE_STORAGE_COSTS = os.environ.get("AWS_FILE_STORAGE_COSTS", 0.00004)
+AWS_DOCKER_STORAGE_COSTS = os.environ.get("AWS_DOCKER_STORAGE_COSTS", 0.04)
+AWS_COMPUTE_COSTS = os.environ.get("AWS_COMPUTE_COSTS", 0.3)
```
jmsmkn, line 135: The `AWS_` namespace is reserved so maybe use `CHALLENGES_` instead (the costs also include non-aws things too such as off site backups). It is a good idea to add units to the names of the variables, and if you are doing things with currency then it is best to use integer arithmetic with (euro) cents as the unit, so `_CENTS_PER_GB_PER_MONTH` etc.

app/grandchallenge/challenges/emails.py
```diff
@@ -1,6 +1,8 @@
+from django.contrib.auth.models import Group
 from django.contrib.sites.models import Site
-from django.core.mail import mail_managers
+from django.core.mail import mail_managers, send_mail
 
+from config import settings
```
jmsmkn, line 5: Do not import settings from config, use `from django.conf import settings`. Maybe we should add a pre-commit check for this?

app/grandchallenge/challenges/emails.py
```diff
@@ -42,3 +44,133 @@ def send_external_challenge_created_email(challenge):
...
+        f"Dear reviewers,\n\n"
+        f"User {challengerequest.creator} has just requested the challenge "
+        f"{challengerequest.title}. To review the challenge, go here:"
+        f"{update_url}\n\n"
+        f"Regards,\n"
+        f"{site.name}\n\n"
+        f"This is an automated service email from {site.domain}."
+    )
+    reviewers = (
+        Group.objects.filter(name=settings.CHALLENGE_REVIEWERS_GROUP_NAME)
```
jmsmkn, line 64: Rather than using the group look for users who have the global permission (either directly or through a group), django guardian has methods to do that.

app/grandchallenge/challenges/emails.py
```diff
@@ -42,3 +44,133 @@ def send_external_challenge_created_email(challenge):
...
+        addition += (
+            f"For your type 2 challenge, we have calculated the following budget estimate "
+            f"based on the information you provided:\n"
+            f"{budget}\n\n"
+        )
+
+    message = (
+        f"Dear {challengerequest.creator.username},\n\n"
+        f"Your challenge request has been sent to the reviewers. You will "
+        f"receive an email informing you of our decision within the next 4 weeks. "
```
jmsmkn, line 96: 4 weeks is quite a long time for challenge admins, would be good if this is quicker!

app/grandchallenge/challenges/emails.py
```diff
@@ -42,3 +44,133 @@ def send_external_challenge_created_email(challenge):
...
+
+def send_challenge_status_update_email(challengerequest, challenge=None):
+    site = Site.objects.get_current()
+    message = ""
+    if (
+        challengerequest.status
+        == challengerequest.ChallengeRequestStatusChoices.ACCEPTED
+    ):
+        message = (
+            f"Dear {challengerequest.creator.username},\n\n"
```
jmsmkn, line 121: These emails are currently sent plain text, but it is good practise to always use `format_html` to automatically escape any user input when building templates. You can see these reflective things in a lot of the spam email, where spammers create accounts with usernames like `<a href="my-phising-site">CLICK ME</a>` etc etc.

app/grandchallenge/challenges/forms.py
```diff
@@ -179,3 +190,265 @@ class Meta:
...
+        )
+        widgets = {
+            "start_date": forms.TextInput(attrs={"type": "date"}),
+            "end_date": forms.TextInput(attrs={"type": "date"}),
+            "challenge_type": forms.Select(
+                attrs={"onchange": "updateForm();"}
+            ),
+        }
+        help_texts = {
+            "challenge_short_name": "Short name that will be used in the URL "
```
jmsmkn, line 245: This is quite difficult to parse and find the other keys, could you wrap the help text in brackets? eg:

```
"challenge_short_name": (
    "Short name that will be used in the URL "
    ...
)
```

app/grandchallenge/challenges/models.py
```diff
@@ -1,5 +1,6 @@
 import datetime
 import logging
+from functools import cached_property
```
jmsmkn, line 3: I think that you want to use the one from django, not functools.

jmsmkn: Oh wait, why was that import changed?

app/grandchallenge/challenges/models.py
```diff
@@ -676,3 +685,284 @@ def delete_challenge_follows(*_, instance: Challenge, **__):
         app_label=instance._meta.app_label, model=instance._meta.model_name
     ).get()
     Follow.objects.filter(object_id=instance.pk, content_type=ct).delete()
+
+
+class ChallengeRequest(models.Model):
+    class ChallengeTypeChoices(models.TextChoices):
+        """Challenge type choices."""
+
+        T1 = "TYPE_1", _("Type 1 - prediction submission")
```
jmsmkn, line 694: You could use `IntegerChoices` for this.

app/grandchallenge/challenges/models.py
```diff
@@ -676,3 +685,284 @@ def delete_challenge_follows(*_, instance: Challenge, **__):
...
+        T2 = "TYPE_2", _("Type 2 - algorithm submission")
+
+    class ChallengeRequestStatusChoices(models.TextChoices):
+        ACCEPTED = "ACPT", _("Accepted")
+        REJECTED = "RJCT", _("Rejected")
+        PENDING = "PEND", _("Pending")
+
+    created = models.DateTimeField(auto_now_add=True)
```
jmsmkn, line 702: Always good to add `modified` too, maybe switch this to a UUID model?

app/grandchallenge/challenges/models.py
```diff
@@ -676,3 +685,284 @@ def delete_challenge_follows(*_, instance: Challenge, **__):
...
+    affiliated_event = models.CharField(
+        blank=True,
+        max_length=50,
+        help_text="Is this challenge part of a workshop or conference? If so, which one?",
+    )
+    structured_challenge_submission_form = models.FileField(
+        null=True,
+        blank=True,
+        upload_to=get_pdf_path,
+        storage=public_s3_storage,
```
jmsmkn, line 754: Public storage will mean that these are publicly accessible, I don't think that we want that. Probably you want this in protected storage and then an additional method to serve them to users that have permission. Check that `get_pdf_path` really puts it in the place that you want, especially if you move to UUIDs. Make sure it includes the namespace for the app and model.

jmsmkn: Also wonder if we should use the pdf here, or add the form fields to our site?

app/grandchallenge/challenges/models.py
```diff
@@ -676,3 +685,284 @@ def delete_challenge_follows(*_, instance: Challenge, **__):
...
+    )
+    task_types = models.ManyToManyField(
+        TaskType, blank=True, help_text="What type of task is this challenge?"
+    )
+    modalities = models.ManyToManyField(
+        ImagingModality,
+        blank=True,
+        help_text="What imaging modalities are used in this challenge?",
+    )
+    structures = models.ManyToManyField(
```
jmsmkn, line 768: A lot of these fields seem to be copied from the Challenge models, maybe it is best to define a mixin (eg, like UUIDModel)?

app/grandchallenge/challenges/models.py
```diff
@@ -676,3 +685,284 @@ def delete_challenge_follows(*_, instance: Challenge, **__):
...
+        help_text="Please indicate if you plan to coordinate a publication "
+        "of the challenge results."
+    )
+    code_availability = models.TextField(
+        help_text="Will the participants’ code be accessible after the challenge?"
+    )
+    expected_number_of_teams = models.IntegerField(
+        help_text="How many teams do you expect to participate in your challenge?"
+    )
+    average_algorithm_container_size = models.IntegerField(
```
jmsmkn, line 799: Good idea to include units in the variable here.

jmsmkn: And in `inference_time_limit`, `average_size_of_test_image`.

app/grandchallenge/challenges/models.py
```diff
@@ -676,3 +685,284 @@ def delete_challenge_follows(*_, instance: Challenge, **__):
...
+    def get_absolute_url(self):
+        return reverse("challenges:request-detail", kwargs={"pk": self.pk})
+
+    def save(self, *args, **kwargs):
+        adding = self._state.adding
+        super().save(*args, **kwargs)
+        if adding:
+            send_challenge_requested_email_to_reviewers(self)
+            send_challenge_requested_email_to_requester(self)
+        if self._orig_status != self.status:
```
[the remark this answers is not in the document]
jmsmkn: I would not do this in the same method, maybe move the challenge creation to an admin task?

app/grandchallenge/challenges/models.py
```diff
@@ -676,3 +685,284 @@ def delete_challenge_follows(*_, instance: Challenge, **__):
...
+            challenge = None
+            if self.status == self.ChallengeRequestStatusChoices.ACCEPTED:
+                challenge = Challenge.objects.create(
```
jmsmkn, line 859: This would be slightly better:

```
challenge = Challenge(
                    title=self.title,
                    short_name=self.challenge_short_name,
                    creator=self.creator,
                    hidden=True,
                    contact_email=self.contact_email,
                )
challenge.full_clean()
challenge.save()
```

Just in case the models get out of sync.

app/grandchallenge/challenges/models.py
```diff
@@ -676,3 +685,284 @@ def delete_challenge_follows(*_, instance: Challenge, **__):
...
+                )
+                challenge.task_types.set(self.task_types.all())
+                challenge.modalities.set(self.modalities.all())
+                challenge.structures.set(self.structures.all())
+
+                if self.challenge_type == self.ChallengeTypeChoices.T2:
+                    phase = challenge.phase_set.get()
+                    phase.submission_kind = phase.SubmissionKind.ALGORITHM
+                    phase.creator_must_be_verified = True
+                    phase.save()
```
jmsmkn, line 874: Again, `full_clean` before save.

app/grandchallenge/challenges/templates/challenges/challengerequest_detail.html
```diff
@@ -0,0 +1,53 @@
...
+        <li class="breadcrumb-item"><a href="{% url 'challenges:request-list' %}">Requests</a></li>
+        <li class="breadcrumb-item active">{{ object.title }}</li>
+    </ol>
+{% endblock %}
+
+{% block content %}
+    <h2>{{ object.title }}</h2>
+    <br>
+    <div class="row">
+        <p class="col-6">{% if object.status == 'ACPT' %}<span class="badge badge-success">Accepted</span>{% elif object.status == 'RJCT' %}<span class="badge badge-danger">Rejected</span>{% endif %} Challenge request by {{ object.creator|user_profile_link }}, created on {{ object.created }}</p>
```
jmsmkn, line 21: Do not use the raw status codes, use the variables you create in the enum.

app/grandchallenge/challenges/templates/challenges/challengerequest_form.html
```diff
@@ -0,0 +1,45 @@
...
+    {% else %}
+        <h2>Approve or reject "{{ object.title }}"</h2>
+        {% crispy form %}
+    {% endif %}
+{% endblock %}
+
+{% block script %}
+    {{ block.super }}
+    <script>
+        function updateForm(){
```
jmsmkn, line 28: This can go in its own file, we shouldn't add any new in-line js (and in fact move everything to files). Will help us with a CSP later.

app/grandchallenge/challenges/templates/challenges/challengerequest_list.html
```diff
@@ -0,0 +1,52 @@
...
+                {% for object in page_obj %}
+                    <tr>
+                        <td class="text-center">
+                            <a class="btn btn-dark btn-sm" href="{% url 'challenges:request-detail' pk=object.pk %}"><i class="fa fa-search"></i></a>
+                        </td>
+                        <td class="text-center align-middle">{{ object.title }}</td>
+                        <td class="text-center align-middle">{{ object.creator|user_profile_link }}</td>
+                        <td class="text-center align-middle">{{ object.challenge_type }}</td>
+                        <td class="text-center align-middle">
+                            {% if object.status == 'ACPT' %}
```
jmsmkn, line 39: Use the enum rather than the values.

app/grandchallenge/challenges/urls.py
```diff
@@ -32,4 +36,16 @@
...
+    path("request-list/", ChallengeRequestList.as_view(), name="request-list"),
+    path(
+        "request/<pk>/",
+        ChallengeRequestDetail.as_view(),
+        name="request-detail",
+    ),
+    path(
+        "request/<pk>/update/",
+        ChallengeRequestUpdate.as_view(),
+        name="request-update",
```
jmsmkn, line 49: We have a URL structure for this:

```
requests/ (list view, note the plural, name would be requests-list)
requests/create/ (create view, requests-create)
requests/<pk>/ (detail view)
requests/<pk>/update/ (update view)
```

app/tests/challenges_tests/factories.py
```diff
@@ -0,0 +1,33 @@
+from datetime import timedelta
+
+from django.utils.timezone import now
+
+from grandchallenge.challenges.models import ChallengeRequest
+from tests.factories import ChallengeRequestFactory
+
+
+def generate_type_2_challenge_request(creator):
```
jmsmkn, line 9: These should be pytest fixtures, you can then remove `generate_` from the name.

app/tests/challenges_tests/test_permissions.py
```diff
@@ -76,3 +79,59 @@ def _get_view():
...
 
             remove_perm(permission, u, obj)
+
+
+@pytest.mark.django_db
+def test_request_challenge_only_when_verified(client):
+    user = UserFactory()
+    assert not Verification.objects.filter(user=user)
+
+    response = get_view_for_user(
```
jmsmkn, line 89: Now we have some time to look at the tests It is better to test the view directly rather than using get_view_for_user. Have you looked at Adam Johnsons book on Django testing?

app/tests/challenges_tests/test_views.py
```diff
@@ -318,3 +324,116 @@ def test_challenge_card_status(
         assert f"{expected_status} for {title}" in response.rendered_content
     else:
         assert expected_status in response.rendered_content
+
+
+@pytest.mark.django_db
+def test_challenge_request_type_2_budget_fields_required(client):
```
jmsmkn, line 330: This test is good, do you really need to create and verify the user to test the form? If not, see if you can find a way to remove the `django_db` marker, everything will run much quicker this way. Better place for this test would be in `test_forms`.

jmsmkn (commented):
Looks great! Nice use of django crispy forms. It would be good if you can use some of this time to learn how to write faster tests, our test suite is getting really long to run as we (mostly me!) have not done a good job of writing fast tests, far too easy to reach for a full view test which adds a lot of overhead. It might also inform the design of your forms etc to make them testable and ensure a good separation of concerns.

app/grandchallenge/challenges/models.py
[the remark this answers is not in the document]
amickan: I'm not sure how that would work though. We need the challegne to be created before sending the acceptance email, so separating the challenge creation and email sending doesn't work. I could make the entire status update an admin action, but I think part of the point of this pitch is to avoid that the challenge review is dependent on one of us, right?

app/grandchallenge/challenges/emails.py
amickan: Agreed, we;ll actually be doing the review every 2 weeks, but I think it's better to give a generous time frame in this email or else we'll get swamped by emails asking for a status update. Same as we do for verifications.

app/grandchallenge/challenges/models.py
amickan: The pdf contains a lot of questions that I don't think we want to copy over into our form. I put it into public storage because the PDFs are actually publically available on the [BIAS website](https://www.biomedical-challenges.org/miccai2022/Overview). I also considered making this a field where the requester can paste the url to the form rather than uploading the form, but you can't link to a specific form on that page it turns out, so went with the PDF upload after all. 

But protected storage is fine as well if you think that's better. ;-)

app/grandchallenge/challenges/models.py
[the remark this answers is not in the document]
jmsmkn: I see, still I think the best thing to do is make this a separate method on the model. You could then call that automatically from the view when the state changes, or from an admin task if you like. That would still allow changing it from the admin or shell (etc) without automatically having these side effects.

app/grandchallenge/challenges/models.py
jmsmkn: I think protected because they're not there until after MICCAI review, I think users would be surprised if they are made public by us before they are made public by MICCAI.

amickan: Good point! Hadn't thought of that for some reason...

app/grandchallenge/challenges/models.py
amickan: Not sure actually! Using the one from django now.

app/tests/challenges_tests/test_views.py
amickan: The verification is indeed not necessary, but the user is I'm afraid because the ChallengeRequestForm needs a creator to be instantiated. So I think the test does need database access unfortunately, right?

jmsmkn: Would be strange if it needs the user saved in the db, it cannot take a user that is just built by `UserFactory.build()`?

amickan: The form instantiation does work with `UserFactory.build()` but the validation still requires database access (I've tried `is_valid()` and `full_clean()` and `clean()`. The first two require db access, the latter complains that `cleaned_data` does not exist.

jmsmkn: OK then no problem!

app/grandchallenge/challenges/emails.py
```diff
@@ -1,18 +1,25 @@
...
+    message = format_html(
+        "Dear manager,\n\n"
+        "User {} has just created the challenge {} at {}.\n\n"
+        "Regards,\n{}\n\n"
+        "This is an automated service email from {}.",
+        challenge.creator,
+        challenge.short_name,
+        challenge.get_absolute_url(),
+        site.name,
+        site.domain,
```
jmsmkn, line 22: Suggestion for the future - when you have a lot of arguments to `format_html` it is nicer to use kwargs as tracking these becomes difficult, eg

```
format_html(
        "Dear manager,\n\n"
        "User {user} has just created the challenge {challenge_name} at {challenge_url}.\n\n"
        "Regards,\n{site_name}\n\n"
        "This is an automated service email from {site_domain}.",
        user=challenge.creator,
        challenge_name=challenge.short_name,
        challenge_url=challenge.get_absolute_url(),
        site_name=site.name,
        site_domain=site.domain,
```

app/grandchallenge/challenges/urls.py
```diff
@@ -32,4 +36,20 @@
         ExternalChallengeUpdate.as_view(),
         name="external-update",
     ),
+    path("requests/", ChallengeRequestList.as_view(), name="requests-list"),
+    path(
+        "requests/create",
```
jmsmkn, line 41:
```suggestion
        "requests/create/",
```

app/grandchallenge/challenges/views.py
```diff
@@ -66,10 +79,10 @@ def get_context_data(self, *args, **kwargs):
                     (
                         "Here is an overview over the medical image analysis"
                         " challenges that have been hosted on Grand Challenge."
-                        "<br>Please <a href='{}'>contact us</a> if you would like "
-                        "to host your own challenge."
+                        "<br>Please fill in <a href='{}'>this form</a> "
+                        "if you would like to host your own challenge."
                     ),
-                    random_encode("mailto:support@grand-challenge.org"),
+                    random_encode(reverse("challenges:requests-create")),
```
jmsmkn, line 85: `random_encode` is not necessary here, this is only for making email addresses harder for spammers to scrape.

app/grandchallenge/serving/views.py
```diff
@@ -131,3 +132,19 @@ def _create_download(*, creator_id, image_id=None, submission_id=None):
...
+        challenge_request = ChallengeRequest.objects.get(
+            pk=challenge_request_pk
+        )
+    except ChallengeRequest.DoesNotExist:
+        raise Http404("Challenge request not found.")
+
+    if request.user.has_perm("view_challengerequest"):
+        return protected_storage_redirect(
+            name=challenge_request.structured_challenge_submission_form.name
+        )
```
jmsmkn, line 150: A functional view should never return None (implicitly done here). You need to raise `PermissionDenied` if the user cannot access the file.

jmsmkn: This method also needs tests.

jmsmkn (commented):
Looks good! Couple of minor things.

PR #2314, opened 2022-03-09
Add filesystem perparation task

jmsmkn:
When running EFS in bursting mode the number of burst credits can drop to zero, and the filesystem performance will drop by 1000x. To work around this we create a temporary credits file that only exists to build up the burst credits, so that the high performance mode is always available.


app/tests/components_tests/stubs.py
```diff
@@ -50,6 +51,20 @@ def _copy_to_output(self):
...
 
 
+class CloudwatchClientStub:
+    def get_metric_statistics(self, **__):
+        return {
+            "Label": "BurstCreditBalance",
+            "Datapoints": [
+                {
+                    "Timestamp": "2022-03-09T14:38:00+00:00",
+                    "Maximum": 2199000000000,  # Be careful changing this value, could create a lot of data
```
miriam-groeneveld, line 61: Maybe use the more readable notation?
```suggestion
                    "Maximum": 2_199_000_000_000,  # Be careful changing this value, could create a lot of data
```

miriam-groeneveld (approved):
Looks good

app/tests/components_tests/stubs.py
jmsmkn: Agreed, also made it a variable as if someone changed the target you may end up with a huge file generated on disk in the tests 😱

PR #2315, opened 2022-03-11
Hanging Protocol App

amickan:
- Adds a simple hanging protocol app as well as create, update, detail and list view 
- Adds hanging protocol as foreign key to algorithms, archives, reader studies

Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/119


app/grandchallenge/hanging_protocols/forms.py
```diff
@@ -0,0 +1,18 @@
...
+
+from grandchallenge.core.widgets import JSONEditorWidget
+from grandchallenge.hanging_protocols.models import HangingProtocol
+
+
+class HangingProtocolForm(forms.ModelForm):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.helper = FormHelper(self)
+        self.helper.layout.append(Submit("save", "Save"))
```
jmsmkn, line 13: There is a mixin for this already, `SaveFormInitMixin` or something like that.

app/grandchallenge/hanging_protocols/urls.py
```diff
@@ -0,0 +1,19 @@
...
+    HangingProtocolList,
+    HangingProtocolUpdate,
+)
+
+app_name = "hanging-protocols"
+
+urlpatterns = [
+    path("", HangingProtocolList.as_view(), name="list"),
+    path("create/", HangingProtocolCreate.as_view(), name="create"),
+    path("<slug:slug>/", HangingProtocolDetail.as_view(), name="detail"),
```
jmsmkn, line 15: The `:slug` is not necessary.

app/grandchallenge/hanging_protocols/views.py
```diff
@@ -0,0 +1,50 @@
...
+from grandchallenge.subdomains.utils import reverse
+
+
+class HangingProtocolList(LoginRequiredMixin, ListView):
+    model = HangingProtocol
+    paginate_by = 50
+
+
+class HangingProtocolCreate(
+    LoginRequiredMixin, SuccessMessageMixin, CreateView
```
jmsmkn, line 19: I think we should limit who can create protocols like we do the workstation configs (etc), can use the global `create_hangingprotocol` permission.

jmsmkn (commented):
Looks great! Only thing missing is limiting who can create them IMO.

app/grandchallenge/hanging_protocols/models.py
```diff
@@ -0,0 +1,102 @@
...
+from django.db import models
+from guardian.shortcuts import assign_perm
+
+from grandchallenge.core.models import TitleSlugDescriptionModel, UUIDModel
+from grandchallenge.core.validators import JSONValidator
+
+
+class ImagePort(models.TextChoices):
+    MAIN = "M", "Main"
+    SECONDARY = "S", "Secondary"
```
HarmvZ, line 11: Do you know why these first 2 have a single character choice while the others are fully typed out? I see you moved this from the ReaderStudy model to here and it was already like this, but I'm just curious why it's like this.

app/grandchallenge/hanging_protocols/models.py
```diff
@@ -0,0 +1,102 @@
...
+    "contains": {
+        "type": "object",
+        "properties": {
+            "viewport_name": {
+                "type": "string",
+                "enum": [port.lower() for port in ImagePort.labels],
+                "pattern": "^main$",
+            },
+        },
+    },
```
HarmvZ, line 36: What does this do exactly? The `items` part below already validates the schema correctly so this does not add any extra functionality, right? Also, how does the combination of `pattern` and `enum` fields work?

HarmvZ (commented):
Nice work! I have two questions.

app/grandchallenge/hanging_protocols/models.py
amickan: This validates that at least one of the items in the array has `main` as the `viewport_name`.  The items part below only validates that the `viewport_name`  is one of the strings specifed in the `enum` but not that the `main` viewport is required. Now that I look at it, the `enum` might not be necessary in the `contains` part here, because `pattern` is already specific enough.

HarmvZ: Ah right, makes sense!

HarmvZ (approved):

HarmvZ (approved):

PR #2317, opened 2022-03-14
Add display set to answer serializer

MikeOverkamp-diag:


app/grandchallenge/reader_studies/serializers.py
```diff
@@ -157,30 +168,43 @@ def validate(self, attrs):
...
         if self.instance:
-            on_commit(
-                lambda: add_scores.apply_async(
-                    kwargs={
-                        "instance_pk": str(self.instance.pk),
-                        "pk_set": list(
-                            map(str, images.values_list("pk", flat=True))
-                        ),
-                    }
+            if images is not None and images.count() > 0:
```
jmsmkn, line 188: Shouldn't we validate that only `images` or `display_set` are set?

app/grandchallenge/reader_studies/tasks.py
```diff
@@ -46,6 +46,27 @@ def add_scores(*, instance_pk, pk_set):
             add_score(instance, ground_truth.answer)
 
 
+@shared_task
```
jmsmkn, line 49: Please do not use the default queue. Please ensure that any new tasks are idempotent and use one of the acks-late queues (depending on their memory/duration requirements).

jmsmkn (commented):
Looks good! Some minor comments.

app/grandchallenge/reader_studies/tasks.py
MikeOverkamp-diag: I think the micro-short queue should be sufficient for this

jmsmkn (approved):

PR #2322, opened 2022-03-17
Add image port mapping field

MikeOverkamp-diag:


amickan (approved):
Looks great!

jmsmkn:
Definitely should only be a warning. Maybe it should be a in the viewer? I imagine that we will one day have the ability to dynamically switch out hanging protocols. I guess you could add the hanging protocol info in the admin section of the algorithm detail page, and show a warning there.

MikeOverkamp-diag:
Do we want the message to always be there on a mismatch? It is fairly easy to add a warning post save in the admin, but I'd say that is of limited use. If we want it to be there permanent, I think that's a bit more work and I'd propose to move that to a separate PR.

MikeOverkamp-diag:
Renamed the field and updated the help text. Accidentally rebased in the process, hence the force push. Apologies.

jmsmkn (approved):

PR #2327, opened 2022-03-21
Hanging protocol serializer field

amickan:
Adds `hanging_protocol` json to algorithm, archive and reader study serializers.

Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/119


HarmvZ (dismissed):

app/tests/archives_tests/test_serializers.py
```diff
@@ -26,13 +27,16 @@
...
+                    "hanging_protocol",
                 ),
             },
         )
     ),
 )
 class TestSerializers:
     def test_serializer_valid(self, serializer_data, rf):
+        hp = HangingProtocolFactory(json=([{"viewport_name": "main"}]))
+        serializer_data["factory"] = ArchiveFactory(hanging_protocol=hp)
```
HarmvZ, line 39: To make this work correctly I think you need remove these lines and use a subclassed version of ArchiveFactory on line 19 where the hanging protocol field is defined.

jmsmkn (approved):

PR #2332, opened 2022-03-22
Fix verification email

MikeOverkamp-diag:


MikeOverkamp-diag:
I'm wondering if we want to allow the email field on the Verification model to blank? I feel it should not be blank, which is why I added the email in the fixtures. However, the field has `blank=True` set, which is why I also included the try.. except in the view function.

jmsmkn:
I think I agree, AFAIK doesn't need to be blank any more, was left over from the initial implementation.

MikeOverkamp-diag:
Removed the blank=True. It's probably still safest to keep the try.. except in place in case of any residual empty email fields.

app/grandchallenge/groups/views.py
```diff
@@ -77,6 +77,10 @@ def get_result_label(self, result):
             is_verified = False
 
         if is_verified:
+            try:
```
jmsmkn, line 80: Are the changes in this file required now?

jmsmkn:
> Removed the blank=True. It's probably still safest to keep the try.. except in place in case of any residual empty email fields.

I've checked and it's safe.

jmsmkn (approved):

PR #2336, opened 2022-03-23
Allow emptying of field value for display sets

MikeOverkamp-diag:


app/tests/reader_studies_tests/test_api.py
```diff
@@ -1676,6 +1676,24 @@ def test_display_set_add_and_edit(client, settings):
     assert ds.values.count() == 2
     assert ds2.values.count() == 2
 
+    # Remove the civ fromthe new display set again
```
miriam-groeneveld, line 1679:
```suggestion
    # Remove the civ from the new display set again
```

miriam-groeneveld (dismissed):

miriam-groeneveld (approved):

PR #2340, opened 2022-03-23
Add view_content field to serializers

amickan:
Adds `view_content` field to the algorithm, archive and rs serializers. Also adds some additional help texts.


miriam-groeneveld (approved):
Looks good, I think the view_content was already added to the serializers earlier today.
But nice to have the links in the help texts, thanks for that!

PR #2346, opened 2022-03-24
Allow copying of archive items to display sets

MikeOverkamp-diag:
Note: I intentionally did not include a link to the new view anywhere yet, as it is only intended for beta testing. I will manually supply the correct url to the testers.

### DIFF
--- a/app/tests/archives_tests/test_views.py
+++ b/app/tests/archives_tests/test_views.py
@@ -467,8 +467,9 @@ def test_archive_items_to_reader_study_update(client, settings):
     rs1 = ReaderStudyFactory(use_display_sets=True)
     rs2 = ReaderStudyFactory(use_display_sets=False)
 
-    editor = UserFactory()
-    archive.editors_group.user_set.add(editor)
+    editor, user = UserFactory(), UserFactory()
+    archive.add_user(user)
+    archive.add_editor(editor)
     rs1.add_editor(editor)
     rs2.add_editor(editor)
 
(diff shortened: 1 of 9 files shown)


jmsmkn (commented):
Looks good but could you add a test for the reader studies that are selectable and for the overall view?

jmsmkn:
Maybe I'm missing something I don't see how that test tests for reader study selection and the overall view. Maybe I wasn't clear. I want to see a test to ensure that:

```
        self.fields["reader_study"].queryset = (
            get_objects_for_user(
                self.user,
                "reader_studies.change_readerstudy",
                accept_global_perms=False,
            )
            .filter(use_display_sets=True)
            .order_by("title")
        )
```

and 

```
    permission_required = (
        f"{Archive._meta.app_label}.use_{Archive._meta.model_name}"
    )
```

are implemented.

app/tests/archives_tests/test_views.py
```diff
@@ -456,3 +457,67 @@ def test_api_archive_item_add_and_update_non_image_file(client, settings):
...
+        viewname="archives:items-reader-study-update",
+        client=client,
+        reverse_kwargs={"slug": archive.slug},
+        follow=True,
+        user=editor,
+    )
+
+    assert response.status_code == 200
+    assert str(rs1.pk) in response.rendered_content
+    assert str(rs2.pk) not in response.rendered_content
```
MikeOverkamp-diag, line 502: The queryset test is included by these two asserts. I'll add a test for someone without edit rights.

MikeOverkamp-diag:
Now that I'm working on the test, I'm unsure if the permission I used is the one we want. Do we want all archive users to be able to add items to a reader study? I know that's the current implementation for images as well, but if we want to chance this, I'd say this is the time.

MikeOverkamp-diag:
Nevermind my previous comment, I don't see why we wouldn't want user to be able to do this. 

I think I covered all possible scenarios in the tests now.

jmsmkn:
> Nevermind my previous comment, I don't see why we wouldn't want user to be able to do this.
> 
> I think I covered all possible scenarios in the tests now.

Yes I think it's right, otherwise you end up with users downloading the images from the archive and re-uploading them to the study.

jmsmkn (approved):

PR #2349, opened 2022-03-25
Add user field to group admin

amickan:


app/grandchallenge/groups/admin.py
```diff
@@ -0,0 +1,47 @@
...
+User = get_user_model()
+
+
+class GroupAdminForm(forms.ModelForm):
+    class Meta:
+        model = Group
+        exclude = []
+
+    users = forms.ModelMultipleChoiceField(
+        queryset=User.objects.all(),
```
jmsmkn, line 16: You cannot use this in production, querying all of the users will OOM the process. You need to use something that does things with filtering. This is what we use elsewhere, unsure if it can be used in the admin https://django-select2.readthedocs.io/en/latest/.

amickan: The correct widget would be `AutocompleteSelectMultiple` (this is also what Django sets as the widget for all fields declared as `autocomplete_fields`), but the `get_queryset()` method of the autocomplete widget fails when calling `get_limit_choices_to` on the `user` field, which simply does not have this defined. I spent all day trying to get around this, with no success. As a suboptimal alternative, I could instead render the user field as a raw-id-field. Adding a user to a group then requires knowing their id, but since this is only necessary for one very specific case, maybe this is an ok work-around for now?

jmsmkn: Yes, the id is pretty easy for an admin to determine so that is a fine compromise.

amickan: Hm I was too quick, the raw-id solution has the same problem...

jmsmkn: Hmm, that's strange, are you sure that the queryset is being evaluated there?

jmsmkn: Using autocomplete fields is not an option? https://docs.djangoproject.com/en/dev/ref/contrib/admin/#django.contrib.admin.ModelAdmin.autocomplete_fields

amickan: No, for some reason autocomplete fields doesn't work. But all it does it set the widget for the field to `AutocompleteSelectMultiple`, so even if it worked, we would run into the same issue as when explicitly defining that widget for the field in the form, I think. I've also tried specifying the `Select2MultipleWidget` we use elsewhere, but that just renders a list not a select field with autocomplete in the admin for some reason. Interestingly, I did get it to work with the `ModelSelect2Multiple` widget from `dal_select2`, but that field gets rendered white with white text and I haven't yet managed to change the color... 
image

jmsmkn: I think that still has the problem that the queryset is fully evaluated to populate the list right? It's no problem to use all users as the queryset, so long as it is not evaluated internally.

amickan: ah yes you're right it does fully evaluate the queryset. Hm, I don't know anymore, all the options I've tried evaluate the queryset, not sure what else to try anymore :-/

jmsmkn: I think that I have found a solution, could you take a look and maybe fix the test?

jmsmkn: The pre commit stuff is fixed on master so that can be ignored/merged in.

jmsmkn (approved):

PR #2352, opened 2022-03-29
Allow each viewport only once in hanging protocol

MikeOverkamp-diag:


amickan (approved):

jmsmkn (commented):
General comment - please try to test forms directly rather than using an integration test with a view. It would be good to try and get our test time down as it's far too much right now (my fault).

PR #2356, opened 2022-03-29
Ensure JSONFields are not set to null

MikeOverkamp-diag:


MikeOverkamp-diag:
That's what I thought as well, but reading this changed my mind a bit: https://stackoverflow.com/a/38126751

jmsmkn:
I see that, but doesn't this get you into the same problem that you have for a nullable char field where `""` and `None` represent empty data, only here it's `[]` and `None`?

MikeOverkamp-diag:
Works like a charm. Also included the other two JSONFields so we don't run into the same issue with them.

jmsmkn (approved):

PR #2357, opened 2022-03-29
Add data license and long term commitment fields to Challenge request form

amickan:
- This adds two checkboxes, one asking if the organizers agree to make their public training data avaiable under a CC-BY license and the other asking if they are willing to suport the challenge for the next 5 years. If these boxes are not checked, they have to explain why they do not agree to these terms.
- This also adds the budget information that the requester entered to the detail view (on top of the actual budget calculation).


jmsmkn (commented):
This looks good, but this is a perfect use case for using HTMX (see https://htmx.org/ and https://github.com/adamchainz/django-htmx). My feeling is that this would simplify this form a lot,  and it would be really interesting to use this elsewhere in the site. Maybe if you have some downtime you can investigate this?

jmsmkn (approved):

amickan:
> This looks good, but this is a perfect use case for using HTMX (see https://htmx.org/ and https://github.com/adamchainz/django-htmx). My feeling is that this would simplify this form a lot, and it would be really interesting to use this elsewhere in the site. Maybe if you have some downtime you can investigate this?

Yes, good idea, I'll have a look soon!

PR #2359, opened 2022-03-30
Speed up display set list view

MikeOverkamp-diag:


jmsmkn (approved):
This still does one db call per interface which will have the same issue (look at you, MPMRI), you could probably get all the information in one call and then transform but up to you.

jmsmkn (approved):

PR #2361, opened 2022-03-30
Add migration to move reader studies to new setup

MikeOverkamp-diag:
We probably should not merge this until Monday, as it adds a warning to any reader study that does not have `use_display_sets` set to True.

### DIFF
--- a/app/grandchallenge/reader_studies/management/commands/migrate_to_display_sets.py
+++ b/app/grandchallenge/reader_studies/management/commands/migrate_to_display_sets.py
@@ -1,0 +1,69 @@
+from django.core.management import BaseCommand
+
+from grandchallenge.cases.models import Image
+from grandchallenge.components.models import (
+    ComponentInterface,
+    ComponentInterfaceValue,
+)
+from grandchallenge.reader_studies.models import (
+    Answer,
+    DisplaySet,
+    ReaderStudy,
+)
+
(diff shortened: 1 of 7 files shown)


jmsmkn (commented):
Looks good, but the migration of each study should be in an atomic block

jmsmkn:
Also do we need to prevent new answers being created for a study whilst it is being migrated?

jmsmkn:
Can you add something to detect this at the end of the migration for each RS?

jmsmkn:
Also please add logging as we go so we have insights on the progress.

MikeOverkamp-diag:
> Can you add something to detect this at the end of the migration for each RS?

Done, but I'm not sure how to add that bit to the test.

MikeOverkamp-diag:
I'm considering not putting this in a management command, but making it a button in the view, like the `Generate hanging list` button. The way I picture this, is that we have users define the view content first, that way we know what interface to use for what image. By going this route, we are not limited to just main and main-overlay, but can use any interface and as many viewports as needed. What do you think @jmsmkn?

jmsmkn:
I think we should still keep this so that we handle it for the simple cases, but indeed other studies will need a path forward and that sounds like a fine idea.

jmsmkn (approved):

PR #2365, opened 2022-03-31
Add display set creation task to micro short queue

MikeOverkamp-diag:
Having it in the 2xl queue might cause a long delay between uploading cases and actually seeing them in the reader study, this might confuse users. I think the task is fast and light enough to justify adding it to the micro short queue.

### DIFF
--- a/app/grandchallenge/reader_studies/tasks.py
+++ b/app/grandchallenge/reader_studies/tasks.py
@@ -84,7 +84,7 @@ def add_image_to_display_set(
         display_set.values.add(civ)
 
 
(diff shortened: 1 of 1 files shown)


jmsmkn:
I'm not so sure, the limit is 10 minutes for the short queue. Creating a new image is quite slow due to the permissions stuff, and someone could have a session with a few hundred images, right? This would then never get added. The short queue can also fill up as this does all of the calls to the ECS API too, we actually scale out the large queue to have more workers, so I don't think that this will make much of a difference. I'm going to close this as I don't think that it will solve the issue, and adds in this extra constraint.

PR #2366, opened 2022-04-01
Limit display list page size to 10 for performance

MikeOverkamp-diag:


jmsmkn:
This seems like a bit of a duct tape solution. How many queries are being done per row?

MikeOverkamp-diag:
You are right and it does not seem to make the load times much better. 

After updating a display set when there are 7 display sets:

TimerPanel_utime;dur=33380.032;desc="User CPU time", TimerPanel_stime;dur=3192.9039999999986;desc="System CPU time", TimerPanel_total;dur=36572.936;desc="Total CPU time", TimerPanel_total_time;dur=35165.398836135864;desc="Elapsed time", SQLPanel_sql_time;dur=39.029598236083984;desc="SQL 29 queries", CachePanel_total_time;dur=11.178255081176758;desc="Cache 30 Calls"

After updating a display set when there are 8 display sets:

TimerPanel_utime;dur=31422.95899999999;desc="User CPU time", TimerPanel_stime;dur=3224.7529999999997;desc="System CPU time", TimerPanel_total;dur=34647.71199999999;desc="Total CPU time", TimerPanel_total_time;dur=33304.01921272278;desc="Elapsed time", SQLPanel_sql_time;dur=57.84440040588379;desc="SQL 32 queries", CachePanel_total_time;dur=16.022920608520508;desc="Cache 34 Calls"

So the difference is only 3 queries after updating a value.

jmsmkn:
Also is there an N+1 problem with cache calls too - also increases from 30 to 34 calls.

app/grandchallenge/datatables/views.py
```diff
@@ -44,7 +45,7 @@ def get(self, request, *args, **kwargs):
         response = super().get(request, *args, **kwargs)
         if request.META.get("HTTP_X_REQUESTED_WITH") == "XMLHttpRequest":
             start = int(request.GET.get("start", 0))
-            page_size = int(request.GET.get("length"))
+            page_size = self.page_size or int(request.GET.get("length"))
```
jmsmkn, line 48: Does this really work with Datatables JS? If it is requesting 50 objects and you only send back 10, is the pagination correct (eg, total number of pages and objects that appear on subsequent pages)? I would imagine that you would set the pagination options client side, and then set a maximum server side, raising a 400 bad request error if the clients selected pagination is over the server side limit.

MikeOverkamp-diag:
I think you are right and we need to find a better solution.

PR #2367, opened 2022-04-01
Hanging protocol visualization

amickan:
This adds a visualization of the hanging protocol to the create and update form as well as to the detail view and the list view. 

image

image

https://user-images.githubusercontent.com/30069334/161499758-123ce6f8-015a-40a3-993c-ea22cb7bc464.mp4

### DIFF
--- a/app/grandchallenge/hanging_protocols/static/js/hanging_protocols/hanging_protocol_visualization.js
+++ b/app/grandchallenge/hanging_protocols/static/js/hanging_protocols/hanging_protocol_visualization.js
@@ -46,9 +46,13 @@ function createViewportDiv(divId, viewportNum, viewportSpec, totalHeight, totalW
     viewportDiv.style.top = isNaN(viewportSpec.y) ? "0%" : (viewportSpec.y / parseFloat(totalHeight).toFixed(2))*100 + '%';
     document.getElementById(divId).appendChild(viewportDiv).classList.add('bg-dark', 'rounded', 'border', 'border-2', 'd-flex', 'flex-column', 'justify-content-center', 'align-items-center');
 
-    viewportDiv.innerHTML += '<p style="font-size: 1.5em">'+ viewportSpec.viewport_name + '</p>';
-    if (viewportSpec.fullsizable == true) {
-        viewportDiv.innerHTML += '<i class="fas fa-expand m-1"></i>'
+    if (/\d/.test(divId)) {
+        viewportDiv.innerHTML += '<p class="mb-0" style="font-size: 0.9em">'+ viewportSpec.viewport_name + '</p>';
+    } else {
+        viewportDiv.innerHTML += '<p style="font-size: 1.5em">' + viewportSpec.viewport_name + '</p>';
+        if (viewportSpec.fullsizable == true) {
+            viewportDiv.innerHTML += '<i class="fas fa-expand m-1"></i>'
+        }
     }
 }
 
@@ -59,7 +63,6 @@ function updateHangingProtocolVisualization(parentDivId, jsonString){
     try {
         let jsonSpec = JSON.parse(jsonString);
         [totalHeight, totalWidth] = getGridDimensions(jsonSpec);
-        console.log(totalWidth, totalHeight)
         removeAllChildNodes(document.getElementById(parentDivId));
         for (let i = 0; i < jsonSpec.length; i++) {
             createViewportDiv(parentDivId, i, jsonSpec[i], totalHeight, totalWidth);
@@ -74,7 +77,6 @@ $(document).ready(function () {
         showOrHideVisualizationDiv("hpVisualization", jsonString);
         updateHangingProtocolVisualization("hpVisualization", jsonString);
         document.getElementById('jsoneditor_id_json').addEventListener('input', function(){updateHangingProtocolVisualization()});
-        document.getElementById('jsoneditor_id_json').addEventListener('keyup', function(){updateHangingProtocolVisualization()});
         document.getElementById('jsoneditor_id_json').addEventListener('paste', function(){updateHangingProtocolVisualization()});
     } else {
         let jsonDivs = document.querySelectorAll("[id^='id_json']");
--- a/app/grandchallenge/hanging_protocols/static/js/hanging_protocols/hanging_protocol_visualization.js
+++ b/app/grandchallenge/hanging_protocols/static/js/hanging_protocols/hanging_protocol_visualization.js
@@ -8,13 +8,11 @@ function removeAllChildNodes(parent) {
     }
 }
 
-function showOrHideVisualizationDiv(jsonString) {
-    console.log(jsonString)
+function showOrHideVisualizationDiv(divId, jsonString) {
     if (jsonString != '' && jsonString != 'null') {
-            console.log("show!")
-            document.getElementById("hpVisualization").style.display = 'block';
+            document.getElementById(divId).style.display = 'block';
         } else {
-            document.getElementById("hpVisualization").style.display = 'none';
+            document.getElementById(divId).style.display = 'none';
     }
 }
 
@@ -37,7 +35,7 @@ function getGridDimensions(json) {
     return [totalHeight, totalWidth]
 }
 
-function createViewportDiv(viewportSpec, totalHeight, totalWidth) {
+function createViewportDiv(divId, viewportSpec, totalHeight, totalWidth) {
     let viewportDiv = document.createElement("div");
     viewportDiv.setAttribute('id', viewportSpec.viewport_name);
     viewportDiv.style.opacity = '0.5';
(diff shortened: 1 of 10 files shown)


app/grandchallenge/hanging_protocols/static/js/hanging_protocols/hanging_protocol_visualization.js
```diff
@@ -0,0 +1,90 @@
...
+    let allW = [];
+    let allH = [];
+
+    for (let i = 0; i < json.length; i++) {
+        json[i].x == undefined ? undefined : allX.push(json[i].x);
+        json[i].y == undefined ? undefined : allY.push(json[i].y);
+        json[i].w == undefined ? undefined : allW.push(json[i].w);
+        json[i].h == undefined ? undefined : allH.push(json[i].h);
+    }
+    let totalWidth = (allX.length != json.length | allW.length != json.length) ? json.length : Math.max(...allX.map(function(num, idx) {return num+allW[idx]}));
```
HarmvZ, line 31: The `|` in the first part of the ternary operator here is a bitwise OR. It think you meant to use the logical OR (=`||` in JS `or` in Python). Functionally it will not make a difference in this case since you apply it on booleans (except for very minor performance increase due to short circuiting) but you should always use the logical OR in cases like these.


Also, the logic on these lines will fail silently if there are multiple viewports defined that only contain some of the `x`, `y`, `w` or `h` fields. I think it is probably best to add an extra requirement to the hanging protocol that either all of these fields for all viewports are filled in, or none of them. That way it will be a lot easier to calculate max heigh/width and it will be a lot clearer also for the users. Then this whole function can also be simplified.

app/grandchallenge/hanging_protocols/static/js/hanging_protocols/hanging_protocol_visualization.js
```diff
@@ -0,0 +1,90 @@
...
+
+    return [totalHeight, totalWidth]
+}
+
+function createViewportDiv(divId, viewportNum, viewportSpec, totalHeight, totalWidth) {
+    let viewportDiv = document.createElement("div");
+    viewportDiv.setAttribute('id', viewportSpec.viewport_name);
+    viewportDiv.style.opacity = '0.5';
+    viewportDiv.style.position = 'absolute';
+    viewportDiv.style.zIndex = '-' + viewportSpec.order;
```
HarmvZ, line 42: Nice!

app/grandchallenge/hanging_protocols/static/js/hanging_protocols/hanging_protocol_visualization.js
```diff
@@ -0,0 +1,90 @@
...
+    viewportDiv.style.width = isNaN(viewportSpec.w) ? (1 / parseFloat(totalWidth).toFixed(2))*100 + '%' : (viewportSpec.w / parseFloat(totalWidth).toFixed(2))*100 + '%';
```
HarmvZ, line 43: The `toFixed()` turns a Number into a string with a certain number of decimals. And javascript tries to change the type of a certain variable so that the operation applied to it makes sense (type coercion). So what will happen when the interpreter comes across this statement (notice the quotes to indicate that it is a string):
```
> const totalWidth = "123.456"
undefined
> parseFloat(totalWidth)
123.456
> parseFloat(totalWidth).toFixed(2)
'123.46'
> 1 / parseFloat(totalWidth).toFixed(2)
0.008099789405475458
> (1 / parseFloat(totalWidth).toFixed(2)) * 100
0.8099789405475458
> (1 / parseFloat(totalWidth).toFixed(2)) * 100 + '%'
'0.8099789405475458%'
```

So that is probably not quite what you wanted to do. I think you wanted to do something like this:
```
(1 / parseFloat(totalWidth) * 100).toFixed(2) + '%'
```
And even here the `parseFloat` is not strictly necessary because the interpreter will already try to convert the string `totalWidth` into a Number when you try to apply the divide operation on it, but it is clearer to add it.

app/grandchallenge/hanging_protocols/static/js/hanging_protocols/hanging_protocol_visualization.js
```diff
@@ -0,0 +1,90 @@
...
+            createViewportDiv(parentDivId, i, jsonSpec[i], totalHeight, totalWidth);
+        }
+    } catch (err) {
+    }
+}
+
+$(document).ready(function () {
+    if (window.location.href.includes("update") | window.location.href.includes("create")) {
+        let jsonString = document.getElementById("id_json").value;
+        showOrHideVisualizationDiv("hpVisualization", jsonString);
```
HarmvZ, line 77: This line is not necessary since it is already called in the `updateHangingProtocolVisualization` function below, right?

app/grandchallenge/hanging_protocols/static/js/hanging_protocols/hanging_protocol_visualization.js
```diff
@@ -0,0 +1,90 @@
...
+    }
+    let totalWidth = (allX.length != json.length | allW.length != json.length) ? json.length : Math.max(...allX.map(function(num, idx) {return num+allW[idx]}));
+    let totalHeight = (allY.length != json.length | allH.length != json.length) ? 1 : Math.max(...allY.map(function(num, idx) {return num+allH[idx]}));
+
+    return [totalHeight, totalWidth]
+}
+
+function createViewportDiv(divId, viewportNum, viewportSpec, totalHeight, totalWidth) {
+    let viewportDiv = document.createElement("div");
+    viewportDiv.setAttribute('id', viewportSpec.viewport_name);
```
HarmvZ, line 39: If `viewportSpec.viewport_name` is not (yet) defined, this will create a div with `id="undefined"`. So if there are multiple empty objects there will be multiple divs with `id="undefined"` but id should always be unique. Also it will show undefined as the name of the div. I think it would be nicer if either the div is not created when the `viewport_name`  field is not (yet) defined, or that it will show no title and for example a red background or something to indicate that it is an invalid viewport.

app/grandchallenge/hanging_protocols/static/js/hanging_protocols/hanging_protocol_visualization.js
```diff
@@ -0,0 +1,90 @@
...
+    viewportDiv.style.opacity = '0.5';
+    viewportDiv.style.position = 'absolute';
+    viewportDiv.style.zIndex = '-' + viewportSpec.order;
+    viewportDiv.style.width = isNaN(viewportSpec.w) ? (1 / parseFloat(totalWidth).toFixed(2))*100 + '%' : (viewportSpec.w / parseFloat(totalWidth).toFixed(2))*100 + '%';
+    viewportDiv.style.height = isNaN(viewportSpec.h) ? "100%" : (viewportSpec.h / parseFloat(totalHeight).toFixed(2))*100 + '%';
+    viewportDiv.style.left = isNaN(viewportSpec.x) ? (viewportNum / parseFloat(totalWidth).toFixed(2))*100 + '%' : (viewportSpec.x / parseFloat(totalWidth).toFixed(2))*100 + '%';
+    viewportDiv.style.top = isNaN(viewportSpec.y) ? "0%" : (viewportSpec.y / parseFloat(totalHeight).toFixed(2))*100 + '%';
+    document.getElementById(divId).appendChild(viewportDiv).classList.add('bg-dark', 'rounded', 'border', 'border-2', 'd-flex', 'flex-column', 'justify-content-center', 'align-items-center');
+
+    if (/\d/.test(divId)) {
```
HarmvZ, line 49: If I understand correctly, this test is to see if the parent div has a digit in it so if you are on the list view, right? I think this should be done in a more robust way. It would be even better if this is not needed at all. I think you can do that by making the font size relative and always show the icon for fullsizable.

app/grandchallenge/hanging_protocols/static/js/hanging_protocols/hanging_protocol_visualization.js
```diff
@@ -0,0 +1,90 @@
...
+        removeAllChildNodes(document.getElementById(parentDivId));
+        for (let i = 0; i < jsonSpec.length; i++) {
+            createViewportDiv(parentDivId, i, jsonSpec[i], totalHeight, totalWidth);
+        }
+    } catch (err) {
+    }
+}
+
+$(document).ready(function () {
+    if (window.location.href.includes("update") | window.location.href.includes("create")) {
```
HarmvZ, line 75: This works but there should be a better way to check if the user is on a page where they are editing the hp form. I think the best way is to remove it from this file and add it to a separate js file that you only include in the form.html page.

HarmvZ (changes requested):
Nice work on this! I like the minimal styling of it and that you also added it to the detail and list views. I have a few comments about the js stuff but other than that it looks great!

app/grandchallenge/hanging_protocols/static/js/hanging_protocols/hanging_protocol_visualization.js
amickan: Good point. I went with a mix of the two. The div only gets created when the `viewport_name` is not undefined, and if the provided `viewport_name` is not one of the possible labels, it gets a red background and a warning that a valid viewport name needs to be selected.

app/grandchallenge/hanging_protocols/static/js/hanging_protocols/hanging_protocol_visualization.js
```diff
@@ -0,0 +1,80 @@
...
+    viewportDiv.style.height = isNaN(viewportSpec.h) ? "100%" : (viewportSpec.h / parseFloat(totalHeight) * 100).toFixed(2) + '%';
+    viewportDiv.style.left = isNaN(viewportSpec.x) ? (viewportNum / parseFloat(totalWidth) * 100).toFixed(2) + '%' : (viewportSpec.x / parseFloat(totalWidth) * 100).toFixed(2) + '%';
+    viewportDiv.style.top = isNaN(viewportSpec.y) ? "0%" : (viewportSpec.y / parseFloat(totalHeight) * 100).toFixed(2) + '%';
+    document.getElementById(divId).appendChild(viewportDiv).classList.add('rounded', 'border', 'border-2', 'd-flex', 'flex-column', 'justify-content-center', 'align-items-center');
+
+    if (!possibleViewPorts.includes(viewportSpec.viewport_name)) {
+        viewportDiv.innerHTML += '<p class="mb-0" style="color: #fff">Invalid viewport name</p>';
+    } else {
+        viewportDiv.innerHTML += '<p class="mb-0">' + viewportSpec.viewport_name + '</p>';
+        if (viewportSpec.fullsizable == true) {
```
HarmvZ, line 60: I missed this on my first passthrough, but it is best to never use double equals in JS because that can lead to weird behavior. So instead of using `==` and `!=` you should always use `===` and `!==`. In this case it's just checking if the value is true so you can simplify to:

```suggestion
        if (viewportSpec.fullsizable) {

```

app/grandchallenge/hanging_protocols/static/js/hanging_protocols/hanging_protocol_visualization.js
```diff
@@ -0,0 +1,80 @@
...
+    }
+}
+
+function updateHangingProtocolVisualization(parentDivId, jsonString){
+    jsonString = jsonString || document.getElementById("id_json").value;
+    parentDivId = parentDivId || "hpVisualization";
+    showOrHideVisualizationDiv(parentDivId, jsonString);
+    try {
+        let jsonSpec = JSON.parse(jsonString);
+        let validJsonSpec = jsonSpec.filter(viewPort => viewPort.viewport_name != undefined);
```
HarmvZ, line 72: Never use `==`. In this case because you are checking if the value is undefined it gets slightly more complex. The way to do that properly is like this:

```suggestion
        let validJsonSpec = jsonSpec.filter(viewPort => typeof viewPort.viewport_name !== "undefined");

```

app/tests/hanging_protocols_tests/test_forms.py
```diff
@@ -102,3 +102,47 @@ def test_hanging_protocol_form(client):
...
+        follow=True,
+        user=user,
+    )
+
+    assert response.status_code == 200
+    assert HangingProtocol.objects.count() == 1
+    assert (
+        "Viewport definition must contain all or none of x, y, w, and h"
+        not in response.rendered_content
+    )
```
HarmvZ, line 148: Maybe add a test for the case where there are multiple viewports and one contains all of the dims and the other none resulting in an error.

HarmvZ (changes requested):
I have a few comments left. I missed some small things on the first review (`==` vs `===`) and I messed up the example I gave you for the form validation. Sorry 😅

HarmvZ (approved):

PR #2369, opened 2022-04-01
Reduce number of queries in ds list view

MikeOverkamp-diag:


app/grandchallenge/reader_studies/models.py
```diff
@@ -1117,13 +1117,12 @@ def value_list(self):
...
             cache.delete(key)
         if self.is_editable:
-            values = self.values.values_list(
-                "interface__slug", "id", "image_id"
-            )
             items = self.reader_study.values_for_interfaces
-            for slug, civ, image in values:
-                items[slug]["selected"] = civ
-                items[slug]["selected_image"] = image or ""
+            for civ in self.values.all():
```
MikeOverkamp-diag, line 1121: The fact that this is better than using the values_list was pretty surprising to me.

MikeOverkamp-diag: Obviously only combined with the `prefetch_related` on the initial queryset.

jmsmkn (approved):

app/grandchallenge/reader_studies/models.py
jmsmkn: Yes that is going to always create more requests and those are the killer, not the amount of data returned. If you would like to reduce the amount of data returned too then I think you can use `Prefetch(..., queryset=..., to_attr=...)`.

PR #2373, opened 2022-04-05
Add algorithm view_content and hanging_protocol to job serializer

MikeOverkamp-diag:


jmsmkn (dismissed):

app/grandchallenge/algorithms/serializers.py
```diff
@@ -78,6 +83,15 @@ class JobSerializer(serializers.ModelSerializer):
     algorithm_title = CharField(
         source="algorithm_image.algorithm.title", read_only=True
     )
+    algorithm_title = CharField(
+        source="algorithm_image.algorithm.title", read_only=True
+    )
```
HarmvZ, line 88: Accidental duplication of lines above?

jmsmkn: Good catch

PR #2374, opened 2022-04-05
Update case_text serializer

MikeOverkamp-diag:


jmsmkn (approved):
Looks good, this will need to be covered by tests before we remove the non-display set option. May be good to go through what is missing for this app on CodeCov and fill that in with (fast!) tests.

PR #2375, opened 2022-04-05
Consistent hanging protocol API

jmsmkn:
This PR moves hanging_protocol and view_content from the top level objects to the display set objects.


MikeOverkamp-diag (dismissed):

HarmvZ:
For the CIRRUS it would be easier if this is only implemented like this for archive item and algorithm job. For reader studies the reader study itself is alway accesible and parsed by CIRRUS so the hanging_protocol and view_content are loaded from there.

jmsmkn:
> For the CIRRUS it would be easier if this is only implemented like this for archive item and algorithm job. For reader studies the reader study itself is alway accesible and parsed by CIRRUS so the hanging_protocol and view_content are loaded from there.

But then we're back to an inconsistent API for everything.

HarmvZ:
Right but I think that inconsistency makes sense here. The hp and view_content will always be the same for all display sets of a reader study, but moving it to the display set requires reloading them each time a display set is loaded when going through a reader study. For algorithms and archives this is not the case because you start the workstation for each separate item or job.

chrisvanrun:
> > For the CIRRUS it would be easier if this is only implemented like this for archive item and algorithm job. For reader studies the reader study itself is alway accesible and parsed by CIRRUS so the hanging_protocol and view_content are loaded from there.
> 
> But then we're back to an inconsistent API for everything.

I agree with Harm. Moreover, this would complicate the current release effort. Display sets are queried in one large batch, adding a duplicated field like this would add to payload needlessly.

It would make more sense to implement this later in an hierarchical way: if present in display_set it overwrites the default provided by the reader_study.

jmsmkn:
The whole point of this cycle has been to get a consistent API between Jobs, Archive Items and Reader Studies (with the new display sets). We need to get the API right before the release as these things stick around. You do not need to worry about payload size. If it is so complicated for CIRRUS to load a protocol then just take the first one from the display set.

jmsmkn:
> Display sets are queried in one large batch, adding a duplicated field like this would add to payload needlessly.

This will anyway be changed in the next cycle, again, one of they key reasons for going to display sets so we can remove a lot of that logic from the viewer.

> It would make more sense to implement this later in an hierarchical way: if present in display_set it overwrites the default provided by the reader_study.

Doing it this way would mean that we would keep the logic in the backend, less to do in the viewer.

HarmvZ:
But the actual consistent API is still there because hanging_protocol and view_content are defined on the archive, algorithm and reader studies. So adding the fields to the serializer of display set archive item and algorithm jobs are just handy shortcuts for CIRRUS. On the display sets however this is not needed because the readerstudy endpoint itself is already accessed and the fields can be loaded from there. So I would keep the fields on archive, algorithm and reader study endpoints and also add them to the archive item and algorithm job endpoints as handy shortcuts for CIRRUS.

jmsmkn:
If we leave them there then CIRRUS will use them, and then it becomes a months long process to remove. We need to do this right first time.

jmsmkn:
If we leave them there then CIRRUS will use them, and then it becomes a months long process to remove. We need to do this right first time.

HarmvZ:
But it is okay to leave them there and let CIRRUS use them because they actually belong there, right? The fields are defined on those models and not on the item/job/displayset models. So they can always remain there and be used. Except they wont be used by CIRRUS for algorithms and archives because of the permissions issue and the fact that there is a shortcut to those fields via these new serializer fields. It's just that this is not necessary for display sets.

jmsmkn:
The viewer should not care where they live in the backend. That may change in the future. The point is consistency to the outside world. CIRRUS does things incorrectly now with loading the entire study, resulting in slow start up times. That will disappear in the next cycle.

I don't see how getting the hanging protocol from the first display set rather than the study is such a difficult thing. That is not a good enough argument to run with an inconsistent API for several months.

HarmvZ (approved):

HarmvZ:
Discussion was continued in a call. Conclusion: we implement it like it's done here so the API is consistent across all three models and we can support custom hanging protocols per display set in the future.

PR #2380, opened 2022-04-07
Use image name for case texts

### LINKED ISSUE #2378: Add command to migrate case text
With the new reader studies the case text format changed. We either need to update it for users or find a way to maintain the existing API.

MikeOverkamp-diag:
Closes #2378


jmsmkn:
Looking at this I'm wondering if migration is the best approach. I didn't realise that this API had changed. I think that the downside is that it will be very fragile as we allow the user to change the images in the display set, so this can easily get out of sync. I think a different approach is needed, either:

1. Fix this client side where CIRRUS matches the image name
2. Generate the new implementation server side. Much like we do with the hanging list where the image name is subbed in with the API URL, except now we do essentially what this PR does but dynamically. Care needs to be taken with requests but I think this is covered.

Maybe this feature needs a re-design, but for now we need to get it fixed. What do you thing @MikeOverkamp-diag and @chrisvanrun ?

MikeOverkamp-diag:
I think I actually implemented this with using image names instead of display set pks, but then decided that using display set pks made more sense. I don't think it's a lot of work to move back towards that, but it'll require a change in Cirrus. Or we can let the user add image names and then find the display set key in the code (I think the latter is maybe also what you meant?)

jmsmkn:
Yes, that is what I mean - change in CIRRUS (1) or doing the conversion in gc (2). If it can be done with minimal performance impact making the change to gc is quicker.

MikeOverkamp-diag:
I'll see if I can fix this in GC first, I think it should not be more expensive than the current implementation.

jmsmkn:
I wonder if we can also put it on the display set serializer so that CIRRUS could later pick it up from there next, then we remove the performance hit from the reader study list/detail views.

chrisvanrun:
I'm not a huge fan of using the indirect content of a CIV for mapping case text. It breaks the meaning of 'case' within the reader studies: being a set of images.

The main reason I discern for using an image-name, and correct me if I am wrong, is that users might start to rearrange single-image display-sets by swapping out the CIVs to CIs in the sets instead of manipulating the ordering.

Moreover, the new suggestion would break the following scenario where two different case texts are to be provided:

- Case 1: 5 images
- Case 2: same 5 + 1 new image

Maybe support both display-set UUIDs and image-names?

jmsmkn:
> I'm not a huge fan of using the indirect content of a CIV for mapping case text. It breaks the meaning of 'case' within the reader studies: being a set of images.

I think you're a little confused here, the case text is not part of a CIV, and I'm not suggesting that it should be.

> The main reason I discern for using an image-name, and correct me if I am wrong, is that users might start to rearrange single-image display-sets by swapping out the CIVs to CIs in the sets instead of manipulating the ordering.
> 
> Moreover, the new suggestion would break the following scenario where two different case texts are to be provided:
> 
>     * Case 1: 5 images
> 
>     * Case 2: same 5 + 1 new image

It wouldn't break if the new image does not have case text. If it does, then yeah, what do we do. However, the previous implementation (non-display sets with hanging lists) would also break in that scenario, it's never been reported, so I assume not an issue.

> Maybe support both display-set UUIDs and image-names?

Make a decision and stick to one. I say go for image names as the UUID is fragile.

MikeOverkamp-diag:
I have now changed the logic so that users can use image names as keys in the case text (as it was before we introduced display sets), we then look for any display set that contains a value with the key as an image name. I can extend that to also check for display set keys. That should be relatively simple, as long as users do not start combining image names an display set keys, as that could lead to values being overwritten in the resulting cleaned case text. We could solve that by just appending values if multiple matches are found.

I also missed the comment about adding it to the display set serializer, but that can be done easily. I assume we still want to keep it in the reader study serializer for now as well? Or will we make the necessary changes in CIRRUS first?

MikeOverkamp-diag:
> It wouldn't break if the new image does not have case text. If it does, then yeah, what do we do. However, the previous implementation (non-display sets with hanging lists) would also break in that scenario, it's never been reported, so I assume not an issue.

I think that for now it is probably best to stick with image names then, as that is what users know and what has worked up until now. If that turns out to be an issue in the case Chris mentioned, we can always re-evaluate.

jmsmkn:
> I also missed the comment about adding it to the display set serializer, but that can be done easily. I assume we still want to keep it in the reader study serializer for now as well? Or will we make the necessary changes in CIRRUS first?

I think that it's good to add there so that CIRRUS can start using it, then we can think about changing the implementation.

jmsmkn:
But yes keep it on the reader study too for backwards compatibility for the time being.

app/grandchallenge/reader_studies/views.py
```diff
@@ -866,7 +866,11 @@ def get_context_data(self, **kwargs):
 class ReaderStudyViewSet(ReadOnlyModelViewSet):
     serializer_class = ReaderStudySerializer
     queryset = ReaderStudy.objects.all().prefetch_related(
-        "images", "questions__options"
+        "images",
+        "questions__options",
+        "display_sets",
+        "display_sets__values",
+        "display_sets__values__image",
```
jmsmkn, line 873: You can just do `"display_sets__values__images"`, all intermediates are returned too.

MikeOverkamp-diag: Never knew that, thank you!

app/grandchallenge/reader_studies/serializers.py
```diff
@@ -73,6 +74,17 @@ class DisplaySetSerializer(HyperlinkedModelSerializer):
     view_content = JSONField(
         source="reader_study.view_content", read_only=True
     )
+    description = SerializerMethodField()
+
+    def get_description(self, obj):
```
jmsmkn, line 79: Could you move this to a `description` `@property` on the model? Then add it to each display set in the display set list view? Would make it easier for editors to track.

This doesn't introduce any N+1 problems?

jmsmkn: Also would make testing easier as you could then just do a model test, rather than going through the API.

MikeOverkamp-diag: Done, and now that I have actually added the prefetch_related to the view it does not.

jmsmkn (approved):

jmsmkn (commented):
I approved it but if you want to add it to the display set list template view and speed up the test (you can now test the object directly rather than going through the API and all the middleware 6x times) please go ahead.

PR #2381, opened 2022-04-08
Allow adding of multiple values to set in archive item update task

HarmvZ:
Fix for https://sentry.io/organizations/grand-challenge/issues/3173146090/


MikeOverkamp-diag (approved):

MikeOverkamp-diag:
You could also use .update on the set and feed it a list, I don't think that makes a difference in performance though.

PR #2382, opened 2022-04-08
Use proper view_content format in migration method

MikeOverkamp-diag:
This PR fixes the migration method to handle the view_content field properly. The initial function still used `*-overlay` keys in the view_content field, but those are no longer there. Instead a list of slugs is used for each viewport. Luckily this has no consequences for the migrations that have already been executed, as we did not use the reader study's view_content field there. However, it is was not usable when using the actual field, as the format of the value gets validated there. This PR fixes that.

### DIFF
--- a/app/grandchallenge/reader_studies/management/commands/migrate_to_display_sets.py
+++ b/app/grandchallenge/reader_studies/management/commands/migrate_to_display_sets.py
@@ -29,9 +29,9 @@ class Command(BaseCommand):
                 )
                 continue
 
-            view_content = {"main": "generic-medical-image"}
+            view_content = {"main": ["generic-medical-image"]}
             if "main-overlay" in content_keys:
-                view_content["main-overlay"] = "generic-overlay"
+                view_content["main"].append("generic-overlay")
 
             try:
                 migrate_reader_study_to_display_sets(rs, view_content)
--- a/app/grandchallenge/reader_studies/utils.py
+++ b/app/grandchallenge/reader_studies/utils.py
@@ -14,7 +14,9 @@ def _get_civ(image, slug):
     except ComponentInterfaceValue.DoesNotExist:
         raise ValueError(f"ComponentInterface {slug} does not exist.")
 
-    civ = ComponentInterfaceValue.objects.filter(image=image, interface=ci)
+    civ = ComponentInterfaceValue.objects.filter(
+        image=image, interface=ci
+    ).first()
     if civ is None:
         civ = ComponentInterfaceValue.objects.create(image=image, interface=ci)
     return civ
--- a/app/grandchallenge/reader_studies/utils.py
+++ b/app/grandchallenge/reader_studies/utils.py
@@ -29,6 +29,9 @@ def migrate_reader_study_to_display_sets(rs, view_content):  # noqa: C901
             ds = DisplaySet.objects.create(reader_study=rs)
             images = []
             for key in item:
+                # view_content does not contain *-overlay keys, the second
+                # value in the lit of slugs should be the overlay. This is
+                # handled in the else clause here.
                 if "overlay" in key:
                     continue
                 try:
--- a/app/grandchallenge/reader_studies/utils.py
+++ b/app/grandchallenge/reader_studies/utils.py
@@ -1,3 +1,5 @@
+import logging
+
 from django.db import transaction
 
 from grandchallenge.components.models import (
@@ -6,6 +8,20 @@ from grandchallenge.components.models import (
 )
 from grandchallenge.reader_studies.models import Answer, DisplaySet
 from grandchallenge.workstations.models import Workstation
+
+logger = logging.getLogger(__name__)
+
+
+def _get_civ(image, slug):
+    try:
+        ci = ComponentInterface.objects.get(slug=slug)
+    except ComponentInterfaceValue.DoesNotExist:
+        raise ValueError(f"ComponentInterface {slug} does not exist.")
+
+    civ, _ = ComponentInterfaceValue.objects.get_or_create(
+        image=image, interface=ci
+    )
+    return civ
 
 
 def migrate_reader_study_to_display_sets(rs, view_content):  # noqa: C901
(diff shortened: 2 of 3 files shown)


jmsmkn:
Several questions:

- What is the motivation behind this?
- In what situations does it occur?
- What are the consequences for the already migrated studies?
- What do we need to fix for the already migrated studies? 

Please add some context to your PRs so I can understand why changes are being proposed.

app/grandchallenge/reader_studies/utils.py
```diff
@@ -17,27 +33,34 @@ def migrate_reader_study_to_display_sets(rs, view_content):  # noqa: C901
...
                     raise ValueError(
-                        f"ComponentInterface {slug} does not exist."
+                        f"No ComponentInterface provided for {key}."
                     )
-
-                civ, _ = ComponentInterfaceValue.objects.get_or_create(
-                    image=image, interface=ci
-                )
-                ds.values.add(civ)
+                else:
```
jmsmkn, line 48: This section needs a test.

app/grandchallenge/reader_studies/utils.py
```diff
@@ -17,27 +33,34 @@ def migrate_reader_study_to_display_sets(rs, view_content):  # noqa: C901
...
+                    if len(slugs) > 2:
+                        logger.warning(
+                            f"More than two interface slugs provided for {key} "
+                            f"in {rs.slug}. Ignoring all values after the second."
+                        )
```
jmsmkn, line 53: Shouldn't this error out instead?

app/grandchallenge/reader_studies/utils.py
```diff
@@ -7,6 +9,20 @@
...
+logger = logging.getLogger(__name__)
+
+
+def _get_civ(image, slug):
+    try:
+        ci = ComponentInterface.objects.get(slug=slug)
+    except ComponentInterfaceValue.DoesNotExist:
+        raise ValueError(f"ComponentInterface {slug} does not exist.")
+
+    civ, _ = ComponentInterfaceValue.objects.get_or_create(
```
jmsmkn, line 21: This can be problematic in cases where there are multiple CIVs for the same interface and image, shouldn't happen but it is an issue in prod due to how things ran historically.

app/grandchallenge/reader_studies/management/commands/migrate_to_display_sets.py
```diff
@@ -29,9 +29,9 @@ def handle(self, *args, **options):
                 )
                 continue
 
-            view_content = {"main": "generic-medical-image"}
+            view_content = {"main": ["generic-medical-image"]}
             if "main-overlay" in content_keys:
-                view_content["main-overlay"] = "generic-overlay"
+                view_content["main"].append("generic-overlay")
```
jmsmkn, line 34: This view content never actually gets saved.

app/grandchallenge/reader_studies/utils.py
```diff
@@ -17,27 +33,34 @@ def migrate_reader_study_to_display_sets(rs, view_content):  # noqa: C901
             ds = DisplaySet.objects.create(reader_study=rs)
             images = []
             for key in item:
-                image = rs.images.get(name=item[key])
-                images.append(image.pk)
-
+                if "overlay" in key:
+                    continue
```
jmsmkn, line 37: Why skip here?

app/tests/reader_studies_tests/test_api.py
```diff
@@ -1842,8 +1839,8 @@ def test_migrate_to_display_sets(client, settings):
     rs.add_editor(editor)
     rs.add_reader(reader)
     rs.view_content = {
-        "main": "generic-medical-image",
-        "secondary": "generic-overlay",
+        "main": ["generic-medical-image"],
+        "secondary": ["generic-overlay"],
```
jmsmkn, line 1843: Shouldn't this be `{"main": ["generic-medial-image", "generic-overlay"]}`?

app/grandchallenge/reader_studies/management/commands/migrate_to_display_sets.py
MikeOverkamp-diag: It does not need to be saved, it's a variable that gets passed to the `migrate_reader_study_to_display_sets` method in utils.

app/grandchallenge/reader_studies/utils.py
MikeOverkamp-diag: So should we use a filter and create a new civ if one does not exist already?

app/grandchallenge/reader_studies/utils.py
MikeOverkamp-diag: Overlays get handled alongside the main image for the viewport. The `*-overlay` key is no longer used in the view_content

app/grandchallenge/reader_studies/utils.py
MikeOverkamp-diag: Everything but the ` if len(slugs) > 2:` is tested. Shall I add a test for that bit?

app/grandchallenge/reader_studies/utils.py
MikeOverkamp-diag: I'm not sure. I think it should be ok to provide more values in the view content, but we cannot assign anything to those interfaces, as with the old hanging list there was only the option to provide a main image and an overlay per viewport.

app/tests/reader_studies_tests/test_api.py
MikeOverkamp-diag: That's handled in the block before this. This is to test multiple viewports and just using the interfaces that are available. I'd be happy to create a new interface for this case though.

MikeOverkamp-diag:
Added a description that addresses your questions.

app/grandchallenge/reader_studies/utils.py
jmsmkn: Yes, filter and take the first, if it doesn't exist create it.

app/grandchallenge/reader_studies/utils.py
jmsmkn: These methods should be ultra conservative as in the worst case scenario it's either data loss or wrong data. If you're unsure, error out.

app/grandchallenge/reader_studies/management/commands/migrate_to_display_sets.py
jmsmkn: Does CIRRUS handle this correctly then when it is unset on the reader study?

MikeOverkamp-diag: Should be ok. From the [documentation](https://grand-challenge.org/documentation/viewer-layout/):

> No view content, no hanging protocol. In this case, the viewer will use a default hanging protocol with a single view. The first image will be shown, optionally with an overlay if one is present. 

It might still be a good idea to set it explicitly. I can write a management command for that and create a separate pr.

jmsmkn: OK - we should check that CIRRUS shows an error if an image cannot be displayed in this configuration.

jmsmkn (approved):

PR #2387, opened 2022-04-12
Update evaluation task routing

jmsmkn:
Moves calculate ranks to large memory queue and moves some atomic transaction context managers to decorators.


amickan (approved):
LGTM! Just out of curiosity: what is the benefit of using the decorator vs the context manager in this case?

jmsmkn:
Not much, just saves an indent!

PR #2390, opened 2022-04-12
Pull container images from the registry

jmsmkn:
Pull container images from the registry rather than using the file storage. Testing of this is removed as it is not possible to use insecure registries in GitHub actions.


amickan (approved):
Looks good. Is the reason for this change that pulling the image from the registry is faster and doesn't require the overhead of keeping the file in memory?

jmsmkn:
It allows re-use of existing layers, so only the layers that do not exist need to be downloaded. Docker also does some downloading in parallel. It was only done the other way as the registry was added later.

PR #2392, opened 2022-04-12
Add answering time as float

pkcakeout:
Adds the answering time for questions to the Answer-model. 

Relates to https://github.com/DIAGNijmegen/rse-roadmap/issues/131.

Todos:
- [ ] Make it a Duration field as suggested by james 
- [ ] Verify that the validation still works as suggested as I need the answer_time to be patchable even if the reader-study is readonly...


jmsmkn:
I would rather that @MikeOverkamp-diag implemented this.

1. Why would the answer be editable if the study is read only? Doesn't make sense to me.
2. Blank is validation related, so that needs to be True as setting the time is not required.
3. `answer_` is redundant, that was only there for the image to differentiate between the image that the answer is for, and the image created by the answer. 
4. This needs tests.

pkcakeout:
Hi James! Yes, not problem. But I need something rather early :-) So this is that "something". Who knows when @MikeOverkamp-diag has time for this. How the data is uploaded is not really important but I need a endpoint that I can patch for testing the integration.

MikeOverkamp-diag:
The only thing I'm wondering is if this is something we want on the answer model, as this is more so related to the display_set than a specific answer I'd say?

jmsmkn:
This is a property of the Answer, it is to time how long each user took creating that answer for that display set. Requirements:

- `last_edit_time`: DurationField is added to the answer model, default `None`, blank true
  - This field must be tracked by Django Simple History
  - This field is added to the Answer serializer, setting it is optional
  - This field can optionally be set on POST or PATCH (if the answer is editable)
- `total_edit_time`: A property added to the `Answer` Model
  - This sums all of the historical `last_edit_time`s for each answer
  - If any of the `last_edit_times`s are `None`, then `total_edit_time` is `None`
  - It should be serialized read only on the Answer
  - When serializing the answers ensure that the historical answers are prefetched for this calculation, or if possible do it in the DB

pkcakeout:
@MikeOverkamp-diag 

I proposed something like attaching this to the DisplaySet, but it gets way too complicated then. We probably need a new table for this kind of data then because display-set level "answers" are not a thing yet. The weird thing will be that there are multiple read-times stored for the same case this way, but whatever :P

pkcakeout:
@jmsmkn Don't we also need some nice overview page that renders out this data on grand-challenge? Or do we only expose it through the API for now?

jmsmkn:
> @jmsmkn Don't we also need some nice overview page that renders out this data on grand-challenge? Or do we only expose it through the API for now?

API only, this is for editors analysis. They can download it through the CSV files which are generated automatically by DRF.

jmsmkn:
>     * [ ]  Verify that the validation still works as suggested as I need the answer_time to be patchable even if the reader-study is readonly...

@MikeOverkamp-diag Following the comment here https://github.com/DIAGNijmegen/rse-roadmap/issues/131#issuecomment-1098961347 this no longer needs to be the case, so don't change anything about when the patching can be done. I would suggest starting on a new branch.

pkcakeout:
@MikeOverkamp-diag 

When do you have time to work on this? I will keep using this branch for now, but ultimately we need to do a proper integration. Would be nice to have some sort of approximate plan for when you have time to work on this... Like right now, in a week?

MikeOverkamp-diag:
Assume it will be finished next Tuesday (as I have the day off tomorrow and am still doing the final migration stuff today). As James suggested in https://github.com/comic/grand-challenge.org/pull/2392#issuecomment-1098924620, you can use the `last_edit_duration` in your post/patch requests. I think it will be easiest to send an integer (time in s or ms), unless you think something else would be better?

jmsmkn:
Use a DRF DurationField (automatic). The representation is a string following this format `'[DD] [HH:[MM:]]ss[.uuuuuu]'` https://www.django-rest-framework.org/api-guide/fields/#durationfield

pkcakeout:
> I think it will be easiest to send an integer (time in s or ms), unless you think something else would be better?

I used float (s) which is accurate enough and easy.

But a duration field is also fine. Gimme something and I will translate it into the apropriate form :laughing:... I have to admit though that that duration format is a bit annoying to parse :-/ Is there a "not-django" standard-function for that? Anyway, I will write my own if not...

jmsmkn:
> > I think it will be easiest to send an integer (time in s or ms), unless you think something else would be better?
> 
> I used float (s) which is accurate enough and easy.
> 
> But a duration field is also fine. Gimme something and I will translate it into the apropriate form 😆... I have to admit though that that duration format is a bit annoying to parse :-/ Is there a "not-django" standard-function for that? Anyway, I will write my own if not...

```
>>> from datetime import timedelta
>>> str(timedelta(seconds=1337, microseconds=2))
'0:22:17.000002'
```

pkcakeout:
Did not know that. But from what I see datetime library does not seem to include a parser for those time-delta formats... :shrug: 

From my googling, there is no standard way. There are pip packages for that though: https://pypi.org/project/pytimeparse/

But I fear that using one-trick pony libraries like these leads to npm levels of fragmentation. Hand-writing seems better to me?! Opinions?

jmsmkn:
Why do you need to parse? The timings should be independent, right?

Here is how Django/DRF does it, open source so can take that with a license:
https://github.com/django/django/blob/deedf5bbc347e47b3be6e15783fc43a9c0a69256/django/utils/dateparse.py#L132-L154

pkcakeout:
When "allow_answer_modification" and "allow_case_navigation" are enabled, I need to add up the old reading time with the new reading time. Check video of https://github.com/DIAGNijmegen/rse-cirrus-core/pull/1868 . So they _can_ be incremental.

And yes, ok, of course. Taking the django implementation seems the logic solution :laughing:

jmsmkn:
The backend will handle summing things together, you only need to provide the timing for the most recent edit, so you can store and track it however you like.

pkcakeout:
Also fine. Details needed. Will wait until next week when the grand-challenge version is done to do the integration correctly. I do the summing-up in the frontend right now. :shrug:

jmsmkn:
See this comment: https://github.com/comic/grand-challenge.org/pull/2392#issuecomment-1098924620

You will only need to post/patch `last_edit_duration`. 

`total_edit_duration` is read only.

pkcakeout:
Yes, sure. That was decided after I was done with this PR. I will rework it when there is an implementation :-)

PR #2393, opened 2022-04-13
Disallow slug editing for pages

jmsmkn:
This PR fixes a long standing bug where challenge admins could change the `title` of their page. Doing this would break existing links. The `title` field was effectively a `slug`, so this PR updates it to use an auto slug field and names it `slug`. This takes the unique together constraint into account so that the same slug can exist for different challenges.

`display_title` could now be renamed to `title` for consistency with other models, but that may have been confusing with the existing fields.


amickan (approved):
LGTM! How are you going to migrate the titles of existing challenges from the `title` field to the `display_title` field? Don't you need a management command for that?

jmsmkn:
The `title` field is automatically renamed to `slug` in the migration:

```
migrations.RenameField(
            model_name="page",
            old_name="title",
            new_name="slug",
        ),
```

The `display_title` stays the same.

PR #2394, opened 2022-04-13
Refactor updating archive item kwargs method

HarmvZ:
Came across this part of the code and it seemed slightly awkward with a side effect that could be removed from the method.


jmsmkn:
Having side effects is fine so long as it is clear that is the intent, and IMO it is pretty clear that is what will happen from the docs. I don't think it's necessary to refactor this, and if it is then the proposed implementation needs to be fixed as it doesn't send on the updated sets to downstream tasks. When returning multiple objects from a method a named tuple should be used.

PR #2395, opened 2022-04-13
Split interface overview into inputs and outputs

jmsmkn:
This should provide users with more context, and provides more explanation on what an interface is.


app/grandchallenge/algorithms/templates/components/componentinterface_io_switch.html
```diff
@@ -0,0 +1,64 @@
...
+    </ol>
+{% endblock %}
+
+{% block content %}
+    <h2>Algorithm Interface Options</h2>
+
+    <p>
+        You will need to define the inputs to, and the outputs from, your algorithm.
+        The set of inputs and outputs that your algorithm uses is known as its interface.
+        You can build up your algorithms interface by selecting from the current set of interface options.
```
amickan, line 22:
```suggestion
        You can build up your algorithm's interface by selecting from the current set of interface options.
```

app/grandchallenge/algorithms/templates/components/componentinterface_io_switch.html
```diff
@@ -0,0 +1,64 @@
...
+        Use the links below to see the options that are available for your algorithms interface.
```
amickan, line 23:
```suggestion
        Use the links below to see the options that are available for your algorithm's interface.
```

amickan (approved):
This is super clear I think! Thank you!

amickan (approved):

PR #2396, opened 2022-04-14
Add display-set filter to answer

chrisvanrun:
This PR adds a display-set filter to the (mine) end-point of answers.

This is needed to selectively retrieve answers per-case in CIRRUS. 

In addition, I've added a display_set addition to the development fixture reader-study.

### DIFF
--- a/app/grandchallenge/reader_studies/views.py
+++ b/app/grandchallenge/reader_studies/views.py
@@ -1134,7 +1134,11 @@ class AnswerViewSet(
     serializer_class = AnswerSerializer
     queryset = (
         Answer.objects.all()
-        .select_related("creator", "question__reader_study")
+        .select_related(
+            "creator",
+            "question__reader_study",
+            "display_set",
+        )
         .prefetch_related("images")
     )
     permission_classes = [DjangoObjectPermissions]
--- a/app/grandchallenge/reader_studies/views.py
+++ b/app/grandchallenge/reader_studies/views.py
@@ -1137,7 +1137,6 @@ class AnswerViewSet(
         .select_related(
             "creator",
             "question__reader_study",
-            "display_set",
         )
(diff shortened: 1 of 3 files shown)


MikeOverkamp-diag (dismissed):

app/grandchallenge/reader_studies/views.py
```diff
@@ -1134,7 +1134,11 @@ class AnswerViewSet(
     serializer_class = AnswerSerializer
     queryset = (
         Answer.objects.all()
-        .select_related("creator", "question__reader_study")
+        .select_related(
+            "creator",
+            "question__reader_study",
+            "display_set",
```
jmsmkn, line 1140: I don't think adding display_set here is necessary, did you check in the debug toolbar?

chrisvanrun: I queried Mike about if it would make sense to add it here. What would you expect to see in the debug toolbar? (Inexperience in Django optimalisations).

chrisvanrun: Heh, just a single glance to remind me what `select_related` does already points myself to the correct check. If any DB queries are performed to fetch the actual display_set... they should be added.

Will check.

jmsmkn: Yes, that's it. You only need it if you need any information from the display sets in this view. If you're just serializing a hyperlink to the display set then that can already be generated without fetching the extra data as that is a foreign key already in the answer table. Always best to check by counting the number of SQL queries for the view with multiple rows.

chrisvanrun: Doesn't seem to matter much. So I have removed the select_related. Good catch @jmsmkn .

MikeOverkamp-diag (approved):

PR #2398, opened 2022-04-14
Enable html emails in admin

amickan:
This adds the action email template from [mailgun ](https://github.com/mailgun/transactional-email-templates)to our email templates and enables mass sending of html emails through the Django admin. 

The email template is relatively simple, just a box of max-width 600px that contains whatever content we want and ends with "Your Grand Challenge Team" as well as the unsubscribe link underneath the box. 

SimpleTest

This is what the most recent newsletter would look like: 

NewsletterTest

I tested how the email displays in different clients with [email on acid](https://www.emailonacid.com/). After some small adjustments to the template, it now looks fine on all tested clients (although image display depends on how you add them in the markdown editor!). 

image

image

image

We could consider adding a banner similar to the one we have on the homepage to the top of the email to make it more recognizable. 

Addresses https://github.com/DIAGNijmegen/rse-roadmap/issues/130

### DIFF
--- a/app/grandchallenge/emails/admin.py
+++ b/app/grandchallenge/emails/admin.py
@@ -1,7 +1,9 @@
 from django.contrib import admin, messages
 from django.db.transaction import on_commit
+from django.forms import ModelForm
 from markdownx.admin import MarkdownxModelAdmin
 
+from grandchallenge.core.widgets import MarkdownEditorAdminWidget
 from grandchallenge.emails.models import Email
 from grandchallenge.emails.tasks import send_bulk_email
 from grandchallenge.emails.utils import SendActionChoices
@@ -23,9 +25,17 @@ def schedule_emails(modeladmin, queryset, request, action):
         )
 
 
+class EmailAdminForm(ModelForm):
+    class Meta:
+        model = Email
+        widgets = {"body": MarkdownEditorAdminWidget}
+        exclude = ()
+
+
 class EmailAdmin(MarkdownxModelAdmin):
     list_display = ("subject", "sent", "sent_at")
     actions = [*SendActionChoices]
+    form = EmailAdminForm
 
     @admin.action(description="Send to mailing list", permissions=["change"])
     def send_to_mailing_list(self, request, queryset):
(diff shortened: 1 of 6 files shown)


app/grandchallenge/emails/tasks.py
```diff
@@ -76,20 +93,29 @@ def send_bulk_email(action, email_pk):
...
         for recipient in paginator.page(page_nr).object_list:
             user = get_user_model().objects.get(pk=recipient.pk)
+            link = reverse(
+                "profile-update", kwargs={"username": user.username}
+            )
+            html_content = render_to_string(
+                "vendor/mailgun_transactional_emails/action.html",
+                {
+                    "title": subject,
+                    "content": safe(html_body),
```
jmsmkn, line 103: `safe` shouldn't be needed here.

app/grandchallenge/emails/tasks.py
```diff
@@ -76,20 +93,29 @@ def send_bulk_email(action, email_pk):
         messages = []
         for recipient in paginator.page(page_nr).object_list:
             user = get_user_model().objects.get(pk=recipient.pk)
+            link = reverse(
```
jmsmkn, line 96: FYI `action.html` is treated just like any other django template. You can load extensions, filters and do things in the template like normal, rather than doing it here.

app/grandchallenge/emails/admin.py
```diff
@@ -23,9 +25,17 @@ def schedule_emails(modeladmin, queryset, request, action):
         )
 
 
+class EmailAdminForm(ModelForm):
+    class Meta:
+        model = Email
+        widgets = {"body": MarkdownEditorAdminWidget}
```
jmsmkn, line 31: Does the email render correctly in the preview or do we need a separate view?

amickan: The content that we enter there renders correctly yes. You don't see the final email there though, because the content is only added to the template later. 

The markdown editor in general (we use it also for blog posts) doesn't look good in the admin though. I'm not sure how easy it is to improve, but in any case that would be something for a separate PR.

image

app/grandchallenge/emails/tasks.py
amickan: Yes, makes sense, but I think for the url it's just as easy to construct it here. Otherwise I would need to pass the username to the template.

app/grandchallenge/emails/tasks.py
amickan: Oh yes, you're right. This was a relic of my debugging of earlier implementations that didn't work. Removed! ;-)

jmsmkn (approved):

PR #2402, opened 2022-04-15
Fix markdowneditor widget for admin

amickan:
The `MarkdownEditorAdminWidget` didn't display nicely in dark mode. The toolbar was almost invisible and in the preview some of the text would show in black rather than white and hence be invisible as well. Also, the main div had a white rather than a black background. 

image

This PR fixes that: 

image

image

### DIFF
--- a/app/grandchallenge/core/static/css/markdown.css
+++ b/app/grandchallenge/core/static/css/markdown.css
@@ -1,0 +1,31 @@
+@media (prefers-color-scheme: dark) {
+    body {
+        background-color: black
+    }
+
+    .module {
+        background-color: black
+    }
+
+    h2 {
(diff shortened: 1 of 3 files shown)


app/grandchallenge/core/static/css/markdown.css
```diff
@@ -0,0 +1,31 @@
+@media (prefers-color-scheme: dark) {
```
HarmvZ, line 1: Maybe add a comment in this file to explain why it is necessary (bootstrap vs markdown css  conflicts.)

HarmvZ (approved):
Clean fix!

HarmvZ (approved):

PR #2406, opened 2022-04-19
Email create, update and detail views

amickan:
This adds create, update and detail views for emails (restricted through global permissions). 

I also noticed that I had gotten rid of the "Dear user" preprending to the email body, so added that back to template. 

Addresses: https://github.com/DIAGNijmegen/rse-roadmap/issues/130

### DIFF
--- a/app/grandchallenge/emails/templates/emails/email_detail.html
+++ b/app/grandchallenge/emails/templates/emails/email_detail.html
@@ -27,19 +27,17 @@
             </div>
         </div>
         <div class="col-4">
-            {% if not object.sent and "change_email" in perms %}
+            {% if not object.sent and "emails.change_email" in perms %}
                 <div class="alert alert-warning ml-3" role="alert">
                         This email has not been sent yet.
                         You can come back and edit it. When it's ready to be sent,
                         please contact support@grand-challenge.org to send it for you.
                 </div>
                 <div class="text-right">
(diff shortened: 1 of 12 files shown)


app/grandchallenge/emails/templates/emails/email_detail.html
```diff
@@ -0,0 +1,47 @@
+{% extends "base.html" %}
+{% load url %}
+{% load bleach %}
+
+{% block title %}{{ object.subject }} - Emails - {{ block.super }}{% endblock %}
+
+{% block breadcrumbs %}
+    <ol class="breadcrumb">
+        <li class="breadcrumb-item text-light">Emails</li>
```
jmsmkn, line 9: Maybe add a list view too so that people can find their emails back, both drafts and previously sent?

app/grandchallenge/emails/views.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class EmailUpdate(
+    LoginRequiredMixin,
+    PermissionRequiredMixin,
+    UpdateView,
+):
+    model = Email
+    form_class = EmailForm
+    permission_required = "emails.change_email"
```
jmsmkn, line 27: We should also check that the email has not been sent, maybe raise a permission denied if it has?

jmsmkn (approved):

PR #2407, opened 2022-04-19
Add edit duration fields to Answer model

MikeOverkamp-diag:


app/tests/reader_studies_tests/test_api.py
```diff
@@ -1692,3 +1701,111 @@ def test_display_set_delete(client):
...
+        viewname="api:reader-studies-answer-detail",
+        user=u,
+        client=client,
+        method=client.patch,
+        reverse_kwargs={"pk": pk},
+        content_type="application/json",
+        data={"answer": "bar", "last_edit_duration": "00:30"},
+    )
+    assert response.status_code == 200
+    assert response.json()["last_edit_duration"] == "00:00:30"
```
jmsmkn, line 1746: Is this correct? Shouldn't we ignore the value on patch if the total is `None`?

app/tests/reader_studies_tests/test_api.py
```diff
@@ -1692,3 +1701,111 @@ def test_display_set_delete(client):
...
+        user=u,
+        client=client,
+        method=client.patch,
+        reverse_kwargs={"pk": pk},
+        content_type="application/json",
+        data={"answer": "bar", "last_edit_duration": "00:30"},
+    )
+
+    assert response.status_code == 200
+    assert response.json()["last_edit_duration"] == "00:00:30"
```
jmsmkn, line 1810: Same here.

app/tests/reader_studies_tests/test_api.py
MikeOverkamp-diag: I thought it would make more sense to handle the value that has been patched, but just not update the total, since that is a caclulated value. I'd be happy to change it though.

jmsmkn: I see, yes I think this makes sense.

jmsmkn (approved):

PR #2408, opened 2022-04-19
Add disjoint interface validation to forms

### LINKED ISSUE #2404: Cannot update algorithm interface if set to duplicates
There is a problem with the algorithm clean method where the duplicate interfaces are only set by checking the current m2m relations:

https://github.com/comic/grand-challenge.org/blob/9cce6958f2be5bc77e5e3bb327e4be1e63ddbd10/app/grandchallenge/algorithms/models.py#L251-L262

If a user sets a duplicate interface then the clean method won't catch this as the current input and output are fine. Then the algorithm object is saved the m2m updated. Then when the user tries to save the object again the clean method will fail, so they cannot correct this.

Checking the m2m needs to be done in the forms instead. Also add this check to the django admin for the algorithms and phases.

jmsmkn:
Closes #2404

### DIFF
--- a/app/grandchallenge/algorithms/admin.py
+++ b/app/grandchallenge/algorithms/admin.py
@@ -1,7 +1,9 @@
 from django.contrib import admin
 from django.db.models import Count
+from django.forms import ModelForm
 from guardian.admin import GuardedModelAdmin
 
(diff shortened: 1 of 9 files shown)


amickan (approved):
LGTM!

PR #2413, opened 2022-04-20
Add index for display sets

MikeOverkamp-diag:


app/grandchallenge/reader_studies/views.py
```diff
@@ -1024,17 +1032,22 @@ def destroy(self, request, *args, **kwargs):
...
+        # Note: if more fields besides 'reader_study' are added to the
+        # filter_set fields, we cannot call super anymore before randomizing
+        # as we only want to filter out the display sets for a specific
+        # reader study.
+        reader_study = self.reader_study
+        queryset = super().filter_queryset(queryset)
+        if reader_study and reader_study.shuffle_hanging_list:
+            set_seed(1 / int(self.request.user.pk))
+            queryset = queryset.order_by("?")
+            self.randomized_qs = list(queryset)
```
jmsmkn, line 1045: Do you really need every value here? Might this be a huge request (and memory use)?

app/grandchallenge/reader_studies/views.py
```diff
@@ -1047,20 +1060,11 @@ def get_queryset(self):
                 answers__creator=self.request.user,
                 answer_count__gte=answerable_question_count,
             )
-        return queryset
+            if reader_study and reader_study.shuffle_hanging_list:
+                pks = queryset.values_list("pk", flat=True)
```
jmsmkn, line 1064: Doesn't this fetch the queryset twice as it's done above in L1045?

app/grandchallenge/core/utils/query.py
```diff
@@ -8,3 +11,11 @@ def index(queryset, obj):
             return index
 
     return -1
+
+
+def set_seed(seed):
+    # Note: this only works for postgres, if we ever switch dbs, this
+    # may need changing.
+    cursor = connection.cursor()
+    cursor.execute(f"SELECT setseed({seed});")
```
jmsmkn, line 20: Could you explain how this works? From the docs this looks like it sets the seed for the current session. Am I right in thinking this only affects the current request? There's no way that some other worker could call this?

MikeOverkamp-diag: It only effects the current request and only the very first subsequent query.

app/grandchallenge/reader_studies/views.py
MikeOverkamp-diag: This is the filtered list, so it's a different list. In order to retain the proper index we need to filter the filtered values out of the initial full queryset. I hope that makes sense.

app/grandchallenge/reader_studies/views.py
MikeOverkamp-diag: We do need the full queryset, but I think we can get away with just storing the pks here now that I think of it.

app/grandchallenge/reader_studies/views.py
jmsmkn: Would be worth adding some comments to help us in the future.

app/grandchallenge/reader_studies/views.py
jmsmkn: Can you add a test for the comment here?

MikeOverkamp-diag: A test only to ensure that `reader_study` is the only filter_field?

jmsmkn: Yes, and referring the user back to this method if that does change.

MikeOverkamp-diag: Using just the pks didn't work unfortunately, as this list is used later on if `unanswered_by_user=True`

app/grandchallenge/reader_studies/serializers.py
```diff
@@ -79,6 +79,16 @@ class DisplaySetSerializer(HyperlinkedModelSerializer):
...
         source="reader_study.view_content", read_only=True
     )
+    index = SerializerMethodField()
+
+    def get_index(self, obj):
+        if obj.reader_study.shuffle_hanging_list:
+            try:
+                return self.context["view"].randomized_qs.index(obj)
+            except ValueError:
+                return obj.standard_index
```
jmsmkn, line 89: Why would this occur?

app/tests/reader_studies_tests/test_api.py
```diff
@@ -1657,6 +1658,88 @@ def test_display_set_add_and_edit(client, settings):
...
+@pytest.mark.django_db
+def test_display_set_index(client):
+    rs = ReaderStudyFactory()
+    r = UserFactory()
+    rs.add_reader(r)
+
+    for _ in range(20):
+        civ = ComponentInterfaceValueFactory(image=ImageFactory())
+        ds = DisplaySetFactory(reader_study=rs)
+        ds.values.add(civ)
```
jmsmkn, line 1670: This is going to be quite slow. Can you do this with less objects? Also remember about the create/build batch methods of factory boy. You might be able to hardcode the users pk for seed setting to make this deterministic.

app/grandchallenge/reader_studies/serializers.py
MikeOverkamp-diag: In the detail view the list is empty. The index is incorrect there for shuffled hanging lists. I checked with @chrisvanrun if he uses the index in the detail view, but I'm fairly sure the answer is no.

jmsmkn: OK, please add a comment and a test. Would returning `None` here to note that it is invalid make sense?

MikeOverkamp-diag: A lot more sense probably :slightly_smiling_face:

jmsmkn (approved):

PR #2414, opened 2022-04-20
Catch validation error if challenge short name already exists

amickan:
Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/11

### DIFF
--- a/app/grandchallenge/challenges/admin.py
+++ b/app/grandchallenge/challenges/admin.py
@@ -42,7 +42,10 @@ class ChallengeRequestAdmin(ModelAdmin):
 
     @admin.display(description="Total cost")
     def total_cost(self, obj):
-        return obj.budget.get("Total")
+        if obj.budget:
+            return obj.budget.get("Total")
+        else:
+            return None
 
     @admin.action(description="Create challenge for this request")
     def create_challenge(self, request, queryset):
--- a/app/grandchallenge/challenges/forms.py
+++ b/app/grandchallenge/challenges/forms.py
@@ -14,7 +14,6 @@ from django.utils.text import format_lazy
 from django_select2.forms import Select2MultipleWidget
 from django_summernote.widgets import SummernoteInplaceWidget
 
-from grandchallenge.challenges.emails import send_challenge_status_update_email
 from grandchallenge.challenges.models import (
     Challenge,
     ChallengeRequest,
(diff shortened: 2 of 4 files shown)


app/grandchallenge/challenges/views.py
```diff
@@ -361,7 +362,14 @@ def form_valid(self, form):
                 form.instance.status
                 == form.instance.ChallengeRequestStatusChoices.ACCEPTED
             ):
-                challenge = form.instance.create_challenge()
+                try:
+                    challenge = form.instance.create_challenge()
+                except ValidationError:
+                    raise ValidationError(
```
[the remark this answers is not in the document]
jmsmkn: Maybe this has gotten a little complex, the responsibility of `form_valid` is to save the instance. Usually here we add in any side effects like adding an async task. If something is going to raise a validation error then it should be done and handled in the forms clean methods.

app/grandchallenge/challenges/forms.py
```diff
@@ -534,3 +535,26 @@ def __init__(self, *args, **kwargs):
             self.fields["status"].widget.attrs["disabled"] = True
         self.helper = FormHelper(self)
         self.helper.layout.append(Submit("save", "Save"))
+
+    def clean_status(self):
```
jmsmkn, line 539: The responsibility of a `clean_` method is only to clean the forms field, too much is happening here. Please think carefully about what a method should do, and where things should be put. You will likely need to split up the functionality here.

app/grandchallenge/challenges/forms.py
```diff
@@ -534,3 +535,26 @@ def __init__(self, *args, **kwargs):
...
+        if (
+            self.instance._orig_status
+            == self.instance.ChallengeRequestStatusChoices.PENDING
+            and self.instance._orig_status != status
+        ):
+            if status == self.instance.ChallengeRequestStatusChoices.ACCEPTED:
+                try:
+                    challenge = self.instance.create_challenge()
+                except ValidationError:
+                    raise ValidationError(
```
jmsmkn, line 550: Why catch and raise the same thing? The underlying Validation method should give the correct error. Maybe you need to clean the short name instead? Or is that already done?

app/grandchallenge/challenges/forms.py
```diff
@@ -534,3 +535,26 @@ def __init__(self, *args, **kwargs):
...
+    def clean_status(self):
+        status = self.cleaned_data.get("status")
+        if (
+            self.instance._orig_status
+            == self.instance.ChallengeRequestStatusChoices.PENDING
+            and self.instance._orig_status != status
+        ):
+            if status == self.instance.ChallengeRequestStatusChoices.ACCEPTED:
+                try:
+                    challenge = self.instance.create_challenge()
```
jmsmkn, line 548: Objects should not be created in a clean method.

app/grandchallenge/challenges/forms.py
```diff
@@ -534,3 +535,26 @@ def __init__(self, *args, **kwargs):
...
+                except ValidationError:
+                    raise ValidationError(
+                        f"There already is a challenge with short "
+                        f"name: {self.instance.short_name}. Contact "
+                        f"support@grand-challenge.org to accept this request.",
+                    )
+            else:
+                challenge = None
+            send_challenge_status_update_email(
```
jmsmkn, line 557: Emails should not be sent in a clean method.

app/grandchallenge/challenges/forms.py
amickan: The short name is cleaned, but it only checks if a challenge request with the same short name exists, not if a challenge with the same short name exists. I could add that to the validation, but I think I should wait with doing that until we stop asking organizers of already existing challenges to file a request in retrospect. They would then have to come up with a new short name and things would get a little messy. Of course, for brand new challenges is makes sense to also check whether the short name exists among challenges.

app/grandchallenge/challenges/admin.py
```diff
@@ -42,7 +42,10 @@ class ChallengeRequestAdmin(ModelAdmin):
 
     @admin.display(description="Total cost")
     def total_cost(self, obj):
-        return obj.budget.get("Total")
+        if obj.budget:
```
amickan, line 45: This addresses the following issue: https://sentry.io/organizations/grand-challenge/issues/3212784983/?project=303639&query=is%3Aunresolved

app/grandchallenge/challenges/forms.py
amickan: And as per raising the ValidationError, I'm doing this to add a custom error message (adding the contact support part).

jmsmkn (approved):

PR #2415, opened 2022-04-22
Add validation for hl shuffling with case navigation

MikeOverkamp-diag:
As discussed in https://github.com/DIAGNijmegen/rse-roadmap/issues/135


jmsmkn (approved):
LGTM - would be good to split this test out so the form is tested directly, rather than using a full integration test.

PR #2416, opened 2022-04-22
Prevent error in values_for_interfaces

MikeOverkamp-diag:
Somehow the `interfaces` dict sometimes has `None` keys, which breaks sorting (see https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/12). I haven't been able to reproduce this behavior, but this PR will at least prevent the display set list view from erroring out if it happens.


jmsmkn:
Unsure about this fix, I think we need to get to the root cause of why some of those values are None.

jmsmkn:
Closing as this only occurred for beta testers of the display sets, and this is likely resolved now.

PR #2419, opened 2022-04-26
Challenge submission form doi field

amickan:
This adds a field where users can enter the DOI of their BIAS challenge submission form (in addition to or instead of uploading the PDF). The DOI makes it easier to share with possible external reviewers. 

This also includes some minor changes to the help texts and the request list and detail views. 

I did not end up adding reviewer comments or an 'assigned to' field because the 10% teams has decided to organize the review offline on Teams and double administration does not make sense.


jmsmkn (approved):
LGTM - would be nice to have this as a custom field now that we can see the reuse.

Issue #2420, opened 2022-04-28
Reader-study display-sets with "unanswered_by_user=true" calculated incorrectly


pkcakeout:
# Describe the bug

A very coarse explanation from the CIRRUS-side of things:

https://user-images.githubusercontent.com/11632379/165755616-42fedc50-eb4c-4886-b686-ab977c97e6bb.mp4

More details: It seems that queries of the style: `https://grand-challenge.org/api/v1/reader-studies/display-sets/?reader_study=1e5946ed-46f8-4501-92fb-d85c91fa853c&unanswered_by_user=true` are affected. These somehow still return the already answered cases as valid candidates for answering even though they should not. This is also evident from the fact that the reader-study progress is not moving (I think - that might be a CIRRUS bug as well).

PR #2421, opened 2022-04-28
Make "True" comparison case insensitive

pkcakeout:
Issue #2420

### DIFF
--- a/app/grandchallenge/reader_studies/views.py
+++ b/app/grandchallenge/reader_studies/views.py
@@ -1048,7 +1048,7 @@ class DisplaySetViewSet(
         unanswered_by_user = self.request.query_params.get(
             "unanswered_by_user"
         )
-        if unanswered_by_user == "True":
+        if str(unanswered_by_user).lower() == "true":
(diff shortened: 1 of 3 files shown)


app/grandchallenge/reader_studies/views.py
```diff
@@ -1048,7 +1048,7 @@ def filter_queryset(self, queryset):
         unanswered_by_user = self.request.query_params.get(
             "unanswered_by_user"
         )
-        if unanswered_by_user == "True":
+        if str(unanswered_by_user).lower() == "true":
```
pkcakeout, line 1051: `str()` to coerce the type which can be "None" or anything else, I guess, into a string.

jmsmkn:
Ooh, there is a `strtobool` method we define in settings to standardise this, maybe use that? https://github.com/comic/grand-challenge.org/blob/c974fada1707df888e0c5a0e209503e1bdb261bc/app/config/settings.py#L26-L28

jmsmkn (approved):

PR #2423, opened 2022-05-02
Exclude slug from historical record

jmsmkn:
This causes a RuntimeError when there are many pages with
the same name.

Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/17


amickan (approved):
LGTM! More of a general question: how often do we make use of the model history for challenge pages? Do we get requests from users to retrieve previous versions a lot or do we use the records for something else?

jmsmkn:
Nothing yet, we could offer an interface to restore a previous version of the page, visualise changes, but is only there just in case right now.

PR #2432, opened 2022-05-03
Fix reader study copying

### LINKED ISSUE #2431: Reader study copying does not work with display sets.
It looks like you end up with images set on the reader study again, and no display sets present.

https://github.com/comic/grand-challenge.org/blob/c68d359bcd23e45becdc1899e4e299859ed3e3ee/app/grandchallenge/reader_studies/views.py#L603

MikeOverkamp-diag:
Closes #2431

### DIFF
--- a/app/grandchallenge/reader_studies/tasks.py
+++ b/app/grandchallenge/reader_studies/tasks.py
@@ -130,8 +130,6 @@ def add_image_to_answer(*, upload_session_pk, answer_pk):
 def copy_reader_study_display_sets(*, orig_pk, new_pk):
     orig = ReaderStudy.objects.get(pk=orig_pk)
     new = ReaderStudy.objects.get(pk=new_pk)
-    if orig.display_sets.count() == new.display_sets.count():
-        return
     with transaction.atomic():
         for ds in orig.display_sets.all():
             new_ds = DisplaySet.objects.create(reader_study=new)
--- a/app/grandchallenge/reader_studies/tasks.py
+++ b/app/grandchallenge/reader_studies/tasks.py
@@ -124,3 +124,15 @@ def add_image_to_answer(*, upload_session_pk, answer_pk):
         add_image(answer, image)
     else:
         raise ValueError("Upload session for answer does not match")
+
+
+@shared_task(**settings.CELERY_TASK_DECORATOR_KWARGS["acks-late-2xlarge"])
+def copy_reader_study_display_sets(*, orig_pk, new_pk):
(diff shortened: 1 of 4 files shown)


app/grandchallenge/reader_studies/views.py
```diff
@@ -599,10 +599,14 @@ def form_valid(self, form):  # noqa: C901
...
         )
         rs.add_editor(self.request.user)
-        if form.cleaned_data["copy_images"]:
-            rs.images.set(reader_study.images.all())
-        if form.cleaned_data["copy_hanging_list"]:
-            rs.hanging_list = reader_study.hanging_list
+        if form.cleaned_data["copy_display_sets"]:
+            for ds in reader_study.display_sets.all():
+                new_ds = DisplaySet.objects.create(reader_study=rs)
+                new_ds.values.set(ds.values.all())
```
jmsmkn, line 605: Could this be moved to an asynchronous task? This operation will take a while (see, migrating studies to display sets) and we will have problems with users getting timeout errors and then trying again, copying the study many times. Add a success message to the response to let the user know that the task will be done asynchronously. Should be on the long tasks queue, using an atomic block.

app/grandchallenge/reader_studies/tasks.py
```diff
@@ -124,3 +124,15 @@ def add_image_to_answer(*, upload_session_pk, answer_pk):
         add_image(answer, image)
     else:
         raise ValueError("Upload session for answer does not match")
+
+
+@shared_task(**settings.CELERY_TASK_DECORATOR_KWARGS["acks-late-2xlarge"])
+def copy_reader_study_display_sets(*, orig_pk, new_pk):
+    orig = ReaderStudy.objects.get(pk=orig_pk)
+    new = ReaderStudy.objects.get(pk=new_pk)
+    if orig.display_sets.count() == new.display_sets.count():
```
jmsmkn, line 133: Why is this check needed?

jmsmkn: That should only happen in development though. If you did that in production this check wouldn't help as the two tasks would likely run in parallel. If the task is only called once then the atomic block is enough protection as SQS and celery with acks late guarantees only one execution of the task, but it could get called many times before completion which is where the atomic block adds its protection.

MikeOverkamp-diag: Ah thank you for explaining! I was not sure if the atomic block was enough when I accidentally double called it, but that makes a lot of sense. I'll remove the check.

jmsmkn (approved):

PR #2433, opened 2022-05-04
Custom identifier model field

amickan:


app/grandchallenge/publications/utils/__init__.py
```diff
@@ -1,11 +1,75 @@
...
+
+identifier_validator = RegexValidator(regex=f"{DOI_REGEX}|{ARXIV_REGEX}")
+
+
+class PublicationType(models.TextChoices):
+    DOI = "D"
+    ARXIV = "A"
+
+
+class IdentifierField(models.CharField):
```
jmsmkn, line 23: Interesting stuff!

I think a better home for this would be in a `fields.py`. Would be great if you could move `get_publication_type`, `get_identifier_csl` and `get_publication_url` to `@properties` on the field now, so we could do things like `publication.identifier.kind` (note not type as that is reserved), `publication.identifier.csl` and `publication.identifier.url` rather than using separate methods.

app/grandchallenge/publications/utils/__init__.py
```diff
@@ -1,11 +1,75 @@
...
+            "help_text"
+        ] = "The DOI, e.g., 10.1002/mrm.25227, or the arXiv id, e.g., 2006.12449"
+        kwargs["validators"] = [identifier_validator]
+        super().__init__(*args, **kwargs)
+
+    def deconstruct(self):
+        name, path, args, kwargs = super().deconstruct()
+        del kwargs["max_length"]
+        del kwargs["help_text"]
+        del kwargs["validators"]
```
jmsmkn, line 41: Does this mean that users cannot change the help_text or validators? I think we would still want to keep that as an option.

app/grandchallenge/publications/utils/__init__.py
```diff
@@ -1,11 +1,75 @@
...
+
+    description = (
+        "The DOI e.g., 10.1002/mrm.25227, or the arXiv id, e.g., 2006.12449"
+    )
+
+    def __init__(self, *args, **kwargs):
+        kwargs["max_length"] = 255
+        kwargs[
+            "help_text"
+        ] = "The DOI, e.g., 10.1002/mrm.25227, or the arXiv id, e.g., 2006.12449"
```
jmsmkn, line 33: Duplicate?

amickan: The description is different from the help text. It's a way to document what the field is used for. It would be shown in the admin docs, but I just realized that we don't use that feature, so I guess I can leave the description out entirely.

app/grandchallenge/publications/utils/__init__.py
amickan: Yes, indeed, as it is now the `help_text` and the `validators` cannot be changed by users. I agree that it makes sense to leave the `help_text` customizable, but why would we want the `validators` to be customizable? Isn't the validation **the** thing that defines this custom field?

jmsmkn: Not really - we're using a custom field here to add helper methods to make the field more usable (see my first review comment). You want to force that this validator is called, but you also want the user to be able to extend the validation used (say, I only want arxiv ids). The way to do this is not to force the validators to be only this one, but to either extend the list of validators so the one you want is always there, or to extend the fields `validate` method:

> The validate() method on a Field handles field-specific validation that is not suitable for a validator. It takes a value that has been coerced to a correct datatype and raises ValidationError on any error. This method does not return anything and shouldn’t alter the value. You should override it to handle validation logic that you can’t or don’t want to put in a validator.
https://docs.djangoproject.com/en/4.0/ref/forms/validation/

e.g. https://github.com/django/django/blob/27b07a3246bc033cd9ded01238c6dc64731cce35/django/db/models/fields/json.py#L104-L113.

This way the fields special validation is always done, but the consumer can do additional things.

The same would go for the max_length field, you can set a default but this forcing pattern makes the field more difficult to use.

jmsmkn: Here is a good example with a default validator and max_length preset https://github.com/django/django/blob/27b07a3246bc033cd9ded01238c6dc64731cce35/django/db/models/fields/__init__.py#L2523-L2545

app/grandchallenge/publications/utils/__init__.py
amickan: Hm, I cannot get it to work like this. If I add a property to the field and then try to call `publication.identifier.kind`, for example, it throws an error saying that a string doesn't have an attribute 'kind'. The property is registered on the Field class, but I can't call it.

jmsmkn: Ah yes, my mistake, you will need to return and take a different class with those properties, rather than a string. Worth reading through this documentation https://docs.djangoproject.com/en/4.0/howto/custom-model-fields/ and especially https://docs.djangoproject.com/en/4.0/howto/custom-model-fields/#converting-values-to-python-objects-1

amickan: Sorry, I'm still stuck. I created a new class `PublicationIdentifier` with those properties and I overwrite the `from_db_value` and `to_python` methods on the custom field to return an instance of this new class. That seems to work, but now saving a new `Publication` to the database does not work. It can't adapt the type 'PublicationIdentifier', but I haven't been able to figure out what I'm missing.

amickan: It took forever, but I think I got it now.

app/grandchallenge/publications/fields.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+class PublicationType(models.TextChoices):
+    DOI = "D"
+    ARXIV = "A"
+
+
+class PublicationIdentifier:
+    def __init__(self, identifier):
+        self.identifier = identifier
```
jmsmkn, line 25: It would be nicer to make this protected as you don't really want it to be changed outside of the class.

```suggestion
    def __init__(self, identifier):
        self.__identifier = identifier
```

jmsmkn (approved):

PR #2434, opened 2022-05-05
Add `opacity` and `parent_id` fields to viewitem definition in hanging protocol.

HarmvZ:


app/tests/hanging_protocols_tests/test_forms.py
```diff
@@ -166,3 +166,93 @@ def test_hanging_protocol_dimension_validation(client):
...
         not in response.rendered_content
     )
+
+
+@pytest.mark.django_db
+def test_hanging_protocol_parent_id_draggable(client):
+    user = UserFactory()
+    assign_perm("hanging_protocols.add_hangingprotocol", user)
+
+    response = get_view_for_user(
```
jmsmkn, line 176: Please re-write this test to test the form directly, rather than going through the API. You should be able to remove the django_db marker.

app/tests/hanging_protocols_tests/test_forms.py
```diff
@@ -166,3 +169,51 @@ def test_hanging_protocol_dimension_validation(client):
         "Either none or all viewports must have x, y, w, and h keys."
         not in response.rendered_content
     )
+
+
+def test_hanging_protocol_parent_id_draggable(client):
```
jmsmkn, line 174:
```suggestion
def test_hanging_protocol_parent_id_draggable():
```

jmsmkn (approved):

PR #2437, opened 2022-05-06
Challenge type badge

amickan:


app/grandchallenge/challenges/models.py
```diff
@@ -583,6 +584,17 @@ def status(self):
             status = StatusChoices.CLOSED
         return status
 
+    @property
+    def type(self):
+        phase_types = {phase.submission_kind for phase in self.phase_set.all()}
+        # as long as one of the phases is type 2,
+        # the challenge is classified as type 2
+        if 3 in phase_types:
```
jmsmkn, line 592: Magic number 3?

app/grandchallenge/challenges/models.py
```diff
@@ -583,6 +584,17 @@ def status(self):
             status = StatusChoices.CLOSED
         return status
 
+    @property
+    def type(self):
```
jmsmkn, line 588: Don't use `type` for a property as this is confusing with the `type()` built in.

app/grandchallenge/evaluation/templates/evaluation/evaluation_list.html
```diff
@@ -101,7 +101,7 @@ <h2>Evaluations</h2>
                             {% endif %}
                         </td>
                         <td>
-                            {% if evaluation.submission.phase.submission_kind == evaluation.submission.phase.SubmissionKind.ALGORITHM %}
+                            {% if evaluation.submission.phase.submission_kind == SubmissionKind.ALGORITHM %}
```
jmsmkn, line 104: Does this actually work? I wouldn't have thought that SubmissionKind would be in the context. Same for the leaderboard row. You can still define `SubmissionKind = SubmissionKind` on the `Phase` model so that all of the existing things still work, then becomes a much smaller changeset.

jmsmkn: Yeah, the only problem is if those two references somehow shift, but, that is unlikely.

Issue #2401, opened 2022-04-14
Finalise the display set migration


jmsmkn:
We now need to finalise the migration. I have already removed a few things with image removal and hanging list generation.

For `ReaderStudy`
- [x] Remove all usages of `use_display_sets`, this is now always true so take the `True` blocks when needed.
- [x] Rewrite tests to ensure that functionality is still covered
- [x] Remove use of `hanging_list`
- [x] Remove use of `images`
- [x] Mark the `images`, `hanging_list`, `use_display_sets`
- [x] Remove any unused methods

Note that in the serializers some methods need to stay until CIRRUS has gone through 2 more release cycles to get these changes: https://github.com/DIAGNijmegen/rse-gcapi/commit/f35686648add70006819bf21dd1864a770753f2a and https://github.com/DIAGNijmegen/rse-gcapi/commit/1ba3199be6c992a8634c0deebbda3357a763936d

For `Answer`
- [x] Remove use of `images`
- [x] Mark `images` as deprecated

Make a PR and we'll merge that. Then make a 2nd PR removing all the deprecated fields for good.

MikeOverkamp-diag:
Status update: still fixing/rewriting tests, which is quite a bit of work. After that I still need to check for unused functions and properties. I will continue on Monday.

PR #2438, opened 2022-05-09
Deprecate images

MikeOverkamp-diag:
This deprecates all fields related to reader study images. When this is merged, we can safely remove the fields from the model. and close https://github.com/comic/grand-challenge.org/issues/2401


jmsmkn (approved):
LGTM - I took a stab at fixing some smaller things but they're all reversible.

PR #2444, opened 2022-05-12
Improve performance of db query

jmsmkn:
This seems counter-intuitive as it is more db requests (4 vs 1), but
the Q object approach would skip indexes so was really slow.

https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/5


MikeOverkamp-diag (approved):
In that case, this seems like a solid solution. Nice find.

PR #2445, opened 2022-05-12
Ensure no duplicate CIs for archive items

### LINKED ISSUE #2397: Adding of duplicate CIs to an archive item
The set of CIs for an archive item should be unique. However, it is currently possible to add multiple CIV with the same CI for a single archive item. This is because the check to see if an a CI already exists and should be removed for an archive item is done asynchronously. So while the task to add a CIV to an item is queued, another task can be spawned to add another CIV for the same CI. This will result in both CIVs being added to the item so duplicate CIs on a single item. 

I'm not sure how to mitigate this issue. We could try to check the task queue (if that is even possible) before adding a new task to add a CIV if there are already tasks for that CI in it. Then the new task should be cancelled/not triggered. Because of the asynchronous nature it will then probably be impossible to properly show an error about this to the user. Probably there is a better way, but I'm not seeing it currently.

This issue is further magnified because it is not clear that an upload is processing when adding a new CIV to an archive item. When you add a file type CIV to an archive item it will redirect you back to the archive item list view. When the CIV adding task is still queued after you are redirected which is quite likely, you will not see any indication that it is in the process of being added on that page. Only when you check https://grand-challenge.org/cases/uploads/ you can see the upload session status. So a user could think something went wrong and try to add it again which will result in the issue described above. Maybe it is good to redirect the user to the cases/uploads page after adding a CIV so that they see that the task is queued.

MikeOverkamp-diag:
Closes #2397 

Unfortunately, I do not see a way to add a test for the issue this fixes, as we lose the asynchronous nature of the celery tasks in the test suite.

### DIFF
--- a/app/grandchallenge/archives/tasks.py
+++ b/app/grandchallenge/archives/tasks.py
@@ -74,21 +74,10 @@ def add_images_to_archive_item(
         )
 
 
-@shared_task(**settings.CELERY_TASK_DECORATOR_KWARGS["acks-late-micro-short"])
-def update_archive_item_values(
-    *, archive_item_pk, civ_pks_to_remove, civ_pks_to_add
-):
-    with transaction.atomic():
-        archive_item = ArchiveItem.objects.get(pk=archive_item_pk)
-        archive_item.values.remove(*civ_pks_to_remove)
-        archive_item.values.add(*civ_pks_to_add)
-
-
 def update_archive_item_update_kwargs(
     instance,
     interface,
     civ_pks_to_add,
-    civ_pks_to_remove,
     upload_pks,
     value=None,
     image=None,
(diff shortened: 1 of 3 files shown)


app/grandchallenge/archives/tasks.py
```diff
@@ -139,18 +118,41 @@ def update_archive_item_update_kwargs(
             civ = interface.create_instance(value=value)
             civ_pks_to_add.add(civ.pk)
 
-    return civ_pks_to_add, civ_pks_to_remove, upload_pks
+    return civ_pks_to_add, upload_pks
```
jmsmkn, line 121: We shouldn't return multiple values from a method. If needed, can you change this to a named tuple and update any callers?

jmsmkn:
Is it possible to add a database constraint for this to ensure that it doesn't happen? Or would we need to set that on the through table?

app/grandchallenge/archives/tasks.py
MikeOverkamp-diag: The return value is actually never used, as the sets are updated in place. I'm not sure how I feel about that tbh. I could just return nothing here. What do you think?

MikeOverkamp-diag:
> Is it possible to add a database constraint for this to ensure that it doesn't happen? Or would we need to set that on the through table?

I'll see if I can make this work with a unique constraint. I'm not sure if that would work.

MikeOverkamp-diag:
`unique_together` cannot be used on manytomany fields (which makes sense) and I do not see another easy way to enforce this in the db. What we could possibly do is add an m2mchanged signal that raises an error if there is a duplicate.

jmsmkn:
Don't we have a similar check for Display Sets? Or do we potentially have the same problem there?

MikeOverkamp-diag:
For display sets the part where we add civs to them is handled synchronously, so this wouldn't be an issue there. Whether or not doing this synchronously is the right choice is another question though.

app/grandchallenge/archives/tasks.py
jmsmkn: Then yes if it's not needed remove it.

jmsmkn (approved):

PR #2448, opened 2022-05-17
Add more information to verification form

amickan:


app/grandchallenge/verifications/templates/verifications/verification_form.html
```diff
@@ -19,6 +19,17 @@ <h2>Request Verification</h2>
         The email address that you provide here will be only be used once to verify your identity,
         and will not be used for any further communications.
     </p>
+    <p>
+        Before submitting this form, make sure that your profile information is complete and valid. You must provide
+        your legal full name and the affiliation you indicate should match the email you use for verification below.
```
jmsmkn, line 24: I would remove "legal" here for D&I reasons, we only need the name people would like to be known as.

```suggestion
        your full name and the affiliation you indicate should match the email you use for verification below.
```

amickan: Ok, sure. I copied this from our terms of service. Should we change it there as well then?

jmsmkn: We have to ask a lawyer before changing that but IMO we should, there are many reasons why people shouldn't have to use their legal name rather than the name they're known as, it's not something we should enforce anyway.

jmsmkn (approved):

PR #2449, opened 2022-05-18
Reduce maximum algorithm duration

jmsmkn:
SageMaker currently only supports up to 1 hour of compute per job


amickan (approved):
LGTM! Does this affect any of the existing phases?

jmsmkn:
Maybe, but it's only validation and not a db constraint, that's why I added the filter.

PR #2450, opened 2022-05-18
Add index to display set api detail view

amickan:


app/grandchallenge/reader_studies/views.py
```diff
@@ -1093,6 +1091,28 @@ def filter_queryset(self, queryset):
...
+            return self.get_paginated_response(serializer.data)
+
+        serializer = self.get_serializer(queryset, many=True)
+        return Response(serializer.data)
+
+    def get_object(self):
+        obj = super().get_object()
+        # retrieve the full queryset and save its shuffled version to later
+        # determine the shuffled index for this object
+        if obj.reader_study and obj.reader_study.shuffle_hanging_list:
```
MikeOverkamp-diag, line 1106: You don't need to check for `obj.reader_study` here, as that's a non nullable field on the model, so if it's not there something is very wrong and we probably want an exception.

app/grandchallenge/reader_studies/views.py
```diff
@@ -1053,19 +1053,17 @@ def destroy(self, request, *args, **kwargs):
...
         # as we only want to filter out the display sets for a specific
         # reader study.
         reader_study = self.reader_study
-        queryset = super().filter_queryset(queryset)
         if reader_study and reader_study.shuffle_hanging_list:
-            set_seed(1 / int(self.request.user.pk))
-            queryset = queryset.order_by("?")
+            queryset = self.shuffle_qs(queryset=queryset)
             # Save the queryset to determine each item's index in the serializer
             self.randomized_qs = list(queryset)
```
MikeOverkamp-diag, line 1066: I'd say this should be part of the `shuffle_qs` method, as you do it after each call to that anyway.

MikeOverkamp-diag (commented):
LGTM, couple of small things.

MikeOverkamp-diag (approved):

PR #2451, opened 2022-05-19
Add interface field to question model

HarmvZ:
- Adds `interface` field to question model allowing a user to select the Component Interface that is related to the question.
- Validates if the Component Interface is allowed to be set for the answer type according to this map: https://github.com/comic/grand-challenge.org/blob/78e4c5a00751665e397e677981a4b44c1b207d7f/app/grandchallenge/reader_studies/models.py#L918-L953
- Only shows the allowed Component Interfaces in the select box in the form.


app/grandchallenge/reader_studies/models.py
```diff
@@ -958,6 +1000,9 @@ class ScoringFunction(models.TextChoices):
         default=ScoringFunction.ACCURACY,
     )
     order = models.PositiveSmallIntegerField(default=100)
+    interface = models.ForeignKey(
+        ComponentInterface, on_delete=models.SET_NULL, null=True, blank=True
```
MikeOverkamp-diag, line 1004: Just `null=True`

MikeOverkamp-diag (changes requested):
Just one small thing. I also wonder if we want to create a widget for the interface selector, but we can do that later on once everything is working.

app/tests/reader_studies_tests/test_forms.py
```diff
@@ -382,6 +392,48 @@ def test_image_port_only_with_bounding_box(
...
+    rs = ReaderStudyFactory()
+    for ci_kind in ci_kinds:
+        ci = ComponentInterfaceFactory(kind=ci_kind)
+        f = QuestionForm(
+            {
+                **defaults,
+                "reader_study": rs,
+                "answer_type": answer_type,
+                "image_port": "M" if answer_type in q.annotation_types else "",
+                "interfac": ci,
```
MikeOverkamp-diag, line 416: typo here is making the tests fail I think.

MikeOverkamp-diag (approved):

app/grandchallenge/reader_studies/models.py
```diff
@@ -911,6 +915,44 @@ class AnswerType(models.TextChoices):
...
+    AnswerType.MULTIPLE_LINES: [InterfaceKindChoices.MULTIPLE_LINES],
+    AnswerType.CHOICE: [InterfaceKindChoices.CHOICE],
+    AnswerType.MULTIPLE_CHOICE: [InterfaceKindChoices.MULTIPLE_CHOICE],
+    AnswerType.MULTIPLE_CHOICE_DROPDOWN: [
+        InterfaceKindChoices.MULTIPLE_CHOICE
+    ],
+    AnswerType.MASK: [
+        InterfaceKindChoices.IMAGE,
+        InterfaceKindChoices.HEAT_MAP,
+        InterfaceKindChoices.SEGMENTATION,
```
jmsmkn, line 951: Only segmentations should be valid here.

app/grandchallenge/reader_studies/models.py
```diff
@@ -911,6 +915,44 @@ class AnswerType(models.TextChoices):
     MULTIPLE_LINES = "MLIN", "Multiple lines"
 
 
+ANSWER_TYPE_TO_INTERFACE_KIND_MAP = {
+    AnswerType.SINGLE_LINE_TEXT: [],
+    AnswerType.MULTI_LINE_TEXT: [],
```
jmsmkn, line 920: Shouldn't these map to `InterfaceKindChoices.STRING`?

app/grandchallenge/reader_studies/models.py
```diff
@@ -958,6 +1000,9 @@ class ScoringFunction(models.TextChoices):
         default=ScoringFunction.ACCURACY,
     )
     order = models.PositiveSmallIntegerField(default=100)
+    interface = models.ForeignKey(
+        ComponentInterface, on_delete=models.SET_NULL, null=True
```
jmsmkn, line 1004: I would do `models.PROTECT` as it would mess up a reader study otherwise.

app/tests/reader_studies_tests/test_forms.py
```diff
@@ -382,6 +396,48 @@ def test_image_port_only_with_bounding_box(
     assert Question.objects.all().count() == questions_created
 
 
+@pytest.mark.django_db
+@pytest.mark.parametrize(
+    "answer_type,ci_kinds", ANSWER_TYPE_TO_INTERFACE_KIND_MAP.items()
+)
+def test_question_interface_field(client, answer_type, ci_kinds):
```
jmsmkn, line 403: `client` isn't needed here. Can you do this without `@pytest.mark.django_db`? This test is going to be quite slow as it is creating a lot of `ci`s. I think there is also too much magic going on as it is constructing the test cases from the form of the object in the codebase. It's better if this is explicit.

app/grandchallenge/reader_studies/static/js/reader_studies/interface_choice.js
```diff
@@ -0,0 +1,17 @@
+$(document).ready(function() {
```
jmsmkn, line 1: Could this be done with htmx insead?

MikeOverkamp-diag: Looking into that now. I haven't used it before.

jmsmkn: Anne added it and did some really nice things with the Challenge Request forms. Take a look there but this looks like the example we'd need https://htmx.org/examples/value-select/. You should speak with Anne about it but she is off today.

jmsmkn (commented):
Looks good but this needs tests.

app/tests/reader_studies_tests/test_forms.py
```diff
@@ -364,6 +370,28 @@ def test_question_update(client):
     assert question.order == 100
 
 
+@pytest.mark.django_db
+@pytest.mark.parametrize(
+    "answer_type,interfaces",
+    ((k, v) for k, v in ANSWER_TYPE_TO_INTERFACE_KIND_MAP.items()),
+)
+def test_question_form_interface_field(answer_type, interfaces):
```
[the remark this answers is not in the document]
MikeOverkamp-diag: With this test I want to test whether the correct interfaces are provided for the interface form field and also that all other interfaces are not available. Would it then be sufficient to explicitly provide answer types and interface kinds as parameters (which would also remove the need for the conditionals)?

Als the part [here](https://github.com/comic/grand-challenge.org/pull/2451/files#diff-5c8f5b6a7f08bada7f0d190db9189f635ed1217c1a68ee25dd9f44a861e72e37R320) tests whether an invalid interface can be set for a question, but also does so implicitly. Should I make that more explicit as well?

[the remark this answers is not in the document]
jmsmkn: Sounds good to me.

jmsmkn (approved):

PR #2452, opened 2022-05-23
Enable editors to view users unanswered ds

amickan:
Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/91

### DIFF
--- a/app/grandchallenge/reader_studies/views.py
+++ b/app/grandchallenge/reader_studies/views.py
@@ -1065,6 +1065,21 @@ class DisplaySetViewSet(
         unanswered_by_user = strtobool(
             self.request.query_params.get("unanswered_by_user", "False")
         )
+        user_pk = self.request.query_params.get("user", False)
+        if user_pk:
+            user = get_user_model().objects.filter(pk=user_pk).get()
+            if (
+                user != self.request.user
+                and not self.request.user.groups.filter(
+                    pk=self.reader_study.editors_group.pk
+                ).exists()
+            ):
+                raise PermissionDenied(
+                    "You do not have permission to view the answers of this user."
+                )
+        else:
+            user = self.request.user
+
         if unanswered_by_user is True:
             if reader_study is None:
                 raise DRFValidationError(
@@ -1075,7 +1090,7 @@ class DisplaySetViewSet(
             queryset = (
                 queryset.annotate(answer_count=Count("answers"))
                 .exclude(
-                    answers__creator=self.request.user,
+                    answers__creator=user,
                     answer_count__gte=answerable_question_count,
                 )
                 .order_by("order", "created")
--- a/app/grandchallenge/reader_studies/views.py
+++ b/app/grandchallenge/reader_studies/views.py
@@ -1072,17 +1072,13 @@ class DisplaySetViewSet(
                 "Specifying a user is only possible when retrieving unanswered"
                 " display sets."
             )
-
         if username:
             user = get_user_model().objects.filter(username=username).get()
-            if (
-                user != self.request.user
-                and not self.request.user.groups.filter(
-                    pk=self.reader_study.editors_group.pk
-                ).exists()
+            if user != self.request.user and not self.request.user.has_perm(
+                "change_readerstudy", self.reader_study
             ):
                 raise PermissionDenied(
-                    "You do not have permission to retrieve this users unanswered"
+                    "You do not have permission to retrieve this user's unanswered"
                     " display sets."
                 )
         else:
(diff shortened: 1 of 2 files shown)


app/grandchallenge/reader_studies/views.py
```diff
@@ -1065,6 +1065,21 @@ def list(self, request, *args, **kwargs):
         unanswered_by_user = strtobool(
             self.request.query_params.get("unanswered_by_user", "False")
         )
+        user_pk = self.request.query_params.get("user", False)
+        if user_pk:
+            user = get_user_model().objects.filter(pk=user_pk).get()
+            if (
+                user != self.request.user
+                and not self.request.user.groups.filter(
+                    pk=self.reader_study.editors_group.pk
```
MikeOverkamp-diag, line 1074: Would it make sense to extend this to superusers as well?

MikeOverkamp-diag (approved):
LGTM. Just one small thing.

app/grandchallenge/reader_studies/views.py
amickan: The `ObjectPermissionsFilter` only checks for the display set permissions, not the permissions on the answers associated with the display sets. I thought that checking whether the user is an editor would be much more efficient than checking for the `view_answer` permissions.

amickan: I was confused for a moment here. Checking for answer perms makes no sense, since this endpoint does not return the answers, but simply the display sets. The user is only used to return unanswered display sets for that user. I have added a check for that, and also clarrified the error message to make things clearer.

jmsmkn: No you should always use the permissions directly - that way you don't have to maintain the business logic about who should have access to a particular thing in multiple places in your codebase.

The permission message through me off, this approach seems very strange to me. Why should the display set list care about answer permissions? Answers should be entirely unrelated to this. My feeling is that this view is doing too much. What are you trying to achieve here? Maybe it needs a separate endpoint?

amickan: This is needed to show a user's  progress (that is not the current user, see [here](https://github.com/DIAGNijmegen/rse-roadmap/issues/91)). I'm not sure if a seperate endpoint would be helpful here. It's only a tiny change to the existing list endpoint and if I create a new endpoint, we would also need to refactor all places in Cirrus where we're currently retrieving unanswered display sets (for the current user) through the list view.

app/grandchallenge/reader_studies/views.py
```diff
@@ -1065,6 +1065,30 @@ def list(self, request, *args, **kwargs):
...
+                ).exists()
+            ):
+                raise PermissionDenied(
+                    "You do not have permission to retrieve this users unanswered"
+                    " display sets."
+                )
+        else:
+            user = self.request.user
+            user = self.request.user
```
jmsmkn, line 1090: Duplicated

app/grandchallenge/reader_studies/views.py
```diff
@@ -1065,6 +1065,30 @@ def list(self, request, *args, **kwargs):
         unanswered_by_user = strtobool(
             self.request.query_params.get("unanswered_by_user", "False")
         )
+        user_pk = self.request.query_params.get("user", False)
```
jmsmkn, line 1068: The users pk should not be exposed. This should use the username instead.

app/grandchallenge/reader_studies/views.py
jmsmkn: I know, but this is anyway not used by CIRRUS yet. By not taking care during implementation this is how we end up with overly complex APIs, that are difficult to maintain and more importantly difficult to change. 

Creating a new endpoint with clear responsibilities makes sense to me, as you are doing something different, not just listing the display sets. Then you can have clear permissions on that.

amickan: Retrieving unanswered display sets is used by Cirrus, I believe. The only thing I'm adding here is the option to do this for another user other than the current user. I will create a new endpoint, but that will mean updating Cirrus.

jmsmkn: OK whatever, keep it as it is but update the permission check to use `change_readerstudy` as this is what we use for editors.

pkcakeout:
FYI: Tested this locally against CIRRUS, and it works great, it seems! :-)

app/grandchallenge/reader_studies/views.py
amickan: I talked to Paul about how we would go about creating a new `unanswered_sets` endpoint.  I could add the new endpoint, but leave the current queryparameter on the list view `(unanswered_by_user)` functional until after the next Cirrus release. Once Cirrus is released again and all references have been updated there, we can deprecate the option to query `unanswered_for_user` from the list view. The initial implementation would mean that there is some duplicate code, but that would be ok. What do you think, James?

jmsmkn: Lets keep it as is and move on.

jmsmkn (approved):

PR #2457, opened 2022-05-24
Button for admin view

amickan:
This adds a button to the readerstudy progress page to view user's answers.

image

### DIFF
--- a/app/config/settings.py
+++ b/app/config/settings.py
@@ -1071,6 +1071,7 @@ WORKSTATIONS_READY_STUDY_QUERY_PARAM = "readerStudy"
 WORKSTATIONS_ALGORITHM_JOB_QUERY_PARAM = "algorithmJob"
 WORKSTATIONS_ARCHIVE_ITEM_QUERY_PARAM = "archiveItem"
 WORKSTATIONS_CONFIG_QUERY_PARAM = "config"
+WORKSTATIONS_USER_QUERY_PARAM = "viewAsUser"
 # The name of the network that the workstations will be attached to
 WORKSTATIONS_NETWORK_NAME = os.environ.get(
     "WORKSTATIONS_NETWORK_NAME", "grand-challengeorg_workstations"
--- a/app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_progress.html
+++ b/app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_progress.html
@@ -3,6 +3,7 @@
 {% load url %}
 {% load profiles %}
 {% load humanize %}
+{% load workstations %}
 
 {% block title %}
     Users Progress - {{ block.super }}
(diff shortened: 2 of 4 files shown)


app/grandchallenge/workstations/templatetags/workstations.py
```diff
@@ -64,4 +65,7 @@ def workstation_query(
             }
         )
 
+    if user:
+        query.update({settings.WORKSTATIONS_USER_QUERY_PARAM: user.username})
```
jmsmkn, line 69: This should only be passed in combination with reader study, so I think the place for it is between R40 and R41.

app/tests/workstations_tests/test_templatetags.py
```diff
@@ -18,6 +18,7 @@ def test_workstation_query(settings):
     algorithm_job = AlgorithmJobFactory()
     config = WorkstationConfigFactory()
     archive_item = ArchiveItemFactory()
+    user = UserFactory()
```
jmsmkn, line 21: Whilst you're here could you try refactoring this to not use the DB? You should be able to switch to using the `build` methods of the factories.

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_progress.html
```diff
@@ -26,33 +27,33 @@ <h2>Users Progress for {{ reader_study.title }}</h2>
...
+                                {{ user.progress.questions|floatformat }}%
                             </div>
                         </div>
                     </div>
-                    <div class="col-2 mb-1">
-                        <div class="h-100 d-flex align-items-center">
-                            <div>
-                                &nbsp;Progress: {{ user.progress.questions|floatformat }}%
-                            </div>
+                    <div class="col-6 col-md-3 p-1">
```
jmsmkn, line 47: Yes, if you want to hide this you can add a feature flag. Make a new boolean setting which is named consistently with the existing reader study settings that we can control with environment variables, pass this through to the template, default false.

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_progress.html
```diff
@@ -26,33 +27,33 @@ <h2>Users Progress for {{ reader_study.title }}</h2>
...
-                                     aria-valuenow="{{ user.progress.questions }}"
-                                     aria-valuemin="0" aria-valuemax="100">
-                                </div>
+                    <div class="col-6 col-lg-4 col-md-3 p-1">
+                        <div class="progress h-100">
+                            <div class="progress-bar" role="progressbar"
+                                 style="width: {{ user.progress.questions }}%"
+                                 aria-valuenow="{{ user.progress.questions }}"
+                                 aria-valuemin="0" aria-valuemax="100">
+                                {{ user.progress.questions|floatformat }}%
```
jmsmkn, line 43: Why did you move this? We had lots of issues with that before.

jmsmkn: Doesn't it disappear at some point? When the value approaches 100%

amickan: It disppeared on small screens but that was because of how the divs were set up. I fixed that and I can see it on all screen sizes with a value of 100% and 50%. I can play around some more to make sure it works, but if you prefer moving it back to how it was, also fine with me. 
Large screen
image
Small screen (not particularly pretty but I presume this screen isn't used very often?)
image

MikeOverkamp-diag: I don't recall what the exact issue was, but I think it wasn't at 100% but rather something in the 85-95 range. Maybe check that and 0 just to be sure.

amickan: Ah I see that the text partially disappears in cases of 5% (the too big to display on the blue bar). I'll see if I can fix that and otherwise revert to how it was.

amickan: Changing the position of the text to `absolute` and adding a shadow solves the issue: 
image

I guess it makes sense that the default behaviour is that the text disappears. It's supposed to label only the blue part... So what do you think?

MikeOverkamp-diag: I like what you did, it's better than no text imo.

amickan: Well, we could revert to what we did before, which is printing the progress text in a seperate column next to the progress bar.

app/config/settings.py
```diff
@@ -1115,6 +1115,8 @@
 }
 # Number of minutes grace period before the container is stopped
 WORKSTATIONS_GRACE_MINUTES = 5
+# Feature Flag for reader study admin view
+READER_STUDY_VIEW_AS_USER_FEATURE = False
```
jmsmkn, line 1119: We should be able to change this with an environment variable.

app/tests/workstations_tests/test_templatetags.py
```diff
@@ -11,14 +11,14 @@
 
 @pytest.mark.django_db
```
jmsmkn, line 12: Can the marker go now?

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_progress.html
jmsmkn: I would say keep it as a separate column, then the numbers are aligned when you have multiple readers with different progressions, looks tidier.

app/tests/workstations_tests/test_templatetags.py
amickan: Ah yes sorry!

app/config/settings.py
```diff
@@ -1114,6 +1115,10 @@
 }
 # Number of minutes grace period before the container is stopped
 WORKSTATIONS_GRACE_MINUTES = 5
+# Feature Flag for reader study admin view
+READER_STUDY_VIEW_AS_USER_FEATURE = os.environ.get(
+    "READER_STUDY_VIEW_AS_USER_FEATURE", "False"
+)
```
jmsmkn, line 1121: This won't give you the intended behaviour. In this case the default will be `READER_STUDY_VIEW_AS_USER_FEATURE == "False"`, so when you do `{% if reader_study_view_as_user_feature %}` in your template that will resolve to `True` as `bool("False") == True`. You need to use `strtobool`, see the other boolean settings.

jmsmkn (approved):

PR #2458, opened 2022-05-24
Replace docker volumes with minio/S3 for components IO

jmsmkn:
This PR changes the docker components backend to use minio/S3 rather than docker volumes. It does this by adding an executable, [SageMaker Shim](https://github.com/DIAGNijmegen/rse-sagemaker-shim), to each container image in the registry on import. There are several changes:

- The setup of minio is moved to a separate script. This adds two new buckets, `grand-challenge-components-inputs` and `grand-challenge-components-outputs`. A new user IAM is added that can read from the input bucket, and write to the output bucket. A new private network is added so that the component containers can read and write to S3.
- Each container image has a new layer added to it with everything that is needed to run SageMaker Shim. This binary and neceaary directories are added to the web container image at build time, and then patched in the repo by `crane` in `_mutate_container_image`. This adds the original `cmd` and `entrypoint` used by the container as base64 encoded environment variables, and sets the new `cmd` and `entrypoint` to run SageMaker Shim which runs the original command in a subprocess.
- All container images have a new field, `latest_shimmed_version`, which tracks what version of SageMaker Shim they have in the container registry.
- All component backends now use the image label (which we control) to select which containers to run, rather than the sha256 of the image. This allows us to select between the original (unshimmed) container images, and the shimmed ones.
- Containers run by the docker backend:
  - now run with networking enabled to communicate with minio/S3, this is an internal network in development so they do not have access to the outside world. With the new IAM roles on minio they only have access to the components buckets. A test is added to ensure the containers cannot communicate out.
  - are now responsible for downloading the inputs from minio/S3, decompressing them, and then uploading them to minio/S3. This is all handled by SageMaker Shim.
  - are now responsible for decompressing their inputs, this is handled by SageMaker Shim (TODO, adding tests there)
  - now get run with a command `["invoke", "--pk", job_id, ...]` with the inputs and outputs as expected by SageMaker Shim.
- Several new environment variables are added:
  - `COMPONENTS_SHIM_IMAGES`: whether to shim the container images or not
  - `COMPONENTS_INPUT_BUCKET_NAME`: the name of the bucket that contains the inputs for the job
  - `COMPONENTS_OUTPUT_BUCKET_NAME`: the name of the bucket that contains the outputs for the job
  - `COMPONENTS_DOCKER_NETWORK_NAME`: the name of the network that will be added to the containers run by the docker backend
  - `COMPONENTS_DOCKER_TASK_SET_AWS_ENV`: whether to set the AWS environment for the containers run by the docker backend
  - `COMPONENTS_DOCKER_TASK_AWS_ACCESS_KEY_ID`: the IAM access key added to the containers run by the docker backend
  - `COMPONENTS_DOCKER_TASK_AWS_SECRET_ACCESS_KEY`: the IAM secret key added to the containers run by the docker backend
- `crccheck/hello-world` has been replaced by our own containerized web server for tests

In terms of testing this is just a refactoring so is covered by many integration tests. This is low risk for deployment as we do not use this backend in production, the only new process will be the shimming of the new container images that currently will not be used. We might end up with many new images shimmed with an older version of SageMaker Shim that we eventually not use, but we can clean that up later, I wanted to keep this PR as small as possible.

https://github.com/DIAGNijmegen/rse-roadmap/issues/129


app/grandchallenge/components/backends/docker.py
```diff
@@ -479,53 +493,61 @@ def _create_images_result(self, *, interface, reader):
...
+
         try:
-            with TemporaryDirectory() as tmpdir:
-                temp_file = Path(safe_join(tmpdir, interface.relative_path))
-                get_file(container=reader, src=output_file, dest=temp_file)
-                with open(temp_file, "rb") as f:
-                    civ = interface.create_instance(fileobj=f)
-        except NotFound:
-            raise ComponentException(f"File {output_file} was not produced")
+            with SpooledTemporaryFile(max_size=MAX_SPOOL_SIZE) as fileobj:
```
MikeOverkamp-diag, line 536: What happens of the file is over 1GB? Or is that guaranteed to not happen?

scripts/minio.py
```diff
@@ -0,0 +1,158 @@
...
+def _setup_public_storage():
+    """
+    Add anonymous read only to public S3 storage.
+
+    Only used in development. In production, set a similar policy on the S3 bucket.
+    """
+    host_alias = "local"
+    policy_name = "public_read"
+    policy = {
+        "Version": "2012-10-17",
```
MikeOverkamp-diag, line 36: Would it make sense to move this to a setting as well?

MikeOverkamp-diag (commented):
Just some questions for now, but I'll have another look later.

app/grandchallenge/components/backends/docker.py
jmsmkn: It then gets written to disk, nice way to keep things fast for small files but still be able to handle larger ones without OOMing:

> This class operates exactly as [TemporaryFile()](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryFile) does, except that data is spooled in memory until the file size exceeds max_size, or until the file’s fileno() method is called, at which point the contents are written to disk and operation proceeds as with [TemporaryFile()](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryFile).

https://docs.python.org/3/library/tempfile.html#tempfile.SpooledTemporaryFile

scripts/minio.py
jmsmkn: No IMO, this really as defines the schema of the policy, not the version of the policy itself https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_version.html

MikeOverkamp-diag (dismissed):

jmsmkn:
Lets not merge this until after the long weekend, just in case.

PR #2459, opened 2022-05-24
Separate interface lists for reader studies and archives

miriam-groeneveld:
Well, I thought that would be a quick and simple fix, but was a bit more complicated than I thought. I opted for creating new pages for the different contexts, that required a copy of the AddCasesForm for reader studies to be able to reference the correct url. Alternative was to cram everything into the single componentsinterface_list.html, which became quite messy.

### DIFF
--- a/app/grandchallenge/algorithms/templates/components/componentinterface_list.html
+++ b/app/grandchallenge/algorithms/templates/components/componentinterface_list.html
@@ -26,7 +26,7 @@
         You can select multiple options for your algorithms {{ list_type|lower }}.
         However, the same option cannot be used for both input to and output from your algorithm.
         If an option does not exist for your use case please contact support with the title, description, and kind
-        for your new interface option.
+        for your new interface.
     </p>
 
     <div class="table-responsive">
--- a/app/grandchallenge/archives/forms.py
+++ b/app/grandchallenge/archives/forms.py
@@ -193,9 +193,9 @@ class AddCasesForm(UploadRawImagesForm):
             (
                 'See the <a href="{}">list of interfaces</a> for more '
                 "information about each interface. "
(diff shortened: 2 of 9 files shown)


jmsmkn (commented):
Looks good, shame that it seems like a lot of duplication, this looks like it should be solvable with parametrisation (maybe by moving the view out of algorithms/archives/reader studies and into the components app, and have it routed from the top level). Could you add tests for the new views and form?

miriam-groeneveld:
Agree about the duplication, but using parameters became very messy with the breadcrumbs. I tried that as well: https://github.com/comic/grand-challenge.org/compare/specified_interface_lists_2?expand=1
Not sure which one I prefer.

jmsmkn:
Yes, that's what I mean about moving it to the top level - avoid the breadcrumb problem by moving it out of each app and to the top level. No good example but something like this https://grand-challenge.org/about/, would then be like an interfaces flatpage (but not actually a flatpage). You could even make it a special part of the documentation app. Maybe that's better?

miriam-groeneveld:
> Yes, that's what I mean about moving it to the top level - avoid the breadcrumb problem by moving it out of each app and to the top level. No good example but something like this https://grand-challenge.org/about/, would then be like an interfaces flatpage (but not actually a flatpage). You could even make it a special part of the documentation app. Maybe that's better?

Ah, I see what you mean regarding moving it to the toplevel, that would be better yes. How would moving it to the documentation app work, create a interfaces folder within the documentation templates folder and move it all there? With comparable parameters to distinguish the various contexts?

jmsmkn:
Maybe @amickan has some thoughts on the structure?

miriam-groeneveld:
Closed in favour of new PR: #2485

PR #2463, opened 2022-05-31
Show main image name on ds list view

amickan:
This adds the image name of the first image to the display set list view. 
image

On the list view this does not add any queries, on the statistics view, however, this adds one query per display set and I couldn't find a way to reduce the queries, so leaving it at the display set id for the statistics page.

Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/148

### DIFF
--- a/app/config/settings.py
+++ b/app/config/settings.py
@@ -1285,7 +1285,7 @@ DATA_UPLOAD_MAX_NUMBER_FIELDS = int(
 RETINA_GRADERS_GROUP_NAME = "retina_graders"
 RETINA_ADMINS_GROUP_NAME = "retina_admins"
 
-ENABLE_DEBUG_TOOLBAR = True
+ENABLE_DEBUG_TOOLBAR = False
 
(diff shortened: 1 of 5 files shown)


app/config/settings.py
```diff
@@ -1285,7 +1285,7 @@
 RETINA_GRADERS_GROUP_NAME = "retina_graders"
 RETINA_ADMINS_GROUP_NAME = "retina_admins"
 
-ENABLE_DEBUG_TOOLBAR = False
+ENABLE_DEBUG_TOOLBAR = True
```
jmsmkn, line 1288: This should not be changed

jmsmkn:
> this adds one query per display set

Yes, please revert, this is textbook N+1, we have to avoid that completely.

jmsmkn (approved):
This might be a problem with https://github.com/DIAGNijmegen/rse-roadmap/issues/133 but seems OK for now.

PR #2471, opened 2022-06-02
Challenge stats page correction

amickan:
The phase slug is not unique, so we need to search by pk instead. Sorry that I didn't notice this earlier!


jmsmkn (approved):

app/grandchallenge/pages/views.py
```diff
@@ -153,7 +153,7 @@ def delete(self, request, *args, **kwargs):
 
 def get_average_job_duration_for_phase(phase):
     algorithm_images = Submission.objects.filter(
-        phase__slug=phase.slug
+        phase__pk=phase.pk
```
jmsmkn, line 156: Even simpler, you can just do:

```suggestion
        phase=phase
```

PR #2472, opened 2022-06-02
Add support for .obj files

### LINKED ISSUE #2468: Add support for .obj files
Many challenges (SLCN, 3D Teeth Seg) use surface meshes, I think we're at the point where they should be supported. No need for viewer integration right now.

As [discussed on teams](https://teams.microsoft.com/l/message/19:376eeffc772144cabed3236a2cbb3ad8@thread.tacv2/1650442890564?tenantId=b208fe69-471e-48c4-8d87-025e9b9a157f&groupId=45c973c4-5543-45ad-a8b8-79ea596e6c69&parentMessageId=1650442890564&teamName=Research%20Software%20Engineering&channelName=cirrus-dev&createdTime=1650442890564) it would be good to add support for single `.obj` files in the same way that we do for `.pdf`.

We do not need to do any validation of the file, beyond the mime-type for now.

STL, doesn't seem as widely supported.

amickan:
Closes #2468


app/grandchallenge/components/models.py
```diff
@@ -514,6 +517,11 @@ def get_file_mimetypes(cls, *, kind):
                 "application/x-sqlite3",
                 "application/vnd.sqlite3",
             )
+        elif kind == InterfaceKind.InterfaceKindChoices.OBJ:
+            return (
+                "text/plain",
+                "application/octet-stream",
+            )
```
jmsmkn, line 524: What file did you test this with?

jmsmkn: OK, I'm waiting to get access to the 3D Teeth Seg dataset, we should double check with those.

amickan: I tested with a bunch of files from the 3D teeth set and they are all of mimetype `text/plain`. The only file that was of mimetype `application/octet-stream` was a file that I had created myself yesterday, so I think we can safely ignore that then.

jmsmkn (approved):

PR #2485, opened 2022-06-03
Separate interface lists for reader studies and archives

miriam-groeneveld:
Moved interface pages to the components app.
Changed naming in archive (no more Cases, just Images or Items)so it matches the naming convention in the Interfaces pages.

Old urls
https://grand-challenge.org/algorithms/interfaces/
https://grand-challenge.org/algorithms/interfaces/inputs/
https://grand-challenge.org/algorithms/interfaces/outputs/
New urls:
https://grand-challenge.org/components/interfaces/algorithms/
https://grand-challenge.org/components/interfaces/inputs/
https://grand-challenge.org/components/interfaces/outputs/

We will also need to update the documentation once this is deployed.


jmsmkn:
If you remove URLs we must remember to add redirects to the application after deployment. Can you add a list of the old -> new URLs in your main PR comment?

cc @amickan

app/grandchallenge/reader_studies/forms.py
```diff
@@ -470,3 +471,25 @@ def clean_ground_truth(self):
         values = [x for x in rdr]
 
         return values
+
+
+class AddCasesForm(UploadRawImagesForm):
```
jmsmkn, line 476: This new form should have a test, is it only the url in the help text that is changing? Then maybe add that to the forms init kwarg and use the same form in both places.

jmsmkn (dismissed):

PR #2486, opened 2022-06-03
Fix challenges_for_algorithms cache

jmsmkn:
The value now only stores the attrs used in the related view,
and the key is set never to timeout as it would previously
expire after 300s, yet only set once per day.

Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/47


amickan (approved):
Ah I had completely forgotten that we implemented this. Thanks for fixing this!

Issue #2379, opened 2022-04-07
Display set list view slow with a lot of images


jmsmkn:
**Describe the bug**
The display set list view is slow when there are a large number of images

**To Reproduce**
1. New reader study
2. Upload a bunch of images
3. Navigate to /display-sets/

**Additional context**
I don't think this is an N+1, but might be. More likely rendering all of the images in each drop down selection. If not an N+1 then needs moving to select2 autocomplete with appropriate permissions.

MikeOverkamp-diag:
Shouldn't be an N+1, as that was resolved in #2369. I'll look into it though.

PR #2488, opened 2022-06-03
Refactor display set list view

MikeOverkamp-diag:
Closes https://github.com/comic/grand-challenge.org/issues/2379

This PR:

- Adds a django form for display sets to replace the old html solution
- Refactors the display set list view. It now only loads read only values initially and adds edit buttons which load a django form via htmx to change display sets
- Adds support for interfaces added to questions to provide default anwers (https://github.com/DIAGNijmegen/rse-roadmap/issues/137)
- Removes the harsh caching on the interface values for display sets. The caching causes quite some issues in production, specifically when the code for the cached functions is updated. Performance is good without the caching, but ideally we'd want to chache `values_for_interfaces` a bit harsher than by just using `cached_property` but not as harshly as previously implemented.

display-sets


app/grandchallenge/reader_studies/models.py
```diff
@@ -828,34 +832,13 @@ class Meta:
...
-                    civ.image_id or ""
-                )
-        else:
-            items = {
-                val.interface.slug: {
-                    "title": val.title,
-                    "selected_image": val.image_id,
-                }
-                for val in self.values.all()
+        items = {
```
jmsmkn, line 835: This is much cleaner, the property name `value_list` is a little unclear as it returns a dict. I wonder if this is needed at all now? Could you go straight to iterating over the items? Where does `image_id` come from?

jmsmkn (commented):
Looks good, a few questions.

app/grandchallenge/reader_studies/models.py
MikeOverkamp-diag: I think this can probably be factored out. I'll have a look. `image_id` is provided by django for the foreignkey relation between civs ans images.

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_images_list.html
```diff
@@ -88,23 +87,27 @@ <h5 class="modal-title" id="removeCaseModalLabel">Remove
...
                     complete: (response) => {
-                        window.location.replace(window.location.href);
+                        htmx.ajax('GET', target.data("hx-get"), {target: target.data("hx-target"), swap: target.data("hx-swap")});
                     }
                 })
+
+            });
+
+            // Trigger htmx ajax request here, because using hx- attributes does not work in html loaded by datatables.js
+            // TODO: replace datatables.js with htmx?
```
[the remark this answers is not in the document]
MikeOverkamp-diag: I'll create an issue

app/grandchallenge/reader_studies/views.py
```diff
@@ -1274,3 +1254,88 @@ class QuestionInterfacesView(View):
...
+):
+    template_name = "reader_studies/readerstudy_display_set_update.html"
+    model = DisplaySet
+    form_class = DisplaySetForm
+    permission_required = (
+        f"{ReaderStudy._meta.app_label}.change_{DisplaySet._meta.model_name}"
+    )
+    raise_exception = True
+
+    def form_valid(self, form):
```
jmsmkn, line 1283: Shouldn't this all be in the forms save method? Isn't the call to `super().form_valid()` missing?

app/grandchallenge/reader_studies/views.py
```diff
@@ -353,15 +359,6 @@ class ReaderStudyStatistics(
     # TODO: this view also contains the ground truth answer values.
     # If the permission is changed to 'read', we need to filter these values out.
 
-    def get_context_data(self, *args, **kwargs):
-        context = super().get_context_data(*args, **kwargs)
-        context.update(
-            {
-                "reader_study_display_set_view_feature": settings.READER_STUDY_DISPLAY_SET_VIEW_FEATURE,
```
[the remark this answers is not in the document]
MikeOverkamp-diag: Had a look and it is still used in the statistics view.

app/grandchallenge/reader_studies/views.py
jmsmkn: Could you add a comment in the code there, important for future users to know why.

I think it should redirect, this always seems like a fairly fundamental thing for web development (https://en.wikipedia.org/wiki/Post/Redirect/Get).

app/grandchallenge/reader_studies/urls.py
```diff
@@ -38,6 +40,16 @@
         QuestionInterfacesView.as_view(),
         name="question-interfaces",
     ),
+    path(
+        "display-sets/<uuid:pk>/",
+        DisplaySetDetail.as_view(),
+        name="display-set-detail",
+    ),
+    path(
+        "display-sets/<uuid:pk>/update",
```
jmsmkn, line 49:
```suggestion
        "display-sets/<uuid:pk>/update/",
```

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_display_set_detail.html
```diff
@@ -0,0 +1,40 @@
...
+    </tr>
+  </table>
+  <div class="d-flex justify-content-between px-1 py-2">
+    {% if object.is_editable %}
+      <button class="btn btn-primary" hx-get="{% url 'reader-studies:display-set-update' pk=object.pk %}" hx-target="#ds-content-{{ object.id }}" hx-swap="outerHTML">Edit</button>
+    {% else %}
+      <div></div>
+    {% endif %}
+    <a href="{% url 'workstations:workstation-session-create' slug=object.reader_study.workstation.slug %}?{% workstation_query display_set=object config=reader_study.workstation_config %}" class="btn btn-primary">
+      <i class="fa fa-eye"></i> View Case
```
jmsmkn, line 37:
```suggestion
      <i class="fa fa-eye"></i> View Display Set
```

jmsmkn (approved):

app/grandchallenge/reader_studies/views.py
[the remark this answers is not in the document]
jmsmkn: This LOC should have no effect on the statistics view, it is still there on L399.

PR #2489, opened 2022-06-05
Add Amazon SageMaker Batch backend

jmsmkn:
This PR adds a new component backend using Amazon SageMaker Batch. The approach taken schedules one task at a time as this is what fits in best with the existing architecture, but can be made more efficient in the future. There are a few assumptions in the backend about this pattern which would need to change if we go to multiple jobs (in particular about logs, duration and metrics parsing). The backend works by placing an invocations json file onto S3, then starting the SageMaker batch transform job that will run a job parametrised by the contents of that json file, and then parsing the outputs as normal. It is event driven so output parsing will happen through EventBridge integration (set up in terraform). Key points:

- For all container image models a new field is added, `runtime_metrics`, that allows us to track the CPU/GPU memory/utilization.

image

- We now have a wider array of instance types to choose from, if we have opted in to them on AWS. Users will no longer need to select a GPU if they just need more memory for their algorithm.
- The time limit for this backend is at most 3600 seconds, this is set by AWS.
- Test use the built in stubber from boto3
- New environment variables:
  - `COMPONENTS_CREATE_SAGEMAKER_MODEL`: whether to create sagemaker models or not, default to False
  - `COMPONENTS_AMAZON_SAGEMAKER_EXECUTION_ROLE_ARN`: the IAM ARN that is passed to the SageMaker model
  - `COMPONENTS_AMAZON_SAGEMAKER_SECURITY_GROUP_ID`: the security group that the SageMaker model will use
  - `COMPONENTS_AMAZON_SAGEMAKER_SUBNETS`: CSV list of the subnets the model will run in

https://github.com/DIAGNijmegen/rse-roadmap/issues/129
#2458


jmsmkn:
This is ready to go but builds are failing due to https://status.python.org/incidents/lgpr13fy71bk, lets wait until Monday.

PR #2493, opened 2022-06-07
Allow blank interfaces for Questions validations

chrisvanrun:
Quick fix


MikeOverkamp-diag:
What problem are you solving with this? We really do not want two different values to indicate an empty field,

chrisvanrun:
> What problem are you solving with this? We really do not want two different values to indicate an empty field,

Sorry, lacking context.

I am busy creating dev fixtures for integration tests for CIRRUS. For this I want to:
* create a Question object
* run full_clean over it (to catch problems early in the fixtures)

One cannot run a full_clean at the moment, since it will complain over the empty interface.

chrisvanrun:
Also, `blank` is for validation purposes, `null` is for database entries. IIRC

MikeOverkamp-diag:
Hmm I think you're right in a foreignkey situation and that this is only problematic in string-type fields. I'd prefer to have @jmsmkn opinion on it before merging though.

MikeOverkamp-diag (approved):

jmsmkn:
The Django Docs sum it up well:

## Blank

> If True, the field is allowed to be blank. Default is False.

> Note that this is different than [null](https://docs.djangoproject.com/en/4.0/ref/models/fields/#django.db.models.Field.null). [null](https://docs.djangoproject.com/en/4.0/ref/models/fields/#django.db.models.Field.null) is purely database-related, whereas [blank](https://docs.djangoproject.com/en/4.0/ref/models/fields/#django.db.models.Field.blank) is validation-related. If a field has blank=True, form validation will allow entry of an empty value. If a field has blank=False, the field will be required.

https://docs.djangoproject.com/en/4.0/ref/models/fields/#blank

## Null

> If True, Django will store empty values as NULL in the database. Default is False.

> Avoid using [null](https://docs.djangoproject.com/en/4.0/ref/models/fields/#django.db.models.Field.null) on string-based fields such as [CharField](https://docs.djangoproject.com/en/4.0/ref/models/fields/#django.db.models.CharField) and [TextField](https://docs.djangoproject.com/en/4.0/ref/models/fields/#django.db.models.TextField). If a string-based field has null=True, that means it has two possible values for “no data”: NULL, and the empty string. In most cases, it’s redundant to have two possible values for “no data;” the Django convention is to use the empty string, not NULL. One exception is when a [CharField](https://docs.djangoproject.com/en/4.0/ref/models/fields/#django.db.models.CharField) has both unique=True and blank=True set. In this situation, null=True is required to avoid unique constraint violations when saving multiple objects with blank values.

> For both string-based and non-string-based fields, you will also need to set blank=True if you wish to permit empty values in forms, as the [null](https://docs.djangoproject.com/en/4.0/ref/models/fields/#django.db.models.Field.null) parameter only affects database storage (see [blank](https://docs.djangoproject.com/en/4.0/ref/models/fields/#django.db.models.Field.blank)).

https://docs.djangoproject.com/en/4.0/ref/models/fields/#null

As `blank=True` is purely validation related, and in forms will set `required=False` on the fields widget. We don't hit this problem in production right now as `required=False` was explicitly set on the form in #2451 (https://github.com/comic/grand-challenge.org/pull/2451/files#diff-8a186bcd8cc7ba8f12b2048ab340f4d429a63f6dc1020a0158c669e43049242eR363). However, we would hit this same problem if creating/editing Questions in the Django admin.

jmsmkn:
And as the foreign keys are non-string based (they are UUID Fields or Integers) then we don't have the issue with multiple representations of "no data".

jmsmkn (approved):

PR #2497, opened 2022-06-07
Two-Factor Authentication

amickan:
This adds Two-Factor Authentication through the [`django-allauth-2fa`](https://github.com/valohai/django-allauth-2fa) package. 

I tested a few other packages, most notably [`django_two_factor_auth`](https://github.com/jazzband/django-two-factor-auth). It appears to be better maintained than `django-allauth-2fa`, but there are some issues with it. Forcing 2fa for existing users that don't have a totp device set-up yet, for example, does not work - you get stuck in an infinite loop. This is a known issue, there is a [workaround](https://github.com/jazzband/django-two-factor-auth/issues/219#issuecomment-494382380) for it and also a [PR to fix it](https://github.com/jazzband/django-two-factor-auth/pull/370), but this PR has been around for while and it's not clear when it will be incorporated. This and some other issues with integrating it made me decide to not use this package. 

I also checked out [`django-mfa-2`](https://github.com/mkalioby/django-mfa2/), but that requires a lot of custom code to integrate. 

With `django-allauth-2fa`, any user of GC has the option to enable 2FA through a link in the profile: 

It works with the most common Authenticator apps and it also support backup tokens. Users can also remove 2FA again. 

image

image

For staff users, 2FA is obligatory. 

Getting this to work with social sign up required some custom code. The other packages also don't support 2FA with social sign up out of the box. 

Back-up tokens:
At the moment, the backup tokens are always shown. That behaviour can be changed by setting `ALLAUTH_2FA_ALWAYS_REVEAL_BACKUP_TOKENS` to False in the settings.


pkcakeout:
Wow... this looks pretty fan-ceee!

app/grandchallenge/profiles/templates/allauth_2fa/setup.html
```diff
@@ -0,0 +1,47 @@
...
+
+<h4>
+  {% trans 'Step 1' %}:
+</h4>
+
+<p>
+  {% trans 'Scan the QR code below with a token generator of your choice (e.g., Google Authenticator, Microsoft Authenticator).' %}
+</p>
+
+<img src="{{ qr_code_url }}" />
```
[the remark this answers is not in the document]
amickan: You can get the token with `get_token_from_totp_device(tot_device)` where `tot_device=user.totdevice_set.last()`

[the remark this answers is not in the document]
jmsmkn: Yes, but in order to enter it into users apps they will need to see the code in the template, rather than, or in addition to, a QR code. This is what it looks like on the GitHub Page, we would need the same (the "enter this text code" bit):

image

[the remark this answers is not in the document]
amickan: Ah ok, sorry, misunderstood your comment. I'll add that!

app/config/urls/root.py
```diff
@@ -58,6 +59,12 @@ def handler500(request):
         name="django.contrib.sitemaps.views.sitemap",
     ),
     path(settings.ADMIN_URL, admin.site.urls),
+    re_path(
+        r"accounts/two_factor/setup/?$",
+        TwoFactorSetup.as_view(),
+        name="two-factor-setup",
+    ),
```
jmsmkn, line 66: This doesn't need to be a `re_path`.

```suggestion
    path(
        "accounts/two_factor/setup/",
        TwoFactorSetup.as_view(),
        name="two-factor-setup",
    ),
```

app/config/urls/root.py
```diff
@@ -4,7 +4,7 @@
 from django.contrib.auth.decorators import login_required
 from django.contrib.sitemaps.views import sitemap
 from django.template.response import TemplateResponse
-from django.urls import path
+from django.urls import path, re_path
```
jmsmkn, line 7:
```suggestion
from django.urls import path
```

app/config/settings.py
```diff
@@ -419,6 +421,13 @@
     "django.contrib.flatpages.middleware.FlatpageFallbackMiddleware",
     # Redirects last as they're a last resort
     "django.contrib.redirects.middleware.RedirectFallbackMiddleware",
+    # 2FA middleware, needs to be after subdomina, flatpage and redirect middleware
+    # TwoFactorMiddleware resets the login flow if another page is loaded
+    # between login and successfully entering two-factor credentials. We're subsetting
+    # the original allauth_2fa middleware to pass the correct urlconf.
+    "grandchallenge.core.middleware.TwoFactorMiddleware",
+    # Force 2AF for staff users
+    "grandchallenge.core.middleware.RequireStaffAndSuperuser2FAMiddleware",
```
jmsmkn, line 430: There are a few typos in this, 2AF should be 2FA and subdomina should be subdomain.

Why do they need to be after FlatpageFallbackMiddleware and RedirectFallbackMiddleware? Does this affect our flatpages and redirects, especially on sub-domains?

jmsmkn: What tests would break if these were out of order? Or not after the subdomain middleware?

amickan: The Flatpages redirect tests fail if I put the middleware before the flatpage middleware. With the current order of the middleware everything works as intended, but I will take another, closer look at the 2FA middleware (also add a test for it).

jmsmkn: Yes I think that makes sense.

app/config/urls/root.py
amickan: Somehow the partial path turns out to be necessary. I'm not sure why, yet, but if I just use path it goes to the original setup view and not the custom one, even thought this url is defined before I include the allauth-2fa urls.

amickan:
- Added some explanations for the back-up tokens. Is this clear?
image
image
- Changed "Two-Factor Authentication Settings" to "2FA settings"
image
- Disabling 2FA now requires entering a 2FA token first. I took inspiration from how this is handled in other 2fa packages. If you prefer to require the user's password instead of a 2FA token, fine as well. 
- The superuser issue is not specific to this PR. I also have it on main. I think this started for me when we moved away from docker compose. I will take a look and fix this in a seperate PR.

jmsmkn (approved):
LGTM, nice work!

PR #2500, opened 2022-06-08
Include form media for json editor

### LINKED ISSUE #2499: JSONEditor fails to load for component interface in display set form
**Describe the bug**
JSONEditor fails to load for component interface in display set form.

**To Reproduce**
1. Create a reader study containing a question where component interface field is set to an interface that is defined using JSON.
1. Go to the cases list view for that reader study.
1. Open a display set and click on edit.
1. The JSONEditor widget is not loaded for the interface field.
1. An error is shown in the console.

**Expected behavior**
JSONEditor widget loads correctly.

**Screenshots**

image

See the empty space below "Region-of-interest". The JSONEditor widget should show up there. As the error in the console indicates, the following files (JSONEditorWidget dependencies) are somehow not loaded:

https://github.com/comic/grand-challenge.org/blob/5a193d8970a66cfd53617262414da90b1ab0b2aa/app/grandchallenge/core/widgets.py#L17-L19

MikeOverkamp-diag:
Closes #2499

### DIFF
--- a/app/grandchallenge/components/form_fields.py
+++ b/app/grandchallenge/components/form_fields.py
@@ -29,6 +29,12 @@ def _join_with_br(a, b):
 
 
 class InterfaceFormField:
+    _possible_widgets = {
+        UserUploadMultipleWidget,
+        UserUploadSingleWidget,
+        JSONEditorWidget,
+    }
+
     def __init__(
         self,
         *,
--- a/app/grandchallenge/reader_studies/forms.py
+++ b/app/grandchallenge/reader_studies/forms.py
@@ -516,10 +516,4 @@ class DisplaySetForm(Form):
                         required=False,
                     )
 
-                if self.fields[slug].widget not in self._possible_widgets:
-                    # The forms media must be pre-loaded in the list view
-                    # which means we need to know all possible media
-                    # from each form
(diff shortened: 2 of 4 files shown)


jmsmkn:
This creates coupling between the view and form, why not include `{{ form.media }}` in the response as usual?

MikeOverkamp-diag:
It seemed like potentially including the same static files 20 times would be a bit much, so I opted to just include it once in the list. Would you prefer to include it for each form?

jmsmkn:
Yes, the browser will ignore any duplicates.

jmsmkn:
> Once the media has been added for one of the forms, the widget works wel for all subsequent forms that are loaded. However, that leaves us with the first form that is loaded. The only solution I have been able to come up with is to check if the media was there on form load and if not, add it and rerender the form. That still leaves an ugly javascript error on first load (and some other small issues with js variables that should be fixable) and a sort of flash when loading the first form. I really do not see any other way to do this without adding the static files to the list view before any forms are loaded unfortunately. I'd love to hear any other ideas on how to solve this issue.

PR added that should do this #2503

jmsmkn (approved):

jmsmkn (approved):

PR #2504, opened 2022-06-09
Add a field for private notes to the algorithm description

nlessmann:
This adds a field "editor_notes" to the Algorithm model, which can be edited in the "Algorithm description" form and which will be visible in the "Admin info" section on the algorithm homepage.

Looks like this:


amickan (approved):
LGTM!

PR #2509, opened 2022-06-10
Allow archive item creation via api

### LINKED ISSUE #2508: Users cannot create archive items without images in the html templates
**Describe the bug**
The only way for a user to create a new archive item through the html templates is to upload an image. Some algorithms work without images, so a user needs a way to upload non-image files directly to new archive items. I think that they can use the API client to work around this, but unsure, @MikeOverkamp-diag could you confirm ASAP?

**Expected behavior**
A user uploads non-image interfaces and archive items are created for them.

**Additional context**
Related to #2506

MikeOverkamp-diag:
Closes #2508

### DIFF
--- a/app/grandchallenge/archives/apps.py
+++ b/app/grandchallenge/archives/apps.py
@@ -19,6 +19,10 @@ def init_archiveitem_permissions(*_, **__):
         f"{ArchiveItem._meta.app_label}.change_{ArchiveItem._meta.model_name}",
         g,
     )
+    assign_perm(
+        f"{ArchiveItem._meta.app_label}.add_{ArchiveItem._meta.model_name}",
+        g,
+    )
 
 
 class ArchivesConfig(AppConfig):
--- a/app/grandchallenge/archives/serializers.py
+++ b/app/grandchallenge/archives/serializers.py
@@ -1,6 +1,7 @@
 from django.db.transaction import on_commit
 from guardian.shortcuts import get_objects_for_user
 from rest_framework import serializers
+from rest_framework.exceptions import ValidationError
 from rest_framework.fields import JSONField, ReadOnlyField, URLField
 from rest_framework.relations import HyperlinkedRelatedField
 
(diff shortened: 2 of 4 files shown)


amickan (approved):
LGTM! Maybe add a check to the test that a normal user cannot create archive items.

app/grandchallenge/archives/apps.py
```diff
@@ -19,6 +19,10 @@ def init_archiveitem_permissions(*_, **__):
         f"{ArchiveItem._meta.app_label}.change_{ArchiveItem._meta.model_name}",
         g,
     )
+    assign_perm(
+        f"{ArchiveItem._meta.app_label}.add_{ArchiveItem._meta.model_name}",
```
jmsmkn, line 23: I don't think the permissions are correct here. A user should only be able to create an archive item for an archive if they have the "upload_archive" permission. As implemented, I think that it means archive users users will have permission to create archive items for the archive.

MikeOverkamp-diag: You are correct. I'm not sure how to properly handle this, as drf automatically checks for global add permissions in the create view. I could add an explicit check in the view/serializer?

jmsmkn: Django views have `get_permission_object`. Probably there is something in the case upload endpoints, pretty sure there are views that rely on permissions of related objects, have a look through all of the POST endpoints in DRF and see how they're handled.

The permission for archives on an `ArchiveItemPostSerializer` is `archives.use_archive`, that might just be plain wrong already.

jmsmkn: Update the test to ensure that it catches this current bug, then change https://github.com/comic/grand-challenge.org/blob/d82b8a7cff0b6651fc3800f08448d11656eb65cb/app/grandchallenge/archives/serializers.py#L82 to `archive.upload_archive` and see if anything else breaks.

MikeOverkamp-diag: Relying on the CI to see if anything breaks. I think if something does break, than that would be a bug anyway.

jmsmkn (approved):

PR #2514, opened 2022-06-13
Remove reader study feature flags

amickan:
This removes the two feature flags `READER_STUDY_DISPLAY_SET_VIEW_FEATURE` and `READER_STUDY_VIEW_AS_USER_FEATURE`. They are no longer necessary since Cirrus now supports them.


jmsmkn (approved):
Were these ever set in terraform? If so they also need removing there.

MikeOverkamp-diag (dismissed):

jmsmkn (approved):
Test failure is spurious.

PR #2516, opened 2022-06-13
Migrate to `docker compose`

jmsmkn:
This PR simplifies a lot of our development tooling:

- Uses the newer `docker compose` rather than `docker-compose`
- Uses the universal docker compose file format
- Removes the local dev option (AFAIK no one was using this)
- Removes `./cycle-docker-compose.sh` and replaces it with `make runserver`
- Moves all environment setting into the `Makefile`
- Adds watchfiles to automatically restart the celery workers
- Runs the tests in the celery worker rather than the web container

Changes to the developers process:

- Use `make runserver` rather than `./cycle-docker-compose.sh`
- Use `docker compose ...` rather than `docker-compose ...`
- Use `docker compose run ... celery_worker_evaluation pytest ...` rather than `docker-compose run ... web pytest ...`


docker-compose.yml
```diff
@@ -151,7 +133,7 @@ services:
       DEBUG: 'true'
       PYTHONDONTWRITEBYTECODE: 1
     restart: always
-    command: "celery -A config worker -l info -c 1 -Q acks-late-2xlarge,acks-late-micro-short"
+    command: watchfiles "celery -A config worker -l info -c 1 -Q acks-late-2xlarge,acks-late-micro-short" /app
```
chrisvanrun, line 136: Cool package.

chrisvanrun (approved):
Good clean-up. Can confirm it works on Win11+WSL2 with Docker-Desktop (4.9.0, latest).

PR #2522, opened 2022-06-14
Validate view_content type

MikeOver1:
Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/51

### DIFF
--- a/app/grandchallenge/hanging_protocols/forms.py
+++ b/app/grandchallenge/hanging_protocols/forms.py
@@ -2,6 +2,8 @@ from crispy_forms.helper import FormHelper
 from crispy_forms.layout import ButtonHolder, Div, Layout, Submit
 from django import forms
 from django.core.exceptions import ValidationError
+from jsonschema import ValidationError as JSONValidationError
+from jsonschema import validate
 
 from grandchallenge.components.models import ComponentInterface
 from grandchallenge.core.forms import SaveFormInitMixin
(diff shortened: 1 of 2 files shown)


app/grandchallenge/hanging_protocols/forms.py
```diff
@@ -88,6 +88,14 @@ def clean_json(self):
 class ViewContentMixin:
     def clean_view_content(self):
         mapping = self.cleaned_data["view_content"] or {}
+        if not isinstance(mapping, dict):
```
jmsmkn, line 91: Shouldn't we be checking against the schema server side, not just that it is an object?

app/grandchallenge/hanging_protocols/forms.py
```diff
@@ -88,6 +88,14 @@ def clean_json(self):
...
     def clean_view_content(self):
         mapping = self.cleaned_data["view_content"] or {}
+        if not isinstance(mapping, dict):
+            self.add_error(
+                error=f"Value {str(mapping)} is not valid. "
+                "Should be of type `object`.",
+                field="view_content",
+            )
+            # return, because the ci check assumes the value is a dict
+            return mapping
```
jmsmkn, line 98: Why return? Shouldn't a Validation error be raised here? I thought `add_error` should only be used on the model clean method, not individual fields.

MikeOverkamp-diag: Looked it up and you're absolutely right. I'll update the form and the test.

app/grandchallenge/hanging_protocols/forms.py
MikeOverkamp-diag: Sorry, completely missed this remark.

jmsmkn: Do you know why this validation wasn't happening in the first place? It is set on the model field right? So why is it invalid here? Do the model validations just happen later?

jmsmkn: > I think the model validation happens after the form validates

Can you confirm this?

MikeOverkamp-diag: It does, in ModelForms, model.full_clean() is called after the form's clean method.

jmsmkn (approved):

Issue #1738, opened 2021-02-26
API schema has warnings


jmsmkn:
In #1737 we added an OpenAPI3.0 schema, but we should go through this and ensure that the spec is complete, things I know are missing:

* [ ] The `CharField(source="get_*_display",...)` should be replaced with custom `ChoiceFields` that use the display values for their representations, so that we get enums for these back in the schema
* [ ] The Schemas for JSONFields with validators should be included in the top level schema
* [x] There are warnings where return types have been inferred, these are mostly correct anyway but need to be fixed, see the TODO in `app/tests/api_tests/test_schema.py`

jmsmkn:
We're on quite an old version of `drf-spectacular`, this also needs to be updated https://github.com/comic/grand-challenge.org/blob/e5145fd991740aa45f9ed56483205c8590900d6b/pyproject.toml#L89

PR #2527, opened 2022-06-16
Add missing filter fields to swagger docs

### LINKED ISSUE #2479: `unanswered_by_user` and `user` options are missing from the Display Set documentation
**Describe the bug**
The swagger documentation should show all the filters available, they are missing from https://grand-challenge.org/api/#/reader-studies/reader_studies_display_sets_list

**Screenshots**
image

MikeOverkamp-diag:
Closes #2479

### DIFF
--- a/app/grandchallenge/reader_studies/views.py
+++ b/app/grandchallenge/reader_studies/views.py
@@ -38,6 +38,12 @@ from django.views.generic import (
     View,
 )
 from django_filters.rest_framework import DjangoFilterBackend
+from drf_spectacular.types import OpenApiTypes
+from drf_spectacular.utils import (
+    OpenApiParameter,
+    extend_schema,
+    extend_schema_view,
+)
 from guardian.mixins import LoginRequiredMixin, PermissionListMixin
 from guardian.mixins import (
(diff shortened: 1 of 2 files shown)


jmsmkn:
Could you add a test for this? We're using an ancient, beta, version of spectacular and the API may have changed by the time we upgrade (#1738)

MikeOverkamp-diag:
This is based on the docs for the latest version btw: https://drf-spectacular.readthedocs.io/en/latest/customization.html

jmsmkn:
Probably it is doing something to the serializer and you can test that directly.

MikeOverkamp-diag:
I think the test is a bit iffy still, but it's the best I could do. At least we'll have a reminder to look at this part of the code if we ever do upgrade.

jmsmkn:
Why not test against the actual generated schema returned by this view? https://github.com/comic/grand-challenge.org/blob/7a5403adacbed60d596ea8ea749440bf94393bcb/app/grandchallenge/api/urls.py#L185

jmsmkn (approved):

PR #2529, opened 2022-06-17
Fix superuser script

amickan:


chrisvanrun:
> Actually could you move this to a Make command like we have for the algorithm evaluation fixtures? Better to have one way of doing things. Would be nice just to do `make superuser`, don't need to configure the username.

That would add a lot of garble to the Make file if you would do it directly. I'd opt to make this a python script and call it via the Makefile, similar to the development_fixtures.

jmsmkn:
Yes:

> _like we have for the algorithm evaluation fixtures_

PR #2534, opened 2022-06-17
Fix ground truth issues

### LINKED ISSUE #2477: Ground truth for optional choice questions
**Describe the bug**
It's not clear how to provide a ground truth when a reader study contains optional questions. What should be the value if an optional question should remain empty?

**To Reproduce**
1. Open a reader study with an optional question of type choice
2. Go to ground truth, download the example, change the values and upload.

**Expected behavior**
A description of what is wrong with the csv and how to fix it.

**Screenshots**
Depending on the content of the csv, it either states invalid values (for example with '"None'):
image

or error (for example with ""):
image

MikeOverkamp-diag:
Closes #2477


app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
```diff
@@ -417,4 +422,22 @@ <h2>Copy</h2>
...
+            $.ajax({
+                url: "{% url 'api:reader-study-remove-ground-truth' pk=object.pk %}",
+                type: "POST",
+                complete: () => {location.reload()},
+                headers: {
+                    'X-CSRFToken': window.drf.csrftoken,
+                    'Content-Type': 'application/json'
+                },
+            });
+        };
```
jmsmkn, line 440: Please move this to its own js file, we have to get rid of all inline js so that we can set a proper CSP (TODO for a long time).

app/grandchallenge/reader_studies/views.py
```diff
@@ -939,7 +943,15 @@ def get_context_data(self, **kwargs):
         return context
 
 
-class ReaderStudyViewSet(ReadOnlyModelViewSet):
+class DeleteGroundTruthPermission(DjangoObjectPermissions):
```
jmsmkn, line 946: Why are you using change permissions for a delete view? Surely `delete_` should be used?

app/grandchallenge/reader_studies/views.py
```diff
@@ -959,6 +971,22 @@ def _check_change_perms(self, user, obj):
         if not (user and user.has_perm(self.change_permission, obj)):
             raise Http404()
 
+    def get_permissions(self, *args, **kwargs):
+        if self.action == "remove_ground_truth":
+            return [DeleteGroundTruthPermission()]
+        return super().get_permissions(*args, **kwargs)
+
+    @action(detail=True, url_path="delete-ground-truth", methods=["post"])
```
jmsmkn, line 979: POST for deletion? Why use DRF for this? Wouldn't a normal view be better, then it could be hooked up with HTMX if needed.

MikeOverkamp-diag: I went through the options for what type of request to use and came across this: https://itnext.io/mass-delete-via-http-rest-how-do-you-do-it-1bff0f5eb72d. The arguments there made sense to me, which is why I opted for POST rather than DELETE. I think moving this to a regular view makes sense though.

app/grandchallenge/reader_studies/views.py
MikeOverkamp-diag: I wasn't sure on this, my reasoning was that any editor for a reader study should be allowed to delete the ground truth for it, which is what I wanted to reflect in the permission check here.

app/grandchallenge/reader_studies/views.py
jmsmkn: Interesting, but that is a bit out of date, it looks like OpenAPI now supports request bodies with DELETE https://github.com/OAI/OpenAPI-Specification/pull/1937 so I would say we could/should use it in this situation.

app/grandchallenge/reader_studies/views.py
jmsmkn: Yes, so give those permissions to the editors group, they probably already have it as they can already delete answers that belong to their reader study from any user.

MikeOverkamp-diag: Currently reader study editors do not have delete permissions for answer objects, no one does actually. So we need to assign them for all existing answers, which will probably take some time. Shall I include a management command for that?

jmsmkn: Not super necessary, it's a quick thing we can do in the shell:

```python
for rs in ReaderStudy.objects.all():
  assign_perm("delete_answer", rs.editors.group, rs.answers.all())
```

MikeOverkamp-diag:
Update on this: I'm a bit stuck, thought I had everything working properly and went to finish things up, but when I return a redirect on the delete method, I end up with a `405 - Method not allowed (DELETE)` on `reader-studies/reader-study`, even though that's only the redirect url, not the actual url I send the delete request to.

```
[23/Jun/2022 19:16:42] "DELETE /reader-studies/reader-study/remove-answers/ HTTP/1.0" 302 -
Method Not Allowed (DELETE): /reader-studies/reader-study/
Method Not Allowed: /reader-studies/reader-study/
[23/Jun/2022 19:16:42] "DELETE /reader-studies/reader-study/ HTTP/1.0" 405 -
```
The weird thing is that whenever I step through the code in a debug session, the error does not appear and I just finish with the 302, which I would expect. I'll have a fresh look tomorrow.

MikeOverkamp-diag:
[This comment](https://github.com/axios/axios/issues/2133#issuecomment-581037874) offered a solution that worked for me: using a 303 status code rather than a 302 for the redirect. This is not something that is used in django at all, so I'm not sure if it's the proper solution.

jmsmkn:
That makes sense to me given https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/303. You could have:

```
class HttpResponseSeeOther(HttpResponseRedirect):
    status_code = 303

return HttpResponseSeeOther("/reader-studies/reader-study/")
```

MikeOverkamp-diag:
Went with that first, but eventually went with a 200 status code and a `HX-Redirect` header to be able to use htmx with the redirect. The htmx header only works of the status code is 200.

app/grandchallenge/reader_studies/views.py
```diff
@@ -822,35 +830,76 @@ def get_context_data(self, **kwargs):
...
+        return reverse(
+            "reader-studies:users-progress",
+            kwargs={"slug": self.kwargs["slug"]},
+        )
+
+
+class AnswersRemoveGroundTruth(AnswerBatchDelete):
+    success_message = "Ground truth removed"
+
+    def get_permission_objects(self, request):
```
jmsmkn, line 896: Why is there a difference between `get_permission_objects` and `get_queryset` - shouldn't they be the same?

app/grandchallenge/reader_studies/views.py
```diff
@@ -939,7 +988,15 @@ def get_context_data(self, **kwargs):
         return context
 
 
-class ReaderStudyViewSet(ReadOnlyModelViewSet):
+class DeleteGroundTruthPermission(DjangoObjectPermissions):
+    def get_required_permissions(self, method, model):
+        return [f"{model._meta.app_label}.change_{model._meta.model_name}"]
```
jmsmkn, line 993: This is still using a change permission?

MikeOverkamp-diag: Ah sorry, this needs to be removed, it is no longer used, the action is no longer part of the viewset.

app/grandchallenge/reader_studies/models.py
```diff
@@ -1341,6 +1353,11 @@ def assign_permissions(self):
             self.question.reader_study.editors_group,
             self,
         )
+        assign_perm(
+            f"delete_{self._meta.model_name}",
+            self.question.reader_study.editors_group,
```
jmsmkn, line 1358: Where is the management command to assign this permission to existing objects?

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
```diff
@@ -417,4 +424,5 @@ <h2>Copy</h2>
 {% block script %}
     {{ block.super }}
     <script src="{% static "js/refresh_sidebar.js" %}"></script>
+    <script src="{% static "rest_framework/js/csrf.js" %}"></script>
```
jmsmkn, line 427: Why is this line added?

app/grandchallenge/reader_studies/views.py
```diff
@@ -939,7 +985,7 @@ def get_context_data(self, **kwargs):
         return context
 
 
-class ReaderStudyViewSet(ReadOnlyModelViewSet):
+class ReaderStudyViewSet(ModelViewSet):
```
jmsmkn, line 988: Why is this now a ModelViewSet? Doesn't this add create/delete views in the API?

app/grandchallenge/reader_studies/views.py
```diff
@@ -822,35 +830,73 @@ def get_context_data(self, **kwargs):
...
-        return super().form_valid(form)
-
     def get_success_url(self):
         return self.reader_study.get_absolute_url()
 
 
+class AnswersRemoveForUser(AnswerBatchDelete):
+    def get_queryset(self):
+        data = QueryDict(self.request.body)
+        user = data["user"]
```
jmsmkn, line 876: This is a strange construction and results in strange API calls in the tests - `data=f"user={r1.id}"`. Shouldn't a form with validation of the user field, or using a URL with a capture value for the username (`"/remove-answers/<str:username>/"`, `creator__username=...`) be nicer?

app/tests/reader_studies_tests/test_views.py
```diff
@@ -135,6 +135,43 @@ def test_answer_remove(client):
     assert Answer.objects.filter(creator=r2).count() == 1
 
 
+@pytest.mark.django_db
+def test_answer_remove_ground_truth(client):
```
jmsmkn, line 139: These tests do not check that answers and ground truth for other reader studies are unaffected.

app/grandchallenge/reader_studies/views.py
```diff
@@ -822,35 +830,73 @@ def get_context_data(self, **kwargs):
...
     )
     raise_exception = True
+    success_message = "Answers removed"
 
-    def get_permission_object(self):
-        return self.reader_study
+    def check_permissions(self, request):
+        permission_objects = self.get_queryset()
+        forbidden = any(
+            not request.user.has_perm(self.permission_required, obj)
```
jmsmkn, line 843: N+1 problem here.

app/grandchallenge/reader_studies/models.py
MikeOverkamp-diag: I didn't include it, as you set it could easily be done via the shell:

> Not super necessary, it's a quick thing we can do in the shell:
> 
> ```python
> for rs in ReaderStudy.objects.all():
>   assign_perm("delete_answer", rs.editors_group, rs.answers.all())
> ```

But I can add it if you'd prefer that :slightly_smiling_face:

app/grandchallenge/reader_studies/views.py
MikeOverkamp-diag: I'll include it in the url, as html5 does not support the delete method for forms.

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_progress.html
```diff
@@ -49,6 +49,7 @@ <h2>Users Progress for {{ reader_study.title }}</h2>
                                 &nbsp;{{ user.progress.questions|floatformat }}%
                             </div>
                         </div>
+                values: {csrfmiddlewaretoken: target.data("csrf")},
```
jmsmkn, line 52: What is this doing?

app/grandchallenge/reader_studies/views.py
```diff
@@ -828,35 +832,73 @@ def get_context_data(self, **kwargs):
...
+                "HX-Refresh": True,
+            },
+        )
 
     @property
     def reader_study(self):
         return get_object_or_404(ReaderStudy, slug=self.kwargs["slug"])
 
-    def form_valid(self, form):
-        form.remove_answers(reader_study=self.reader_study)
```
jmsmkn, line 853: Is the `remove_answers` method on the form still being used?

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_progress.html
MikeOverkamp-diag: Probably an accidental paste. It's gone now

jmsmkn (approved):

PR #2536, opened 2022-06-20
Add compute costs to challenge stats

amickan:
This adds the number of archive items and the average compute costs to the challenge stats page and gives the challenge reviewers (aka 10%) access to these pages.


jmsmkn (approved):
LGTM - we have to be really careful with this view now, do not refresh it as it can DOS the site. We should add a periodic task to set this information in the cache in the background for each challenge on the long running queue.

PR #2537, opened 2022-06-20
Check for number of images in ds update task

MikeOverkamp-diag:


app/grandchallenge/reader_studies/tasks.py
```diff
@@ -54,7 +55,15 @@ def add_image_to_display_set(
...
-    image = Image.objects.get(origin_id=upload_session_pk)
+    upload_session = RawImageUploadSession.objects.get(pk=upload_session_pk)
+    try:
+        image = Image.objects.get(origin_id=upload_session_pk)
+    except (Image.DoesNotExist, Image.MultipleObjectsReturned):
+        error_message = "Image imports should result in a single image"
+        upload_session.status = RawImageUploadSession.FAILURE
+        upload_session.error_message = error_message
+        upload_session.save()
+        raise ImageImportError(error_message)
```
jmsmkn, line 66: We probably don't need to know about this, so I think we can register this exception with Celery to mark the task as failed https://docs.celeryq.dev/en/stable/userguide/tasks.html#Task.throws

jmsmkn (approved):

PR #2554, opened 2022-06-28
Fixes 2FA for social login

### LINKED ISSUE #2544: 2FA reports failure for login but succeeds
* Enable 2FA as per request at first login of my 'admin' account
* Try to log in with my 'admin' account, which is a google account (for extra security).
* Provide 2FA code

Get reported:
Screenshot 2022-06-24 093215

But it seems indeed successful as I can continue to GC.

amickan:
Closes #2544 

I spent quite a lot of time exploring different solutions for the problem, and ultimately went with the solution provided in a pending [PR](https://github.com/valohai/django-allauth-2fa/pull/131) on the django-allauth-2fa repo.


app/grandchallenge/profiles/adapters.py
```diff
@@ -44,6 +54,62 @@ def clean_email(self, email):
...
+            return response
+
+        # Require two-factor authentication if it has been configured or if the user is staff user.
+        if self.has_2fa_enabled(user) or user.is_staff or user.is_superuser:
+            self.stash_pending_login(request, user, **kwargs)
+            # For the social login the Require2FA middleware does not work, so check here again
+            if (
+                user.is_staff or user.is_superuser
+            ) and not self.has_2fa_enabled(user):
+                redirect_url = reverse("two-factor-setup")
```
amickan, line 72: The `BaseRequire2FAMiddleware` middleware does not work with social login because at the time that the middleware is evaluated, the user is not logged in yet. I tried to find a fix for that in the middleware itself, but did not succeed and spent too much time trying. Checking here again is not a pretty solution, but it works.

jmsmkn: I'm a little bit stuck on exactly how to review this PR, honestly I don't know enough about Django AllAuth 2FA to work this out. It is not clear to me exactly what comes from the linked PR, what has been changed, and what can be removed when that PR is released. Could you maybe add some more comments to the new methods, identifying what is verbatim and what you have changed?

amickan: I found out that my custom addition here actually breaks the login for staff users who have not set-up 2FA yet. So I removed the custom part. I now copied the linked PR as is, which means that once the linked PR has been merged, we can drop the `pre_login()`m `stash_pending_login()` and `unstash_pending_login_kwargs()`. 

The `post_login` stuff is necessary if we want to force 2FA set-up for admin users who come from the social route, because the `BaseRequire2FAMiddleware` does not appear to work with social login. I honestly think it would be fine to drop this as well since this only concerns Harm as far as I know and we can just ask him to enable 2FA, no need to force it from the Django side.

jmsmkn: Yes removing it sounds fine to me. Maybe on the users list page in the admin you can add the 2FA status, then site admins can enforce whatever policy they like through social controls.

app/grandchallenge/profiles/admin.py
```diff
@@ -41,6 +42,12 @@ class UserProfileAdmin(UserAdmin):
     )
     actions = (deactivate_users,)
 
+    @admin.display(description="User has 2FA enabled")
+    def has_2fa_enabled(self, obj):
+        return obj.totpdevice_set.filter(confirmed=True).exists()
```
jmsmkn, line 47: I think we need a select/prefetch related in the admin list view now.

amickan: Annotating turned out to be more efficient than prefetching.

jmsmkn (approved):

PR #2559, opened 2022-06-29
Add support for multi category masks

MikeOverkamp-diag:
This implements all required changes to support segmentation masks with multiple categories, as outlined in https://github.com/DIAGNijmegen/rse-roadmap/issues/136.


app/grandchallenge/components/models.py
```diff
@@ -847,6 +912,10 @@ def __str__(self):
     def clean(self):
         super().clean()
 
+        if self.interface.kind == InterfaceKindChoices.SEGMENTATION:
+            self.interface._validate_pixel_values(
+                self.image_file, self._image_file.image_type
+            )
```
jmsmkn, line 918: This can go in the self.interface.is_image_kind block.

Won't this be called every time the CIV is saved? How have you ensured that this only runs in the Celery task on the acks-late-2xlarge queue?

app/grandchallenge/reader_studies/tasks.py
```diff
@@ -72,9 +73,14 @@ def add_image_to_display_set(
...
         )
-        civ, _ = ComponentInterfaceValue.objects.get_or_create(
+        civ = ComponentInterfaceValue.objects.filter(
             interface=interface, image=image
-        )
+        ).first()
+        if civ is None:
+            civ = ComponentInterfaceValue(interface=interface)
+        civ.image = image
+        civ.full_clean()
```
jmsmkn, line 82: The cleaning can no longer be done on the micro short queue.

app/grandchallenge/reader_studies/tasks.py
```diff
@@ -102,12 +113,31 @@ def create_display_sets_for_upload_session(
...
     ):
-        add_image(answer, image)
+        image_file = image.files.filter(
+            image_type__in=[
+                ImageFile.IMAGE_TYPE_MHD,
+                ImageFile.IMAGE_TYPE_TIFF,
+            ]
+        ).get()
+        try:
+            question._validate_pixel_values(
```
jmsmkn, line 129: This is being done on the default queue. This task need to be re-worked so that it is idempotent and moved to the 2xlarge queue.

app/grandchallenge/components/models.py
jmsmkn: Thinking about this approach I think loading Image files in clean methods is asking for trouble. IMO the design should be revisited. Maybe something like this:

- A `segments` JSON field is added to `Image`
  - It is nullable (Default) or contains an array of the unique values in the image
  - We need to ensure that all of the values of an image are not dumped in there, so it has a maximum length, `MAXIMUM_SEGMENTS_LENGTH` (see below)
- When validation of a image is done anywhere the two sets of segments are compared
- `panimg` is responsible for all image interactions and getting the data so that we can validate images later
  - We extract the `segments` from each image there:
    - This is a array of numbers, or default `None`
    - We have to avoid doing this for large images so:
      - Segments should always be None for TIFF type images
      - Segments should always be None if the METAIO image type is not one of the Mask types (https://github.com/Kitware/MetaIO/blob/ffe3ce141c5a2394e40a0ecbe2a667cc0566baf5/src/metaTypes.h#L63-L94 - select the subset that are actually used as masks from there)
      - `panimg` adds a variable `MAXIMUM_SEGMENTS_LENGTH = 32`.
      - If the `number of segments > MAXIMUM_SEGMENTS_LENGTH` then `segments = None` for that image
- `MAXIMUM_SEGMENTS_LENGTH` is also added to the overlay schema 

This will save re-writing and guarding the clean method of the answer and CIV.

poetry.lock
```diff
@@ -2515,24 +2515,7 @@ babel = [
...
-    {file = "backports.zoneinfo-0.2.1-cp37-cp37m-win_amd64.whl", hash = "sha256:a76b38c52400b762e48131494ba26be363491ac4f9a04c1b7e92483d169f6582"},
-    {file = "backports.zoneinfo-0.2.1-cp38-cp38-macosx_10_14_x86_64.whl", hash = "sha256:8961c0f32cd0336fb8e8ead11a1f8cd99ec07145ec2931122faaac1c8f7fd987"},
-    {file = "backports.zoneinfo-0.2.1-cp38-cp38-manylinux1_i686.whl", hash = "sha256:e81b76cace8eda1fca50e345242ba977f9be6ae3945af8d46326d776b4cf78d1"},
-    {file = "backports.zoneinfo-0.2.1-cp38-cp38-manylinux1_x86_64.whl", hash = "sha256:7b0a64cda4145548fed9efc10322770f929b944ce5cee6c0dfe0c87bf4c0c8c9"},
-    {file = "backports.zoneinfo-0.2.1-cp38-cp38-win32.whl", hash = "sha256:1b13e654a55cd45672cb54ed12148cd33628f672548f373963b0bff67b217328"},
-    {file = "backports.zoneinfo-0.2.1-cp38-cp38-win_amd64.whl", hash = "sha256:4a0f800587060bf8880f954dbef70de6c11bbe59c673c3d818921f042f9954a6"},
-    {file = "backports.zoneinfo-0.2.1.tar.gz", hash = "sha256:fadbfe37f74051d024037f223b8e001611eac868b5c5b06144ef4d8b799862f2"},
-]
+"backports.zoneinfo" = []
```
jmsmkn: FYI you need to update poetry to version 1.1.14 and then run `poetry cache clear --all pypi`.

Updating dependencies should be done with `poetry update panimg` so that only that dependency is modified.

app/grandchallenge/components/models.py
```diff
@@ -648,9 +694,24 @@ def create_instance(self, *, image=None, value=None, fileobj=None):
...
         super().clean()
+        self._clean_overlay_segments()
         self._clean_store_in_database()
         self._clean_relative_path()
 
+    def _clean_overlay_segments(self):
+        try:
+            ci = ComponentInterface.objects.get(pk=self.pk)
+        except ComponentInterface.DoesNotExist:
+            return
```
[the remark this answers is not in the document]
MikeOverkamp-diag: This is to get the current value for the overlay segments before saving a new value. Of course, this could be anewly created interface, in which case there is no current db value.

app/grandchallenge/components/models.py
```diff
@@ -648,9 +694,24 @@ def create_instance(self, *, image=None, value=None, fileobj=None):
...
+        if (
+            ComponentInterfaceValue.objects.filter(interface=self).exists()
+            and ci.overlay_segments != self.overlay_segments
```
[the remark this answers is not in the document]
MikeOverkamp-diag: But if we go this route then that check is not needed of course.

app/grandchallenge/cases/models.py
```diff
@@ -322,6 +344,12 @@ def shape_without_color(self) -> List[int]:
         result.append(self.width)
         return result
 
+    def full_clean(self):
+        if self.segments is not None:
+            # panimg returns a tuple, which does not play nicely with jsonschema
```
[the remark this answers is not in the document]
MikeOverkamp-diag: That is what I initially had there, but because the `PanImg` class in panimg is a `dataclass` its fields need to be hashable. Apparently we could work around that by either implementing our own `__hash__` method or by excluding non hashable fields from the hash, but this seemed like a better solution to me. (However, I did not realize using a tuple would give issues with validation in gc.)

app/grandchallenge/cases/models.py
```diff
@@ -277,7 +295,11 @@ class Image(UUIDModel):
     series_description = models.CharField(
         max_length=64, default="", blank=True
     )
-
+    segments = models.JSONField(
+        null=True,
```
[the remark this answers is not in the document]
MikeOverkamp-diag: It should be ok to exclude it, unless our db somehow uses something other then `null` as a default to populate empty fields. So I'll add it to be safe.

jmsmkn (approved):

PR #2560, opened 2022-06-30
Algorithm for challenge phase form and view

amickan:
This implements the changes described here: https://github.com/DIAGNijmegen/rse-roadmap/issues/157 

Creating an algorithm for a Type 2 challenge is now possible through a dedicated `AlgorithmforPhase` form and view. The view checks that:
- user is verified
- user is an admin of the challenge or a participant and the phase is open for submission
- and that the phase: is configured for algorithm submissions (has inputs and outputs and an archive set)

Challenge organizers can define viewer properties on the phase and these (along with a few other things) get automatically set on the algorithm that gets created.


app/grandchallenge/algorithms/urls.py
```diff
@@ -31,6 +32,11 @@
 urlpatterns = [
     path("", AlgorithmList.as_view(), name="list"),
     path("create/", AlgorithmCreate.as_view(), name="create"),
+    path(
+        "create-for-phase/<phase_pk>/",
+        PhaseAlgorithmCreate.as_view(),
+        name="create-for-phase",
+    ),
```
jmsmkn, line 39: I think it would be nicer to have this view in the evaluation app, maybe at `"<slug>/algorithms/create/"`. We shouldn't expose phase pk's, this is better done through the slug and challenge.

app/grandchallenge/algorithms/views.py
```diff
@@ -92,6 +98,75 @@
 logger = logging.getLogger(__name__)
 
 
+class PhaseConfiguredForAlgorithmSubmissionMixin(UserPassesTestMixin):
```
jmsmkn, line 101: A better location for `PhaseConfiguredForAlgorithmSubmissionMixin` and `UserCanSubmitToPhaseMixin` is in the evaluations app, not the algorithms.

app/grandchallenge/algorithms/views.py
```diff
@@ -109,6 +184,45 @@ def form_valid(self, form):
         return response
 
 
+class PhaseAlgorithmCreate(
+    LoginRequiredMixin,
+    PhaseConfiguredForAlgorithmSubmissionMixin,
+    UserCanSubmitToPhaseMixin,
+    VerificationRequiredMixin,
```
jmsmkn, line 191: Have you double checked that using multiple `UserPassesTestMixin` works? I seem to remember there being issues with this a while back that only the first one would be applied.

app/grandchallenge/algorithms/views.py
```diff
@@ -109,6 +184,45 @@ def form_valid(self, form):
...
+        self.object.display_editors = True
+        self.object.contact_email = self.request.user.email
+        self.object.modalities.set(self.phase.challenge.modalities.all())
+        self.object.structures.set(self.phase.challenge.structures.all())
+        self.object.inputs.set(self.phase.algorithm_inputs.all())
+        self.object.outputs.set(self.phase.algorithm_outputs.all())
+        if self.phase.workstation:
+            self.object.workstation = self.phase.workstation
+        self.object.save()
+        return response
```
jmsmkn, line 213: `form_valid` is only responsible for saving the form. You should only need to change this method if you need to add in things like signals. The form itself is responsible for setting any other parameters and doing the object validation. Probably what you need here is a separate form, with the inherited values from the phase set as non-editalble, hidden, fields.

app/grandchallenge/algorithms/views.py
jmsmkn: It might not be worth separating these out into mixins yet, if the functionality is not being used it is premature optimisation.

app/grandchallenge/evaluation/forms.py
```diff
@@ -103,7 +114,12 @@ class Meta:
         }
 
 
-class PhaseUpdateForm(PhaseTitleMixin, forms.ModelForm):
+class PhaseUpdateForm(
+    PhaseTitleMixin,
+    WorkstationUserFilterMixin,
+    forms.ModelForm,
+    ViewContentMixin,
```
jmsmkn, line 121: Shouldn't the ModelForm be last?

app/grandchallenge/algorithms/views.py
amickan: Yes, this is what I was alluding to with my comment above. It works, but it requires explicitly calling `super().test_func()` and then checking the the returned response. If you don't that, indeed only the last mixin that is called will be applied. I agree that it might not be worth seperating at the moment, so I'll combine them into one (also the verification check).

jmsmkn (dismissed):
LGTM!

app/grandchallenge/evaluation/views/__init__.py
```diff
@@ -43,6 +46,66 @@
...
+    Mixin that checks if a user is either an admin of a challenge
+    or a participant of the challenge and that the phase is configured for
+    algorithm submission. If the user is a participant, it also checks that the phase
+    is open for submissions.
+    """
+
+    def test_func(self):
+        response = super().test_func()
+        if response:
+            if not (
```
jmsmkn, line 60: Actually one thing - the user (whether admin or participant), must be verified.

jmsmkn: Sorry my mistake, it's already there.

jmsmkn (approved):

PR #2565, opened 2022-07-01
Add periodic task for phase statistics

amickan:
The challenge statistics page is quite slow for some challenges. This adds a periodic task to retrieve the necessary information and stores it on the Phase model.


app/grandchallenge/evaluation/models.py
```diff
@@ -416,6 +416,21 @@ class Phase(UUIDModel):
...
+    )
+    cached_archive_item_count = models.PositiveIntegerField(
+        editable=False, blank=True, null=True
+    )
+    cached_average_compute_cost = models.DecimalField(
+        max_digits=10, decimal_places=2, editable=False, blank=True, null=True
+    )
+    cached_total_compute_cost = models.DecimalField(
+        max_digits=10, decimal_places=2, editable=False, blank=True, null=True
+    )
```
jmsmkn, line 433: Rather than storing these on the model could you store these in the cache instead? We didn't do this before as we didn't have a shared cache, we have that now. In the periodic task create the key for each phase (include the `pk` in the name) and set it to never expire.

jmsmkn (approved):

PR #2573, opened 2022-07-07
Make challenge stats tab visible to reviewers

amickan:


app/grandchallenge/challenges/templates/challenges/challenge_topbar.html
```diff
@@ -90,7 +90,7 @@
                             </a>
                         </div>
                     {% endif %}
-                    {% if request.user.is_staff %}
+                    {% if request.user.is_staff or perms.challenges.change_challengerequest %}
```
jmsmkn, line 93: This does not match the permission test in the view.

jmsmkn (approved):

Issue #2553, opened 2022-06-28
Displayset view shows wrong widget for interfaces with super-kind File


miriam-groeneveld:
**Describe the bug**
Interfaces can be defined with super_kind File. For annotations this means that the annotation should be provided as a json file. For these interfaces, the "Edit display set" view in a reader study shows the wrong widget (JSON widget instead of file upload widget). This also leads to conflicting database objects: 

![image](https://user-images.githubusercontent.com/51329283/176149508-5f5efcda-b2c2-44ac-97bd-984eb77bbc16.png)

**To Reproduce**
1. Go to a reader study with a question of type Multiple 2D bounding boxes that is linked to interface Regions of interest, for example  [this reader study](https://grand-challenge.org/reader-studies/cirruscoreweb-release-default-answers/)
2. Edit a display set
3. The widget for the default answer for the 2d Bounding box is a JSON widget.

**Expected behavior**
A file upload widget

jmsmkn:
The interface is correct, values can either be saved as files or in the db, see how it is handled in Archive Items.

jmsmkn:
This is what you need to use https://github.com/comic/grand-challenge.org/blob/cace4b153b67ec93e50c9fd706ff8c1d805a94c3/app/grandchallenge/components/models.py#L629-L648

MikeOverkamp-diag:
But then I wouldn't expect a file upload widget for the interface in this issue, as the initial value is also defined in a json editor. The `interface_type_json` and `interface_type_file` functions are used to determine which widget to use, which is what is also done in the display set form.

jmsmkn:
It does not matter. The issue is with the file saving, not the widget.

MikeOverkamp-diag:
This issue focuses specifically on the widget though, which imo is not what needs te be changed, but rather, like you said, the way the value is saved. I'd rather be clear on that before implementing a fix :slightly_smiling_face:

miriam-groeneveld:
That's my bad, sorry for confusing matters. I don't really care what kind of widget is shown. Although it's a bit weird maybe, to paste a huge json object, I think a file upload makes more sense (especially when users download these from other answers or algorithm outputs already anyway)

jmsmkn:
> I'd rather be clear on that before implementing a fix 

I don't know if I can be more clear: just fix the saving.

MikeOverkamp-diag:
[...]

PR #2575, opened 2022-07-10
Support all interface types in display set UI

MikeOverkamp-diag:
Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/50, closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/12, closes https://github.com/comic/grand-challenge.org/issues/2553

This PR implements all changes for https://github.com/DIAGNijmegen/rse-roadmap/issues/159

- It adds a create view for a single display set
- It fixes the widget for json type interface values that are stored as a file
- It adds the option to upload a new file/image for an interface in the display set update view
- It adds the option to add a new interface to a display set via the update view

Peek 2022-07-11 13-23


app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_display_set_update.html
```diff
@@ -1,10 +1,22 @@
...
+            <button type="button" class="close"
+                    data-dismiss="alert" aria-label="Close">
+                <span aria-hidden="true">&times;</span>
+            </button>
+        </div>
+    {% endfor %}
+  {% endif %}
   {% if object.is_editable %}
-    <form action="{% url 'reader-studies:display-set-update' pk=object.pk %}" method="post" class="ds-form" data-hx-get="{% url 'reader-studies:display-set-detail' pk=object.pk %}" data-hx-target="#ds-content-{{ object.id }}" data-hx-swap="outerHTML">
+    <form action="{% url 'reader-studies:display-set-update' pk=object.pk %}" method="post" class="ds-form" data-hx-get="{% url 'reader-studies:display-set-detail' pk=object.pk %}" data-hx-target="#ds-content-{{ object.id }}" data-hx-swap="outerHTML" id="form-{{ object.pk }}">
```
amickan, line 16: Why do you need the `data-hx-get` attribute here? This does a post request to the ds update view, which returns the ds detail view on success. So then this attribute should be redundant here, right?

app/grandchallenge/reader_studies/templates/reader_studies/display_set_add_interface.html
```diff
@@ -0,0 +1,15 @@
+{% load crispy_forms_tags %}
+
+<div class="card-body py-1" id="ds-content-{{ object.id }}">
+  <form action="{% url 'reader-studies:display-set-add-interface' pk=object.pk %}" method="post" class="ds-form" data-hx-get="{% url 'reader-studies:display-set-update' pk=object.pk %}" data-hx-target="#ds-content-{{ object.id }}" data-hx-swap="outerHTML" id="form-{{ object.pk }}">
```
amickan, line 4: Again, not sure if you need the `data-hx-get` attribute here (see the other comment).

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_display_set_update.html
MikeOverkamp-diag: You're right, we only need the `data-hx-target` attribute, that is used here: https://github.com/comic/grand-challenge.org/blob/main/app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_images_list.html#L96. This form gets processed using js :slightly_smiling_face:

app/grandchallenge/components/models.py
```diff
@@ -713,7 +725,10 @@ def _clean_overlay_segments(self):
             )
 
     def _clean_relative_path(self):
-        if self.is_json_kind:
+        if (
+            self.is_file_kind
+            and self.kind in InterfaceKind.interface_type_json()
+        ):
```
amickan, line 731: I was confused by this at first. Maybe it would be clearer if you added a property `is_json_file_kind`?

app/grandchallenge/reader_studies/forms.py
```diff
@@ -500,6 +525,82 @@ def __init__(self, *args, instance=None, **kwargs):
...
+        super().__init__(*args, **kwargs)
+
+        self.fields["user_upload"].queryset = get_objects_for_user(
+            user, "uploads.change_userupload", accept_global_perms=False
+        ).filter(status=UserUpload.StatusChoices.COMPLETED)
+        self.interface = interface
+        self.display_set = display_set
+
+    def save(self):
+        civ = self.display_set.values.get(interface=self.interface)
```
amickan, line 560: Could there ever be more than one CIV of a certain interface in a display set, in some sort of inbetween state when a previous user upload has not finished yet?

app/grandchallenge/reader_studies/urls.py
```diff
@@ -96,6 +103,11 @@
         AddDisplaySetsToReaderStudy.as_view(),
         name="add-displaysets",
     ),
+    path(
+        "<slug>/display-sets/<uuid:pk>/<interface_pk>/images/update/",
```
amickan, line 107: Is `slug` necessary here? You can get the reader study through the display set. I would have intuitively put this with the other display-set urls above.

app/grandchallenge/reader_studies/forms.py
```diff
@@ -479,16 +487,33 @@ def __init__(self, *args, instance=None, **kwargs):
...
+
+                    else:
+                        # Use the field/widget provided by InterfaceFormField,
+                        # which includes proper validation
+                        self.fields[slug] = InterfaceFormField(
+                            kind=civ["kind"],
+                            schema=civ["schema"],
+                            initial=val.value if val else None,
+                            required=False,
+                        ).field
```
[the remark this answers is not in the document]
MikeOverkamp-diag: Yes, I'm currently finishing up the create form for display sets, in which this has been reworked. I could add that work to this PR?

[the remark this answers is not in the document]
MikeOverkamp-diag: Or better yet, I could make a PR to merge that into this PR.

amickan (commented):
LGTM. Just a few comments / questions. 

One small thing I noticed is that if a user leaves the json editor empty, the form can't be saved (as desired), but they don't get an error message. Not super important, but maybe something to fix. 

When testing this locally, I also noticed that the display set list view is still very slow and does a lot of duplicate queries - not related to this PR, but we should probably look into optimizing the view, if possible.

MikeOverkamp-diag:
I think it is getting slower because we initialize all the forms in the list view on load. I'll have a look at that after i finish the create view and after I process the other comments on this one.

app/grandchallenge/reader_studies/forms.py
MikeOverkamp-diag: That shouldn't be possible as we always remove existing civs for an interface before assigning a new one.

app/grandchallenge/reader_studies/forms.py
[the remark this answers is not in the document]
MikeOverkamp-diag: https://github.com/comic/grand-challenge.org/pull/2582

app/grandchallenge/reader_studies/templates/reader_studies/display_set_add_images.html
```diff
@@ -0,0 +1,13 @@
+{% load crispy_forms_tags %}
+
+
+<div class="card-body py-1" id="ds-content-{{ display_set }}">
+    <form action="{% url 'reader-studies:add-files-to-display-set' slug=reader_study pk=display_set interface_pk=interface %}" method="post" class="ds-form" data-hx-target="#ds-content-{{ display_set}}" id="form-{{ display_set }}">
```
amickan, line 5: `slug=reader_study` needs to go since the URL path changed. This does not appear to be covered by the tests either.

app/grandchallenge/reader_studies/templates/reader_studies/select_upload_widget.html
```diff
@@ -0,0 +1,6 @@
+<select name="{{ widget.name }}"{% include "django/forms/widgets/attrs.html" %}>{% for group_name, group_choices, group_index in widget.optgroups %}{% if group_name %}
+  <optgroup label="{{ group_name }}">{% endif %}{% for option in group_choices %}
+  {% include option.template_name with widget=option %}{% endfor %}{% if group_name %}
+  </optgroup>{% endif %}{% endfor %}
+</select>
+<a href="" hx-get="{% url 'reader-studies:add-files-to-display-set' slug=widget.attrs.reader_study_slug pk=widget.attrs.display_set_pk interface_pk=widget.attrs.interface_pk %}" hx-target="#form-{{ widget.attrs.display_set_pk }}" hx-swap="outerHTML">Upload new</a>
```
amickan, line 6: `slug=widget.attrs.reader_study` needs to go here as well.

amickan:
Also it appears that updating a display set with a json file (json type but not stored in database) does not work. The `DisplaySetUpdate` view tries to save the file as a value (see `form_valid()` on L1414 in this PR, can't seem to comment there directly).

app/grandchallenge/components/models.py
```diff
@@ -957,8 +974,7 @@ def _validate_value_only(self):
             )
 
     def _validate_value(self):
-        if self.interface.saved_in_object_store:
-            self._validate_file_only()
+        if self.interface.saved_in_object_store and self.file:
```
jmsmkn, line 977: Why has the `_validate_file_only` gone? Now you can get things that have a file and value.

jmsmkn (changes requested):
The changes in `form_fields.py` need to be reverted/reworked. Think about the API and how the methods are used elsewhere in the codebase.

app/grandchallenge/components/models.py
```diff
@@ -527,6 +527,11 @@ def get_file_mimetypes(cls, *, kind):
             )
         elif kind == InterfaceKind.InterfaceKindChoices.OBJ:
             return ("text/plain", "application/octet-stream")
+        elif kind in InterfaceKind.interface_type_json():
```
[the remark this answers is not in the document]
MikeOverkamp-diag: I will revert the changes to `is_file_kind` and `is_json_kind`, but I tink this change is still needed here, as we want to use a file upload widget for json types that are stored in the dbm so we need to know what mimetypes to allow there.

app/grandchallenge/components/models.py
jmsmkn: This is still unresolved.

MikeOverkamp-diag: Sorry, I somehow thought I had changed that. Resolved now.

jmsmkn:
> @jmsmkn I reverted the changes to the `is_file_kind` and `is_json_kind` methods, but I think we need to keep the changes made to the form_fields and the additional line for allowed mimetypes. We could get rid of the form_fields changes only if we add the interface instance or the `store_in_database` property, as it is a class method that has no knowledge of the instance.

If the default widget is a property of the instance and not of a class why wouldn't the `classmethod` be changed to an instance method?

MikeOverkamp-diag:
Good point! I'll change it.

app/grandchallenge/components/models.py
```diff
@@ -686,6 +636,67 @@ def saved_in_object_store(self):
...
 
+    @property
+    def requires_file(self):
+        return (
+            self.is_file_kind
+            or self.is_json_kind
+            and not self.store_in_database
+        )
+
+    def get_default_field(self):
```
jmsmkn, line 647: `get_default_field` and `get_file_mimetypes` can now be `@properties`: `default_field` and `file_mimetypes`.

app/grandchallenge/reader_studies/forms.py
```diff
@@ -478,41 +480,174 @@ def clean_ground_truth(self):
         return values
 
 
-class DisplaySetForm(Form):
+class SelectUploadWidget(Select):
```
jmsmkn, line 483: Not a form, should be in a `reader_studies/widgets.py` file.

app/grandchallenge/reader_studies/forms.py
```diff
@@ -478,41 +480,174 @@ def clean_ground_truth(self):
...
+    template_name = "reader_studies/select_upload_widget.html"
+
+
+class DisplaySetBaseForm(Form):
     _possible_widgets = {
-        Select,  # Default for ModelChoiceField
+        SelectUploadWidget,
         *InterfaceFormField._possible_widgets,
     }
+    user = None
```
jmsmkn, line 492: Why is user being set as a class attribute?

app/grandchallenge/reader_studies/forms.py
```diff
@@ -478,41 +480,174 @@ def clean_ground_truth(self):
...
+        self.instance = instance
+        self.reader_study = reader_study
+        self.user = user
+        for slug, values in reader_study.values_for_interfaces.items():
+            current_value = None
+            if instance:
+                current_value = instance.values.filter(
+                    interface__slug=slug
+                ).first()
+            interface = ComponentInterface.objects.get(slug=slug)
```
jmsmkn, line 505: How many db queries is this doing? 2 per interface?

app/grandchallenge/reader_studies/forms.py
```diff
@@ -478,41 +480,174 @@ def clean_ground_truth(self):
...
+
+        order = (
+            instance.order if instance else reader_study.next_display_set_order
+        )
+        self.fields["order"] = IntegerField(initial=order)
+
+    def _get_default_field(self, interface, current_value):
+        return InterfaceFormField(
+            instance=interface,
+            initial=current_value.value if current_value else None,
```
jmsmkn, line 527: Shouldn't this check `if current_value is None`? What if the `current_value` is `0`, `[]`, `{}`, etc?

app/grandchallenge/reader_studies/forms.py
```diff
@@ -478,41 +480,174 @@ def clean_ground_truth(self):
...
+            required=False,
+            user=self.user,
+        ).field
+
+    def _get_json_field(self, interface, current_value):
+        return self._get_default_field(interface, current_value)
```
jmsmkn, line 533: Why this intermediary method?

app/grandchallenge/reader_studies/forms.py
```diff
@@ -478,41 +480,174 @@ def clean_ground_truth(self):
...
+                attrs={
+                    "reader_study_slug": self.reader_study.slug,
+                    "display_set_pk": self.instance.pk,
+                    "interface_pk": interface.pk,
+                }
+            ),
+        )
+
+    def _get_file_field(self, interface, values, current_value):
+        return self._get_image_field(interface, values, current_value)
```
jmsmkn, line 550: Why this intermediary method?

app/grandchallenge/reader_studies/forms.py
```diff
@@ -478,41 +480,174 @@ def clean_ground_truth(self):
...
+        super().__init__(
+            *args,
+            instance=None,
+            reader_study=reader_study,
+            user=user,
+            **kwargs,
+        )
+
+    def _get_image_field(self, interface, values, current_value):
+        return self._get_default_field(interface, current_value)
```
jmsmkn, line 575: Why is this different? Why is this not handled in `InterfaceFormField`?

app/grandchallenge/reader_studies/forms.py
```diff
@@ -478,41 +480,174 @@ def clean_ground_truth(self):
...
+            *args,
+            instance=instance,
+            reader_study=instance.reader_study,
+            user=user,
+            **kwargs,
+        )
+
+
+class DisplaySetCreateForm(DisplaySetBaseForm):
+    def __init__(self, *args, instance=None, reader_study, user, **kwargs):
```
jmsmkn, line 565: This is taking `instance` as a kwarg an then discarding it, why overload the `__init__` method at all?

app/grandchallenge/reader_studies/forms.py
```diff
@@ -478,41 +480,174 @@ def clean_ground_truth(self):
...
+                }
+            ),
+        )
+
+    def _get_file_field(self, interface, values, current_value):
+        return self._get_image_field(interface, values, current_value)
+
+
+class DisplaySetUpdateForm(DisplaySetBaseForm):
+    def __init__(self, *args, instance, user, **kwargs):
```
jmsmkn, line 554: This `__init__` method has a different signature to the base, dangerous. Why overload it at all?

app/grandchallenge/reader_studies/forms.py
```diff
@@ -478,41 +480,174 @@ def clean_ground_truth(self):
...
+        self.fields["user_upload"].queryset = get_objects_for_user(
+            user, "uploads.change_userupload", accept_global_perms=False
+        ).filter(status=UserUpload.StatusChoices.COMPLETED)
+        self.interface = interface
+        self.display_set = display_set
+
+    def save(self):
+        civ = self.display_set.values.get(interface=self.interface)
+        user_upload = self.cleaned_data["user_upload"]
+        user_upload.copy_object(to_field=civ.file)
```
jmsmkn, line 601: This cannot be done in the request-response cycle. Needs to be done in an async celery task.

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_display_set_update.html
```diff
@@ -1,10 +1,22 @@
...
+    {% for message in messages %}
+        <div class="alert alert-{{ message.level_tag }}">
+            {{ message }}
+            <button type="button" class="close"
+                    data-dismiss="alert" aria-label="Close">
+                <span aria-hidden="true">&times;</span>
+            </button>
+        </div>
+    {% endfor %}
+  {% endif %}
```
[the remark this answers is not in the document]
MikeOverkamp-diag: Same as above.

app/grandchallenge/reader_studies/urls.py
```diff
@@ -96,6 +109,16 @@
         AddDisplaySetsToReaderStudy.as_view(),
         name="add-displaysets",
     ),
+    path(
+        "<slug>/display-sets/add-single/",
+        AddDisplaySetToReaderStudy.as_view(),
+        name="add-displayset",
+    ),
+    path(
+        "<slug>/display-sets/new/add-interface/",
```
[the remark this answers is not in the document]
MikeOverkamp-diag: The `new` here is to indicate that an interface is added to a display set that is yet to be created. Would it be ok to keep that in and replace all `add`s by `create`s?

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_display_set_detail.html
```diff
@@ -7,6 +7,19 @@
...
+    {% for message in messages %}
+        <div class="alert alert-{{ message.level_tag }}">
+            {{ message }}
+            <button type="button" class="close"
+                    data-dismiss="alert" aria-label="Close">
+                <span aria-hidden="true">&times;</span>
+            </button>
+        </div>
+    {% endfor %}
+{% endif %}
```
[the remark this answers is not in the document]
jmsmkn: Yes, so why not include the partial rather than copy and pasting?

app/grandchallenge/reader_studies/forms.py
MikeOverkamp-diag: current_value is a civ, not a civ value, so it can either be None or a civ instance.

app/grandchallenge/reader_studies/forms.py
MikeOverkamp-diag: Yes, 2 per interface for the update form and 1 per interface for the create form. They're small queries though, so I didn't think it would be a problem.

app/grandchallenge/reader_studies/forms.py
MikeOverkamp-diag: I changed the form and methods a little, which I hope clarifies things a bit and makes a bit more sense as well. Basically for the create view I think we want to go straight to uploading new files, whereas the update view has the option to select an existing civ with a link to upload a new file.

app/grandchallenge/reader_studies/forms.py
MikeOverkamp-diag: The only thing I'm unsure of here is how to let the user know if their file does not pass the civ validation, for instance if it does not match the schema provided for the interface. We cannot update the upload session status, as this is merely a UserUpload instance, which get deleted.

jmsmkn: A notification?

app/grandchallenge/reader_studies/urls.py
```diff
@@ -92,9 +85,39 @@
...
+        DisplaySetInterfacesCreate.as_view(),
+        name="display-set-new-interfaces-create",
+    ),
+    path(
+        "<slug>/display-sets/<uuid:pk>/interface/create/",
+        DisplaySetInterfacesCreate.as_view(),
+        name="display-set-interfaces-create",
+    ),
+    path(
+        "<slug>/display-sets/<uuid:pk>/interfaces/<interface_pk>/files/update/",
```
jmsmkn, line 118: Interface slugs should be used, not pks:

```suggestion
        "<slug>/display-sets/<uuid:pk>/interfaces/<slug:interface_slug>/files/update/",
```

app/grandchallenge/reader_studies/views.py
```diff
@@ -1429,3 +1468,326 @@ def form_valid(self, form):
...
+
+    def get_form_kwargs(self):
+        kwargs = super().get_form_kwargs()
+        kwargs.update(
+            {
+                "pk": self.kwargs.get("pk"),
+                "reader_study": self.reader_study,
+                "interface": self.request.GET.get("interface"),
+                "user": self.request.user,
+                "auto_id": f"id-{random.randint(0, 10**10)}-%s",
```
jmsmkn, line 1587: There is not a structured way to do this?

app/grandchallenge/reader_studies/views.py
```diff
@@ -1429,3 +1468,326 @@ def form_valid(self, form):
...
+                user_upload = data[slug]
+                transaction.on_commit(
+                    partial(
+                        add_file_to_display_set.apply_async,
+                        kwargs={
+                            "user_upload_pk": str(user_upload.pk),
+                            "interface_pk": str(interface.pk),
+                            "display_set_pk": str(ds.pk),
+                        },
+                    )
```
jmsmkn, line 1743: Stick to one way of constructing these, you use a lambda above and a partial here. This construction is the easiest to debug as you get the function signature back when you inspect the callbacks:

```
on_commit(
  add_file_to_display_set.signature(
    kwargs={
      "user_upload_pk": str(user_upload.pk),
      "interface_pk": str(interface.pk),
      "display_set_pk": str(ds.pk),
    }
  ).apply_async
)
```

MikeOverkamp-diag: Ah that should work as well. I changed this one to a partial because lambdas don't work well in for loops (which took me a little while to figure out).

app/grandchallenge/reader_studies/views.py
MikeOverkamp-diag: I couldn't think of a better way unfortunately. We need a unique id to use as the htmx target, but there could be more than one of these forms loaded in the create view

app/grandchallenge/reader_studies/views.py
```diff
@@ -1429,3 +1468,326 @@ def form_valid(self, form):
...
+                )
+            elif interface.is_json_kind:
+                civ = ComponentInterfaceValue.objects.create(
+                    value=data[slug], interface=interface
+                )
```
jmsmkn, line 1748: Shouldn't `create_instance` be used?

app/grandchallenge/reader_studies/views.py
```diff
@@ -1429,3 +1468,326 @@ def form_valid(self, form):
...
+        try:
+            self._process_new_interfaces()
+        except ValidationError as e:
+            errors.update(e.message_dict)
+        return self.return_errors(errors)
+
+    def form_valid(self, form):
+        try:
+            data = self._process_new_interfaces()
+        except ValidationError as e:
```
jmsmkn, line 1779: Why is the validation being done in the view and not in the form? Validation errors should not occur here.

app/grandchallenge/reader_studies/views.py
```diff
@@ -1429,3 +1468,326 @@ def form_valid(self, form):
...
+                        "interface_pk": str(interface.pk),
+                        "display_set_pk": self.kwargs["pk"],
+                    }
+                )
+            )
+            messages.add_message(
+                self.request, messages.SUCCESS, "File copy queued."
+            )
+        else:
+            civ = interface.create_instance(value=value)
```
jmsmkn, line 1629: ValidationError unhandled.

app/grandchallenge/reader_studies/views.py
```diff
@@ -1429,3 +1468,326 @@ def form_valid(self, form):
...
+    @cached_property
+    def display_set(self):
+        return DisplaySet.objects.get(pk=self.kwargs["pk"])
+
+    def get_context_data(self, *args, **kwargs):
+        context = super().get_context_data(*args, **kwargs)
+        context.update(
+            {
+                "display_set": self.kwargs["pk"],
+                "interface": self.kwargs["interface_pk"],
```
jmsmkn, line 1503: `interface_pk` still used here and in other places.

app/grandchallenge/reader_studies/views.py
jmsmkn: A UUID?

app/grandchallenge/reader_studies/views.py
MikeOverkamp-diag: The validation is done un the form, but this view combines multiple forms and collects the errors from each of them.

app/grandchallenge/reader_studies/views.py
MikeOverkamp-diag: Yes, was too quick to commit via the web UI, sorry!

miriam-groeneveld:
It looks like something goes wrong when adding a json file for interface Multiple 2D bounding boxes, the value never gets added and the interface is also not added to the case. There is no error shown.

When uploading a json file for interface Multiple 2D bounding boxes, the help text below the upload widget says: "The following file formats are supported: .m2db". Which should be ".json".

MikeOverkamp-diag:
> It looks like something goes wrong when adding a json file for interface Multiple 2D bounding boxes, the value never gets added and the interface is also not added to the case. There is no error shown. I tested with a json file with invalid content, no error is shown to the user.
> 
> When uploading a json file for interface Multiple 2D bounding boxes, the help text below the upload widget says: "The following file formats are supported: .m2db". Which should be ".json".

Where did you test this and was it on the latest commit?

MikeOverkamp-diag:
Ah, I can reproduce this. The error message gets too long for the notification message field. I'll also check the help text.

MikeOverkamp-diag:
It is now fixed. You get any validation errors for your files via a notification.

app/grandchallenge/reader_studies/tasks.py
```diff
@@ -90,6 +92,62 @@ def add_image_to_display_set(
...
+    civ_pk=None,
+):
+    user_upload = UserUpload.objects.get(pk=user_upload_pk)
+    display_set = DisplaySet.objects.get(pk=display_set_pk)
+    interface = ComponentInterface.objects.get(pk=interface_pk)
+    with transaction.atomic():
+        if civ_pk is None:
+            civ = ComponentInterfaceValue.objects.create(interface=interface)
+        else:
+            civ = ComponentInterfaceValue.objects.get(pk=civ_pk)
```
jmsmkn, line 110: Changing the value of the CIV should never be done. A new CIV should always be created (unless the file, value and image is already none, but then it is strange to create the CIV without a value elsewhere first).

Check the usage of the form too, and other places where you may do this.

app/grandchallenge/reader_studies/tasks.py
```diff
@@ -90,6 +92,62 @@ def add_image_to_display_set(
...
+        user_upload.copy_object(to_field=civ.file)
+        try:
+            civ.full_clean()
+        except ValidationError as e:
+            civ.delete()
```
jmsmkn, line 115: Should be `transaction.rollback()`

app/grandchallenge/reader_studies/tasks.py
```diff
@@ -90,6 +92,62 @@ def add_image_to_display_set(
...
+            transaction.on_commit(
```
jmsmkn, line 116: `on_commit` does not need to be used here, the notification needs to be sent regardless (and especially if the transaction is not committed).

app/grandchallenge/reader_studies/tasks.py
```diff
@@ -90,6 +92,62 @@ def add_image_to_display_set(
...
+):
+    user_upload = UserUpload.objects.get(pk=user_upload_pk)
+    display_set = DisplaySet.objects.get(pk=display_set_pk)
+    interface = ComponentInterface.objects.get(pk=interface_pk)
+    with transaction.atomic():
+        if civ_pk is None:
+            civ = ComponentInterfaceValue.objects.create(interface=interface)
+        else:
+            civ = ComponentInterfaceValue.objects.get(pk=civ_pk)
+        user_upload.copy_object(to_field=civ.file)
```
jmsmkn, line 111: Is there any way to do the validation without copying the file? This will leave files on S3 if validation fails later.

app/grandchallenge/reader_studies/tasks.py
```diff
@@ -90,6 +92,62 @@ def add_image_to_display_set(
...
+                    immutable=True,
+                ).apply_async
+            )
+        else:
+            civ.save()
+            display_set.values.add(civ)
+
+
+@shared_task
+def send_failed_file_copy_notification(
```
jmsmkn, line 133: Why does this need to be a separate task? Seems excessive for what it does.

app/grandchallenge/reader_studies/views.py
```diff
@@ -1090,6 +1087,23 @@ def get_serializer_class(self):
...
+        image = data.pop("image", None)
+        with transaction.atomic():
+            if interface.is_image_kind:
+                # New images can also be added via the rawimageupload endpoint
+                if image:
+                    civ = ComponentInterfaceValue(interface=interface)
+                    civ.image = image
+                    civ.full_clean()
+                    civ.save()
+                    return civ
```
jmsmkn, line 1102: Why not use `create_instance` here?

app/grandchallenge/reader_studies/views.py
```diff
@@ -1090,6 +1087,23 @@ def get_serializer_class(self):
...
+            elif interface.is_json_kind:
+                civ = interface.create_instance(value=value)
+                return civ
```
jmsmkn, line 1105: Anything that falls through this if statement is silently dropped, do we really want that to happen?

jmsmkn (changes requested):

jmsmkn (approved):

PR #2577, opened 2022-07-11
Change HTML email line-height

amickan:
The images in HTML emails were not showing properly in Outlook. The cropping turns out to be a side effect of setting `mso-line-height-rule: exactly` on the body / all elements, which in turn is necessary to control line-height in Outlook. Setting the line-height on the body element to a percentage value above 100% fixes this. 

We could also drop the `mso-line-height-rule: exactly` setting, but then we lose control over the line height of the text in Outlook.


jmsmkn:
Did you test this with Email on Acid? I'm surprised this didn't pick up the problem first time around.

Looking at https://github.com/mailgun/transactional-email-templates/blob/master/templates/action.html it seems that you didn't include all of the mso (microsoft office) CSS directives in #2398 - was there a reason for that?

jmsmkn (approved):
I see, thanks!

amickan:
Brief update, I just created another dummy account with Email on Acid again and ran the tests for the original template as well as the new one. The tests replicate the cropping issue for Outlook with the original template,  and confirm that it's solved with the fix in this PR. I still have no idea why my earlier tests did have that problem though. 

Old template version 
image


New template version 
image

PR #2582, opened 2022-07-13
Add create view for single display sets

MikeOverkamp-diag:


app/grandchallenge/reader_studies/forms.py
```diff
@@ -486,68 +485,92 @@ class SelectUploadWidget(Select):
...
+            *args,
+            instance=instance,
+            reader_study=instance.reader_study,
+            user=user,
+            **kwargs,
+        )
+
+
+class DisplaySetCreateForm(DisplaySetBaseForm):
+    def __init__(self, *args, instance=None, reader_study, **kwargs):
```
amickan, line 567: Why are you not explicitly passing the `user` in your call to super? It's included in the kwargs, so this works either way, but maybe it would be clearer to do this explicitly, just like you do for the UpdateForm?

app/grandchallenge/reader_studies/views.py
```diff
@@ -1549,15 +1553,23 @@ def get_permission_object(self):
 
     @property
     def display_set(self):
-        return DisplaySet.objects.get(pk=self.kwargs["pk"])
+        if self.kwargs.get("pk"):
+            return DisplaySet.objects.get(pk=self.kwargs["pk"])
+
+    def get_template_names(self):
```
amickan, line 1559: Now that you can this method, the `template_name` definition above can go, right?

app/grandchallenge/reader_studies/static/reader_studies/js/display_set_create.js
```diff
@@ -0,0 +1,64 @@
...
+                  data[entry.name] = entry.value
+              });
+              interfaces.push(data)
+          }
+      )
+      formData.new_interfaces = interfaces;
+      $.ajax({
+          type: 'POST',
+          url:  target.attr("action"),
+          data: JSON.stringify(formData),
```
amickan, line 26: I guess you deliberately decided on doing this in js, but you could also add extra variables to the request with the [`hx-vars`](https://htmx.org/attributes/hx-vars/) attribute in the html (which can call a function that computes the values). Since you would still need custom js to compute the values, I'm not sure if it's worth changing, just wanted to mention it since we're trying to move away from custom js code.

app/grandchallenge/reader_studies/forms.py
```diff
@@ -593,21 +616,31 @@ def __init__(self, *args, pk, interface, user, **kwargs):
...
+            )
+            attrs = {
+                "hx-get": reverse_lazy(
+                    "reader-studies:display-set-add-interface",
+                    kwargs={"pk": pk},
+                ),
+                "hx-target": f"#ds-content-{pk}",
+            }
+        else:
+            qs = ComponentInterface.objects.all()
```
amickan, line 632: Right now, a user could try to add 2 generic medical images. The form would not validate - good - but the error they get is not clear: "Correct the errors below", but there is no error displayed further below to clarify that the interfaces have to be unique. I think either we need to narrow down the options here or we need a clearer error message for this case.

amickan: Actually, they don't get an error. The display set gets created without the duplicate interfaces.

app/grandchallenge/reader_studies/static/reader_studies/js/display_set_create.js
MikeOverkamp-diag: Yes we still need js, as we need to commit multiple forms at the same time basically. The form that has all the interfaces that are already in the reader study, and a form for each of the new interfaces added here. I reworked how to do that multiple times before ending up with this. I went this route because I couldn't think of a way to properly handle everything in the backend without losing information in case the form does not validate.

app/grandchallenge/reader_studies/forms.py
MikeOverkamp-diag: Nice catch, I'll have a look at limiting the options. Did you manage to figure out what the error was the first time? If not, I'll try to reproduce it.

app/grandchallenge/reader_studies/static/reader_studies/js/display_set_create.js
amickan: Yes that makes sense. The error handling complicates things.

app/grandchallenge/reader_studies/forms.py
amickan: So, the first time around the error was `Correct the errors below` which you get if the view returns a server error (caused by my breakpoint).

MikeOverkamp-diag: Ah yes, that message shows on all error response statuses. I;m not sure if we should just add it on 400 status? But in that case, what do we do on other status codes?

MikeOverkamp-diag: Ok, I changed the form such that interfaces can only be included once. I also added a delete button for additional interface forms and changed the error message for non 400 status codes.

PR #2583, opened 2022-07-13
Restrict admin submission

jmsmkn:
https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/78

### DIFF
--- a/app/grandchallenge/evaluation/forms.py
+++ b/app/grandchallenge/evaluation/forms.py
@@ -431,12 +431,12 @@ class SubmissionForm(SaveFormInitMixin, forms.ModelForm):
 
 
 class LegacySubmissionForm(SubmissionForm):
-    def __init__(self, *args, challenge, **kwargs):
+    def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
         self.fields[
             "creator"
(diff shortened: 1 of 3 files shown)


app/grandchallenge/evaluation/forms.py
```diff
@@ -431,14 +431,26 @@ class Meta:
...
             "creator"
-        ].queryset = challenge.participants_group.user_set.all().order_by(
+        ].queryset = self._phase.challenge.participants_group.user_set.all().order_by(
             Lower("username")
         )
 
+    def clean(self):
+        if Evaluation.objects.filter(
+            submission__phase__challenge=self._phase.challenge,
+            status=Evaluation.EXECUTING_PREREQUISITES,
```
amickan, line 446: Why only this status, and not also `PENDING`, `PROVISIONING` and `EXECUTING`?

jmsmkn: They don't cause problematic load on the system - it's one evaluation to hundreds of algorithm jobs, and it's the latter we want to stop.

amickan: Ah of course, that makes sense! Thanks for explaining!

PR #2587, opened 2022-07-14
Add linking options to workstation configurations

chrisvanrun:
Part of pitch:
- https://github.com/DIAGNijmegen/rse-roadmap/issues/160

This PR adds a set of boolean values to the workstation configuration:

# In the details
ConfigView

# In the form
ConfigSettings


app/tests/workstation_config_tests/test_serializers.py
```diff
@@ -0,0 +1,62 @@
...
+)
+from tests.factories import WorkstationConfigFactory
+from tests.serializer_helpers import check_if_field_in_serializer
+
+
+@pytest.mark.django_db
+def test_serializer_fields():
+    serializer = WorkstationConfigSerializer(WorkstationConfigFactory())
+
+    check_if_field_in_serializer(
```
jmsmkn, line 14: What is the value of this test? Those "helpers" are super old and I don't think add value, so we shouldn't do anything new with them. This one is testing DRF is generating an object with some fields. We can trust DRF to do that. If you really want to check that a field is in a serializer it is better to inspect `WorkstationConfigSerializer.Meta.fields` directly. Even then, I'm unsure what the value of doing that is beyond making things more of a chore to change, it doesn't even tell you if it is hooked up into a view, but maybe this is what you want.

chrisvanrun: Last time I added a model field but forgot to add it to the serializer's Meta. I thought this test covered that, but now that I trace the functions more closely I see it doesn't. Will add a custom test for that.

chrisvanrun: A bug in pitch-related features requires my attention. As such, I've decided to currently not spent time on adding a full _is-data-field-in-serializer_ tests. I simply removed the 'old' style tests.

jmsmkn (approved):

PR #2594, opened 2022-07-19
Add retries and locks to archive job scheduling

jmsmkn:
Also adds a new variable `ALGORITHMS_MAX_ACTIVE_JOBS` which allows us to independently set the number of active jobs in the queue.


jmsmkn:
That's correct! There is a lock to schedule the jobs so only one process can do that at a time, but there is no specific ordering. We may still have problems with the limited size of the acks-late-2xlarge worker pool, so if a lot of images or containers are uploaded things back up.

PR #2597, opened 2022-07-20
Fix compute cost calculation

amickan:
Sorry that this slipped my mind earlier, but the compute costs calculation was incorrect for jobs that took longer than 1 day.


amickan:
> This isn't an issue given the time limit of 1 hour, but OK.

But it is for the total compute time, which can be multiple days.

PR #2600, opened 2022-07-21
Update and differentiate storage costs

amickan:


app/grandchallenge/challenges/models.py
```diff
@@ -982,9 +985,8 @@ def budget(self):
 
             budget["Docker storage cost"] = round(
                 self.average_algorithm_container_size_in_gb
-                * self.average_number_of_containers_per_team
```
amickan, line 985: Since we're now removing unused container images once a day, I figured that it no longer makes sense to multiply by the `average_number_of_containers` (which is practically 1 now, and not 10 as is the default for this field).

jmsmkn: We still need to keep it - even though the images are being removed from ECR, we keep them on S3 in case we need to go back.

jmsmkn: Also - participants tend to create many algorithms and only upload one container for each, those we never remove (maybe we should - if an algorithm hasn't been used in n months we shut it down).

app/config/settings.py
```diff
@@ -128,8 +128,13 @@
 FLATPAGE_ABOUT_URL = os.environ.get("FLATPAGE_ABOUT_URL", "/about/")
 
 # Costs (in US dollar cents)
-CHALLENGES_STORAGE_COST_CENTS_PER_TB_PER_YEAR = os.environ.get(
-    "CHALLENGES_STORAGE_COST_CENTS_PER_TB_PER_YEAR", 4000
+# based on 0.023 / GB / month S3 standard pricing
+CHALLENGES_S3_STORAGE_COST_CENTS_PER_TB_PER_YEAR = os.environ.get(
+    "CHALLENGES_S3_STORAGE_COST_CENTS_PER_TB_PER_YEAR", 27600
+)
```
jmsmkn, line 134: You will need to cast the environment variables to an `int` as they always return a string, check the usage elsewhere (CHALLENGES_S3_STORAGE_COST_CENTS_PER_TB_PER_YEAR, CHALLENGES_ECR_STORAGE_COST_CENTS_PER_TB_PER_YEAR, CHALLENGES_COMPUTE_COST_CENTS_PER_HOUR)

```suggestion
CHALLENGES_S3_STORAGE_COST_CENTS_PER_TB_PER_YEAR = int(os.environ.get(
    "CHALLENGES_S3_STORAGE_COST_CENTS_PER_TB_PER_YEAR", 27600
))
```

app/grandchallenge/challenges/models.py
amickan: Ok, and 10 is a reasonable default? The Docker storage costs then amount to 6000 Euros for the standard 50 teams, 10 containers per team of around 10GB each. Does that sound about right to you?

jmsmkn: From MIDOG the numbers were 48 and 314, so 50 and 5 may be good defaults.

amickan: Ok, I can change the default value to 5 so that future requests have the updated value. I don't think it's worth writing a management command to update the existing requests. Their not really relevant anymore, and it's so few that I could also do that manually in the admin (13 type 2 requests).

amickan: One other thing: I'm only counting the container storage costs once here, but technically they have to be counted for each phase, right?

jmsmkn: I think that's correct - people should not be creating new images for the subsequent phases.

amickan: True, makes perfect sense, I just got confused between phases and challenge with multiple tasks (sorry)

PR #2603, opened 2022-07-25
Add form and view to create a display set from a job

jmsmkn:
This PR adds a new button for users with the global `add_readerstudy` permission to create a display set from a successful job. The view is idempotent and will redirect the user to a workstation session with the display set. The user requires the `change_readerstudy` and `view_job` object permissions.

image

image

https://github.com/DIAGNijmegen/rse-roadmap/issues/162

### DIFF
--- a/app/grandchallenge/algorithms/forms.py
+++ b/app/grandchallenge/algorithms/forms.py
@@ -593,6 +593,7 @@ class DisplaySetFromJobForm(SaveFormInitMixin, Form):
         queryset=Job.objects.none(),
         disabled=True,
         required=True,
+        widget=HiddenInput(),
     )
 
     def __init__(self, *args, user, job, **kwargs):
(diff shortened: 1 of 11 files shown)


amickan (approved):
LGTM! But we're not checking whether it makes sense to add a particular job to a reader study in terms of the CIVs that are involved. So you can add any job (that you have view permissions for) to any reader study (you are an editor of), regardless of what the other display sets in that reader study look like.

jmsmkn:
> LGTM! But we're not checking whether it makes sense to add a particular job to a reader study in terms of the CIVs that are involved. So you can add any job (that you have view permissions for) to any reader study (you are an editor of), regardless of what the other display sets in that reader study look like.

Yes, depends on the algorithm I/O, would need to check with the CIRRUS crew, we could add a filter on the view content to the form.

miriam-groeneveld:
> This PR adds a new button for users with the global add_readerstudy permission to create a display set from a successful job. The view is idempotent and will redirect the user to a workstation session with the display set. The user requires the change_readerstudy and view_job object permissions.

Why is the `add_readerstudy` permission required for this button? Are `change_readerstudy` and `view_job` permissions not strict enough?

jmsmkn:
Just to limit access for now. `change_readerstudy` is an object permission, so it is more expensive to check if the reader has the change permission for any reader study, add reader study is a global permission and a rough shorthand for that.

PR #2604, opened 2022-07-25
Add slice_plane_indicator field to custom hanging protocol definition

HarmvZ:
Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/146


app/grandchallenge/hanging_protocols/forms.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, *args, **kwargs):
             ButtonHolder(Submit("save", "Save")),
         )
 
-    def clean_json(self):
+    def clean_json(self):  # noqa: C901
```
jmsmkn, line 50: Could you split the method out into sub-validation methods? Make the aim of each method clear so the future dev can see what is happening from `clean_json` (which should do say 4 method calls).

app/grandchallenge/hanging_protocols/forms.py
```diff
@@ -85,6 +85,18 @@ def clean_json(self):
                     field="json",
                 )
 
+        for viewport in [v for v in value if "slice_plane_indicator" in v]:
```
jmsmkn, line 88: The list comprehension in a for loop is confusing, could you move the filter to the main loop?

app/grandchallenge/hanging_protocols/models.py
```diff
@@ -110,6 +110,10 @@ class Orientation(models.TextChoices):
                 "minimum": 0,
                 "maximum": 1,
             },
+            "slice_plane_indicator": {
+                "type": "string",
+                "enum": [port.lower() for port in ImagePort.labels],
```
jmsmkn, line 115: We discussed using labels in #2539, might be worth creating a separate text choice for that as the labels become part of the external API otherwise. See https://github.com/comic/grand-challenge.org/pull/2539/commits/6f58d5398676dde4bc5f9996ce6ca168c53be262 for what changed.

jmsmkn (approved):

app/grandchallenge/hanging_protocols/forms.py
```diff
@@ -107,15 +109,17 @@ def clean_json(self):
         ]
         for viewport in json:
             if "parent_id" in viewport:
-                errors.append(
-                    *self._validate_parent_id(viewport, viewport_names)
-                )
+                errors = [
+                    *errors,
+                    *self._validate_parent_id(viewport, viewport_names),
+                ]
```
jmsmkn, line 115: Nit: cleaner to use `errors.extend(self._validate_parent_id(viewport, viewport_names))` etc.

jmsmkn: Or rather than creating, returning and appending errors, can't you call `self.add_error(error=error, field="json")` each time?

jmsmkn (approved):

PR #2612, opened 2022-07-26
Add a cooldown period before users can post

jmsmkn:
https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/87


amickan (approved):
Nice! Hopefully that'll keep the spamming at bay a little.

PR #2615, opened 2022-07-28
Add form to import algorithm from remote instance

jmsmkn:
This PR adds a form to import an algorithm from a remote instance to the current instance:

https://user-images.githubusercontent.com/12661555/182618661-3f421cc2-2130-45df-b42d-8d5430a5a2e7.mp4

The user needs to provide:
- The remote algorithm URL
- Their API token for the remote instance
- The name of the storage bucket that the source algorithm is stored on

On the local instance a new algorithm will be created with the same pk and slug. The metadata that is serialised over DRF will be set. The logo is added. The interfaces are added, creating them on the fly if necessary. Checks are done to ensure that the interfaces are consistent. The user must be an editor of the remote algorithm. The remote instance must be available to the local instance.

On save, a celery task is launched to import the latest ready algorithm image. The local instance must have list and get permissions on the prefix for the remote image.


app/grandchallenge/algorithms/forms.py
```diff
@@ -622,3 +639,309 @@ class AlgorithmPublishForm(AlgorithmPublishValidation, ModelForm):
...
+            )
+        }
+
+        self.new_interfaces = []
+        for slug, remote_interface in remote_interfaces.items():
+            try:
+                self._validate_existing_interface(
+                    slug=slug, remote_interface=remote_interface
+                )
+            except ObjectDoesNotExist:
```
amickan, line 827: Don't you also need to catch the possible `ValidationError` that `_validate_existing_interface()` could throw?

amickan: Ah no, you won't that to error out and not create a new interface in that case.

app/grandchallenge/algorithms/forms.py
```diff
@@ -622,3 +639,309 @@ class AlgorithmPublishForm(AlgorithmPublishValidation, ModelForm):
...
+            # Force the given slug to be used
+            interface.instance.slug = interface.initial_data["slug"]
+
+            # Set the store in database correctly, for most interfaces this is
+            # False, then switch it if the super kind is different
+            if interface.initial_data[
+                "super_kind"
+            ] != interface.get_super_kind(obj=interface.instance):
+                interface.instance.store_in_database = True
+            interface.instance.save()
```
amickan, line 885: This makes sense, but if we serialized `store_in_database` on the `ComponentInterfaceSerializer` this would be much more straightforward. Is there a reason why we are not adding that to the serializer?

amickan (approved):

app/grandchallenge/algorithms/forms.py
jmsmkn: Too many details to expose to external consumers, it's an implementation detail.

PR #2616, opened 2022-07-29
Pdf report for display set answers

amickan:
This adds a view that generates a PDF report for a given `display_set` and `user` in a `reader_study`. All non-image and non-annotation type answers are rendered. For now, the PDF is very plain, answers are just printed as is together with some general information. View is restricted to reader study editors.

image

Rendering headers and footers as described in the [docs](https://xhtml2pdf.readthedocs.io/en/latest/format_html.html#example-with-2-static-frames-and-1-content-frame) appears to be [broken](https://github.com/xhtml2pdf/xhtml2pdf/issues/564), so for now no page numbers etc.

Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/163


app/grandchallenge/reader_studies/models.py
```diff
@@ -926,6 +926,24 @@ class AnswerType(models.TextChoices):
...
+            AnswerType.MULTIPLE_POINTS,
+            AnswerType.POLYGON,
+            AnswerType.MULTIPLE_POLYGONS,
+            AnswerType.MASK,
+            AnswerType.LINE,
+            AnswerType.MULTIPLE_LINES,
+            AnswerType.ANGLE,
+            AnswerType.MULTIPLE_ANGLES,
+        ]
+
```
amickan, line 946: This change was necessary to use this in the answer filter for the PDF. Not sure why this needed to be a property of the Question model. Making it a staticmethod on the `AnswerType` enum seemed a little more logical to me. This is also how we do it for interface types, for example.

app/grandchallenge/reader_studies/models.py
```diff
@@ -926,6 +926,24 @@ class AnswerType(models.TextChoices):
     ANGLE = "ANGL", "Angle"
     MULTIPLE_ANGLES = "MANG", "Multiple angles"
 
+    @staticmethod
+    def annotation_types():
```
jmsmkn, line 930: As this is now a method could you rename it `get_annotation_types`?

app/grandchallenge/reader_studies/views.py
```diff
@@ -1438,3 +1441,71 @@ def form_valid(self, form):
...
+            answer.question.question_text: answer.answer_text
+            for answer in self.display_set.answers.select_related(
+                "creator", "question", "answer_image"
+            )
+            .filter(
+                Q(creator=self.user),
+                Q(is_ground_truth=False),
+                ~Q(question__answer_type__in=AnswerType.annotation_types()),
+                Q(answer_image__isnull=True),
+            )
```
jmsmkn, line 1501: Better not to use Q objects:

```suggestion
            .filter(
                creator=self.user,
                is_ground_truth=False,
                answer_image__isnull=True,
            )
            .exclude(
                question__answer_type__in=AnswerType.annotation_types(),
            )
```

jmsmkn (commented):
Great start!

app/grandchallenge/reader_studies/models.py
```diff
@@ -1143,28 +1161,10 @@ def clean(self):
...
-            self.AnswerType.MULTIPLE_LINES,
-            self.AnswerType.ANGLE,
-            self.AnswerType.MULTIPLE_ANGLES,
-        ]
-
     @property
     def allow_null_types(self):
         return [
-            *self.annotation_types,
+            *AnswerType.annotation_types(),
```
jmsmkn, line 1167: You can keep the reference to `self` here and on L1139.

app/grandchallenge/reader_studies/models.py
```diff
@@ -1165,28 +1183,10 @@ def clean(self):
...
-            self.AnswerType.MULTIPLE_LINES,
-            self.AnswerType.ANGLE,
-            self.AnswerType.MULTIPLE_ANGLES,
-        ]
-
     @property
     def allow_null_types(self):
         return [
-            *self.annotation_types,
+            *self.AnswerType.get_annotation_types(),
```
jmsmkn, line 1189: Does this really work? I would have thought this would result in an attribute error? `self.AnswerType.get_annotation_types()` should be `self.get_annotation_types()`. Same on L1161

amickan: `self.get_annotatuion_types()` doesn't work because `get_annotation_types` is no longer a property or method on the `Question` model

amickan: And yes it works, `get_annotation_types()` returns a list, and here we unpack that list. Or why did you think we would get an attribute error here?

jmsmkn: Ah yes, confusing construction but I see why it works now.

jmsmkn (approved):

PR #2617, opened 2022-07-29
Create and use PkToHyperlinkedRelatedField for session field on feedback serializer

HarmvZ:
Specialized implementation of the HyperlinkedRelatedField that maintains the hyperlink representation when reading serialized data but allows us to use only the pk when writing to the field. 

This is useful for the feedback endpoint because we only know the session pk when writing to that endpoint and cannot reverse it into the full URL (at least not without hardcoding it).

### DIFF
--- a/app/grandchallenge/core/serializer_fields.py
+++ b/app/grandchallenge/core/serializer_fields.py
@@ -1,0 +1,11 @@
+from rest_framework import serializers
+
+
+class PkToHyperlinkedRelatedField(serializers.HyperlinkedRelatedField):
+    """
+    A HyperlinkedRelatedField that allows you to use the primary key of the
+    model when writing, but returns the URL of the object when reading.
+    """
+
+    def to_internal_value(self, data):
+        return self.get_queryset().get(pk=data)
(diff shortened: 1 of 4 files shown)


jmsmkn:
I really don't like this, sorry, it's really nothing for the client to add the URL to the request. That adds some confirmation that you're linking the correct object as PKs can be duplicated across objects, includes the host now that we have multiple instances, and allows discovery of the API.

HarmvZ:
I disagree. I feel like adding the URL in cirrus is duplicating the source of truth and it just makes cirrus more fragile. If the URL changes in GC it will break the request in cirrus. Since the pk is a UUID it will always uniquely point to a single object. Discovery of the API is still possible since the field still serializes to a URL when reading.

jmsmkn:
That's ok! We don't have to agree. There are different trade-offs when designing an API. The `HyperLinkedRelatedField` does have those problems, but, I think they're acceptable for the benefits it brings. This mixing is not the solution, if a developer uses this proposed field and adds an update method then the client would need to parse out the `pk` from the hyperlink anyway to post it back. If you didn't want to deal with hyperlinks then a `PkRelatedField` should be used instead, but, I also think we should stick with `HyperLinkedRelatedField` for consistency across the API.

PR #2623, opened 2022-08-03
AnswerType specific answer rendering for PDF report

amickan:
For the reader study PDF report:
- render all options for choice type answers and underline the selected option(s) (circling selected options is not possible because the `border-radius` css property is not supported by `xhtml2pdf`)
- render either a green checkmark or a red cross for boolean answers

Only render answers with a value, leave out empty answers.

image

Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/163

### DIFF
--- a/app/grandchallenge/reader_studies/models.py
+++ b/app/grandchallenge/reader_studies/models.py
@@ -927,6 +927,14 @@ class AnswerType(models.TextChoices):
     MULTIPLE_ANGLES = "MANG", "Multiple angles"
 
     @staticmethod
+    def get_choice_types():
+        return [
+            AnswerType.CHOICE,
+            AnswerType.MULTIPLE_CHOICE,
+            AnswerType.MULTIPLE_CHOICE_DROPDOWN,
+        ]
+
+    @staticmethod
     def get_annotation_types():
         return [
             AnswerType.BOUNDING_BOX_2D,
--- a/app/grandchallenge/reader_studies/models.py
+++ b/app/grandchallenge/reader_studies/models.py
@@ -1379,7 +1379,7 @@ class Answer(UUIDModel):
                     "title", flat=True
                 ).all()
             ]
-            return ", ".join(list)
+            return format_html(", ".join(list))
         elif self.question.answer_type == Question.AnswerType.BOOL:
             return (
                 format_html('<span class="success">✔</span>')
--- a/app/grandchallenge/reader_studies/models.py
+++ b/app/grandchallenge/reader_studies/models.py
@@ -11,7 +11,6 @@ from django.db.models import Avg, Count, Q, Sum
 from django.db.models.signals import post_delete
 from django.dispatch import receiver
 from django.utils.functional import cached_property
-from django.utils.html import format_html
 from django_extensions.db.models import TitleSlugDescriptionModel
 from guardian.shortcuts import assign_perm, get_objects_for_group, remove_perm
 from jsonschema import RefResolutionError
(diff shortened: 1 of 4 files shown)


app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_display_set_pdf_report.html
```diff
@@ -47,7 +49,7 @@ <h2 style="margin-top:50px;">Answers</h2>
                 {% for question, answer in answers.items %}
                     <tr>
                         <td class="left"><b>{{ question }}:</b></td>
-                        <td>{{ answer }}</td>
+                        <td>{{ answer|safe }}</td>
```
jmsmkn, line 52: It is always best to avoid the use of `safe`. Why is this needed here? Could you change the implementation to ensure this doesn't need to be used (either in a template filter or by calling it directly from python).

app/grandchallenge/reader_studies/models.py
```diff
@@ -1363,6 +1364,31 @@ def validate(  # noqa: C901
...
+        ):
+            list = [
+                format_html("<u>{}</u>", option)
+                if option in self.answer_text
+                else format_html("{}", option)
+                for option in self.question.options.values_list(
+                    "title", flat=True
+                ).all()
+            ]
+            return format_html(", ".join(list))
```
jmsmkn, line 1382: This has the same effect - the first argument to `format_html` is marked safe implicitly, and you're including user input there. Can't this whole function be done in the template itself?

amickan: Ok, but `format_html` also passes the elements through a conditional escape, so I thought that would be fine? I can move this to the template itself, of course. It just felt cleaner / easier to keep track of (especially as we potentially add more for annotations later) to do it in python.

jmsmkn: The first argument to `format_html` is not escaped, the rest are. Never, ever, send user input to the first argument of `format_html`, that should only be a raw string that we are in control of and are 100% sure is safe. https://docs.djangoproject.com/en/4.0/ref/utils/#django.utils.html.format_html

Just noticed that in the docs there is `format_html_join`, which could help you here https://docs.djangoproject.com/en/4.0/ref/utils/#django.utils.html.format_html_join

amickan: Oh sorry, I see the difference now! `format_html_join` looks like a good solution. Thanks!

amickan: `format_html_join` is not an option, it only works if all elements are formatted in the same way, which is not the case here (some are underlined, some are not). I will do it in the template. 

But just for my understanding: the elements in `list` are all individually constructed with `format_html`, so the user input, i.e. `option` in my code is escaped before it's marked safe (option is an argument to format_html). So then I'm not sure I understand why marking the joined string as safe is an issue.  Sorry for going on about this, but I want to make sure I fully understand this.

jmsmkn: Maybe in this specific case it's OK, but that would need to be covered by a lot of tests as you're relying on future developers not to touch `list` (btw, you should never use `list` as a variable name as it shadows the built-in `list` function). See the substitutions tests.

Honestly, what you're trying to do is too complex for a small snippet, this is much better solved in a template where all of these cases are handled for you.

amickan: Ok, thanks for explaining. I'll move this to the template.

jmsmkn (approved):

jmsmkn (approved):

PR #2627, opened 2022-08-04
 Fix for showing limited number of github repos

### LINKED ISSUE #2299: `add-repo` view only displays limited number of repos
The dropdown for selecting a repo to link to an algorithm from GitHub only displays at most 30 results. If a user has access to more than 30 repos, the repo that they want to link might not be there.

miriam-groeneveld:
Fix issue where only the first 30 github app installations and repos were retrieved.
Default parameter per_page is increased from 30 (default) to 100(max).
Request is repeated until all items are retrieved.

Closes #2299

### DIFF
--- a/app/grandchallenge/algorithms/views.py
+++ b/app/grandchallenge/algorithms/views.py
@@ -1,4 +1,5 @@
 import logging
+import math
 from datetime import timedelta
 from typing import Dict
 
@@ -753,9 +754,7 @@ class JobUpdate(LoginRequiredMixin, ObjectPermissionRequiredMixin, UpdateView):
 
 
 class DisplaySetFromJobCreate(
-    LoginRequiredMixin,
-    ObjectPermissionRequiredMixin,
-    FormView,
+    LoginRequiredMixin, ObjectPermissionRequiredMixin, FormView
 ):
     form_class = DisplaySetFromJobForm
     permission_required = "algorithms.view_job"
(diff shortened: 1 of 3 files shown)


jmsmkn (changes requested):
Unfortunately a recursive get is not possible in a request-response cycle, makes it really easy for a user to DOS the site by keeping all of the workers occupied making requests to GitHub. We can only really do one  or two calls to an external API per request as even one call occupies a worker for up to 5 seconds - an age when our response time target is 100ms. To fully solve this we will need to add some filtering with the calls hooked in to an auto complete endpoint.

Increasing `per_page` to 100 is a good idea and maybe make a separate PR with just that change?

miriam-groeneveld:
> Unfortunately a recursive get is not possible in a request-response cycle, makes it really easy for a user to DOS the site by keeping all of the workers occupied making requests to GitHub. We can only really do one or two calls to an external API per request as even one call occupies a worker for up to 5 seconds - an age when our response time target is 100ms. To fully solve this we will need to add some filtering with the calls hooked in to an auto complete endpoint.
> 
> Increasing `per_page` to 100 is a good idea and maybe make a separate PR with just that change?

Ah, that's not something I considered. What about caching the result? The github endpoint does not seem to have a way to filter or search, so only increasing the default will not solve the issue. DIAGNijmegen has more than 500 repos already.

jmsmkn:
Then rather than using autocomplete we would need to implement pagination and select.

miriam-groeneveld:
Simple increase of default for now; addressed in #2634

PR #2633, opened 2022-08-08
Do case sensitive username match

amickan:
Addresses https://sentry.io/organizations/grand-challenge/issues/3475841542/?project=303639&query=is%3Aunresolved+is%3Afor_review+assigned_or_suggested%3A%5Bme%2C+none%5D&sort=inbox


jmsmkn:
Usernames are case insensitive. The sentry problem occurs as there are old users that have the same username when checked case insensitively. This happened as there was a bug that did not validate the username for case insensitive clashes when the user created their account. The validation was fixed a couple of years ago. The solution for this is to update one of the usernames when a clash occurs.

PR #2637, opened 2022-08-09
HP json schema for specialized views

amickan:


app/grandchallenge/hanging_protocols/models.py
```diff
@@ -37,7 +37,7 @@ class Orientation(models.TextChoices):
 
 HANGING_PROTOCOL_SCHEMA = {
     "definitions": {},
-    "$schema": "http://json-schema.org/draft-06/schema#",
+    "$schema": "https://json-schema.org/draft/2019-09/schema#",
```
jmsmkn, line 40: What was the reason for upgrading? Is this version compatible with both the python library we use for validation and the json widget?

amickan: I upgraded because the `if ...else` validators were only introduceed in draft 7 (version 2019-09 is not necessary after all). Both `jsonschema`  and `jsoneditor` (which uses `ajv` for validation) have full support for draft 7.

jmsmkn (approved):

PR #2640, opened 2022-08-10
Add validation for CIV value saving

amickan:
This adds checks when saving a CIV to ensure:

- The value of image can only be changed from None to something, never from something to something else
- The value of value can only be changed from None or its default to something, never from something to something else
- The value of file can only be changed from "" or None to something, never from something to something else
- One, and only one, of image, value or file should be set

With these checks in place, the `DisplaySetUpdate` view needed a small change as well.

### DIFF
--- a/app/grandchallenge/components/models.py
+++ b/app/grandchallenge/components/models.py
@@ -962,8 +962,43 @@ class ComponentInterfaceValue(models.Model):
         else:
             return f"Component Interface Value {self.pk} for {self.interface}"
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._value_orig = self.value
+        self._image_orig = self.image
+        self._file_orig = self.file
+
+    def save(self, *args, **kwargs):
+        if (
+            (
+                self._value_orig not in (None, self.interface.default_value)
+                and self.value is not None
+                and self._value_orig != self.value
+            )
+            or (self._image_orig and self._image_orig != self.image)
+            or (
+                self._file_orig.name not in (None, "")
+                and self._file_orig != self.file
+            )
+        ):
+            raise ValidationError(
+                "You cannot change the value, file or image of an existing CIV. "
+                "Please create a new CIV instead."
+            )
+        super().save(*args, **kwargs)
+
     def clean(self):
         super().clean()
+        attributes = [
+            attribute
+            for attribute in [self.value, self.image, self.file.name]
+            if attribute is not None
+            if attribute != ""
+        ]
+        if len(attributes) > 1:
+            raise ValidationError(
+                "Only one of image, value and file can be defined."
+            )
 
         if self.interface.is_image_kind:
             self._validate_image_only()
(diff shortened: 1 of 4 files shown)


app/grandchallenge/components/models.py
```diff
@@ -962,8 +962,43 @@ def __str__(self):
...
+        super().__init__(*args, **kwargs)
+        self._value_orig = self.value
+        self._image_orig = self.image
+        self._file_orig = self.file
+
+    def save(self, *args, **kwargs):
+        if (
+            (
+                self._value_orig not in (None, self.interface.default_value)
+                and self.value is not None
```
MikeOverkamp-diag, line 975: Wouldn't this allow you to still set the value to `None`?

app/grandchallenge/reader_studies/views.py
```diff
@@ -1393,13 +1393,11 @@ def form_valid(self, form):
...
-                else:
-                    val = ComponentInterfaceValue.objects.create(
-                        interface=ci, value=civ
-                    )
-                    instance.values.add(val)
+                    assigned_civs.append(current_value)
+                val = ComponentInterfaceValue.objects.create(
+                    interface=ci, value=civ
+                )
+                instance.values.add(val)
```
MikeOverkamp-diag, line 1400: I have also made some changes to this logic in #2575, but I'll be sure to check those if this PR gets merged first.

MikeOverkamp-diag (changes requested):

app/grandchallenge/components/models.py
amickan: Yes, that's correct. That is currently necessary for  `civ_value_to_file()` to keep functioning.

app/grandchallenge/reader_studies/views.py
amickan: Oh right! I see that we actually made the same changes. So that should be easy to integrate if this goes through. I thought I had checked your PR, but clearly missed those changes. Would it make more sense to merge this PR into your branch?

app/grandchallenge/components/models.py
MikeOverkamp-diag: Ah that makes sense then :slightly_smiling_face:

MikeOverkamp-diag (approved):

app/grandchallenge/reader_studies/views.py
MikeOverkamp-diag: Let's keep them separate and I'll fix any conflicts if they arise.

app/tests/reader_studies_tests/test_views.py
```diff
@@ -422,11 +422,11 @@ def test_display_set_update(client):
...
     assert not ds1.values.filter(pk=civ_img.pk).exists()
     assert ds1.values.filter(pk=civ_img_new.pk).exists()
-    civ_str.refresh_from_db()
-    assert civ_str.value == "new-title"
+    assert not ds1.values.filter(pk=civ_str.pk).exists()
+    assert ds1.values.filter(interface__kind="STR").get().value == "new-title"
 
-    # A new ds should have been created for civ_img
-    assert DisplaySet.objects.count() == 3
+    # Two new ds should have been created, one for civ_img and one for civ_str
```
jmsmkn, line 428: Is that correct? Shouldn't the old one be removed?

MikeOverkamp-diag: I changed this so that we only create new display sets for image type civs in #2575

PR #2642, opened 2022-08-11
Add autocomplete for interfaces in reader study forms

MikeOverkamp-diag:


app/grandchallenge/reader_studies/forms.py
```diff
@@ -492,6 +494,8 @@ def clean_ground_truth(self):
 
 class DisplaySetCreateForm(Form):
     _possible_widgets = {
+        SelectUploadWidget,
+        autocomplete.ModelSelect2,
```
jmsmkn, line 498: How does this form produce those widgets?

app/grandchallenge/reader_studies/forms.py
```diff
@@ -628,6 +632,7 @@ def save(self):
 class DisplaySetInterfacesCreateForm(Form):
     _possible_widgets = {
         *InterfaceFormField._possible_widgets,
+        autocomplete.ModelSelect2,
```
jmsmkn, line 635:
```suggestion
        autocomplete.ModelSelect2,
        Select,
```

app/grandchallenge/reader_studies/forms.py
MikeOverkamp-diag: Good point, it doesn't.

app/grandchallenge/reader_studies/forms.py
```diff
@@ -643,28 +647,55 @@ def __init__(self, *args, pk, interface, reader_study, user, **kwargs):
...
+                    "disabled": selected_interface is not None,
+                }
+            )
+        widget_kwargs["attrs"] = attrs
+
         self.fields["interface"] = ModelChoiceField(
             initial=selected_interface,
             queryset=qs,
-            widget=Select(attrs=attrs),
+            widget=widget(**widget_kwargs),
```
amickan, line 698: I'm not sure why, but trying this out, it doesn't give my any options in the select on the update and add-single-ds views. It's just an empty dropdown (even though the queryset that is passed here contains a bunch of options). So something is going wrong, but I haven't been able to find the issue.

MikeOverkamp-diag: Having a look now.

MikeOverkamp-diag: This is now fixed. It broke with the last commit.

jmsmkn: Doesn't that reveal an implementation issue as that form does not produce those widgets?

MikeOverkamp-diag: The issue is that the `DisplaySetInterfacesCreateForm` is only ever loaded within either the update or create form, but the media for `_possible_widgets` is only loaded for the view's form_class, which is are the `DisplaySetCreateForm` for the create view and the `DisplaySetUpdateForm` for the update view. So in a sense the possible widgets for the `DisplaySetInterfacesCreateForm` are also widgets for the both the create and the update form. I think it would be clearer to add them as `*DisplaySetInterfacesCreateForm._possible_widgets` though.

jmsmkn: Those are different forms, it’s kind of hacky to fix a problem in the view by trying to fix it in the form attributes.

MikeOverkamp-diag: Alternatively, instead of using the `form_class` attribute on the view to determine which media to include, we add a new attribute `included_form_classes` which is a list of all forms that could possibly be loaded in the view. Does that sounds better?

jmsmkn: Yes, then you have the possible forms for the view which provide the possible widgets.

app/grandchallenge/reader_studies/forms.py
```diff
@@ -563,7 +561,7 @@ def _get_default_field(self, *, interface, current_value):
 class DisplaySetUpdateForm(DisplaySetCreateForm):
     _possible_widgets = {
         SelectUploadWidget,
-        *DisplaySetCreateForm._possible_widgets,
+        *InterfaceFormField._possible_widgets,
```
jmsmkn, line 564: Why is this being replaced? `DisplaySetUpdateForm` subclasses `DisplaySetCreateForm` so can include its widgets.

app/grandchallenge/reader_studies/forms.py
```diff
@@ -643,28 +651,55 @@ def __init__(self, *args, pk, interface, reader_study, user, **kwargs):
...
-                "hx-target": f"#ds-content-{pk}",
-            }
+            attrs.update(
+                {
+                    "hx-get": reverse_lazy(
+                        "reader-studies:display-set-interfaces-create",
+                        kwargs={"pk": pk, "slug": reader_study.slug},
+                    ),
+                    "hx-target": f"#ds-content-{pk}",
+                    "hx-trigger": "interfaceSelected",
```
amickan, line 681: A shame that the simple `change` trigger doesn't work here, but nice work around!

amickan (approved):
Looks like everything is working now, and the htmx stuff looks good to me as well!

PR #2645, opened 2022-08-15
Fix json file input on try out algorithm

amickan:
Addresses https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/101

### DIFF
--- a/app/grandchallenge/algorithms/views.py
+++ b/app/grandchallenge/algorithms/views.py
@@ -497,7 +497,7 @@ class AlgorithmExperimentCreate(
                     civ = ComponentInterfaceValue.objects.create(interface=ci)
                     civs.append(civ)
                     upload_pks[civ.pk] = create_upload(value)
-            elif ci.is_file_kind:
+            elif ci.requires_file:
                 civ = ComponentInterfaceValue.objects.create(interface=ci)
                 value.copy_object(to_field=civ.file)
                 civ.full_clean()
--- a/app/grandchallenge/components/serializers.py
+++ b/app/grandchallenge/components/serializers.py
@@ -132,7 +132,7 @@ class ComponentInterfaceValuePostSerializer(serializers.ModelSerializer):
             )
             instance.full_clean()
 
-        if interface.is_file_kind:
+        if interface.requires_file:
             if not attrs.get("user_upload"):
                 raise serializers.ValidationError(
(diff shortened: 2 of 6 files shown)


jmsmkn (approved):

jmsmkn:
@amickan I'm looking at uses of `is_file_kind` - does the same problem exist in `AlgorithmExperimentCreate`, `update_archive_item_update_kwargs`, `ArchiveEditArchiveItem`, or `ComponentInterfaceValuePostSerializer`? Is the test `test_algorithm_multiple_inputs` correct? I find this API a little confusing.

amickan:
> @amickan I'm looking at uses of `is_file_kind` - does the same problem exist in `AlgorithmExperimentCreate`, `update_archive_item_update_kwargs`, `ArchiveEditArchiveItem`, or `ComponentInterfaceValuePostSerializer`? Is the test `test_algorithm_multiple_inputs` correct? I find this API a little confusing.

Good point. I will double check and make changes as necessary.

amickan:
I have checked all places where we use `is_file_kind` and replaces them with `requires_file` when necessary. I also added some tests. 

One thing that is not covered yet is running an algorithm job with a json input (not stored in db) through the API. That will require some changes in gcapi, maybe @MikeOverkamp-diag can have a look?

jmsmkn:
> I have checked all places where we use `is_file_kind` and replaces them with `requires_file` when necessary. I also added some tests.
> 
> One thing that is not covered yet is running an algorithm job with a json input (not stored in db) through the API. That will require some changes in gcapi, maybe @MikeOverkamp-diag can have a look?

Is that being done in https://github.com/DIAGNijmegen/rse-gcapi/pull/136?

MikeOverkamp-diag:
> > I have checked all places where we use `is_file_kind` and replaces them with `requires_file` when necessary. I also added some tests.
> > One thing that is not covered yet is running an algorithm job with a json input (not stored in db) through the API. That will require some changes in gcapi, maybe @MikeOverkamp-diag can have a look?
> 
> Is that being done in [DIAGNijmegen/rse-gcapi#136](https://github.com/DIAGNijmegen/rse-gcapi/pull/136)?

No that's just the updated api for reader studies. I can have a look at the algorithm stuff this week.

jmsmkn (approved):

PR #2648, opened 2022-08-16
Add management command to populate overlay_segments

MikeOverkamp-diag:


app/grandchallenge/components/tasks.py
```diff
@@ -979,3 +981,33 @@ def civ_value_to_file(*, civ_pk):
...
+@shared_task(**settings.CELERY_TASK_DECORATOR_KWARGS["acks-late-2xlarge"])
+def validate_voxel_values(*, civ_pk):
+    civ = get_model_instance(
+        pk=civ_pk,
+        app_label="components",
+        model_name="componentinterfacevalue",
+    )
+
+    if civ.image.segments is None:
+        image_file = civ.image.files.first()
```
jmsmkn, line 995: What about image that are `mhd` and `zraw` pairs? You need both.

TIFF and DZI need to be excluded.

app/grandchallenge/components/management/commands/add_overlay_segments.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class Command(BaseCommand):
+    def add_arguments(self, parser):
+        parser.add_argument("json", type=str)
+
+    def handle(self, *args, **options):
+        json_str = options["json"]
+        mapping = json.loads(json_str)
+        for slug in mapping:
```
jmsmkn, line 20: Would be better to have the slug as a required argument and just do this one interface at a time.

app/grandchallenge/components/management/commands/add_overlay_segments.py
```diff
@@ -0,0 +1,46 @@
...
+                )
+            interface.overlay_segments = [
+                {
+                    "voxel_value": int(voxel_value),
+                    "name": mapping[slug][voxel_value],
+                    "visible": True,
+                }
+                for voxel_value in mapping[slug]
+            ]
+            interface.save()
```
jmsmkn, line 35: No validation of `overlay_segments`?

app/grandchallenge/components/tasks.py
MikeOverkamp-diag: Aren't these guaranteed to be mha and not mhd/zraw though?

jmsmkn: No, both exist.

app/tests/components_tests/test_management_commands.py
```diff
@@ -0,0 +1,42 @@
...
+    ci = ComponentInterfaceFactory(
+        title="foo", kind=InterfaceKind.InterfaceKindChoices.SEGMENTATION
+    )
+    ComponentInterfaceValueFactory(interface=ci, image=im)
+
+    im.segments = [2]
+    im.save()
+
+    # This raises a ValidationError, but unfortunetaly that gest swallowed in
+    # the context manager.
```
jmsmkn, line 28: I don't think this is correct. The callback is `apply_async`, which [when invoked](https://docs.celeryq.dev/en/stable/reference/celery.app.task.html#celery.app.task.Task.apply_async) returns an [AsyncResult](https://docs.celeryq.dev/en/stable/reference/celery.result.html#celery.result.AsyncResult) instance, even when using eager task execution. This has a `status` and `traceback` attributes that can be used to determine whether the `ValidationError` was raised e.g.

```python
with capture_om_commit_callbacks() as callbacks:
    ...

for callback in callbacks:
    maybe_task = callback()
    if isinstance(maybe_task, AsyncResult):
        # Do something with maybe_task.traceback
```

MikeOverkamp-diag: Nice, I'll look into that. Thank you!

app/grandchallenge/components/management/commands/add_overlay_segments.py
```diff
@@ -0,0 +1,50 @@
...
+        parser.add_argument("json", type=str)
+
+    def handle(self, *args, **options):
+        slug = options["slug"]
+        json_str = options["json"]
+        mapping = json.loads(json_str)
+        try:
+            interface = ComponentInterface.objects.get(slug=slug)
+        except ComponentInterface.DoesNotExist:
+            self.stdout.error(f"Could not find interface with slug: {slug}.")
```
jmsmkn, line 27: Does this exit? Would raising `CommandError` be better here? https://docs.djangoproject.com/en/4.1/howto/custom-management-commands/#command-exceptions

app/grandchallenge/components/tasks.py
```diff
@@ -979,3 +981,40 @@ def civ_value_to_file(*, civ_pk):
         )
         civ.value = None
         civ.save()
+
+
+def _get_sitk_image(image_file):
+    with NamedTemporaryFile() as tmp:
```
jmsmkn, line 987: The `zraw` file needs to have the correct name as it is referenced by the header. The filename should be preserved, you can ensure it's safe to write by using `safe_join`

app/grandchallenge/components/tasks.py
```diff
@@ -979,3 +981,40 @@ def civ_value_to_file(*, civ_pk):
...
+        pk=civ_pk,
+        app_label="components",
+        model_name="componentinterfacevalue",
+    )
+
+    if civ.image.segments is None:
+        for file in civ.image.files.all():
+            if not file.image_type == ImageFile.IMAGE_TYPE_MHD:
+                continue
+            simple_itk_image = _get_sitk_image(file)
```
jmsmkn, line 1014: This still won't work for mhd/(z)raw pairs. You can enable this mode in the grand challenge settings (I think) to try it out. You need to create a temporary directory in this method, then download the files, then load the mhd.

MikeOverkamp-diag: The only thing I seem to able to find is that we use the mhd image builder from panimg, which always saves the output as mha. I can't find any setting that changes this, though I do see an admin filter for zraw/mhd.

jmsmkn: Hmm, yeah, I guess we removed it when migrating to panimg. If you can monkey patch panimg and change the extension here from `.mha` to `mhd` it should do it https://github.com/DIAGNijmegen/rse-panimg/blob/9e46f1f5c4bcf892b26f58fddc140808a86bdcd8/panimg/models.py#L367

jmsmkn (approved):

PR #2650, opened 2022-08-19
Allow file type inputs for job create in api

MikeOverkamp-diag:


app/grandchallenge/components/tasks.py
```diff
@@ -961,6 +963,50 @@ def add_images_to_component_interface_value(
...
+        Notification.send(
+            type=NotificationType.NotificationTypeChoices.FILE_COPY_STATUS,
+            actor=user_upload.creator,
+            message=f"File for interface {civ.interface.title} failed validation.",
+            target=target,
+            description=(
+                f"File for interface {civ.interface.title} failed validation:\n{error}."
+            ),
+        )
+
```
amickan, line 1008: For archive items, I think we delete the user upload object after its file contents have been moved to the CIV field.

app/grandchallenge/components/tasks.py
```diff
@@ -961,6 +963,50 @@ def add_images_to_component_interface_value(
...
+        else:
+            user_upload.copy_object(to_field=civ.file)
+            civ.save()
+
+    if error is not None:
+        Notification.send(
+            type=NotificationType.NotificationTypeChoices.FILE_COPY_STATUS,
+            actor=user_upload.creator,
+            message=f"File for interface {civ.interface.title} failed validation.",
+            target=target,
```
amickan, line 1003: I don't think you need a target for this type of notification, do you? The receivr is the upload session creator and th only thing that is printed is the description (i.e. the message).

MikeOverkamp-diag: Target is not nullable: https://github.com/comic/grand-challenge.org/blob/main/app/grandchallenge/notifications/models.py#L94. But I think the notification links to the target right? So this might actually be a nice way to give some context.

amickan:
I think some tests within GC (next to the tests you will add in gcapi) would be good. Otherwise, this looks good to me!

app/grandchallenge/components/tasks.py
amickan: Oh right, forgot about that. For the user, the link is not visible (unless you print it explicitly), but for us admins the context is there indeed.

app/grandchallenge/components/tasks.py
```diff
@@ -961,6 +963,51 @@ def add_images_to_component_interface_value(
...
+        model_name=target_model,
+    )
+    error = None
+    with transaction.atomic():
+        try:
+            civ.validate_user_upload(user_upload)
+            civ.full_clean()
+        except ValidationError as e:
+            transaction.set_rollback(True)
+            error = str(e)
```
amickan, line 993: Maybe I'm misunderstanding this, but if this fails, the civ remains in the database without a value or file attached, because it was already created and saved in the previous step, right? Or what does the rollback do in this case? Is that what we want?

MikeOverkamp-diag: This is the way James suggested to handle this in https://github.com/comic/grand-challenge.org/pull/2575, so I went with that here as well.

amickan: You mean in the `add_file_to_display_set`? But there you create the CIV within the atomic block, so that's different, right? Or do you mean a different place?

MikeOverkamp-diag: Ah you're right!

amickan (approved):

PR #2652, opened 2022-08-24
Add fixture for non-image algorithm

MikeOverkamp-diag:


scripts/algorithm_evaluation_fixtures.py
```diff
@@ -53,6 +59,10 @@ def _get_inputs():
     )
 
 
+def _get_non_image_inputs():
+    return ComponentInterface.objects.filter(slug__in=["predictions-csv-file"])
+
```
amickan, line 64: Don't we want a test with json type interface that is not stored in the DB as well?

amickan (approved):

jmsmkn:
Why was this done? Is it used anywhere or only for convenience?

The suffix incrementing has made working with this command annoying - this is often run multiple times when developing the algorithm evaluation feature and you get several algorithms with the same title but different inputs, unclear which to select when submitting to the demo algorithm.

MikeOverkamp-diag:
This was added to be able to test https://github.com/DIAGNijmegen/rse-gcapi/pull/137 properly.

jmsmkn:
So is it used in the gcapi tests?

MikeOverkamp-diag:
Yes.

PR #2653, opened 2022-08-29
Include display sets in civ serving permission check

MikeOverkamp-diag:


jmsmkn:
Could you add a test for this?

jmsmkn (approved):

PR #2659, opened 2022-09-09
Do not fetch image to determine original civ image

MikeOverkamp-diag:
Fetching the full object results in n+1 problems for any query that somehow includes image type civs.


MikeOverkamp-diag:
Already covered by this test: https://github.com/comic/grand-challenge.org/blob/main/app/tests/components_tests/test_models.py#L305

jmsmkn (approved):

PR #2665, opened 2022-09-15
Set accept_global_perms to False by default

MikeOverkamp-diag:


.pre-commit-config.yaml
```diff
@@ -12,18 +12,18 @@ repos:
...
         args: [--py38-plus]
   - repo: https://github.com/ambv/black
-    rev: 22.3.0
+    rev: 22.8.0
     hooks:
       - id: black
         language: python
   - repo: https://gitlab.com/pycqa/flake8
-    rev: 4.0.1
+    rev: 3.9.2
```
jmsmkn, line 26: Why did this get downgraded?

app/grandchallenge/core/shortcuts.py
```diff
@@ -0,0 +1,16 @@
+from functools import partial
```
jmsmkn, line 1: Would be clearer if this was `guardian.py` and also included `PermissionListMixin`, and `ObjectPermissionRequiredMixin` with `accept_global_perms = False` set explicitly, just in case they change it upstream.

jmsmkn (approved):

PR #2671, opened 2022-09-23
Add Cornerstone3D viewer for GPU enabled clientside viewing of volumes

HarmvZ:
Roadmap issue: https://github.com/DIAGNijmegen/rse-roadmap/issues/174

Adds a view that loads an image with [Cornerstone3D](https://www.cornerstonejs.org/). It supports basic image manipulation options like flipping x/y, invert and changing orientation. RGB images are currently not supported. Vtk.js (used internally for rendering the volumes) currently only supports either uint8 or float32 datatypes for voxel data. Therefore any image with a datatype larger than 8 bit is converted to float32. This conversion is done in the background by using a worker (`typed_array_to_float32.js`). 

https://user-images.githubusercontent.com/2001094/191953166-185beb7e-3afd-44ff-81fd-0e14ac1ff156.mp4

See main repository for more information: https://github.com/DIAGNijmegen/rse-gc-cornerstone-poc


pkcakeout:
Why is the image sideways? I can imagine two valid orientations, but sideways seems incorrect. That is the MeVisLab test image you are using, right? The 3D brain scan of a 40x40x40 resolution (or something like that).

HarmvZ:
It's indeed the mevislab probandT1 image. It is not a demo of the flipping. This is how it is loaded by default. Other images I have tried were not flipped, so I think it has something to do with one of the special headers in the file that are not parsed correctly by cornerstone or this custom loader.

pkcakeout:
I think you are facing a problem that Chris faced before... he will post an answer.

PR #2676, opened 2022-10-04
Retrieve workstation name and url via api

miriam-groeneveld:
Add api call to get some basic workstation information. Will add test tomorrow.

### DIFF
--- a/app/grandchallenge/workstations/serializers.py
+++ b/app/grandchallenge/workstations/serializers.py
@@ -1,8 +1,8 @@
-from rest_framework.fields import CharField
+from rest_framework.fields import CharField, URLField
 from rest_framework.relations import HyperlinkedRelatedField
 from rest_framework.serializers import ModelSerializer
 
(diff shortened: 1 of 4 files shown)


app/grandchallenge/workstations/serializers.py
```diff
@@ -29,3 +29,11 @@ def __init__(self, *args, **kwargs):
...
                 creator=user
             ).all()
+
+
+class WorkstationSerializer(ModelSerializer):
+    url = URLField(source="get_absolute_url", read_only=True)
+
+    class Meta:
+        model = Workstation
+        fields = ("pk", "title", "slug", "url")
```
jmsmkn, line 39: The pk should not be serialized.

jmsmkn: Actually, hmm, it's UUID, so OK.

jmsmkn (approved):

PR #2678, opened 2022-10-17
Disallow IsAuthenticated DRF permission class

amickan:


jmsmkn:
LGTM, could you add tests for the two permission classes?

jmsmkn (approved):

PR #2681, opened 2022-10-24
Documentation updates

miriam-groeneveld:
Adds all the possible elements to the annotation examples.
Small type


jmsmkn (commented):
Looks good, but could you clarify in the documentation which attributes are mandatory and which are optional for each structure?

miriam-groeneveld:
You want me to add the schema for each object?

jmsmkn:
Not the schema, but right now if I were a user I would assume that I would need to write all of the keys, that is not the case as only a couple are required, the rest optional.

app/grandchallenge/components/models.py
```diff
@@ -150,111 +150,149 @@ def interface_type_json():
         * Chart
 
         Example json for 2D bounding box annotation
+            required: ``type``, ``corners``, ``version``
```
jmsmkn, line 153: For consistency:

```suggestion
            required: "type", "corners", "version"
```

app/grandchallenge/components/models.py
```diff
@@ -284,88 +327,113 @@ def interface_type_json():
...
             {
+                "name": "Some angles",
                 "type": "Multiple angles",
                 "angles": [
                     {
+                        "name": "First angle",
                         "lines": [[[110, 135, 0.5], [60, 165, 0.5]],[[70, 25, 0.5], [85, 65, 0.5]]],
-                        "name": "First angle"
+                        "probability": 0.82
+
```
jmsmkn, line 426:
```suggestion
```

jmsmkn (approved):

PR #2704, opened 2022-11-21
Add slice_plane_indicator_fade_ms setting to custom hanging protocols

chrisvanrun:
Extension of this CIRRUS PR:
- https://github.com/DIAGNijmegen/rse-cirrus-core/pull/2191

Once that is merged this PR can be reviewed/merged as well.

### DIFF
--- a/app/grandchallenge/hanging_protocols/models.py
+++ b/app/grandchallenge/hanging_protocols/models.py
@@ -118,6 +118,10 @@ HANGING_PROTOCOL_SCHEMA = {
                 "type": "string",
                 "enum": ViewportNames.values,
             },
+            "slice_plane_indicator_fade_ms": {
+                "type": "number",
+                "minimum": 0,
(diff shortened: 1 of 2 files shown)


jmsmkn:
Do we really need such fine grained configuration? We have far too many options already, this does not seem important enough to warrant yet another option.

chrisvanrun:
It relates to the swivel tool. For mpMRI-related reader studies, it is imperative that it fades out, for anything swivel related however it is needed to stay to be an important cue for which plane the viewer is currently in. The only way I see to marry these two needs is to create a configuration option.

It could have been a boolean switch, but I thought we might as well make it a fine configuration option.

chrisvanrun:
Pre-commits should be fixed via: https://github.com/comic/grand-challenge.org/pull/2715

jmsmkn (approved):

PR #2710, opened 2022-11-30
Challenge cost overview

amickan:
https://user-images.githubusercontent.com/30069334/204804687-20556346-1738-433d-ad1c-9f5587fbe0e4.mp4


amickan:
I can't reproduce the test failures locally and looking at the trace back, I don't understand why the pk in the reverse_kwargs gets set to '' somewhere down the line in get_view_for_user when I do pass the pk to that function initially. Very strange.

amickan:
> I can't reproduce the test failures locally and looking at the trace back, I don't understand why the pk in the reverse_kwargs gets set to '' somewhere down the line in get_view_for_user when I do pass the pk to that function initially. Very strange.

I spent more time trying to figure this test failure out, but I'm stumped. I cannot figure out what's going on. Locally all tests pass.

jmsmkn:
I can reproduce it on my machine with `docker compose run --rm celery_worker_evaluation pytest tests/challenges_tests/test_views.py -n 1 --reuse-db`

jmsmkn:
Note that the problem is coming from `{% url ... %}` the template, not the lookup in the test. It's likely that you have some cache data set, which isn't set or updated in the test.

Doing this using the cache is more difficult than setting an attribute on the model. Probably best to change the design.

amickan:
> Note that the problem is coming from `{% url ... %}` the template, not the lookup in the test. It's likely that you have some cache data set, which isn't set or updated in the test.
> 
> Doing this using the cache is more difficult than setting an attribute on the model. Probably best to change the design.

Ohhh! Of course. Right, moving it to the model also works, true, I'll try that.

amickan:
I moved the challenge compute and docker costs to the model and out of the cache. The cache still contains the monthly challenge cost break down, since that's not challenge specific, of course, and the usual phase statistics (left those in the cache to not mess with the previous implementation, but could also move those stats to the model).

app/config/settings.py
```diff
@@ -1144,6 +1144,10 @@
         "task": "grandchallenge.challenges.tasks.update_challenge_results_cache",
         "schedule": crontab(minute="*/5"),
     },
+    "update_challenge_cost_statistics": {
+        "task": "grandchallenge.challenges.tasks.update_challenge_cost_statistics",
+        "schedule": crontab(minute=0, hour=0),
```
jmsmkn, line 1149: Could you put this in time order for the dictionary? There is another task running at 00:00, so maybe move this to the next free slot at 05:00.

app/grandchallenge/challenges/models.py
```diff
@@ -361,6 +361,8 @@ class Challenge(ChallengeBase):
         default="",
         help_text="This email will be listed as the contact email for the challenge and will be visible to all users of Grand Challenge.",
     )
+    accumulated_compute_cost = models.FloatField(default=0, blank=True)
+    accumulated_docker_storage_cost = models.FloatField(default=0, blank=True)
```
jmsmkn, line 365: The name of the field should tell us if this is dollars or cents. Shouldn't we be storing things as Cents (int)?

app/grandchallenge/challenges/urls.py
```diff
@@ -56,4 +61,29 @@
...
+    path(
+        "costs/costs-per-year",
+        ChallengeCostsPerYearView.as_view(),
+        name="costs-per-year",
+    ),
+    path(
+        "costs/year-row",
+        YearCostsRow.as_view(),
+        name="year-cost-row",
+    ),
```
jmsmkn, line 88: 3x Missing trailing slashes.

app/grandchallenge/challenges/views.py
```diff
@@ -389,3 +399,158 @@ def form_valid(self, form):
         response = HttpResponse()
         response["HX-Refresh"] = "true"
         return response
+
+
+class ChallengeCostViewPermissionMixin(UserPassesTestMixin):
+    def test_func(self):
+        return self.request.user.is_staff or self.request.user.has_perm(
```
jmsmkn, line 406: I think we should just work with the `challenges.view_challengerequest` permission, then the Mixin can be removed too.

app/grandchallenge/evaluation/tasks.py
```diff
@@ -562,19 +564,80 @@ def get_average_job_duration_for_phase(phase):
...
+    end_date = datetime.datetime.now()
+    delta = relativedelta.relativedelta(end_date, start_date)
+    monthly_costs = {}
+    algorithms_submitted_per_month = {}
+    for year in [start_date.year, start_date.year + delta.years]:
+        if not year == start_date.year + delta.years:
+            months = range(1, 13)
+        else:
+            months = range(1, 2 + delta.months)
+        for month in months:
```
jmsmkn, line 577: This is going to do a lot of queries, one per month per challenge. It should be possible with an aggregate instead, no?

app/grandchallenge/evaluation/tasks.py
```diff
@@ -562,19 +564,80 @@ def get_average_job_duration_for_phase(phase):
...
+            except (TypeError, KeyError):
+                monthly_costs[year] = {}
+                algorithms_submitted_per_month[year] = {}
+                monthly_costs[year][
+                    job_start_date.strftime("%B")
+                ] = compute_cost
+                algorithms_submitted_per_month[year][
+                    job_start_date.strftime("%B")
+                ] = submitted_algorithms
+
```
jmsmkn, line 622: This method is now doing a lot more than `get_average_job_duration_for_phase`, probably best to split it out or rename.

app/grandchallenge/evaluation/tasks.py
```diff
@@ -589,13 +652,17 @@ def update_phase_statistics():
...
             "average_duration", None
         )
-        accumulated_algorithm_job_run_time = avg_duration.get(
+        accumulated_algorithm_job_run_time = duration_dict.get(
             "total_duration", None
         )
+        monthly_costs = duration_dict.get("monthly_costs", None)
+        algorithms_submitted_per_month = duration_dict.get(
+            "algorithms_submitted_per_month", None
+        )
```
jmsmkn, line 665: When would the duration dict ever not contain those fields? The `None` is anyway unnecessary, I think the `get` is too.

app/grandchallenge/pages/views.py
```diff
@@ -165,8 +166,3 @@ def get_context_data(self, **kwargs):
         )
 
         return context
-
-    def test_func(self):
-        return self.request.user.is_staff or self.request.user.has_perm(
```
jmsmkn, line 170: Missed this before but I think the same about only using `challenges.view_challengerequest`

app/grandchallenge/evaluation/tasks.py
amickan: It's not clear to me yet how this would work with an aggregate, but I'm looking into it.

app/tests/challenges_tests/test_tasks.py
```diff
@@ -137,3 +147,182 @@ def test_challenge_request_budget_calculation(challenge_request):
...
+                    "May": [],
+                    "June": [],
+                    "July": [],
+                    "August": [],
+                    "September": [],
+                    "October": [
+                        str(s3.algorithm_image.algorithm.pk),
+                        str(s4.algorithm_image.algorithm.pk),
+                        str(s5.algorithm_image.algorithm.pk),
+                    ],
```
jmsmkn, line 310: Do we need to keep the pks of each algorithm submitted? The object saved in the cache would be huge. Won't a count do instead?

jmsmkn (approved):

PR #2711, opened 2022-12-01
Fix json file validation

amickan:
The schema validation in `validate_user_upload` didn't work because the file contents weren't parsed correctly. This function is only called in two places (ds file updating and adding a file to a civ). The tests for both scenarios were incomplete and didn't catch the issue. I updated the tests accordingly.


jmsmkn:
Does this fix #2713

amickan:
Yes, it does! Hadn't seen the issue, sorry!

jmsmkn (approved):

PR #2716, opened 2022-12-12
Flexible image widget

amickan:
Implements a flexible image widget as described here: https://github.com/DIAGNijmegen/rse-roadmap/issues/190

Users can choose between uploading a new image and selecting an existing image. They can search for an existing image by name and pk and only get to choose from images that they have view permissions for. 

image

image

This widget is used in archives, algorithms and reader studies.


app/grandchallenge/reader_studies/views.py
```diff
@@ -1430,50 +1432,16 @@ def form_valid(self, form):
...
-        for assigned in assigned_civs:
-            if assigned.interface.is_image_kind and (
-                not instance.reader_study.display_sets.exclude(pk=instance.pk)
-                .filter(values=assigned)
-                .exists()
-            ):
-                ds = DisplaySet.objects.create(
-                    reader_study=instance.reader_study
-                )
-                ds.values.add(assigned)
```
amickan, line 1475: I'm not sure if this is necessary or even desired behaviour. We don't do this for archive items either. The only reason I can see is that a user loses view rights for an image when its no longer connected to the ds (and also not to an archive or job etc). But then for consistency, this should also happen for archive items, shouldn't it? Or maybe we could consider not removing the view permission for the original uploader of the image (image origin --> imageuploadsession.creator always keep the view permission)?

app/grandchallenge/algorithms/views.py
```diff
@@ -495,10 +496,25 @@ def create_upload(image_files):
...
+                            civ.full_clean()
+                            civ.save()
+                        civs.append(civ)
+                    elif widget == WidgetChoices.IMAGE_UPLOAD:
+                        # create civ without image, image will be added when import completes
+                        civ = ComponentInterfaceValue.objects.create(
+                            interface=ci
+                        )
+                        civs.append(civ)
+                        upload_pks[civ.pk] = create_upload(value)
```
jmsmkn, line 517: We probably should have an `else` block here that raises a runtime exception to remind future developers to extend the block when changes are made, you never know.

app/grandchallenge/archives/views.py
```diff
@@ -427,9 +428,16 @@ def create_upload(image_files):
             ci = ComponentInterface.objects.get(slug=slug)
 
             if ci.is_image_kind:
-                upload_session = create_upload(value)
+                widget = form.data[f"WidgetChoice-{ci.slug}"]
+                if widget == WidgetChoices.IMAGE_SEARCH.name:
+                    image = value
+                    upload_session = None
+                elif widget == WidgetChoices.IMAGE_UPLOAD.name:
```
jmsmkn, line 435: And here.

app/grandchallenge/cases/static/js/cases/ignore_enter_button.js
```diff
@@ -0,0 +1,5 @@
+$(document).keypress(function (e) {
+    if (e.which == 13) {
+        return false;
+    }
+});
```
jmsmkn, line 5: Why this method?

app/grandchallenge/cases/templates/cases/image-search-widget.html
```diff
@@ -0,0 +1,30 @@
...
+                id="selected-image-{{ widget.name }}">
+            <option value="">No image selected</option>
+        </select>
+    </div>
+</div>
+
+<div id="search-{{ widget.name }}"></div>
+
+{% if widget.attrs.help_text and widget.attrs.help_text != "" %}
+    <small id="hint_{{ widget.attrs.id }}" class="form-text text-muted">{{ widget.attrs.help_text|safe }}</small>
```
jmsmkn, line 29: Help text is not necessarily safe.

app/grandchallenge/cases/templates/cases/image-search-widget.html
```diff
@@ -0,0 +1,30 @@
...
+{% load static %}
+
+<div class="form-group mb-1"
+     id="search-image-{{ widget.name }}"
+     hx-get="{% url 'cases:image-search' %}"
+     hx-trigger="load once, keyup[target.value.length > 1] delay:500ms from:#query-{{ widget.name }}, click[keyCode==8] delay:500ms from:#query-{{ widget.name }}"
+     hx-target="#selected-image-{{ widget.name }}"
+     hx-swap="outerHTML"
+     hx-vals='{"interface": "{{ widget.name }}"}'
+     hx-include="[id='query-{{ widget.name }}']"
```
jmsmkn, line 11: I'm still curious about the need for `hx-vals` (both here and other places) and `hx-include`, but this comes from my lack of knowledge about htmx. Are they really necessary? Why aren't these things being sent along with a form?

app/grandchallenge/cases/views.py
```diff
@@ -159,3 +176,81 @@ def get_object(self):
...
+
+class ImageWidgetSelectView(LoginRequiredMixin, View):
+    def get(self, request, *args, **kwargs):
+        query_params = request.GET.copy()
+        interface = query_params.pop("interface", None)[0]
+        widget_name = query_params.pop("WidgetChoice-" + interface, None)[0]
+        help_text = query_params.pop("help_text", None)
+
+        if widget_name:
+            if widget_name == WidgetChoices.IMAGE_SEARCH.name:
```
jmsmkn, line 189: Shouldn't be a need for nested if statements here, `not widget_name` should be caught by L223.

app/grandchallenge/cases/templates/cases/image_search_result_select.html
```diff
@@ -0,0 +1,13 @@
+<select id="selected-image-{{ interface }}"
+        name="{{ interface }}"
+        class="custom-select"
+>
+    {% if object_list %}
```
jmsmkn, line 5: Is there a limit to the length of the object list?

app/grandchallenge/cases/static/js/cases/ignore_enter_button.js
amickan: The search field is nested in the form that has a submit button. Pressing enter after typing a search term (quite a common thing to do I think) will trigger the entire form to submit. So I thought it would make most sense to disable enter for the entire form. The search happens automatically when they type.

app/grandchallenge/cases/templates/cases/image-search-widget.html
amickan: `hx-include` is used to include additional element values in the request (from elements outside of the form that the htmx is configured for). Values from elements within the form are sent along automatically. Looking at this case here again, it's actually redudant! It's also the only place in the gc repo where we use this at the moment. I'll remove it. I believe in the gc-private repo it was necessary everytime I used it, but I'll double check that assumption when I have a moment.

`hx-vals` is used to add extra parameters to a request that are not coming from other elements in the DOM. I think the only other way of sending the interface slug along would be to add a hidden input element to the form, but I don't see how that would be any better than passing it through hx-vals in this particular case where we control the value (it's not user generated).

amickan: Ah so for the `hx-include`: it turns out to be necessary because I'm not using a form element, but a div here. Using a form is not possible, because then we'd have a form nested in a form, which leads to issues with the final form submission.

app/grandchallenge/algorithms/views.py
```diff
@@ -495,10 +496,29 @@ def create_upload(image_files):
...
-                    civs.append(civ)
-                    upload_pks[civ.pk] = create_upload(value)
+                    widget = form.data[f"WidgetChoice-{ci.slug}"]
+                    if widget == WidgetChoices.IMAGE_SEARCH:
+                        (
+                            civ,
+                            created,
+                        ) = ComponentInterfaceValue.objects.get_or_create(
+                            interface=ci, image=value
+                        )
```
amickan, line 506: I'm wondering now if the `get_or_create` calls will cause issues. There might be images for which there are multiple CIVs for a given interface already (the generic interfaces mostly, but maybe also others), right? So I guess this should be out inside a try .. except statement and if multiple civs exist we choose the latest?

amickan: I've addressed this in the latest commit

app/grandchallenge/algorithms/views.py
```diff
@@ -494,10 +496,35 @@ def create_upload(image_files):
...
+                    widget = form.data[f"WidgetChoice-{ci.slug}"]
+                    if widget == WidgetChoices.IMAGE_SEARCH:
+                        try:
+                            (
+                                civ,
+                                created,
+                            ) = ComponentInterfaceValue.objects.get_or_create(
+                                interface=ci, image=value
+                            )
+                        except MultipleObjectsReturned:
```
jmsmkn, line 508: I don't think we should handle `MultipleObjectsReturned` - this _shouldn't_ happen, and if it does we should be made aware of it from sentry so that we can correct the entries in the db.

app/grandchallenge/algorithms/views.py
```diff
@@ -466,14 +468,14 @@ def get_context_data(self, *args, **kwargs):
...
         return context
 
-    def form_valid(self, form):
-        def create_upload(image_files):
-            upload_session = RawImageUploadSession.objects.create(
-                creator=self.request.user
-            )
-            upload_session.user_uploads.set(image_files)
-            return upload_session.pk
+    def create_upload(self, image_files):
```
jmsmkn, line 471:
```suggestion
    @staticmethod
    def create_upload(image_files):
```

app/grandchallenge/archives/tasks.py
```diff
@@ -95,13 +96,28 @@ def update_archive_item_update_kwargs(
...
-                civ.image = image
-                civ.full_clean()
+                try:
+                    (
+                        civ,
+                        created,
+                    ) = ComponentInterfaceValue.objects.get_or_create(
+                        interface=interface, image=image
+                    )
+                except MultipleObjectsReturned:
```
jmsmkn, line 107: See previous.

app/grandchallenge/cases/templates/cases/image-search-widget.html
jmsmkn: > I think the only other way of sending the interface slug along would be to add a hidden input element to the form

But what about the case above with the help text? We allow the user to customise that in some places. If we think some values are safe now it is very difficult to keep track of all those usages, IMO it is better to be conservative and use hidden form elements.

app/grandchallenge/cases/utils.py
```diff
@@ -52,3 +53,8 @@ def get_sitk_image(*, image):
             raise
 
     return sitk_image
+
+
+class WidgetChoices(TextChoices):
```
jmsmkn, line 58: Why in `utils.py` and not `widgets.py`?

app/grandchallenge/cases/widgets.py
```diff
@@ -0,0 +1,114 @@
...
+)
+from django.forms.widgets import ChoiceWidget
+from django.utils.datastructures import MultiValueDictKeyError
+
+from grandchallenge.cases.models import Image
+from grandchallenge.cases.utils import WidgetChoices
+from grandchallenge.uploads.widgets import UserUploadMultipleWidget
+
+
+class ImageSearchMixin(ChoiceWidget):
```
jmsmkn, line 17: Why make this a Mixin and not just put it all in `ImageSearchWidget`?

app/grandchallenge/cases/widgets.py
```diff
@@ -0,0 +1,114 @@
...
+            try:
+                # this happens if the data comes from the DS update form
+                value = data[f"WidgetChoice-{name}"]
+            except MultiValueDictKeyError:
+                value = None
+        if value:
+            if Image.objects.filter(pk=value).exists():
+                return [value, None]
+            else:
+                return [None, [value]]
```
jmsmkn, line 85: Are these covered by tests because TBH I have no idea what is happening here.

app/grandchallenge/reader_studies/tasks.py
```diff
@@ -79,17 +84,19 @@ def add_image_to_display_set(
...
+            )
+            try:
+                civ.full_clean()
+            except ValidationError as e:
+                # this should only happen for new uploads
+                upload_session.status = RawImageUploadSession.FAILURE
+                upload_session.error_message = e.message
+                upload_session.save()
         else:
             civ.save()
```
jmsmkn, line 98: Why is saving the CIV reqired here now? It shouldn't be now that you're using create.

app/grandchallenge/reader_studies/views.py
jmsmkn: I think that it's fine to remove.

app/grandchallenge/reader_studies/views.py
```diff
@@ -1488,9 +1457,68 @@ def form_valid(self, form):
...
+            ):
+                assigned_civs.append(current_value)
+                try:
+                    (
+                        civ,
+                        created,
+                    ) = ComponentInterfaceValue.objects.get_or_create(
+                        interface=ci, image=new_value
+                    )
+                except MultipleObjectsReturned:
```
jmsmkn, line 1484: See previous comment.

app/grandchallenge/reader_studies/forms.py
```diff
@@ -571,8 +574,14 @@ class DisplaySetUpdateForm(DisplaySetCreateForm):
     }
 
     def _get_image_field(self, *, interface, values, current_value):
-        return self._get_select_upload_widget_field(
-            interface=interface, values=values, current_value=current_value
+        return FlexibleImageField(
+            image_queryset=get_objects_for_user(self.user, "cases.view_image"),
+            upload_queryset=get_objects_for_user(
+                self.user, "uploads.change_userupload"
+            ),
```
jmsmkn, line 581: Do these filters have tests?

app/grandchallenge/reader_studies/views.py
```diff
@@ -1617,22 +1634,39 @@ def get_context_data(self, *args, **kwargs):
...
-                    immutable=True,
+            if isinstance(value, Image):
+                try:
+                    (
+                        civ,
+                        created,
+                    ) = ComponentInterfaceValue.objects.get_or_create(
+                        interface=interface, image=value
+                    )
+                except MultipleObjectsReturned:
```
jmsmkn, line 1645: See previous comment.

app/grandchallenge/reader_studies/views.py
```diff
@@ -1776,19 +1815,36 @@ def create_display_set(self, data):
...
-                        immutable=True,
+                if isinstance(data[slug], Image):
+                    try:
+                        (
+                            civ,
+                            created,
+                        ) = ComponentInterfaceValue.objects.get_or_create(
+                            interface=interface, image=data[slug]
+                        )
+                    except MultipleObjectsReturned:
```
jmsmkn, line 1826: See previous comment.

app/grandchallenge/uploads/templates/uploads/widget.html
```diff
@@ -44,4 +44,7 @@
                 {% include "django/forms/widgets/attrs.html" %}
         >
     {% endif %}
+    {% if widget.attrs.help_text %}
+        <small id="hint_{{ widget.attrs.id }}" class="form-text text-muted">{{ widget.attrs.help_text|safe }}</small>
```
jmsmkn, line 48: Help text is not safe.

app/grandchallenge/cases/widgets.py
amickan: Good point, left over from an earlier implementation where that made sense, just forgot to refactor.

app/grandchallenge/cases/views.py
```diff
@@ -159,3 +176,81 @@ def get_object(self):
...
+
+
+class ImageSearchView(LoginRequiredMixin, ListView):
+    template_name = "cases/image_search_result_select.html"
+    search_fields = ["pk", "name"]
+    model = Image
+    paginate_by = 50
+
+    def get_queryset(self):
+        return get_objects_for_user(self.request.user, "cases.view_image")
```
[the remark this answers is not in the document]
amickan: Not a direct one, I will add one.

app/grandchallenge/cases/widgets.py
amickan: Yes, although I see some room for improvements. Will make things a bit more explicit

app/grandchallenge/reader_studies/forms.py
amickan: I will add a test

app/grandchallenge/reader_studies/tasks.py
amickan: You're right, it's not necessary.

app/grandchallenge/cases/templates/cases/image-search-widget.html
amickan: Fair enough. I have added hidden elements and include them with `hx-include`

app/grandchallenge/cases/views.py
```diff
@@ -161,6 +173,88 @@ def get_object(self):
         return img
 
 
+class ImageWidgetSelectView(LoginRequiredMixin, View):
+    def get(self, request, *args, **kwargs):
+        query_params = request.GET.copy()
+        interface = query_params.pop("interface", None)[0]
+        widget_name = query_params.pop("WidgetChoice-" + interface, None)[0]
+        help_text = query_params.pop("help_text", None)
+        current_value = query_params.pop("current_value", None)
```
jmsmkn, line 182: Why not access these directly from `request.GET`? Here, if `interface` or `WidgetChoice-` are not present then an type error will be raised as `None` is not sub-scriptable. 

```suggestion
        interface = request.GET["interface"][0]
        widget_name = request.GET[f"WidgetChoice-{interface}"][0]
        help_text = request.GET.get("help_text")
        current_value = request.GET.get("current_value")
```

app/grandchallenge/reader_studies/tasks.py
```diff
@@ -79,17 +84,17 @@ def add_image_to_display_set(
...
-        try:
-            civ.full_clean()
-        except ValidationError as e:
-            upload_session.status = RawImageUploadSession.FAILURE
-            upload_session.error_message = e.message
-            upload_session.save()
-        else:
-            civ.save()
-            display_set.values.add(civ)
+            civ = ComponentInterfaceValue.objects.create(
```
jmsmkn, line 87: Should this be a get_or_create?

jmsmkn (commented):
LGTM! The search view might be quite slow in production due to the large number of images. We may have to use `filter_by_permission` instead of `get_objects_for_user` in `ImageSearchView`. Lets see.

app/grandchallenge/reader_studies/tasks.py
amickan: On this specific line no, but two lines above, we should probably better use a get_or_create, saves some lines of code as well

setup.cfg
```diff
@@ -48,3 +48,5 @@ ignore =
     W503
     # N818 Exceptions should be named Error (TODO)
     N818
+    # B028 Suggest !r when formatted value in f-string is surrounded by quotes.
+    B028
```
amickan, line 52: Pre-commit was failing since the latest flake8 update which includes a check for quotes outside of the braces in an f-string (B028). In most cases in our codebase, these quotes are actually necessary, so I suggest to ignore this new warning.

app/grandchallenge/reader_studies/static/reader_studies/js/remove_hidden_input_fields.js
```diff
@@ -0,0 +1,9 @@
+form_id = document.getElementById("form-id").dataset.formid
+document.querySelector('#form-'+form_id).addEventListener("submit", function(e){
+    // this removes the hidden interface input fields that flexible-image-widget.html and image-search-widget.html add
+    // to prevent multiple interface values from being submitted with the form
```
jmsmkn, line 4: Why filter these out client side rather than server side?

amickan: Changing the request data server side (in the post() function on the view seemed like a bad idea, the request object is probably immutable for a good reason). Handling the multiple values for the `interface` key in the data further down the line required changes in multiple places in the codebase, and so preventing them from being sent along in the first place seemed best. How exactly would you suggest doing this server side?

jmsmkn: Doing this client side is hacky. You shouldn't have to rely on the client running something for validation and use. There is nothing to stop a client sending multiple values for the interface now right, so it would need to be handled server side anyway.

amickan: That makes sense. I have added a server-side solution, but I'm not sure if updating the request data is the best way to go. It's definitely the most straight-forward since it only requires changes in one place, but it does feel a bit hacky as well.

jmsmkn: Yes, it's still not great but is better to do on the server side. Can't you do this as part of the value validation on the form?

Is the problem using htmx for this? Would a select 2 (multiple) not be better?

amickan: We're using a select2 for the interface here already, but for the FlexibleImageWidget we use htmx to update the form in place after a user has chosen a subwidget, and this is when the extra interface values get added. 

Value validation on the `clean_field` method doesn't work because that method never gets called because the clean method on the field (which is run before) already fails validation. I don't think you can (or at least couldn't figure out how to) override the clean method of the ModelChoiceField (which is only defined in the form's init method) So the only way (as far as I can see) to fix this is to override the full_clean method and change the form data before any other clean methods are called. But this is essentially what I did before on the post method.

jmsmkn: Can't you give the field a different name on the sub widget?

amickan: That will affect all other places where the widget is used. I'll check if it causes similarly complicated issues elsewhere.

amickan: Changing the name fixes the problem without introducing issues elsewhere. Feeling a bit stupid for not thinking of that solution myself.

jmsmkn: I mean it took us both a while to get there, it's the new year after all!

PR #2717, opened 2022-12-13
Adds annotation type component interface to development fixtures

miriam-groeneveld:
Needed to be able to test https://github.com/DIAGNijmegen/rse-gcapi/pull/144

### DIFF
--- a/scripts/development_fixtures.py
+++ b/scripts/development_fixtures.py
@@ -466,7 +466,7 @@ def _create_algorithm_demo(users):
             ]
         },
     ]
-    for res, det in zip(results, detections):
+    for res, det in zip(results, detections, strict=False):
         _create_job_result(users, algorithm_image, cases_image, res, det)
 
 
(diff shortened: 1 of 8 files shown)


chrisvanrun:
This doesn't work because the `strict=True` is a python 3.10 feature.

scripts/development_fixtures.py
```diff
@@ -535,6 +535,14 @@ def _create_reader_studies(users):
...
     )
     image.save()
+    annotation_interface = ComponentInterface(
+        store_in_database=True,
+        relative_path="annotation.json",
+        slug="annotation",
+        title="Annotation",
+        kind=ComponentInterface.Kind.TWO_D_BOUNDING_BOX,
+    )
+    annotation_interface.save()
```
chrisvanrun, line 545: This should likely be cherry-picked on top of:
- https://github.com/comic/grand-challenge.org/pull/2715

chrisvanrun (changes requested):
Doesn't work for python 3.8 (default python Django runs underatm)

jmsmkn (approved):
LGTM, there is a tight coupling from the gcapi tests to some of the things in this file, probably they should be separated out and added to a new separate set of fixtures managed in gcapi itself.

PR #2721, opened 2022-12-15
Set credits per job based on the compute usage

jmsmkn:
Many algorithm administrators do not set the credits per job, or set it to a very low value. This causes budget problems. This PR:

- Removes the users ability to set `credits_per_job`
- Each day sets the number of credits per job based on the algorithms average duration and compute requirements and costs
- For short running algorithms sets a maximum of `ALGORITHMS_MAX_DEFAULT_JOBS_PER_MONTH` (50)
- Assigns users by default `ALGORITHMS_CENTS_PER_MONTH`

As `Credit._meta.get_field("credits").get_default() == settings.ALGORITHMS_USER_CENTS_PER_MONTH` it means that 1 credit = 1 cent of compute costs.

If users are running into issues with the new limits then they can be assigned more credits based on their project and funding.

Closes https://github.com/DIAGNijmegen/rse-roadmap/issues/200


amickan (approved):
LGTM!

PR #2735, opened 2023-01-10
Dependency update

jmsmkn:


jmsmkn:
> We no longer need the PDF report for display sets? How come?

They were never used, for RSNA we ended up going in a different direction with the algorithm generating it and viewing the results in the Answer list view was preferred.

amickan:
> > We no longer need the PDF report for display sets? How come?
> 
> They were never used, for RSNA we ended up going in a different direction with the algorithm generating it and viewing the results in the Answer list view was preferred.

Oh ok, wasn't aware that it wasn't being used on GC, but then it makes sense to remove indeed.

amickan (approved):

PR #2736, opened 2023-01-11
Live budget updating

amickan:
- Automatically update the cost estimate on the request form when values in the budget fields change
- Round up cost estimates to nearest 10 Euro
- Add and update some extra warnings

### DIFF
--- a/app/tests/challenges_tests/test_tasks.py
+++ b/app/tests/challenges_tests/test_tasks.py
@@ -1,3 +1,5 @@
+import math
+
 import pytest
 from django.conf import settings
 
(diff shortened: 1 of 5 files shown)


app/tests/challenges_tests/test_tasks.py
```diff
@@ -54,64 +56,84 @@ def test_challenge_creation_from_request(challenge_request):
...
+        challenge_request.budget["Data storage cost for phase 1"]
+        == math.ceil(
+            challenge_request.phase_1_number_of_test_images
+            * challenge_request.average_size_of_test_image_in_mb
+            * settings.CHALLENGES_S3_STORAGE_COST_CENTS_PER_TB_PER_YEAR
+            / 1000000
+            / 100
+            / 10
+        )
+        * 10
```
jmsmkn, line 69: FYI When writing tests it is better to hardcode the values rather than duplicating the logic. You can protect yourself from any changes to the variables (or even test changes to the variables) by overriding them for the test.

jmsmkn (approved):

PR #2738, opened 2023-01-12
Fix partially unanswered Display Set filtering

chrisvanrun:
Relates to: https://github.com/DIAGNijmegen/rse-cirrus-core/issues/2183

An error in reader-study sequencing: partially answered display sets were excluded when unanswered display sets were requested.

This only occurred in the corner case where:
- not all answers for a display set were saved (error in saving or newly created questions)
- **and** there are other user' answers for that specific display set that total more than the number of answerable questions

Pretty hard to think of when writing tests, but happens easily in production.

### DIFF
--- a/app/tests/reader_studies_tests/test_api.py
+++ b/app/tests/reader_studies_tests/test_api.py
@@ -1441,11 +1441,14 @@ def test_display_set_list_filters(client, settings):
     settings.task_eager_propagates = (True,)
     settings.task_always_eager = (True,)
 
-    r = UserFactory()
+    r1 = UserFactory()
+    r2 = UserFactory()
 
     rs1, rs2 = (ReaderStudyFactory() for _ in range(2))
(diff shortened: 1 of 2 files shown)


app/tests/reader_studies_tests/test_api.py
```diff
@@ -2122,23 +2158,6 @@ def test_query_unanswered_display_sets_for_another_user(client, settings):
...
-        user=editor,
-        client=client,
-        method=client.get,
-    )
-    assert response.status_code == 400
-    assert (
-        "Specifying a user is only possible when retrieving unanswered display sets."
-        in str(response.rendered_content)
-    )
-
```
chrisvanrun, line 2141: Moved this check as it makes more sense in the general display filter tests.

amickan (approved):
Good catch! LGTM!

PR #2742, opened 2023-01-17
Workstation annotation style options

chrisvanrun:
Part of the pitch:
- https://github.com/DIAGNijmegen/rse-roadmap/issues/205

This PR adds two new fields:
- `default_brush_size`
- `default_annotation_color`

For the annotation color, a special color-editor widget is added based on [jscolor ](https://jscolor.com/).

## Create/Update View
image

## Detail View
image


chrisvanrun:
Had to do a quick rebase and a merge to fix some CI problems.

jmsmkn:
I made a tool to manage our JS dependencies, jscolor is on cdnjs, I need to fix amass, then we can manage things by adding the following entry to the block in `pyproject.toml`:

```
jscolor = { version = "==2.5.1", include = ["jscolor(.min)?.js$"] }
```

and running `amass lock` and `amass install`.

app/grandchallenge/core/widgets.py
```diff
@@ -19,6 +20,48 @@ class Media:
...
+        )
+
+    def __init__(self, attrs=None, format="auto"):
+        self.format = format
+        super().__init__(attrs)
+
+    def get_context(self, name, value, attrs=None):
+        context = {}
+        context.update(self.attrs.copy() or {})
+        context.update(attrs or {})
```
jmsmkn, line 49: Not to sure about dumping all of the attrs into the context, better to be explicit about the ones that are used IMO.

app/grandchallenge/workstation_configs/models.py
```diff
@@ -270,6 +271,21 @@ class ImageInterpolationType(models.TextChoices):
         validators=[MinValueValidator(limit_value=0.01)],
     )
 
+    default_brush_size = models.DecimalField(
+        blank=True,
+        null=True,
+        max_digits=4,
+        decimal_places=3,
+        validators=[MinValueValidator(limit_value=0.001)],  # 1µm
```
jmsmkn, line 279: Is this OK for pathology?

app/grandchallenge/workstation_configs/templates/workstation_configs/workstationconfig_detail.html
```diff
@@ -279,6 +292,18 @@ <h2>Linking Configuration</h2>
...
 
+{% block extra_css %}
+    {{  block.super }}
+    <style>
+        .color-box {
+            display: inline-block;
+            border: black solid 1px;
+            min-width: 0.9375rem;
+            min-height: 0.9375rem;
+        }
```
jmsmkn, line 303: Move this to our css files, the block is only there for includes from other applications, not for inlines.

jmsmkn:
I rebased on main sorry, you will need to re-create the branch locally.

app/grandchallenge/workstation_configs/models.py
chrisvanrun: Good question, average length of a cell is 120um, so I thought I'd be enough to do 1um. However, mitochondria and other organoids in cells might require a finer touch. (Or viri particles/bacteria). I'll make it 1nm. That should def be clear on anything organic.

app/grandchallenge/core/widgets.py
chrisvanrun: I also wasn't too sure, the Django docs suggest it's common practice for widgets and colorfield did this originally. So I took the safe route. However, I agree with you and it seems weird from a Pythonic way. So let's be explicit.

app/grandchallenge/workstation_configs/templates/workstation_configs/workstationconfig_detail.html
chrisvanrun: Done.

app/grandchallenge/core/widgets.py
chrisvanrun: I've looked more seriously at how the contexts are scoped for widgets and how the superclasses handle these. I've made my changes thus.

jmsmkn (approved):

Issue #2705, opened 2022-11-22
Users should not be able to use their email address as a username


jmsmkn:
We display users usernames in a lot of places. Some users use their email address as their username, this should fail a validation check.

james-milburn-crowe:
Can grab this as a first issue! (but can't assign myself yet!)

jmsmkn:
What is the status of this?

james-milburn-crowe:
Have a single unrelated test failure here, https://github.com/comic/grand-challenge.org/tree/username_not_email
I'm trying to work out why

james-milburn-crowe:
Just moved all the validation logic to a separate validator, and it seems to be working. Just waiting for tests to complete on jenkins,

jmsmkn:
> Just waiting for tests to complete on jenkins,

FYI we aren't using Jenkins for our CI (for anything we do now), for this repo everything is run using GitHub Actions.

PR #2746, opened 2023-01-20
Check for active sessions

amickan:
This adds a active-sessions API endpoint, which is queried in the session-control.js. If there are no active sessions, a new session is created, if there are active sessions, we extract the region from the latest spawned session to define the `sessionOrigin` for the `postMessage` call. 

Unsure if doing this call synchronously is the best idea, or if we better make the call on page load and disable the "View in viewer" button until the call is finished and set the region as a data-attribute.


app/grandchallenge/workstations/views.py
```diff
@@ -75,6 +75,24 @@ def keep_alive(self, *_, **__):
                 status=HTTP_400_BAD_REQUEST,
             )
 
+    @action(detail=False)
+    def active_sessions(self, request):
```
jmsmkn, line 79: I don't think this needs a new endpoint, can't it be achieved by only adding the status to a django filter on the list endpoint?

app/grandchallenge/workstations/static/workstations/js/session-control.js
```diff
@@ -6,14 +6,28 @@ function openWorkstationSession(element) {
...
 
         if (event.ctrlKey) {
             window.open(creationURI);
             return;
         }
 
-        // PoC Hack: ToDo query any existing session
-        const sessionOrigin = element.dataset.sessionOrigin;
+        let activeSessions;
+        $.ajax({
```
jmsmkn, line 18: This adds an API call on each click, which does a filter on the values in the db. Given we only have at most 3 active regions, isn't it quicker just to broadcast the message client side to each of the active regions in turn?

app/grandchallenge/workstations/static/workstations/js/session-control.js
```diff
@@ -27,20 +29,33 @@ function openWorkstationSession(element) {
...
-            }
-            workstationWindow.postMessage(msg, sessionOrigin);
+
+            const fallback = setTimeout(() => {
+                // Assume window is non-responsive
+                window.open(creationURI, windowIdentifier);
+            }, timeout);
+
+            potentialSessionOrigins.forEach((origin) => {
+                sendSessionControlMessage(workstationWindow, origin, {loadQuery: query}, () => {
```
jmsmkn, line 43: `urlencode` is used at the end of `workstation_query(...)` in `workstations.py` to convert a dictionary of query parameters to a URL string. This is no longer necessary as it is now being communicated over this interface. It would be better to pass the dictionary straight across by removing the final `urlquote` and then handling the changed API in CIRRUS.

This should get over the problem of users having non-unreserved characters in their username which are then quoted (https://github.com/python/cpython/blob/e5b08ddddf1099f04bf65e63017de840bd4b5980/Lib/urllib/parse.py#L835-L838), we saw this example with users who use their email as their username (#2705) but will also occur for non-ascii characters:

```
>>> from urllib.parse import quote_plus
>>> quote_plus("dfasg@")
'dfasg%40'
>>> quote_plus("김")
'%EA%B9%80'
```

chrisvanrun: I've actually considered this before. Mainly as a cleaner way of going about things. However,  we're dependent on it being a query string for the initial session start: we aren't too sure when the viewer can handle the session control so we'd dump it in the URL and let the viewer pick it up from there.

A workaround is to build in a wait for the session to get up and send some ack/ready to the GC JS. But that is a hornet's nest of checks and fallbacks. For instance: in 40 seconds of boot-up time I can get the user to navigate away from the original view spawn page. Alternatively, you can park a first session-control action at the session object level.

jmsmkn: I see, yes we need to solve it for both the URL bar and message. Lets sit down and see how the components are decoded on the CIRRUS side.

amickan:
This branch is now an exact copy of the just merged `[playwright-session-control-tests](https://github.com/comic/grand-challenge.org/tree/playwright-session-control-tests)` branch. The only additional change I have made is change `domains` to `workstation_domains`

chrisvanrun:
Since this was nearly a cc from PR:
- #2758

I am just going to approve and merge this.

PR #2752, opened 2023-01-26
added validation to forms.py, so that a users username cannot be an email address

james-milburn-crowe:
Pull request for the above
Logic is included in forms, not in model, because:

1) We're already doing the validation there

2) I'm worried that doing the validation in the model would affect existing users with email usernames

Separately, there's a cleanup task to handle existing users with email usernames, in some way - I'd recommend that we could truncate, by default, their username, and remove everything after the @

One failing test, which I do not understand, as it seems to be failing in each branch, but not in main.


app/grandchallenge/profiles/forms.py
```diff
@@ -74,6 +76,14 @@ def clean(self):
             # accounts with names fooAB barAB etc.
             raise ValidationError("Account details invalid")
 
+        username = self.cleaned_data.get("username", "")
+
+        if re.match(
+            r"(^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$)", username
+        ):
+            # standard python email regex, avoids the RFC nightmare email regex
```
jmsmkn, line 84: There is an existing validator that can be re-used, see https://docs.djangoproject.com/en/4.1/_modules/django/core/validators/#EmailValidator, though here we want the inverse.

app/tests/profiles_tests/test_views.py
```diff
@@ -181,3 +182,60 @@ def test_organization_update(self, client):
         )
 
         assert org1.title not in response.content.decode()
+
+
+@pytest.mark.django_db
+class TestAccountCreation(TestCase):
```
jmsmkn, line 188: Please write pytest style tests. Subclassing `TestCase` is not necessary and we should avoid the weird `.assertWhatever` methods.

app/tests/profiles_tests/test_views.py
```diff
@@ -181,3 +182,60 @@ def test_organization_update(self, client):
...
+            "department": "test",
+            "country": "AX",
+            "website": "https://test.org",
+            "receive_notification_emails": "on",
+            "receive_newsletter": "on",
+            "accept_terms": "on",
+            "password1": "NPAEu_5=fX4d",
+            "password2": "NPAEu_5=fX4d",
+        }
+        response = self.client.post(
```
jmsmkn, line 205: You shouldn't need to use the django db or make a post request to test a form.

app/tests/profiles_tests/test_views.py
```diff
@@ -181,3 +182,60 @@ def test_organization_update(self, client):
...
+            reverse("account_signup"),
+            data,
+            follow=True,
+            secure=True,
+        )
+        assert response.template_name == ["account/verification_sent.html"]
+
+        # self.assertFalse(response.content.decode())
```
jmsmkn, line 213: Remove this commented out line.

jmsmkn:
Rebasing this on main should fix the test.

app/grandchallenge/profiles/forms.py
james-milburn-crowe: frustratingly, unlike the regex and url validator, it doesn't have an inverse - could use the regex validator with the above regex though - would that that be better?

jmsmkn: What about something like:

```python
try:
    EmailValidator()(username)
except ValidationError:
    # Username is not an email address
    pass
else:
    raise ValidationError("You cannot use an email address as a username")
```

james-milburn-crowe: I'll take it - it is cleaner for sure

james-milburn-crowe: I was always told that except...pass statements were somewhat unpythonic, but it's probably considerably more pleasant than a mystery regex

jmsmkn: That's fine in most cases, but here we're relying on the exception to do the thing we want to totally fine to suppress it.

app/tests/profiles_tests/test_views.py
james-milburn-crowe: fixed!

app/grandchallenge/profiles/forms.py
james-milburn-crowe: fixed this to use the EmailValidator - thanks for the pointers!

app/tests/profiles_tests/test_views.py
james-milburn-crowe: think this is fixed - tests are now pytest style, I think

app/grandchallenge/profiles/validators.py
```diff
@@ -0,0 +1,17 @@
...
+    try:
+        EmailValidator()(username)
+    except ValidationError:
+        # Username is not an email address
+        pass
+    else:
+        raise ValidationError("You cannot use an email address as a username")
+
+
+# List is needed as
```
jmsmkn, line 16: The incomplete comment can be removed.

app/grandchallenge/profiles/validators.py
```diff
@@ -0,0 +1,17 @@
+from django.contrib.auth.validators import UnicodeUsernameValidator
```
jmsmkn, line 1: How did you know to include `UnicodeUsernameValidator`, are there other default validators we are now missing from the list?

app/tests/profiles_tests/test_validators.py
```diff
@@ -0,0 +1,13 @@
...
+from grandchallenge.profiles.validators import username_is_not_email
+
+
+def test_username_is_not_email():
+    username_is_not_email("test")
+
+
+def test_username_is_email_validator():
+    with pytest.raises(ValidationError):
+        username_is_not_email("test@gmail.com")
```
jmsmkn, line 13: See if you can re-write these tests as one [using parametrize](https://docs.pytest.org/en/6.2.x/parametrize.html) (hint: you will need to use `null_context`, example should be in the codebase).

app/grandchallenge/profiles/validators.py
james-milburn-crowe: it's described in here: https://docs.djangoproject.com/en/4.1/ref/contrib/auth/#django.contrib.auth.validators.UnicodeUsernameValidator as the default validator, and we don't configure any others

app/tests/profiles_tests/test_validators.py
james-milburn-crowe: rewritten!

jmsmkn (approved):

PR #2758, opened 2023-01-31
Add playwright session control test

amickan:


docker-compose.yml
```diff
@@ -132,6 +132,7 @@ services:
       COMPONENTS_DEFAULT_BACKEND: 'grandchallenge.components.backends.docker.DockerExecutor'
       DEBUG: 'true'
       PYTHONDONTWRITEBYTECODE: 1
+      DJANGO_LIVE_TEST_SERVER_ADDRESS: 'localhost:9200'
```
jmsmkn, line 135: I think it's better to add this as a default when you use `getenv` in `settings.py`.

app/tests/settings.py
```diff
@@ -10,9 +10,15 @@
 from config.settings import *  # noqa: F401, F403, E402, I251
 
 SESSION_COOKIE_DOMAIN = ".testserver"
-ALLOWED_HOSTS = [SESSION_COOKIE_DOMAIN]
+ALLOWED_HOSTS = [SESSION_COOKIE_DOMAIN, ".localhost"]
```
jmsmkn, line 13: It seems like you're fighting the existing setting of `.testserver` here, especially when modifying the default site object in `conftest.py`.

Would it work if the `SESSION_COOKIE_DOMAIN` was `.localhost`? If so, it might be better to create a different PR to main where that is changed for all of the tests, and then you can start with something that is aligned with what you're trying to do here.

app/tests/settings.py
```diff
@@ -10,9 +10,15 @@
...
 SECURE_SSL_REDIRECT = False
 
+TEST_TEMPLATE_DIR = os.path.join(
+    SITE_ROOT,  # noqa 405
```
jmsmkn, line 17: We can use pathlib in all of the (non-third-party) Django settings since Django 3.1. Might be nicer to update `SITE_ROOT` to use that.

app/tests/utils.py
```diff
@@ -226,3 +227,18 @@ def recurse_callbacks(callbacks):
 
     if new_callbacks:
         recurse_callbacks(callbacks=new_callbacks)
+
+
+@contextlib.contextmanager
+def playwright_trace(context, directory=None, name="trace"):
+    directory.mkdir(parents=True, exist_ok=True)
```
jmsmkn, line 234: Won't this leave a bunch of test files on disk? Why not use a temporary directory in a context handler?

app/tests/workstations_tests/test_session_control.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class SessionCreationView(TemplateView):
+    url_route = "new-session/"
+    template_name = "new_session.html"
+
+
+@pytest.mark.playwright
+def test_viewer_session_control(playwright_live_server, page):
+    with playwright_trace(page.context, directory=Path("/app/test_results")):
```
jmsmkn, line 36: You should be able to achieve the same thing with a pytest fixture rather than using a big context manager.

docker-compose.yml
amickan: Actually that does not work. The `live_server` then does not pick this up if it's an environment variable.

jmsmkn: OK then it is better to set it as an ENV var for the test docker container. There is also a problem here that the port may be occupied. Is there a way to get around hardcoding it?

app/tests/utils.py
amickan: I will leave this for Chris to comment on and potentially change since he added this, but if this is a temporary file, it will be gone when the test has run and then we can't open it after, right?

jmsmkn: Yes but you could run the tests with `--pdb --lf` and get a breakpoint on the exception with access to the files.

docker-compose.yml
amickan: The reason we hardcoded it is because we need to know the port for the `SessionControlView`, which is constructed outside of the test and hence is independent from the live_server that is set up for the test from scratch every time.

jmsmkn: In that case why can't you use the `settings` fixture and pass it through that way?

amickan: Ah! Hadn't thought of that. That should work.

amickan: Hm I can't get that work. The settings fixture in the test is seperate from the settings that retrieve the variable from in the get_context_data function on the view.

amickan: Ah never mind, got it to work!

app/tests/settings.py
amickan: Only setting `SESSION_COOKIE_DOMAIN` to `.localhost` without also overwriting the domain (as we do in conftest) does not work.

amickan: So changing this for all tests would require getting rid of the `django_db_setup` fixture that sets the domain to `testserver`. Not sure what consequences that would have.

jmsmkn: > Only setting SESSION_COOKIE_DOMAIN to .localhost without also overwriting the domain (as we do in conftest) does not work.

Yes, I wouldn't expect that to work, it's just that you have two of them now fighting each other, one which is auto loaded and one which is used as a fixture. The idea would be to unify them and use one.

app/tests/utils.py
chrisvanrun: ATM is a debug insertion to get a trace. The traces need to be viewed via a GUI; the quickest way is to dump them somewhere where I can directly read them. (Without going through the hassle of doing a docker copy; running these tests take long enough already.)

app/tests/workstations_tests/test_session_control.py
chrisvanrun: That was the first version: however that would always require a context as (first) positional argument. The context manager is more dynamic in that you can, at any point, provide it with context.

chrisvanrun:
Anne and I have decided to throw the towel into the ring and ignore the subdomains. Let's hope production doesn't have any surprises in that regard.

app/tests/utils.py
amickan: We can also just get rid of it. It was just for debugging purposes anyway.

chrisvanrun: Yes, generating traces is not something we'd like to do all the time, anyway.

app/tests/settings.py
amickan: We no longer need to fight the testerver now that we're not testing the subdomains anymore, so closing this

app/tests/workstations_tests/test_session_control.py
```diff
@@ -0,0 +1,79 @@
...
+    )
+    assert "acknowledge" in sent_msg["sessionControl"]["header"]
+
+    # Test that if acknowledge sent too late, a new session will be created
+    viewer_page.evaluate("enableMockAcksWithDelay()")
+    page.get_by_role("button", name="Launch new Session").click()
+    viewer_page.get_by_text(SessionCreationView.template_name).wait_for()
+    assert viewer_page.url.startswith(session_create_view)
+
+    # Test if disabling the acknowledgment will result in a new session being created
```
chrisvanrun, line 71: I think we should:
* add a data-timeout setting (default it to 3000)
* add a secondary button that sets this 0
* use that secondary button to test the timeout behaviour

app/tests/templates/workstation.html
```diff
@@ -0,0 +1,55 @@
...
+        function handleMessage(event) {
+            const eventString = JSON.stringify(event.data)
+            const node = document.createElement("li");
+            node.innerHTML = eventString;
+            const messages = document.getElementById("messages");
+            messages.appendChild(node);
+
+            const msg = {
+                sessionControl: {
+                    header: {
```
chrisvanrun, line 24: Could we change this from `header` to `meta`?

app/tests/workstations_tests/test_session_control.py
amickan: What is the advantage of using a button to test the timeout behaviour?

app/tests/settings.py
```diff
@@ -13,6 +13,11 @@
 ALLOWED_HOSTS = [SESSION_COOKIE_DOMAIN]
 SECURE_SSL_REDIRECT = False
 
+TEST_TEMPLATE_DIR = os.path.join(
+    SITE_ROOT / "tests/templates",  # noqa 405
+)
+TEMPLATES[0]["DIRS"].append(TEST_TEMPLATE_DIR)  # noqa 405
```
jmsmkn, line 19:
```suggestion
TEST_TEMPLATE_DIR = os.path.join(
    SITE_ROOT / "tests/templates",  # noqa 405
)
TEMPLATES[0]["DIRS"].append(SITE_ROOT / "tests" / "templates")  # noqa 405
```

app/tests/workstations_tests/test_session_control.py
```diff
@@ -0,0 +1,79 @@
...
+    template_name = "new_session.html"
+
+
+@pytest.mark.playwright
+def test_viewer_session_control(live_server, page, settings):
+    url = urlparse(live_server.url)
+    settings.DJANGO_LIVE_TEST_SERVER_ADDRESS = f"{url.netloc}"
+    session_create_view = f"{url.scheme}://{url.netloc}/new-session/"
+    session_control_view = f"{url.scheme}://{url.netloc}/session-control/"
+    mock_workstation_view = f"{url.scheme}://{url.netloc}/workstation/"
```
jmsmkn, line 39: Why are the URLs hardcoded and not using `reverse` in the tests or `{% url ... %}` in the templates?

app/grandchallenge/workstations/static/workstations/js/session-control.js
```diff
@@ -15,8 +15,14 @@ function openWorkstationSession(element) {
...
         const regions = JSON.parse(document.getElementById('workstation-regions').textContent);
-        const potentialSessionOrigins = getSessionOrigins(domain, regions);
 
+        let potentialSessionOrigins;
+        if (domain.includes("localhost")) {
+            const port = window.location.port;
+            potentialSessionOrigins = Array(`http://${domain}${port ? ':' + port : ''}`);
+        } else{
+            potentialSessionOrigins = getSessionOrigins(domain, regions);
+        }
```
jmsmkn, line 25: Why isn't this handled in `getSessionOrigins`. Checking if the domain contains localhost does not seem the best way to go about this.

app/tests/templates/workstation.html
amickan: only heere or do you also want to change that in the postmessage that we send in session-control.js (where we send the id along)?

app/tests/workstations_tests/test_session_control.py
amickan: Makes sense to use {% url %} in the templates. In the tests we still need to construct the urls so that we get absolute urls and not just the paths, but I will replace the hardcoded paths with reverse()

app/tests/templates/workstation.html
chrisvanrun: Everywhere, as the default place to add meta items such as id and acks.

app/tests/workstations_tests/test_session_control.py
chrisvanrun: Because we set the data- fields on buttons. So the secondary button would have a 0 timeout.

app/grandchallenge/workstations/static/workstations/js/session-control.js
amickan: Yes, I also don't like it. Maybe we could instead add a data-debug setting on the button that is only set in the tests and use that.

app/tests/workstations_tests/test_session_control.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+class SessionCreationView(TemplateView):
+    template_name = "new_session.html"
+
+
+@pytest.mark.playwright
+def test_viewer_session_control(live_server, page, settings):
+    url = urlparse(live_server.url)
+    settings.DJANGO_LIVE_TEST_SERVER_ADDRESS = f"{url.netloc}"
```
jmsmkn, line 37: Does this variable still need to be set? If not, is there a need to parse `live_server.url`? Can't that just be used directly?

amickan: It still needs to be set because we need it in the view template. It doesn't serve any other purpose than that, so we could call it anything.

jmsmkn: Really, the values don't appear to be used in the templates? Or am I missing something?

jmsmkn: Ah wait, domain is, netloc isn't, but maybe neither need to be handled if the other comment is solved to do more things on the Python side.

app/grandchallenge/workstations/static/workstations/js/session-control.js
```diff
@@ -48,11 +49,15 @@ function openWorkstationSession(element) {
...
 }
 
-function getSessionOrigins(hostname, regions) {
-    const protocol = "https";
+function getSessionOrigins(hostname, regions, debug) {
+    const protocol = window.location.protocol;
     const port = window.location.port;
 
-    return regions.map((region) => `${protocol}://${region}.${hostname}${port ? ':' + port : ''}`);
+    if (debug){
```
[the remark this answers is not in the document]
amickan: But then we'd be matching on the domain name again, right?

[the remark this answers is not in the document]
jmsmkn: No, change the `workstation_regions` context processor to do something like:

```
def workstation_domains(*_, **__):
    return [
      *["https://{region}.domain" for region in settings.WORKSTATIONS_ACTIVE_REGIONS],
      *settings.EXTRA_WORKSTATION_BROADCAST_DOMAINS 
    ]
```

Then settings.py

```
EXTRA_WORKSTATION_BROADCAST_DOMAINS = []
```

Then in your test

```
def test_blah(settings, live_server, ...):
   settings.EXTRA_WORKSTATION_BROADCAST_DOMAINS = [live_server.url]
```

Then update your JS to rather than take a list of subdomains that you then construct a list of URLs from to just take a list of URLs.

app/tests/workstations_tests/test_session_control.py
amickan: Indeed no longer necessary after the rework from the above comment!

app/grandchallenge/workstations/static/workstations/js/session-control.js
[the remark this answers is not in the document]
amickan: Ah this is very elegant. I had misunderstood your previous comment. This works like a charm! Thanks!

jmsmkn (approved):

jmsmkn:
LGTM, Would be good to change the context variable `domains` to `workstation_domains` just so that it is clear what the variable means.

PR #2761, opened 2023-02-03
Add environment variable for session control timeout

amickan:
This adds a `WORKSTATIONS_SESSION_CONTROL_TIMEOUT` environment variable to easily change the timout in production.

### DIFF
--- a/app/config/settings.py
+++ b/app/config/settings.py
@@ -388,6 +388,7 @@ TEMPLATES = [
                 "grandchallenge.core.context_processors.newsletter_signup",
                 "grandchallenge.core.context_processors.viewport_names",
                 "grandchallenge.core.context_processors.workstation_domains",
+                "grandchallenge.core.context_processors.workstation_session_control_timeout",
                 "machina.core.context_processors.metadata",
             ],
             "loaders": [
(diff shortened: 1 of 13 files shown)


jmsmkn:
Is this not way too fine detail? I can't see sites wanting to modify this, if we get it wrong first time we could change the value in the js file and re-deploy.

chrisvanrun:
> Is this not way too fine detail? I can't see sites wanting to modify this, if we get it wrong first time we could change the value in the js file and re-deploy.

It is indeed fine. But I suspect this will likely need ongoing tweaking during integration and hooks heavily on what CIRRUS does:
* we will want to likely integrate these changes next week which will not see the CIRRUS release, we want to reduce the extra start-up time for sessions as much as possible.
* once CIRRUS is released we can see if it generally feels ok or is too flaky at 1s.

If re-deployment is just as easily done (or nearly) as tweaking production envs @amickan I suggest we put it at 1s by default and see what it does.

amickan:
Yes, both cases require a one line change and a redeploy, so it doesn't really matter which way we go. But just changing the default in the js code clutters the buttons less, so maybe that is preferred after all. I will close this PR and change the default for the timeout to 1s in the other PR I'm working on.

chrisvanrun:
Check! I am used to CIRRUS where deploying is something quite heavy.

PR #2762, opened 2023-02-03
[Session Control] module scoping JS

chrisvanrun:
This PR adds:
* Alt-click behavior to copy the query string (supports CIRRUS devs debugs)
* Module scopes session-control.js

The other session JS code is only being loaded on a single duty page, however, the session-control code will be included on a lot of different pages. Making it a module prevents it from conflicting with other (potential) JS. This does entail that overwriting the session-control data attribute querySelector is no longer possible. Tbh, this will likely never be used and another path can be followed to re-implement this if needed.


amickan (approved):
LGTM!

PR #2763, opened 2023-02-03
Add spinner while waiting for session control hooks to run 

amickan:
This replaces the eye icon on the launch button with a spinner until the ack message has been received or the timeout has been reached and a new window is opened and focussed. 

I also decreased the timeout to 1s as per Chris' earlier request. 

https://user-images.githubusercontent.com/30069334/216616855-3d3477f8-1df5-428f-958e-bd4852f26e33.mp4

### DIFF
--- a/app/grandchallenge/workstations/static/workstations/js/session-control.mjs
+++ b/app/grandchallenge/workstations/static/workstations/js/session-control.mjs
@@ -43,8 +43,10 @@ function openWorkstationSession(element) {
 
             potentialSessionOrigins.forEach((origin) => {
                 sendSessionControlMessage(workstationWindow, origin, {loadQuery: query}, () => {
-                    clearTimeout(fallback)
-                }, element);
+                    clearTimeout(fallback);
+                    workstationWindow.focus(); // absolutely necessary in Firefox, in Chrome/Edge window.open() already focuses the window
+                    removeSpinner(element);
+                });
             });
         }
     }
@@ -57,7 +59,7 @@ function hookSessionControllers() {
     }
 }
 
-function sendSessionControlMessage(targetWindow, origin, action, ackCallback, triggeringElement) {
+function sendSessionControlMessage(targetWindow, origin, action, ackCallback) {
     const messageId = UUIDv4();
     const msg = {
         sessionControl: {
@@ -80,8 +82,6 @@ function sendSessionControlMessage(targetWindow, origin, action, ackCallback, tr
         }
         if (ack.id === messageId) {
             ackCallback();
-            targetWindow.focus(); // absolutely necessary in Firefox, in Chrome/Edge window.open() already focusses the window
-            removeSpinner(triggeringElement);
             window.removeEventListener('message', checkAckMessage);
         }
     }
--- a/app/grandchallenge/workstations/static/workstations/js/session-control.mjs
+++ b/app/grandchallenge/workstations/static/workstations/js/session-control.mjs
@@ -6,7 +6,7 @@ function openWorkstationSession(element) {
         const url = element.dataset.createSessionUrl;
         const query = element.dataset.workstationQuery;
         const creationURI = `${url}?${query}`;
-        const timeout = element.dataset.timeout || 3000;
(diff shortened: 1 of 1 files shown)


app/grandchallenge/workstations/static/workstations/js/session-control.mjs
```diff
@@ -80,6 +80,8 @@ function sendSessionControlMessage(targetWindow, origin, action, ackCallback) {
         }
         if (ack.id === messageId) {
             ackCallback();
+            targetWindow.focus(); // absolutely necessary in Firefox, in Chrome/Edge window.open() already focusses the window
+            removeSpinner(triggeringElement);
```
chrisvanrun, line 84: Can't you place this in the ackCallback() ?

app/grandchallenge/workstations/static/workstations/js/session-control.mjs
```diff
@@ -32,19 +34,17 @@ function openWorkstationSession(element) {
         }
 
         if (workstationWindow === null || isBlankContext) {
-            window.open(creationURI, windowIdentifier);
+            createNewSessionWindow(creationURI, windowIdentifier, element);
```
chrisvanrun, line 37: I suggest replacing this with something along the lines: this keeps it more DRY:
```Javascript
function fallbackAction() {
   window.open(creationURI, windowIdentifier);
   removeSpinner(element);
}
...
fallbackAction();
...
fallbackAction();
```

amickan: I'm already doing that here with the `createNewSessionWindow` function though, or what do you mean?

app/grandchallenge/workstations/static/workstations/js/session-control.mjs
amickan: The ackCallback just clears the timer, right So then that code would be run if I understand it correctly.

app/grandchallenge/workstations/static/workstations/js/session-control.mjs
chrisvanrun: You are, but you are repeating yourself in a way. If I would want to change the thing that happens when the session-control gives up advanced methods (i.e. post messages). I would need to change things in multiple locations.

app/grandchallenge/workstations/static/workstations/js/session-control.mjs
chrisvanrun: It does now, yes. But it can easily be extended. Adding these lines 'dirties' the sendSessionControlMessage function and makes it have multiple responsibilities.

app/grandchallenge/workstations/static/workstations/js/session-control.mjs
amickan: I'm confused. It already is just one location that you would need to update (the `createNewSessionWindow` function). What am I missing?

chrisvanrun: It is somewhat of a style choice. So let's leave it. 

To go a bit deeper into my thoughts, just to clear the confusion: It works now, but imagine if you'd again need to add behavior to the fallback which would require additional arguments. You would then need to add that in two locations whereas, in the **inline**-function definition I suggested, you only need to do it at one spot.

app/grandchallenge/workstations/static/workstations/js/session-control.mjs
```diff
@@ -101,6 +103,24 @@ function copyTextToClipboard(text) {
...
 
+function setSpinner(element) {
+    element.querySelector("i").style.display = "none";
+    const spinner = document.createElement("span");
+    spinner.classList.add("spinner-border", "spinner-border-sm");
+    element.prepend(spinner);
+}
+
+function removeSpinner(element) {
+    const spinner = element.querySelector("span");
```
chrisvanrun, line 114: Would the cleaner way, rather than selecting on `<span>`, be to select on `spinner-border` class? I could forsee someone adding a <span> within a button.

chrisvanrun (approved):

PR #2764, opened 2023-02-03
Add test for wrong session control ack

amickan:


chrisvanrun (approved):
Looks good to me, I think this might just work!

PR #2765, opened 2023-02-06
Add cross-origin communication with workstations

amickan:
This PR adds methods to communicate directly with workstations using `window.postMessage`. This allows Grand Challenge to send messages to workstation instances, so that new display sets or jobs can be displayed without having to go through a full page reload of the workstation. If the workstation does not support this communication channel a new workstation window will be launched. This is detected by a posted message being unacknowledged from all rendering sub-domains.

See https://github.com/DIAGNijmegen/rse-roadmap/issues/186


app/config/settings.py
```diff
@@ -1112,6 +1114,8 @@
 # Number of minutes grace period before the container is stopped
 WORKSTATIONS_GRACE_MINUTES = 5
 
+WORKSTATIONS_EXTRA_BROADCAST_DOMAINS = []
```
jmsmkn, line 1117:
```suggestion
# Extra domains to broadcast workstation control messages to. Used in tests.
WORKSTATIONS_EXTRA_BROADCAST_DOMAINS = []
```

app/grandchallenge/workstations/static/workstations/js/session-control.mjs
```diff
@@ -0,0 +1,134 @@
+'use strict';
```
jmsmkn, line 1: Is this necessary for modules?

app/grandchallenge/workstations/static/workstations/js/session-control.mjs
```diff
@@ -0,0 +1,134 @@
...
+
+function openWorkstationSession(element) {
+    return (event) => {
+        const windowIdentifier = element.dataset.workstationWindowIdentifier;
+        const url = element.dataset.createSessionUrl;
+        const query = element.dataset.workstationQuery;
```
jmsmkn, line 7: `url` and `query` don't appear to be used separately. Why not pass through the `creationURI` instead?

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
```diff
@@ -429,4 +433,7 @@ <h2>Copy</h2>
 {% block script %}
     {{ block.super }}
     <script src="{% static "js/refresh_sidebar.js" %}"></script>
+    {{ workstation_domains|json_script:"workstation-domains" }}
+    <script type="module" src="{% static 'workstations/js/session-control.mjs' %}"></script>
+
```
jmsmkn, line 438: Could make this one include from a partial template, may make updating easier.

app/grandchallenge/workstations/static/workstations/js/session-control.mjs
amickan: Ah apparently not (double checked [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#strict_mode_for_modules))

app/grandchallenge/workstations/templatetags/workstations.py
```diff
@@ -79,3 +80,27 @@ def workstation_query(  # noqa: C901
         )
 
     return urlencode(query)
+
+
+@register.simple_tag()
+def workstation_query(**kwargs):
```
jmsmkn, line 86: This template tag shouldn't be used anywhere else now, right?

app/grandchallenge/workstations/templatetags/workstations.py
```diff
@@ -79,3 +80,27 @@ def workstation_query(  # noqa: C901
...
+):
+    create_session_url = reverse(
+        "workstations:workstation-session-create",
+        kwargs={"slug": workstation.slug},
+    )
+    workstation_query_string = get_workstation_query_string(**kwargs)
+    window_identifier = f"workstation-{context_object._meta.app_label}"
+    return (
+        f"data-session-control "
+        f"data-create-session-url= {create_session_url} "
```
jmsmkn, line 102: I think the values should be quoted, just in case.

app/grandchallenge/workstations/templatetags/workstations.py
```diff
@@ -79,3 +80,27 @@ def workstation_query(  # noqa: C901
...
+
+
+@register.simple_tag()
+def workstation_query(**kwargs):
+    return get_workstation_query_string(**kwargs)
+
+
+@register.simple_tag()
+def workstation_session_control_data(
+    workstation=None, context_object=None, timeout=None, **kwargs
```
jmsmkn, line 92: This is quite a complex method so enforce kwargs only.

```suggestion
    *, workstation=None, context_object=None, timeout=None, **kwargs
```

app/grandchallenge/workstations/templatetags/workstations.py
jmsmkn: I can see that it is in `rawimageuploadsession_detail.html` - this should be replaced with a button for the new session control, good to find. 

Also used in `display_set.workstation_url`, but I _think_ that property can be updated to use the query string. That seems only to be used when creating a display set from an algorithm (we can decide if opening the viewer for that new display set is the best idea or not).

app/grandchallenge/workstations/static/workstations/js/session-control.mjs
```diff
@@ -0,0 +1,132 @@
...
+            window.open(creationURI);
+            return;
+        }
+
+        if (event.altKey) {
+            copyTextToClipboard(query);
+            return;
+        }
+
+        setSpinner(element);
```
HarmvZ, line 19: Maybe put the next part of this function in a try/catch block and put `removeSpinner()` in the catch block so that it always is removed even in case of an unexpected error in the next part. You can then always throw the error again so that sentry will catch it.

app/grandchallenge/workstations/static/workstations/js/session-control.mjs
```diff
@@ -0,0 +1,132 @@
+function openWorkstationSession(element) {
+    return (event) => {
+        const windowIdentifier = element.dataset.workstationWindowIdentifier;
```
HarmvZ, line 3: You might want to disable the button when this function is running so that it cannot be triggered multiple times by impatient users. Removing the disabled attribute should then also be done in the catch block as with the spinner removal.

HarmvZ (commented):
Looks good, I just have a few small comments. 
The playwright stuff is nice!

app/grandchallenge/workstations/static/workstations/js/session-control.mjs
```diff
@@ -0,0 +1,132 @@
...
+        const workstationWindow = window.open('', windowIdentifier);
+
+        // check if we just opened a blank or existing context
+        let isBlankContext = false;
+        try {
+            isBlankContext = workstationWindow.document.location.href === "about:blank";
+        } catch (err) {
+            // ignore any errors with getting the href as it entails we did not open
+            // the context
+            console.warn(err);
```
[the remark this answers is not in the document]
chrisvanrun: I think this should ignore the specific error that we expect but otherwise raise it.

app/grandchallenge/workstations/templatetags/workstations.py
amickan: Ah good one. I tested both of these cases and it works.

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
amickan: You mean putting these two lines in a partial template and then including that template everywhere where we are currently using these two lines, and not just here, right?  Good idea.

app/grandchallenge/workstations/static/workstations/js/session-control.mjs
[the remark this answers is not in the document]
amickan: This catches all / any errors not just errors specific to getting href, right? If we only want to catch a specific error, we should be explicit. I'm not sure what errors we'd want to ignore and which ones we don't want to ignore. When a user navigates to a different website on the workstationWindow right after it's opened, we'd get a DOMException that our CORS request was blocked. We might want to catch that and return, probably?

app/grandchallenge/workstations/templatetags/workstations.py
```diff
@@ -79,3 +81,25 @@ def workstation_query(  # noqa: C901
...
+        "workstations:workstation-session-create",
+        kwargs={"slug": workstation.slug},
+    )
+    workstation_query_string = get_workstation_query_string(**kwargs)
+    window_identifier = f"workstation-{context_object._meta.app_label}"
+    data_attrs = {
+        "data-create-session-url": create_session_url,
+        "data-workstation-query": workstation_query_string,
+        "data-workstation-window-identifier": window_identifier,
+        "data-timeout": timeout if timeout else False,
```
jmsmkn, line 100: Minor cleanup to make it easier to read

```suggestion
    *, workstation, context_object, timeout=False, **kwargs
):
    create_session_url = reverse(
        "workstations:workstation-session-create",
        kwargs={"slug": workstation.slug},
    )
    workstation_query_string = get_workstation_query_string(**kwargs)
    window_identifier = f"workstation-{context_object._meta.app_label}"
    data_attrs = {
        "data-create-session-url": create_session_url,
        "data-workstation-query": workstation_query_string,
        "data-workstation-window-identifier": window_identifier,
        "data-timeout": timeout,
```

app/tests/workstations_tests/test_templatetags.py
```diff
@@ -151,24 +162,55 @@ def test_workstation_query_for_images(settings):
...
+    assert (
+        f'data-workstation-window-identifier="workstation-{obj._meta.app_label}"'
+        in data
+    )
+
+    data2 = workstation_session_control_data(
+        workstation=wk, context_object=obj, reader_study=obj, timeout=200
+    )
+    assert f'data-workstation-query="readerStudy={obj.pk}"' in data2
+    assert 'data-timeout="200"' in data2
```
jmsmkn, line 216: Why this pattern? If you do `assert data == 'data-session-control data-create-session-url='http://the/url data-timeout="200"'` etc, then you catch missing spaces etc. You can drop in pks as needed.

app/grandchallenge/workstations/templates/workstations/partials/session-control-data.html
```diff
@@ -0,0 +1 @@
+data-session-control {% for name, value in session_control_data %}{% if value is not False %} {{ name }}{% if value is not True %}="{{ value|stringformat:'s' }}"{% endif %}{% endif %}{% endfor %}
```
jmsmkn, line 1: `{% if value is not False %} {{ name }}{% if value is not True %}` makes it hard to follow. Did you copy this from elsewhere? I can't see where it would be used right now, but we could - does this mean that we could drop putting `data-session-control` in the template and adding `"data-session-control": True` to `data_attrs` in the context?

amickan: Yes that is true, we could drop "data-session-control" from the template. I'm using the template that is used in Django forms for dropping in widget attributes. I just couldn't find a way to reference it directly, which is why I copied it over into a new partial template. The original template also takes the attributes from `widget.attrs`, which I changed here to just session_control_data for simplicity.

jmsmkn: `render_to_string('django/forms/attrs.html', ...)` ?

amickan: I get a template does not exist error. I think we would need to extend the TEMPLATES["DIRS"] in settings.py, but I haven't found the correct reference yet.

jmsmkn: Ah, was looking at a jinja2 template. How about:

```python
render_to_string("django/forms/widgets/attrs.html", {"widget": "attrs": data_attrs})
```

amickan: Still can't find the template :-/

jmsmkn: OK, this is interesting, I'm learning a lot about Django internals. See the document on the [Form Rendering API](https://docs.djangoproject.com/en/4.1/ref/forms/renderers/).

I think we could do:

```python
from django.forms.renderers import DjangoTemplates

...

DjangoTemplates().render("django/forms/attrs.html", {"attrs": data_attrs})
```

Alternatively I think we could add [`'django.forms'` to `INSTALLED_APPS`](https://docs.djangoproject.com/en/4.1/ref/forms/renderers/#django.forms.renderers.TemplatesSetting) but I'm unsure on the side effects of that.

amickan: Yes, the DjangoTemplates renderer works!

app/grandchallenge/workstations/static/workstations/js/session-control.mjs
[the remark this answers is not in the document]
amickan: That seems reasonable to me. I'll change it so that we catch and ignore that one error and reraise all others.

app/grandchallenge/reader_studies/templates/reader_studies/display_set_detail.html
```diff
@@ -43,8 +44,11 @@
       >
           Edit
       </button>
-    <a href="{% url 'workstations:workstation-session-create' slug=object.reader_study.workstation.slug %}?{% workstation_query display_set=object config=reader_study.workstation_config %}" class="btn btn-primary">
-      <i class="fa fa-eye"></i> View Display Set
-    </a>
+      <button class="btn btn-primary btn-sm"
+              title="View in viewer"
+              {% workstation_session_control_data workstation=object.reader_study.workstation context_object=object display_set=object config=object.reader_study.workstation_config %}
```
jmsmkn, line 49: The `context_object` is the `display_set` here. Should it be the `reader_study`?

jmsmkn: Actually shouldn't matter as only the app label is used.

app/grandchallenge/workstations/static/workstations/js/session-control.mjs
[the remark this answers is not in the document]
chrisvanrun: It is indeed the security/limiting factor we want to catch because the target window wasn't created in the current browser context.

jmsmkn (approved):
LGTM!

chrisvanrun:
There is a bug with double quotes in the data- attributes.

jmsmkn:
> There is a bug with double quotes in the data- attributes.

Please explain...

chrisvanrun:
> > There is a bug with double quotes in the data- attributes.
> 
> Please explain...

Merging and deploying this PR would break all viewer interactions. 

All data HTML attributes are double-quoted and result in creation URLs such as: `https://gc.localhost/reader-studies/reader-study/"https://gc.localhost/viewers/cirrus-core/sessions/create/"?"readerStudy=4c652ac6-315e-4f17-8fff-9e4f638a5cf7"`

jmsmkn (dismissed):

jmsmkn (approved):

Issue #2759, opened 2023-01-31
`test_display_set_index` flaky


jmsmkn:
A few spurious failures for this test, needs to be made more robust.

Example https://github.com/comic/grand-challenge.org/actions/runs/4052565213

jmsmkn:
Still problematic with `<User: test_user_1615>` https://github.com/comic/grand-challenge.org/actions/runs/4321976481/jobs/7543833591

PR #2769, opened 2023-02-10
Fix session control window error handling

### LINKED ISSUE #2768: Viewer context switching breaks on CORS issues
Catching the expected security issue when an open session (with a different origin) should not result in creating a new session.

https://github.com/comic/grand-challenge.org/blob/fd32bc360bc60aa99348efce3a343c39ab6e9017/app/grandchallenge/workstations/static/workstations/js/session-control.mjs#L30

chrisvanrun:
Closes #2768 

Fix non-homologous-origin session window always resulting in fallback session creation.

### DIFF
--- a/app/grandchallenge/workstations/static/workstations/js/session-control.mjs
+++ b/app/grandchallenge/workstations/static/workstations/js/session-control.mjs
@@ -25,14 +25,13 @@ function openWorkstationSession(element) {
             // check if we just opened a blank or existing context
             let isBlankContext = false;
             try {
-                isBlankContext = workstationWindow.document.location.href === "about:blank";
+                isBlankContext = workstationWindow.document.location.href === 'about:blank';
             } catch (err) {
-                if (err.name == 'SecurityError') {
-                    // catch SecurityErrors (i.e. blocked CORS requests) so that we can open a new window with a new session
+                // A SecurityError (i.e. blocked CORS requests) suggests that the window is likely an existing session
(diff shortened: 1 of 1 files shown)


chrisvanrun:
Anne mentioned that we should likely keep throwing any unexpected errors to Sentry, so I redid the logic a bit.

chrisvanrun:
Tests failed on:
- #2759

HarmvZ (changes requested):
Finally block returning early as discussed in call.

app/grandchallenge/workstations/static/workstations/js/session-control.mjs
```diff
@@ -121,6 +123,9 @@ function setSpinner(element) {
 
 function removeSpinner(element) {
     const spinner = element.querySelector(".spinner-border");
+    if (spinner === null) {
+
+    }
```
HarmvZ, line 128: Leftover?

```suggestion
```

HarmvZ (changes requested):

app/grandchallenge/workstations/static/workstations/js/session-control.mjs
chrisvanrun: Not totally, I got disrupted in person when working on this and completely forgot to finish the copy-paste =/. I noticed that running removeSpinner multiple times doesn't work, this is a fix.

HarmvZ (approved):

PR #2772, opened 2023-02-16
Allow workstation editors to launch debug sessions

jmsmkn:
This PR allows workstation editors to set the region and add extra environment variables for a new session. Limitation is that the editor cannot have more than 1 session running to prevent flooding of the server and confusion about env vars being passed, so they may have to wait for a workstation session to finish.

Workstation detail view:
image

The form (with validation):
image

The env vars being passed:
image

Note that the env vars editors set here are used as the default environment variables, we will overwrite some keys if editors try to set them (e.g. `GRAND_CHALLENGE_UNSAFE`, `GRAND_CHALLENGE_AUTHORIZATION`, `GRAND_CHALLENGE_API_ROOT`). Editors can only create debug sessions for themselves so should be safe.

Also removes `WORKSTATIONS_DNS_RESOLVER` which is no longer used.

Old style forms and tests are used for consistency with the rest of the workstations application.


app/grandchallenge/workstations/models.py
```diff
@@ -412,13 +454,16 @@ def environment(self) -> dict:
...
-        env = {
-            "NGINX_RESOLVER": settings.WORKSTATIONS_DNS_RESOLVER,
-            "GRAND_CHALLENGE_API_ROOT": unquote(reverse("api:api-root")),
-            "WORKSTATION_SENTRY_DSN": settings.WORKSTATION_SENTRY_DSN,
-            "WORKSTATION_SESSION_ID": str(self.pk),
-            "CIRRUS_KEEP_ALIVE_METHOD": "old",
-        }
+        env = {var["name"]: var["value"] for var in self.extra_env_vars}
+
+        env.update(
```
amickan, line 459: You could also additionally prevent those from being set in the schema as well. Still need to be overwritten of course, but it would make it clear that those values cannot be user-defined.

amickan (approved):

app/grandchallenge/workstations/models.py
jmsmkn: Sure, but then you would need to maintain the list of all the un-settable ones in the Regex of the name value validator. I don't think that we'd ever do that, and no big deal if a user tries to set that and it doesn't work.

PR #2773, opened 2023-02-20
Do not give editors permission to view all jobs

amickan:
Messy start on https://github.com/DIAGNijmegen/rse-roadmap/issues/207


app/grandchallenge/algorithms/tasks.py
```diff
@@ -91,33 +104,25 @@ def run_algorithm_job_for_inputs(*, job_pk, upload_pks, user_upload_pks):
...
-    missing_inputs = list(civ for civ in job.inputs.all() if not civ.has_value)
-    if missing_inputs:
-        error_message += (
-            f"Job can't be started, input is missing for "
-            f"{oxford_comma([c.interface.title for c in missing_inputs])}"
-        )
-
     res = self.AsyncResult(task_id).result
-
-    if isinstance(res, ImageImportError):
```
amickan, line 120: Did this ever work? `res` is a instance of a `ChordError`. I couldn't find a better way than string matching so far...

jmsmkn: It must have done at some point, but obviously wasn't covered by a test. Celery has this annoying feature of auto-upgrading chains of groups to chords, which changes their return types. Must have happened when changing the implementation of the image import tasks.

jmsmkn:
Thanks! I will carry this on in another PR.

PR #2776, opened 2023-02-21
Remove `view_job` permission from algorithm editors

jmsmkn:
At the moment algorithm editors get permission to see all of the input and output data for jobs created for their algorithm by other users through the "try out algorithm" button or the job create POST API view. Now that we have a lot of editors and other options for sharing algorithm jobs with other users such as archives or job viewers there does not seem a good reason to keep doing this.

Editors still get the `"view_logs"` permission so that they can view the logs if they are given access to the job.

Most of the changes are refactoring to make the job from inputs path easier to understand as I was struggling with that.


app/grandchallenge/algorithms/tasks.py
```diff
@@ -129,7 +134,7 @@ def on_job_creation_error(self, task_id, *args, **kwargs):
 def execute_algorithm_job_for_inputs(*, job_pk):
     job = Job.objects.get(pk=job_pk)
 
-    # Send an email to the algorithm editors and creator on job failure
+    # Notify the job creator failure
```
amickan, line 137: So `if isinstance(res, ImageImportError):` a couple of lines above works now?

jmsmkn: Don't know, didn't test it, existing bug if not so will follow that up in a later PR if necessary.

amickan (approved):
LGTM! Very nice refactoring. I wasn't aware of / familiar with model managers. Much cleaner and easier to understand now.

PR #2778, opened 2023-02-22
Add widget model for questions

amickan:
This adds a `AnswerWidget` base class as well as a `AcceptRejectFindingsWidget` subclass. The base class is foreign keyed to the `Question` model. This also adds the new `answer_widget` field to the question serializer.

Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/191


amickan:
I don't understand the failing test `api_tests/test_schema.py::test_schema_is_valid`. What is `assert not GENERATOR_STATS` testing exactly?

amickan:
> I don't understand the failing test `api_tests/test_schema.py::test_schema_is_valid`. What is `assert not GENERATOR_STATS` testing exactly?

Ah I see there are warnings!

app/grandchallenge/reader_studies/models.py
```diff
@@ -1176,6 +1191,34 @@ def assign_permissions(self):
         )
 
     def clean(self):
+        super().clean()
+        self._clean_answer_widget()
+        self._clean_answer_type()
+        self._clean_interface()
+
+    def _clean_answer_widget(self):
```
amickan, line 1199: I'm no longer sure whether this is the best place for these validations. We need to do those on the forms `clean` method before actually creating an instance of an `AnswerWidget`. The benefit of doing it here was that this also covers the admin forms, but I don't think we ever really create or edit questions in the admin, do we? If not, I think it simplifies things if we move these checks to the form's clean method.

app/grandchallenge/reader_studies/models.py
```diff
@@ -1508,3 +1552,31 @@ def delete(self):
...
         unique_together = (("reader_study", "user"),)
+
+
+class AnswerWidgetKindChoices(models.TextChoices):
+    ACCEPT_REJECT = "ACCEPT_REJECT", "Accept/Reject Findings"
+
+
+class AnswerWidget(models.Model):
+
+    kind = models.CharField(
```
jmsmkn, line 1563: This should need to be a db field, it is a property of each class and not of each instance. Best thing to do is set it to None and then set it as a class attribute in the subclass. In the base class or serialiser you can add a method to get the string value of the kind through looking it up in the text choices.

app/grandchallenge/reader_studies/models.py
jmsmkn: Having the validation in one place on the model is a good thing, not sure if it is possible.

app/grandchallenge/reader_studies/models.py
```diff
@@ -1508,3 +1552,31 @@ def delete(self):
...
+
+
+class AcceptRejectFindingsWidget(AnswerWidget):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.kind = AnswerWidgetKindChoices.ACCEPT_REJECT
+
+    @staticmethod
+    def supported_answer_types():
+        return AnswerType.get_multiple_annotation_types()
```
jmsmkn, line 1582: Probably better to put each answer type in here directly as this will be closely tied to what is actually implemented in CIrRUS.

app/grandchallenge/reader_studies/models.py
jmsmkn: Actually, hmm, I see now. There is a problem with the designed I proposed as the foreign key must be to a concrete model! (we could use gfk but they are slow and hard to limit to a particular set of models). Rather than having a base class and a bunch of implementations I think we will need one `Widget` class, which has a `kind` like you have done here, then the `Question` can have a `OneToOne` relationship with that. It would become much like the `Answer` model I guess.

Makes configuration more difficult, but we do not need to worry about that for now.

app/grandchallenge/reader_studies/models.py
amickan: Yes, well, we could call `_clean_answer_widget` on the form's instance in the clean method on the form, but that does require setting the attributes on the instance before doing so (otherwise, `answer_widget` and `interface` and `required` will still be the old values and not those entered in the form). That feels a bit hacky and strange to me, so I currently tend to moving these checks to the form instead.

app/grandchallenge/reader_studies/models.py
jmsmkn: Yes, I think a future developer would naturally extend `get_multiple_annotation_types` without realising that it would implicitly mean that CIRRUS would support it. The widget should define what CIRRUS supports.

amickan: Good point.

app/grandchallenge/reader_studies/models.py
amickan: I have removed the `_clean_answer_widget` now and will add these checks on the form in another PR.

app/grandchallenge/reader_studies/models.py
jmsmkn: @amickan did you see my second comment? Sorry!

amickan: I'm not entirely sure I follow along. The base class I added here is not abstract and the Question model does have a OneToOne realtionship with the base class in the current implementation. What's the issue with what we have now?

amickan: If we want to have configuration options in the future, doing it in one model like you suggest here will be messy.

amickan: The current implementation adds an unnecessary table for the base model to the database, but that's not really an issue, is it?

jmsmkn: We need to link the `Question` to a `Widget` somehow. If we have one table per `WidgetType` it can only be done with a GFK. If we have one table for all `Widget`s then it can be done with a `OneToOne` relationship.

Yes, it means that the `Widget` model becomes large, but it is a proven approach that we have taken with `Answer`, `Notification` and `ComponentInterface`.

amickan: > We need to link the `Question` to a `Widget` somehow. If we have one table per `WidgetType` it can only be done with a GFK. If we have one table for all `Widget`s then it can be done with a `OneToOne` relationship.

A concrete base class achieves that as well. The widget linked to the question is then an instance of the base class, but you can still access the subclasses methods and attributes.

amickan: GFKs are out of the question, I agree - I remember what a nightmare they were to work with for the notifications.

jmsmkn: Does that definitely work? You can add a `OneToOne` relationship to `AnswerWidget` and then set the value to that of a `AcceptRejectFindingsWidget`? What happens when we have other classes of widgets? Never tried this so have to be 100% certain this is going to work first.

amickan: > You can add a `OneToOne` relationship to `AnswerWidget` and then set the value to that of a `AcceptRejectFindingsWidget`?

Yes. 


> What happens when we have other classes of widgets?

They would all follow the pattern of the `AcceptRejectFindingsWidget`, so should be no different from working with just one subclass. 


I read this tutorial on the topic: https://realpython.com/modeling-polymorphism-django-python/#concrete-base-model 

Maybe we can quickly have a call about this, to see if I'm missing potential future scenarios that we should test with the current set-up?

jmsmkn: Would doing `question.widget` return an instance of the concrete class, or the base class?

If we had 10 questions, using 10 different widget concrete classes, how many extra queries would that be? I imagine that if we need the concrete classes it could need 10 joins across 10 different tables?

amickan: `question.widget` returns an instance of the base class. You can then access the subclass with `question.widget.subclassname` and yes, that requires joining across the subclass table. That in itself is maybe not so bad, but I just noticed that in order to be able to do that, you need to know which kind the widget is and the classmethod that I added `get_kind` returns None when called on the base class... so then it does become inefficient...

amickan: Ok, I'm glad we discussed this. I think I then agree that one single `Widget` model with a type field is the way to go after all.

jmsmkn: OK, I think that would be relatively common as we would need to serialize all the widgets for the reader study often. The article mentions that `django-polymorphic` addresses some of these issues but that project looks completely dead.

amickan: Indeed, not maintained, so a bit risky to use. I will work on creating one single `Widget` model.

jmsmkn: I've read the article a little more, was interesting. What we do for `ComponentInterfaceValues` and `Answers` covered by their [Semi Structured Model](https://realpython.com/modeling-polymorphism-django-python/#semi-structured-model) section. I still this this is the most appropriate approach. A lot of their downsides they mention are no longer true, and the trade-offs are well understood.

amickan: I agree

app/grandchallenge/reader_studies/models.py
```diff
@@ -1075,6 +1075,9 @@ class ScoringFunction(models.TextChoices):
     interface = models.ForeignKey(
         ComponentInterface, on_delete=models.PROTECT, null=True, blank=True
     )
+    answer_widget = models.ForeignKey(
```
jmsmkn, line 1078: Widgets should not be re-used between questions, so this should be a OneToOne relationship.

app/grandchallenge/reader_studies/models.py
```diff
@@ -1508,3 +1517,30 @@ def delete(self):
...
+
+
+class AnswerWidgetKindChoices(models.TextChoices):
+    ACCEPT_REJECT = "ACCEPT_REJECT", "Accept/Reject Findings"
+
+
+class AnswerWidget(models.Model):
+
+    kind = models.CharField(
+        max_length=255,
```
jmsmkn, line 1529:
```suggestion
        max_length=13,
```

app/grandchallenge/reader_studies/migrations/0038_auto_20230223_1327.py
```diff
@@ -0,0 +1,45 @@
...
+                        primary_key=True,
+                        serialize=False,
+                        verbose_name="ID",
+                    ),
+                ),
+                (
+                    "kind",
+                    models.CharField(
+                        choices=[("ACCEPT_REJECT", "Accept/Reject Findings")],
+                        max_length=255,
```
jmsmkn, line 30:
```suggestion
                        max_length=13,
```

app/grandchallenge/reader_studies/models.py
```diff
@@ -1508,3 +1517,30 @@ def delete(self):
 
     class Meta(RequestBase.Meta):
         unique_together = (("reader_study", "user"),)
+
+
+class AnswerWidgetKindChoices(models.TextChoices):
+    ACCEPT_REJECT = "ACCEPT_REJECT", "Accept/Reject Findings"
+
+
+class AnswerWidget(models.Model):
```
jmsmkn, line 1526: I think `QuestionWidget` would make more sense as these are widgets for questions, not answers. Using `question.widget` would make more sense than `question.answer_widget`.

app/grandchallenge/reader_studies/models.py
```diff
@@ -1075,6 +1075,9 @@ class ScoringFunction(models.TextChoices):
     interface = models.ForeignKey(
         ComponentInterface, on_delete=models.PROTECT, null=True, blank=True
     )
+    widget = models.OneToOneField(
```
jmsmkn, line 1078: Users can delete questions, so we should ensure that any widget is deleted too. You will need to use signals, see what we do with related groups for reader studies, challenges, etc, and ensure it's tested.

amickan: ah so the `on_delete` shouldn't be `PROTECT` because that would keep them from deleting the question if a widget is set, right?

amickan: oh no that's for the opposite way, deleting the widget wouldn't work because it's associated with a question

jmsmkn: Yes, could also be done the other way around - set `question = models.OneToOne(blank=False, null=False, on_delete=models.CASCADE, related_name="widget")`. That might be better actually as it always ensures that a widget has a question, but I don't think enforces that each question has a widget.

amickan: Hm, yes, but then it would be difficult to include the widget in the question serializer, wouldn't it?

jmsmkn: I don't see why, you could still do `question.widget` - that is what the `related_name` is about.

amickan: Yes, you're right. Sorry, got confused for a minute. This reverse set-up should also work just as well on the question form, so I think this a good design.

jmsmkn (approved):

PR #2780, opened 2023-02-23
Refactor algorithm job creation

### LINKED ISSUE #1903: Add validation on REST API for no algorithm credits left
When a user runs out of credits `remaining_jobs` returns a negative number, and then this is used to slice a django queryset. This then raises `AssertionError Negative indexing is not supported`. This needs to be handled, probably by checking that `remaining_jobs > 0`. See https://sentry.io/organizations/grand-challenge/issues/2489707837/?project=303639&query=is%3Aunresolved

We should also add a check to the REST API to return 403 if the number of remaining jobs are already < 1.

Not a blocker as still the jobs are not created, but, all the errors are not nice.

jmsmkn:
This PR refactors the entrypoints for algorithm job creation. It removes the session upload over the HTML forms functionality as it is only possible to use for a subset of algorithms with the single GMI input. Batch upload for all algorithms is available over the REST API instead.

The naming of views and forms are updated to reflect that they are only creating one job. References to "experiment" are removed. Earlier validation is added.

I could have gone further and turned the `JobCreateForm` to a `ModelForm` so that all of the logic is there rather than on the view (would make more sense) but that is too much work for now.

Closes #1903 


app/grandchallenge/algorithms/templates/algorithms/job_form_create.html
```diff
@@ -0,0 +1,77 @@
...
+                href="{{ algorithm.get_absolute_url }}">{{ algorithm.title }}
+        </a></li>
+        <li class="breadcrumb-item active"
+            aria-current="page">Try-out Algorithm
+        </li>
+    </ol>
+{% endblock %}
+
+{% block content %}
+    {% get_obj_perms request.user for algorithm as "algorithm_perms" %}
```
amickan, line 33: These permissions appear to be unused in this template - remove, right?

jmsmkn: Nice find

app/grandchallenge/algorithms/tasks.py
```diff
@@ -330,7 +259,6 @@ def create_algorithm_jobs(
     *,
     algorithm_image,
     civ_sets,
-    creator=None,
```
amickan, line 333: It is clear that this is only for creating anonymous jobs, but maybe -for dummies- that could be explicitly said in the name of the task or a short docstring?

amickan (approved):
Looks good to me!

PR #2781, opened 2023-02-23
Add chart of algorithm usage

jmsmkn:
Show algorithm editors how often their algorithm has been used. I have re-used the rendering of the runtime metrics chart. Bit of a hack, but as part of the statistics pitch I can standardise the rendering of vega charts.

image

### DIFF
--- a/app/grandchallenge/algorithms/models.py
+++ b/app/grandchallenge/algorithms/models.py
@@ -415,7 +415,7 @@ class Algorithm(UUIDModel, TitleSlugDescriptionModel, ViewContentMixin):
     @property
     def usage_chart_status_choices(self):
         """A map of int to string for Job.choice"""
-        return {k: v for k, v in Job.status.field.choices}
+        return dict(Job.status.field.choices)
 
     @cached_property
     def usage_statistics(self):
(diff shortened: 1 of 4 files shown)


app/grandchallenge/algorithms/models.py
```diff
@@ -407,6 +407,98 @@ def get_jobs_limit(self, user):
...
 
+    @property
+    def usage_chart_statuses(self):
+        """What statuses should be included on the chart"""
+        return [Job.SUCCESS, Job.CANCELLED, Job.FAILURE]
+
+    @property
+    def usage_chart_status_choices(self):
+        """A map of int to string for Job.choice"""
+        return {k: v for k, v in Job.status.field.choices}
```
amickan, line 418: Why not just `dict(ComponentJob.STATUS_CHOICES)`?

app/grandchallenge/algorithms/models.py
```diff
@@ -407,6 +407,98 @@ def get_jobs_limit(self, user):
...
+
+    @cached_property
+    def usage_statistics(self):
```
amickan, line 421: You're not calculating user specific usage stats, as described in the pitch. Is there a security concern with that, or are you planning on adding that later?

app/grandchallenge/algorithms/models.py
amickan: ah probably a circular import issue

jmsmkn: Because the relationship between the field `status` and `ComponentJob.STATUS_CHOICES` is defined by `Job.status.field.choices`, but indeed we could do `dict(Job.status.field.choices)`.

app/grandchallenge/algorithms/models.py
jmsmkn: Will maybe add it later.

PR #2783, opened 2023-02-24
Remove usage of ImageImportError

jmsmkn:
Using exceptions in celery tasks to communicate validation problems didn't work well, so this refactors the usage. It allows us to make more tasks idempotent by covering the entire task with an atomic block.


jmsmkn:
I'm unsure how well all of this is covered by tests. I have done my own manual testing and think everything is working as expected, but might be good for you to try and break it.

amickan (approved):
LGMT! I tried to break, but didn't manage. It appears to work just fine ;-) Reader studies and algorithms are covered by tests, but archives are not, so might be worth to add a test for archive item creation / updating with a faulty or missing image.

PR #2786, opened 2023-02-28
Enable selecting a widget for a question

amickan:
This PR adds a dynamic `widget` field to the `Question` form. The widget choices depend on the `answer_type` and are validated on form submission. The `Accept / Reject` widget requires a default answer to be set as well as unchecking the required box. The widget field is disabled once answers exist. 

I'm planning on adding more helptext to explain the requirements (one a per widget basis) a little better, but first want to get the main functionality reviewed. 

Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/191


app/grandchallenge/reader_studies/forms.py
```diff
@@ -310,6 +322,54 @@ def interface_choices(self):
...
+                    self.add_error(
+                        error=ValidationError(
+                            f'To use the {widget_instance.get_kind_display()} widget you must uncheck "required".'
+                        ),
+                        field=None,
+                    )
+            if not self.errors:
+                widget_instance.question = self.instance
+                widget_instance.save()
+                return widget_instance
```
amickan, line 371: At the moment, we're creating a new widget every time someone updates the question (and keeps the Accept / Reject widget chosen). I think it might be worth adding a check here if there already is a widget of the same type for that question and if so return that widget rather than saving the new one. For now this check is simple, in the future, if we have options that would get a little more involved.

app/grandchallenge/reader_studies/models.py
```diff
@@ -1544,3 +1545,13 @@ def supported_answer_types(self):
...
+
+    @classmethod
+    def get_widget_choices_for_answer_type(cls, answer_type):
+        widget_types = []
+        instance = cls()
+        for kind in QuestionWidgetKindChoices:
+            instance.kind = kind
+            if answer_type in instance.supported_answer_types():
+                widget_types.append((kind.name, kind.label))
+        return widget_types
```
amickan, line 1557: Maybe there is a better way to do this, but I thought this is at least better than adding a `ANSWER_TYPE_TO_QUESTION_WIDGET_MAP` dictionary which would require maintaining the mapping in two places (the dictionary and in `supported_answer_types`...

app/grandchallenge/reader_studies/forms.py
```diff
@@ -310,6 +322,54 @@ def interface_choices(self):
...
+        try:
+            return self.instance.widget.kind
+        except ObjectDoesNotExist:
+            return ""
+
+    def clean_widget(self):
+        value = self.cleaned_data.get("widget")
+        if value == "":
+            try:
+                self.instance.widget.delete()
```
jmsmkn, line 346: Should you be deleting objects in clean? I would have thought that should only be done on save?

app/grandchallenge/reader_studies/forms.py
jmsmkn: I think we need to change this somehow, we shouldn't create or delete objects during clean as the responsibility of this method is only to check the set of values. We definitely should not create a new object on every question update. Doing this in the save methods would be more appropriate. You may be able to use the save method directly, or you may have to work with the pre and post save signals.

app/grandchallenge/reader_studies/models.py
jmsmkn: How about:

```python
ANSWER_TYPE_TO_QUESTION_WIDGET = {
  AnswerType.MULTIPLE_2D_BOUNDING_BOXES: [QuestionWidgetKindChoices.ACCEPT_REJECT, ...]
  ...
}

...

    def supported_answer_types(self):
        return [k for k, v in ANSWER_TYPE_TO_QUESTION_WIDGET.items() if self.kind in v]
```

app/grandchallenge/reader_studies/forms.py
amickan: You're totally right. I'm working on handling the widget saving and deleting in the save method now, much cleaner.

app/grandchallenge/reader_studies/forms.py
```diff
@@ -299,6 +312,21 @@ def __init__(self, *args, **kwargs):
...
+            new_widget = self.cleaned_data["widget"]
+            try:
+                old_widget = instance.widget
+            except ObjectDoesNotExist:
+                old_widget = None
+            if old_widget.pk:
+                old_widget.delete()
+            if new_widget:
+                new_widget.save()
+                instance.widget = new_widget
```
jmsmkn, line 327: Why is `instance.widget = new_widget` necessary? I would have though that the changes would automatically get picked up.

app/grandchallenge/reader_studies/forms.py
```diff
@@ -310,6 +338,54 @@ def interface_choices(self):
...
+                if self.cleaned_data.get("required"):
+                    self.add_error(
+                        error=ValidationError(
+                            f'To use the {widget_instance.get_kind_display()} widget you must uncheck "required".'
+                        ),
+                        field=None,
+                    )
+            if not self.errors:
+                try:
+                    return self.instance.widget
```
jmsmkn, line 384: I don't really understand what is happening here. You're creating a new widget in memory, then validating that, but here returning the initial value? What if it is of a different kind? Is it not validated given the existing options?

app/grandchallenge/reader_studies/forms.py
amickan: Ah, you're right, this is not necessary!

app/grandchallenge/reader_studies/forms.py
amickan: Ah good catch, this was a remnant of the previous implementation. It doesn't really matter in the current implementation since `self.instance.widget` == `widget_instance`. The type is limited by the `widget_choices` and automatically validated. But I have removed this part, since really all we want to do here is send along the new widget instance.

app/grandchallenge/reader_studies/forms.py
```diff
@@ -299,6 +312,20 @@ def __init__(self, *args, **kwargs):
             )
         )
 
+    def save(self, *args, **kwargs):
+        instance = super().save(*args, **kwargs)
+        if "widget" in self.changed_data:
+            new_widget = self.cleaned_data["widget"]
+            try:
+                old_widget = instance.widget
```
amickan, line 320: I realized now that this does not work. `instance.widget` == `new_widget`. I'll need to find a different way of getting the preious widget value

amickan: Not setting the `question` on the `widget_instance` in the clean method should do the trick.

app/grandchallenge/reader_studies/forms.py
```diff
@@ -310,6 +337,52 @@ def interface_choices(self):
...
+                        ),
+                        field=None,
+                    )
+                if self.cleaned_data.get("required"):
+                    self.add_error(
+                        error=ValidationError(
+                            f'To use the {widget_instance.get_kind_display()} widget you must uncheck "required".'
+                        ),
+                        field=None,
+                    )
```
jmsmkn, line 382: OK, now things are getting clearer 🙂

Because this validation depends on two fields (widget and interface, and widget and required) in the form this should not be done in `clean_widget`, it must be done in `clean`. See [the Django docs](https://docs.djangoproject.com/en/4.1/ref/forms/validation/#raising-validation-error). 

> The form subclass’s clean() method can perform validation that requires access to multiple form fields. This is where you might put in checks such as “if field A is supplied, field B must contain a valid email address”.

app/grandchallenge/reader_studies/forms.py
```diff
@@ -299,6 +312,20 @@ def __init__(self, *args, **kwargs):
...
+    def save(self, *args, **kwargs):
+        instance = super().save(*args, **kwargs)
+        if "widget" in self.changed_data:
+            new_widget = self.cleaned_data["widget"]
+            try:
+                old_widget = instance.widget
+            except ObjectDoesNotExist:
+                old_widget = None
+            if old_widget.pk:
+                old_widget.delete()
```
jmsmkn, line 324: `if old_widget.pk:` won't this error out if `old_widget = None`? Maybe this is what is intended?

```suggestion
            try:
                instance.widget.delete()
            except ObjectDoesNotExist:
                pass
```

app/grandchallenge/reader_studies/forms.py
```diff
@@ -310,6 +337,52 @@ def interface_choices(self):
...
+        answer_type = self["answer_type"].value()
+        choices = [("", "Default")]
+        if answer_type:
+            try:
+                extra_options = ANSWER_TYPE_TO_QUESTION_WIDGET[answer_type]
+                choices.extend(
+                    (option.name, option.label) for option in extra_options
+                )
+            except KeyError:
+                pass
```
jmsmkn, line 351: We probably need an error here as the `ANSWER_TYPE_TO_QUESTION_WIDGET` would not have been updated by the programmer.

app/grandchallenge/reader_studies/forms.py
```diff
@@ -310,6 +337,52 @@ def interface_choices(self):
     def initial_interface(self):
         return self.interface_choices().first()
 
+    def widget_choices(self):
+        answer_type = self["answer_type"].value()
+        choices = [("", "Default")]
```
jmsmkn, line 343: You could use `BLANK_CHOICE_DASH` for consistency here.

app/grandchallenge/reader_studies/forms.py
```diff
@@ -310,6 +337,52 @@ def interface_choices(self):
...
+                )
+            except KeyError:
+                pass
+        return choices
+
+    def initial_widget(self):
+        try:
+            return self.instance.widget.kind
+        except ObjectDoesNotExist:
+            return ""
```
jmsmkn, line 358: Does this return value need to match the default choice? Same in L361 (`if value == ""`). If so, using `BLANK_CHOICE_DASH` would help readability.

app/grandchallenge/reader_studies/forms.py
```diff
@@ -310,6 +337,52 @@ def interface_choices(self):
...
         return self.interface_choices().first()
 
+    def widget_choices(self):
+        answer_type = self["answer_type"].value()
+        choices = [("", "Default")]
+        if answer_type:
+            try:
+                extra_options = ANSWER_TYPE_TO_QUESTION_WIDGET[answer_type]
+                choices.extend(
+                    (option.name, option.label) for option in extra_options
```
jmsmkn, line 348: Is this possible?

```suggestion
                    option.choice for option in extra_options
```

app/grandchallenge/reader_studies/forms.py
amickan: When Chris and I discussed this, we thought that `Default` would be clearer for the user than a dashed line. The dashed line indicates that no widget is used, which is not really the case, right?

app/grandchallenge/reader_studies/forms.py
amickan: Yes it does, but `BLANK_CHOICE_DASH` returns `[('', '---------')]` and you can't just get the `''`. I could alternatively add the default choice to the `QuestionWidgetKindChoices` choices and then use that in all places for consistency and clarity?

app/grandchallenge/reader_studies/forms.py
amickan: No, the `choice` attribute (or anything similar) doesn't exist on the option...

app/grandchallenge/reader_studies/forms.py
jmsmkn: It means that the user has not made a choice, this pattern exists in many places and we should be consistent. Other places have tried to re-invent this, it is a mess.

app/grandchallenge/reader_studies/forms.py
jmsmkn: `BLANK_CHOICE_DASH[0][0]`

app/grandchallenge/reader_studies/forms.py
```diff
@@ -310,10 +336,38 @@ def interface_choices(self):
...
+            return self.instance.widget.kind
+        except ObjectDoesNotExist:
+            return ""
+
+    def clean_widget(self):
+        value = self.cleaned_data.get("widget")
+        if value == "":
+            return None
+        else:
+            return QuestionWidget(kind=value)
```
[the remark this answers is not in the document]
amickan: The `full_clean()` method will fail here because the `question` is not set. We can't set the question at this point, because then we can no longer check and remove the old widget in the save method.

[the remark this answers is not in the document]
jmsmkn: Can you track the old widget in the model? Override the `__init__` method and add something like `self._widget_orig`, then check after `super().save(...)` in `save`?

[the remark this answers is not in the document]
jmsmkn: Actually, I wonder if we need a separate form all-together for the Widget as it seems we're having to hack around this a lot. Maybe an [inline form set](https://docs.djangoproject.com/en/3.2/topics/forms/modelforms/#inline-formsets) is what we need, with a separate form for the widget?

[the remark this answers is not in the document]
jmsmkn: Or, maybe we should be updating the existing widget rather than creating and deleting?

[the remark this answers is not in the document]
jmsmkn: Or maybe we should give up on this design and do everything with a JSON field on the Question model...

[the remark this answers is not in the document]
amickan: I will think some more and try some stuff.

app/grandchallenge/reader_studies/forms.py
amickan: I'm actually not sure if we want to throw an error here. We need to be able to return an emtpy list of options for a lot of answer types, and the easiest way to do this is to catch the KeyError here, right? The KeyError is also not very informative, a developer could forget to update the mapping and not get a KeyError if it happens to be for one of the already included answer types...

jmsmkn: The key error comes from looking up the answer type in `ANSWER_TYPE_TO_QUESTION_WIDGET`, so to return an empty list of options for a particular type you would define

```python
ANSWER_TYPE_TO_QUESTION_WIDGET = {
  ...
  NoWidgetAnswerType = []
}
```

If the `KeyError` is not informative then you transform it:

```python
try:
   foo["bar"]
except KeyError as e:
   raise MySpecialException("You need to update foo") from e
```

If a developer forgets to update the mapping, then yes, no way to catch that other than by testing.

amickan: Yes, but if we always raise an error whenever the answer_type is not in the mapping dictionary, then a user will be unable to update the question in those cases.

jmsmkn: Yes, they shouldn't be able to as the application has not be programmed correctly.

amickan: They should be able to update a question regardless of what the answer type is, they just shouldn't be able to define a widget for them. If we raise an error whenever there is a KeyError, they would not be able to change anything on the quesiton, the form would not load because the widget_choices() are evaluated every single time you load the question form.

app/grandchallenge/reader_studies/forms.py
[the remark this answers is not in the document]
amickan: Setting `self._widget_orig` doesn't help us here. The `full_clean` method would still fail if we create a new widget for a question that already has a widget, because you can only have one widget associated with a particular quesiton. So for the full_clean method to work we would need to in fact delete the old widget before we call it, which would mean deleting the object in the `clean` method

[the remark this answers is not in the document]
amickan: I'm looking into formsets now. They might be a good option, especially if we want to add options to the widget model in the future, but I still need to work out how to update the widget choices within the formset depending on answer_type.

app/grandchallenge/reader_studies/forms.py
jmsmkn: One of the foundational Python principles is that "Errors should never pass silently.". If you leave this as it is you open yourself up to bugs that are difficult to track down, as you are never aware of where the application is going wrong.

Yes, if the error is raised no one will be able to update the form. But, that is exactly the behaviour we want as the proper solution to this is a developer going back and updating `ANSWER_TYPE_TO_QUESTION_WIDGET` as we know it is out of date.

amickan: I don't follow the logic here, sorry. The mapping is already out of date unless we add all answer types to it and if we add all answer types, there will never be a KeyError. Sorry, but I think I'm either missing something here, or I don't understand the principal.

amickan: Oh, there will be a KeyError if we add a new answer_type. Is that what you mean?

jmsmkn: Yes, `ANSWER_TYPE_TO_QUESTION_WIDGET` should contain all answer types.

app/grandchallenge/reader_studies/forms.py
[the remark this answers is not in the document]
amickan: Formsets don't seem easier to work with than what we have here in terms of validation or updating the choices per answer_type, so I'm hesitant to move towards using formsets. 

Here is a summary of the options I think we have: 
1. Use the current implementation, and do `full_clean(exclude="question")` in the form's `clean` method. That defeats the purpose of calling `full_clean` though. 
2. Use current implementation and only call the `full_clean` method on the final model instance in the form's `save` method - also not great since that's not what the `save` method should be used for. 
3. Do what we currently do for attaching default answers to a question. Users have to contact support for a new widget, which we create in the admin and they can choose from a list of filtered options. Relatively simple to implement in the form (using a dynamic ModelChoiceField), and since we'd only have one single widget for now, also no extra admin overhead for the time being. 
4. Drop the `QuestionWidget` model entirely and instead use a json field on the `Question` model to define the widget. We would define a json schema, for now super simple, but can be made as complex as we need it to be in the future. Gives the user more autonomy than option 3, but if we end up with lots of configuration options, validation gets complex.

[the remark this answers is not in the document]
amickan: I tend towards option 3 right now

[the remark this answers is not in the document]
jmsmkn: I don't think that widgets should be shared across questions, too easy to change and then you affect multiple questions at once. The workflow of asking for a widget from support does not scale with options, so when that happens the same issues that we face now would need to somehow be solved.

I really think that changing the model is the way to go, how about:

- `question.widget = models.CharField(choices=QuestionWidgetChoices, ...)`

Then in the future we add `question.widget_options` as a JSON field. It's really not that different to the approach we're going for right now.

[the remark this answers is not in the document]
amickan: Ok, so then we get rid of the `QuestionWidget` model and only add a `widget` choice field on the `Question` model for now.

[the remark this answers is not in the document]
jmsmkn: Yes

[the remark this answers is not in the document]
jmsmkn: Probably should have gone for that when we realised that multiple concrete models were not feasible.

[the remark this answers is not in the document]
amickan: Probably, I didn't foresee the issues it would cause with validation on the form, my bad. I think it's probably best if I close this PR and start a new one, right?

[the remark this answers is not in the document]
jmsmkn: Ah it's OK, my fault too, at least we got to a good design before it's used. Up to you whether you want to create a new one or re-work this.

PR #2797, opened 2023-03-07
Fix creation of DZI images

jmsmkn:
Tasks can be interrupted when copying files to storage. As we have to exit cleanly, this can leave files dangling with the prefix of `images/{image_pk}/`.

If this happens for MHA or TIFF files, this is no problem as when the task is re-run a new images are created with a new `image_pk`s. Still the old ones hang around, they should be cleaned up at some point, but they do no affect the running on the system.

Now that the creation of DZI has been separated out into a separate task, if that is interrupted then files will be left around at `images/{image_pk}/{image_pk}.dzi` and in `images/{image_pk}/{image_pk}_files/`. However, when the task is re-run here, the `image_pk` does not change, so when looking for a spare filename Django will append some random letters to all of the filenames. This breaks the DZI viewing as it assumes a filename structure.

This PR resolves this issue by adding the `{image_file_pk}` to the storage prefix, such that all image files are now stored in `images/{image_pk}/{image_file_pk}/`. This fixes the DZI issue as when the task is run again the DZI is assigned a new `image_file_pk`.

Still the files created by interrupted tasks should be cleaned up, but this does not cost us much. Permissions to read files should not be affected as users with `view_image` permission can download anything with prefix `images/{image_pk}/`. Some applications may parse out parts of the file URL so would need to be updated, but I don't see why they should be doing that.


miriam-groeneveld:
I have tested uploading various formats and batches of files, all seem to work as expected. Will try in combination with cirrus, but having slow internet issues at the moment, and need to get the cirrus image from doduo. Hope to have results at the end of today

miriam-groeneveld (approved):
LGTM. I've tested with various uploads (large tiff, batch uploads, via UI and via gcapi), didn't find any issues.

PR #2798, opened 2023-03-08
Track total submissions per challenge phase

amickan:
This adds a `total_number_of_submissions_allowed` field to the `Phase` model to track submissions for an algorithm submission phase.  The value of this field needs to be set in the Django admin to the number of teams * max number of submission per team for the phase as recorded on the challenge request.  
When that limit is reached, the phase is automatically closed for submissions. 

The limits can be tracked on the challenge cost page: 

image

And on the challenge stats page:

image

Closes https://github.com/DIAGNijmegen/rse-roadmap/issues/208


app/grandchallenge/evaluation/models.py
```diff
@@ -437,6 +437,10 @@ class Phase(UUIDModel, ViewContentMixin):
         blank=True,
         on_delete=models.SET_NULL,
     )
+    number_of_submissions_limit = models.PositiveSmallIntegerField(
```
jmsmkn, line 440: I think we need to differentiate this from `submissions_limit` more. Maybe `total_number_of_submissions_allowed`? Then maybe add a help text that this is the limit on the total number of submissions for this phase, not per user. Maybe we should also rename `submissions_limit`?

jmsmkn (commented):
Looks good, minor things.

app/grandchallenge/evaluation/models.py
```diff
@@ -640,7 +637,34 @@ def submission_period_is_open_now(self):
...
+            return min(
+                round(
+                    (
+                        self.submission_set.count()
+                        / self.number_of_submissions_limit
+                    )
+                    * 100
+                ),
+                100,
+            )
```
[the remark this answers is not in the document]
amickan: Fair enough, might indeed be useful for currently ongoing challenge phases (once we've set a limit for them that is). I had set the limit to 100 because of the progress bar I use to display these, but that's not a strict requirement of course. 

If I understand the last sentence here correctly, it should not be necessary to turn this into a cached_property since we're prefetching the `submission_set` for all challenge phases in all views where this is used https://docs.djangoproject.com/en/4.1/ref/models/querysets/#django.db.models.query.QuerySet.count  Did I understand this correctly?

app/grandchallenge/evaluation/models.py
amickan: Good point, a helptext and renaming it to `total_number_of_submissions_allowed` sounds like a good idea. `submissions_limit` already has a very clear help text, so I think it's not strictly necessary to rename this field, but we could change it to `user_submission_limit`?

app/grandchallenge/evaluation/models.py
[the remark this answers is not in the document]
jmsmkn: You can double check by either using the debug toolbar or by using `python manage.py shell_plus --print-sql`

app/grandchallenge/evaluation/models.py
jmsmkn: Sounds find, could even go more verbose with `user_number_of_submissions_allowed_per_period`.

amickan: `submissions_limit_per_user_per_period` sounds good, for the total limit I prefer `total_number_of_submissions_allowed`

amickan: Changing `submission_limit` is quite a bit of refactoring. Should we do this in a seperate PR so that this one stays clean?

jmsmkn: Yes separate PR and then rebase.

app/grandchallenge/evaluation/models.py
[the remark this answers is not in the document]
amickan: We're good! No unnecessary duplicate queries.

app/grandchallenge/evaluation/models.py
```diff
@@ -640,7 +638,31 @@ def submission_period_is_open_now(self):
...
+
+    @property
+    def submission_limit_reached(self):
+        return (
+            self.percent_of_total_submission_limit
+            and self.percent_of_total_submission_limit >= 100
+        )
+
+    @property
+    def percent_of_total_submission_limit(self):
```
jmsmkn, line 655: Looks good, might be worth to rename the properties too now the underlying field has changed name. So `percent_of_total_submissions_allowed` and `exceeds_total_number_of_submissions_allowed` ? Similar things for the 3 new challenge properties. Just thinking that if you don't have `submission_limit` anywhere in this PR the refactoring of the other field becomes easier. What do you think?

amickan: True, I had actually thought about changing the name to that but then thought it was unnecessarily long - but I see the value in keeping it consistent with the field name to avoid confusion. Will change them.

jmsmkn (approved):

PR #2799, opened 2023-03-08
Update challenge request accept email

amickan:


miriam-groeneveld (approved):
Looks good to me!

jmsmkn:
Also for me. These snippets are now really long so it would be nicer to separate this out into `.txt` template files and use `render_to_string` with a context instead.

amickan:
> Also for me. Generally, these snippets are now really long so it would be nicer to separate this out into `.txt` template files and use `render_to_string` with a context instead.

Good idea. I will still merge and deploy this now, just because we are in the middle of accepting and requests and that change needs to be made asap. I will then create another PR to use txt files instead.

PR #2802, opened 2023-03-08
Allow data-target attribute on anchor tags

amickan:


jmsmkn (approved):
Please add a description to the PR explaining why this change is necessary before you merge.

PR #2811, opened 2023-03-15
Allow answer modification without navigation

jmsmkn:
This combination was though not to work in CIRRUS, but it has been requested by a user and determined that it is not a problem.


miriam-groeneveld (approved):
LGTM!

PR #2812, opened 2023-03-15
Use widget value instead of label in QuestionSerializer

HarmvZ:
Using the value in the serializer allows us to change the label without breaking functionality in CIRRUS. (As we _should_ have done with answer_type, direction and image_port).

https://github.com/comic/grand-challenge.org/blob/f654c41fbc6850eac0172baa90b0a009746cee34/app/grandchallenge/reader_studies/models.py#L913


jmsmkn:
It was intentional, doing it this way ties it to the database representation which means we cannot change it internally.

PR #2816, opened 2023-03-28
Add a Number Input Widget for Reader Studies

amickan:
This PR adds a `NumberInput` widget for reader study questions. This new widget can only be selected for answers of type `Number`. The user can optionally configure `min`, `max` and `step size` values for this widget, which if defined get validated for answers for this questions. 

Addresses the GC side of https://github.com/DIAGNijmegen/rse-roadmap/issues/220


chrisvanrun:
There might be a need for some additional validation:

* _When step size, min, and max are set the difference between min and max needs to be neatly divisible by the step size._
 Otherwise, you end up having to make a choice between step size and (min/max) value. For instance, if you have min:0, max:2, and step:0.6. You'll end up with a value range of: `[0, 0.6, 1.2, 1.8, ... ]`.

* _When the step size is set: either min or max values need to be set._
Can't really validate a number with step size alone since that defines a relative value. You need an absolute value to base the steps. For instance, a `0.4` step size and a max value of `20.4` should validate during runtime to allow for `20`, but not for `0`.

Alternatively, with the latter validation, we could assume `0` as a base if neither max nor min is set. This might be an OK assumption. Thoughts?

app/grandchallenge/reader_studies/models.py
```diff
@@ -1447,14 +1491,35 @@ def validate(  # noqa: C901
...
-        ):
-            raise ValidationError(
-                "Answer for required question cannot be None"
-            )
+        if question.answer_type == Question.AnswerType.NUMBER:
+            if question.required and answer is None:
+                raise ValidationError(
+                    "Answer for required question cannot be None"
+                )
+            if question.answer_min_value:
```
jmsmkn, line 1499: These 3 checks on the values should be `is not None`.

app/grandchallenge/reader_studies/serializers.py
```diff
@@ -74,8 +76,19 @@ class Meta:
             "overlay_segments",
             "look_up_table",
             "widget",
+            "widget_options",
         )
 
+    def get_widget_options(self, obj: Question) -> dict:
```
jmsmkn, line 82: I think the API is clearer if the values are just serialized directly rather than grouped up into a dictionary, makes it easier for the client to work with.

app/tests/reader_studies_tests/test_models.py
```diff
@@ -417,6 +420,73 @@ def test_validate_answer():
...
+            pytest.raises(ValidationError),
+        ),
+        (
+            AnswerType.NUMBER,
+            4.7,
+            {
+                "answer_min_value": 1,
+                "answer_max_value": 5,
+                "answer_step_size": 0.5,
+            },
```
jmsmkn, line 453: You probably need some more tests with some zeroes and negative numbers :-)

jmsmkn (commented):
Looks good!

miriam-groeneveld:
We need to validate the default answers provided in the display sets as well. That can be tricky, as these can be added before or after the questions are created.

amickan:
> There might be a need for some additional validation:
> 
> * _When step size, min, and max are set the difference between min and max needs to be neatly divisible by the step size._
>   Otherwise, you end up having to make a choice between step size and (min/max) value. For instance, if you have min:0, max:2, and step:0.6. You'll end up with a value range of: `[0, 0.6, 1.2, 1.8, ... ]`.
> * _When the step size is set: either min or max values need to be set._
>   Can't really validate a number with step size alone since that defines a relative value. You need an absolute value to base the steps. For instance, a `0.4` step size and a max value of `20.4` should validate during runtime to allow for `20`, but not for `0`.
> 
> Alternatively, with the latter validation, we could assume `0` as a base if neither max nor min is set. This might be an OK assumption. Thoughts?

Makes sense! Assuming 0 for the second validation seems fair to me, but it's no effort to add a check in the code, and it feels more user friendly to make this explicit to the user, so I will add it as well.

app/grandchallenge/reader_studies/serializers.py
jmsmkn: Yes, the client can then check all the ones that are not null, just like we do server side.

chrisvanrun:
> We need to validate the default answers provided in the display sets as well. That can be tricky, as these can be added before or after the questions are created. Also, as these might come from an algorithm job, maybe just a warning is sufficient, and Cirrus should handle these cornercases.

This can only be done during runtime and hence is the responsibility of CIRRUS.

pkcakeout:
My 2 cents: do not blow the feature out of proportion. I think we do not really need to validate the default_answer in the displaysets. It will cause an obvious problem to the user and the reader-study editor should this occur, and also probably is an obvious error in their input data. We still catch the error, but late in the reader study causing a sub-par experience for the readers.

But I think the principle is sound: We do not really corrupt data if we only check answers when the users enters a new one, for the reader study editor it holds that: Garbage data in, garbage reader study out :D

amickan:
I tend to agree with Paul and Chris here. While we _could_ validate any possibly existing default answers on GC, it would be quite an expensive check that would need to be done in multiple places. For the AcceptReject Widget we also don't do that (we only check that an interface is selected for the question, but not whether there are display sets with that interface for the reader study). I would say that this is indeed the responsibility of the editor to provide sensible and correct default answers when needed.

miriam-groeneveld:
> I tend to agree with Paul and Chris here. While we _could_ validate any possibly existing default answers on GC, it would be quite an expensive check that would need to be done in multiple places. For the AcceptReject Widget we also don't do that (we only check that an interface is selected for the question, but not whether there are display sets with that interface for the reader study). I would say that this is indeed the responsibility of the editor to provide sensible and correct default answers when needed.

Yes, let's leave it to Cirrus. It's also just a CIV, so in itself is a valid object. Only using the value to create an answer should not work, so that's up to Cirrus.

chrisvanrun:
@amickan did you consider this one? I think it should throw an exception.
```Python
 {
      "answer_min_value": 4,
      "answer_max_value": 0,
  },
```

Also what about negative step sizes? I think they are invalidated already but a test should cover it.

amickan:
> @amickan did you consider this one? I think it should throw an exception.
> 
> ```python
>  {
>       "answer_min_value": 4,
>       "answer_max_value": 0,
>   },
> ```
> 
> Also what about negative step sizes? I think they are invalidated already but a test should cover it.

Good one. I'll add validation that max_value needs to be bigger than min_value. 
Negative step sizes are indeed invalidated already, but I'll add a test for that as well.

app/grandchallenge/reader_studies/models.py
```diff
@@ -1268,6 +1296,56 @@ def _clean_widget(self):
...
+
+            if is_min_value_set and is_max_value_set:
+                try:
+                    range = self.answer_max_value - self.answer_min_value
+                    StepValueValidator(self.answer_step_size)(range)
+                except ValidationError:
+                    raise ValidationError(
+                        "The difference between min and max answer value is not "
+                        "a multiple of the step size."
+                    )
```
jmsmkn, line 1338: I've tried to work on the readability of this method. I do not see why this step size validation block is needed. Using:

```html
<form>
  <input type="number" step="0.3">
</form>
```

or

```html
<form>
  <input type="number" step="0.3" min="-1" max="1">
</form>
```

are both perfectly valid. We should keep to the HTML specifications.

app/tests/reader_studies_tests/test_models.py
```diff
@@ -570,3 +650,116 @@ def test_clean_question_widget(answer_type, widget, interface, error):
...
+            QuestionWidgetKindChoices.NUMBER_INPUT,
+            {
+                "answer_min_value": 0,
+                "answer_max_value": 0,
+            },
+            pytest.raises(ValidationError),
+        ),
+    ),
+)
+def test_clean_widget_options(widget, options, error):
```
jmsmkn, line 755: This test would be easier to debug if you also included part of the validation error you expect to find, at the moment Pytest only tells you that test n failed.

amickan: I have also removed the test that checks the database constraints for the step_size (that it needs to be higher than 0.001). We don't usually test those types of constraints since we can trust Django to enforce them.

jmsmkn (approved):

PR #2818, opened 2023-03-30
Fix OpenAPI Schema for client generation

jmsmkn:
Our OpenAPI Schema had a few errors in it related to read or write only fields, this PR fixes them. It also splits the request and response models in the schema to allow for easier generation of clients, see https://drf-spectacular.readthedocs.io/en/latest/client_generation.html for motivation.


jmsmkn:
>  already behave like write-only fields in that they are not serialized (on read operations)

This is because they do not exist on the model, they are only used for setting arguments to the image import task. When you serialise the values back that information is missing, so DRF must just omit it. That's fine, but drf-spectacular wasn't clever enough to work that out - which seems fine to me as setting `write_only` is the way to communicate that behaviour.

PR #2823, opened 2023-03-31
Upgrade to Django 4.1

jmsmkn:
This PR upgrades Django to 4.1 (4.2 LTS is too fresh for us right now). Release notes:

- 4.0: https://docs.djangoproject.com/en/4.1/releases/4.0/
- 4.1: https://docs.djangoproject.com/en/4.1/releases/4.1/

The main issues we had to change:

## CSRF

Implemented the required changes to `CSRF_TRUSTED_ORIGINS`

https://docs.djangoproject.com/en/4.1/releases/4.0/#csrf
https://docs.djangoproject.com/en/4.1/releases/4.0/#csrf-trusted-origins-changes-4-0
https://docs.djangoproject.com/en/4.1/releases/4.0/#csrf-trusted-origins-changes

## Delete Views

`DeleteView` now works with `POST` and not `DELETE` requests. I changed a bunch of delete methods to use `SuccessMessageMixin` instead of adding the message manually in `delete`. In the reader studies app there are 3 bulk delete views that were using `DeleteView`, I have updated those to use a `FormView` instead as the views no longer work with the `DeleteView` that requires a single object. I would suggest manually testing the following views:

- `AnswersRemoveForUser` (especially with answers for other users)
- `AnswersRemoveGroundTruth` (especially with non-ground truth answers for the same user)
- `QuestionDelete` (especially with existing answers)

The tests were updated to use `client.post` rather than `client.delete`. We need to check that any HTMX/JS stuff does the same. I also noticed some incorrect signatures on `model.delete` so fixed those.

https://docs.djangoproject.com/en/4.1/releases/4.0/#generic-views
https://docs.djangoproject.com/en/4.1/releases/4.0/#deleteview-changes

## Migrations

Some noop migrations have been introduced.

https://docs.djangoproject.com/en/4.1/releases/4.0/#migrations-autodetector-changes

## Backport Removal

The backport of `StepSizeValidator` has been removed as that is now available directly from Django.


app/grandchallenge/reader_studies/views.py
```diff
@@ -847,12 +846,13 @@ def get_context_data(self, **kwargs):
         return context
 
 
-class AnswerBatchDelete(LoginRequiredMixin, DeleteView):
+class AnswerBatchDelete(LoginRequiredMixin, SuccessMessageMixin, FormView):
```
amickan, line 849: I think you're missing a `template_name` for this view to work.

amickan: Also, the `FormView` does not except `DELETE` as a method, so you need to change the method to `POST` in `remove_answers.js`

amickan: Also needs to be changed in `readerstudy_detail.html` for deleting the ground truth

jmsmkn: > I think you're missing a template_name for this view to work.

OK that's strange as the test for it pass.

app/grandchallenge/reader_studies/views.py
```diff
@@ -869,17 +869,16 @@ def check_permissions(self, request):
...
-                "HX-Refresh": True,
-            },
-        )
+
+        response = super().form_valid(*args, **kwargs)
+
+        response["HX-Redirect"] = self.get_success_url()
+        response["HX-Refresh"] = True
+
+        return response
```
amickan, line 881: This doesn't work properly. For the `HX-Redirect` header to work properly you need to use a HttpResponse instead of the HtpRedirectResponse from Django. Alternatively, if you want to stick to using HttpRedirectResponse, you need to remove the `Location` header as well as the `HX-Redirect` response header. 

And it needs to be `response["HX-Refresh"] = "true"` not `response["HX-Refresh"] = True`

app/grandchallenge/reader_studies/views.py
amickan: I misdiagnosed the problem originally. The issue is with the method, not the missing template name. Sorry, forgot to delete the first comment.

amickan: But still strange that the tests pass, or do they use `client.post`?

jmsmkn: They were updated to use `client.post`.

amickan:
I looked through the changes, and thoroughly tried out the reader studies views. I have left comments on the things that don't work yet. Everything else looks good to me! :-)

app/grandchallenge/reader_studies/views.py
jmsmkn: OK both issues should be fixed, could you test them again to verify?

amickan: All views work as expected!

PR #2825, opened 2023-04-05
Fix map keys

jmsmkn:
CI started failing about an hour ago due to an upgrade of docker in GitHub actions. The latest version of docker compose includes a stricter yaml parser. Defining `<<` multiple times in the same block is not strictly allowed, but previously worked. This moves the environment variables to their own blocks and uses the list expansion to include multiple blocks instead.


jmsmkn:
I'm going to merge this as it blocks everyone.

PR #2827, opened 2023-04-05
Add option for algorithm editors to mark old image as active

amickan:
This is the second part of https://github.com/DIAGNijmegen/rse-roadmap/issues/223 

It updates the `latest_executable_image` methods on `Algorithms`, `Phases` and `Workstations` to take into account `is_desired_version` and renames this property to `active_image`. 

This also adds a button on the algorithm image detail view that enables algorithm editors to mark an image of their choice as the desired version for an algorithm and some small UI additions related to that.

image

image


app/grandchallenge/algorithms/models.py
```diff
@@ -341,18 +341,21 @@ def assign_workstation_permissions(self):
                     perm=perm, user_or_group=group, obj=self.workstation
                 )
 
-    @cached_property
-    def latest_executable_image(self):
+    @property
```
jmsmkn, line 344: Why is this no longer cached? Would also need to update Methods and Workstations to be cached again.

app/grandchallenge/algorithms/views.py
```diff
@@ -335,6 +336,39 @@ def get_context_data(self, *args, **kwargs):
         return context
 
 
+class AlgorithmImageMarkDesired(
+    LoginRequiredMixin, ObjectPermissionRequiredMixin, View
+):
+    permission_required = "algorithms.change_algorithmimage"
```
jmsmkn, line 342: Should this be change algorithm, not change algorithm image?

app/grandchallenge/algorithms/views.py
```diff
@@ -335,6 +336,39 @@ def get_context_data(self, *args, **kwargs):
...
+    raise_exception = True
+
+    @cached_property
+    def algorithm(self) -> Algorithm:
+        return get_object_or_404(Algorithm, slug=self.kwargs["slug"])
```
jmsmkn, line 347:
```suggestion
        return self.algorithm_image.algorithm
```

app/grandchallenge/algorithms/views.py
```diff
@@ -335,6 +336,39 @@ def get_context_data(self, *args, **kwargs):
...
+
+    @cached_property
+    def algorithm_image(self) -> Algorithm:
+        return get_object_or_404(AlgorithmImage, pk=self.kwargs["pk"])
+
+    def get_permission_object(self):
+        return self.algorithm_image
+
+    def get(self, *args, **kwargs):
```
jmsmkn, line 356: The database should not be changed in a GET view, use a POST view instead.

app/grandchallenge/algorithms/views.py
```diff
@@ -335,6 +336,39 @@ def get_context_data(self, *args, **kwargs):
...
+        self.algorithm_image.mark_desired_version()
```
jmsmkn, line 357: What happens here if the selected image is not in ECR or on SageMaker?

app/grandchallenge/algorithms/views.py
amickan: Hm, I figured since we're modifying a specific algorithm image here, it should be change algorithm image (in line with our algorithm image update view).

app/grandchallenge/algorithms/views.py
amickan: It would still get marked as desired (the only we check in that method is whether `is_manifest_valid` is True. I'm only making the button available to users when an image is also on ECR and SageMaker, but I will add a check here as well.

app/grandchallenge/algorithms/models.py
amickan: Accidentally changed this, sorry!

app/grandchallenge/algorithms/views.py
jmsmkn: I know, we could argue either way. I think it should be for the algorithm as ultimately you are changing which image is active for it. That involves changing multiple algorithm images, not just the one you're checking permissions on right now.

app/grandchallenge/algorithms/views.py
jmsmkn: OK - but we will need it to work for images not on ECR and SageMaker too as they get cleaned up every evening.

jmsmkn: We shouldn't call validate docker image again, the validation takes a long time and doesn't need to be repeated if it has already been done.

app/grandchallenge/algorithms/views.py
```diff
@@ -339,22 +341,47 @@ def get_context_data(self, *args, **kwargs):
...
+        if not self.algorithm_image.is_manifest_valid:
+            messages.error(
+                request, "Cannot mark an invalid image as active image."
+            )
+            return HttpResponseRedirect(
+                self.algorithm_image.get_absolute_url()
+            )
+        if not self.algorithm_image.can_execute:
+            try:
+                upload_to_registry_and_sagemaker(self.algorithm_image)
```
jmsmkn, line 368: This needs to be done asynchronously. Uploading a file from S3 to ECR can take up to 20 minutes as these images are often massive. You need to be really careful here that this is only being done once per algorithm image or algorithm.

amickan: I'm unsure how to make sure this only runs once per image. I would need to be able to check whether there is already an sync task scheduled or running for this image, right?

jmsmkn: Yes, much like we do for the initial importing.

app/grandchallenge/components/models.py
```diff
@@ -1663,7 +1675,12 @@ def get_peer_images(self):
 
     @transaction.atomic
     def mark_desired_version(self):
-        if self.is_manifest_valid:
+        try:
+            del self.can_execute
+        except AttributeError:
+            pass
+
+        if self.is_manifest_valid and self.can_execute:
```
jmsmkn, line 1683: I was really confused by this, maybe create:

```
def clear_can_execute_cache(self):
   ...
```

next to

```
@cached_property
def can_execute(self)
   ...
```

app/grandchallenge/algorithms/views.py
```diff
@@ -335,6 +338,82 @@ def get_context_data(self, *args, **kwargs):
...
 
 
+class AlgorithmImageMarkDesired(
+    LoginRequiredMixin, ObjectPermissionRequiredMixin, View
+):
+    permission_required = "algorithms.change_algorithm"
+    raise_exception = True
+
+    @cached_property
+    def algorithm(self) -> Algorithm:
```
jmsmkn, line 348: FYI, I know that they're there already in some places but we shouldn't add type annotations as they're not being checked, it's worse to have them as they could drift and be wrong.

app/grandchallenge/algorithms/views.py
```diff
@@ -335,6 +338,82 @@ def get_context_data(self, *args, **kwargs):
...
+        return self.algorithm_image.algorithm
+
+    @cached_property
+    def algorithm_image(self) -> AlgorithmImage:
+        return get_object_or_404(AlgorithmImage, pk=self.kwargs["pk"])
+
+    def get_permission_object(self):
+        return self.algorithm
+
+    def post(self, request, *args, **kwargs):
```
jmsmkn, line 358: There is a lot of logic in the post method and its is difficult to follow. Do you think that a `FormView` would be better? Then you can use the forms validation framework. For instance, I don't think that `HttpResponseRedirect` is the correct thing to return for a validation issue, I think this should be handled in a form. All of the validation checks should be moved there.

app/grandchallenge/algorithms/views.py
```diff
@@ -335,6 +338,82 @@ def get_context_data(self, *args, **kwargs):
...
+        if not self.algorithm_image.can_execute:
+            image_upload_in_progress = (
+                self.algorithm_image.import_status
+                == ImportStatusChoices.STARTED
+            )
+            if not image_upload_in_progress:
+                self.algorithm_image.import_status = (
+                    ImportStatusChoices.STARTED
+                )
+                self.algorithm_image.save()
```
jmsmkn, line 376: This should not be set here, this should only be set when a worker claims the job. You may need to add another attribute to the container image model to track this, or maybe it should be set to pending and you check for

```python
            image_upload_in_progress = (
                self.algorithm_image.import_status
                in (ImportStatusChoices.STARTED, ImportStatusChoices.PENDING)
            )
```

Additionally, `image_upload_in_progress` needs to check ALL images for the related algorithm to prevent contesting imports, so something like:

```python
@property
def image_upload_in_progress(self):
    return self.algorithm_image_set.filter(status__in=(ImportStatusChoices.STARTED, ImportStatusChoices.PENDING)).exists()
```

on the algorithm model.

app/grandchallenge/algorithms/views.py
```diff
@@ -335,6 +338,82 @@ def get_context_data(self, *args, **kwargs):
...
+                        "model_name": self.algorithm_image._meta.model_name,
+                        "pk": self.algorithm_image.pk,
+                        "mark_as_desired": True,
+                    }
+                ).apply_async()
+                messages.success(
+                    request,
+                    "Image validation and upload to registry in progress. "
+                    "It can take a while for this image to become active, please be patient.",
+                )
```
jmsmkn, line 389: With a `FormView` you should then use `SuccessMessageMixin`.

app/grandchallenge/algorithms/views.py
```diff
@@ -335,6 +338,82 @@ def get_context_data(self, *args, **kwargs):
...
+                messages.warning(
+                    request, "Image updating already in progress."
+                )
+            return HttpResponseRedirect(
+                self.algorithm_image.get_absolute_url()
+            )
+
+        if (
+            self.algorithm_image.is_manifest_valid
+            and self.algorithm_image.can_execute
```
jmsmkn, line 400: Do you think that these guards could be moved to `mark_desired_version` as we always want that to be true.

jmsmkn (commented):
Looking good! A few comments as there are a few corner cases.

app/grandchallenge/algorithms/views.py
amickan: Good point about checking for contesting imports! I think the "pending" (queued) status works well, but I will still need to set this here on the view - otherwise, calling the view a second time when the task hasn't been picked up by a worker yet would result in another task being scheduled.

app/grandchallenge/algorithms/views.py
amickan: They are already in the method. I can just remove them here actually.

app/grandchallenge/algorithms/views.py
amickan: Hm, not super straightforward the FormView. I think I would then need a form without fields and an empty placeholder template (FormView needs a template) and would do the validation in the clean method on the form, return ValidationErrors accordingly, and then implement the `upload_to_registry_and_sagemaker` method in `form_invalid` (but only if a specific validation error was returned) and call `mark_desired_version` in `form_valid`.

amickan: in `form_invalid` I would still need to return an HttpResponse with a message that contains either the errors or the success message that uploading to the registry is happening. Returning the form with errors, like you would usually do, doesn't work here since there is no form to return.

jmsmkn: You would need a form with an algorithm image right? e.g. the one that you want to mark as active.

amickan: Maybe I'm missing something, but no matter what the form looks like, it doesn't change the fact that I need to change `form_invalid` to return the error messages properly. We don't want to display the form, we want to show the image detail page with the errors messages. So to me that means that the default behaviour of `form_invalid` needs to be overriden.

amickan: I will make a commit to illustrate.

app/grandchallenge/algorithms/views.py
```diff
@@ -335,6 +338,67 @@ def get_context_data(self, *args, **kwargs):
...
+                    "mark_as_desired": True,
+                }
+            ).apply_async()
+            messages.success(
+                self.request,
+                "Image validation and upload to registry in progress. It can take a while for this image to become active, please be patient.",
+            )
+        else:
+            messages.error(self.request, form.non_field_errors())
+        return HttpResponseRedirect(self.get_success_url())
```
amickan, line 396: This is what I was referring to in my comment. This is still awkward, returning to the success url in form_invalid, but we want to always return to the image detail page and show success and error messages there.

jmsmkn: Why not show them on the page where the form is submitted?

amickan: That is the ImageDetail page.

amickan: We're using HTMX to send the post request (through a button on the ImageDetail page)

amickan: I could try to use the image detail template as the template for the form view, a bit awkward but it might work.

app/grandchallenge/algorithms/views.py
```diff
@@ -335,6 +338,67 @@ def get_context_data(self, *args, **kwargs):
...
+                    "app_label": self.algorithm_image._meta.app_label,
+                    "model_name": self.algorithm_image._meta.model_name,
+                    "pk": self.algorithm_image.pk,
+                    "mark_as_desired": True,
+                }
+            ).apply_async()
+            messages.success(
+                self.request,
+                "Image validation and upload to registry in progress. It can take a while for this image to become active, please be patient.",
+            )
```
jmsmkn, line 393: You should find a way to handle this in `form_valid`. The form should only be concerned about if the image can be marked as desired, then importing it into the registry should happen in `form_valid`. Definitely don't raise a validation error on the form just to ignore it and do something with the form data anyway.

app/grandchallenge/algorithms/forms.py
```diff
@@ -555,13 +556,34 @@ class AlgorithmImageUpdateForm(SaveFormInitMixin, ModelForm):
...
         model = AlgorithmImage
-        fields = ("requires_gpu", "requires_memory_gb")
+        fields = ("requires_gpu", "requires_memory_gb", "comment")
         labels = {"requires_gpu": "GPU Supported"}
         help_texts = {
             "requires_gpu": "If true, inference jobs for this container will be assigned a GPU"
         }
 
 
+class ImageMarkDesiredForm(Form):
```
jmsmkn, line 566: The form is missing the selection of the algorithm image. The algorithm image pk should be sent with the form data and not pulled out of the url and passed through the init method.

app/grandchallenge/algorithms/views.py
amickan: But then you're splitting the logic over two places. We would then need to check `can_execute` and `image_upload_in_progress` both in the form and in the `form_valid` method.

app/grandchallenge/algorithms/views.py
jmsmkn: If HTMX is an issue here then don't use it, put the button on the algorithm image detail page instead. If it is an issue then we need to look at HTMX in more detail as this is fundamental to how REST works.

app/grandchallenge/algorithms/views.py
jmsmkn: This is about responsibilities and separation of concerns. The Form is responsible for defining and validating what to change, the view is how to do it. The value of `can_execute` is cached (or should be if it is not), so it doesn't cost anything to call it twice.

app/grandchallenge/algorithms/views.py
```diff
@@ -335,6 +338,67 @@ def get_context_data(self, *args, **kwargs):
...
+    success_message = "Image successfully updated."
+    template_name = "algorithms/mark_desired_version.html"
+
+    @cached_property
+    def algorithm(self):
+        return self.algorithm_image.algorithm
+
+    @cached_property
+    def algorithm_image(self):
+        return get_object_or_404(AlgorithmImage, pk=self.kwargs["pk"])
```
jmsmkn, line 362: This property needs to go, you should get this from the validated data from the form.

app/grandchallenge/algorithms/forms.py
```diff
@@ -555,13 +556,34 @@ class AlgorithmImageUpdateForm(SaveFormInitMixin, ModelForm):
...
+        self.instance = image
+
+    def clean(self):
+        if not self.instance.is_manifest_valid:
+            raise ValidationError(
+                "Cannot mark an invalid image as active image."
+            )
+
+        if not self.instance.can_execute:
+            if self.instance.algorithm.image_upload_in_progress:
```
jmsmkn, line 578: `self.instance.algorithm.image_upload_in_progress` needs to be checked regardless of whether this particular image can be executed or not. Think of when a user uploads a new container, it's not validated yet, but they then try and activate an old one.

jmsmkn (approved):

PR #2837, opened 2023-04-18
Fix missing source maps

jmsmkn:
Django 4.0 a change to [ManifestStaticFilesStorage](https://docs.djangoproject.com/en/4.2/ref/contrib/staticfiles/#django.contrib.staticfiles.storage.ManifestStaticFilesStorage) means that Django tries to hash the JS source maps. Missing maps have been added but these are still often missing from distributions, so before we run `collectstatic` empty source maps are created that fixes the choking of `ManifestStaticFilesStorage`. Really a bit of a hack, but, I couldn't find a better way. The maps will still not be available, but this is OK.


package.json
```diff
@@ -5,18 +5,15 @@
     "@fortawesome/fontawesome-free": "^5.7.2",
     "@github/markdown-toolbar-element": "1.2.0",
     "jquery": "3.5.1",
-    "popper.js": "^1.14.3",
     "bootstrap": "^4.3.1",
-    "bootswatch": "^4.4.1",
-    "datatables.net-bs4": "^1.10.18",
-    "datatables.net-buttons-bs4": "^1.5.4"
```
jmsmkn, line 12: These dependencies are removed as we're already pulling them in from other sources.

amickan (approved):

PR #2842, opened 2023-05-03
Limit the number of algorithms a user can create for a phase

amickan:
This PR introduces a limit to the number of algorithms a user can make to a challenge phase. The limit is defined in an environment variable and defaults to 3. 

The `AlgorithmForPhase` view was adjusted so that a user now first gets to see a confirmation page with links to existing algorithms with matching inputs and outputs:
image

If a user has reached the limit, that page looks as follows: 
image

If a user has not created any algorithms yet, they will be directly redirected to the form to create an algorithm: 
image

Posting form data to this view will fail if a user has reached the limit of algorithms since the necessary input/output and all the other necessary initial form data won't have been set (we only do that when `hide_form` is `False`, which is only ever the case when a user has not reached the limit yet. 

Closes https://github.com/DIAGNijmegen/rse-roadmap/issues/214


app/grandchallenge/algorithms/forms.py
```diff
@@ -402,38 +402,46 @@ def __init__(
...
-        self.fields["contact_email"].disabled = True
-        self.fields["workstation"].initial = (
-            workstation
-            if workstation
-            else Workstation.objects.get(
-                slug=settings.DEFAULT_WORKSTATION_SLUG
+        if hide_form:
+            for field in self.fields:
+                self.fields[field].disabled = True
+                self.fields[field].widget = HiddenInput()
```
jmsmkn, line 413: I'm not sure that this works. Wouldn't it be more straightforward to do this client side with a `d-none` class? https://getbootstrap.com/docs/4.0/utilities/display/#hiding-elements

jmsmkn (commented):
LGTM, small question

app/grandchallenge/algorithms/forms.py
amickan: We are already not showing the form in the template if `hide_form` is `True`. I initially thought that disabling these fields would add an extra layer of security to make sure people can't just send a post request with form data and get around the limits, but I then realized that for that to work all the other hidden fields need to be set as well, so this can actually be removed.

jmsmkn: I think it would be more idiomatic to implement the limit check as a form validation method in `form.clean` instead of in the view and passing around this `hide_form` variable.

amickan: How would that work? We want to either show the form or not, the clean method is only called after a form is submitted, so then we'd always show the form and return validation errors, which is not really what we want, right?

amickan: Or do you mean keeping the check in the view, but also checking this again in the form's clean method (insetad of passing the `hide_form` variable to the form)?

jmsmkn: > Or do you mean keeping the check in the view, but also checking this again in the form's clean method (insetad of passing the hide_form variable to the form)?

Yes, you can put the check on the form, call it from the clean method and call it from the view if necessary (also cache it if needed). There may be a better place for the check, maybe on the phase model? Unsure. But yes, raising a validation error on clean is the way to ensure a form cannot be submitted.

amickan: On the phase model would have been nice, but I couldn't get that to work, so I moved the checks to the form like you suggested

jmsmkn (approved):

Issue #2841, opened 2023-05-03
UI inconsistencies for challenge (admin) page


miriam-groeneveld:
As administrator of a challenge there are a few inconsistencies in the UI that should be relatively easy to fix:
- [x] Rename `[Phase Name] Evaluation Settings` to `[Phase Name] Settings`
- [ ] Rename `Methods` to `Evaluation methods` and move to phase tab 
- [x] Add more helptext to submission date settings to explain that people can enter their local time.
- [x] Updating Phase settings redirects to the Leaderboard tab, should stay on the Admin tab
- [x] Not specific to the Admin role: don't show the `Join` button when the user has joined already.
- [x] Help text for `Hidden`: change "Do not display this Project in any public overview" to "Do not display this Challenge in any public overview" 
- [x] Help text for `Educational`: fix typo in "It is an educational challange"
- [x] Creator must be verified default on.
- [x] Your submissions tab should be labeled "All submissions" for admin users 
- [ ] Simplify the table on the methods list page to clearly show which image is the currently active image and hide unnecessary information 
- [ ] The submission list currently doesn't hold much useful information for admins. For debugging, they use the evaluation list page. To avoid confusion, it would be better to add the relevant submission information to the evaluation list so that there is one go-to place for challenge admins to debug submissions. For participants the submissions page as it is now needs to remain intact.

amickan:
Unsure if `creator_must_be_verified` should always be `True` by default (regardless of submission type), or if it should only be set to True for algorithm submission phases. 
Most of the phases that will be created in the future will be algorithm submission phases, so we could change this in the Phase model and set it to True for all future phases, and then in the form set it to True for algorithm submission phases.

miriam-groeneveld:
> Unsure if `creator_must_be_verified` should always be `True` by default (regardless of submission type), or if it should only be set to True for algorithm submission phases. Most of the phases that will be created in the future will be algorithm submission phases, so we could change this in the Phase model and set it to True for all future phases, and then in the form set it to True for algorithm submission phases.

That sounds good to me.

amickan:
I've decided to not make it default to `True` on the db level, and just added the change to the form for algorithm submission phases. I think this sufficiently addresses the user annoyance.

PR #2843, opened 2023-05-04
Challenge UI improvements

amickan:
Addresses most points mentioned in https://github.com/comic/grand-challenge.org/issues/2841


miriam-groeneveld (approved):
Looks good to me!

PR #2844, opened 2023-05-10
Challenge admin UI improvements part 2

### LINKED ISSUE #2841: UI inconsistencies for challenge (admin) page

amickan:
Here are some proposed challenge admin UI changes that Miriam and I discussed. Changes are mainly based on feedback we received from challenge organizers. 

This PR:
- splits the method, submission and evaluation lists up across challenge phases 
- adds links to the respective filtered list views nested under their respective phase tab in the admin sidebar
- cleans up the method list view to only include immediately relevant information and show more clearly what the active method for a given phase is
- adds crucial submission related information to the evaluation list table, making the submission list redundant for debugging by the admins

This is just a proposal. We can also only implement parts of the current changeset. 

Updated sidebar (the phase subtabs unfold when clicking on a phase):
image

Updated methods list (phase specific now): 
image

The "add new method" button redirects to a form where the phase is already preset and users only need to upload the container: 
image

Updated evaluations list page (phase specific), now includes submission information as well and is shown under the Admin tab, not the Leaderboard tab: 
image

Closes #2841


jmsmkn:
> Updated sidebar (the phase subtabs unfold when clicking on a phase)

Could we have some indication that it is clickable and either folded (right facing chevron) or unfolded (down facing chevron)

app/grandchallenge/evaluation/templates/evaluation/evaluation_list.html
```diff
@@ -144,7 +140,25 @@ <h2>Evaluations</h2>
     <script type="text/javascript">
         $(document).ready(function () {
```
jmsmkn, line 141: Whilst we're here could we move the js to its own file?

app/grandchallenge/evaluation/views/__init__.py
```diff
@@ -200,18 +209,37 @@ def get_permission_object(self):
...
-        kwargs.update({"challenge": self.request.challenge})
+        kwargs.update(
+            {"challenge": self.request.challenge, "phase": self.phase}
+        )
         return kwargs
 
+    def get_context_data(self, *, object_list=None, **kwargs):
+        context = super().get_context_data()
+        context.update({"phase": self.phase})
+        return context
```
jmsmkn, line 220: You can probably add this to your mixin.

app/grandchallenge/evaluation/forms.py
```diff
@@ -211,11 +211,15 @@ class MethodForm(ContainerImageForm):
...
 
-    def __init__(self, *args, challenge, **kwargs):
+    def __init__(self, *args, challenge, phase, **kwargs):
         super().__init__(*args, **kwargs)
 
         self.fields["phase"].queryset = challenge.phase_set.all()
 
+        if phase:
+            self.fields["phase"].initial = phase
+            self.fields["phase"].widget = HiddenInput()
```
jmsmkn, line 221: Is `challenge` still necessary? `phase` should always be set now so the if statement should not be necessary.

app/grandchallenge/evaluation/views/__init__.py
```diff
@@ -200,18 +209,37 @@ def get_permission_object(self):
 
     def get_form_kwargs(self):
```
jmsmkn, line 210: Could also go to the mixin, challenge is no longer necessary I think.

app/grandchallenge/evaluation/views/__init__.py
```diff
@@ -378,6 +412,18 @@ def get_queryset(self):
                 Q(submission__creator__pk=self.request.user.pk)
             )
 
+    def get_context_data(self, *, object_list=None, **kwargs):
```
jmsmkn, line 415: Using `get_context_data(self, **kwargs)` is more defensive.

app/grandchallenge/evaluation/forms.py
amickan: Depends on whether we keep the unfiltered view working or not. At the moment, `phase` can be unset if someone navigates to `/methods/create/` with `?phase=foo` added to the url. I kept that working intentionally for admins who might have bookmarked these urls (same goes for the SubmissionList and EvalList views, they still work without `phase` set). If we don't want to offer the unfiltered views anymore, than the if statement here (and in some other places) is not necessary indeed.

app/grandchallenge/evaluation/views/__init__.py
```diff
@@ -47,6 +47,14 @@
 from grandchallenge.verifications.views import VerificationRequiredMixin
 
 
+class CachedPhaseMixin:
+    @cached_property
+    def phase(self):
+        phase_slug = self.request.GET.get("phase", None)
```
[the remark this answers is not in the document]
amickan: `phase` is not a url kwarg at the moment but just a query parameter. I left it optional and didn't change the url path structure in order to keep the unfiltered view working. Should we deprecate the unfiltered views entirely?

app/grandchallenge/evaluation/forms.py
jmsmkn: Lets remove the unfiltered views as we're not linking to them.

jmsmkn: For the existing pages add a temporary redirect to the first non-hidden phase for the challenge.

app/grandchallenge/evaluation/urls.py
```diff
@@ -21,11 +22,25 @@
...
+    path("submissions/", SubmissionList.as_view(), name="submission-list"),
+    path(
+        "submissions/<uuid:pk>/",
+        SubmissionDetail.as_view(),
+        name="submission-detail",
+    ),
+    path("results/", LeaderboardRedirect.as_view()),
+    path("leaderboard/", LeaderboardRedirect.as_view()),
+    path("<slug>/", EvaluationList.as_view(), name="list"),
+    path("<slug>/admin", EvaluationAdminList.as_view(), name="admin-list"),
```
jmsmkn, line 43: Try to keep the URLs consistent with the rest, slug should come first if it is necessary, list views should be grouped by objects and URLs should end with "/":

```python
    path("<slug>/methods/", MethodList.as_view(), name="method-list"),
    path(
        "<slug>/methods/create/", MethodCreate.as_view(), name="method-create"
    ),
    path("<slug>/methods/<uuid:pk>/", MethodDetail.as_view(), name="method-detail"),
    path("<slug>/evaluations/", EvaluationList.as_view(), name="list"),
    path("<slug>/evaluations/admin/", EvaluationAdminList.as_view(), name="admin-list"),
    path("<slug>/submissions/", SubmissionList.as_view(), name="submission-list"),
    path(
        "<slug>/submissions/<uuid:pk>/",
        SubmissionDetail.as_view(),
        name="submission-detail",
    ),
```
   
The redirects should remain last:
  
```
    path("results/", LeaderboardRedirect.as_view()),
    path("leaderboard/", LeaderboardRedirect.as_view()),
```

amickan: Okay. I see that you also added the `slug` to the `submission-list`, for example. I'm not sure if we want that. We only use this to give participants an overview of all their submission (i.e. phase non-specific) through a link in the Submit menu. Do we really want this to be phase specific as well? We'd then need an extra nav-tab for each phase, which would clutter the UI I think. I would keep the submission list phase unspecific.

jmsmkn: OK thats fine, just thought that you had changed it as the location of those URLs changed.

jmsmkn (approved):

PR #2846, opened 2023-05-10
Validate segmentations for Tiff type images

### LINKED ISSUE #2672: TIFF files are invalid as (new) segmentation values
Validation always fails for TIFF files when uploading as a component-interface value with sub kind segmentation. (When `overlay_segments` are set).

Creating new segmentation component interfaces forces the addition of the `overlay_segments`. However, TIFF files have no value range calculated as their typical size (~4.000.000.000 voxels) makes it impractical to calculate v.s. regular MHA files (~200.000.000) as such the `image.segments` is never filled and the validation will always fail.

HarmvZ:
Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/232

Closes #2672


app/grandchallenge/components/models.py
```diff
@@ -662,32 +662,68 @@ class OverlaySegmentsMixin(models.Model):
...
+        allowed_values = self.voxel_values
         # An implicit background value of 0 is always allowed, this saves the
         # user having to declare it and the annotator mark it
         allowed_values.add(0)
 
         return allowed_values
 
+    @property
+    def is_contiguous(self):
+        values = sorted(list(self.voxel_values))
```
amickan, line 678: The implicit background value does not need to be included in this check?

app/grandchallenge/components/models.py
```diff
@@ -662,32 +662,68 @@ class OverlaySegmentsMixin(models.Model):
...
-                f"The valid voxel values for this segmentation are: "
-                f"{self.voxel_values}. This segmentation is invalid as "
-                f"it contains the voxel values: {invalid_values}."
-            )
+            if image.files.filter(
+                image_type=ImageFile.IMAGE_TYPE_TIFF
+            ).exists():
+                if not self.is_contiguous:
+                    raise ValidationError(
+                        "Voxel values in for the overlay segments must be contiguous "
```
amickan, line 693:
```suggestion
                        "Voxel values in the overlay segments must be contiguous "
```

app/grandchallenge/cases/models.py
```diff
@@ -287,6 +287,8 @@ class Image(UUIDModel):
     color_space = models.CharField(
         max_length=5, blank=False, choices=COLOR_SPACES
     )
+    min_voxel_value = models.IntegerField(null=True, blank=True)
+    max_voxel_value = models.IntegerField(null=True, blank=True)
```
amickan, line 291: These can be positive and negative - is that intended or should this be a `PositiveIntegerField`?

amickan:
This also requires adding the two new fields to the models in panimg, right?

app/grandchallenge/components/models.py
HarmvZ: I wasn't sure about that. It depends on what we allow. For now I left it out so that it has to be explicitly defined. I'm not really sure how this is supposed to work with signed integer datatypes either.

app/grandchallenge/cases/models.py
HarmvZ: It's intended. They can be negative when working with signed integer datatypes.

HarmvZ:
> This also requires adding the two new fields to the models in panimg, right?

Correct, that's part of https://github.com/DIAGNijmegen/rse-panimg/pull/95 (need to check the failing builds, will do it tomorrow).

jmsmkn:
The tests need to be fixed, maybe that happens with a new panimg release? Will look at those PRs today.

app/grandchallenge/components/models.py
jmsmkn: We could drop the contiguous checks now - the TIFF file will now fill out the segments with all the values between min an max and the existing checks will catch if there is a problem. Telling a user that the interface must be contiguous isn't going to be helpful. If we find a better way to determine the segments for the TIFF file in panimg then all this will still work too.

However, it would be helpful that there is a check for contiguous segments for new interface definitions so that we eliminate this corner case for now.

HarmvZ: I've removed the contiguous check in the voxel value validation and added a check to `_clean_overlay_segments` so new and existing component interfaces cannot be saved when the overlay segments are not contiguous.

app/grandchallenge/components/models.py
```diff
@@ -662,30 +662,42 @@ class OverlaySegmentsMixin(models.Model):
 
     @property
     def voxel_values(self):
-        allowed_values = {x["voxel_value"] for x in self.overlay_segments}
+        return {x["voxel_value"] for x in self.overlay_segments}
 
+    @property
+    def allowed_values(self):
```
jmsmkn, line 668: Do we need to split `voxel_values` and `allowed_values`? It's confusing having two now.

jmsmkn (approved):

app/grandchallenge/components/models.py
```diff
@@ -902,6 +912,11 @@ def _clean_overlay_segments(self):
                 "Overlay segments should only be set for segmentations"
             )
 
+        if not self.overlay_segments_is_contiguous:
```
miriam-groeneveld, line 915: Only saw this now. This now requires all overlay_segments to be contiguous, but we have non-contiguous overlay_segments already in the database.

jmsmkn: Yes that's fine, they're not new and if they need to be updated will need to add intermediate values.

PR #2847, opened 2023-05-11
Add base cost to challenge cost estimate

amickan:
This adds the base cost to the challenge cost estimate calculation: 
image

And also fixes the label for the "agree to pricing policy" checkbox where the link wouldn't show correctly.


miriam-groeneveld (approved):
Looks good!

PR #2854, opened 2023-05-22
Add active version to submission form algorithm options

amickan:
This adds the active version to the algorithm dropdown on the submission page. The easiest way to achieve this was to update the string method of the Algorithm object. That required some small changes elsewhere in the codebase, but seemed better than creating a custom select widget for the submission form. 

I also thought that it would make sense to restrict the algorithm shown in the dropdown to those that can be submitted to the phase (i.e. with matching inputs).

image

### DIFF
--- a/app/grandchallenge/algorithms/models.py
+++ b/app/grandchallenge/algorithms/models.py
@@ -241,7 +241,7 @@ class Algorithm(UUIDModel, TitleSlugDescriptionModel, ViewContentMixin):
         ]
 
     def __str__(self):
-        return f"{self.title} (Active version: {self.active_image})"
+        return f"{self.title}"
 
     def get_absolute_url(self):
         return reverse("algorithms:detail", kwargs={"slug": self.slug})
--- a/app/grandchallenge/algorithms/models.py
+++ b/app/grandchallenge/algorithms/models.py
@@ -526,6 +526,9 @@ class AlgorithmImage(UUIDModel, ComponentImage):
     class Meta(UUIDModel.Meta, ComponentImage.Meta):
         ordering = ("created", "creator")
 
+    def __str__(self):
+        return f"Algorithm image {self.pk} for {self.algorithm.title}"
+
     def get_absolute_url(self):
         return reverse(
             "algorithms:image-detail",
--- a/app/grandchallenge/algorithms/models.py
+++ b/app/grandchallenge/algorithms/models.py
@@ -13,6 +13,7 @@ from django.db.models import Count, Min, Q, Sum
 from django.db.models.signals import post_delete
 from django.db.transaction import on_commit
 from django.dispatch import receiver
+from django.template.defaultfilters import truncatewords
 from django.utils import timezone
 from django.utils.functional import cached_property
 from django_extensions.db.models import TitleSlugDescriptionModel
@@ -527,7 +528,7 @@ class AlgorithmImage(UUIDModel, ComponentImage):
         ordering = ("created", "creator")
 
     def __str__(self):
-        return f"Algorithm image {self.pk} for {self.algorithm.title}"
+        return f"Algorithm image {self.pk} for {self.algorithm} ({truncatewords(self.comment, 4)})"
 
     def get_absolute_url(self):
         return reverse(
--- a/app/grandchallenge/algorithms/models.py
+++ b/app/grandchallenge/algorithms/models.py
@@ -241,7 +241,7 @@ class Algorithm(UUIDModel, TitleSlugDescriptionModel, ViewContentMixin):
         ]
 
     def __str__(self):
-        return f"{self.title}"
+        return f"{self.title} (Active version: {self.active_image})"
 
     def get_absolute_url(self):
         return reverse("algorithms:detail", kwargs={"slug": self.slug})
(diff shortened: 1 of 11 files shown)


app/grandchallenge/evaluation/forms.py
```diff
@@ -253,7 +256,7 @@ class SubmissionForm(SaveFormInitMixin, forms.ModelForm):
 
     def __init__(self, *args, user, phase: Phase, **kwargs):  # noqa: C901
         super().__init__(*args, **kwargs)
-
+        self._user = user
```
jmsmkn, line 259: You could also do this in the Mixin, would be a little clearer for later use:

```python
class UserAlgorithmsForPhaseMixin:
  def __init__(self, *args, user, **kwargs):
    self._user = user

...

class SubmissionForm(UserAlgorithmsForPhaseMixin, ...):
  def __init__(self, *args, phase, **kwargs):
    # use self._user here

...
```

app/grandchallenge/algorithms/models.py
```diff
@@ -241,7 +241,7 @@ class Meta(UUIDModel.Meta, TitleSlugDescriptionModel.Meta):
         ]
 
     def __str__(self):
-        return f"{self.title}"
+        return f"{self.title} (Active version: {self.active_image})"
```
jmsmkn, line 244: Did you check for N+1 problems on all the dropdowns where this is used?

jmsmkn (commented):
LGTM

app/grandchallenge/algorithms/models.py
amickan: Ah good point. There is an N+1 problem with this indeed, but somehow prefetching `algorithm_container_images` in the respective view(s) doesn't help. Have to look into it more.

amickan: This is strange. Maybe it has to do with the fact that we filter the `algorithm_container_images` to get the active image?

jmsmkn: It is because `active_image` does a `.get()` for each row. This is fine as we're usually only doing this for one algorithm. I think it would be best if this was removed from the `__str__` method and we tried another approach as I think that is really difficult to solve with all our rules.

We do this weird thing where we get the algorithm in the form and then set the algorithm image from that (https://github.com/comic/grand-challenge.org/blob/255996b8f373d391cac93529e40bbbe89901b6ab/app/grandchallenge/evaluation/views/__init__.py#L296-L300), I wonder if we should change things around so that we get the algorithm image directly instead?

amickan: Ah that's a good idea. We don't have a manager method for that yet, but that can be easily added.

jmsmkn (commented):
Looks good, few usability questions.

app/grandchallenge/algorithms/models.py
```diff
@@ -526,6 +526,9 @@ class AlgorithmImage(UUIDModel, ComponentImage):
     class Meta(UUIDModel.Meta, ComponentImage.Meta):
         ordering = ("created", "creator")
 
+    def __str__(self):
+        return f"Algorithm image {self.pk} for {self.algorithm.title}"
```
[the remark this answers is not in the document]
amickan: Sure, the algorithm string method works! And no N+1 problems as far as I can tell. We're only listing algorithm images in two places and in both those places, I've added a `select_related` to the query.

As for adding the comment - yes, my only concern is that the comment can be quite long...

app/grandchallenge/algorithms/templates/algorithms/job_detail.html
```diff
@@ -130,7 +130,7 @@ <h3>Result Reference Data</h3>
 
                 <dt class="col-sm-3">Algorithm</dt>
                 <dd class="col-sm-9"><a
-                        href="{{ object.algorithm_image.algorithm.get_absolute_url }}">{{ object.algorithm_image.algorithm }}</a>
+                        href="{{ object.algorithm_image.algorithm.get_absolute_url }}">{{ object.algorithm_image.algorithm.title }}</a>
```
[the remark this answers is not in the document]
amickan: Ah yes, that was a remnant from the previous changes I had made. Will revert!

app/grandchallenge/algorithms/models.py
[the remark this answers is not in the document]
jmsmkn: You could truncate the number of words https://docs.djangoproject.com/en/4.2/ref/templates/builtins/#truncatewords - it's a template filter but no reason why you couldn't use that directly in the `str` method.

app/grandchallenge/evaluation/forms.py
```diff
@@ -314,48 +312,25 @@ def __init__(self, *args, user, phase: Phase, **kwargs):  # noqa: C901
...
-                "required by this phase, please update your algorithm "
-                "to produce: "
-                f"{oxford_comma(self._algorithm_outputs)}. "
-            )
-
-        if algorithm.active_image is None:
-            raise ValidationError(
-                "This algorithm does not have a usable container image. "
-                "Please add one and try again."
-            )
```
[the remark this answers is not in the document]
amickan: Yes, there is a test for the queryset filtering.

jmsmkn (approved):

PR #2856, opened 2023-05-26
Set permissive cross-origin opener header to fix session control

HarmvZ:
Fixes broken session control. 

It's broken because this header is added by default in Django 4.0: https://docs.djangoproject.com/en/4.1/releases/4.0/#requests-and-responses
This disables controlling of the window opened by window.open() for different origins. Since the workstation is loaded on a different subdomain this is seen as a different origin.

See long investigation and discussion in: 
https://teams.microsoft.com/l/message/19:54cf52330bfb4c33991e550b49355741@thread.tacv2/1685087955060?tenantId=b208fe69-471e-48c4-8d87-025e9b9a157f&groupId=45c973c4-5543-45ad-a8b8-79ea596e6c69&parentMessageId=1685087955060&teamName=Research%20Software%20Engineering&channelName=General&createdTime=1685087955060&allowXTenantAccess=false

### DIFF
--- a/app/config/settings.py
+++ b/app/config/settings.py
@@ -289,6 +289,10 @@ CSRF_TRUSTED_ORIGINS = [f"https://*{SESSION_COOKIE_DOMAIN}"]
 SECURE_PROXY_SSL_HEADER = ("HTTP_X_FORWARDED_PROTO", "https")
 SECURE_SSL_REDIRECT = True
 
+# Set permissive cross-origin header to allow session control of windows opened that
+# go to a different subdomain (workstations)
+SECURE_CROSS_ORIGIN_OPENER_POLICY = None
+
 # Set the allowed hosts to the cookie domain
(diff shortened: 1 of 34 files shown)


app/config/settings.py
```diff
@@ -289,6 +289,10 @@
 SECURE_PROXY_SSL_HEADER = ("HTTP_X_FORWARDED_PROTO", "https")
 SECURE_SSL_REDIRECT = True
 
+# Set permissive cross-origin header to allow session control of windows opened that
+# go to a different subdomain (workstations)
+SECURE_CROSS_ORIGIN_OPENER_POLICY = None
+
```
jmsmkn, line 295: We probably want to control this per domain, so better take it from the environment variables.

```suggestion
SECURE_CROSS_ORIGIN_OPENER_POLICY = os.environ.get("SECURE_CROSS_ORIGIN_OPENER_POLICY", "same-origin")
```

PR #2830, opened 2023-04-07
Use path parameters in workstation session creation URLs

HarmvZ:
Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/225.

This PR changes the workstation session creation URLs to use path parameters instead of query parameters. 

**This is ready for review, but it cannot be released until the changes at the cirrus side are finalized and released.**

PR #2862, opened 2023-06-01
Add workstation path to session start redirects

HarmvZ:
Propagates path params in redirects when starting a session. Fixes broken session start from #2830.


app/grandchallenge/workstations/views.py
```diff
@@ -301,10 +302,10 @@ def form_valid(self, form):
...
 
         url = session.get_absolute_url()
-
+        url += f"?path={quote_plus(self.kwargs.get('workstation_path', ''))}"
         qs = self.request.META.get("QUERY_STRING", "")
         if qs:
-            url = f"{url}?{qs}"
+            url = f"{url}&qs={quote_plus(qs)}"
 
         return HttpResponseRedirect(url)
```
HarmvZ, line 310: The redirect here needs to load SessionDetail so it cannot have the cirrus path added to it because that would redirect to the session_proxy instead of SessionDetail. Therefore, the cirrus path is sent through as a query parameter. This ensures the correct view is loaded. In the SessionDetail template, the query parameters are parsed and converted back into the correct URL to load the workstation.

jmsmkn (approved):

PR #2868, opened 2023-06-06
Adds challenge url to costs overview

miriam-groeneveld:
To add a bit of convenience


miriam-groeneveld:
I think the failed test is a glitch?

jmsmkn (approved):

jmsmkn:
Yes it's the old #2759

PR #2870, opened 2023-06-08
Consolidate management of vendored JS

jmsmkn:
We had several methods of managing vendored JS sources, this PR removes using two CDNs and NPM and replacing it with amass. This will help us to enable a good content security policy later by only depending on sources we control.

TODO in further PRs:

- [ ] Fix the usage of `vendor/css/base.min.css` in the admin
- [ ] Replace usages of jsdelivr
- [ ] Add security.txt


pkcakeout:
Management is trying to boost their their line-stats for commits? :laughing: 

image

Out of curiosity, what are you doing here? Is this a big old house-cleaning that is happening? Looks like it?!

PR #2873, opened 2023-06-09
Introduce TEXT-type answers

chrisvanrun:
Part of the pitch:

This change-set adds TEXT-type answers including:
* widgets `Text Input`, `Text Area`
* validation params `answer_max_length`, `answer_min_length`, `answer_match_pattern`

Changeset also includes:
* a minor bug in the `NUMBER` + `NUMBER_INPUT` widget that showed when adding an `answer_type` factor to the parameterized tests
* fix not showing the help-text of both the `interface` and the `widget` fields in the form (the `widgets` + `help_text` + `DynamicField` do not seem to work as one would expect.)

It purposely does **not** contain the data migration to collapse the to-be deprecated 'Multi line' and 'Single line' answer types. This requires a stepped release in that the previous release of CIRRUS should support TEXT types.

# Screenshot
The form question UI changes look like this:


chrisvanrun:
I missed a migration, on it.

app/tests/core_tests/test_fields.py
```diff
@@ -0,0 +1,24 @@
...
+            r"[abc]",
+            nullcontext(),
+        ),
+        (
+            r"[abc",  # Missing closing bracket
+            pytest.raises(ValidationError),
+        ),
+    ),
+)
+def test_regex_field_validation(value, error):
```
jmsmkn, line 22: Would be good to check that complex patterns get compiled correctly and the correct regex is returned, e.g. `r"/^(https?:\/\/)?([\da-z\.-]+)\.([a-z\.]{2,6})([\/\w \.-]*)*\/?$/"`.

Also better to go through Django's way of validating fields rather than calling the private validation method here and trusting that it's all hooked up correctly. You can create a temporary models to do this.

app/grandchallenge/reader_studies/models.py
```diff
@@ -1528,6 +1591,18 @@ def validate(  # noqa: C901
                 offset=question.answer_min_value,
             )(answer)
 
+        if question.answer_min_length is not None:
+            MinLengthValidator(question.answer_min_length)(answer)
+
+        if question.answer_max_length is not None:
+            MaxLengthValidator(question.answer_max_length)(answer)
+
+        if question.answer_match_pattern is not None:
```
jmsmkn, line 1600: Should this be `is not None`? It's a non-nullable field. I think `if question.answer_match_pattern` would do.

jmsmkn (commented):
LGTM, general comments:

- `_clean_widget_options` is getting long and complex, may be better to split it up into the relevant groups?
- Can `widget_choices` be made a bit simpler?

app/grandchallenge/reader_studies/models.py
chrisvanrun: Good point! I had thought to replace this in another check but forgot to also do this here.

app/tests/core_tests/test_fields.py
chrisvanrun: Using the a temp model for the field test turned out to be relatively simple:

```Python
class ModelWithRegex(models.Model):
    regex = RegexField()


@pytest.mark.parametrize(
    "value, error",
    (
        (
            r"[abc]",
            nullcontext(),
        ),
        (
            r"[abc",  # Missing closing bracket
            pytest.raises(ValidationError),
        ),
        (
            r"/^(https?:\/\/)?([\da-z\.-]+)\.([a-z\.]{2,6})([\/\w \.-]*)*\/?$/",
            nullcontext(),
        ),
    ),
)
def test_regex_field_validation(value, error):
    model = ModelWithRegex(regex=value)
    with error:
        model.full_clean()
```

However, I am not too sure how to test if the pattern survives DB entry intact without actually using the DB. Do you have a suggestion?

jmsmkn: Yes, the only way to test that would be to save to the db and read back. You are already using the field in a model, so could do that there.

jmsmkn (approved):

PR #2877, opened 2023-06-09
Fix green bg-dark

### LINKED ISSUE #2876: Documentation links are green
c384765a-c50d-4f2c-8670-535a0225cbf0

jmsmkn:
Problem was a mis-match between bootswatch and bootstrap versions. Vendor files updated, no need to review those. Other changes are QOL improvements for development.

Closes #2876

### DIFF
--- a/dockerfiles/web/Dockerfile
+++ b/dockerfiles/web/Dockerfile
@@ -23,7 +23,7 @@ FROM ${GRAND_CHALLENGE_WEB_BASE_REPOSITORY_URI}:${PYTHON_VERSION}-${POETRY_HASH}
 WORKDIR /app
 COPY --chown=django:django ./app/ /app/
 
-RUN python manage.py collectstatic --noinput && python manage.py compress
+RUN python manage.py collectstatic --noinput && python manage.py compress --force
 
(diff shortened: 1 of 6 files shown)


chrisvanrun (approved):
LGTM!

dockerfiles/web/Dockerfile
```diff
@@ -23,7 +23,7 @@ FROM ${GRAND_CHALLENGE_WEB_BASE_REPOSITORY_URI}:${PYTHON_VERSION}-${POETRY_HASH}
 WORKDIR /app
 COPY --chown=django:django ./app/ /app/
 
-RUN python manage.py collectstatic --noinput && python manage.py compress
+RUN python manage.py collectstatic --noinput && python manage.py compress --force
```
chrisvanrun, line 26: Only a small comment here as to why the `--force` was added. Always has a bit of an infra-as-code smell.

jmsmkn: We also force the other command with `--no-input`. `--force` here skips the `COMPRESS_ENABLED` check, otherwise you need to set a default to True, then override it in 3+ other places for development!

PR #2882, opened 2023-06-16
Add Select Multiple widgets

chrisvanrun:
Part of the pitch:

This (tiny) changeset adds the `Select Multiple` and `Checkbox Select Multiple` widgets as options to the `Multiple choice` answer type. The somewhat cryptic 'SLCT_M'/'CHCKB_SLCT_M' is to have it fit the 13 Char limit.

It does not do any migration as of yet because of stepped releases and backward compatibility.


jmsmkn:
Just increase the char limit rather than making them cryptic.

chrisvanrun:
> Just increase the char limit rather than making them cryptic.

I once wanted to do that for the image ports but vaguely remember I was told it required non-trivial migrations. No ill-effects if the char limit is increased?

jmsmkn:
Are you sure? Renaming the keys is non-trivial as it would require a data migration too. Reducing the limit is only non-trivial if you need to renaming the keys in order to do it. Increasing the char limit is trivial.

chrisvanrun:
> Are you sure? Renaming the keys is non-trivial as it would require a data migration too. Reducing the limit is only non-trivial if you need to renaming the keys in order to do it. Increasing the char limit is trivial.

It's likely a combination of both at the time. I'll store it correctly in the memory banks this time around. I assume that increasing the char limit only needs a within-database migration.

jmsmkn (approved):

PR #2904, opened 2023-06-28
Adds example to docstring of extra results columns

miriam-groeneveld:


jmsmkn:
You need to delete the migration you added, rebase the branch from main, and re-generate the migration (or find the last modification of this field and directly update the help text there as changing the help text is a noop).

PR #2906, opened 2023-06-29
Add slider widget for number questions

chrisvanrun:
Part of the pitch:

This tiny PR adds the switch for selecting the 'Slider' widget introduced in the CIRRUS PR:
- https://github.com/DIAGNijmegen/rse-cirrus-core/pull/2553


app/grandchallenge/reader_studies/models.py
```diff
@@ -1024,6 +1024,7 @@ def get_widget_required_types():
 class QuestionWidgetKindChoices(models.TextChoices):
     ACCEPT_REJECT = "ACCEPT_REJECT", "Accept/Reject Findings"
     NUMBER_INPUT = "NUMBER_INPUT", "Number input"
+    SLIDER = "Slider", "Slider"
```
jmsmkn, line 1027: Also needs changing in the migration. Shouldn't we call it `Range` to match the html specs? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range

```suggestion
    SLIDER = "SLIDER", "Slider"
```

chrisvanrun: I think NUMBER_RANGE is the best fit with the others. I'll update things.

jmsmkn (approved):

jmsmkn:
Don't you want to enable `perform_number_validation` with the new widget? Are there any other things you have missed for the new widget types?

chrisvanrun:
> Don't you want to enable `perform_number_validation` with the new widget? Are there any other combinations like this that have missed for the new widget types?

Good point! They were automatically inherited in CIRRUS so I hadn't had them on my sights directly here. On it!

chrisvanrun:
The validation for both widget-config-related members and the answer's value for Number Range has been added.

chrisvanrun:
Tests fail on unrelated code... I think. Investigating.

jmsmkn:
Could you consolidate the migrations? We should only introduce on new migration per app in a PR otherwise they build up too quickly.

chrisvanrun:
Ran into the flaky test again... what are the odds...

jmsmkn (approved):

Issue #2908, opened 2023-06-30
`test_notification_view_permissions` flaky


chrisvanrun:
Seems random, last fail:
- https://github.com/comic/grand-challenge.org/actions/runs/5421079626/jobs/9856055620

chrisvanrun:
I find it hard to read the actual error, but I suspect the cause is calling a render twice (once in the response, once from the object) and the rendering having content generated as:
`naturaltime(self.created)`

There is likely a (small) chance that clock time has shifted between rendering. I suspect it only needs a second. I suspect a fix is easily done with patching datetime.now .

PR #2909, opened 2023-06-30
Fix flaky test

chrisvanrun:
Closes: #2908 

Culrpit was this line:
```Python
  assert (
        notification.print_notification(user=user1)
        in response.rendered_content
    )
```

Forcing the flakiness was easily done by inserting some time between the render calls:
```Python
    response_content = response.rendered_content
    time.sleep(5)
    notification = notification.print_notification(user=user1)
    assert notification in response_content
```

### DIFF
--- a/app/tests/notifications_tests/test_views.py
+++ b/app/tests/notifications_tests/test_views.py
@@ -1,6 +1,3 @@
-from datetime import datetime, timezone
-from unittest.mock import patch
-
 import pytest
 from actstream.actions import follow, is_following
 from actstream.models import Follow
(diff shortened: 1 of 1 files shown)


app/tests/notifications_tests/test_views.py
```diff
@@ -87,7 +90,15 @@ def test_notification_deletion(client):
 
 
 @pytest.mark.django_db
-def test_notification_view_permissions(client):
+@patch(
+    "django.contrib.humanize.templatetags.humanize.datetime",
+    wraps=datetime,
+)
+def test_notification_view_permissions(mock_datetime, client):
+    mock_datetime.now.return_value = datetime.now(
```
jmsmkn, line 98: This also seems like a fragile fix, I think it's more robust to check for the notification objects in `set(response.context["notifications"]) == {notification}`, and an empty set in the other test. Also is it a bug in the test in L132 as the wrong user is being sent to `print_notification`?

chrisvanrun: Fixed now.

jmsmkn (approved):

PR #2910, opened 2023-06-30
Fix focused button box-shadow being clipped

chrisvanrun:
This has been a thorn in my eye for a very long time.

pre_button

Any button rendered near the bottom of a content (e.g. towards the footer) would have its border occluded when pressed.

By adding some margin that matches the border radius of `.btn` this is fixed:

post

For existing elements with `.btn` and `.m-1` (or variant) the margin is removed.

### DIFF
--- a/app/grandchallenge/core/static/css/core.css
+++ b/app/grandchallenge/core/static/css/core.css
@@ -242,7 +242,7 @@ ul.socialaccount_providers > li {
 }
 
 .btn {
-    margin: 0.25rem;
+    margin: 0.2rem;
 }
 
 .color-box {
--- a/app/grandchallenge/core/static/css/core.css
+++ b/app/grandchallenge/core/static/css/core.css
@@ -241,10 +241,6 @@ ul.socialaccount_providers > li {
     box-shadow: 0 0 1rem 0 rgba(0, 0, 0, .1);
 }
 
-.btn {
-    margin: 0.2rem;
-}
-
 .color-box {
     display: inline-block;
     border: black solid 1px;
--- a/app/grandchallenge/core/static/css/core.css
+++ b/app/grandchallenge/core/static/css/core.css
@@ -241,6 +241,10 @@ ul.socialaccount_providers > li {
     box-shadow: 0 0 1rem 0 rgba(0, 0, 0, .1);
 }
 
+.btn {
+    margin: 0.25rem;
+}
+
 .color-box {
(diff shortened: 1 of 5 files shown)


app/grandchallenge/core/static/css/core.css
```diff
@@ -241,6 +241,10 @@ ul.socialaccount_providers > li {
     box-shadow: 0 0 1rem 0 rgba(0, 0, 0, .1);
 }
 
+.btn {
+    margin: 0.25rem;
```
jmsmkn, line 245: I don't think adding all margins to all buttons is the way to go to fix this issue, right? Isn't this an issue instead with the padding of the content/footer?

chrisvanrun: TBH I am not too sure either. Though the manual margin fixes seem to suggest that is needed every now and then. Most buttons aren't being pushed to the side of containers of another visible element that could occlude it.

The pressed-button border isn't being taken into account when displaying the buttons, that much is clear. Adding some margin to the footer isn't really the correct fix IMHO. 

Maybe there is some other CSS magic that forces the browser to take the border into account? Some `display` choice?

chrisvanrun: Alternatively, we can fix it only for the buttons where I've noticed it (login/logout, some detail pages).

jmsmkn: I think the problem is somewhere in `<main>`, maybe with `<div class="container pb-3 mt-3">`, or the footer.

chrisvanrun: In the context of detail pages, that makes sense to blame the surroundings. However, in different contexts, it still shows:
image

Diving a bit deeper it seems like it is the box-shadow that is being used to create the visual outline at focus:
```CSS
.btn:focus, .btn.focus {
  outline: 0;
  box-shadow: 0 0 0 .2rem rgba(44,62,80,0.25);
}
```
The only way I can see to take this into account is to reserve space for it. Albeit it should be .2 instead of .25.

jmsmkn: Can we get the .2rem value from the Bootstrap SCSS and make the correction in those files rather than hardcoding instead?

chrisvanrun: Done.

jmsmkn (approved):

PR #2922, opened 2023-07-07
Fix not being able to post empty answers for non-required questions.

### LINKED ISSUE #2916: Non required questions with null answers are being validated
Questions that are not required and to which an empty answer is being posted are being validated incorrectly.

See this sentry issue, as a result of a number type question with a minimum limit set: 
- https://grand-challenge.sentry.io/issues/4295894591/?project=303639&query=is%3Aunresolved&referrer=issue-stream&stream_index=0

chrisvanrun:
Closes #2916

This changeset prevents running validators on non-required questions with an explicit empty answer. Special allowance had to be made for blank-type answer types and image types.

This required some work to understand the broken tests and their initial meaning. I've added a few cases to those tests to further the scope and some corner cases.

### DIFF
--- a/app/grandchallenge/reader_studies/models.py
+++ b/app/grandchallenge/reader_studies/models.py
@@ -1090,6 +1090,32 @@ ANSWER_TYPE_TO_QUESTION_WIDGET_CHOICES = {
     for answer_type, options in ANSWER_TYPE_TO_QUESTION_WIDGET.items()
 }
 
+EMPTY_ANSWER_VALUES = {
+    AnswerType.SINGLE_LINE_TEXT: "",
+    AnswerType.MULTI_LINE_TEXT: "",
+    AnswerType.TEXT: "",
+    AnswerType.NUMBER: None,
+    AnswerType.CHOICE: None,
+    AnswerType.BOOL: None,
+    AnswerType.BOUNDING_BOX_2D: None,
+    AnswerType.MULTIPLE_2D_BOUNDING_BOXES: None,
+    AnswerType.DISTANCE_MEASUREMENT: None,
+    AnswerType.MULTIPLE_DISTANCE_MEASUREMENTS: None,
+    AnswerType.POINT: None,
+    AnswerType.MULTIPLE_POINTS: None,
+    AnswerType.POLYGON: None,
+    AnswerType.MULTIPLE_POLYGONS: None,
+    AnswerType.LINE: None,
+    AnswerType.MULTIPLE_LINES: None,
+    AnswerType.MASK: None,
+    AnswerType.ANGLE: None,
+    AnswerType.MULTIPLE_ANGLES: None,
+    AnswerType.ELLIPSE: None,
+    AnswerType.MULTIPLE_ELLIPSES: None,
+    AnswerType.MULTIPLE_CHOICE: [],
+    AnswerType.MULTIPLE_CHOICE_DROPDOWN: [],
+}
+
 
 class Question(UUIDModel, OverlaySegmentsMixin):
     AnswerType = AnswerType
@@ -1440,45 +1466,19 @@ class Question(UUIDModel, OverlaySegmentsMixin):
             )
 
     @property
-    def allow_null_types(self):
-        return [
-            *self.AnswerType.get_annotation_types(),
-            self.AnswerType.CHOICE,
-            self.AnswerType.NUMBER,
-        ]
-
-    @property
-    def allow_blank_types(self):
-        return [
-            self.AnswerType.TEXT,
-            self.AnswerType.SINGLE_LINE_TEXT,
-            self.AnswerType.MULTI_LINE_TEXT,
-        ]
-
-    @property
-    def allow_empty_list_types(self):
-        return [
-            self.AnswerType.MULTIPLE_CHOICE,
-            self.AnswerType.MULTIPLE_CHOICE_DROPDOWN,
-        ]
-
-    @property
-    def empty_answer_values(self):
-        """Returns a list of answer values which are considered to be empty"""
-        result = []
-        if self.answer_type in self.allow_null_types:
-            result.append(None)
-        if self.answer_type in self.allow_blank_types:
-            result.append("")
-        if self.answer_type in self.allow_empty_list_types:
-            result.append([])
-        return result
+    def empty_answer_value(self):
+        """Returns the answer value which is considered to be empty"""
+        if self.answer_type not in EMPTY_ANSWER_VALUES:
+            raise RuntimeError(
+                f"{self.answer_type} has no representation of an empty value."
+            )
+        return EMPTY_ANSWER_VALUES[self.answer_type]
 
     def is_answer_valid(self, *, answer):
         """Validates ``answer`` against ``ANSWER_TYPE_SCHEMA``."""
         allowed_types = [{"$ref": f"#/definitions/{self.answer_type}"}]
 
-        if self.answer_type in self.allow_null_types:
+        if self.empty_answer_value is None:
             allowed_types.append({"$ref": "#/definitions/null"})
 
         try:
@@ -1671,9 +1671,11 @@ class Answer(UUIDModel):
                     "Provided options are not valid for this question"
                 )
 
-        answer_is_empty = answer in question.empty_answer_values
-        answer_is_required = question.required and not question.is_image_type
-        if answer_is_required and answer_is_empty:
+        answer_is_empty = (
+            answer == question.empty_answer_value
+            and not question.is_image_type
+        )
+        if question.required and answer_is_empty:
             raise ValidationError(
                 "Answer for required question cannot be empty"
             )
--- a/app/grandchallenge/reader_studies/models.py
+++ b/app/grandchallenge/reader_studies/models.py
@@ -1671,6 +1671,8 @@ class Answer(UUIDModel):
                     "Provided options are not valid for this question"
                 )
 
+        # Image types can have empty answers while an image is being uploaded,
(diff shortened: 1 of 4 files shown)


app/tests/reader_studies_tests/test_api.py
```diff
@@ -797,9 +802,11 @@ def test_answer_creator_is_reader(client):
         (Question.AnswerType.MULTIPLE_POLYGONS, None, 201),
         (Question.AnswerType.ANGLE, None, 201),
         (Question.AnswerType.MULTIPLE_ANGLES, None, 201),
-        (Question.AnswerType.CHOICE, None, 400),
+        (Question.AnswerType.CHOICE, None, 201),
```
jmsmkn, line 805: Hmm, now we're saying that non-answers for Choice and Number are now `None` rather than `""`. Has something changed in CIRRUS? Do all widgets now return these answers rather than `""`? It must have worked before surely? We would need to check the database and existing values are consistent.

chrisvanrun: I assumed non-answers representation based on the snippets below and the errors that the validators are reporting about None types. I'll double-check what CIRRUS outputs atm.


https://github.com/comic/grand-challenge.org/blob/8e1e08b5d6d2a94752c2b243f98105a128449a3d/app/grandchallenge/reader_studies/models.py#L1436-L1449

chrisvanrun: Widgets haven't changed AFAIK. The current output of CIRRUS with no answers:

```
AnswerType.SINGLE_LINE_TEXT ""
AnswerType.MULTI_LINE_TEXT ""
AnswerType.TEXT ""
AnswerType.BOOL False
AnswerType.NUMBER None
AnswerType.CHOICE None
AnswerType.MULTIPLE_CHOICE []
AnswerType.POINT None
AnswerType.MULTIPLE_POINTS None
AnswerType.DISTANCE_MEASUREMENT None
AnswerType.MULTIPLE_DISTANCE_MEASUREMENTS None
AnswerType.BOUNDING_BOX_2D None
AnswerType.MULTIPLE_2D_BOUNDING_BOXES None
AnswerType.POLYGON None
AnswerType.MULTIPLE_POLYGONS None
AnswerType.LINE None
AnswerType.MULTIPLE_LINES None
AnswerType.MASK None
AnswerType.ANGLE None
AnswerType.MULTIPLE_ANGLES None
AnswerType.ELLIPSE None
AnswerType.MULTIPLE_ELLIPSES None
```

chrisvanrun: `CHOICE` and `NUMBER` all annotations have always (or at least for a while) had non-answer as `None`. I wasn't too sure about the `NUMBER` but I checked what CIRRUS (pre-widget split) does and v6.0.0 of CIRRUS already produces `None` for `NUMBER`.

We do sort of break with the Widget (HTML) wise in that the number inputs always generate a `""` when left empty. But then again, that is always a string... so even a basic interpretation should put it to `None`/`Null` before processing.

So sorry, I cannot follow where you see the hiccup.

jmsmkn: OK, strange, then I have no idea what is/was going on with these tests. We indeed only have values for `None` for those types:

```python
>>> Answer.objects.filter(answer__isnull=True, question__answer_type=AnswerType.CHOICE).count()
124
>>> Answer.objects.filter(answer="", question__answer_type=AnswerType.CHOICE).count()
0
>>> Answer.objects.filter(answer__isnull=True, question__answer_type=AnswerType.NUMBER).count()
4536
>>> Answer.objects.filter(answer="", question__answer_type=AnswerType.NUMBER).count()
0
```

The tests here would suggest that they wouldn't be accepted. Were they changed at an earlier point?

jmsmkn: The test for the CHOICE field were added here, an apparently was never nullable? https://github.com/comic/grand-challenge.org/commit/377d1496fca0c04dc90cbc586dcde7aff91ee055 This doesn't make sense to me, what am I missing?

jmsmkn: The first null NUMBER or CHOICE answers were created last year:

```python
>>> Answer.objects.filter(answer__isnull=True, question__answer_type=AnswerType.NUMBER).order_by("created").first().created
datetime.datetime(2022, 3, 23, 14, 21, 53, 800920, tzinfo=datetime.timezone.utc)
>>> Answer.objects.filter(answer__isnull=True, question__answer_type=AnswerType.CHOICE).order_by("created").first().created
datetime.datetime(2021, 10, 5, 14, 13, 34, 282831, tzinfo=datetime.timezone.utc)
```

How? Is this test actually testing what it is supposed to?

jmsmkn: I think we should have a separate test with that checks that blank answers are accepted regardless of the required status:

```python
@pytest.mark.parametrize(
  "answer_type,answer",
  (
AnswerType.SINGLE_LINE_TEXT ""
AnswerType.MULTI_LINE_TEXT ""
AnswerType.TEXT ""
AnswerType.BOOL False
AnswerType.NUMBER None
AnswerType.CHOICE None
AnswerType.MULTIPLE_CHOICE []
AnswerType.POINT None
AnswerType.MULTIPLE_POINTS None
AnswerType.DISTANCE_MEASUREMENT None
AnswerType.MULTIPLE_DISTANCE_MEASUREMENTS None
AnswerType.BOUNDING_BOX_2D None
AnswerType.MULTIPLE_2D_BOUNDING_BOXES None
AnswerType.POLYGON None
AnswerType.MULTIPLE_POLYGONS None
AnswerType.LINE None
AnswerType.MULTIPLE_LINES None
AnswerType.MASK None
AnswerType.ANGLE None
AnswerType.MULTIPLE_ANGLES None
AnswerType.ELLIPSE None
AnswerType.MULTIPLE_ELLIPSES None
))
@pytest.mark.parametrize(
  "is_required",
  (False, True, ),
)
def test_blank_answers(answer_type, answer, is_required):
  ...
  assert response.status_code == 201
```

But obviously valid python.

chrisvanrun: I have added your suggested test. TIL you can chain `parametrize()`. This focuses on the required answers and subtypes. 

However, I don't think we can/should revert the `test_answer_is_correct_type`. For that specific test, the `None` answers for annotations would stumble over the new logic for required questions and answers. Setting the test question to not required allows us to test the typing better: both empty answers and regular answers are testable for validity.

I re-arranged the cases a bit to bundle the answer_types together. Imho this makes this a bit more readable.

app/grandchallenge/reader_studies/models.py
```diff
@@ -1441,6 +1447,33 @@ def allow_null_types(self):
...
+    def empty_answer_values(self):
+        """Returns a list of answer values which are considered to be empty"""
+        result = []
+        if self.answer_type in self.allow_null_types:
+            result.append(None)
+        if self.answer_type in self.allow_blank_types:
+            result.append("")
+        if self.answer_type in self.allow_empty_list_types:
+            result.append([])
+        return result
```
jmsmkn, line 1475: There's a risk here that you still have multiple empty answer representations. I think it is better to build a map of answer type to empty answer representation, e.g.

```python
EMPTY_ANSWER_VALUES = {
  AnswerType.TEXT: "",
  AnswerType.MULTIPLE_CHOICE: [],
  ...
}

...

  def empty_answer_values(self):
    return EMPTY_ANSWER_VALUES[self.answer_type]
```

That way we ensure there is only one representation and it is clear to the developer what to expect. If a key is missing you can either leave it as is or make it explicit that the programmer needs to extend the map.

chrisvanrun:
I've discovered that MASK types can have `None` as a value even when answered, but only temporarily. The two client steps for MASK creation are:
1. Create `None` answer, gain an answer pk
2. Upload image, with mention of pk

In the duration between the two steps, the MASK has always been considered to be explicitly not answered. Handling this nicely would require some kind of placeholder for to-be-uploaded masks and a patch in CIRRUS. I don't have the appetite for this atm nor do I think it terribly important. It does entail that `None` MASK answers are allowed even when the question is set to be required.

jmsmkn:
I thought this was handled here, that the answer posted by CIRRUS is an object with `{"upload_session_pk": <uuid>}`, which is then cross checked in https://github.com/comic/grand-challenge.org/blob/115fb964708c19a3163b7d7fa95e30d4e5458797/app/grandchallenge/reader_studies/tasks.py#L179-L182

chrisvanrun:
> I thought this was handled here, that the answer posted by CIRRUS is an object with `{"upload_session_pk": <uuid>}`, which is then cross checked in
> 
> https://github.com/comic/grand-challenge.org/blob/115fb964708c19a3163b7d7fa95e30d4e5458797/app/grandchallenge/reader_studies/tasks.py#L179-L182

I sort of expected that as well, but when checking out the client code, I concluded that an empty answer is created first.

https://github.com/DIAGNijmegen/rse-cirrus-core/blob/3fe524fc4f690234732659705522e88cef6d2479/src/cirrus/DIAG/CIRRUSCore/Modules/Scripts/python/cirrus/workstations/core_web/app_logic/readerstudy/sequencing.py#L368-L383

jmsmkn:
Ah no, I see, you cannot create an Upload Session without an answer. I thought you could create it and then patch it later. I'm fairly sure that is how it used to work, ah well.

jmsmkn (approved):

chrisvanrun:
Ups, unexpected tests fail. On it!

PR #2930, opened 2023-07-13
Fix display set delete

### LINKED ISSUE #2921: Cannot Delete Display Sets
MicrosoftTeams-image (7)

Likely this is because of the new strict in-line JS prevention.

jmsmkn:
Closes #2921


chrisvanrun (approved):
Looks good! This prevents the usage of HTMX for DRF and at the same time cleans up the old image cases thingies in reader studies.

PR #2947, opened 2023-07-25
Add "openseadragon" specialized viewer and fix bug in validator

pkcakeout:
Issue https://github.com/DIAGNijmegen/rse-roadmap/issues/245

Adds a "specialized_view" called "openseadragon". Question mainly is if we want to go the generic route, for example, "clientside_renderer" or specific with "openseadragon". I kind of do not know... I know that we want to experiment with a few types so going the specific route makes it easy to load multiple implementations of client-side viewers side-by-side, but in the long run we probably do not want to support 20 different clientside viewers so we would need to remove them again, which would invalidate stored hanging protocols. So what is better? I do not know...

Anyhow, fixed a pre-existing issue with the "if" statement in the json-schema as well.


pkcakeout:
@HarmvZ  CCing you as well... `specialized_views` already existed. I think it makes sense to carry this through...

jmsmkn:
My guess is that it should be something generic in case we don't want to run with OSD forever, but want to pick an alternative pathology client side viewer. But doesn't matter so much, will leave it up to you.

HarmvZ (approved):
I think this is fine for now. If we want to change it later it's not that hard to migrate existing hanging protocols anyway.

HarmvZ (dismissed):
Just thought about this some more and actually, changing it is a bit more complex because it also needs to be changed in CIRRUS so we need a whole deprecation strategy again. Let's make it more generic. I would say something like `client-rendered-pathology`.

pkcakeout:
Okay, generic it is!

pkcakeout:
> Just thought about this some more and actually, changing it is a bit more complex because it also needs to be changed in CIRRUS so we need a whole deprecation strategy again. Let's make it more generic. I would say something like `client-rendered-pathology`.

Short explanation: I chose a higher-level of "generic", because then we can specialize further later. If we do not choose the "most specialized option" now, we should use the "most generic option" for now so that it is more likely that we - hopefully - won't need any extra classes and can do everything with this one. Right now , we do not have a clientside viewer for non-pathology scans, but I think that is fine, we can fallback to the other viewers for those images!

If we would use a `pathology-client-side` view item, we guarantee that we need to add an additional class later for other images, which is what we would seek to avoid by choosing the generic option.

jmsmkn (approved):

PR #2948, opened 2023-07-27
Blocked domain helptext

miriam-groeneveld:
Provides some more information on how users can verify their account when their email address domain is from a blocked list.


jmsmkn (approved):
We should really extend the Verification form to gather this information rather than using email.

PR #2957, opened 2023-08-04
Add last update-time of combined leaderboard

chrisvanrun:
Part of the pitch:
- https://github.com/DIAGNijmegen/rse-roadmap/issues/243

Adds:
image

### DIFF
--- a/app/grandchallenge/evaluation/models.py
+++ b/app/grandchallenge/evaluation/models.py
@@ -1082,19 +1082,30 @@ class CombinedLeaderboard(TitleSlugDescriptionModel, UUIDModel):
             raise NotImplementedError
 
     @cached_property
-    def combined_ranks(self):
-        combined_ranks = cache.get(self.combined_ranks_cache_key)
+    def _combined_ranks_object(self):
+        result = cache.get(self.combined_ranks_cache_key)
 
         if (
-            combined_ranks is None
-            or combined_ranks["phases"]
-            != {phase.pk for phase in self.public_phases}
-            or combined_ranks["combination_method"] != self.combination_method
+            result is None
+            or result["phases"] != {phase.pk for phase in self.public_phases}
+            or result["combination_method"] != self.combination_method
         ):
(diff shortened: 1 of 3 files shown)


app/grandchallenge/evaluation/models.py
```diff
@@ -1143,6 +1158,7 @@ def update_combined_ranks_cache(self):
         combined_ranks = []
         num_phases = self.public_phases.count()
 
+        now = timezone.now()
         for user, evaluations in self.users_best_evaluation_per_phase.items():
             if len(evaluations) == num_phases:  # Exclude missing data
```
miriam-groeneveld, line 1163: Maybe not part of this pr perse, but isn't this incorrect? What if a challenge has 3 public phases, and a combined leaderboard for 2 phases (2+3). If a user submitted to all 3 phases, their results are excluded I think?

chrisvanrun: The `public_phases` are property of `CombinedLeaderboard` and hence are scoped to linked phases for the board, not the challenge as a whole.

chrisvanrun: Kicking results out if one of the phases is not public is a stop-gap to prevent result leakage. Maybe a warning should be added for the challenge admins only that signifies this on the leaderboard detail page?

miriam-groeneveld: Ah, sorry, my bad.

miriam-groeneveld (approved):
LGTM

app/grandchallenge/evaluation/models.py
jmsmkn: > If a user submitted to all 3 phases, their results are excluded I think?

Shouldn't be an issue. `self.users_best_evaluation_per_phase` is also limited to `self.public_phases` (e.g. the phases we are interested in), so results will not be kicked out if the user has submitted to additional public/private phases.

Issue #2893, opened 2023-06-23
Algorithm job runtime metrics do not seem to be present


jmsmkn:

jmsmkn:
There are two warnings:

```
WARN Infinite extent for field "hoursminutesseconds_Timestamp": [Infinity, -Infinity] [vega.min.b43e106fae93.js:1:1416](https://d19vzld1wvxwbr.cloudfront.net/static/vendored/vega/vega.min.b43e106fae93.js)
WARN Infinite extent for field "Percent100": [Infinity, -Infinity]
```

Seems related to https://github.com/vega/vega-lite/issues/6922, goes away when we remove the transform for the hover element. The Vega editor also does not show this graph immediately and only on refresh.

jmsmkn:
Source


```js
{
  "$schema": "https://vega.github.io/schema/vega-lite/v5.json",
  "width": "container",
  "padding": 0,
  "title": "ml.g4dn.2xlarge / 8 CPU / 32 GB Memory / 1x T4 GPU",
  "data": {
    "values": [
      {
        "Metric": "CPUUtilization",
        "Timestamp": "2023-06-27T12:53:00+00:00",
        "Percent": 1.8073500000000002
      },
      {
        "Metric": "DiskUtilization",
        "Timestamp": "2023-06-27T12:53:00+00:00",
        "Percent": 0.0981705
      },
      {
        "Metric": "GPUMemoryUtilization",
        "Timestamp": "2023-06-27T12:53:00+00:00",
        "Percent": 0
      },
      {
        "Metric": "GPUUtilization",
        "Timestamp": "2023-06-27T12:53:00+00:00",
        "Percent": 0
      },
      {
        "Metric": "MemoryUtilization",
        "Timestamp": "2023-06-27T12:53:00+00:00",
        "Percent": 0.0152067
      }
    ]
  },
  "layer": [
    {
      "transform": [{"calculate": "100*datum.Percent", "as": "Percent100"}],
      "encoding": {
        "x": {
          "timeUnit": "hoursminutesseconds",
          "field": "Timestamp",
          "title": "Local Time / HH:MM:SS"
        },
        "y": {
          "field": "Percent100",
          "type": "quantitative",
          "title": "Utilization / %"
        },
        "color": {"field": "Metric", "type": "nominal"}
      },
      "layer": [
        {"mark": "line"},
        {
          "transform": [{"filter": {"param": "hover", "empty": false}}],
          "mark": "point"
        }
      ]
[...]

PR #2958, opened 2023-08-07
Added a fix for charts not appearing in result logs

james-milburn-crowe:
A fix for #2893 - Algorithm job runtime metrics do not seem to be present

The charts were present on the page, but not rendering

As per https://vega.github.io/vega-lite/docs/size.html#specifying-responsive-width-and-height 
Vega charts need a resize event called if the container size gets altered - this happens when we switch tabs

I've tried to minimise the number of these events, by setting it to trigger only when the logs tab loads. 

Other options:
We could load the chart only when users swap to the tab - this would probably mean a faster initial page load, but more calls to the server if users swap back and forth between logs and info

Setting the chart to display:inline, but with visible:hidden means the overlay shows up with mouseover

### DIFF
--- a/app/grandchallenge/algorithms/static/algorithms/js/display_result_text.mjs
+++ b/app/grandchallenge/algorithms/static/algorithms/js/display_result_text.mjs
@@ -13,7 +13,7 @@ $('#resultInfoModal').on('show.bs.modal', function (event) {
     modal.find('#resultDetailLink').attr("href", `${button.data("pk")}/`);
 })
 
-$('[href="#v-pills-logs"]').on('shown.bs.tab', function (event) {
+$('#v-pills-logs-tab').on('shown.bs.tab', function (event) {
     // a resize is needed when switching to the logs tab to make the vega chart load correctly
     window.dispatchEvent(new Event('resize'))
 })


app/grandchallenge/algorithms/static/algorithms/js/display_result_text.mjs
```diff
@@ -12,3 +12,8 @@ $('#resultInfoModal').on('show.bs.modal', function (event) {
     modal.find('.modal-title').text(button.data('title'));
     modal.find('#resultDetailLink').attr("href", `${button.data("pk")}/`);
 })
+
+$('[href="#v-pills-logs"]').on('shown.bs.tab', function (event) {
```
chrisvanrun, line 16: I think this works, but just curious: why not the id of the target element (i.e. `v-pills-logs-tab`)?

james-milburn-crowe: I'll replace that! I had a good reason, but on further digging it's not needed in this version of bootstrap. Just tested and it seems fine to just swap to the element

chrisvanrun (approved):
LGTM!

Issue #2892, opened 2023-06-23
Datatables.js broken for Algorithm/Archive permission requests


jmsmkn:
See errors in JS console on the permission request list pages.

james-milburn-crowe:
related to this - https://stackoverflow.com/questions/28454203/jquery-datatables-cannot-read-property-adatasort-of-undefined

Error vanishes if you set order to 7 in 
``` $(document).ready(function () {
    $('#usersTable').DataTable({
        order: [[8, "asc"], [0, "desc"]], ```

in permission_request_displaytable.mjs

looks like we have 8 columns, and an off-by one error

PR #2966, opened 2023-08-10
Fixed error that was being thrown on the algorithm permissions page

james-milburn-crowe:
In permission_data_tables.js , we were trying to sort by column 9, when there are only 8 columns

resolves issue #2892


chrisvanrun:
Well that was straightforward =P

chrisvanrun (approved):

PR #2969, opened 2023-08-11
Update makefile

miriam-groeneveld:
Do not use buildx for the dist container as it does not support custom networks
See also this [commit](https://github.com/comic/grand-challenge.org/commit/f3bf654753c43cc6ae466347bbcdc0fa3f352297). I think the update to poetry caused a new docker build of the web-base image, which fails due to the use of dockerx.


jmsmkn:
The reasoning here doesn't make sense to me as that build step is not using a custom network.

jmsmkn:
Indeed, the build fails at a later point. Updating poetry always causes a base image rebuild. What has likely happened is that the docker version in GitHub actions has been updated so that now using buildkit is the default, and now we need to opt out.

PR #2993, opened 2023-08-23
Add support for multiple hanging protocols

james-milburn-crowe:
Grand challenge side of optional hanging protocols feature

Added support for optional hanging protocols in:
-Algorithm
-Archive
-Reader studies
-Phase

Checked that we are testing that settings are propagated across properly when creating an algorithm from a phase. I've tested it too on the test site, and it seems to display/remember settings as we'd expect.

I've also added "hanging protocol" creation permissions to the demo development fixture user - I can drop that, but it was useful to be able to do while testing.


james-milburn-crowe:
grand challenge side of issue #2993

james-milburn-crowe:
just fixing the migration I tried to squash, sorry

james-milburn-crowe:
I'm not sure I understand why the pull test is failing - I had ended up with two migrations for phase, so I merged them, as they both represented tiny changes. It seems to run the migration correctly on my machine, but break on the test server. Is it doing something incrementally from the last commit?

james-milburn-crowe:
wait, this is broken because I need to merge main into it, which has a migration of the same name, I think.

james-milburn-crowe:
I think this is ready for merge now. I've pulled in origin/main and resolved the previous migration conflict

app/grandchallenge/algorithms/models.py
```diff
@@ -103,6 +103,11 @@ class Algorithm(UUIDModel, TitleSlugDescriptionModel, ViewContentMixin):
         blank=True,
         on_delete=models.SET_NULL,
     )
+    optional_hanging_protocols = models.ManyToManyField(
+        "hanging_protocols.HangingProtocol",
+        related_name="optional_hanging_protocols_of_algorithm",
```
jmsmkn, line 108: In this case for the reverse lookup you would have something like `HangingProtocol.objects.filter(optional_hanging_protocols_of_algorithm=algorithm)`. Thats a bit verbose. Maybe `optional_for_algorithm`?

Similar comment for the other 3 apps.

app/grandchallenge/algorithms/models.py
```diff
@@ -103,6 +103,11 @@ class Algorithm(UUIDModel, TitleSlugDescriptionModel, ViewContentMixin):
         blank=True,
         on_delete=models.SET_NULL,
     )
+    optional_hanging_protocols = models.ManyToManyField(
```
jmsmkn, line 106: You need to define an explicit through table for new ManyToManyFields as this helps with long term maintenance https://docs.djangoproject.com/en/4.2/ref/models/fields/#django.db.models.ManyToManyField.through

Same comment for the other 3 apps.

app/grandchallenge/algorithms/migrations/0043_algorithm_optional_hanging_protocols.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 4.1.10 on 2023-08-16 13:02
```
jmsmkn, line 1: It is best to delete the new migrations that you're adding and then regenerating them when rebasing on main.

Same comment for the other 3 apps.

scripts/development_fixtures.py
```diff
@@ -171,6 +171,7 @@ def _set_user_permissions(users):
     add_archive_perm = Permission.objects.get(codename="add_archive")
     users["archive"].user_permissions.add(add_archive_perm)
     users["demo"].user_permissions.add(add_archive_perm)
+    assign_perm("hanging_protocols.add_hangingprotocol", users["demo"])
```
jmsmkn, line 174: This can be removed.

app/grandchallenge/archives/models.py
```diff
@@ -89,6 +89,12 @@ class Archive(UUIDModel, TitleSlugDescriptionModel, ViewContentMixin):
         blank=True,
         on_delete=models.SET_NULL,
     )
+    optional_hanging_protocols = models.ManyToManyField(
+        "hanging_protocols.HangingProtocol",
+        related_name="optional_hanging_protocols_of_archive",
+        blank=True,
+        help_text="Optional alternative hanging protocols for this archive",
```
jmsmkn, line 96: Why does this one have help text and the others not? Make it consistent.

jmsmkn (commented):
I cannot find a test for the serializers, if CIRRUS is relying on a new API then it should be tested with no optional protocols set, and many optional protocols set.

app/grandchallenge/algorithms/models.py
james-milburn-crowe: just testing this now, but think I've got it

james-milburn-crowe: I also see why it's best practice, too - you can't alter field to switch to a through table, you have to drop and re-add the whole thing.

james-milburn-crowe: I think this is right, as per the latest commit

app/grandchallenge/algorithms/models.py
james-milburn-crowe: fixed

app/grandchallenge/archives/models.py
james-milburn-crowe: made them all consistent

scripts/development_fixtures.py
james-milburn-crowe: removed!

james-milburn-crowe:
Added in tests for optional hanging protocol serializers, to match the existing hanging protocol serializers. We aren't testing it in phase there. Is there a good reason for that, or should I set it up to test?

jmsmkn:
Phases do not have an API endpoint, so they don't have a Serializer, why would they need to be added?

app/grandchallenge/algorithms/models.py
```diff
@@ -103,6 +103,13 @@ class Algorithm(UUIDModel, TitleSlugDescriptionModel, ViewContentMixin):
...
+        "hanging_protocols.HangingProtocol",
+        through="OptionalProtocolsAlgorithm",
+        related_name="optional_of_algorithm",
```
jmsmkn, line 109: I don't think that "of" is the correct preposition, IMO it should be "for".

`HangingProtocol.objects.filter(optional_for_algorithm=algorithm)` => Give me the hanging protocols that are optional for this algorithm.

Same for the other apps.

app/grandchallenge/algorithms/models.py
```diff
@@ -923,3 +930,12 @@ def delete(self, *args, **kwargs):
 
     class Meta(RequestBase.Meta):
         unique_together = (("algorithm", "user"),)
+
+
+class OptionalProtocolsAlgorithm(models.Model):
```
jmsmkn, line 935: Lets have consistency of naming with the rest of the codebase. `Protocols` should be `HangingProtocol`, non-plural.

Same for the other apps.

app/grandchallenge/algorithms/serializers.py
```diff
@@ -96,6 +96,12 @@ class JobSerializer(serializers.ModelSerializer):
         read_only=True,
         allow_null=True,
     )
+    optional_hanging_protocols = HangingProtocolSerializer(
+        many=True,
+        source="algorithm_image.algorithm.optional_hanging_protocols",
```
jmsmkn, line 101: We need to ensure that this does not introduce any N+1 queries. Inspect the SQL calls with the django debug toolbar, or directly with shell plus. Ensure that the API list and detail views are not having to do multiple database requests when multiple objects have multiple optional hanging protocols. If so, select related or prefetch related statements may need to be added to the queryset in the view.

Same for the other apps.

james-milburn-crowe:
Just making these changes, and looking at the database calls. 

I spotted a phase serializer in the evaluation serializers, but I didn't check that it was actually used by the api. If there's no need to test it, then that's fine!

jmsmkn:
> I've got prefetch working on each of the views, so I think the N+1 problem is fixed. I think that optional_hanging_protocols is only fetched by phase on phase update, where I think we don't want to prefetch. Is that correct?

You would still have the same problem there so yes, you would need it.

Please tidy up the code so it is consistent with the rest of the codebase. Class attributes should be in a single block. Filtering the queryset happens early in the view, so should be after any permission check methods and before form and context generation. Using a class attribute to filter the queryset is preferable to a method, so use those if possible.

james-milburn-crowe:
Just got this updated. I went back through with the debugger, and found two additional places where the N+1 error for hanging protocols was occurring, and I think I've tided up the code to be in line with the rest of the codebase. Ran the tests just now, so I'm updating it

jmsmkn (approved):

PR #3005, opened 2023-08-30
Add SVG icon computed property to Hanging protocol model

HarmvZ:
Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/247

- Adds svg icon computed property to the model
- Adds the icon and other required properties to the serializer

### List view
image

### Detail view:
image

### DIFF
--- a/app/grandchallenge/hanging_protocols/models.py
+++ b/app/grandchallenge/hanging_protocols/models.py
@@ -216,7 +216,7 @@ class HangingProtocol(UUIDModel, TitleSlugDescriptionModel):
             )
 
     @property
-    def svg_icon(self):
+    def svg_icon(self) -> str:
         width = len(self.json)
         height = 1
 
--- a/app/grandchallenge/hanging_protocols/serializers.py
+++ b/app/grandchallenge/hanging_protocols/serializers.py
@@ -1,17 +1,9 @@
-from drf_spectacular.types import OpenApiTypes
-from drf_spectacular.utils import extend_schema_field
 from rest_framework import serializers
 
 from grandchallenge.hanging_protocols.models import HangingProtocol
 
 
 class HangingProtocolSerializer(serializers.ModelSerializer):
-    svg_icon = serializers.SerializerMethodField()
-
-    @extend_schema_field(OpenApiTypes.STR)
-    def get_svg_icon(self, object):
-        return object.svg_icon
-
     class Meta:
         model = HangingProtocol
         fields = ["json", "title", "pk", "svg_icon"]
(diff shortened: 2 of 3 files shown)


app/grandchallenge/hanging_protocols/models.py
```diff
@@ -213,6 +213,33 @@ def save(self, *args, **kwargs):
...
+            h = height_px - 2 * padding
+            x = padding + i * w
+            y = padding
+            if "x" in self.json[0]:
+                w = (width_px - 2 * padding) * vi["w"] / width
+                h = (height_px - 2 * padding) * vi["h"] / height
+                x = padding + (width_px - 2 * padding) * vi["x"] / width
+                y = padding + (height_px - 2 * padding) * vi["y"] / height
+            svg += f'\t<rect x="{x}" y="{y}" width="{w}" height="{h}" stroke-width="{stroke_width}" />\n'
+        svg += "</svg>"
```
jmsmkn, line 240: Could you move this part to a Django template so that values are properly escaped?

app/grandchallenge/hanging_protocols/templates/hanging_protocols/hangingprotocol_detail.html
```diff
@@ -18,7 +18,10 @@
 {% endblock %}
 
 {% block content %}
-    <h3>{{ object.title }}</h3>
+    <h3>
+        <span style="stroke: #000">{{ object.svg_icon|safe }}</span>
```
jmsmkn, line 22: Best to avoid using `safe` in templates. Same for `app/grandchallenge/hanging_protocols/templates/hanging_protocols/hangingprotocol_list.html`.

app/grandchallenge/hanging_protocols/models.py
jmsmkn: Actually you might be able to do it with 3 instances of `format_html` and passing the kwargs.

app/grandchallenge/hanging_protocols/serializers.py
```diff
@@ -6,4 +6,4 @@
 class HangingProtocolSerializer(serializers.ModelSerializer):
     class Meta:
         model = HangingProtocol
-        fields = ["json"]
+        fields = ["json", "title", "id", "svg_icon"]
```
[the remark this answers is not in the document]
jmsmkn: Also please add a test.

app/grandchallenge/hanging_protocols/models.py
HarmvZ: I ended up converting it to a template, that way it's also easy to embed in the list and detail templates so we can avoid using the `safe` filter there.

jmsmkn (approved):
Take a look at the two commits I made. The first uses `format_html` which makes reuse of `svg_icon` much easier (no mixin/settings) and avoids the use of safe. The second is more straightforward for defining the schema.

PR #3006, opened 2023-08-30
added title and id field for hanging protocol serializers

james-milburn-crowe:
Added title and id field to hanging protocol serializers, as part of #2993 

Added these to both hanging_protocol and optional_hanging_protocol serializer tests


jmsmkn:
Done already in #3005

Issue #2949, opened 2023-07-27
Homepage carousel looks bad on small screens


jmsmkn:
748px wide:

<img width="749" alt="image" src="https://github.com/comic/grand-challenge.org/assets/12661555/f25c6f5b-83e9-4efd-b18b-17f3bf752a33">

500px wide:

<img width="505" alt="image" src="https://github.com/comic/grand-challenge.org/assets/12661555/46c8fa95-e900-4b39-bc38-a7ae506cd170">

james-milburn-crowe:
Think the fixes here would be: 
 < 500px - can be very similar, but ensure the image is centered
500-748px - I'll try a couple of options, but image and text should not be cut, and whitespace in the middle should prehaps reduce in size

pkcakeout:
Probably better to use `dp` instead of `px` - `px` really does not work well on high-density screens if I am not mistaken!

james-milburn-crowe:
ok, looks like the issue here is that we use columns inside a carousel, and it looks like this makes it overflow the container borders (So at medium breakpoints, the cutoff image extends out into the whitespace, but is hidden). 

Removing col-12 col-md-3 for the carousel items gives a good result, and we could just let bootstrap carousel handle the sizing

james-milburn-crowe:
![Screenshot 2023-08-08 180846](https://github.com/comic/grand-challenge.org/assets/14093674/97321259-58f2-4dde-bb2e-6647fd0f4ce0)
 Here's what's currently happening

there's this way of adding columns inside a carousel - https://stackoverflow.com/questions/48648916/bootstrap-4-carousel-columns-instead-of-carousel-item  we just have to make some inner row/column setups for it

james-milburn-crowe:
ended up working yesterday on some of the file deduplication things, instead of this, but it's finished today

jmsmkn:
I think this issue already existed but the corners of the image are not rounded on md+ screens:

<img width="749" alt="image" src="https://github.com/comic/grand-challenge.org/assets/12661555/f25c6f5b-83e9-4efd-b18b-17f3bf752a33">

james-milburn-crowe:
strangely, this is only on the first image loaded - all others are rounded on md+

jmsmkn:
Could also be the shape - the image is square or portrait rather than landscape.

james-milburn-crowe:
just took a quick look at this at the end of the day - it looks like carousel-img-container is the culprit - from core.css

```
[...]

PR #3007, opened 2023-08-31
fix for carousel images being displayed with non rounded corners

james-milburn-crowe:
Carousel image was being displayed without rounded corners, as it was overflowing the bounds of its surrounding div

Proposed fix is to apply the image sizing class directly to the image, rather than the containing div, forcing it to not overflow. I've altered the class name, as it no longer applies to the surrounding container (we do not use it anywhere else)

Below are pics with the altered class applied to the image div.

capture_with_sizing
Capture_with_sizing_rectange
Capture_smaller_size


james-milburn-crowe:
Solves #2949

jmsmkn (approved):

Issue #3009, opened 2023-09-01
Hanging protocol previews sometimes broken in list view


jmsmkn:
**Describe the bug**
See Main with draggable smaller view or Tertiary on https://grand-challenge.org/hanging-protocols/:

<img width="362" alt="Screenshot 2023-09-01 at 08 44 14" src="https://github.com/comic/grand-challenge.org/assets/12661555/2afe6464-9566-4f34-8be3-4677ec63dcef">
<img width="362" alt="Screenshot 2023-09-01 at 08 44 26" src="https://github.com/comic/grand-challenge.org/assets/12661555/41877790-2ebd-4692-b1d5-b2a1fb8edd82">

The SVG is correct but the grey previews with names are wrong. On the detail pages they look fine:

https://grand-challenge.org/hanging-protocols/main-with-draggable-smaller-view/
<img width="1148" alt="image" src="https://github.com/comic/grand-challenge.org/assets/12661555/557f3639-d9ed-4dee-99e2-aff8f33ebbce">
https://grand-challenge.org/hanging-protocols/tertiary/
<img width="1144" alt="image" src="https://github.com/comic/grand-challenge.org/assets/12661555/5868d20e-6c20-4dd1-a1d3-4379cff2f760">

james-milburn-crowe:
great - looking at it

james-milburn-crowe:
looks like all the elements are still being placed, but that they're ending up covered by the main box

james-milburn-crowe:
z index is being set to -1 and -2 on those elements, causing them to appear behind the card.

james-milburn-crowe:
Ok, a little more research: The only ones that use a negative z-index are those hanging protocols, and there it doesn't render correctly. It seems to render fine on the other view, because it's not got a card it can end up behind.

Digging into the js, it looks like we translate "order" to the inverse of its z-index, and these are the only one using orders

I think shunting the cards to z-index -99 or just turning off their background color is probably the simplest fix

alternately, we could start our highest ordered object at a z index equal to the number of viewports. That way, when someone creates a 5000 viewport hanging protocol, it'll still render correctly (At least with regard to correct layering)

jmsmkn:
From the schema we know that there are a maximum of 20 viewports.

jmsmkn:
Closed by #3011

PR #3011, opened 2023-09-01
fix for hanging protocols displaying incorrectly on hanging protocol cards

james-milburn-crowe:
A small fix for #3009  - the issue was that z-indexes were being set as negative, causing viewports with an order >0 to appear behind the card

We have a max of 20 viewpoints, and therefore I've set z-index to count down from 20, rather than going negative. 

This keeps the correct ordering, but stops things from being placed behind the cards

We could also set card z-index to -99 or similar, but I think that is more likely to cause some future unforeseen bootstrap issues


HarmvZ:
I slightly improved your implementation by using the max value for `order` to determine the values for z-index. This is better because we currently do not put constraints on the usage of `order` except that it should be an integer. This improvement should work for any integer value.

In the meantime, I got in a fight with pycharm/git and line endings - that's fixed now.

james-milburn-crowe:
ah, thanks! that makes sense - but I think this might go negative - we'd like to avoid z-index being negative, as it puts the item behind the cards

james-milburn-crowe:
just fixing, not ready for merge - it's a much nicer solution, but maxOrder is currently being set to NaN, and so z-index is just removed completely. Which still makes the rendering work much more nicely, so honestly we could consider doing that

jmsmkn:
Converting to draft in that case.

james-milburn-crowe:
> > but I think this might go negative
> 
> It shouldn't. How do you think so?

Ah, sorry, my maths was off! works perfectly!

jmsmkn (approved):

Issue #2998, opened 2023-08-24
Make Algorithm Creation Button more obvious


jmsmkn:
Include it on the algorithms page with a selection of open Phases, and make it a big button on the submission page.

james-milburn-crowe:
Was just going to pick up some small things - hope that's the right thing to be doing - this one looks nice, as I found it surprisingly tough to find the button when testing

james-milburn-crowe:
![algorithm_button](https://github.com/comic/grand-challenge.org/assets/14093674/80c5e124-f487-4d18-b386-ef2c5735a233)

on the algorithm page

I think I'd need to do some more things to be able to view it on the test server demo page - I'll look into it, and post in teams if I can't work it out

james-milburn-crowe:
Just an update, got the fixtures working, and a button on the algorithm submission page - trying to style it nicely, and sort the other one

jmsmkn:
The other one is probably going to be to difficult, not worth it.

james-milburn-crowe:
I had a bit of work to do for health-ri (prisma) this afternoon, but think this is working now. I ended up moving it from the template to the form, as it is where the old link was set (and it means it styles nicely)

I've not done a lot with crispy forms though, so this may be completely the wrong way of setting it. I'll make a PR, and it might be totally incorrect

james-milburn-crowe:
![Capture_al_submit](https://github.com/comic/grand-challenge.org/assets/14093674/1533614b-9b0c-444d-b3f5-4ce2b792ac52)

![Capture_al_submit_2](https://github.com/comic/grand-challenge.org/assets/14093674/fee6bc30-10b1-4553-b781-232b4825dd69)

It took a little time to figure out how to sort, but there's a small screen here that now appears when there are no valid algorithms. I've had a stab at wording, but there's a little more formatting to do

PR #3013, opened 2023-09-04
Add Alert to Submission Page When User has No Algorithms

james-milburn-crowe:
Replaces the small create algorithm link with a large create algorithm button on the phase submission page

Capture

Part of #2998

### DIFF
--- a/app/grandchallenge/evaluation/forms.py
+++ b/app/grandchallenge/evaluation/forms.py
@@ -357,7 +357,7 @@ class SubmissionForm(
                     },
                 ),
             )
-            if not self.fields["algorithm_image"]:
+            if not self.fields["algorithm_image"].queryset:
                 self.helper.layout = Layout(
                     ButtonHolder(
                         HTML(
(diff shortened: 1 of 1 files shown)


jmsmkn:
Buttons there are to do with form submission, should be reserved for saving, cancelling, deleting etc. I think a new section is better, that is rendered instead of the form if the user does not have a valid algorithm, that explains that this is an algorithm submission phase and that they first need to set up their algorithm.

james-milburn-crowe:
> Buttons there are to do with form submission, should be reserved for saving, cancelling, deleting etc. I think a new section is better, that is rendered instead of the form if the user does not have a valid algorithm, that explains that this is an algorithm submission phase and that they first need to set up their algorithm.

I'll put a section together

james-milburn-crowe:
It took a little work to get this showing up properly - 
Capture45
I have one last thing to check tomorrow, but if this looks ok, then it'll be done!

james-milburn-crowe:
This is all done and working (no rush on reviews, I just wanted to check it wasn't languishing because you were waiting for input from me)

PR #3031, opened 2023-09-18
Fix creating permalinks in emails HTML renders

### LINKED ISSUE #3030: Email markdown trouble with rendering to HTML
Any header like markdown (i.e. `### TITLE`) gets automagically rendered to include an anchor link:

097bab37-96ea-4d01-a193-851521e3a125

A quick workaround is manually defining the `<hx></hx>` tags in the Markdown so the plugin does not parse them.

chrisvanrun:
Closes #3030 

See comment on the related ticked. This fixes the:
* detail email page
* the actually sent email

However, it does not fix the preview pane of the `MarkdownWidget` of the form since it is a rabbit hole.

### DIFF
--- a/app/config/settings.py
+++ b/app/config/settings.py
@@ -11,7 +11,6 @@ from django.contrib.messages import constants as messages
 from django.core.exceptions import ImproperlyConfigured
 from django.urls import reverse
 from machina import MACHINA_MAIN_STATIC_DIR, MACHINA_MAIN_TEMPLATE_DIR
-from markdown.extensions.toc import TocExtension
 from sentry_sdk.integrations.celery import CeleryIntegration
 from sentry_sdk.integrations.django import DjangoIntegration
 from sentry_sdk.integrations.logging import ignore_logger
@@ -779,10 +778,6 @@ BLEACH_STRIP = strtobool(os.environ.get("BLEACH_STRIP", "True"))
 
 # The markdown processor
 MARKDOWNX_MEDIA_PATH = datetime.now().strftime("i/%Y/%m/%d/")
-MARKDOWNX_MARKDOWN_PERMALINK_EXTENSION = TocExtension(
-    permalink=True,
-    permalink_class="headerlink text-muted small pl-1",
-)
 MARKDOWNX_MARKDOWN_EXTENSIONS = [
     "markdown.extensions.fenced_code",
     "markdown.extensions.tables",
--- a/app/config/settings.py
+++ b/app/config/settings.py
@@ -779,15 +779,16 @@ BLEACH_STRIP = strtobool(os.environ.get("BLEACH_STRIP", "True"))
 
 # The markdown processor
 MARKDOWNX_MEDIA_PATH = datetime.now().strftime("i/%Y/%m/%d/")
+MARKDOWNX_MARKDOWN_PERMALINK_EXTENSION = TocExtension(
+    permalink=True,
+    permalink_class="headerlink text-muted small pl-1",
+)
 MARKDOWNX_MARKDOWN_EXTENSIONS = [
     "markdown.extensions.fenced_code",
     "markdown.extensions.tables",
     "markdown.extensions.sane_lists",
     "markdown.extensions.codehilite",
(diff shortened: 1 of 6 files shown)


app/config/settings.py
```diff
@@ -779,15 +779,16 @@
 
 # The markdown processor
 MARKDOWNX_MEDIA_PATH = datetime.now().strftime("i/%Y/%m/%d/")
+MARKDOWNX_MARKDOWN_PERMALINK_EXTENSION = TocExtension(
```
jmsmkn, line 782: I think that it's better to just initialise this in `md2html`. The namespace `MARKDOWNX_` belongs to the markdownx package, and `MARKDOWNX_MARKDOWN_PERMALINK_EXTENSION` is not an option there.

app/grandchallenge/core/templatetags/bleach.py
```diff
@@ -28,14 +28,30 @@ def clean(html: str):
...
+    return md2html(
+        markdown=markdown,
+        link_blank_target=True,
+        create_permalink_for_headers=False,
+    )
+
+
+@register.filter
+def md2html(
+    markdown: str | None,
```
jmsmkn, line 41: Getting complex, can we add a `*` for kwargs only after `markdown`?

app/grandchallenge/core/templatetags/bleach.py
```diff
@@ -28,14 +28,30 @@ def clean(html: str):
 
 
 @register.filter
-def md2html(markdown: str | None, link_blank_target=False):
+def md2html_email(markdown: str | None):
```
jmsmkn, line 31: Do we need another filter or is this possible to achieve with just the options?

chrisvanrun: It's possible with the options, but since they are booleans it gets verbose really quick to keep it readable:
```Jinja
{% set args = { 'option_a': True, 'option_b': False } %}
{% body | md2html(args)
```

Since this specific function needs to be called at two places, we'd need to repeat the exact pattern at both places. 

Alternatively, we could add a single boolean switch to mark it an email that is to be rendered...

chrisvanrun:
TIL that jinja2 registered filters do caching, but only do so for the (first) piped argument, ignoring any options or other arguments.

Editing an email on the form meant the `md2html` filter got directly called for the preview. This made a cache entry that the follow-up `md2email_html` call (which calls the `md2html`) had an unintentional hit on.

A lot of head-to-desk banging was involved in figuring out this flow of events.

jmsmkn:
> TIL that jinja2 registered filters do caching, but only do so for the (first) piped argument, ignoring any options or other arguments.
> 
> Editing an email on the form meant the `md2html` filter got directly called for the preview. This made a cache entry that the follow-up `md2email_html` call (which calls the `md2html`) had an unintentional hit on.
> 
> A lot of head-to-desk banging was involved in figuring out this flow of events.

I don't really understand this. We're not using jinja2, this is Django templates. If it was doing any caching it shouldn't be across requests.

chrisvanrun:
> I don't really understand this. We're not using jinja2, this is Django templates. If it was doing any caching it shouldn't be across requests.

Sorry, got CIRRUS focused here. For me, they have very similar syntax and I always mix them up. I'll create a simple example of what I think went wrong in my initial PR.

chrisvanrun:
I made the wrong assumptions. What I thought must be caching since it made 0 sense any other way I finally found what was actually going on and it turned out that `main` had a bug.

 I finally noticed that the settings themselves were being updated every time md2html was being called with `link_blank_target=True`. A small bug initially, easy to miss since its effect isn't that big:
- Some links accidentally got turned into blank targets when they should not have. 
- The `md2html` filter got progressively slower since it had an increasingly larger number of `LinkBlankTargetExtension` that it had to apply for each render.

I've now fixed that and it also solves the indirect result I saw with the email preview not rendering correctly.

jmsmkn:
Oh wow! Great find.

jmsmkn (approved):

PR #3056, opened 2023-10-05
Remove mention of algorithm editors having access to try-out results of algorithms

chrisvanrun:
Found this slightly incorrect help text for an algorithm result that is generated by a try-out:

image

Algorithm editors do **not** have access.


app/grandchallenge/algorithms/migrations/0046_alter_job_public.py
```diff
@@ -0,0 +1,21 @@
...
+        ("algorithms", "0045_optionalhangingprotocolalgorithm_and_more"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="job",
+            name="public",
+            field=models.BooleanField(
+                default=False,
+                help_text="If True, allow anyone to download this result along with the input image. Otherwise, only the job creator will have permission to download and view this result.",
```
jmsmkn, line 18: For changes to help text you do not need to create a migration, find the last migration where this field was changed and change it there instead (Algorithms Migration 0001 Line 281)

chrisvanrun: Done!

jmsmkn (approved):

PR #3061, opened 2023-10-17
Move link to evaluation from leaderboard metrics to ordinal rank

chrisvanrun:
One of the Seg.A. orgs was not too sure how to get to the evaluation details of submission during MICCAI. It struck me as something other users might have trouble with as well. There is a certain non-intuitiveness to it, so I thought to throw this suggestion out here and see what others think.

Move the link from each of the aggregate metrics to the ordinal rank.

# Current
image

# Suggestion
image

## Review note
The GitHub integrated differential seems to go haywire on this. Regular PyCharm does a better job:

image

### DIFF
--- a/app/grandchallenge/evaluation/templates/evaluation/leaderboard_row.html
+++ b/app/grandchallenge/evaluation/templates/evaluation/leaderboard_row.html
@@ -4,8 +4,9 @@
 {% load humanize %}
 {% load remove_whitespace %}
 {% load url %}
-
-{{ object.rank|ordinal }}
+<a href="{{ object.get_absolute_url }}">
+    {{ object.rank|ordinal }}
+</a>
 <split></split>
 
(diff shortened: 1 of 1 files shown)


jmsmkn:
Why not just add the link? No problem to have it in both places.

jmsmkn:
Okay, then lets leave it as is as this is what people will be used to clicking.

chrisvanrun:
But does it make sense for new people that the metrics represent a link to get to the evaluation details? That was what this PR was aiming to address.

jmsmkn:
It was always see more information about this metric as that is what is displayed. If we are going to change things I would rather it be done systematically rather than on an ad-hoc basis. In some places the pk is used (but those are ugly), in others an icon (but those take up space), and here a metric or score (as that is what more detail is being provided about). For me it doesn't make sense to now change that to rank. In the combined leaderboard the links are also on the right.

Issue #3062, opened 2023-10-19
Scrolling to anchors goes incorrectly


chrisvanrun:
Using anchors to jump directly to documentation sections doesn't work very nicely. 

Visiting https://grand-challenge.org/documentation/making-a-challenge-submission/#submitting-your-predictions-files 

# Expected

<img width="842" alt="image" src="https://github.com/comic/grand-challenge.org/assets/7871310/4626c86d-686a-4b1a-b744-8ccd4c3f2940">

# Actual
![image](https://github.com/comic/grand-challenge.org/assets/7871310/2d4d45e7-60e5-434b-b86b-6c662dad3ee4)

PR #3063, opened 2023-10-19
Add link to algorithm results in evaluation details

chrisvanrun:
This PR adds a link to the algorithm results in the admin section following a discussion here:
- https://github.com/DIAGNijmegen/rse-roadmap/issues/267

image

In addition, I noticed that the length of the metrics.json is generally longer in production than the bare minimum in the development fixtures. In practice challenge hosts tend to add a per-case metric, blowing up the number of lines to 100+.

I think having an admin section on the detail page is pretty neat. However, it is hard to get to with these numerous metric lines. To tackle this I've added a 'jump to' button in this PR as well. As a bonus, this also makes the existence of the section more obvious for new hosts:

## Small additions
- Pulled in the submission file download link into the new 'Results' section
- Added a scroll padding to take the sticky main nav-bar height into consideration when scrolling anchors. This is nice to have here, but fixes: #3062

### DIFF
--- a/app/grandchallenge/algorithms/templates/algorithms/algorithm_card_body.html
+++ b/app/grandchallenge/algorithms/templates/algorithms/algorithm_card_body.html
@@ -28,7 +28,7 @@
         <a class="badge badge-info above-stretched-link p-1 mr-1"
            href="{{ publication.identifier.url }}"
            title="{{ publication.title }}">
-            <i class="fas fa-file fa-fw mr-1"></i>Article
+            <i class="fas fa-file fa-fw"></i>&nbsp;Article
         </a>
     {% endfor %}
 
--- a/app/grandchallenge/algorithms/templates/algorithms/algorithm_card_body.html
+++ b/app/grandchallenge/algorithms/templates/algorithms/algorithm_card_body.html
@@ -28,7 +28,7 @@
         <a class="badge badge-info above-stretched-link p-1 mr-1"
            href="{{ publication.identifier.url }}"
            title="{{ publication.title }}">
-            <i class="fas fa-file fa-fw"></i>&nbsp;Article
+            <i class="fas fa-file fa-fw mr-1"></i>Article
         </a>
     {% endfor %}
 
--- a/app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
+++ b/app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
@@ -247,6 +247,13 @@
                 </div>
             {% endif %}
 
+            <div class="card-body">
+                <h3 class="card-title">Results</h3>
+                   <button class="btn btn-primary" target="_blank" href="{% url 'algorithms:job-list' slug=object.submission.algorithm_image.algorithm.slug %}">
+                       Go to results of {{ object.submission.algorithm_image.algorithm.title }} <i class="fas fa-angle-double-right"></i>
+                   </button>
+            </div>
+
             {% if object.status == object.SUCCESS %}
                 <div class="card-body">
                     <h3 class="card-title">Visibility</h3>
--- a/app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
+++ b/app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
@@ -294,10 +294,10 @@
                             {% csrf_token %}
                             <input type="hidden" name="published"
                                    value="false">
-                            <button type="submit" class="btn btn-danger">
+                            <a type="submit" class="btn btn-danger">
                                 <i class="fas fa-eye-slash"></i>&nbsp;&nbsp;
                                     Exclude this result from the leaderboard(s)
-                            </button>
+                            </a>
                         </form>
                     {% else %}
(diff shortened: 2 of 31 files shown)


app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
```diff
@@ -247,20 +254,41 @@ <h3 class="card-title">Prerequisite Jobs</h3>
                 </div>
             {% endif %}
 
+            <div class="card-body">
+                <h3 class="card-title">Results</h3>
```
jmsmkn, line 258: Results says to me here "The results of this evaluation". Would Predictions be a better title?

app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
```diff
@@ -32,6 +32,13 @@
 
     <h2>Evaluation</h2>
 
+    {% if "change_challenge" in challenge_perms %}
+        <a href="#admin" class="btn btn-danger mb-3">
+            <i class="fas fa-arrow-down"></i>&nbsp;&nbsp;
+            Admin section
```
jmsmkn, line 38: I don't think this is necessary - if tonnes of metrics are included then the admins should hide the metrics and only show the relevant ones, the scrollbar on the right or page down are enough to get to the section.

app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
```diff
@@ -286,23 +315,12 @@ <h3 class="card-title">Predictions</h3>
                     <a href="data:text/plain;charset=utf-8,{{ predictions|json_dumps|urlencode }}"
                         download="predictions.json"
                         class="btn btn-primary">
-                        <i class="fa fa-download"></i>
+                        <i class="fa fa-download"></i>&nbsp;&nbsp;
```
jmsmkn, line 318: Are the non-breakable spaces really necessary?

app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
chrisvanrun: Good point. I had some confusion about this as well. Looking at the different subsections I'm also spotting a predictions.json download button under 'Predictions'. Going to combine them.

app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
chrisvanrun: How would new hosts know that that section exists?

app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
jmsmkn: Yes, do not tackle layout with spaces, breaks, tables, etc. Remove it so it is consistent with the rest of the site.

app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
jmsmkn: By looking at the contents of the page.

app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
chrisvanrun: Ok. I had actually added it to _make_ it consistent with the major UX item on the page: the challenge nav bar ( 😅 ). I'll smack that as well.

app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
chrisvanrun: Would they not just assume it's just the details of the evaluation? I would.

chrisvanrun: Removed.

chrisvanrun:
Tagged on a tiny fix for:

chrisvanrun:
Double-checked locally, seems like test `test_permission_request_notifications_flow_for_manual_review` is flaky.

app/grandchallenge/algorithms/templates/algorithms/algorithm_card_body.html
```diff
@@ -28,7 +28,7 @@
         <a class="badge badge-info above-stretched-link p-1 mr-1"
            href="{{ publication.identifier.url }}"
            title="{{ publication.title }}">
-            <i class="fas fa-file fa-fw"></i>&nbsp;Article
```
jmsmkn, line 31: OK - one `&nbsp;` is fine - this says "These things belong together on the same line, don't break them like a normal space". It was your use of two of them to force a certain spacing that was problematic. Please keep your PRs simple. They should do one thing, not touch 31 files to add a link to another page.

chrisvanrun: Roger. I try to balance 'while I have my hands in this code ...' with keeping PRs focused. I oversteped here.

jmsmkn (approved):

PR #3064, opened 2023-10-23
Add `image_pk_description_map` to serializers

pkcakeout:
Adds `image_pk_description_map` field to serialized display-set data.

Needed for https://github.com/DIAGNijmegen/rse-roadmap/issues/265 where we want to show individual case texts on each view item, so the catch-all `description` field does not suffice anymore.

The map is using the `image.pk` as a key and maps each image to the _rendered_ markdown (html). The `description` field needs to remain in place for now to service older workstations.

Other than that, I hope that `image_pk_description_map` is not too cryptic of a name...

### DIFF
--- a/app/grandchallenge/reader_studies/models.py
+++ b/app/grandchallenge/reader_studies/models.py
@@ -1,4 +1,5 @@
 import json
+from collections import OrderedDict
 
 from actstream.models import Follow
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


pkcakeout:
If that is your preferred method, we can certainly patch the current and prev-version of CIRRUS to do the interpolation on its own... sure! I will check how much work that would be...

pkcakeout:
Converted this to a Draft as well for better "visibility management".

jmsmkn:
Closing this as we decided on the route of CIV rendering in View Items instead. Users will enable this by adding the interface to the view content.

PR #3068, opened 2023-10-30
Make description required for public reader study

miriam-groeneveld:
I was looking at the reader study info page, and thought it would be nice to make is similar to the algorithm information page, so showing the logo and having the buttons on the right hand side, including a (new) Publish button:
image

Maybe that's too much, and I should simply add a requirement on the description field when checking the `public` checkbox?


jmsmkn:
> Maybe that's too much, and I should simply add a requirement on the description field when checking the public checkbox?

Yes, changing the forms clean method is all that is needed here.

Makefile
```diff
@@ -33,7 +33,7 @@ build_web_test:
 
 build_web_dist:
 	@docker pull $(GRAND_CHALLENGE_WEB_BASE_REPOSITORY_URI):$(PYTHON_VERSION)-$(POETRY_HASH) || { \
-		docker buildx build \
```
jmsmkn, line 36: This is still required on GitHub actions.

miriam-groeneveld:
Closed in favour of https://github.com/comic/grand-challenge.org/pull/3070

PR #3073, opened 2023-11-01
Fix display of string type job outputs

### LINKED ISSUE #3067: Display of Prostate Cancer Risk Score borked in Job list view
Displaying `ig` and `o` rather than `High` or `Low`.

amickan:
Closes #3067

### DIFF
--- a/app/grandchallenge/algorithms/templates/algorithms/job_detail.html
+++ b/app/grandchallenge/algorithms/templates/algorithms/job_detail.html
@@ -80,11 +80,7 @@
                         {% for object in json %}
                             <tr>
                                 <td>{{ object.interface.title }}</td>
-                                {% if object.interface.kind == 'STR' %}
-                                    <td>{{ object.value|slice:"1:-1" }} </td>
-                                {% else %}
-                                    <td>{{ object.value }}</td>
-                                {% endif %}
+                                <td>{{ object.value }}</td>
                             </tr>
                         {% endfor %}
                     </tbody>
--- a/app/grandchallenge/algorithms/templates/algorithms/job_list_row.html
+++ b/app/grandchallenge/algorithms/templates/algorithms/job_list_row.html
@@ -79,11 +79,7 @@
 {% for interface in outputs_list_display.JSON %}
     {% get_dict_values object.slug_to_output interface.slug as json %}
     {% if json %}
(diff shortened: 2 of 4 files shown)


app/grandchallenge/algorithms/templates/algorithms/job_detail.html
```diff
@@ -80,11 +80,7 @@ <h2>Algorithm Result</h2>
                         {% for object in json %}
                             <tr>
                                 <td>{{ object.interface.title }}</td>
-                                {% if object.interface.kind == 'STR' %}
-                                    <td>{{ object.value|slice:"1:-1" }} </td>
-                                {% else %}
-                                    <td>{{ object.value }}</td>
-                                {% endif %}
+                                <td>{{ object.value }}</td>
```
jmsmkn, line 83: It seems to me that this was handling the json `"` character. Could there be a difference in how these data are stored? Did you try this with interfaces with values stored in the db and in the object store? I know that the contents of the file `text-report.json` will be `"My fancy report"`.

amickan: Hm, I didn't know that we had string interfaces not stored in the DB, sorry! Strings stored as json files would currently not show at all (i.e. you would see `None`) because we print `civ.value`. Do we want to show the contents of those json files in the table though? We don't this for other json interfaces not stored in the db. They could potentially contain a lot of content, maybe a download button would be better then. We don't show the

jmsmkn: I agree, we shouldn't show them. I'm just a bit confused why the `"` was necessary in the first place now. But, if it's definitely not required then we can go ahead, let me know. We could of course do conditional removal (e.g. if the first and last char are `"` then remove them) with a template tag.

amickan: Yes, I was trying to understand that as well. No idea to be honest. I can confirm that there is no need for removing `"`, but we do need to add some code to make sure that interfaces not stored in the DB don't get added to the display. I will commit those additions in a bit.

jmsmkn: OK lets keep an eye on it after we deploy

PR #3075, opened 2023-11-01
Only show additional metrics button when they exist

### LINKED ISSUE #3027: Additional metrics buttons are displayed even when there are no additional metrics
Screenshot 2023-09-14 at 16 29 41

https://surgtoolloc23.grand-challenge.org/evaluation/final-testing-phase/leaderboard/

amickan:
Closes #3027

### DIFF
--- a/app/grandchallenge/evaluation/views/__init__.py
+++ b/app/grandchallenge/evaluation/views/__init__.py
@@ -740,6 +740,7 @@ class LeaderboardDetail(TeamContextMixin, PaginatedTableListView):
                 "now": now().isoformat(),
                 "limit": 1000,
                 "user_teams": self.user_teams,
+                "additional_metrics": True if self.phase.extra_results_columns else False,
                 "allow_metrics_toggling": len(self.phase.extra_results_columns)
                 > 0
                 or self.phase.scoring_method_choice is not self.phase.ABSOLUTE,
--- a/app/grandchallenge/evaluation/views/__init__.py
+++ b/app/grandchallenge/evaluation/views/__init__.py
@@ -740,9 +740,7 @@ class LeaderboardDetail(TeamContextMixin, PaginatedTableListView):
                 "now": now().isoformat(),
                 "limit": 1000,
                 "user_teams": self.user_teams,
-                "additional_metrics": True
-                if self.phase.extra_results_columns
-                else False,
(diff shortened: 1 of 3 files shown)


app/grandchallenge/evaluation/views/__init__.py
```diff
@@ -740,6 +740,9 @@ def get_context_data(self, *args, **kwargs):
                 "now": now().isoformat(),
                 "limit": 1000,
                 "user_teams": self.user_teams,
+                "additional_metrics": True
+                if self.phase.extra_results_columns
+                else False,
```
jmsmkn, line 745:
```suggestion
                "additional_metrics": bool(self.phase.extra_results_columns),
```

app/grandchallenge/evaluation/views/__init__.py
```diff
@@ -740,6 +740,7 @@ def get_context_data(self, *args, **kwargs):
                 "now": now().isoformat(),
                 "limit": 1000,
                 "user_teams": self.user_teams,
+                "additional_metrics": bool(self.phase.extra_results_columns),
```
jmsmkn, line 743: How does `additional_metrics` differ from `allow_metrics_toggling`? I'm not sure we actually need the new variable and just the logic in `allow_metrics_toggling` needs to be fixed.

amickan: Good point. I just saw the ranking was taken into consideration and assumed that it needed to stay in place. Looking closer at it now I don't see what the ranking method has to do with toggling - either there are extra results to show or hide, or there are not, right? So indeed, I think, we can update `allow_metrics_toggling` to only check whether there are extra results.

amickan: The ranking methods `mean` and `median` only make sense if there are extra results columns. From what I can tell we don't check that, but I guess that's not really an issue since it doesn't matter for the calculations.

amickan: Ah I understand now why we check the scoring method - the comparison does not work though (i.e. evaluates to True when it shouldn't). I'll probably only have the time to take a calm look on Monday to post a fix

amickan: Changing `is not` to `!=` does the trick

jmsmkn (approved):

PR #3085, opened 2023-11-08
Add archive item check to submission form

### LINKED ISSUE #3076: Reject submission creation when there are no valid archive items for an algorithm submission phase
Only needs an update of the forms clean method, we still should show the form to help challenge admins with their phase set up.

amickan:
Closes #3076 

I realized that we don't have tests for any of the checks on the SubmissionForm. Let me know if you'd like me to add some.

### DIFF
--- a/app/grandchallenge/evaluation/forms.py
+++ b/app/grandchallenge/evaluation/forms.py
@@ -369,6 +369,11 @@ class SubmissionForm(
     def clean_algorithm_image(self):
         algorithm_image = self.cleaned_data["algorithm_image"]
 
+        if self._phase.count_valid_archive_items == 0:
+            raise ValidationError(
+                "There are no valid archive items in the archive linked to this phase."
+            )
+
         if Submission.objects.filter(
             algorithm_image__image_sha256=algorithm_image.image_sha256,
             phase=self._phase,


app/grandchallenge/evaluation/forms.py
```diff
@@ -369,6 +369,11 @@ def __init__(self, *args, user, phase: Phase, **kwargs):  # noqa: C901
     def clean_algorithm_image(self):
         algorithm_image = self.cleaned_data["algorithm_image"]
 
+        if self._phase.count_valid_archive_items == 0:
```
jmsmkn, line 372: The check should be on the Phase, not the algorithm image, right? The error message is also too technical, maybe something simple first like - this phase is not ready for submissions yet, then the rest.

And yes, tests would be good.

amickan: Yes, logically this should be done in the phase clean method, but since we want to return an error to the user and we don't expose the phase field on the form, I added it to the algorithm image field. I can add the error to the non-field errors on the form, so they show up on top, if you prefer, but it's quite crowded there already with all warnings etc, so wasn't sure.

jmsmkn: Yes I think hoisting it is fine. Maybe this should anyway be added to all those warnings as no one will be able to submit with this new check too. It's really hard to communicate all of the tasks the admins must do to get everything set up.

app/grandchallenge/evaluation/forms.py
```diff
@@ -366,6 +366,15 @@ def __init__(self, *args, user, phase: Phase, **kwargs):  # noqa: C901
                 "uploads.change_userupload",
             ).filter(status=UserUpload.StatusChoices.COMPLETED)
 
+    def clean_phase(self):
+        phase = self.cleaned_data["phase"]
+        if phase.count_valid_archive_items == 0:
```
jmsmkn, line 371: Missing an "and phase.submission_type == SubmissionType.Algorithm" ?

app/grandchallenge/evaluation/forms.py
```diff
@@ -366,6 +366,15 @@ def __init__(self, *args, user, phase: Phase, **kwargs):  # noqa: C901
...
+            self.errors["__all__"] = [
```
jmsmkn, line 372: This will wipe out the existing content, I think that `add_error` with `field=None` should be used instead (https://docs.djangoproject.com/en/4.2/ref/forms/api/#django.forms.Form.add_error). If so, could you check for other usages of this pattern in the codebase and correct them too?

amickan: Found a few, none critical (as in we didn't override important error messages in those places), but updated the usage anyway.

jmsmkn: Great, and it definitely works as expected with ErrorList? Not too sure how good our test coverage is on that.

amickan: Yes, I played around with it in the shell. ErrorList is really just a list of error strings and the `add_error` method adds all of them.

Issue #3087, opened 2023-11-10
Display Set can have CIV without value


miriam-groeneveld:
**Describe the bug**

**To Reproduce**
1. Go to https://grand-challenge.org/reader-studies/cirruscoreweb-release-default-answers/ 
2. Click on `Cases`, then `View/Delete Cases` 
3. Expand the first display set. Note that there is no value for the interface `regions-of-interest`:
![image](https://github.com/comic/grand-challenge.org/assets/51329283/9dd6d49b-63ed-4d79-856a-72078c59ffe3)
4. Click on `Edit Case`
5. Dropdown for `regions-of-interest` shows empty entries
![image](https://github.com/comic/grand-challenge.org/assets/51329283/b800a054-9c7c-49e5-892a-8f31309c346d)
6. Selecting the `---` option and save has no effect (should be a differnt bug maybe?)
7. Click on `View Display Set`
8. Cirrus will show error:
![image](https://github.com/comic/grand-challenge.org/assets/51329283/f227762c-140a-4cfa-b322-6f12002f227f)

The empty dropdown list issue can be reproduced by editing the Display Set and selecting an existing json file for the text-report interface, see video:

https://github.com/comic/grand-challenge.org/assets/51329283/a629f56c-83f0-44fd-925a-46a05f28f986

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

amickan:
I found out why the regions of interest file gets deleted when you add a text report file, but still not entirely sure how the empty CIVs got created in the first place. Could it be that the regions-of-interest interface used to be stored in the DB? There is a second issue in the DS update view where updating a json interface stored in the DB results in empty CIVs being created, so that could explain it.

amickan:
Knowing of this issue, there are probably more empty CIVs in the database (specifically of json interfaces stored in the db or those that used to be stored in the db). Could you check if that's true, James?

jmsmkn:
> Could it be that the regions-of-interest interface used to be stored in the DB?

No, it is not possible to change the storage type directly. When we do we run a management command. These were definitely created with the storage set correctly.

> Knowing of this issue, there are probably more empty CIVs in the database (specifically of json interfaces stored in the db or those that used to be stored in the db). Could you check if that's true, James?

There are, but this is not coming from changing of storage type and happens regardless of how the interface is defined. There are even images that are empty where the storage type is not relevant.

[...]

PR #3088, opened 2023-11-13
Fix display set value updating 

amickan:
Fixes the following issues:
- when partially updating the values of a DS, (json) file type interface values used to get deleted when they were unchanged
- deleting the value of a json interface would result in the creation of an empty CIV 
- for values of json interfaces, a new CIV would be created regardless of whether they passed validation, again resulting in empty CIVs when validation failed

Part of #3087

### DIFF
--- a/app/grandchallenge/reader_studies/views.py
+++ b/app/grandchallenge/reader_studies/views.py
@@ -1411,13 +1411,12 @@ class DisplaySetUpdate(
         for ci_slug, new_value in form.cleaned_data.items():
             if ci_slug == "order":
                 continue
-            with transaction.atomic():
-                instance, assigned_civs = self.create_civ_for_value(
-                    instance=instance,
-                    ci_slug=ci_slug,
-                    new_value=new_value,
-                    assigned_civs=assigned_civs,
-                )
+            instance, assigned_civs = self.create_civ_for_value(
+                instance=instance,
+                ci_slug=ci_slug,
+                new_value=new_value,
+                assigned_civs=assigned_civs,
+            )
         instance.values.remove(*assigned_civs)
 
         if (
--- a/app/grandchallenge/reader_studies/views.py
+++ b/app/grandchallenge/reader_studies/views.py
@@ -1411,12 +1411,13 @@ class DisplaySetUpdate(
         for ci_slug, new_value in form.cleaned_data.items():
             if ci_slug == "order":
                 continue
-            instance, assigned_civs = self.create_civ_for_value(
-                instance=instance,
-                ci_slug=ci_slug,
-                new_value=new_value,
-                assigned_civs=assigned_civs,
-            )
+            with transaction.atomic():
+                instance, assigned_civs = self.create_civ_for_value(
+                    instance=instance,
+                    ci_slug=ci_slug,
+                    new_value=new_value,
+                    assigned_civs=assigned_civs,
+                )
         instance.values.remove(*assigned_civs)
 
         if (
@@ -1437,10 +1438,11 @@ class DisplaySetUpdate(
             if current_value and current_value.value != new_value:
                 assigned_civs.append(current_value)
             if new_value and current_value.value != new_value:
-                val = ComponentInterfaceValue.objects.create(
+                civ = ComponentInterfaceValue.objects.create(
                     interface=ci, value=new_value
                 )
-                instance.values.add(val)
+                civ.full_clean()
+                instance.values.add(civ)
         elif isinstance(new_value, Image):
             if not current_value or (
                 current_value and current_value.image != new_value
@@ -1467,7 +1469,7 @@ class DisplaySetUpdate(
                 )
             )
         else:
-            if current_value != new_value:
+            if current_value and current_value != new_value:
                 assigned_civs.append(current_value)
                 if new_value:
                     # If there is already a value for the provided civ's interface in
--- a/app/grandchallenge/reader_studies/views.py
+++ b/app/grandchallenge/reader_studies/views.py
@@ -1409,7 +1409,7 @@ class DisplaySetUpdate(
         instance = self.get_object()
         assigned_civs = []
         for ci_slug, new_value in form.cleaned_data.items():
-            if ci_slug == "order" or new_value is None:
+            if ci_slug == "order":
                 continue
             instance, assigned_civs = self.create_civ_for_value(
                 instance=instance,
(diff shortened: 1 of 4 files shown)


app/grandchallenge/reader_studies/views.py
```diff
@@ -1409,14 +1409,15 @@ def form_valid(self, form):
...
-            if ci_slug == "order" or new_value is None:
+            if ci_slug == "order":
                 continue
-            instance, assigned_civs = self.create_civ_for_value(
-                instance=instance,
-                ci_slug=ci_slug,
-                new_value=new_value,
-                assigned_civs=assigned_civs,
-            )
+            with transaction.atomic():
```
jmsmkn, line 1414: Why is this needed? We should be using atomic requests.

amickan: Ok, brainfog on Monday evening. It turns out to be unnecessary here, just tested things again. I think what happened is that I misplaced this on Monday evening, because it does appear to be necessary in the `DisplaySetInterfacesCreate` view, but I don't understand why. We call `create_instance` there and I see in backends/base.py that you also wrapped the output interface creation in an atomic block, explaining there that this is necessary because `create_instance` needs to create a CIV to store the files. I don't follow this explanation though. We create CIVs in a lot of places where the extra atomic block is not necessary, so why with this specific method then?

jmsmkn: The only thing I can think of is that those are used in non-atomic Celery tasks, in which case the atomicity should be moved into the task itself (sometimes we don't want a task to be completely atomic, e.g. with image imports where you want to inform the user that something is being consumed first then a large amount of work happens). You have to be really careful with refactoring those. I just pointed it out here as there shouldn't be a reason to introduce any new ones outside of a Celery task.

There was a time when we were not using atomic commits due to an error on my part (put the setting in the wrong place for years!) and there might be remnants of that, but it is not a cause of the issue you are trying to fix.

amickan: While trying to write a test to illustrate the issue in the `DisplaySetInterfacesCreate`, I realized that the atomic block is only necessary if the request is an HTMX request. I couldn't reproduce the issue in the test until I added the Htmx specific Http header to the request. I have updated the test to illustrate the issue.

jmsmkn: Extremely strange. Good luck debugging that one!

amickan: False alarm with the Htmx stuff - embarassing, but phew... 

I still can't reproduce the CIV creation in the test though. In the test the form immediately evaluates to invalid (as it should) whereas in the UI if I click my way through adding a new interface with an incorrect value, the form first evaluates to true (and hence hits the `form_valid` method on the view) and only gets marked as invalid after another round of validation, triggered by the creation of the CIV. It's strange that the form initially evaluates to valid, it shouldn't in my eyes (see snippet that Mike added): 

https://github.com/comic/grand-challenge.org/blob/ea980be59a473e8cf225a458d4db6b9caa2a5243/app/grandchallenge/reader_studies/views.py#L1644-L1653

**In this snippet, since we catch the validation error, the transaction isn't rolled back and so we end up with a CIV without a value.** 

Does this make any sense? I have looked at this for a bit too long in a row today, and will do something else for the rest of today to regain some clarity.

jmsmkn: OK that makes more sense. Then maybe you need to do things the other way around - do not do `CIV.create(...)`, instead create the CIV directly then do the validation and if it passes save it.

jmsmkn: And yes, the validation should definitely be completed before form_valid is hit! It's no good adding errors to the form afterwards.

amickan: Figured it out, I think! The interface field was not required, although it should be on this form. So empty values passed validation on the form's clean method, but failed validation on CIV creation. I suppose that Mike added the try... except statement in the form_valid method to catch this scenario and return the error to the user. If we make the field required, we no longer need that though. All other errors will then be caught in the form's clean method, or for the more complex cases in the asynchronous tasks (where we use atomic blocks everywhere to ensure that the transaction gets reverted if validation fails).

app/grandchallenge/reader_studies/views.py
```diff
@@ -1434,12 +1434,14 @@ def create_civ_for_value(
...
         current_value = instance.values.filter(interface=ci).first()
         if ci.is_json_kind and not ci.requires_file:
-            if current_value:
+            if current_value and current_value.value != new_value:
                 assigned_civs.append(current_value)
-            val = ComponentInterfaceValue.objects.create(
-                interface=ci, value=new_value
-            )
-            instance.values.add(val)
+            if new_value and current_value.value != new_value:
```
jmsmkn, line 1439: Can these three if statements be tidied up? Quite a lot going on there and some duplication.

app/grandchallenge/reader_studies/views.py
```diff
@@ -1429,28 +1429,33 @@ def form_valid(self, form):
...
-                    interface=ci, image=new_value
-                )
-                if created:
-                    civ.full_clean()
-                instance.values.add(civ)
+            civ.full_clean()
+            instance.values.add(civ)
+        elif not new_value:
+            # if the new value is None, remove the old CIV from the display set
+            assigned_civs.append(current_civ)
```
jmsmkn, line 1444: Sorry, trying to understand this code still. I remember being lots of issues with it which is why there are a lot of existing comments. Why this construction where values are added one by one, but collected and then removed in bulk by passing round a list? Surely it's better just to remove them directly with `instance.values.remove(current_civ)` - then you could remove all instances of `assigned_civs`, right? Or am I missing something?

jmsmkn: I also don't see why `instance` needs to be returned as it is not being modified, right?

app/grandchallenge/reader_studies/views.py
```diff
@@ -1476,8 +1486,28 @@ def create_civ_for_value(
...
+                instance, ci, current_civ, new_value, assigned_civs
+            )
+        elif ci.is_image_kind:
+            return self.create_civ_for_image(
+                instance, ci, current_civ, new_value, assigned_civs
+            )
+        elif ci.requires_file:
+            return self.create_civ_for_file(
+                instance, ci, current_civ, new_value, assigned_civs
+            )
```
jmsmkn, line 1509: Should we catch anything in a final `else` block here?

amickan: Good idea, yes. At this point, the ifs cover all cases, but that could of course change in the future.

app/grandchallenge/reader_studies/views.py
amickan: You're right. Removing them directly works and is definitely better.

amickan: And passing the instance around also shouldn't be necessary. I'll double check, of course!

amickan:
Unsure how much refactoring should be done in one PR, but the `DisplaySetUpdateView` and the `DisplaySetInterfacesCreate` views both have an update method that does essentially the same thing. We could move that method to the display set model instead and use the same method in both places, I think. Maybe I'm overseeing issues here, so will need to try this out, but I can also do that in a seperate PR. If it works, it would make future debugging easier I think.

jmsmkn:
Moving these to model methods is a great idea! Then you can loose a bunch of the `get_object` calls. Maybe do that first for these 4 methods here, then follow up with a PR for consolidation.

jmsmkn (approved):
Nice cleanup!

PR #3092, opened 2023-11-16
Remove inline event handler for notification bulk update

### LINKED ISSUE #3091: Broken "Select all" button on notifcations page
**Describe the bug**
The "select all" button on the notifications page does not work anymore. 

**To Reproduce**
1. Go to https://grand-challenge.org/notifications/
2. Click on 'Select all'
3. Nothing happens

**Expected behavior**
All notifications on the page should get selected.

amickan:
With the introduction of a CSP, the "Select all" button, for which we used an inline event handler, no longer worked. This fixes that.

Closes #3091

### DIFF
--- a/app/grandchallenge/notifications/static/js/notifications/bulk_delete_update.js
+++ b/app/grandchallenge/notifications/static/js/notifications/bulk_delete_update.js
@@ -1,6 +1,8 @@
-window.toggleCheckboxes = function toggleCheckboxes(source) {
+document.getElementById("SelectCheckboxes").addEventListener("click", toggleCheckboxes);
+
+function toggleCheckboxes(source) {
     $('input[name="checkbox"]').each(function () {
-        this.checked = source.checked;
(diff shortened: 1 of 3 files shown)


app/grandchallenge/notifications/static/js/notifications/bulk_delete_update.js
```diff
@@ -1,6 +1,8 @@
-window.toggleCheckboxes = function toggleCheckboxes(source) {
+document.getElementById("SelectCheckboxes").addEventListener("click", toggleCheckboxes);
+
+function toggleCheckboxes(source) {
     $('input[name="checkbox"]').each(function () {
-        this.checked = source.checked;
+        this.checked = source.srcElement.checked;
```
jmsmkn, line 5: I think that `srcElement` is deprecated for `target` https://developer.mozilla.org/en-US/docs/Web/API/Event/srcElement

jmsmkn (approved):

PR #3093, opened 2023-11-19
Add Direct Messages

jmsmkn:
This PR adds the Direct Messages application. It introduces three main new models:

- `Conversation`: Many users are `participants` of one conversation. Many permissions are added to each conversation - view, create_conversation_direct_message, mark_conversation_read, mark_conversation_message_as_spam.
- `DirectMessage`: Has a foreign key to a `conversation`, `sender` and which users have not read this message. The sender can delete the message and the conversation participants can mark the message as spam. If a direct message is created the other participants in the group have the message marked as unread. If a message is deleted or marked as spam then all of the unread markers are deleted.
- `Mute`: tracks which users have muted which other users. If a direct message is created and the receiver has muted the sender the message is not marked as unread by the receiver.

The views use HTMX. As a conversation is loaded all of the messages are marked as read as a callback. For now only challenge admins can message their participants.

See https://github.com/DIAGNijmegen/rse-roadmap/issues/255


app/grandchallenge/direct_messages/models.py
```diff
@@ -0,0 +1,199 @@
...
+
+        if self.is_deleted or self.is_reported_as_spam:
+            DirectMessageUnreadBy.objects.filter(direct_message=self).delete()
+
+        if adding:
+            self.assign_permissions()
+
+    def delete(self, *args, **kwargs):
+        self.is_deleted = True
+        self.save()
```
amickan, line 66: Why are we only removing the many to many relationship and keep the message itself?

jmsmkn: The unread markers are removed as there is no message to read. The message is kept but the content not displayed so that users know that a message was sent but deleted. We keep the content so we can check for abuse and also the user can un-do the deletion.

amickan: Ok, makes sense!

app/grandchallenge/direct_messages/forms.py
```diff
@@ -0,0 +1,199 @@
...
+            total_participants_count=Count("participants", distinct=True),
+            relevant_participants_count=Count(
+                "participants",
+                filter=Q(participants__in=participants),
+                distinct=True,
+            ),
+        ).filter(
+            total_participants_count=len(participants),
+            relevant_participants_count=len(participants),
+        ):
```
jmsmkn, line 55: This can be moved to the QuerySet

app/grandchallenge/direct_messages/views.py
```diff
@@ -0,0 +1,304 @@
...
+                queryset=DirectMessage.objects.order_by("created").annotate(
+                    unread_by_user=Case(
+                        When(unread_by=self.request.user, then=Value(True)),
+                        default=Value(False),
+                        output_field=BooleanField(),
+                    )
+                ),
+            ),
+            "direct_messages__sender__user_profile",
+        )
```
jmsmkn, line 123: This can be moved to the QuerySet

scripts/development_fixtures.py
```diff
@@ -111,12 +116,15 @@ def _create_flatpages():
...
     users = {}
+    fake = Faker()
 
     for username in usernames:
         user = get_user_model().objects.create(
             username=username,
             email=f"{username}@example.com",
             is_active=True,
+            first_name=fake.first_name(),
+            last_name=fake.last_name(),
```
jmsmkn, line 127: Also fill out their department, institution and country.

app/grandchallenge/core/templates/grandchallenge/partials/userlinks.html
```diff
@@ -42,19 +42,31 @@ <h6 class="dropdown-header">{{ request.site.name }} Support</h6>
     </div>
 </li>
 {% if user.is_authenticated %}
-    {% if ACTSTREAM_ENABLE %}
     <li class="nav-item mr-auto">
-        {% if user.user_profile.has_unread_notifications %}
+        {% if user.directmessageunreadby_set.exists %}
```
amickan, line 46: Since this feature is only available to challenge admins (and when messaged to participants), shouldn't we hide the entire icon if there are no messages?

jmsmkn: Yes, thought about that but then it would be an extra lookup on every single page. We don't have an abstract user so I couldn't easily work out how to attach a `prefetch_related` to `request.user`. So, knowing that we have to make one extra call for this I went for the quickest one that also avoids any joins.

In the longer term we could move both the messages and notifications icon and alert to a HTMX view instead, then we don't slow down each page load.

app/grandchallenge/direct_messages/templates/direct_messages/conversation_detail.html
```diff
@@ -0,0 +1,156 @@
...
+        {% endifchanged %}
+
+        <div class="d-flex w-100 justify-content-start">
+            {% if direct_message.unread_by_user %}
+                <span class="text-danger"><i class="fa fa-xs fa-circle"></i></span>&nbsp;
+            {% endif %}
+
+            <div>
+                {% include "direct_messages/partials/directmessage_detail.html" with preview=False %}
+            </div>
```
amickan, line 101: As I was trying things out locally, I found it a bit difficult to tell my own messages apart from those of other people in the conversation. We could right align the users messages and left align the rest (see whatsapp and other message apps). Idk, just a thought.

app/grandchallenge/direct_messages/views.py
```diff
@@ -0,0 +1,304 @@
...
+        return super().form_valid(form)
+
+
+class DirectMessageCreate(
+    LoginRequiredMixin, ObjectPermissionRequiredMixin, CreateView
+):
+    permission_required = "direct_messages.create_conversation_direct_message"
+    raise_exception = True
+    model = DirectMessage
+    form_class = DirectMessageForm
```
amickan, line 210: Muted users can still create messages, and won't know that they have been muted by someone. I guess that's on purpose? Is that to protect the user who does the blocking? I feel like at least some sort of indication that the messages aren't making it through to the other user would be nice for the blocked user.

app/grandchallenge/core/templates/grandchallenge/partials/userlinks.html
amickan: Fair enough! I hadn't thought this far yet, but I agree that it's definitely not worth the extra call and also not really worth adding a custom manager for the user model to do the necessary prefetching.

app/grandchallenge/direct_messages/templates/direct_messages/conversation_detail.html
jmsmkn: Indeed, we can change it in the future, I think this is OK for the initial version.

app/grandchallenge/direct_messages/views.py
```diff
@@ -0,0 +1,304 @@
...
+        return form_kwargs
+
+
+class DirectMessageReportSpam(
+    LoginRequiredMixin, ObjectPermissionRequiredMixin, UpdateView
+):
+    permission_required = "direct_messages.mark_conversation_message_as_spam"
+    raise_exception = True
+    model = DirectMessage
+    form_class = DirectMessageReportSpamForm
```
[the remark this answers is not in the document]
jmsmkn: Yes, we can monitor this in the admin and potentially take action. Later we can maybe add reports for us to be notified if a particular user is sending a bunch of spam, or is blocked by multiple people.

amickan (approved):
Impressive! I worked my way through the code and tried to break it, but it all appears to be working as intended :-) I think some more tests would be good though. 

I really like the look and feel of the UI for the messages as well.

PR #3097, opened 2023-11-22
Refactor DisplaySet views

amickan:
All DisplaySet create and update views now use the CIV creation method on the model.


jmsmkn (approved):
Nice cleanup!

PR #3099, opened 2023-11-23
My challlenge requests list

### LINKED ISSUE #3060: Unclear how to view own challenge request
As far as I can tell it is currently not possible to review your own challenge request.

## Current
Current workflow, as a potential challenge organizer:
- Click 'Host your own Challenge'
- Answer all fields
- Hit Save
- Get dropped at the challenge list page

## Suggestion

Twofold: 
1. add a 'my challenge requests' button that allows a user to see all created requests by the user. Make sure to include the state in the overview.
2. add a configurable URL redirect after the save, and configure it in production to the 'https://grand-challenge.org/documentation/create-your-own-challenge/#from-challenge-request-to-challenge-launch'. Update the docs slightly to make it a nicer redirect land.
3. Update the sent email to include a link to the created request.

amickan:
This updates the challenge requests list view so that both reviewers and users can access it and view the requests they have access to. To do this, this adds object permissions for challenge requests. All related views are updated with new permissions and a reviewer group is added. 

Once this is merged, we will need to add the 10% team members to the new reviewer group, as well as Chris, Miriam, me and James. 

Closes #3060


app/config/settings.py
```diff
@@ -88,6 +88,7 @@
 USER_LOGIN_TIMEOUT_DAYS = 14
 REGISTERED_USERS_GROUP_NAME = "__registered_users_group__"
 REGISTERED_AND_ANON_USERS_GROUP_NAME = "__registered_and_anonymous_users__"
+CHALLENGES_REVIEWERS_GROUP_NAME = "challengerequest_reviewers"
```
jmsmkn, line 91: For consistency

```suggestion
CHALLENGES_REVIEWERS_GROUP_NAME = "__challengerequest_reviewers__"
```

app/grandchallenge/challenges/models.py
```diff
@@ -957,9 +957,18 @@ def save(self, *args, **kwargs):
         super().save(*args, **kwargs)
 
         if adding:
+            self.assign_permissions()
             send_challenge_requested_email_to_reviewers(self)
             send_challenge_requested_email_to_requester(self)
 
+    def assign_permissions(self):
+        assign_perm("view_challengerequest", self.creator, self)
+        reviewers, _ = Group.objects.get_or_create(
```
jmsmkn, line 966: I think only get the reviewers here. You can create the group in a post migration step in `apps.py`, see the implementation of the registered users group.

app/grandchallenge/challenges/templates/challenges/challenge_list.html
```diff
@@ -17,7 +18,7 @@
             </a>
         </div>
         <div class="col-xl-4 col-lg-5 col-md-6 mb-3 text-md-right">
-            {% if perms.challenges.change_challengerequest %}
+            {% if request.user|is_challenge_reviewer %}
```
jmsmkn, line 21: The object based permissions should be used here, see `get_perms_for_user` for other objects.

jmsmkn: In fact, you use it in the next file, why the difference?

app/grandchallenge/challenges/emails.py
```diff
@@ -26,20 +25,21 @@ def send_challenge_requested_email_to_reviewers(challengerequest):
         site_name=site.name,
         site_domain=site.domain,
     )
-    reviewers = get_user_model().objects.filter(
-        Q(groups__permissions__codename="change_challengerequest")
-        | Q(user_permissions__codename="change_challengerequest")
+    reviewers, _ = Group.objects.get_or_create(
```
jmsmkn, line 28: Should only be a get here.

app/grandchallenge/challenges/urls.py
```diff
@@ -10,13 +10,19 @@
...
     UsersChallengeList,
+    UsersChallengeRequestList,
 )
 
 app_name = "challenges"
 
 urlpatterns = [
     path("", ChallengeList.as_view(), name="list"),
     path("my-challenges/", UsersChallengeList.as_view(), name="users-list"),
+    path(
```
jmsmkn, line 21: Why do we need two views? Why not one view and then filter and separate any actions based on the users permission?

app/grandchallenge/core/templatetags/permissions.py
```diff
@@ -0,0 +1,16 @@
+from django import template
+from django.conf import settings
+
+register = template.Library()
+
+
+@register.filter
+def is_challenge_reviewer(user):
```
jmsmkn, line 8: Placeholder as I don't think this file is necessary.

app/grandchallenge/challenges/templates/challenges/challenge_list.html
jmsmkn: Oh! I see, again, add and use the global perm for this.

app/grandchallenge/pages/templates/pages/challenge_statistics.html
```diff
@@ -51,7 +52,7 @@ <h4>Submissions</h4>
         {% endfor %}
     </dl>
 
-    {% if perms.challenges.view_challengerequest %}
+    {% if request.user|is_challenge_reviewer %}
```
jmsmkn, line 55: Use global permissions for this.

app/grandchallenge/serving/views.py
```diff
@@ -165,7 +165,9 @@ def serve_structured_challenge_submission_form(
     except ChallengeRequest.DoesNotExist:
         raise Http404("Challenge request not found.")
 
-    if request.user.has_perm("challenges.view_challengerequest"):
+    if request.user.groups.filter(
+        name=settings.CHALLENGES_REVIEWERS_GROUP_NAME
+    ).exists():
```
jmsmkn, line 170: Should use object based permissions here.

app/grandchallenge/challenges/templates/challenges/challenge_list.html
amickan: Ok, I wasn't sure what would be better, checking for global permissions or checking for group membership. Why is it better to use global permissions here (apart from the fact that it makes the custom template tag unnecessary here)? And shouldn't we try and be consistent and use global permissions everywhere else then (also in the emails for example)?

app/grandchallenge/challenges/emails.py
amickan: See comment above, shouldn't we better check for global permissions here as well for consistency, rather than group membership?

app/grandchallenge/challenges/urls.py
amickan: Good point actually. I thought they were too different, but that's actually not true! I'll combine them then

app/grandchallenge/challenges/emails.py
amickan: Ah I think the look-up is more efficient this way actually.

app/grandchallenge/serving/views.py
amickan: I'm wondering which permission to use, view or change challenge request? So far we only allowed reviewers to download the submission form, but I don't think that makes any sense. I would say that the creator should be able to download it as well, and hence use the view permission. Would you agree?

app/grandchallenge/challenges/templates/challenges/challenge_list.html
jmsmkn: Yes it is better to use global permissions otherwise you need to duplicate things into the codebase about who has access, this is the shortcut for that so that we can set the authorisation in one place. Then it will also work if you want to give another group permission.

And yes, be consistent about it.

app/grandchallenge/challenges/emails.py
```diff
@@ -40,6 +40,7 @@ def send_challenge_requested_email_to_reviewers(challengerequest):
 
 def send_challenge_requested_email_to_requester(challengerequest):
     site = Site.objects.get_current()
+    link = reverse("challenges:requests-list")
```
jmsmkn, line 43: I can't make a comment on this line but on Line 29 you could use `User.objects.with_perm('change_challengerequest')`, bit cleaner, but check this.

app/grandchallenge/challenges/templates/challenges/challenge_list.html
```diff
@@ -17,7 +17,7 @@
             </a>
         </div>
         <div class="col-xl-4 col-lg-5 col-md-6 mb-3 text-md-right">
-            {% if perms.challenges.change_challengerequest %}
+            {% if perms.challenges.view_challengerequest %}
```
jmsmkn, line 20: This should probably still be change right? Or we could remove the button altogether as its now in the nav.

amickan: well both the global view and the global change permissions are only given to reviewers, so doesn't really matter.

app/grandchallenge/challenges/emails.py
amickan: To do that we need to specify the authentication backend... That would be "django.contrib.auth.backends.ModelBackend", correct? And also, I thought it's better not to refer to User directly, but isntead use get_user_model()?

app/grandchallenge/challenges/views.py
```diff
@@ -134,23 +137,25 @@ def get_form_kwargs(self):
...
-    permission_required = "challenges.view_challengerequest"
+    permission_required = "view_challengerequest"
+    raise_exception = True
+    login_url = reverse_lazy("account_login")
 
 
 class ChallengeRequestDetail(
     LoginRequiredMixin,
-    PermissionRequiredMixin,
+    ObjectPermissionRequiredMixin,
```
jmsmkn, line 152: Please make sure that there are tests for all of these changes - that one request creator cannot access those of another, that the list view is filtered etc.

app/grandchallenge/challenges/emails.py
jmsmkn: Yes, took the example from the docs. Strange that you have to specify the backend, whatever, just leave it.

app/grandchallenge/challenges/templates/challenges/challenge_list.html
jmsmkn: I think get rid of the button as it's no longer needed.

jmsmkn (approved):

PR #3100, opened 2023-11-24
Add a YouTube tag for MarkDown fields

chrisvanrun:
This PR adds a markdown notation `[youtube ID]` to embed youtube into markdown on pages. These can be created by users, require `<iframe>`s and are shown to users.

This markdown:
```Markdown
# A Title
This is a demo reader study.

## Instructions
Just look at how wonderful this loads. Note that it's set to load things on a 16:9, hence the best result is gotten by recording on that aspect ratio.

[youtube QCYYhkTlnhQ]
```
image

## Custom Tags
A post-processing step to parse new custom tags is now made. This is done after (!) the bleaching process and operates on clean HTML. As such, a custom tag should produce clean HTML itself. 

It should be pretty straight forward to extend this with new tags.


app/grandchallenge/core/templatetags/bleach.py
```diff
@@ -6,18 +8,45 @@
...
-def clean(html: str):
+def clean(html: str, allow_iframes=False):
     """Clean the html with bleach."""
+
+    allowed_tags = settings.BLEACH_ALLOWED_TAGS
+    allowed_attributes = settings.BLEACH_ALLOWED_ATTRIBUTES
+
+    if allow_iframes:
+        allowed_tags = [*allowed_tags, "iframe"]
+        allowed_attributes.update({"iframe": allowed_iframe})
```
jmsmkn, line 44: This updates the setting as it's not a (deep)copy, right?

app/grandchallenge/core/templatetags/bleach.py
```diff
@@ -6,18 +8,45 @@
...
-from grandchallenge.core.utils.markdown import LinkBlankTargetExtension
+from grandchallenge.core.utils.markdown import (
+    EmbedYoutubeExtension,
+    LinkBlankTargetExtension,
+)
 
 register = template.Library()
 
 
+def allowed_iframe(tag, name, value):
```
jmsmkn, line 19: Interesting construction. Is this a documented feature of bleach? This library is deprecated, so we want to move away from it. I don't want to add more dependency on features of it as it would make that migration harder. Is it possible to achieve the same with `nh3` https://nh3.readthedocs.io/en/latest/ ?

app/grandchallenge/core/templatetags/bleach.py
```diff
@@ -38,12 +67,24 @@ def md2email_html(markdown: str | None):
     )
 
 
+@register.filter
+def md2page_html(markdown: str | None):
+    """Converts markdown to clean html intended for showing as a page"""
```
jmsmkn, line 72: I don't really understand why we need a new filter here, in what situations would I use each of:

- `md2html`: this was for all rendering on the site, apart from emails
- `md2html_email`: this was for rendering for emails
- `md2html_page`: don't really get it, `Page` objects do not use markdown and this seems to be used everywhere

app/grandchallenge/core/templatetags/bleach.py
chrisvanrun: It is a [documented ](https://bleach.readthedocs.io/en/latest/clean.html#using-functions)feature of bleach. I didn't know it was deprecated.

Looking at nh3. It should be nigh equivalent to using `attribute_filter` callables. I don't think it can easily integrate with bleach at the same time, might as well replace it completely then. Shall I bloat the PR with that or run the current bleach implementation and replace it in a new PR?

app/grandchallenge/core/templatetags/bleach.py
chrisvanrun: Oh yes! Good find. Almost made the same error here as I fixed last time =P

app/grandchallenge/core/templatetags/bleach.py
```diff
@@ -6,18 +8,45 @@
...
 register = template.Library()
 
 
+def allowed_iframe(tag, name, value):
+    """Returns true if the iframe tag is allowed, false otherwise"""
+    if tag != "iframe":
+        return False
+    if name in settings.BLEACH_ALLOWED_FRAME_ATTRIBUTES:
+        return True
+    if name == "src":
```
jmsmkn, line 25: Is there a way to only allow iframes with `sandbox`?

app/grandchallenge/core/utils/markdown.py
```diff
@@ -41,3 +44,72 @@ def run(self, root):
...
+        )
+
+
+class EmbedYouTubeInLineProcessor(InlineProcessor):
+    def handleMatch(self, m, data):  # noqa: N802
+        youtube_id = m.group(1)
+
+        el = ElementTree.Element("iframe")
+
+        el.set(
```
jmsmkn, line 68: Don't love this way of constructing a HTML fragment. Could you use a Django partial template instead and then load it as an element? Find those much easier to follow.

app/grandchallenge/reader_studies/models.py
```diff
@@ -484,7 +484,7 @@ def remove_reader(self, user):
     @property
     def help_text(self) -> str:
         """The cleaned help text from the markdown sources"""
-        return md2html(self.help_text_markdown, link_blank_target=True)
+        return md2page_html(self.help_text_markdown)
```
jmsmkn, line 487: Maybe this goes back to which filter am I supposed to use, but why do we need to embed youtube videos in help text?

app/grandchallenge/core/templatetags/bleach.py
chrisvanrun: I imagine them roughly for:

- `md2html`: all rendering without a specific filter, sane defaults. Generally used for things that leave GC (i.e. go to CIRRUS).
- `md2html_email`: rendering emails, email specific HTML features (should really remove the BS4 additions here)
- `md2page_html`: rendering information pages to be shown as information or blog or documentation

app/grandchallenge/core/templatetags/bleach.py
jmsmkn: Just investigate, install nh3 in an empty python project and see if it is possible to do the same, and what the solution would look like.

app/grandchallenge/reader_studies/models.py
chrisvanrun: I got confused by this name as well. But it is actually the text that gets shown on the reader study Information page.

app/grandchallenge/core/templatetags/bleach.py
jmsmkn: I still don't get why we need the new split?

app/grandchallenge/core/templatetags/bleach.py
chrisvanrun: Yes, but not straightforward. nh3 doesn't seem to support it either. We could introduce a custom markdown extension for that. But that runs before the cleaning. 

It also has minimum benefit: we'll allow for 'weak' sandboxes so someone could still use a `sandbox="allow-x allow-y"` and have a not-effective sandbox. We could introduce sub-attribute checks.... but that might be going overboard.

jmsmkn: I'm wondering if we're going about this the wrong way because we don't have much control here. Maybe we should first convert the markdown to html and end up with a safestring. Then, we do our own post-processing on that and insert our own html, then no messing about with bleach. We used to have such a thing that worked with `{% tag argument %}` but it could be easily changed for `[tag argument]`:

https://github.com/comic/grand-challenge.org/blob/c2737876c91e15626f2e6908af741279e22d588a/app/grandchallenge/pages/substitutions.py

https://github.com/comic/grand-challenge.org/blob/c2737876c91e15626f2e6908af741279e22d588a/app/tests/pages_tests/test_substitutions.py

chrisvanrun: I thought your suggestion was a very good one: why bother with bleaching and clean content when we actually generate the HTML ourselves! As such I have re-introduced the `Substitution` class but named it `TagSubstitution` and updated a bit to work with Callables.

Extending with new custom tags should now be really, _really_ easy.

app/grandchallenge/core/templatetags/bleach.py
chrisvanrun: No longer relevant after the refactor.

app/grandchallenge/core/templatetags/bleach.py
chrisvanrun: Not relevant any more. However, just to comment since I just had my hands on bleach and its settings. I think `nh3` is nigh perfect drop-in replacement for `bleach`.

jmsmkn:
I think something went wrong with merging/rebasing 

Screenshot 2023-11-24 at 18 16 57

chrisvanrun:
> I think something went wrong with merging/rebasing
> 
> Screenshot 2023-11-24 at 18 16 57

Yep. I did a git pull. Not sure what kind of strategy was applied. I'll clean up a bit.

chrisvanrun:
OK. Cleaned it up again by just dropping the new (?) commits. I'm definitely lacking the focus to process further.

chrisvanrun:
Ping!

jmsmkn:
I know, it's on my TODO list sorry, higher priorities right now.

chrisvanrun:
> I know, it's on my TODO list sorry, higher priorities right now.

No worries. Reasonable and understood! Just wanted to make sure my last comment wasn't leaving this hanging somewhere.

app/tests/core_tests/test_tag_substitutions.py
```diff
@@ -0,0 +1,102 @@
...
+        (
+            mark_safe("[ foo safe ] [ foo safe ]"),
+            _arg_dependent_replacement,
+            SafeString,
+        ),
+    ],
+)
+def test_safe_substitutions_with_arguments(inp, content, typ):
+    result = TagSubstitution(tag_name="foo", replacement=content)(inp)
+    assert isinstance(result, typ)
```
jmsmkn, line 102: What is this test doing? I do not really understand what is going on with `_arg_dependent_replacement`. Why is it being added as a parameter anyway? Seems to be the same in all tests.

app/grandchallenge/core/utils/tag_substitutions.py
```diff
@@ -0,0 +1,98 @@
...
+        if isinstance(self.replacement, Callable):
+            self.num_args = self._get_num_positional_args(self.replacement)
+
+        if not re.search(r"\A[a-zA-Z_\-]+\Z", self.tag_name):
+            raise ValueError(f"{self.tag_name} is not a valid tag name.")
+
+    @property
+    def pattern(self) -> str:
+        pattern = rf"{self.markers[0]}\s*{self.tag_name}"
+        for _ in range(0, self.num_args):
```
jmsmkn, line 59:
```suggestion
        for _ in range(self.num_args):
```

app/grandchallenge/core/templatetags/process_tags.py
```diff
@@ -0,0 +1,25 @@
...
+from django.template.loader import render_to_string
+
+from grandchallenge.core.utils.tag_substitutions import TagSubstitution
+
+register = template.Library()
+
+process_youtube_tags = TagSubstitution(
+    tag_name="youtube",
+    replacement=lambda _id: render_to_string(
+        "partials/youtube_embed.html",
```
jmsmkn, line 11: The partial should be moved to a namespace:

```suggestion
        "grandchallenge/partials/youtube_embed.html",
```

app/tests/core_tests/test_tag_processing.py
```diff
@@ -0,0 +1,32 @@
...
+
+@pytest.mark.parametrize(
+    "input_html, expected_html",
+    [
+        (
+            "[youtube QCYYhkTlnhQ]",
+            EXPECTED_YOUTUBE_EMBED,
+        ),
+        (  # Random white-spaces and newlines
+            "[  youtube\n\t QCYYhkTlnhQ ]",
```
jmsmkn, line 25: What is the output when people forget to include the id argument?

jmsmkn (dismissed):

jmsmkn (approved):

app/tests/core_tests/test_tag_substitutions.py
chrisvanrun: Did you end up changing it? I am also not sure why past Chris didn't just add it as fixed.

jmsmkn (approved):

PR #3107, opened 2023-11-28
switch `pre-commit` | bot

Borda:
I would suggest switching to a [pre-commit bot](https://github.com/marketplace/pre-commit-ci), which would also commit fixes within each PR, which overall simplifies endorsing formatting and maintenance. Note that pre-commit is already used by this repo, and the bot is free for all public projects :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,22 +11,6 @@ concurrency:
 
 jobs:
 
-  precommit:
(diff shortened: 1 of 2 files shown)


jmsmkn:
Thanks but no thanks, we do not want to add an additional dependency on more 3rd party infrastructure so do everything in GitHub actions.

PR #3109, opened 2023-11-29
Add option for open algorithm phase

jmsmkn:
See https://github.com/DIAGNijmegen/rse-roadmap/issues/257


app/grandchallenge/evaluation/models.py
```diff
@@ -436,6 +436,23 @@ class Phase(UUIDModel, ViewContentMixin):
             MaxValueValidator(limit_value=3600),
         ],
     )
+    give_algorithm_editors_job_view_permissions = models.BooleanField(
```
amickan, line 439: Should we make this field read_only in the admin to prevent it from being changed accidentally?

amickan: Also, minor and maybe non-sense, but I was a bit confused by the name at first. I read `algorithm editors` and somehow thought `challenge admins`. Algorithm editors is definitely the correct term here, of course, but maybe adding some extra clarification for dummies to the help text could prevent confusion for challenge admins? Something like `...algorithm editors (i.e. challenge participants)`. Of course the algorithm creator is not necessarily the only editor, so that's also inaccurate. Hm, unsure, just thought I'd mention my initial confusion. :-)

amickan (dismissed):

app/grandchallenge/evaluation/models.py
jmsmkn: > Should we make this field read_only in the admin to prevent it from being changed accidentally?

Good idea, are there any other settings you can think of that we shouldn't touch?

jmsmkn: > ...algorithm editors (i.e. challenge participants)

Agreed!

amickan (dismissed):

amickan (approved):

PR #3113, opened 2023-11-30
Add credit limits for algorithm editors

amickan:
This adds credit limits for algorithm editor through an environment variable with a default of 5. See pitch for more details.

I have not yet made verification required for the view or the `add_editor` method since that touches a lot of tests and hence seemed better to do seperately.

Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/215

### DIFF
--- a/app/grandchallenge/algorithms/models.py
+++ b/app/grandchallenge/algorithms/models.py
@@ -436,13 +436,15 @@ class Algorithm(UUIDModel, TitleSlugDescriptionModel, ViewContentMixin):
             algorithm_image__image_sha256=self.active_image.image_sha256,
         ).values_list("pk", flat=True)
         user_credit = Credit.objects.get(user=user)
+        credits_left = max(user_credit.credits, 0)
+        credits_per_job = max(self.credits_per_job, 1)
 
         if (
             self.is_editor(user=user)
             and len(editor_jobs) < settings.ALGORITHMS_JOB_LIMIT_FOR_EDITORS
         ):
             return (
-                (max(user_credit.credits, 0) // max(self.credits_per_job, 1))
+                (credits_left // credits_per_job)
                 + settings.ALGORITHMS_JOB_LIMIT_FOR_EDITORS
                 - len(editor_jobs)
             )
@@ -451,10 +453,8 @@ class Algorithm(UUIDModel, TitleSlugDescriptionModel, ViewContentMixin):
                 pk__in=editor_jobs[: settings.ALGORITHMS_JOB_LIMIT_FOR_EDITORS]
             ).spent_credits(user=user)
             if jobs["total"]:
-                credits_left = user_credit.credits - jobs["total"]
-            else:
-                credits_left = user_credit.credits
-            return max(credits_left, 0) // max(self.credits_per_job, 1)
+                credits_left = credits_left - jobs["total"]
+            return max(credits_left, 0) // credits_per_job
 
     @cached_property
     def user_statistics(self):
--- a/app/grandchallenge/algorithms/models.py
+++ b/app/grandchallenge/algorithms/models.py
@@ -437,12 +437,12 @@ class Algorithm(UUIDModel, TitleSlugDescriptionModel, ViewContentMixin):
                     creator__in=self.editors_group.user_set.all(),
                     algorithm_image__image_sha256=self.active_image.image_sha256,
                 ).count()
-                n_free_jobs = max(
-                    settings.ALGORITHMS_JOB_LIMIT_FOR_EDITORS - n_editor_jobs,
-                    0,
-                )
             else:
-                n_free_jobs = settings.ALGORITHMS_JOB_LIMIT_FOR_EDITORS
+                n_editor_jobs = 0
+            n_free_jobs = max(
+                settings.ALGORITHMS_JOB_LIMIT_FOR_EDITORS - n_editor_jobs,
+                0,
+            )
             user_credits += n_free_jobs * self.credits_per_job
 
         credits_left = user_credits - spent_credits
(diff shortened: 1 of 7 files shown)


app/grandchallenge/algorithms/models.py
```diff
@@ -427,17 +427,34 @@ def remove_user(self, user):
 
     def get_jobs_limit(self, user):
         """Get the maximum number of jobs a user can schedule now"""
-        if self.is_editor(user=user):
-            # Not limited by credits
-            return
+        if not self.active_image:
```
jmsmkn, line 430: I'm finding `get_jobs_limit` quite hard to follow. I would have thought that you could have done the normal `spent_credits` lookup, and then have an if statement for whether they're an admin, something along the lines of:

```python
def get_jobs_limit(self, user):
  user_credits = Credit.objects.get(user=user)

  # This following line is also a bit confusing
  spent_credits = Job.objects.spent_credits(user=user)["total"]

  if self.is_editor(user=user) and self.active_image:
    # Including all jobs here as failed jobs still use compute
    spent_by_editors = Job.objects.filter(
      creator__in=self.editors_group.user_set.all(),
      algorithm_image__image_sha256=self.active_image.image_sha256,
    ).aggregate(
      total=Sum("algorithm_image__algorithm__credits_per_job")
    )["total"]

    user_credits += settings.ALGORITHMS_JOB_LIMIT_FOR_EDITORS * self.credits_per_job
    
    spent_credits -= spent_by_editors

  credits_left = user_credits - spent_credits

  return max(credits_left, 0) // max(self.credits_per_job, 1)
```

Or am I missing something here?

amickan: That should work as well. I think it ultimately boils down to the same result, but is easier to follow. Thanks! 

Some questions:
- will 5 be enough for debugging if we include failed jobs in the count?
- Why only check for the existence of an `active_image` if the user is an editor? If this method is supposed to return how many jobs a user can currently schedule, it should return None regardless of whether the user is an editor or not if there is no active_image (hence why I pulled this to the top of the method). Or am I missing something?

app/grandchallenge/algorithms/models.py
```diff
@@ -427,17 +427,34 @@ def remove_user(self, user):
...
+            return (
+                (credits_left // credits_per_job)
+                + settings.ALGORITHMS_JOB_LIMIT_FOR_EDITORS
+                - len(editor_jobs)
+            )
         else:
-            user_credit = Credit.objects.get(user=user)
-            jobs = Job.objects.spent_credits(user=user)
+            jobs = Job.objects.exclude(
+                pk__in=editor_jobs[: settings.ALGORITHMS_JOB_LIMIT_FOR_EDITORS]
```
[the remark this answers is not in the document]
amickan: If we were to exclude all editor jobs here, all editors would always end up with the full amount of credits (i.e. nothing spent). And it doesn't matter which ones we exclude, right? All jobs cost equally much. 

But I think your suggestion to increase the editors user credits rather than decreasing the spent credits is easier to follow, so I will adopt that approach.

app/grandchallenge/algorithms/models.py
jmsmkn: > will 5 be enough for debugging if we include failed jobs in the count?

We can see, they also have their normal credits.

> Why only check for the existence of an active_image if the user is an editor? If this method is supposed to return how many jobs a user can currently schedule, it should return None regardless of whether the user is an editor or not if there is no active_image (hence why I pulled this to the top of the method). Or am I missing something?

I think they're separate questions, right? This way they get an answer as if there were an active image. The logic about whether they can actually schedule a job or not elsewhere is handled elsewhere.

There might be a small bug in the example - `spent_by_editors` should be no more than `settings.ALGORITHMS_JOB_LIMIT_FOR_EDITORS * self.credits_per_job` (pull that to another variable if needs be).

amickan: > The logic about whether they can actually schedule a job or not elsewhere is handled elsewhere.

Good point. That makes sense.

amickan: Wait, actually, I'm confused by the above. Why are you both increasing the user credits for editors and decreasing their spent credits? We should only be doing one of the two I think.

jmsmkn: You're right, I've update it and fixed the logic bug I mentioned too.

amickan: Ok, yes, now I think it's only missing the check for the `active_image` (otherwise the query will fail) and another if statement because spent_credits can be None.

jmsmkn: > Ok, yes, now I think it's only missing the check for the `active_image` (otherwise the query will fail) and another if statement because spent_credits can be None.

Ah, to get around that you can set `default=0` as an argument to `Sum()` https://docs.djangoproject.com/en/4.2/topics/db/aggregation/#aggregating-on-empty-querysets-or-groups

jmsmkn: One last suggestion for readability:

```python
            if self.active_image:
                # Including all jobs here as failed jobs still use compute
                n_editor_jobs = Job.objects.filter(
                    creator__in=self.editors_group.user_set.all(),
                    algorithm_image__image_sha256=self.active_image.image_sha256,
                ).count()
            else:
                n_editor_jobs = 0

            n_free_jobs = max(
                settings.ALGORITHMS_JOB_LIMIT_FOR_EDITORS - n_editor_jobs,
                0,
            )
            user_credits += n_free_jobs * self.credits_per_job
```

What do you think?

jmsmkn (approved):

PR #3116, opened 2023-12-01
Check if image exists before deleting civs for archive items

### LINKED ISSUE #3104: Annotation CIV disapears after uploading image CIV
**Describe the bug**
When uploading a Generic Overlay CIV to an archive item that contains one or multiple annotations CIVs (tested with 2D bounding box and point annotations), those annotations disappear.

**To Reproduce**
1. Go to an archive item that contains some annotation CIVs
2. Upload an image CIV to the archive item.
3. Refresh to see the annotation CIVs disappeared from the archive item.

**Expected behavior**
The annotation CIVs should remain.

**Screenshots**

https://github.com/comic/grand-challenge.org/assets/2001094/7e54b5c1-4d17-40ae-8ebc-9acf743861cd

amickan:
When uploading a new image to an existing archive item, all non-image CIVs of that same item used to get deleted. The reason for this was that we checked for existing CIVs with the same image and deleted those. In the case of new image uploads, however, civ.image is still empty at the time of checking, and so it matched all CIVs with image=None. 

I was confused why we need this check at all. After some digging, I realized that it's necessary to enable updating of the interface of an image through the api (say from generic medical image to something more specific). This was a common use case when it was added. I'm in doubt whether we still need or want to offer this feature. It means that you can never have an archive item with the same image attached to different interfaces. Seems like a reasonable restriction to me, but it's a restriction that we don't enforce anywhere else (i.e. for display sets or jobs) as far as I know. 

Closes #3104

### DIFF
--- a/app/grandchallenge/archives/static/archives/js/archive_item_update.js
+++ b/app/grandchallenge/archives/static/archives/js/archive_item_update.js
@@ -1,0 +1,10 @@
+$('#ajaxDataTable').on( 'init.dt', function() {
+   allSelectElements = document.querySelectorAll('[id^="interfaceSelect"]');
+   allSelectElements.forEach(function(elem) {
+        elem.addEventListener("change", loadUpdateView);
+   });
+});
+
+function loadUpdateView(source) {
+    window.location.href = source.target.value
+}
--- a/app/grandchallenge/archives/templates/archives/archive_items_list.html
+++ b/app/grandchallenge/archives/templates/archives/archive_items_list.html
@@ -30,4 +30,5 @@
 {% block script %}
     {{ block.super }}
     {% include 'workstations/partials/session-control.html' %}
+    <script src="{% static 'archives/js/archive_item_update.js' %}"></script>
 {% endblock %}
--- a/app/grandchallenge/archives/tasks.py
+++ b/app/grandchallenge/archives/tasks.py
@@ -133,7 +133,9 @@ def update_archive_item_values(*, archive_item_pk, civ_pks_to_add):
             ).values_list("pk", flat=True):
                 civ_pks_to_remove.append(civ_pk)
         # for images, check if there are any CIVs with the provided image
-        if civ.interface.is_image_kind:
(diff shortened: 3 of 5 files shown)


jmsmkn:
> It means that you can never have an archive item with the same image attached to different interfaces. Seems like a reasonable restriction to me, but it's a restriction that we don't enforce anywhere else (i.e. for display sets or jobs) as far as I know.

Test it out, I don't think that restriction is necessary. If we do allow this, do users have an easy way of removing values from a display set?

app/grandchallenge/archives/static/archives/js/archive_item_update.js
```diff
@@ -0,0 +1,10 @@
+$('#ajaxDataTable').on( 'init.dt', function() {
```
jmsmkn, line 1: Can you do this without jquery?

app/grandchallenge/archives/templates/archives/archive_items_list.html
```diff
@@ -30,4 +30,5 @@ <h2>Items for {{ archive.title }}</h2>
 {% block script %}
     {{ block.super }}
     {% include 'workstations/partials/session-control.html' %}
+    <script src="{% static 'archives/js/archive_item_update.js' %}"></script>
```
jmsmkn, line 33: Can you make this a js module?

app/grandchallenge/archives/static/archives/js/archive_item_update.js
amickan: hm, I'm not sure how actually. I went with what is recommended in the datatables documentation. I also tried adding an event listener to the data table, but couldn't get that work. How else would I do it?

jmsmkn: Ah, I neglected that this is datatables, then indeed not possible, sorry my fault!

jmsmkn (approved):

jmsmkn (approved):

PR #3119, opened 2023-12-04
Require verification for alg editors and job creation

amickan:
Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/215 
Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/201


jmsmkn (dismissed):

jmsmkn (approved):

jmsmkn:
@amickan could you check the tests, I cannot work out why they're now failing? I needed to merge as there was a conflict.

app/grandchallenge/groups/views.py
```diff
@@ -36,7 +36,11 @@ def get_success_url(self):
         return self.obj.get_absolute_url()
 
     def form_valid(self, form):
-        form.add_or_remove_user(obj=self.obj)
+        try:
+            form.add_or_remove_user(obj=self.obj)
+        except RuntimeError as e:
+            form.add_error(None, e)
```
jmsmkn, line 42: Runtime errors should not be added to forms, these could be things deep within our systems whose message should not be revealed to the user. The error shouldn't be raised here anyway as the problem should be caught in form validation, right?

amickan: Hm, ok. I tried to add the check to the user clean method on the form, but since I don't know what the object is in the EditorsForm, the check would be applied also for archives and reader studies. Is that something that we're ok with? 

Otherwise, I think I would need to update the form to take the object as argument as well. Or is there another way?

jmsmkn: You could take an argument on init. Or just drop this requirement. Does it really matter if editors are not verified? It's running jobs or uploading containers that we really care about.

amickan: Yeah ok, it's no use to be an editor of an algorithm if you're not verified since you can't run jobs without verification, but indeed it's not strictly necessary to check that. I honestly think that I'll drop the requirement then, seems like unnecessary work to adjust the form for too little added benefit.

PR #3125, opened 2023-12-08
Turn display set list of cards into table

### LINKED ISSUE #3042: Unclear when uploading for a DisplaySet fails
# Description
* Click "Add Case (single)"
* Provide files for all the interfaces
* Hit the save button

# Expected:
The status of the upload(s) is visible somewhere

# What actually happens:
You are presented with an empty display set (none) with no idea what is going on:

image

amickan:
This PR turns the display set list of cards into a table and adds a seperate update view. 

New list view:
image

Update view:
image

Closes #3042


app/grandchallenge/reader_studies/views.py
```diff
@@ -1323,8 +1317,126 @@ class DisplaySetDetail(
...
+            errors.update(e.message_dict)
+
+        if errors:
+            return self.return_errors(errors)
+        else:
+            messages.success(
+                self.request,
+                "Display set created. Image and file import jobs have been queued.",
+            )
+            return JsonResponse({"redirect": self.get_success_url()})
```
amickan, line 1384: This is again a bit awkward: we're returning errors on form_valid. I took this over as is because I can't think of a different way of doing this. The reason we do this is that we add new interfaces to a display set and they can't be validated as part of the normal form.

jmsmkn: Without looking at this in detail shouldn't it be done in a custom `clean` method in a form?

amickan: Yes, I'm actually just working on that! ;-)

app/grandchallenge/components/views.py
```diff
@@ -84,3 +89,46 @@ def get_queryset(self):
...
+class InterfaceProcessingMixin:
+    def process_data_for_object(self, data):
+        raise NotImplementedError
+
+    def form_invalid(self, form):
+        return JsonResponse(form.errors, status=400)
+
+    def form_valid(self, form):
+        self.process_data_for_object(form.cleaned_data)
+        return JsonResponse({"redirect": self.get_success_url()})
```
amickan, line 103: This might seem strange, but we want to prevent the default from happening here since the form can contain extra, dynamically added fields (new interfaces) for which we want to be able to show errors as well. The default here would just return the form with the errors for the fields that were declared in the form init function. The fields that are added by the user would just disappear.

chrisvanrun:
Post merge / release don't forget to update: https://grand-challenge.org/documentation/define-your-cases/

app/grandchallenge/components/forms.py
```diff
@@ -79,3 +84,166 @@ def save(self, *args, **kwargs):
...
+            required=False,
+            user=self.user,
+        ).field
+
+    def clean(self):
+        cleaned_data = super().clean()
+        try:
+            cleaned_data.update(self._process_new_interfaces())
+        except ValidationError as e:
+            self.errors.update(e.message_dict)
```
jmsmkn, line 151: We don't need to raise something here?

app/grandchallenge/components/forms.py
```diff
@@ -79,3 +84,166 @@ def save(self, *args, **kwargs):
...
+            )
+        else:
+            selected_interface = None
+
+        widget_kwargs = {}
+        attrs = {
+            "hx-get": htmx_url,
+            "hx-trigger": "interfaceSelected",
+            "disabled": selected_interface is not None,
+            "hx-target": f"#form-{kwargs['auto_id'][:-3]}",
```
jmsmkn, line 216: Why cut the last chars?

app/grandchallenge/components/static/components/js/autocomplete_htmx.mjs
```diff
@@ -1,23 +1,20 @@
...
-    });
-    if (vals.length) {
-      vals = vals.map(val => `{"type": "const", "dst": "interface-${val}", "val": "${val}"}`);
-    }
-    vals.push(`{"type": "const", "dst": "reader-study", "val": "${$('#readerStudySlug').data('slug')}"}`);
-    $(".dal-forward-conf script").text(`[${vals.join(',')}]`);
-  });
-
-});
+$(document).ready(() => {
```
jmsmkn, line 1: In general it would be good to do things without JQuery (although we cannot remove it as we're using a large tables app built with it). Most things should be available as native browser functionality now, see https://youmightnotneedjquery.com.

app/grandchallenge/components/views.py
jmsmkn: Hmm, seems a bit anti-Django. Is there another approach? Could you append the existing fields on the form?

app/grandchallenge/components/views.py
```diff
@@ -84,3 +89,46 @@ def get_queryset(self):
...
     def get_result_label(self, result):
         return result.title
+
+
+class InterfaceProcessingMixin:
+    def process_data_for_object(self, data):
+        raise NotImplementedError
+
+    def form_invalid(self, form):
+        return JsonResponse(form.errors, status=400)
```
jmsmkn, line 99: Missing `super()`?

app/tests/reader_studies_tests/test_views.py
```diff
@@ -454,8 +425,9 @@ def test_display_set_update(client):
         },
         user=user,
         method=client.post,
+        content_type="application/json",
     )
-    assert response.status_code == 302
+    assert response.status_code == 200
```
jmsmkn, line 430: Why are we no longer redirecting on creation? https://en.wikipedia.org/wiki/Post/Redirect/Get

app/grandchallenge/components/views.py
amickan: In this case, super().form_invalid() only renders the form with error messages, right?. We can call it, but we wouldn't do anything with the output since we need to return the errors in json format for our custom js code to pick them up and add them to the form correctly.

jmsmkn: That may be the case now but you do not know that in the future for all uses of this Mixin. It is best practise to always call the super method as the view could be doing something custom.

app/grandchallenge/components/views.py
amickan: Maybe I have found a way. Needs more work still, but I think I'm getting there.

amickan: Got it to work, but I'm not sure if this implementation is 100% kosher either, and it does come with some drawbacks as well. I'll commit and see what you think. From a user perspective, it is a bit weird that the look of the form changes when a validation error is returned:

Before submitting the form with a new interface (number):
image

and after validation:

image

The box around the new interface disappears and also the option to remove it.

app/grandchallenge/components/forms.py
amickan: Changed implementation, ValidationError is now raised in `_process_new_interfaces`.

app/grandchallenge/components/views.py
jmsmkn: What happened to the submitted value and the metics json file widget?

amickan: That the widget disappears is a seperate issue with the json editor, I will create a seperate PR (it's related to https://github.com/comic/grand-challenge.org/issues/3162). That the value of number disappears is something that I'm still trying to understand.

amickan: Got the values to persist. Implementation complete, but I'm now wondering if using htmx to send along the new interfaces (instead of the custom js script that currently does this) might also work and be neater. We could use hx-include to include them. Any objections before I go down the potential rabbit whole of getting that to work?

amickan: I don't see why we wouldn't use htmx, makes for a lot less code. I'll prepare a commit

app/grandchallenge/reader_studies/views.py
```diff
@@ -1456,17 +1432,18 @@ def form_valid(self, form):
...
     def get_success_url(self):
         return reverse(
-            "reader-studies:display-set-detail",
-            kwargs={"pk": self.kwargs["pk"], "slug": self.kwargs["slug"]},
+            "reader-studies:display_sets",
+            kwargs={"slug": self.kwargs["slug"]},
         )
 
 
 class DisplaySetInterfacesCreate(ObjectPermissionRequiredMixin, FormView):
```
amickan, line 1440: For this view, we're no longer using the post method. We're only using this to get the correct form fields for newly added interfaces. Validating form data and adding the interface and value to the display set is done through the create and update views (they are now fields on the create / update form). So that means that the `form_valid` method here can go, and maybe we should explicitly deny the post method on this view? Something like 

```python
def post(self):
    return HttpResponseNotAllowed()
```

or we raise a RunTimeError so that we are made aware of this happening?

amickan: The same goes for the `DisplaySetFilesUpdate` view, which I haven't updated yet in the interest of keeping the changeset smaller, but it'll follow the same logic as the `InterfacesCreate` view.

app/tests/reader_studies_tests/test_views.py
amickan: In the latest implementation the reason for this is that the view we're calling here is an HTMX view. In order to redirect on success, we need to return an HttpResponse with the appropriate HX- headers. Django's HttpResponseRedirect unfortunately does not support these headers, and so we are stuck with HttpResponse, hence the 200 status code.

app/grandchallenge/reader_studies/views.py
jmsmkn: Sound like a strange construction but from what you say my guess is that it should be a `TemplateView` where you add in the form handling yourself.

amickan: Yeah, I guess that makes sense. I will change it to a TemplateView then

app/grandchallenge/components/forms.py
```diff
@@ -79,3 +86,156 @@ def save(self, *args, **kwargs):
...
+                    interface__slug=slug
+                ).first()
+
+            self.init_interface_field(
+                interface_slug=slug, current_value=current_value, values=values
+            )
+        # Add fields for dynamically added new interfaces:
+        # These are sent along as form data like all other fields, so we can't
+        # tell them apart from the form fields initialized above. Hence
+        # the check if they already have a corresponding field on the form or not.
```
jmsmkn, line 118: Maybe flu-brain still but I cannot work out this construction. Wouldn't it make more sense to have two separate forms, one for adding new fields (`CreateView`) and one for updating the existing ones (`UpdateView`)?

app/grandchallenge/components/forms.py
```diff
@@ -79,3 +86,156 @@ def save(self, *args, **kwargs):
...
+                self.init_interface_field(
+                    interface_slug=slug,
+                    current_value=None,
+                    values=[],
+                )
+
+    def save(self):
+        # we need a dummy save method for form_valid() when called from a
+        # generic CreateView/UpdateView
+        pass
```
jmsmkn, line 133: Is this still necessary now that you switched to a `TemplateView`?

amickan: Yes, because this is the form for the DisplaySetCreate and DisplaySetUpdate views.

app/grandchallenge/components/forms.py
amickan: The form names are probably a bit confusing, let me explain, and then we can settle on better names: 
- `MultipleCIVCreateForm`: this form is the underlying form for both the DisplaySetCreate and DisplaySetUpdate form. It's a form that is dynamically constructed to contain fields for all interfaces that exist on other display sets of a given reader study, as well as potentially added new interfaces (through htmx).
- `ComponentInterfaceCreateForm`: this form is used to add one single new interface. Initially it contains an autocomplete field that lets the user choose an interface, when an interface is sent along with the get request to the view, it also contains the appropriate field for the chosen interface.

app/grandchallenge/components/views.py
```diff
@@ -84,3 +89,50 @@ def get_queryset(self):
...
+        # since this is an HTMX view, we need to trigger
+        # the redirect through inclusion of the HX headers
+        # HttpResponseRedirect does not support the inclusion of these custom headers
+        messages.add_message(
+            self.request, messages.SUCCESS, self.success_message
+        )
+        return HttpResponse(
+            self.get_success_url(),
+            headers={
+                "HX-Redirect": self.get_success_url(),
```
jmsmkn, line 110: Instead of `self.get_success_url()` I think this needs to be `response.url` with `response = super().form_valid(form)`.

app/grandchallenge/reader_studies/templates/reader_studies/display_set_create.html
```diff
@@ -39,11 +39,9 @@ <h2>Add {{ type_to_add }} to {{ object }}</h2>
...
     {{ block.super }}
     {{ form_media }}
-    <script src="{% static 'reader_studies/js/display_set_create.js' %}" type="text/javascript"></script>
-    <script src="{% static 'reader_studies/js/autocomplete_htmx.js' %}" type="text/javascript"></script>
+    <script type="module" src="{% static 'components/js/create_extra_interfaces.mjs' %}" type="text/javascript"></script>
+    <script type="module" src="{% static 'components/js/autocomplete_htmx.mjs' %}" type="text/javascript"></script>
     <script type="text/javascript" src="{% static "cases/js/ignore_enter_button.js" %}"></script>
-
-    {# TODO: This might not be used #}
     {% include 'grandchallenge/partials/drf_csrf.html' %}
```
jmsmkn, line 45: This partial is definitely necessary?

app/grandchallenge/reader_studies/views.py
```diff
@@ -1342,47 +1327,41 @@ class DisplaySetUpdate(
...
+    included_form_classes = (
+        DisplaySetUpdateForm,
+        ComponentInterfaceCreateForm,
+        FileForm,
+    )
+    success_message = format_html(
+        "Display set updated. Image and file import jobs have been queued. "
+        "You will be notified about errors related to image and file imports "
+        "via a <a href={}>notification</a>.",
+        "https://grand-challenge.org/notifications/",
```
jmsmkn, line 1339: Lets start to always use the kwargs to `format_html`, these things always end up quite complex.

```suggestion
        "via a <a href={url}>notification</a>.",
        "url=https://grand-challenge.org/notifications/",
```

app/grandchallenge/reader_studies/views.py
```diff
@@ -1540,131 +1512,32 @@ class AddDisplaySetToReaderStudy(
     )
     included_form_classes = (
         DisplaySetCreateForm,
-        DisplaySetInterfacesCreateForm,
+        ComponentInterfaceCreateForm,
+    )
+    success_message = format_html(
```
jmsmkn, line 1517: Can you remove the duplication of the success message?

app/grandchallenge/components/forms.py
amickan: > Doesn't it mean that you skip saving if you do add this mixin to a form that actually updates something? Again, you seem to be fighting the framework too much with this design.

That is true. Maybe changing the Create and Update views to FormViews would be more appropriate then. It would not require the dummy save method on the form mixin, and in the end neither the create nor the update views for display sets are traditional create and update views, since the form is dynamically constructed.

app/tests/reader_studies_tests/test_views.py
jmsmkn: So why not `HttpResponse` with the headers and status code 302?

amickan: That would work. I thought I should better not change the status codes of these classes, but I see now that HttpResponse is a pretty generic response, unlike the specific ones like HttpResponseRedirect etc, so ok.

app/grandchallenge/components/forms.py
amickan: I turned the create and update views into FormViews and have also updated the form names. Hope this clarifies things.

app/grandchallenge/reader_studies/templates/reader_studies/display_set_create.html
amickan: Actually not, also not in the update html. In the list view template it is necessary for removing display sets.

jmsmkn (approved):
Nice refactor!

PR #3127, opened 2023-12-11
Add 3 point angle annotation type

pkcakeout:
- Add json schema(s)
- Add test data
- Add migrations

Pitch https://github.com/DIAGNijmegen/rse-roadmap/issues/283


app/grandchallenge/components/schemas.py
```diff
@@ -491,6 +491,67 @@
...
+            },
+            "required": ["angle"],
+            "additionalProperties": False,
+        },
+        "M3AN": {
+            "type": "object",
+            "properties": {
+                "name": {"type": "string"},
+                "type": {"enum": ["Multiple three-point angles"]},
+                "boxes": {
```
jmsmkn, line 542: Wouldn't `"angles"` be more appropriate than `"boxes"` here?

app/grandchallenge/components/schemas.py
```diff
@@ -491,6 +491,67 @@
...
+                        "minItems": 3,
+                        "maxItems": 3,
+                    },
+                    "minItems": 3,
+                    "maxItems": 3,
+                },
+                "version": {"$ref": "#/definitions/version-object"},
+                "probability": {"type": "number", "minimum": 0, "maximum": 1},
+            },
+            "required": ["angle"],
```
jmsmkn, line 534: Shouldn't `"version"` also be required?

app/tests/components_tests/test_serializers.py
```diff
@@ -195,6 +195,45 @@
...
+            [76.801, -57.62, 84.43],
+            [77.10, -18.97, 115.48],
+        ],
+        "probability": 0.92,
+        "version": {"major": 1, "minor": 0},
+    },
+    "M3AN": {
+        "name": "Some annotations",
+        "type": "Multiple three-point angles",
+        "angles": [
```
jmsmkn, line 212: How did this test pass?

app/grandchallenge/components/schemas.py
pkcakeout: Sorry yes! Too many updates necessary, this is a copy-paste mistake!

app/grandchallenge/components/schemas.py
pkcakeout: Reviewing this myself yields that you are right! For the "3ANG", this should be required. The "xxx-object" types do not have those... fixing!

jmsmkn: > The "xxx-object" types do not have those... 

I think that is correct, they're only used in the Multi-x-object, the versioning on those is done at the top level.

app/tests/components_tests/test_serializers.py
pkcakeout: I forgot to include the 3-angle types in an enumeration that happened for the pytest-parameterized tests... that was the problem. Fixed!

app/grandchallenge/reader_studies/models.py
```diff
@@ -1204,6 +1216,8 @@ class QuestionWidgetKindChoices(models.TextChoices):
     AnswerType.MULTIPLE_ELLIPSES: None,
     AnswerType.MULTIPLE_CHOICE: [],
     AnswerType.MULTIPLE_CHOICE_DROPDOWN: [],
+    AnswerType.THREE_POINT_ANGLE: None,
+    AnswerType.MULTIPLE_THREE_POINT_ANGLES: None,
```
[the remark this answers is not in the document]
pkcakeout: Added 2fbad11a372114af2a369d094a5e75e78481f6e0 - I assume that is what you maybe meant for the reader-study tests? Thanks for the hint!

jmsmkn (approved):

PR #3143, opened 2023-12-19
Add grand-challenge-forge JSON descriptors

chrisvanrun:
Part of the pitch:
- https://github.com/DIAGNijmegen/rse-roadmap/issues/237

This adds fields to the admin pages for `Phase` and `Challenge`.

The value of the fields can be used as input for https://github.com/DIAGNijmegen/rse-grand-challenge-forge

### DIFF
--- a/app/grandchallenge/challenges/admin.py
+++ b/app/grandchallenge/challenges/admin.py
@@ -1,6 +1,9 @@
+import json
+
 from django.contrib import admin, messages
 from django.contrib.admin import ModelAdmin
 from django.core.exceptions import ValidationError
+from django.utils.safestring import mark_safe
 
 from grandchallenge.challenges.emails import send_challenge_status_update_email
 from grandchallenge.challenges.models import (
@@ -15,11 +18,17 @@ from grandchallenge.core.admin import (
     UserObjectPermissionAdmin,
 )
 from grandchallenge.core.templatetags.costs import millicents_to_euro
+from grandchallenge.core.utils.grand_challenge_forge import (
+    get_forge_json_description,
+)
 
 
 @admin.register(Challenge)
 class ChallengeAdmin(ModelAdmin):
-    readonly_fields = ("creator",)
+    readonly_fields = (
+        "creator",
+        "challenge_forge_json",
+    )
     autocomplete_fields = ("publications",)
     ordering = ("-created",)
     list_display = (
@@ -39,6 +48,11 @@ class ChallengeAdmin(ModelAdmin):
 
     def available_compute_euros(self, obj):
         return millicents_to_euro(obj.available_compute_euro_millicents)
+
+    @staticmethod
+    def challenge_forge_json(obj):
+        json_desc = get_forge_json_description(challenge=obj)
+        return mark_safe(f"<pre>{json.dumps(json_desc, indent=2)}</pre>")
 
 
 @admin.register(ChallengeRequest)
--- a/app/grandchallenge/core/utils/grand_challenge_forge.py
+++ b/app/grandchallenge/core/utils/grand_challenge_forge.py
@@ -1,0 +1,54 @@
+from grandchallenge.evaluation.models import Phase
+from grandchallenge.evaluation.utils import SubmissionKindChoices
+from grandchallenge.subdomains.utils import reverse
+
+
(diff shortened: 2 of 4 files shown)


app/grandchallenge/evaluation/admin.py
```diff
@@ -78,13 +84,24 @@ class PhaseAdmin(admin.ModelAdmin):
...
     def open_for_submissions(self, instance):
         return instance.open_for_submissions
 
+    @staticmethod
+    def challenge_forge_json(obj):
+        json_desc = get_forge_json_description(
+            challenge=obj.challenge,
+            phases_queryset=Phase.objects.filter(id=obj.id),
+        )
+        return mark_safe(f"<pre>{json.dumps(json_desc, indent=2)}</pre>")
```
jmsmkn, line 103: Sane here.

app/grandchallenge/evaluation/admin.py
```diff
@@ -78,13 +84,24 @@ class PhaseAdmin(admin.ModelAdmin):
...
 
     @admin.display(boolean=True)
     def open_for_submissions(self, instance):
         return instance.open_for_submissions
 
+    @staticmethod
+    def challenge_forge_json(obj):
+        json_desc = get_forge_json_description(
+            challenge=obj.challenge,
+            phases_queryset=Phase.objects.filter(id=obj.id),
```
jmsmkn, line 101: Small improvement is to avoid a second lookup here. You could instead do:

```python
json_desc = get_forge_json_description(
             challenge=obj.challenge,
             phase_pks=[obj.pk], # FYI it is best to avoid using id as that is overloaded, pk is more explicit that you're dealing with a Django object
)

...

def get_forge_json_description(challenge, phase_pks=None):
     """
     Generates a JSON description of the challenge and phases suitable for
     grand-challenge-forge to generate a challenge pack.
     """
     
     # FYI going through the reverse relation means that we know that all the phases belong to the challenge
     phases = challenge.phase_set.filter(
         archive__isnull=False,
         submission_kind=SubmissionKindChoices.ALGORITHM,
     ).prefetch_related("archive", "inputs", "outputs")

     if phase_pks is not None:
         phases = phases.filter(pk__in=phase_pks)

     ...
```

app/grandchallenge/challenges/admin.py
```diff
@@ -40,6 +49,11 @@ def get_queryset(self, *args, **kwargs):
     def available_compute_euros(self, obj):
         return millicents_to_euro(obj.available_compute_euro_millicents)
 
+    @staticmethod
+    def challenge_forge_json(obj):
+        json_desc = get_forge_json_description(challenge=obj)
+        return mark_safe(f"<pre>{json.dumps(json_desc, indent=2)}</pre>")
```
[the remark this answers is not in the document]
chrisvanrun: Makes sense. I don't think there are currently any fields that would introduce an XSS vector. But better be safe!

app/grandchallenge/evaluation/admin.py
chrisvanrun: The `pk` versus `id` is a good point here. In a silly habit, I looked at the model and straight up read the `id`. 

About the second lookup, to clarify, I thought query sets were lazily evaluated and passing an unevaluated one could be filtered before it would be.

jmsmkn: Yes, but this way we ensure that the phases passed actually belong to the challenge so it is internally consistent.

app/grandchallenge/core/utils/grand_challenge_forge.py
```diff
@@ -0,0 +1,53 @@
...
+
+    phases = challenge.phase_set.filter(
+        archive__isnull=False,
+        submission_kind=SubmissionKindChoices.ALGORITHM,
+    ).prefetch_related("archive", "inputs", "outputs")
+
+    if phase_pks is not None:
+        phases = phases.filter(pk__in=phase_pks)
+
+    archives = {p.archive.id: p.archive for p in phases}.values()
```
jmsmkn, line 19:
```suggestion
    archives = {p.archive for p in phases}
```

jmsmkn (approved):

PR #3144, opened 2023-12-21
Adds example json for some new annotation types

miriam-groeneveld:
Ellipse, Multiple ellipses, Three-point angle and Multiple three-point angle annotations.

### DIFF
--- a/app/grandchallenge/components/models.py
+++ b/app/grandchallenge/components/models.py
@@ -474,14 +474,15 @@ class InterfaceKind:
                 "ellipses": [
                     {
                         "major_axis": [[-44, 535, 0.5], [-112, 494, 0.5]],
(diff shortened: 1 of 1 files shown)


app/grandchallenge/components/models.py
```diff
@@ -443,6 +447,82 @@ def interface_type_json():
...
+                    {
+                        "name": "first",
+                        "angle": [[300, 237, 0.5], [263, 282, 0.5], [334, 281, 0.5]]
+                    },
+                    {
+                        "name": "second",
+                        "angle": [[413, 237, 0.5], [35, 160, 0.5], [367, 293, 0.5]]
+                    }
+                ],
+                "probability": 0.003,
```
chrisvanrun, line 522: Not sure that we support probability on the collection level of 'multiple' types.

jmsmkn (approved):

PR #3147, opened 2024-01-02
Minor fix for forge JSON creation

chrisvanrun:
Fix using default input and output (and not algorithm) in forge JSON creation.

Accidentally used the `inputs` instead of the `algorithm_inputs` for generating the JSON

### DIFF
--- a/app/tests/core_tests/test_grand_challenge_forge.py
+++ b/app/tests/core_tests/test_grand_challenge_forge.py
@@ -65,6 +65,17 @@ def test_get_forge_json_description():
                 ]:
                     assert ci_key in component_interface
 
+    # Quick check on CI input and outputs
+    for index, ci in enumerate(
+        description["challenge"]["phases"][0]["inputs"]
+    ):
+        assert inputs[index].slug == ci["slug"]
+
+    for index, ci in enumerate(
+        description["challenge"]["phases"][0]["outputs"]
+    ):
+        assert outputs[index].slug == ci["slug"]
+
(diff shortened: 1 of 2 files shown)


app/tests/core_tests/test_grand_challenge_forge.py
```diff
@@ -65,6 +65,17 @@ def test_get_forge_json_description():
                 ]:
                     assert ci_key in component_interface
 
+    # Quick check on CI input and outputs
+    for index, ci in enumerate(
+        description["challenge"]["phases"][0]["inputs"]
```
miriam-groeneveld, line 70: shouldn't it be algorithm_input here as well?

chrisvanrun: It could? In this case, this refers to the JSON that is fed into the forge for package creation.

Since the pack is only generated for the algorithm-type phase, it seemed odd to call it `'algorithm_inputs'` in the pack/forge package itself.

miriam-groeneveld: I think it would make it clearer, also because the evaluation method will have inputs and outputs.

chrisvanrun: I now know why I did this: it was because of the predictions.json also mapping it to "inputs" and "outputs". Of, there it is clearly an algorithm job.

chrisvanrun: I've updated will relying on CI to double-check if I replaced it correctly.

jmsmkn (approved):

PR #3152, opened 2024-01-09
Set default answer roll over to 0 and improve help text

### LINKED ISSUE #3151: Missing default for answer roll over field in reader studies
When creating a reader study, the answer roll over field is required but empty by default. The help text does not clearly mention that it should be set to 0 to "disable" the feature. I think this should be added to the help text and the default value should be set to 0.

HarmvZ:
Closes #3151


jmsmkn (approved):

chrisvanrun:
These changes make sense. Thank you Harm!

PR #3156, opened 2024-01-10
Streamline CIV creation and validation code for archive items and display sets

### LINKED ISSUE #3138: Validation errors on upload content should return a 4xx
Was working on testing an upload script and got back the following:

image

IMHO a client-side error is more appropriate here: for instance [415 Unsupported Media Type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/415)

amickan:
Closes #3138 

With this change to use the same logic and code for CIV creation for display sets and archive items, it is no longer possible to update the interface of an image in an archive item. We already discussed this [here ](https://github.com/comic/grand-challenge.org/pull/3116)briefly and had decided that this was no longer necessary.


app/grandchallenge/archives/static/archives/js/archive_item_update.mjs
```diff
@@ -1,5 +1,5 @@
 $('#ajaxDataTable').on( 'init.dt', function() {
-   allSelectElements = document.querySelectorAll('[id^="interfaceSelect"]');
+   var allSelectElements = document.querySelectorAll('[id^="interfaceSelect"]');
```
jmsmkn, line 2:
```suggestion
   let allSelectElements = document.querySelectorAll('[id^="interfaceSelect"]');

```

jmsmkn:
> With this change to use the same logic and code for CIV creation for display sets and archive items, it is no longer possible to update the interface of an image in an archive item. We already discussed this here briefly and had decided that this was no longer necessary.

How would a user achieve this now?

app/grandchallenge/components/models.py
```diff
@@ -1929,3 +1929,87 @@ def update_size_in_storage(self):
         else:
             self.size_in_storage = self.image.size
             self.size_in_registry = self.calculate_size_in_registry()
+
+
+class CIVForObjectMixin:
```
jmsmkn, line 1934: Could you make these methods keyword arg only (`self, *, ...`)? Their signatures are quite complex.

app/grandchallenge/archives/serializers.py
```diff
@@ -94,41 +89,15 @@ def __init__(self, *args, **kwargs):
...
-                }
-            ).apply_async
-        )
-
+        values = validated_data.pop("values")
+        for value in values:
+            interface = value.get("interface", None)
+            user_upload = value.get("user_upload", None)
+            image = value.get("image", None)
+            value = value.get("value", None)
```
jmsmkn, line 101: Do we validate that only one of these is set?

app/grandchallenge/components/models.py
```diff
@@ -1929,3 +1929,87 @@ def update_size_in_storage(self):
...
+    def create_civ(self, ci_slug, new_value, user=None):
+        ci = ComponentInterface.objects.get(slug=ci_slug)
+        current_civ = self.values.filter(interface=ci).first()
```
jmsmkn, line 1937: Please don't use the pattern of `filter` and `first`. It should be a `get` and then handling the error appropriately (which might mean 5xx as something has gone seriously wrong if there are two values for the same interface).

app/grandchallenge/components/tasks.py
```diff
@@ -1139,3 +1139,106 @@ def validate_voxel_values(*, civ_pk):
...
+    object_pk,
+    interface_pk,
+    upload_session_pk,
+):
+    from grandchallenge.components.models import (
+        ComponentInterface,
+        ComponentInterfaceValue,
+    )
+
+    with transaction.atomic():
```
jmsmkn, line 1160: You can hoist this up to before the function:

```python
@shared_task(...)
@transaction.atomic
def add_image_to_object(...):
  ...
```

app/grandchallenge/components/tasks.py
```diff
@@ -1139,3 +1139,106 @@ def validate_voxel_values(*, civ_pk):
...
+            interface=interface, image=image
+        )
+        if created:
+            try:
+                civ.full_clean()
+            except ValidationError as e:
+                # this should only happen for new uploads
+                upload_session.status = RawImageUploadSession.FAILURE
+                upload_session.error_message = e.message
+                upload_session.save()
```
jmsmkn, line 1190: Shouldn't we exit here?

app/grandchallenge/components/tasks.py
```diff
@@ -1139,3 +1139,106 @@ def validate_voxel_values(*, civ_pk):
...
+            upload_session.status = RawImageUploadSession.FAILURE
+            upload_session.error_message = error_message
+            upload_session.save()
+            return
+
+        object.values.remove(*object.values.filter(interface=interface))
+        civ, created = ComponentInterfaceValue.objects.get_or_create(
+            interface=interface, image=image
+        )
+        if created:
```
jmsmkn, line 1183: Add some breathing room:

```suggestion

        if created:
```

app/grandchallenge/components/tasks.py
```diff
@@ -1139,3 +1139,106 @@ def validate_voxel_values(*, civ_pk):
...
+
+    user_upload = UserUpload.objects.get(pk=user_upload_pk)
+    object = get_model_instance(
+        pk=object_pk,
+        app_label=app_label,
+        model_name=model_name,
+    )
+    interface = ComponentInterface.objects.get(pk=interface_pk)
+    error = None
+    with transaction.atomic():
```
jmsmkn, line 1217: Again can be hoisted, makes it clearer that the task is already atomic.

app/grandchallenge/components/tasks.py
```diff
@@ -1139,3 +1139,106 @@ def validate_voxel_values(*, civ_pk):
...
+        civ = ComponentInterfaceValue.objects.create(interface=interface)
```
jmsmkn, line 1218: Why use create and then rollback on error? Better to instantiate, clean and handle error/save, no?

app/grandchallenge/components/tasks.py
```diff
@@ -1139,3 +1139,106 @@ def validate_voxel_values(*, civ_pk):
...
+                civ.full_clean()
+            except ValidationError as e:
+                # this should only happen for new uploads
+                upload_session.status = RawImageUploadSession.FAILURE
+                upload_session.error_message = e.message
+                upload_session.save()
+        object.values.add(civ)
+
+
+@shared_task(**settings.CELERY_TASK_DECORATOR_KWARGS["acks-late-2xlarge"])
```
jmsmkn, line 1194: Why does this need to be on the large queue?

app/tests/components_tests/test_tasks.py
```diff
@@ -295,3 +307,111 @@ def test_update_sagemaker_shim(
...
 
     assert "MANIFEST_UNKNOWN" in error.value.stderr.decode()
+
+
+@pytest.mark.django_db
+def test_add_image_to_object(settings):
+    settings.task_eager_propagates = (True,)
+    settings.task_always_eager = (True,)
+
+    ds = DisplaySetFactory()
```
jmsmkn, line 317: These two tests only use display sets and not archive items? Is that tested elsewhere.

jmsmkn (commented):
Good cleanup, probably most of my questions come from the existing implementation.

amickan:
> > With this change to use the same logic and code for CIV creation for display sets and archive items, it is no longer possible to update the interface of an image in an archive item. We already discussed this here briefly and had decided that this was no longer necessary.
> 
> How would a user achieve this now?

With the changes here, they can no longer do this. Do we really need that functionality? It has always been a pretty incomplete feature (only for archive items, only for images and only possible through the API). It was originally meant to enable users to change existing archive items with generic medical images or generic overlays into something more specific. The specific interfaces have been around for a while now, I would assume it's less of an issue now? Also, there is always the alternative that we delete existing archive items and the user uploads from scratch. Or, maybe better, we enable deleting images from archive items / display sets (we already have this for all other interface kinds anyway), then there won't be a need to change the interface.

jmsmkn:
> Do we really need that functionality?

I think it's pretty common but not a blocker for this PR.

> Or, maybe better, we enable deleting images from archive items / display sets (we already have this for all other interface kinds anyway), then there won't be a need to change the interface.

We should enable that in a separate PR. Workflow would have to be to add the image as a new value, then remove the old one value. The other way around they would loose permission to add the image back.

app/grandchallenge/archives/serializers.py
amickan: Yes, when serializing the civ data: https://github.com/comic/grand-challenge.org/blob/72153c085ed6cec5708b5288b7f5a77a257aa6ca/app/grandchallenge/components/serializers.py#L114-L123

app/tests/components_tests/test_tasks.py
amickan: Only indirectly (i.e. we test the archive item updating views), but I'll adjust this test for completeness.

app/grandchallenge/archives/serializers.py
jmsmkn: But then by doing `or` on the values, what happens when the values are falsey (0, "", intentional None etc)? In this construction you might end up with None when 0 was intended.

app/grandchallenge/components/tasks.py
amickan: I just kept that setting from the original implementation and didn't re-evaluate, sorry. The short queue should be sufficient indeed.

app/grandchallenge/archives/serializers.py
amickan: Hm, you mean in the call to `create_civ`? I just tried this out, if `user_upload` and `image` are unset (i.e. None) and `value` is `""` or `0`, the or construction would return `""` or `0` respectively, just like we want. Did I misunderstand? 

Intentional `None` as value is a bit more tricky down the line, though. In `create_civ_for_value` we delete any currently existing civ if the new value is `None`. This enables deleting a value from a display set etc. So you wouldn't be able to add a value of None. I thought that that wouldn't ever be necessary, but maybe I'm wrong. Can you give an example of when we would want to do that?

app/grandchallenge/components/tasks.py
amickan: True!

app/grandchallenge/archives/serializers.py
jmsmkn: The JSON (Anything) type allows nulls.

app/grandchallenge/components/tasks.py
amickan: That indeed makes more sense. Again a copy-paste oversight.

app/grandchallenge/archives/serializers.py
amickan: I see. In that case, we want to check whether None is a valid value for the interface and only delete the old CIV if None is not a valid value for the interface. Also noticed that `required` should be set to False for the interface form field for interfaces that allow None (or for boolean interfaces) - fixed that as well.

app/grandchallenge/algorithms/forms.py
```diff
@@ -103,13 +103,16 @@ def __init__(self, *args, algorithm, user, **kwargs):
...
 
         for inp in self._algorithm.inputs.all():
-            self.fields[inp.slug] = InterfaceFormField(
-                instance=inp,
-                initial=inp.default_value,
-                user=self._user,
-                required=(inp.kind != InterfaceKindChoices.BOOL),
-                help_text=clean(inp.description) if inp.description else "",
-            ).field
+            if inp is not None:
```
jmsmkn, line 106: Just for my understanding - when would `None` be returned here?

amickan: Sorry, this is a copy paste error!

jmsmkn (approved):

PR #3158, opened 2024-01-11
Initialize json editor widgets on htmx load

amickan:
The json editor widgets weren't being initilized after htmx content loaded.

### DIFF
--- a/app/grandchallenge/core/static/js/jsoneditor_widget.mjs
+++ b/app/grandchallenge/core/static/js/jsoneditor_widget.mjs
@@ -26,10 +26,17 @@ function initialize_jsoneditor_widget(jsoneditorWidgetID) {
     }
 }
 
-$(document).ready(function () {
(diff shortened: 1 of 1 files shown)


app/grandchallenge/core/static/js/jsoneditor_widget.mjs
```diff
@@ -26,10 +26,17 @@ function initialize_jsoneditor_widget(jsoneditorWidgetID) {
...
+function search_for_jsoneditor_widgets() {
     const jsoneditorWidgets = document.getElementsByClassName("jsoneditorWidget");
-
     for (let jsoneditorWidget of jsoneditorWidgets) {
         initialize_jsoneditor_widget(jsoneditorWidget.dataset.widgetId);
     }
-})
+}
+
+$(document).ready(function () {
```
jmsmkn, line 36:
```suggestion
document.addEventListener("DOMContentLoaded", function(event) {
```

PR #3166, opened 2024-01-16
Default annotation line width

chrisvanrun:
Part of the pitch:
- https://github.com/DIAGNijmegen/rse-roadmap/issues/287

The pretty straightforward addition of `default_annotation_line_width` to workstation configuration.

image


chrisvanrun:
Mhh, this seemingly seems to randomly fail on `make docs`. Waiting for the Djanog tests to finish before giving that specific job a retry.

app/grandchallenge/workstation_configs/models.py
```diff
@@ -291,6 +292,12 @@ class ImageInterpolationType(models.TextChoices):
         help_text="Default color for displaying and creating annotations",
     )
 
+    default_annotation_line_width = PositiveSmallIntegerField(
+        blank=True,
+        null=True,
+        help_text="Default line width for displaying and creating annotations",
```
jmsmkn, line 298: Are there any units to this? mm, voxels, px, cm, mm, in, pt, pc etc etc?

chrisvanrun: It is not obvious from any documentation, but some practical interactions would suggest it is pixels. I've updated the help text.

jmsmkn (approved):

PR #3171, opened 2024-01-18
Add `default_annotation_color` for `Question`s

jmsmkn:
Note that the behaviour is slightly different to `WorkstationConfig.default_annotation_color`, here we do not allow `None` values, so the empty choice is always `""`.

See https://github.com/DIAGNijmegen/rse-roadmap/issues/276#issuecomment-1898643477


chrisvanrun:
Will have a detailed look tomorrow. Do note: https://github.com/DIAGNijmegen/rse-roadmap/issues/276#issuecomment-1898859539

chrisvanrun (approved):
Looks good to me! I'll quickly inspect if `""` is OK for the parsing constructs at CIRRUS. Thanks James!

PR #3172, opened 2024-01-19
Refactor DisplaySetFilesUpdate view

amickan:
This refactors the DisplaySetFilesUpdate view and turns it into a TemplateView. This view is used to return an upload field when a user wants to upload a new file for a file interface on the DisplaySetUpdate view.

This is the last part of the display set refactoring step for the pitch https://github.com/DIAGNijmegen/rse-roadmap/issues/274


app/grandchallenge/components/views.py
```diff
@@ -153,3 +156,33 @@ def get_success_message(self, cleaned_data):
             success_message=self.success_message,
             url=reverse("notifications:list"),
         )
+
+
+class FileUpdateMixin(ObjectPermissionRequiredMixin, TemplateView):
```
jmsmkn, line 161: Including `TemplateView` here makes it a view, not a `Mixin` (a mixin can be used with any class, this class is limited as it cannot be used with any other view). You could either remove the `TemplateView` from this, or move it to views.py as a base view.

app/grandchallenge/components/views.py
```diff
@@ -153,3 +156,33 @@ def get_success_message(self, cleaned_data):
...
+        context = super().get_context_data(*args, **kwargs)
+        context.update(
+            {
+                "form": self.form_class(**self.get_form_kwargs()),
+            }
+        )
+        return context
+
+    def get_form_kwargs(self):
+        return {
```
jmsmkn, line 185: Missing super.

app/grandchallenge/components/views.py
amickan: Ah true, it's already in views.py, but I will change it's name to `FileUpdateBaseView`

app/tests/reader_studies_tests/test_views.py
```diff
@@ -465,6 +470,40 @@ def test_display_set_update(client):
...
+                ci_json.slug: '{"foo": "new"}',
+                ci_img.slug: str(im2.pk),
+                f"WidgetChoice-{ci_img.slug}": WidgetChoices.IMAGE_SEARCH.name,
+                ci_json_file.slug: str(upload.pk),
+                "order": 11,
+            },
+            user=user,
+            method=client.post,
+        )
+    assert response.status_code == 302
```
jmsmkn, line 497: It's a good idea to test which URL you are being directed to too as you could be say being pushed towards the login page.

app/grandchallenge/components/views.py
amickan: Since we're inheriting from a TemplateView super doesn't have the `get_form_kwargs` method. This used to be a FormView, but since we're only using this view to get the correct form, not to post any data to it, I changed it to a TemplateView, just like I did for the InterfacesCreate view in yesterdays PR.

app/grandchallenge/components/forms.py
```diff
@@ -151,8 +165,60 @@ def _get_image_field(self, *, interface, current_value):
...
+                # or the UUID of a new UserUpload object
+                try:
+                    uuid.UUID(current_value)
+                    return self._get_default_field(
+                        interface=interface, current_value=current_value
+                    )
+                except ValueError as e:
+                    if ComponentInterfaceValue.objects.filter(
+                        pk=current_value
+                    ).exists():
```
jmsmkn, line 188: You're looking at **any** CIV here? Thinking data security, should it be filtered by permission or by membership of the object they definitely have permission to view/change?

app/grandchallenge/components/views.py
jmsmkn: Ah, I see, that's a little confusing. Do you need to override `get_form_kwargs` in the subclasses? If not, just throw them in directly.

amickan: I don't need to override the method in the subclass, so will indeed just add them to the form directly. Thanks!

app/grandchallenge/components/forms.py
```diff
@@ -151,8 +165,60 @@ def _get_image_field(self, *, interface, current_value):
...
+                # this happens on initial form load when there already is a CIV for
+                # the interface
+                return self._get_select_upload_widget_field(
+                    interface=interface,
+                    values=values,
+                    current_value=current_value,
+                )
+            else:
+                # on form submit, current_value either is the pk of an existing CIV
+                # or the UUID of a new UserUpload object
```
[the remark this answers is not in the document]
amickan: You mean prepend this to `current_value` itself, or what do you mean by identifiers?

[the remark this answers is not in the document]
jmsmkn: The value of `current_value`.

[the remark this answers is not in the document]
amickan: I think it would be best to pass along another parameter that identifies the widget type. I'll see if that breaks the user upload widget in other places first though

app/grandchallenge/components/forms.py
amickan: Hm, good point. That would be quite an expensive query though, right? I think it's not strictly necessary because we check that the value is part of the queryset (the `values` parameter that we pass to the widget) and we retrieve that from the DB in `base_obj.values_for_interfaces`.

app/grandchallenge/components/forms.py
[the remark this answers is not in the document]
amickan: As far as I can tell this works and doesn't break the upload widget in other places

app/grandchallenge/components/forms.py
amickan: ah yes of course, don't know why I didn't see that earlier!

app/grandchallenge/components/forms.py
```diff
@@ -151,8 +164,63 @@ def _get_image_field(self, *, interface, current_value):
...
+                    if (value and int(value) in values) or not value:
+                        return self._get_select_upload_widget_field(
+                            interface=interface,
+                            values=values,
+                            current_value=ComponentInterfaceValue.objects.get(
+                                pk=value
+                            )
+                            if value
+                            else None,
+                        )
```
jmsmkn, line 195: This is hard to understand now as there are lots of packed clauses, can we try to make it easier on our future selves? Maybe split it out a bit? This is what I think is going on, but I am unsure:


```python
try:
    if int(value) in values:
        current_value = ComponentInterfaceValue.objects.get(
              pk=value
        )
    else:
        # The selected CIV does not belong to the set of values
        raise PermissionDenied
except TypeError:
    current_value = None

return self._get_select_upload_widget_field(
                             interface=interface,
                             values=values,
                             current_value=current_value,
                         )
```

amickan: That is exactly what is going on, except that `int(None)` returns a `ValueError` not a TypeError. Thanks!

jmsmkn: Are you sure?

```
>>> int(None)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```

jmsmkn: Also made a change that puts the exception handling next to the line we expect to find it so that we don't have huge try blocks.

amickan: Ah, sorry for the confusion, since we're parsing `value` from a string with `current_value.split("_")`, `value` won't be `None` but `''` instead, and  `int('')` raises a ValueError.

jmsmkn: Ah okay, then update the comment about it as it's misleading.

jmsmkn (approved):

PR #3185, opened 2024-01-22
Add archive item create view

### LINKED ISSUE #3008: Cannot create new Archive Items from non-image files via the web iterface
Must use the API if you have say mp4 only input.

amickan:
This adds a `ArchiveItemCreateView` using the same underlying functionality as the `DisplaySetCreateView`, as well as the view for interface creation for archive items. 

There is no button for the create view anywhere yet. I will add that in a seperate PR together with an update of the archive item update and list views and their respective templates. 

Closes #3008


app/grandchallenge/archives/views.py
```diff
@@ -696,3 +702,97 @@ def get_serializer_class(self):
...
+    success_message = "Archive item has been created."
+    type_to_add = ArchiveItem._meta.verbose_name
+    base_object_type = Archive._meta.model_name
+
+    def get_permission_object(self):
+        return self.base_object
+
+    @property
+    def base_object(self):
+        return Archive.objects.filter(slug=self.kwargs["slug"]).get()
```
jmsmkn, line 728:
```suggestion
        return Archive.objects.get(slug=self.kwargs["slug"])
```

app/grandchallenge/components/models.py
```diff
@@ -2046,3 +2046,43 @@ def create_civ_for_file(self, *, ci, current_civ, new_value):
...
+        from grandchallenge.reader_studies.models import ReaderStudy
+
+        if self.__class__ == ReaderStudy:
+            qs = self.display_sets.select_related(
+                "values", "values__interface", "values__image"
+            )
+        elif self.__class__ == Archive:
+            qs = self.items.select_related(
+                "values", "values__interface", "values__image"
+            )
```
jmsmkn, line 2064: This is a bit backwards as the Mixin needs to know about all its consumers. The way to do this would be something like:

```python
class Mixin:
  @property
  def related_attr(self): # think of a better name
    raise NotImplementedError
  
  def values_for_interfaces(self):
    queryset = self.related_attr.select_related(
                 "values", "values__interface", "values__image"
             )

...

class Consumer(Mixin):
  @property
  def related_attr(self):
    return self.items # or display_sets
```

app/grandchallenge/components/models.py
```diff
@@ -2046,3 +2046,43 @@ def create_civ_for_file(self, *, ci, current_civ, new_value):
...
+                "values__id",
+            )
+            .order_by("values__id")
+            .distinct()
+        )
+        interfaces = [x["values__interface__slug"] for x in vals]
+        if self.__class__ == ReaderStudy:
+            interfaces += self.questions.filter(
+                interface__isnull=False
+            ).values_list("interface__slug", flat=True)
```
jmsmkn, line 2078: Any modification should happen in a method of the subclass.

app/grandchallenge/components/models.py
```diff
@@ -2046,3 +2046,43 @@ def create_civ_for_file(self, *, ci, current_civ, new_value):
             # if no new value is provided (user selects '---' in dropdown)
             # delete old CIV
             self.values.remove(current_civ)
+
+
+class ValuesForInterfacesMixin:
+    @cached_property
+    def values_for_interfaces(self):
```
jmsmkn, line 2053: This is very complex and is doing a lot of work in python. It seems like it should be possible to offload it to the database but I am unsure. I cannot really see any unit tests for this method so don't know what it needs to return.

app/grandchallenge/archives/views.py
```diff
@@ -696,3 +702,97 @@ def get_serializer_class(self):
...
+    permission_required = (
+        f"{Archive._meta.app_label}.change_{ArchiveItem._meta.model_name}"
+    )
+
+    def get_permission_object(self):
+        return self.object
+
+    @property
+    def object(self):
+        if self.kwargs.get("pk"):
```
jmsmkn, line 776: Why is None allowed here?

app/grandchallenge/archives/views.py
```diff
@@ -696,3 +702,97 @@ def get_serializer_class(self):
...
+        return self.return_url
+
+
+class ArchiveItemInterfaceCreate(InterfacesCreateBaseView):
+    permission_required = (
+        f"{Archive._meta.app_label}.change_{ArchiveItem._meta.model_name}"
+    )
+
+    def get_permission_object(self):
+        return self.object
```
jmsmkn, line 772: `self.object` is `None` when the pk does not exist? How does the permission check work there?

app/tests/archives_tests/test_views.py
```diff
@@ -849,3 +853,113 @@ def test_archive_item_add_value(
...
+    client, settings, django_capture_on_commit_callbacks
+):
+    settings.task_eager_propagates = (True,)
+    settings.task_always_eager = (True,)
+
+    archive = ArchiveFactory()
+    editor = UserFactory()
+    archive.add_editor(editor)
+    ai1 = ArchiveItemFactory(archive=archive)
+    ci_str = ComponentInterfaceFactory(kind="STR")
```
jmsmkn, line 902: Please use the `ImageKind` objects rather than the codes directly.

app/grandchallenge/components/models.py
amickan: Ok, makes sense. And further below instead of  checking `self.__class__ == ReaderStudy` when adding the question interfaces for reader studies, a try ... except statement would be better?
``

jmsmkn: See the other suggestion, I think it's better to be explicit, e.g.

```python

class Mixin:
  @property
  def relevant_interfaces(self): # Again, naming...
    return self.related_attr.select_related(
                 "values", "values__interface", "values__image"
             )

  def values_for_interfaces(self):
    interfaces = self.relevant_interfaces
    ... # Do stuff with the interfaces

class Archive(Mixin):
  pass # we're happy with the default

class ReaderStudy(Mixin):
  @property
  def relevant_interfaces(self):
    interfaces = super().relevant_interfaces
    interfaces += self.questions.filter(...)
    return interfaces
```

amickan: Never mind, got my answer in your other comment

app/grandchallenge/components/models.py
amickan: I can't think of a way to do this in the DB, but I'll add a test so you can see if you do.

app/grandchallenge/archives/views.py
amickan: Very good question. I had just taken this bit over from Mike's code without questioning it since everything worked as expected, but looking into it now, the answer is that the permission check will pass if the permission object is None. That is strange,, but lokks like it is intended to work this way in Django guardian.

For us, I'm actually thinking now that we don't need to protect this view with object permissions since it's just a template view that is used within the UpdateView where we check permissions. Do you agree?

jmsmkn: My guess is that by passing `None` for the object you then do a check against the global permission. I think it should still have permissions attached as you're getting information about an object. Is this the same for a reader study view?

app/grandchallenge/archives/models.py
```diff
@@ -33,7 +34,12 @@
 from grandchallenge.subdomains.utils import reverse
 
 
-class Archive(UUIDModel, TitleSlugDescriptionModel, ViewContentMixin):
+class Archive(
+    UUIDModel,
+    TitleSlugDescriptionModel,
+    ViewContentMixin,
```
jmsmkn, line 40: You will need to merge from main as `ViewContentMixin` -> `HangingProtocolMixin`.

app/grandchallenge/components/views.py
```diff
@@ -182,3 +206,64 @@ def get_context_data(self, *args, **kwargs):
             }
         )
         return context
+
+
+class ObjectCreateMixin:
```
jmsmkn, line 211: `ObjectCreate` I think is too unspecific. I like the terminology of `DisplaySet` as this is what is used in DICOM, and much prefer it over `ArchiveItem`. I think using `DisplaySetCreate` would be too confusing as that has a different concrete class. These are sets of interface values right, so maybe `CIVSetCreate`? What do you think?

app/grandchallenge/components/models.py
```diff
@@ -2046,3 +2046,39 @@ def create_civ_for_file(self, *, ci, current_civ, new_value):
...
+            .filter(values__interface__slug__isnull=False)
+            .values(
+                "values__interface__slug",
+                "values__id",
+            )
+            .order_by("values__id")
+            .distinct()
+        )
+        interfaces = [x["values__interface__slug"] for x in vals]
+        return set(interfaces), vals
```
jmsmkn, line 2071: Please use a named tuple when returning mutiple values and then access them by attribute name.

app/grandchallenge/components/views.py
jmsmkn: Also, why is this separate from `MultipleCIVProcessingBaseView`? It's never used independently, right?

app/grandchallenge/reader_studies/views.py
```diff
@@ -1374,89 +1371,85 @@ def get_success_url(self):
...
     permission_required = (
         f"{ReaderStudy._meta.app_label}.change_{ReaderStudy._meta.model_name}"
     )
     included_form_classes = (
         DisplaySetCreateForm,
-        SingleCIVForm,
+        *MultipleCIVProcessingBaseView.included_form_classes,
     )
     success_message = "Display set has been created."
+    type_to_add = DisplaySet._meta.verbose_name
```
jmsmkn, line 1415: Better to use the model name rather than the verbose name as the latter can change with the locale.

app/grandchallenge/reader_studies/views.py
```diff
@@ -1374,89 +1371,85 @@ def get_success_url(self):
...
-            "pk": self.kwargs.get("pk"),
-            "base_obj": self.reader_study,
-            "interface": self.request.GET.get("interface"),
-            "user": self.request.user,
-            "auto_id": f"id-{uuid.uuid4()}",
-            "htmx_url": self.get_htmx_url(),
-        }
+    @property
+    def base_object(self):
+        return ReaderStudy.objects.get(slug=self.kwargs["slug"])
```
jmsmkn, line 1384: Shouldn't the base object be `self.object.reader_study`?

app/grandchallenge/components/views.py
jmsmkn: We could then call `related_item_model` `civ_sets_field`.

app/grandchallenge/archives/views.py
amickan: Yes, it's the same for the reader study view. Ok, I'll add the permission check again. 

It does not check global permissions because we have `accept_global_perms` set to False. `check_permissions` simply returns `None` and if that happens the view is executed.

jmsmkn: Yes, please ensure permissions are checked in both places and ensure there is a test for this.

app/grandchallenge/components/views.py
amickan: I like your wording, very clear, so `CIVSetCreate` and `civ_sets_related_manager` it is. Thanks!

app/grandchallenge/reader_studies/views.py
amickan: No, because `self.object` can be `None` here,but we still need the reader study base object in that case

app/grandchallenge/reader_studies/views.py
amickan: The only downside is that I'm using this in the templates for the breadcrumbs, where the verbose name would be better or else we end up with "Archiveitems" rather than "Archive Items" etc. I can add another variable to the context that contains the verbose name, or do you see another option?

amickan: Or I add a template tag to retrieve the verbose name of an object. Would avoid cluttering the context of the view. What do you think?

amickan: ah yeah, but then it might change with locale which we probably don't want. Then maybe `type_to_add` and `base_model_name` shouldn't rely on the object's meta information but be supplied as strings on the view instead.

app/grandchallenge/reader_studies/views.py
jmsmkn: Is this bit used for permission checks? If so, you need to be incredibly careful that you do not allow users to be able to modify objects they do not have access to. This often comes from looking up the object via the pk, but then using the URL parameters to get the permission object.

app/grandchallenge/reader_studies/views.py
jmsmkn: Yes template tags to get the information would be the way to go. `object._meta.model_name` is safe to use as an identifier.

amickan:
I'm confused by the test failures. I don't think they are related to the changeset here, but I merged main, and on main they don't fail from what I can see

amickan:
Ah, it'sflaky tests, I will create a seperate PR to fix them. Indeed unrelated to this changeset

app/tests/reader_studies_tests/test_views.py
```diff
@@ -696,6 +696,22 @@ def test_display_set_interfaces_create_permissions(client):
...
+    )
+    assert response.status_code == 200
+
+    response = get_view_for_user(
+        viewname="reader-studies:display-set-new-interfaces-create",
+        client=client,
+        reverse_kwargs={"slug": rs.slug},
+        user=u2,
+    )
+    assert response.status_code == 200
```
jmsmkn, line 713: Why is a normal user getting a 200 response here? They shouldn't have any need to access this view, right?

amickan: That's because the permission object is None here, then everyone passes the permission check for the view. 

Maybe the way to go would be to check change permissions for the reader study rather than the display set. The display set might not exist, but the reader study will always exist when this view is called, so we wouldn't end up with the permission object being None. Does that sound good? Or, we could make both the permission and the permission object dependent on whether there is an object and check reader study change permission when there is not and display set change permission when there is.

jmsmkn: > That's because the permission object is None here, then everyone passes the permission check for the view

Then, if I understand this correctly, that's a serious bug. Anyone could then add display sets to any reader study. This pattern needs to change everywhere where it occurs.

If there is no existing display set then the user needs to have permission to add display sets for that reader study, which may be being done with the change reader study object permission.

If there is an existing display set then the user needs to have permission to add a value to that display set, which may be being done with the change display set object permission.

amickan: We're talking about the InterfaceCreate view here, which is a TemplateView, no adding or updating is done through that view. We do check permissions correctly on the create and update views. 

I will make the changes to the permission handling for the interface view as well though.

amickan: Permission checks on the interface views should now be complete. I have also added a more comprehensive test.

jmsmkn: Thanks!

PR #3188, opened 2024-01-22
Add more view item validation

### LINKED ISSUE #3183: Add view content, reader study and question validation for non-image types
- [x] Add validation to view-content to ensure that each view port contains at most 1 item of the kinds Chart, PDF, MP4, and Thumbnail.
- [x] Add checks in GC to ensure that viewports configured to display these kinds, are not selectable as viewports for reader study questions.

See https://github.com/DIAGNijmegen/rse-roadmap/issues/282

jmsmkn:
This removes `ViewContentMixin` as there was a lot of duplication, we should probably add a `ValidatedJSONField(schema=...)` that sets the validator and the default widget.

Closes #3183


jmsmkn:
Some tests are failing as the reader study is being set on the question only in `form_valid`, this needs to be changed so it is set in the form instead.

amickan (approved):

PR #3189, opened 2024-01-22
Add instant verification field to reader study options

HarmvZ:
Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/289


app/grandchallenge/reader_studies/forms.py
```diff
@@ -165,6 +166,18 @@ def clean(self):
...
             )
 
+        if (
+            self.cleaned_data["instant_verification"]
+            and not self.cleaned_data["is_educational"]
+        ):
+            self.add_error(
+                error=ValidationError(
+                    "Reader study must be educational when instant verification is enabled.",
+                    code="required",
```
jmsmkn, line 176: That code is too broad, it is not used anywhere else so can be dropped.

```suggestion
```

jmsmkn (approved):

PR #3192, opened 2024-01-23
Remove the Retina API

jmsmkn:
See https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/236


HarmvZ (approved):
What's the strategy for removing the annotations and registrations models?

jmsmkn:
> What's the strategy for removing the annotations and registrations models?

Once we have deployed this and nothing new will be created we will export a dump of the database tables, then create a migration to remove those tables and deploy.

PR #3196, opened 2024-01-24
Add support for OCI container images

### LINKED ISSUE #3195: Importing of containers exported with Docker 25.0.0 fails
**Describe the bug**

I tried to import a viewer on grand-challenge right now, and it does not work:

image

The failed images after the green, imported one are the ones I tried to upload:

- 1&2 are builds and clean re-builds of a dev branch
- 3 is a build of the CIRRUS main branch
- 4 is a known-good downloaded image from aws codebuild

None of them are importing right now, all fail with:

~~~~~~~~~~~~
Validation Errors
    blobs/sha256/e73667bff9013e92f65a7074744cbabffd2ce629e3da19adab4.json not found at the root of the container image file. Was this created with docker save?
~~~~~~~~~~~~

I manually opened the uploaded containers and the blobs are there. The files are also importing fine into docker. So I think that this time I have an actual grand-challenge issue at hand!

**To Reproduce**

Take a known-good image from aws-code build and try to upload to a viewer on grand-challenge.

**Expected behavior**

The image imports sucessfully.

jmsmkn:
Closes #3195


pkcakeout:
I suspect that this will do the trick. Should I test it for you? Or I could give you a tiny exported test-image to write a test against? Would that help you?

pkcakeout:
Here... was super easy actually:

[bare-oci.tar.zip](https://github.com/comic/grand-challenge.org/files/14039025/bare-oci.tar.zip)

Zip contains a docker-image that you can import... it is only 20 KB in size!

jmsmkn:
Actually, I need to wait until I have Docker 25 to do this, too many changes. It's not been released for macos desktop yet.

app/tests/components_tests/test_tasks.py
```diff
@@ -422,3 +425,22 @@ def test_add_file_to_object(settings, object_type):
...
     )
     assert ComponentInterfaceValue.objects.filter(interface=ci).count() == 1
+
+
+@pytest.mark.parametrize(
+    "container_image_file",
+    ("hello-scratch-oci.tar.gz", "hello-scratch-docker-v2.tar.gz"),
+)
+@pytest.mark.django_db
+def test_get_image_config_and_sha256(container_image_file):
```
amickan, line 435: Should there be tests for all the ValidationErrors raised by that and the related methods?

jmsmkn: I don't think that it's worth creating corrupt container image files for this.

amickan: Fair enough!

amickan (approved):

Issue #3169, opened 2024-01-17
Bulk delete option for display sets 


amickan:
Request from Colin through teams: https://teams.microsoft.com/l/message/19:6daf914d105b412989baaffa5b066dfa@thread.tacv2/1705501050956?tenantId=b208fe69-471e-48c4-8d87-025e9b9a157f&groupId=97a88c45-447f-4147-9e80-5e7c013f7501&parentMessageId=1705501050956&teamName=DIAG&channelName=Grand%20Challenge&createdTime=1705501050956&allowXTenantAccess=false

We allow the user to delete display sets, but don't have a bulk delete option through the UI. It would be nice to have a button to delete all display sets of a reader study. 

This should only be possible for display sets without answers.

Do we want the same for archive items? We currently don't allow deleting of archive items at all.

jmsmkn:
> Do we want the same for archive items? We currently don't allow deleting of archive items at all.

I think we want the two things to behave in the same way, so yes.

snorthman:
The gcapi documentation is no longer valid. https://grand-challenge.org/documentation/gc-api-for-reader-studies/ (at the bottom):

```python
display_sets = client.reader_studies.display_sets.iterate_all(
    params={"reader_study": rs["pk"]}
)
while True:
    ds = next(display_sets)
    client.reader_studies.display_sets.delete(ds["pk"])
```
example:

```python
client = gcapi.Client(token=gcapi_token)
client.reader_studies.display_sets.delete('6e1e49b8-bc36-40eb-98fe-25bc90d934a2')

httpx.HTTPStatusError: Client error '405 Method Not Allowed' for url '[https://grand-challenge.org/api/v1/reader-studies/display-sets/6e1e49b8-bc36-40eb-98fe-25bc90d934a2/'](https://grand-challenge.org/api/v1/reader-studies/display-sets/6e1e49b8-bc36-40eb-98fe-25bc90d934a2/%27)
For more information check: https://httpstatuses.com/405
```

Any alternative way to bulk delete display_sets via gcapi? @amickan

amickan:
At the moment there is no alternative, sorry, but I'm hoping to get to this this cycle.

chrisvanrun:
I see that the docs have also been updated. Jeej! 🥳

PR #3198, opened 2024-01-24
Enable archive item deletion through the API

amickan:

After merging this, permissions for existing archive items need to be updated by running `assign_permissions()` on all items.

### DIFF
--- a/app/tests/archives_tests/test_views.py
+++ b/app/tests/archives_tests/test_views.py
@@ -849,3 +849,25 @@ def test_archive_item_add_value(
             )
     assert response.status_code == 200
     assert ArchiveItem.objects.get().values.first().value
+
+
+@pytest.mark.django_db
+def test_api_archive_item_deletion(client):
+    archive = ArchiveFactory()
+    editor = UserFactory()
+    archive.add_editor(editor)
+    i1, i2 = ArchiveItemFactory.create_batch(2, archive=archive)
+
+    assert ArchiveItem.objects.count() == 2
+
+    response = get_view_for_user(
+        viewname="api:archives-item-detail",
+        reverse_kwargs={"pk": i1.pk},
+        user=editor,
+        client=client,
+        method=client.delete,
+    )
+    assert response.status_code == 204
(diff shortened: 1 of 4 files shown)


jmsmkn:
Why does this need to be an API endpoint? I would have thought that the normal view would be sufficient. That way, you can get confirmation of the consequences, I don't really see the need to automate this from another application.

app/tests/archives_tests/test_views.py
```diff
@@ -849,3 +849,25 @@ def test_archive_item_add_value(
             )
     assert response.status_code == 200
     assert ArchiveItem.objects.get().values.first().value
+
+
+@pytest.mark.django_db
+def test_api_archive_item_deletion(client):
```
jmsmkn, line 855: Needs tests for object based permissions, e.g.
- ensuring that a user can only delete an archive item if they have the "delete_archite_item" bit 
- ensuring that the only users and groups with the "delete_archite_item" bit are the editors group for this reader study
See usages of `get_groups_with_set_perms`

amickan:
> Why does this need to be an API endpoint?

I went with the API view since we discussed wanting a bulk delete option in the future (https://github.com/comic/grand-challenge.org/issues/3169). And this is also how display set deleting is done at the moment (although that's not a reason to use an API view per se of course).

jmsmkn:
A bulk delete should be a single view, you shouldn't implement bulk deletion by iterating over individual deletion endpoints, that would take forever.

jmsmkn:
> The latter is not a reason to use an API view per se of course, but if we go for a normal delete view for archive items, we should do that for display sets as well, or else it'll be difficult to use the same templates etc. across the apps

I would rather use a normal view for both. API views are only for things that need to go into other apps (e.g. CIRRUS, user upload scripts via GC API).

amickan:
> I would rather use a normal view for both.

Alright. I understand. I will add normal delete views for both then.

amickan:
I will create a new branch for the delete views.

PR #3200, opened 2024-01-25
Add display set and archive item delete views

amickan:

Once this is merged, the archive item permissions need updating by running `assign_permissions` for all existing archive items.


app/grandchallenge/core/templatetags/meta_attr.py
```diff
@@ -7,3 +7,18 @@
...
+
+
+@register.filter
+def verbose_name_plural(obj):
+    return obj._meta.verbose_name_plural
+
+
+@register.filter
+def model_name(obj):
+    return obj._meta.model_name
```
amickan, line 24: This is the same as in the other open PR. Depending on which PR gets merged first, I will update the remaining PR.

app/grandchallenge/archives/views.py
```diff
@@ -798,3 +799,10 @@ def get_htmx_url(self):
                 "archives:item-new-interface-create",
                 kwargs={"slug": self.base_object.slug},
             )
+
+
+class ArchiveItemDeleteView(CIVSetDeleteView):
+    model = ArchiveItem
+    permission_required = (
+        f"{Archive._meta.app_label}.change_{ArchiveItem._meta.model_name}"
```
jmsmkn, line 807: Should be `delete_{ArchiveItem._meta.model_name}`

jmsmkn (commented):
Can the delete display set API view also be removed?

app/tests/components_tests/test_views.py
```diff
@@ -190,3 +191,75 @@ def test_interfaces_create_permissions(
...
+    response = get_view_for_user(
+        viewname=viewname,
+        client=client,
+        user=collaborator,
+        reverse_kwargs={"slug": base_obj.slug, "pk": obj.pk},
+    )
+    if base_object_factory == ReaderStudyFactory:
+        assert response.status_code == 403
+    elif base_object_factory == ArchiveFactory:
+        assert response.status_code == 200
```
jmsmkn, line 243: Why do uploaders get to delete? I would have thought they would only be able to upload.

amickan:
> Can the delete display set API view also be removed?

It's no longer used on the website, so yes, could be removed, but there might be users out there that are using the delete endpoint through the API in their own scripts, so I wasn't sure.

app/tests/components_tests/test_views.py
amickan: I wasn't sure about this. My reasoning was the whoever can create and edit archive items should also be able to delete them. Both editors and uploaders can create and edit. But I'm also fine with restricting the delete view to editors.

app/tests/components_tests/test_views.py
```diff
@@ -190,3 +191,75 @@ def test_interfaces_create_permissions(
...
+        client=client,
+        method=client.post,
+        user=editor,
+        reverse_kwargs={"slug": base_obj.slug, "pk": obj.pk},
+    )
+    assert response.status_code == 302
+    if base_object_factory == ReaderStudyFactory:
+        object_count = DisplaySet.objects.count()
+    elif base_object_factory == ArchiveFactory:
+        object_count = ArchiveItem.objects.count()
```
jmsmkn, line 264: Try to avoid if statements in tests, they should be doing straightforward things. In this case you can check both:

```
assert DisplaySet.objects.count() == 0
assert ArchiveItem.objects.count() == 0
```

Further up you could remove the if statement with a `setattr(base_obj, add_collaborator_attr, collaborator)`, moving add_collaborator_attr to the parameterisation ("add_reader", etc.)

jmsmkn:
> It's no longer used on the website, so yes, could be removed, but there might be users out there that are using the delete endpoint through the API in their own scripts, so I wasn't sure.

It can go.

app/tests/components_tests/test_views.py
jmsmkn: I always saw uploaders as system users who were pushing data from external systems. I think that it is fine not to give them delete permission.

jmsmkn (approved):

PR #3202, opened 2024-01-25
Catch exceptions unrelated to validation when adding files to an object

### LINKED ISSUE #3199: Updating or creating new display sets and adding non-image component interface values fails silently
**Describe the bug**

When I do either of these:
- create a new display set
- edit an existing display set
- delete all display sets and recreate all of them

Non-image component interface values are silently rejected.

**To Reproduce**

1. Go to a reader study
2. Create a new display set by clicking "Add Display Set (Single)"
3. Add and image, `fixed-image` and `generic-medical-image` are the ones I tested. Upload data.
4. Add a `text-report` (and some thumbnail, and some video - if you want) and upload valid data.
5. Click "save"

After some time, the images appear as part of the new display set. But the text-report, and video, and thumbnails do not. A message informs the uploader that they should check the notifications for errors, but no notifications are generated.

**Expected behavior**

Either an error is communicated or the upload succeeds (as they did before).

amickan:
Closes #3199


app/tests/components_tests/test_tasks.py
```diff
@@ -422,3 +423,49 @@ def test_add_file_to_object(settings, object_type):
...
+    with pytest.raises(RandomException):
+        add_file_to_object(
+            app_label=obj._meta.app_label,
+            model_name=obj._meta.model_name,
+            object_pk=obj.pk,
+            user_upload_pk=us.pk,
+            interface_pk=ci.pk,
+        )
+
+    notification = Notification.objects.last()
```
amickan, line 469: This test fails and I cannot work out why. I stepped through the code and verified that the notification is created, but it is not visible in the test. I think this has to do with different DB transactions, but I cannot figure out what needs to change in the test for the notification to be visible.

jmsmkn: I can't immediately see it. Honestly, I don't think we really need to solve this issue, we have the error in Sentry and investigate each occurrence anyway. Probably best to move on and close this PR as it's not straightforward.

amickan: Ok. I agree.  It shouldn't happen very often anyway. And if we encounter issues more frequently, we can revisit this.

Issue #2871, opened 2023-06-08
Cannot remove a value from an archive item


jmsmkn:
If a user adds a value to an archive item it can be changed, but not removed. They should be able to do this.

jmsmkn:
Now possible through the django admin a3c1497

amickan:
This is now partially possible, i.e. for values and files, not for images yet. I will add an issue for image removal.

Issue #2869, opened 2023-06-07
Unclear where to see the Archive Item or Image UUID


jmsmkn:
Users have trouble setting up their archive and often need to remove images and/or archive items. Finding the correct UUID is often difficult. They should be more obvious on the archive pages, possibly an Archive Item detail view is also needed.

chrisvanrun:
Same holds for reader studies / display sets. Can't even copy the uid from the overview because the CSS doesn't allow for that. Rather you need to go into the HTML itself to copy it.

chrisvanrun:
We could be inspired by the way GitHub goes about it by adding a nice way to copy it, but that would require some JS magic so maybe plain text is safer:
![image](https://github.com/comic/grand-challenge.org/assets/7871310/69588953-f04d-4e93-a030-cff0b85b1aab)

pkcakeout:
Not quite sure if we are tagging on too much stuff now, but I also found a problem:

![image](https://github.com/comic/grand-challenge.org/assets/11632379/5eb661c8-570f-4aa5-b3b6-5ec51eb608ac)

If you want to cherry-pick a single item, to add to a reader-study, it is a lot of work! It always is a click-party _or_ you need to hold backspace until all images are deleted.

I think this page would better be served by a checkmark + list view. That allows one to use `Ctrl+F` to find an item and then click the single item. Add a select all/invert selection/deselect all button, and it would be awesome, I think. I am daring to copy paste an example from wordpress here, which I think better when it comes to this feature:

![image](https://github.com/comic/grand-challenge.org/assets/11632379/0b0b4d2b-953e-4b68-8128-61f252a9dcb6)

(that comes from my wordpress plugin, but I think this is a great design for this function. (notice the search-users field in the top right as well!)

pkcakeout:
Ok, nvm then! I hid my post to reduce noise!

PR #3210, opened 2024-01-26
Archive item update and list views

amickan:
This updates the archive item update and list views, using the same underlying code and templates as the respective display set views. 

This also removes the ArchiveCasesList view which used to show a list of images in an archive (with associated meta information). 

Closes https://github.com/comic/grand-challenge.org/issues/2869
Closes https://github.com/comic/grand-challenge.org/issues/2871


app/tests/archives_tests/test_views.py
```diff
@@ -731,19 +732,17 @@ def test_archive_item_add_image(
...
                     ci_img.slug: image.pk,
                     f"WidgetChoice-{ci_img.slug}": WidgetChoices.IMAGE_SEARCH.name,
                 },
             )
-    assert response.status_code == 200
-    assert item.values.first().image.pk == image.pk
-    assert item.values.first() != old_civ
+    assert response.status_code == 302
+    assert item.values.filter(interface=ci_img).get().image.pk == image.pk
+    assert item.values.filter(interface=ci_img).get() != old_civ_img
```
jmsmkn, line 745: Watch out for this unnecessary pattern, seen it a few times now.

```suggestion
    assert item.values.get(interface=ci_img).image.pk == image.pk
    assert item.values.get(interface=ci_img) != old_civ_img
```

app/grandchallenge/components/views.py
```diff
@@ -288,3 +283,80 @@ def get_success_url(self):
...
         return f"{self.object._meta.verbose_name.title()} was successfully deleted"
+
+
+class BaseModelOptions(TextChoices):
+    ARCHIVE = Archive._meta.model_name
+    READER_STUDY = ReaderStudy._meta.model_name
+
+
+class CivSetListView(
+    LoginRequiredMixin, ObjectPermissionRequiredMixin, PaginatedTableListView
```
jmsmkn, line 294: Why are we using object permissions for a list view?

amickan: We're checking if the user has change/use permission to the base object (archive or reader study) here. I took this over from how the list views used to be implemented since that made sense to me. Would it be better to use `PermissionListMixin` and check for permissions on each listed object?

jmsmkn: Yes, absolutely!

jmsmkn: It should always be this way btw, that way we control our rules about what users can do in one place (`assign_permissions`) and then they views do not have any logic about who can do what with what.

app/grandchallenge/reader_studies/views.py
```diff
@@ -379,65 +379,23 @@ class ReaderStudyStatistics(
...
 
 
-class ReaderStudyDisplaySetList(
-    LoginRequiredMixin, ObjectPermissionRequiredMixin, PaginatedTableListView
-):
+class ReaderStudyDisplaySetList(CivSetListView):
     model = DisplaySet
     permission_required = (
-        f"{ReaderStudy._meta.app_label}.change_{ReaderStudy._meta.model_name}"
+        f"{ReaderStudy._meta.app_label}.change_{DisplaySet._meta.model_name}"
```
jmsmkn, line 385: Should be view permission, no?

```suggestion
        f"{ReaderStudy._meta.app_label}.view_{DisplaySet._meta.model_name}"
```

app/tests/components_tests/test_views.py
```diff
@@ -261,3 +263,84 @@ def test_civset_delete_view(
...
+
+    response = get_view_for_user(
+        viewname=viewname,
+        client=client,
+        user=collaborator,
+        reverse_kwargs={"slug": base_obj.slug},
+    )
+    assert response.status_code == 200
+    assert (
+        len(response.context["object_list"]) == collaborator_visible_obj_count
```
jmsmkn, line 334: Maybe add some civ sets for other objects to make sure that they're not visible too, then we know the filtering is hooked up correctly.

app/grandchallenge/reader_studies/views.py
amickan: Then readers would be able to see the cases, which I don't think we want, do we? At the moment, we check for the change reader study permission, so readers don't get access to this view.

jmsmkn: Ah good point! That is something to be careful about and is unexpected for a list view. Could you add that as a comment and make sure it is covered with a test?

app/grandchallenge/components/views.py
```diff
@@ -288,3 +286,77 @@ def get_success_url(self):
...
+
+    @cached_property
+    def base_object(self):
+        return NotImplementedError
+
+    def render_row(self, *, object_, page_context):
+        return render_to_string(
+            self.row_template,
+            context={**page_context, "object": object_},
+        ).split("<split></split>")
```
jmsmkn, line 334: How does this differ from the implementation in `PaginatedTableListView`?

app/grandchallenge/components/views.py
```diff
@@ -288,3 +286,77 @@ def get_success_url(self):
...
+    columns = [
+        Column(title="Values"),
+        Column(title="View"),
+        Column(title="Edit"),
+        Column(title="Remove"),
+    ]
+
+    def __init__(self):
+        if self.extra_columns:
+            self.columns = [*self.extra_columns, *self.columns]
```
jmsmkn, line 324: This is mixing class and instance properties. Rather than adding `extra_prefetch_fields` you could remove that altogether and I think do something like:

```python
class ReaderStudyDisplaySetList(CivSetListView):
   columns = [
         *CivSetListView.columns,
         Column(title="DisplaySet ID", sort_field="pk"),
         Column(title="Order", sort_field="order"),
     ]
```

That way, `columns` remains a Class property of `ReaderStudyDisplaySetList` rather than an instance property - there may be some assumptions of that (we've run into bugs before due to confusion there).

app/grandchallenge/components/views.py
amickan: Doesn't differ at all and can go, sorry, didn't realize the duplication there when creating this base view.

app/grandchallenge/components/views.py
amickan: Ah ok, that makes sense for columns. For get_queryset with the prefetch fields, I'm not sure yet what's best. To avoid calling instance properties in the base view, I think it would actually be best to move the entire `get_queryset` method to the instance and set a property `prefetch_fields` on the base view to call from the instance method. Would you agree?

If we leave the queryset method on the base view (doing the select_related and the prefetching of the common fields in the base view and then add an extra prefetch_related in the instance method), we'd add an extra db call, which would be less efficient db-query wise than doing it all in one go from what I understand. So seems to me that leaving `get_queryset` to the instances is better, even if that means a bit of code duplication.

jmsmkn: This is about mixing of class and instance properties - the way you have it now means that columns is sometimes a class attribute, sometimes an instance attribute. It means the class attribute is no longer safe to use. So, it means either revert to having columns as a class property, or to turn everything into an instance property. The former is less work.

I think the new way is clearer as there is less indirection, but there is no need to call `CivSetListView` in an instance method, you can use `super()` here: `fields_to_prefetch = super().prefetch_fields`.

One thing you could consider is doing:

```python
class CivSetListView:
  def get_queryset(self):
    queryset = super().get_queryset()
    return queryset.prefetch_related("values", "values__image", "values__interface")

class ReaderStudy:
  def get_queryset(self):
    queryset = super().get_queryset()
    return queryset.prefetch_related("answers").filter(reader_study=self.base_object).select_related("reader_study")
```

I think that works but I'm unsure.

amickan: > One thing you could consider is doing:

Hm, I'm not sure I like this better than what I have now. I feel like it's clearer if there is only one place of `get_queryset` to look at rather than two. Also, I think this way would be slightly less efficient, right? We'd be doing an unnecessary extra DB query, plus we're calling select_related after the prefetch_related call...

amickan: > So, it means either revert to having columns as a class property, or to turn everything into an instance property.

Now I'm a bit confused. I thought that in my current implementation `columns` _is_ a class rather than an instance property of the list views. They're defined directly in the class definition, not in the init method any longer. Am I misunderstanding your comment?

jmsmkn: > Hm, I'm not sure I like this better than what I have now. I feel like it's clearer if there is only one place of get_queryset to look at rather than two. 

This is more like how inheritance should work. With it as it is now the consumers of the View need to know about the custom internal expectations (they have to call `prefetch_related` with an attribute). With what I propose, the consumers don't care, and are only concerned with their own concrete implementation of the view. This allows for looser coupling.

> We'd be doing an unnecessary extra DB query, plus we're calling select_related after the prefetch_related call

AFAIK that shouldn't matter as selects and prefetches get bundled. But maybe they don't? You can check with the `.query` attribute of the queryset.


> Now I'm a bit confused.

Sorry, I wrote the comment before realising that you had made changes. That part looks good now.

amickan: > AFAIK that shouldn't matter as selects and prefetches get bundled. But maybe they don't? You can check with the `.query` attribute of the queryset.

Looks like you are right. The generated queries are identical. I will update the code to use what you proposed.

app/grandchallenge/archives/views.py
```diff
@@ -365,236 +363,82 @@ def get_context_data(self, **kwargs):
...
-                "files",
-                "componentinterfacevalue_set__algorithms_jobs_as_input__algorithm_image__algorithm",
-            )
-            .select_related(
-                "origin__creator__user_profile",
-                "origin__creator__verification",
-            )
+            queryset.filter(archive=self.base_object)
+            .select_related("archive")
+            .distinct()
```
jmsmkn, line 441: Why is a `distinct` needed here? This does make the query slower.

jmsmkn: Maybe for the search? This is just for my understanding.

amickan: I thought that since our query involves prefetching a many-to-many relationship, there is a chance of getting duplicate rows. A call to distinct() does make the query slower though and I cannot produce issues with the query right now without distinct, and examples I can find of this issue all involve filtering, so maybe this is not an issue with prefetching after all.

jmsmkn (approved):

PR #3228, opened 2024-02-01
Add data migration for archive item delete permission

amickan:


app/grandchallenge/archives/migrations/0017_archive_item_delete_permission_granting.py
```diff
@@ -0,0 +1,35 @@
...
+    Archive = apps.get_model("archives", "Archive")  # noqa: N806
+    ArchiveItem = apps.get_model("archives", "ArchiveItem")  # noqa: N806
+
+    archives = (
+        Archive.objects.order_by("-created")
+        .prefetch_related("editors_group", "items")
+        .all()
+    )
+    for batch in archives.iterator(chunk_size=100):
+        for archive in batch:
```
jmsmkn, line 15:
```suggestion
    for archive in archives.iterator(chunk_size=100):

```

Was this tested?

app/grandchallenge/archives/migrations/0017_archive_item_delete_permission_granting.py
```diff
@@ -0,0 +1,35 @@
+from django.db import migrations
+from guardian.shortcuts import assign_perm
+
+
+def add_archive_item_delete_permissions(apps, schema_editor):
+    Archive = apps.get_model("archives", "Archive")  # noqa: N806
+    ArchiveItem = apps.get_model("archives", "ArchiveItem")  # noqa: N806
```
jmsmkn, line 7: ArchiveItem is unused.

jmsmkn: Ah, it is.

amickan:
I found this while looking for solutions: https://sleepy.yaks.industries/posts/set-permissions-django-migrations/. So adding the permission in the migration appears to be the way to go.

app/grandchallenge/archives/migrations/0017_archive_item_delete_permission_granting.py
```diff
@@ -0,0 +1,47 @@
...
+        ArchiveItemGroupObjectPermission.objects.bulk_create(
+            objs=[
+                ArchiveItemGroupObjectPermission(
+                    content_object=item,
+                    group=archive.editors_group,
+                    permission=delete_permission,
+                )
+                for item in archive.items.all()
+            ],
+            ignore_conflicts=True,
```
amickan, line 33: If the DB is not empty and there already are items with the given permission (which can happen because new items get the delete permission assigned at creation), this will fail with an integrity error. Ignoring those conflicts does the trick, although it feels a bit strange to do.

jmsmkn: For my understanding - does that mean that only the conflicting rows are ignored? It doesn't ignore the entire batch?

amickan: From the documentation `On databases that support it (all but Oracle), setting the ignore_conflicts parameter to True tells the database to ignore failure to insert any rows that fail constraints such as duplicate unique values.` 

To me his reads as only the conflicting rows will be ignored, but I will do some tests to verify

amickan: Only the failing rows are ignored

app/grandchallenge/archives/migrations/0017_archive_item_delete_permission_granting.py
```diff
@@ -0,0 +1,47 @@
...
+        "archives", "ArchiveItemGroupObjectPermission"
+    )
+    Permission = apps.get_model("auth", "Permission")  # noqa: N806
+
+    queryset = Archive.objects.order_by("-created").prefetch_related(
+        "editors_group", "items"
+    )
+    for app_config in apps.get_app_configs():
+        app_config.models_module = True
+        create_permissions(app_config, verbosity=0)
```
jmsmkn, line 18: This seems strange. Would it be better to check `if ArchiveItem.objects.exists()`? Then we know the data migration needs to do work and is being done in a non-empty db.

app/grandchallenge/archives/migrations/0017_archive_item_delete_permission_granting.py
```diff
@@ -0,0 +1,47 @@
...
+    delete_permission = Permission.objects.get(
+        codename=f"delete_{ArchiveItem._meta.model_name}",
```
jmsmkn, line 20: I think that we're missing the check on the content type here, should be something like `content_type__app_name="archives"`.

app/grandchallenge/archives/migrations/0017_archive_item_delete_permission_granting.py
amickan: True, that also works!

app/grandchallenge/archives/migrations/0017_archive_item_delete_permission_granting.py
```diff
@@ -0,0 +1,45 @@
...
+    ArchiveItem = apps.get_model("archives", "ArchiveItem")  # noqa: N806
+    ArchiveItemGroupObjectPermission = apps.get_model(  # noqa: N806
+        "archives", "ArchiveItemGroupObjectPermission"
+    )
+    Permission = apps.get_model("auth", "Permission")  # noqa: N806
+
+    queryset = Archive.objects.order_by("-created").prefetch_related(
+        "editors_group", "items"
+    )
+    if queryset:
```
jmsmkn, line 15: Doing `if queryset` will execute the queryset and fetch all the archives and archive items, will be very slow as it would fetch a tonne of data from the db at once. Using `exists()` will be much faster and keeps the advantage of doing the iterator later:

```suggestion
    if queryset.exists():

```

jmsmkn (approved):

PR #3233, opened 2024-02-05
Remove IMAGE .xy help with one that generates a .gz

chrisvanrun:


jmsmkn:
SNAP! #3234 also includes a change to the first line of the help text.

PR #3248, opened 2024-02-14
Add one click unsubscribe view

amickan:
This adds a `SubscriptionUpdate` view that can be used to update notification and newsletter email preferences. 

To enable one-click unsubscribe functionality, we subclass this view for both notifications and newsletters, set the initial form values accordingly and auto-submit the form. 

On success, the requester gets redirected to a subscription overview page with the same form, but this time without automatically submitting the form. image

Note that this does not yet add the unsubscribe link and necessary headers to the emails. That will be done in a seperate PR.

Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/295


app/tests/profiles_tests/test_views.py
```diff
@@ -179,3 +180,74 @@ def test_organization_update(self, client):
...
+
+@pytest.mark.parametrize(
+    "viewname,subscription_attr",
+    (
+        ("newsletter-unsubscribe", "receive_newsletter"),
+        ("notification-unsubscribe", "receive_notification_emails"),
+    ),
+)
+@pytest.mark.django_db
+def test_one_click_unsubscribe(client, settings, viewname, subscription_attr):
```
jmsmkn, line 193: It's Django functionality by we should also check that modifying the token (e.g. changing the username) does not allow users to (un)subscribe other users.

app/tests/profiles_tests/test_views.py
```diff
@@ -179,3 +180,74 @@ def test_organization_update(self, client):
...
+        client=client,
+        viewname=viewname,
+        reverse_kwargs={"token": token},
+        method=client.post,
+    )
+    assert response.status_code == 302
+    user.user_profile.refresh_from_db()
+    assert not getattr(user.user_profile, subscription_attr)
+
+    # reset preferences
```
jmsmkn, line 216: Try not to create long tests unless necessary. Debugging very long tests can hide errors. You should try for the setup - modify - assert pattern. Some people go to extremes and only have one assert per test but I think there is a halfway house. That you have 3 comments in here, and that you have a reset, suggests that there should be 3 separate tests. Have a look at some of the things Adam Johnson has put out (e.g. https://adamj.eu/tech/2020/06/15/how-to-unit-test-a-django-form/).

app/tests/profiles_tests/test_views.py
```diff
@@ -179,3 +180,74 @@ def test_organization_update(self, client):
...
+    (
+        ("newsletter-unsubscribe", "receive_newsletter"),
+        ("notification-unsubscribe", "receive_notification_emails"),
+    ),
+)
+@pytest.mark.django_db
+def test_one_click_unsubscribe(client, settings, viewname, subscription_attr):
+    # Override the celery settings
+    settings.task_eager_propagates = (True,)
+    settings.task_always_eager = (True,)
```
jmsmkn, line 196: Why is Celery required for the test?

amickan: Because we are creating/updating a verification user set if there is a user mismatch and I'm calling the asynchronous task `update_verification_user_set` to do that.

app/grandchallenge/profiles/views.py
```diff
@@ -197,3 +204,66 @@ def form_invalid(self, form):
...
+            }
+        )
+        return kwargs
+
+    def form_valid(self, form):
+        self.user_to_unsubscribe = form.cleaned_data["user"]
+        self.user_to_unsubscribe.user_profile.unsubscribe(
+            subscription_type=self.subscription_type
+        )
+        return super().form_valid(form)
```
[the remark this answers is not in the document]
amickan: Making this an UpdateView and using a ModelForm hadn't occurred to me. That makes it a bit more straight-forward. I will try that out. Thanks for the suggestion

app/grandchallenge/profiles/templates/profiles/subscription_form.html
```diff
@@ -0,0 +1,16 @@
+{% extends "base.html" %}
+{% load static %}
+{% load crispy_forms_tags %}
+
+
+{% block content %}
+    <h2>Subscriptions</h2>
+    <p>Your current subscription preferences are as follows:</p>
+    {% csrf_token %}
```
jmsmkn, line 9: Just for my understanding - I thought that crispy included the csrf token?

app/grandchallenge/profiles/views.py
```diff
@@ -197,3 +206,94 @@ def form_invalid(self, form):
...
+            not get_user_model()
+            .objects.filter(username__iexact=token, is_active=True)
+            .exists()
+        ):
+            return False
+
+        return True
+
+    def validate_token(self):
+        return Signer(salt=UNSUBSCRIBE_SALT).unsign(self.kwargs.get("token"))
```
jmsmkn, line 234: In `test_func` the result of this method is called `token` when it is returning a username. Might make the usage a little clearer?

```suggestion
    @property
    def username_from_token(self):
        return Signer(salt=UNSUBSCRIBE_SALT).unsign(self.kwargs.get("token"))
```

app/grandchallenge/profiles/views.py
```diff
@@ -197,3 +206,94 @@ def form_invalid(self, form):
         # and display an error message
         messages.add_message(self.request, messages.ERROR, "Incorrect token.")
         return response
+
+
+class SubscriptionUpdate(SuccessMessageMixin, UserPassesTestMixin, UpdateView):
+    model = UserProfile
+    form_class = SubscriptionPreferenceForm
+    template_name = "profiles/subscription_form.html"
+    success_message = "You successfully updated your subscription preferences."
```
jmsmkn, line 215: Could we add information in the success message about what they have changed their preferences to? Something like:

`"You successfully updated your email preferences. You will no longer receive newsletter/notification emails. You will continue to receive newsletter/notification emails."`

app/grandchallenge/profiles/views.py
```diff
@@ -197,3 +206,94 @@ def form_invalid(self, form):
         # and display an error message
         messages.add_message(self.request, messages.ERROR, "Incorrect token.")
         return response
+
+
+class SubscriptionUpdate(SuccessMessageMixin, UserPassesTestMixin, UpdateView):
```
jmsmkn, line 211: Should this be subscription update or email preferences? It still keeps people subscribed to notifications and forum tracks, just stops sending them emails.

jmsmkn:
Looks good! Much easier to understand using update views and model forms.

app/grandchallenge/profiles/views.py
amickan: Good point! `EmailPreferencesUpdate` is better!

app/grandchallenge/profiles/templates/profiles/subscription_form.html
amickan: Also correct. In a first implementation I didn't use `{% crispy form %}` and when changing it I forgot to remove the csrf template tag.

app/grandchallenge/profiles/views.py
```diff
@@ -197,3 +206,94 @@ def form_invalid(self, form):
...
+            and self.object.user != self.request.user
+        ):
+            update_verification_user_set.signature(
+                kwargs={
+                    "usernames": [
+                        self.request.user.username,
+                        self.object.user.username,
+                    ]
+                }
+            ).apply_async()
```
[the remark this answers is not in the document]
amickan: 1. True, I hadn't though about GET. I think it makes sense to do that always. I'm unsure where to move this check to though. Moving it to `test_func` would be straight forward, but then test_func does more than just a permission checking. Other methods that run before rendering the form are `get_context_data` and `get_form_kwargs` - semantically, neither of those fit the purpose either. So I'm tempted to add it to test_func. What do you think?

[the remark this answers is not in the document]
jmsmkn: I think that doing it in `test_func` is probably fine as it's an action we want to take in response to a dodgy permissions check. Otherwise, you would do it in `dispatch` or `get`, but then you might end up doing double work elsewhere.

jmsmkn (approved):

PR #3251, opened 2024-02-21
Add `List-Unsubscribe` email headers

amickan:
This adds the `List-Unsubscribe` and `List-Unsubscribe-Post` headers to all subscription emails on GC and standardizes email sending so that all emails now use a common html template. 

Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/295 

Example notification email:
image


app/grandchallenge/emails/emails.py
```diff
@@ -1,34 +1,111 @@
...
+            "List-Unsubscribe-Post": "List-Unsubscribe=One-Click",
+        }
+    else:
+        return None
+
+
+def create_email_object(
+    recipient, subject, message, connection, subscription_type=None
+):
+    site = Site.objects.get_current()
```
jmsmkn, line 82: By moving this here I think that you now do one query per email sent. This introduces an n+1 problem, which will be an issue when sending out the mailing list emails.

app/grandchallenge/emails/emails.py
```diff
@@ -1,34 +1,111 @@
...
+        {
+            "title": subject,
+            "username": recipient.username,
+            "content": message,
+            "unsubscribe_link": unsubscribe_link,
+            "subscription_type": subscription_type,
+            "site": site,
+        },
+    )
+    html_content_without_linebreaks = html_content.replace("\n", "")
```
jmsmkn, line 97: How do the plain text emails look? Doesn't this approach strip out all paragraph breaks and links?

Did you consider the approach of writing all of our emails in Markdown, which is human readable, sending the rendered markdown as the html and the plain markdown as the plaintext?

app/grandchallenge/emails/emails.py
```diff
@@ -1,34 +1,111 @@
...
+                subject=subject,
+                message=message,
+                subscription_type=subscription_type,
+                connection=connection,
+            )
+        )
+    return connection.send_messages(messages)
+
+
+def get_unsubscribe_link(recipient, subscription_type):
```
jmsmkn, line 54: Would prefer `kwarg` only arguments for functions, especially if they have more than one argument.

app/grandchallenge/emails/emails.py
```diff
@@ -1,34 +1,111 @@
...
+def filter_recipients(recipients, subscription_type=None):
+    filtered_recipients = []
+    for recipient in recipients:
+        if not recipient.is_active:
+            # Do not send emails to inactive users
+            continue
 
-    send_mail(
+        if (
+            subscription_type == EmailSubscriptionTypes.NOTIFICATION
```
jmsmkn, line 19: Maybe we should extend EmailSubscriptionTypes with SYSTEM? Then we can set subscription type to always be required and know that it is being correctly handled.

amickan: Good idea!

app/grandchallenge/emails/emails.py
amickan: Oh wow, you're absolutely right!

app/grandchallenge/emails/emails.py
amickan: I changed all email messages to html and removed line breaks, so they look fine, but I agree that using markdown would be nicer. Thanks for the suggestion!

amickan: Hm, looks like replacing of linebreaks is still necessary because the `md2email_html` function doesn't handle line breaks:

`First line\r\nSecond line` --> `<p>First line<br>\nSecond line</p>`
`First line\rSecond line` --> `<p>First line\nSecond line</p>`
`First line\nSecond line` --> `<p>First line\nSecond line</p>`

and the \n in there needs to go...

amickan: We can circumvent this issue in our service emails by using the `<br/>` tag in the message instead of `\n`, but this is an issue for the newsletter emails since they will be written in the markdown editor and will result in the first example. So I suggest we keep `html_content_without_linebreaks = html_content.replace("\n", "")`.

amickan: and using the markdown message as the plain text for the email would be incomplete (without username and the unsubscribe footer). We can do the conversion from md to html in the standard method, but I will keep all other steps in there.

jmsmkn: Are you sure your test is correct here?

If you type `\n` in a markdown file, then that is the literal `\` character followed by the `n` character, so will be rendered. If you put `\n` in a string that passes through the function, it is the newline character, so will not be rendered. It is no problem to send newline character in plain text emails. In fact, they should be there. In markdown, a paragraph break is represented by two newlines (`\n\n`), in which case a new paragraph break will be inserted.

We should not have carriage returns (`\r`) anywhere in our codebase as we use unix line endings.

amickan: I tried this out by creating an email in the UI (like the 10% team would do for the newsletter):
image

then retrieving the email in the shell:
```
>>> email.body
'First line\r\n\r\nSecond line'`
>>> md2email_html(email.body)
'<p>First line</p>\n<p>Second line</p>'
```

The `\n` shouldn't be there, or at least we don't want it there because it will get inserted as is in the html email template and would look like this: 

image

jmsmkn: For example, for the newsletter, you have `email.body`, which is a markdown string. You would then send this to `send_standard_email_batch` as `message=email.body` (maybe a better name for `message` is `message_markdown`).

Then, in `send_standard_email_batch` you generate `message_html = md2email_html(message_markdown)` and send out `message_html` and `message_markdown`.

For the other emails you create a template, e.g. app/grandchallenge/challenges/templates/challenges/partials/challenge_request_confirmation_email.md

```markdown
Your challenge request has been sent to the reviewers. You will receive an email informing you of our decision within the next 4 weeks.

The reviewers might contact you for additional information during that time. You can track the status of your request [here]({{ link }}).

We have calculated the following budget estimate. This estimate is based on the information you provided in the form and reflects a rough estimation of the costs we expect to incur:

{% for key, value in budget.items %}
    - {{ key }}: {{ value }} €
{% endfor %}
```

You then do:

```python
message_markdown = render_to_string(
         "challenges/partials/challenge_request_confirmation_email.md",
         context,
     )
```

For the other emails probably what we had before was almost valid markdown:

```python
 message_markdown = format_html(
         (
             "You have {unread_notification_count} new notification{suffix}.\n"
             "Read and manage your notifications [here]({url})."
         ),
         unread_notification_count=n_notifications,
         suffix=pluralize(n_notifications),
         url=reverse("notifications:list"),
     )
```

The `\n` in there is fine as plain text.

amickan: Yes, the last two examples work (except for the fact that `message_markdown` does not contain the "Dear username" part and also not the unsubscribe footer, but in terms of linebreaks, all good.

But there is still the issue with the newsletter email, see my comment above.

jmsmkn: Why isn't this an issue with the current way of sending out the newsletter?

jmsmkn: > (except for the fact that message_markdown does not contain the "Dear username" part and also not the unsubscribe footer, but in terms of linebreaks, all good.

Ah of course, you would need a standard template to put it in to for that.

amickan: > Why isn't this an issue with the current way of sending out the newsletter?

It is an issue with the current way of sending out the newsletter, which is why we do `html_content_without_linebreaks = html_content.replace("\n", "")` to get rid of the unnecessary line breaks in the html message.

amickan: > Ah of course, you would need a standard template to put it in to for that.

Yes, so that's why I now take the rendered html message and turn that back into text with `strip_tags`. Otherwise, we would need two templates, one html and one plain text...

amickan: The plain text is only added as a fall back if someone's email client doesn;t support html emails, so I think it would be fine to only include the core message there. The unsubscribe link is in the header of the email, so we would still fullfill the new requirements.

jmsmkn: Yes, but we want to avoid using `strip_tags` as that gets rid of all the links, right?

jmsmkn: > The plain text is only added as a fall back if someone's email client doesn;t support html emails, so I think it would be fine to only include the core message there. The unsubscribe link is in the header of the email, so we would still fullfill the new requirements.

We would not, the requirements are to have a prominent link at the footer of the email too.

jmsmkn: I still do not understand why `{{ object.body|md2email_html }}` works fine in `email_detail.html`, but we need extra processing when doing `md2email_html(object.body)` followed by `render_to_string(...)`. Why not do `{{ content|md2email_html }}` in `action.html`? Does it have the same problem?

amickan: Doing `{{ content|md2email_html }}` in `action.html` actually seems to work!

amickan: > We would not, the requirements are to have a prominent link at the footer of the email too.

Where do you read that? The RFCs only mention the headers. Of course, it is much user friendlier to add a link in the footer, and I can create a second template to use for the plain text content, but strictly-speaking, just adding the headers is enough to fullfill the requirements as far as I know. It's only if you don't have the headers, that Google and Yahoo will keep accepting the mail until June 2024 as long as you do have a prominent link in the footer.

jmsmkn: https://support.google.com/a/answer/81126?sjid=10053208798928542167-EU&visit_id=638442052921949914-1483516976&rd=1#zippy=%2Crequirements-for-sending-or-more-messages-per-day

Marketing messages and subscribed messages must support one-click unsubscribe, and include a clearly visible unsubscribe link in the message body.

jmsmkn: I think having a secondary standard `.txt` template that adds the greeting, content block, sign-off and unsubscribe link is fine. We should be able to keep the two synced and you will end up with better looking email than by stripping the tags.

app/grandchallenge/emails/templates/vendored/mailgun_transactional_emails/action.html
```diff
@@ -1,49 +1 @@
...
... clear: both; color: #999; margin: 0; padding: 20px;"> <table width="100%" style="font-family: 'Helvetica Neue',Helvetica,Arial,sans-serif; box-sizing: border-box; font-size: 14px; margin: 0;"> <tr style="font-family: 'Helvetica Neue',Helvetica,Arial,sans-serif; box-sizing: border-box; font-size: 14px; margin: 0;"> <td class="aligncenter content-block" style="font-family: 'Helvetica Neue',Helvetica,Arial,sans-serif; box-sizing: border-box; font-size: 12px; vertical-align: top; color: #999; text-align: center; margin: 0; padding: 0 0 20px;" align="center" valign="top"> This is an automated email from {{ site.domain.lower }}.<br> {% if unsubscribe_link %} If you no longer wish to receive {{ subscription_type|lower }} emails from us, <a href="{{ unsubscribe_link }}" style="font-family: 'Helvetica Neue',Helvetica,Arial,sans-serif; box-sizing: border-box; font-size: 12px; color: #999; text-decoration: underline; margin: 0;">unsubscribe here</a>. {% endif %} </td> </tr> </table> </div> </div> </td> <td style="font-family: 'Helvetica Neue',Helvetica,Arial,sans-serif; box-sizing: border-box; font-size: 14px; vertical-align: top; margin: 0;" valign="top"></td> </tr> </table> </body> </html>
```
jmsmkn, line 1: All the linebreaks and tabs have disappeared from this file, hard to read now. Was it intentional?

app/grandchallenge/challenges/emails.py
```diff
@@ -71,25 +72,25 @@ def send_challenge_status_update_email(challengerequest, challenge=None):
...
+            "challenges/partials/challenge_request_acceptance_email.html",
             context,
         )
     elif (
         challengerequest.status
         == challengerequest.ChallengeRequestStatusChoices.REJECTED
     ):
         message = render_to_string(
-            "challenges/partials/challenge_request_rejection_email.txt",
+            "challenges/partials/challenge_request_rejection_email.html",
```
jmsmkn, line 83: All templates should be in markdown format otherwise the plain text emails will contain all of the <p> tags etc.

app/grandchallenge/emails/templates/vendored/mailgun_transactional_emails/action.html
amickan: Yes, because they otherwise end up as \n in the email.

jmsmkn: Okay, then I really do not understand what is going on. Why does `{{ content|md2email_html }}`, which can contain newlines, work then, but the template, which does contain newlines, does not?

amickan: I also don't! Actually when trying things out again, even `{{ content|md2email_html }}` doesn't seem to work after all. Can't work out why it (seemed) to work before. I will try this out tomorrow again, with a fresh head. Sorry for the back and forth!

jmsmkn: I think there is a problem with your testing and there is no need to do any newline parsing. I cannot reproduce the problem. My process:
- Trigger the email which logs the raw email to the celery console
- Put the content into a `.eml` file
- Open that email with Apple Mail

Attached are the three emails that are produced when logging in as superuser:

`raw.html` created with 59e1f02:
[raw.eml.zip](https://github.com/comic/grand-challenge.org/files/14376766/raw.eml.zip)
Screenshot 2024-02-22 at 17 44 20

`raw-expanded.html` created with 883856f:
[raw-expanded.eml.zip](https://github.com/comic/grand-challenge.org/files/14376793/raw-expanded.eml.zip)
Screenshot 2024-02-22 at 17 46 26

`raw-expanded-with-lots-of-newlines` created with 883856f but setting `markdown_message="First line\r\n\r\nSecond line\n\nWe noticed a new login to your account.\n\n\n\n\n\n\n\n\nIf this was you, you don't need to do anything. If not, please change your password and update your 2FA device."`:
[raw-expanded-with-lots-of-newlines.eml.zip](https://github.com/comic/grand-challenge.org/files/14376839/raw-expanded-with-lots-of-newlines.eml.zip)
Screenshot 2024-02-22 at 17 47 51

Also did the same for the newsletter
[raw-newsletter.eml.zip](https://github.com/comic/grand-challenge.org/files/14376938/raw-newsletter.eml.zip)
Screenshot 2024-02-22 at 17 54 12

jmsmkn (dismissed):
This is good to go if you are happy with how I tested this with Apple Mail and Outlook

app/tests/emails_tests/test_tasks.py
```diff
@@ -168,16 +169,136 @@ def test_filter_recipients(subscription_type, expected_recipients):
...
+
+    assert str(error.value) == "Inactive users cannot be emailed"
+
+
+@pytest.mark.django_db
+def test_can_always_email_system_messages():
+    user = UserFactory(is_active=True)
+
+    user.user_profile.receive_newsletter = True
+    user.user_profile.receive_notification_emails = True
```
amickan, line 222: Shouldn't these be set to `False` to test that we can always send system emails?

app/grandchallenge/emails/templates/vendored/mailgun_transactional_emails/action.html
amickan: Ok, thanks for testing this as well! Looks great, I can also reproduce this now.

I had tested things in the shell (up until render_to_string with the action template). Saving the output of that to file and opening it in with Outlook looks fine for me as well, but I didn't test things that way, I instead copy pasted the rendered string into https://app.postdrop.io/ which apparently interprets the linebreaks differently.

app/tests/emails_tests/test_tasks.py
```diff
@@ -79,14 +85,220 @@ def test_email_content(settings):
...
```
jmsmkn, line 222:
```suggestion
    user.user_profile.receive_newsletter = False
    user.user_profile.receive_notification_emails = False
```

app/grandchallenge/emails/templates/vendored/mailgun_transactional_emails/action.html
jmsmkn: Okay, then yes, you needed to `print` the rendered string first before copying. Using the actual email document is more robust, you can also check the headers there too. Rather than copying from the console we could use the file backend, but this concatenates everything into a single file rather than individual messages https://docs.djangoproject.com/en/5.0/topics/email/#file-backend.

app/tests/emails_tests/test_tasks.py
jmsmkn: Good catch!

jmsmkn (approved):

PR #3259, opened 2024-02-26
Add dummy letter to upload widget ID for interfaces starting with a number

### LINKED ISSUE #3256: Unsupported upload interface type breaks later file uploads widgets
I have a reader study with a 3d-liver-model interface (.obj kind). The following likely also extends to archive items.

We don't support uploading these via the UI.

However, when one of these interfaces is present all subsequent file upload widgets are not loaded on page: https://grand-challenge.org/reader-studies/chris-test-reader-study-the-second-coming/display-sets/create-single/

image

I've since removed the display set with the interface to quickly do a workaround.

Not sure if this needs to be fixed or is so corner case that we can safely ignore it.

amickan:
An id name cannot start with a number in CSS (despite being valid HTML). Interfaces starting with a number break our upload widget because of that. 

Closes #3256

### DIFF
--- a/app/grandchallenge/uploads/widgets.py
+++ b/app/grandchallenge/uploads/widgets.py
@@ -1,3 +1,5 @@
+import re
+
 from django.forms.widgets import HiddenInput, MultipleHiddenInput
 
 
@@ -11,8 +13,11 @@ class UserUploadWidgetMixin:
 
     def get_context(self, *args, **kwargs):
         context = super().get_context(*args, **kwargs)
+        widget_id = context["widget"]["attrs"]["id"]
+        if re.match(r"^\d", widget_id):
+            context["widget"]["attrs"]["id"] = f"X_{widget_id}"
         context["widget"]["allowed_file_types"] = {
-            "id": f"{context['widget']['attrs']['id']}AllowedFileTypes",
+            "id": f"{widget_id}AllowedFileTypes",
             "value": self.allowed_file_types,
         }
         return context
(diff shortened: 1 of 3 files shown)


amickan:
I just realized while testing that this is also an issue for htmx and thus for the FlexibleImageWidget We currently don't have any image interfaces that start with a number, so it's not a problem yet. I'm wondering if instead of adding a similar workaround for the htmx views where necessary, it might be better to addd a restriction for interface names?

app/grandchallenge/uploads/widgets.py
```diff
@@ -11,8 +13,12 @@ def __init__(self, *args, allowed_file_types=None, **kwargs):
 
     def get_context(self, *args, **kwargs):
         context = super().get_context(*args, **kwargs)
+        widget_id = context["widget"]["attrs"]["id"]
+        if re.match(r"^\d", widget_id):
+            # id names cannot start with a number in CSS
```
jmsmkn, line 18: Why not prefix everything for consistency? Or just flip it around to `allowedFileTypes{context['widget']['attrs']['id']`?

jmsmkn:
> I just realized while testing that this is also an issue for htmx and thus for the FlexibleImageWidget We currently don't have any image interfaces that start with a number, so it's not a problem yet. I'm wondering if instead of adding a similar workaround for the htmx views where necessary, it might be better to addd a restriction for interface names?

I think fix the frontend instead, would be a strange restriction.

app/grandchallenge/uploads/widgets.py
amickan: True, prefixing all makes sense.

jmsmkn (approved):

PR #3260, opened 2024-02-26
Add empty answer confirmation and related label options to reader-study questions

chrisvanrun:
Part of the pitch:
- https://github.com/DIAGNijmegen/rse-roadmap/issues/271

Adds the required fields on GC side.

Renders as:
image


app/grandchallenge/reader_studies/models.py
```diff
@@ -1370,6 +1379,28 @@ def _clean_answer_type(self):
...
+                "Cannot have answer confirmation and have a question be "
+                "required at the same time"
+            )
+
+        if self.answer_type in (
+            AnswerType.HEADING,
+            AnswerType.BOOL,
+            AnswerType.CHOICE,
+            AnswerType.MULTIPLE_CHOICE,
+        ):
```
jmsmkn, line 1397: Isn't it safer to go the other way around and only check for Questions where we know this is supported? Or will this be automatically supported as new annotation types (etc) are added?

chrisvanrun: The way it's implemented in CIRRUS is that it is automatically supported for all widgets and widgets need to explicitly disable or ignore it (e.g. Accept/Reject workflow).

It's an invalid (widget) configuration for the tuple of types where it checks here.

I think it is safer to keep it as an 'is unsupported' check, because I **know** I'll forget to add any new types to the supported list.

Moreover, it should be supported/checked on the level of widgets in GC. However, since we don't have widgets for HEADING or BOOLEAN and it applies to all AnswerTypes of CHOICE and MULTIPLE_CHOICE, I though the answer_type is a decent place for now.

jmsmkn: Okay - make sure you error out at runtime on the client-side if you are asked to do this but it is not implemented.

jmsmkn: > because I know I'll forget to add any new types to the supported list.

Given how often we release GC I don't see that as an issue. I would rather go for safety.

chrisvanrun: Ok. I'll switch the checks around.

jmsmkn (approved):

PR #3265, opened 2024-02-28
Form to turn phase into algorithm phase

### LINKED ISSUE #2942: Form to turn phase into algorithm phase
Automatic creation of phases was removed in #2941 due to problems with validation, and they would all be called "challenge". The setup for an algorithm phase is quite involved. This can all be automated with a form that collects:

- The source phase
- The input and output interfaces

It would then automatically create the archive, add the challenge editors to that archive, set the phases submission type, verification status and interfaces.

amickan:
This PR adds a new view to turn prediction phases into algorithm phases. It adds a button to do so to the Challenge list view. 
The form takes the phases and the algorithm inputs and outputs as values: 
image

The link to this form is both on the challenge request detail page: 
image

and in the challenge admin:
image

Closes #2942


app/grandchallenge/evaluation/admin.py
```diff
@@ -58,6 +65,84 @@ def clean(self):
...
+                "title": "Turn phase(s) into algorithm phase(s)",
+            }
+        )
+        return context
+
+    def get_success_url(self):
+        return reverse("admin:evaluation_phase_changelist")
+
+    def turn_phase_into_algorithm_phase(self, *, phase, inputs, outputs):
+        archive = Archive.objects.create(
```
jmsmkn, line 127: The archive is missing a `full_clean` (I think that it's not valid anyway as it requires a logo).

app/grandchallenge/evaluation/admin.py
```diff
@@ -58,6 +65,84 @@ def clean(self):
         return cleaned_data
 
 
+class ConfigureAlgorithmPhasesForm(Form):
+    phases = ModelMultipleChoiceField(
+        queryset=Phase.objects.select_related("challenge")
+        .filter(submission_kind=SubmissionKindChoices.CSV)
+        .all()
+    )
```
jmsmkn, line 73: Unsure if we should allow this for multiple phases, it might be a bit safer if we only allow it for individual ones.

Maybe we should exclude phases that have a method and phases that have submissions for more safety?

app/grandchallenge/evaluation/admin.py
```diff
@@ -58,6 +65,84 @@ def clean(self):
...
+    algorithm_inputs = ModelMultipleChoiceField(
+        queryset=ComponentInterface.objects.all()
+    )
+    algorithm_outputs = ModelMultipleChoiceField(
+        queryset=ComponentInterface.objects.all()
+    )
```
jmsmkn, line 79: Could we make these the autocomplete fields like we have on the normal form, these are too fiddly with so many interfaces. You should be able to reuse how the admin does it, I think:

Screenshot 2024-02-29 at 15 46 32

app/grandchallenge/evaluation/admin.py
```diff
@@ -58,6 +65,84 @@ def clean(self):
...
+
+    def clean(self):
+        cleaned_data = super().clean()
+        try:
+            for phase in self.cleaned_data["phases"]:
```
jmsmkn, line 84: Does this matter? I thought that the slug field will automatically choose a free slug?

app/grandchallenge/evaluation/admin.py
```diff
@@ -58,6 +65,84 @@ def clean(self):
         return cleaned_data
 
 
+class ConfigureAlgorithmPhasesForm(Form):
+    phases = ModelMultipleChoiceField(
```
jmsmkn, line 69: All of the fields on the form should be required.

app/grandchallenge/evaluation/admin.py
```diff
@@ -58,6 +65,84 @@ def clean(self):
...
+        messages.success(self.request, "Phases were successfully updated")
+        return super().form_valid(form)
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context.update(
+            {
+                **admin.site.each_context(self.request),
+                "opts": Phase._meta,
+                "title": "Turn phase(s) into algorithm phase(s)",
```
jmsmkn, line 118: There is a lot going on in this admin integration that I do not really understand and may be fragile. Would it be more straightforward using a normal view? Especially now you have the specialised permission.

app/grandchallenge/evaluation/admin.py
amickan: I feel like allowing to do this for multiple phases is one of the main use cases of this form, right? We often have to configure two (or even more phases) in exactly the same way. 

But I agree that exlcuding phases with a method and submissions is a good idea to lock it down more.

app/grandchallenge/evaluation/admin.py
amickan: Ah you might be right, I will double check.

app/grandchallenge/evaluation/admin.py
amickan: I can get the autocomplete widget to work for the inputs and outputs, but not for the phases.

app/grandchallenge/evaluation/admin.py
amickan: They are required by default.

app/grandchallenge/evaluation/admin.py
amickan: I can turn it into a normal view, sure. The extra context here is so that the sidebar and the breadcrumbs get displayed.

jmsmkn: I think a normal view is best for this. The very first version of GC had lots of custom modifications of the admin which were tedious to maintain long term, so I ended up rewriting everything as normal views.

app/grandchallenge/evaluation/admin.py
jmsmkn: Okay, then I think I prefer `CheckboxSelectMultiple` over `SelectMultiple` as it's nicer to use (feel like too many mistakes can creep in with cmd+click+scroll), but up to you.

app/grandchallenge/evaluation/admin.py
amickan: Ok, understood. Unsure now about the permissions: the specific new permission I added is enough if you only grant it to staff users, but normally admin views are locked down to staff users additionally. Should I add a check for that as well, just in case and to be in line with other admin views?

jmsmkn: Staff should be limited to checking for the Django admin access. Having a global permission check on a user for a normal view should be sufficient, right? It's the same as what we have for say, blog post creation. We could eventually create a group of normal users (like the DIAG 10% team) who can onboard challenges.

app/grandchallenge/evaluation/templates/evaluation/configure_algorithm_phases_form.html
```diff
@@ -0,0 +1,25 @@
...
+            aria-current="page">Configure algorithm phases
+        </li>
+    </ol>
+{% endblock %}
+
+{% block content %}
+    <h3>Configure algorithm submission phases</h3>
+    <form method="post">
+        {% crispy form %}
+    </form>
```
jmsmkn, line 19: Form block is unnecessary

```suggestion
    {% crispy form %}
```

app/grandchallenge/evaluation/forms.py
```diff
@@ -539,3 +547,24 @@ def clean(self):
             raise ValidationError("This challenge has exceeded its budget")
 
         return cleaned_data
+
+
+class ConfigureAlgorithmPhasesForm(SaveFormInitMixin, Form):
+    phases = ModelMultipleChoiceField(
```
jmsmkn, line 553: Shouldn't the phases be filtered by the current challenge?

app/grandchallenge/evaluation/views/__init__.py
```diff
@@ -980,3 +987,65 @@ def get_success_url(self):
...
+                short_name=phase.challenge.short_name
+            )
+            phase.algorithm_time_limit = (
+                challenge_request.inference_time_limit_in_minutes
+            )
+        except ObjectDoesNotExist:
+            messages.warning(
+                self.request,
+                f"Algorithm run time limit could not be configured for {phase}.",
+            )
```
jmsmkn, line 1044: It might be better to add this to the form and use the value from the challenge request as the initial value if it exists, then you avoid a lookup in `form_valid`.

app/grandchallenge/evaluation/forms.py
jmsmkn: Ah wait, you have it on the challenges list, this might be better on the per-challenge level (e.g. url in the evaluation app and link from a page in there). I don't see a reason to do this for multiple challenges at once? Maybe I'm missing something. That way you can also add the time limit from the challenge request in the form.

amickan: What current challenge? I implemented the view outside of the challenge context.

app/grandchallenge/evaluation/views/__init__.py
amickan: Ok, then we would presume that there can only be one time limit, so we can only configure phases from one challenge at a time.

app/grandchallenge/evaluation/forms.py
amickan: Sorry, your comment somehow wasn't showing when I wrote mine. Ok, I think you're right that we mostly want to configure phases from one and the same challenge and never across challenges. So yes, moving this to the challenge level makes sense. I'm only wondering, from a usability perspective, staff users don't have access to the challenge pages, so a link on those pages won't be visible to us. We'll have to go straight to the url to do this then.

app/grandchallenge/evaluation/views/__init__.py
amickan: Ignore my last comment. I hadn't seen your earlier comment.

amickan:
> Somehow my latest commit isn't showing here. Not sure what happened. Git push was successful.

Making another commit did the trick...

jmsmkn (approved):

PR #3280, opened 2024-03-20
Instant notification email option

amickan:
This migrates the `receive_notification_emails` field on the user profile from a boolean field to a choice field; and enables instant email notifications for users who opted in to this. 

Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/300


app/grandchallenge/direct_messages/emails.py
```diff
@@ -73,11 +77,10 @@ def send_new_unread_direct_messages_email(
...
         url=reverse("direct-messages:conversation-list"),
     )
-    site = Site.objects.get_current()
     send_standard_email_batch(
         site=site,
         subject=subject,
         markdown_message=msg,
         recipients=[user],
-        subscription_type=EmailSubscriptionTypes.SYSTEM,
+        subscription_type=EmailSubscriptionTypes.NOTIFICATION,
```
amickan, line 85: I noticed that we check the notification email preference when determining whether or not to email users about unread messages. I wasn't aware of that before, and wrongly classified them as system messages. It's also not clear from the help text of the model field that this is the case. I wonder if a seperate email preference setting would make sense for direct messages? That could be done in a seperate PR. Else I will just adapt the help text.

app/grandchallenge/profiles/migrations/0015_alter_userprofile_receive_notification_emails.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("profiles", "0014_userprofile_unread_messages_email_last_sent_at"),
+    ]
+
+    operations = [
+        migrations.AlterField(
```
jmsmkn, line 25: Just to double check - have you tested that this migration actually works? I don't know what the consequences of casting a bool field to char are.

app/grandchallenge/profiles/models.py
```diff
@@ -55,9 +67,13 @@ class UserProfile(models.Model):
         help_text="Display the organizations that you are a member of in your profile.",
     )
 
-    receive_notification_emails = models.BooleanField(
-        default=True,
-        help_text="Whether to receive notification emails",
+    receive_notification_emails = models.CharField(
+        max_length=50,
```
jmsmkn, line 71: Max length only needs to be 13

app/grandchallenge/profiles/migrations/0015_alter_userprofile_receive_notification_emails.py
```diff
@@ -0,0 +1,40 @@
...
+                    ("DAILY_SUMMARY", "Send me an email with a daily summary"),
+                    ("DISABLED", "Do not send me notification emails"),
+                    ("INSTANT", "Send me an email immediately"),
+                ],
+                default="DAILY_SUMMARY",
+                help_text="Whether to receive emails about unread notifications and how often (immediately vs. once a day if necessary).",
+                max_length=50,
+            ),
+        ),
+        migrations.RunPython(migrate_notification_email_preferences),
```
jmsmkn, line 39: This should be marked as elidable as there is nothing to do with an empty db so can be removed when we squash the migrations.

app/grandchallenge/profiles/models.py
```diff
@@ -55,9 +67,13 @@ class UserProfile(models.Model):
         help_text="Display the organizations that you are a member of in your profile.",
     )
 
-    receive_notification_emails = models.BooleanField(
-        default=True,
-        help_text="Whether to receive notification emails",
+    receive_notification_emails = models.CharField(
```
jmsmkn, line 70: Unfortunately we cannot just alter the field. There will be code running after the migration has happened but is still assuming that a boolean field is in use. This will only be 10 minutes but our application generates a lot of notifications. With a char field all of the `if self.receive_notification_emails` will pass, and we would end up emailing people who have opted out.

The approach needs to change:

- a new field should be added to store the choice (maybe `notification_email_choice`?), this will be one migration file.
- a data migration (e.g. a second migration file as AFAIK db modification and data changes cannot happen in the same transaction in postgres) should be made to fill out the new field
- the code should be updated to use the new field (search the codebase to ensure there are no more instances of `receive_notification_emails`)

That's 1 PR.

Once that is deployed a 2nd PR needs to be created where `receive_notification_emails` should be marked as deprecated with `deprecate_field` and a new migration made.

Then, once that is deployed a 3rd PR should be created to remove the `receive_notification_emails` field and a new migration made.

I won't check the rest of the PR right now as unfortunately this is a lot to change, sorry.

app/grandchallenge/direct_messages/emails.py
jmsmkn: Good catch! I think help text modification is fine.

app/grandchallenge/profiles/migrations/0015_alter_userprofile_receive_notification_emails.py
amickan: I tested this, yes, and it worked. Here is what I did: I first spun up the dev environment with the old model field, updated user preferences to have some who opted in and some who opted out, then changed the field, created the migration (and added the extra migration step) and then ran `migrate` and checked that the user preferences were migrated correctly.

app/grandchallenge/profiles/migrations/0015_alter_userprofile_receive_notification_emails.py
amickan: Ah true, good catch.

app/grandchallenge/profiles/models.py
amickan: Ok, I understand the step wise approach, and understand that this is generally how one should ago about making these types of db schema changes, but in this specific case, I (maybe wrongly) reasoned that the 10 minutes or so after the migration runs and before the new code is running don't matter since we only send the email once a day at 4am and so by the time we send emails, the new code would be running (instant email wouldn't have an effect since no user has that set as their preference after migrating). Does that make sense?

I will change the approach, just wanted to explain why I did what I did. :-)

PR #3282, opened 2024-03-21
Enable instant notification emails

amickan:
This adds a new `notification_email_choice` field to the user profile and populates this field based on each user's `receive_notification_emails` preference. The code is refactored to use the new field, and instant email notifications are enabled for users who opted in to this.

Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/300


app/grandchallenge/profiles/models.py
```diff
@@ -146,20 +166,58 @@ def get_unsubscribe_link(self, *, subscription_type):
...
                 f"Unknown subscription type: {subscription_type}"
             )
 
+    def dispatch_unread_notifications_email(
+        self, *, site, unread_notification_count=None
+    ):
+        self.notification_email_last_sent_at = now()
+        self.save(update_fields=["notification_email_last_sent_at"])
+
+        send_unread_notifications_email(
```
jmsmkn, line 192: `send_unread_notifications_email` doesn't seem to be used anywhere else and in this case should not be used directly. Maybe get rid of the `send_unread_notifications_email` function and put everything it's doing in this new method? Same with `send_new_unread_direct_messages_email`. That way we do not need to make sure that developers only call the dispatch methods.

app/tests/profiles_tests/test_models.py
```diff
@@ -69,3 +81,80 @@ def test_submit_newsletter_preference(client):
...
+        setattr(user.user_profile, subscription_attr, subscription_preference)
+    with expectation:
+        link = user.user_profile.get_unsubscribe_link(
+            subscription_type=subscription_type
+        )
+        if link:
+            assert link == reverse(
+                unsubscribe_viewname,
+                kwargs={"token": user.user_profile.unsubscribe_token},
+            )
```
jmsmkn, line 160: Only 1 action should be performed in a `with pytest.raises` block. Each statement should be in its own context. That way, the test ensures that the exception is raised only in the exact statement where you expect it. Remember that the exception is raised and caught, so any code after the exception will not be executed.

app/grandchallenge/profiles/models.py
```diff
@@ -146,20 +166,58 @@ def get_unsubscribe_link(self, *, subscription_type):
...
+                unread_notification_count
+                if unread_notification_count
+                else self.unread_notification_count
+            ),
+        )
+
+    def dispatch_unread_direct_messages_email(
+        self, *, site, new_unread_message_count=None
+    ):
+        new_senders = [s.first_name for s in get_new_senders(user=self.user)]
```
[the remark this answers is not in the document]
amickan: True, if they receive mutliple DMs at the same time, they would get multiple emails that each mention multiple senders, rather than just the one sender that the email was triggered by. I will add another argument to the function to fix that.

jmsmkn (approved):

PR #3286, opened 2024-03-25
Simplify json error reporting

amickan:
Json validation errors used to include the instance (the submitted json), the schema and the error. This changes the reporting to only print the error message without the schema and instance. This is a temporary fix until this is implemented: https://github.com/python-jsonschema/jsonschema/issues/1218 

It also standardizes the message for `FILE_COPY_STATUS` notifications, opting for the less verbose version.

### DIFF
--- a/app/grandchallenge/core/validators.py
+++ b/app/grandchallenge/core/validators.py
@@ -161,7 +161,9 @@ class JSONValidator:
         try:
             validate(value, self.schema, registry=self.registry)
         except JSONValidationError as e:
-            raise ValidationError(f"JSON does not fulfill schema: {e}")
+            raise ValidationError(
+                f"JSON does not fulfill schema: instance {e.message.replace(str(e.instance), '')}"
+            )
 
     def __eq__(self, other):
         return isinstance(other, JSONValidator) and self.schema == other.schema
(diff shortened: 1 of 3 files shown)


app/grandchallenge/core/validators.py
```diff
@@ -161,7 +161,9 @@ def __call__(self, value):
         try:
             validate(value, self.schema, registry=self.registry)
         except JSONValidationError as e:
-            raise ValidationError(f"JSON does not fulfill schema: {e}")
+            raise ValidationError(
+                f"JSON does not fulfill schema: instance {e.message.replace(str(e.instance), '')}"
```
amickan, line 165: This is not ideal since the format of the message can change any time, but there is currently no better way (that I can think of at least).

amickan:
Some of the tests that involve checking the returned json validation error currently don't pass. I wanted to get feedback on the general approach first before I change the tests. Maybe we decide not to change anything and wait for the feature request to be implemented.

jmsmkn:
Yes the approach is fine for now. Would be happy for you to work on fixing this upstream if you can.

amickan:
> Would be happy for you to work on fixing this upstream if you can.

I think I will wait this one out a little longer. From the conversation on the issue it sounds like the person who proposed it has a pretty clear idea of how to implement it already (whereas I don't, yet) and I have other things higher up on my priority list at the moment. I might give it a shot later though, if it hasn't been tackled by then.

jmsmkn (approved):

PR #3290, opened 2024-03-27
Add CIVSet bulk delete view

### LINKED ISSUE #3169: Bulk delete option for display sets

amickan:
This adds a bulk delete view for CIV sets and implements this for display sets.

A user can select a subset of (editable) display sets or all display sets (for the given reader study):
image

... and needs to confirm their selection for actual deletion to happen.
image

The initial selection of items is sent to a FormView via a GET request. The FormView prepopulates the multi-model select field on the confirmation form with the preselected items and renders that. The user could deselect a few of the options on the confirmation view if they wanted, upon submission, the items are deleted. 

The approach presented here aligns with how the Django admin does things. I also played around with an alternative implementation where the list view is essentially a form with just one `ModelMultipleChoiceField` using the `CheckboxSelectMultiple` widget to select items. Customizing the display of the items in the select widget to show all the other information that is currently displayed, however, was messy and nearly impossible to align to make it looks like a table. Also, enabling sorting and filtering would require a lot of custom code. So dropped that option. 

Closes #3169

### DIFF
--- a/app/grandchallenge/components/views.py
+++ b/app/grandchallenge/components/views.py
@@ -15,12 +15,17 @@ from rest_framework.viewsets import ReadOnlyModelViewSet
 from grandchallenge.algorithms.forms import NON_ALGORITHM_INTERFACES
 from grandchallenge.api.permissions import IsAuthenticated
 from grandchallenge.archives.models import Archive
-from grandchallenge.components.forms import NewFileUploadForm, SingleCIVForm
+from grandchallenge.components.forms import (
+    CIVSetDeleteForm,
+    NewFileUploadForm,
+    SingleCIVForm,
+)
 from grandchallenge.components.models import ComponentInterface, InterfaceKind
 from grandchallenge.components.serializers import ComponentInterfaceSerializer
 from grandchallenge.core.guardian import (
     ObjectPermissionRequiredMixin,
     PermissionListMixin,
+    get_objects_for_user,
 )
 from grandchallenge.datatables.views import Column, PaginatedTableListView
 from grandchallenge.reader_studies.models import ReaderStudy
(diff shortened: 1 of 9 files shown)


app/grandchallenge/components/views.py
```diff
@@ -337,3 +342,75 @@ def get_queryset(self):
...
+                ),
+            }
+        )
+        return kwargs
+
+    @cached_property
+    def selected_objects(self):
+        # Selecting happens on the ListView through a GET request to this view
+        # on POST the selected objects are part of the form field
+        if self.request.method != "GET":
```
jmsmkn, line 396: Nitpick - it is preferable to do positive checks as they're easier to parse by the reader and change. So in this case it would be `== "GET"` and then swap the two blocks around.

jmsmkn (approved):
Looks great! Up to you whether to solve the nit.

jmsmkn (dismissed):

app/grandchallenge/components/views.py
```diff
@@ -337,3 +342,75 @@ def get_queryset(self):
...
+                    self.selected_objects
+                    if self.selected_objects
+                    else self.get_queryset()
+                ),
+            }
+        )
+        return kwargs
+
+    @cached_property
+    def selected_objects(self):
```
jmsmkn, line 393: Actually, is there a reason for having this as a separate property rather than doing in `get_queryset`? If it were done there you could avoid all of the `self.selected_objects if self.selected_objects else self.get_queryset()`.

app/grandchallenge/components/views.py
```diff
@@ -337,3 +342,75 @@ def get_queryset(self):
...
+                return self.get_queryset()
+            else:
+                selected = self.request.GET.getlist(
+                    "selected-for-deletion", None
+                )
+                # filtering the original queryset
+                # so that only objects with permission are included
+                return self.get_queryset().filter(pk__in=selected)
+        else:
+            return None
```
jmsmkn, line 408: You could instead return `self.model.objects.none()` so that you're always dealing with a queryset downstream, that would probably remove the need for `self.selected_objects.count() if self.selected_objects else 0`.

app/grandchallenge/components/views.py
amickan: I found it clearer to seperate, but it can be combined. We just need to make sure not to return an empty queryset on POST then because else validation on the form will fail. So then technically "delete_count" is no longer a delete count on POST.  That doesn't really matter since we only use that in the context of a GET request, but I have a slight preference for leaving the methods separate.

app/grandchallenge/components/views.py
amickan: Ah good idea!

jmsmkn (approved):

PR #3292, opened 2024-03-28
Add bulk delete view for archive items

amickan:
Adds bulk delete view for archive items. I noticed that we have a custom str method for archive items, so the confirmation view will look as follows and won't contain the archive item pk. Should I update the str method to also show the pk or remove the custom str method since we're not using it anywhere else critical as far as I can tell? 

image


jmsmkn (approved):

jmsmkn:
Your comment is related to https://github.com/comic/grand-challenge.org/issues/3018, would be good to solve it.

PR #3301, opened 2024-04-10
Fix drf ajax calls

### LINKED ISSUE #3293: Cannot delete notifications
**Describe the bug**
It is no longer possible for the user to delete their notifications. 

**To Reproduce**
1. Go to https://grand-challenge.org/notifications/
2. Select notifications
3. Click on "Delete selected"
4. Nothing happens
5. The console reveals that the CSRF token is missing

**Expected behavior**
The notifications should get deleted. The CSRF token should automatically be sent along with the ajax request. 

This used to work not too long ago. I can't readily identify what happened that broke this. Also couldn't fix it quickly. Needs some more investigation.

amickan:
DRF expects an element with id `drf_csrf` that contains the csrf token and header. That was removed at some point, breaking the ajax calls for updating and deleting notifications and subscriptions.

Closes #3293


app/grandchallenge/core/templates/grandchallenge/partials/drf_csrf.html
```diff
@@ -9,5 +9,7 @@
     {{ ""|json_script:"csrfToken" }}
 {% endif %}
 
+<div id="drf_csrf"></div>
```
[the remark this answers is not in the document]
amickan: > Does DRF require the global variable? If not, why not write `drf_csrf` directly with `json_script` and a template tag/filter?

That would be preferred, but I couldn't get that to work. We first need to build the dictionary with the csrf token and header and then write that with json_script. I don't think the Django template language is made for doing that, something like the following doesn't work: 
`{{ {'csrfHeaderName': csrfHeaderName, 'csrfToken': csrfToken}|json_script:"drf_csrf" }}` 

Is there another way of achieving this?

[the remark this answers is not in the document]
jmsmkn: Make a new template tag or template filter https://docs.djangoproject.com/en/5.0/howto/custom-template-tags/#simple-tags

If it is a tag, use it with:

```
{% my_new_tag as drf_data %}
{{ drf_data|json_script:"drf_csrf" }}
```

app/grandchallenge/core/templates/grandchallenge/partials/drf_csrf.html
```diff
@@ -1,13 +1,7 @@
 {% load static %}
-{% load string %}
+{% load csrf %}
 
-{{ csrf_header_name|default:'X-CSRFToken'|json_script:"csrfHeaderName" }}
+{% csrf_data csrf_token as drf_data %}
```
jmsmkn, line 4: Don't you also need to pass `csrf_header_name`?

The `bool(request) == False` path doesn't seem to be handled?

amickan: I made it so that you can optionally pass `csrf_header_name`. If you don't the default of "X-CSRFToken" is used. 

And why do we need to handle `bool(request) == False`? It will just be an empty string then right?

jmsmkn: > I made it so that you can optionally pass csrf_header_name. If you don't the default of "X-CSRFToken" is used.

Yes, but you're not sending `csrf_header_name` to the template tag function. It should be sent, right?

> And why do we need to handle bool(request) == False? It will just be an empty string then right?

Don't know. It was added here, I am not sure why https://github.com/comic/grand-challenge.org/pull/2891/commits/781e6c2195f87eb65c66b8bafca094289eff1e51. Maybe it was a problem with the `to_string` method. So okay, lets leave that out.

amickan: > but you're not sending `csrf_header_name` to the template tag function. It should be sent, right?

No, if it's not sent along, its default value is used "X-CSRFToken".

jmsmkn: So why not send it along? In the original implementation if the value of `csrf_header_name` changes this bit of code automatically works. In this implementation if the value of `csrf_header_name` changes this bit of code will break again. As there is no automated test for it, it would only be caught in production.

amickan: Yeah ok, that's a good point.

jmsmkn (approved):

PR #3311, opened 2024-04-12
Fix ValidationError printing

### LINKED ISSUE #3309: 'ValidationError' object has no attribute 'message'
See https://grand-challenge.sentry.io/issues/5181371764/?project=303639&query=%21logger%3Acsp+is%3Aunresolved&referrer=issue-stream&statsPeriod=14d&stream_index=2

Introduced in #3296

amickan:
Closes #3309

### DIFF
--- a/app/grandchallenge/core/utils/error_messages.py
+++ b/app/grandchallenge/core/utils/error_messages.py
@@ -1,0 +1,7 @@
+def format_error_message(error):
+    if hasattr(error, "message"):
+        return error.message
+    elif hasattr(error, "messages"):
+        return ", ".join(error.messages)
+    else:
+        return str(error)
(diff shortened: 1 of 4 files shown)


app/grandchallenge/core/utils/error_messages.py
```diff
@@ -0,0 +1,7 @@
+def format_error_message(error):
```
jmsmkn, line 1: This is specific to validation errors, right? Only there because of the construction of this class in django.

amickan: Yes. Should I add a check `if isinstance(error, ValidationError)` and error out if not?

jmsmkn: I think maybe just rename the method so developers know when to use it.

PR #3318, opened 2024-04-16
Update algorithm job detail API: add url and algorithm, remove algorithm_title

chrisvanrun:
Part of the pitch:
- https://github.com/DIAGNijmegen/rse-roadmap/issues/314

This PR adds a `url` to the job detail API to directly link to the regular detail view. To be added to CIRRUS.

It also removes `algorithm_title` and replaces it with a hyperlinked relation to the algorithm itself. The rationale is that the title has no place in this serialisation. However, to get the title, a requester would need to traverse via the `AlgorithmImage`, which only admins of an algorithm would have access to (?). Adding a direct rel to the algorithm enables a callee to retrieve the algorithm title via a secondary call to get details on the algorithm. The algorithm details are public.

### DIFF
--- a/app/grandchallenge/algorithms/serializers.py
+++ b/app/grandchallenge/algorithms/serializers.py
@@ -84,7 +84,6 @@ class JobSerializer(serializers.ModelSerializer):
     """Serializer without hyperlinks for internal use"""
 
     algorithm_image = StringRelatedField()
-    algorithm = StringRelatedField()
 
     inputs = ComponentInterfaceValueSerializer(many=True)
     outputs = ComponentInterfaceValueSerializer(many=True)
@@ -113,7 +112,6 @@ class JobSerializer(serializers.ModelSerializer):
             "url",
             "api_url",
             "algorithm_image",
-            "algorithm",
             "inputs",
             "outputs",
             "status",
@@ -143,7 +141,10 @@ class HyperlinkedJobSerializer(JobSerializer):
     outputs = HyperlinkedComponentInterfaceValueSerializer(many=True)
 
     class Meta(JobSerializer.Meta):
-        pass
+        fields = [
+            *JobSerializer.Meta.fields,
+            "algorithm",
+        ]
 
 
 class JobPostSerializer(JobSerializer):
--- a/app/grandchallenge/algorithms/serializers.py
+++ b/app/grandchallenge/algorithms/serializers.py
@@ -84,13 +84,13 @@ class JobSerializer(serializers.ModelSerializer):
     """Serializer without hyperlinks for internal use"""
 
     algorithm_image = StringRelatedField()
+    algorithm = StringRelatedField()
+
     inputs = ComponentInterfaceValueSerializer(many=True)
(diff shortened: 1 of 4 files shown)


app/grandchallenge/algorithms/serializers.py
```diff
@@ -110,13 +110,14 @@ class Meta:
...
             "pk",
+            "url",
             "api_url",
             "algorithm_image",
+            "algorithm",
             "inputs",
             "outputs",
             "status",
             "rendered_result_text",
-            "algorithm_title",
```
jmsmkn, line 119: I'm not sure that algorithm_title can be removed. Won't it need to be deprecated and go through two CIRRUS release cycles first?

jmsmkn: Hmm, it was added in 2020 without any context, any ideas why @HarmvZ ? https://github.com/comic/grand-challenge.org/pull/1633

chrisvanrun: https://github.com/DIAGNijmegen/rse-cirrus-core/commit/c33fa4b6a8fd0f44b184b54c5ed709dbe5dac84f

That plugin has seen major changes since.

jmsmkn: Great, and that was removed in https://github.com/DIAGNijmegen/rse-cirrus-core/commit/de699e337cb10a7588afc3cd80f97375f911e193#diff-b3e0218cb1d9d5340a73e38d6226ea7dc783dd8784f94ed749f4c766c957b0b1.

app/grandchallenge/algorithms/serializers.py
```diff
@@ -84,13 +84,13 @@ class JobSerializer(serializers.ModelSerializer):
     """Serializer without hyperlinks for internal use"""
 
     algorithm_image = StringRelatedField()
+    algorithm = StringRelatedField()
```
jmsmkn, line 87: Is this field being filled out anywhere? It is used in predictions.json.

jmsmkn: BTW, it is unnecessary for the predictions.json, and could be used only in the serialiser for API view. Make sure the API list view does not have an n + 1 problem.

chrisvanrun: It turns out that the API list view already HAD an n +1 problem. The culprit was the optional_hanging_protocols.

I am not sure what the best practice for testing the n + 1 problems is. I've added a somewhat naive test for it, but any feedback on that is welcome.

chrisvanrun: Oh, and I can confirm the predictions.json is solely reliant on the precursor of the serializer: JobSerializer. So it doesn't show up at the predictions.json.

app/grandchallenge/algorithms/serializers.py
chrisvanrun: FYI. This did show up in the predictions.json, but I do not think anyone is actively reliant upon it. Or I would hope not...

jmsmkn (approved):

PR #3322, opened 2024-04-22
Ignore B909

amickan:
With the most recent bugbear release, more checks were added to the B909 check and some of our code was flagged. I checked all instances and think what we do is safe. I don't think we want to ignore B909 entirely (?), so I instead added the noqa comments wherever necessary.


jmsmkn:
This is probably https://github.com/PyCQA/flake8-bugbear/issues/467, pinning bugbear (or excluding this version) is a better option.

jmsmkn (approved):

PR #3323, opened 2024-04-22
Dependent algorithm phases

amickan:
This adds a `parent` field to the Phase model which enables dependent phases. 

Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/309


app/tests/evaluation_tests/test_admin.py
```diff
@@ -1,17 +1,38 @@
...
+        2,
+        submission_kind=SubmissionKindChoices.ALGORITHM,
+        challenge=ChallengeFactory(),
+    )
+    p1.parent = p2
+    p1.save()
+    form = PhaseAdmin.form(
+        instance=p1,
+    )
+    for field in p1.read_only_fields_for_dependent_phases:
```
jmsmkn, line 37: The test is a little bit circular. If `read_only_fields_for_dependent_phases` returns no data, nothing will be tested. I think it's more robust to be explicit about what fields you expect to be disabled in this scenario.

app/tests/evaluation_tests/test_models.py
```diff
@@ -740,3 +742,110 @@ def test_give_algorithm_editors_job_view_permissions_only_for_algorithm_phase():
...
+    p1.save()
+    p2.save()
+    p3.save()
+
+    for p in [p2, p3, p4]:
+        assert p in p1.parent_phase_choices
+    assert p1 not in p1.parent_phase_choices
+    for p in [p3, p4]:
+        assert p in p2.parent_phase_choices
+    assert p1 not in p2.parent_phase_choices
```
jmsmkn, line 799: Wouldn't doing a set comparison be clearer?

```suggestion
    assert set(p1. parent_phase_choices) == {p2, p3, p4}
    assert set(p2.parent_phase_choices) == {p3, p4}
```

app/tests/evaluation_tests/test_models.py
```diff
@@ -740,3 +742,110 @@ def test_give_algorithm_editors_job_view_permissions_only_for_algorithm_phase():
...
+    with pytest.raises(ValidationError) as e:
+        p1.clean()
+    assert (
+        "The parent phase needs to have at least 1 valid archive item."
+        in str(e)
+    )
+
+    p1.parent = p2
+    with nullcontext():
+        p1.clean()
```
jmsmkn, line 834: `nullcontext` isn't doing anything here

```suggestion
    p1.clean()
```

app/grandchallenge/evaluation/models.py
```diff
@@ -509,6 +510,14 @@ class Phase(FieldChangeMixin, HangingProtocolMixin, UUIDModel):
         default=0,
         help_text="The total compute cost for this phase in Euro Cents, including Tax",
     )
+    parent = models.ForeignKey(
+        "self",
+        on_delete=models.PROTECT,
+        related_name="children",
+        null=True,
+        blank=True,
+        help_text="Is this phase dependent on another phase? If selected, submissions to the current phase will only be possible after a succesful submission has been made to the parent phase.",
```
jmsmkn, line 519: We should probably add a warning about enabling this. On deadline day if you require a successful submission to a sanity check phase in order to submit to the test phase it could prevent people submitting if the sanity check takes a long time to execute. Don't know how to phrase this better!

app/grandchallenge/evaluation/models.py
```diff
@@ -858,6 +895,59 @@ def send_give_algorithm_editors_job_view_permissions_changed_email(self):
...
+                    filter=Q(algorithm_outputs__in=algorithm_outputs),
+                    distinct=True,
+                ),
+            }
+            extra_filters = {
+                "total_input_count": len(algorithm_inputs),
+                "total_output_count": len(algorithm_outputs),
+                "relevant_input_count": len(algorithm_inputs),
+                "relevant_output_count": len(algorithm_outputs),
+            }
```
jmsmkn, line 935: I think I recognise this code from somewhere else? Or am I thinking of algorithm jobs and archive items? Are there any possibilities for reuse?

jmsmkn (commented):
Looks great! Hadn't thought of the circular depdency checking

app/tests/evaluation_tests/test_models.py
amickan: Ah I was looking for a better way but doing a set comparison didn't occur to me. Thanks!

app/tests/evaluation_tests/test_admin.py
amickan: Yeah ok, I understand. My thinking was that since I'm testing the specific fields in another test, it would suffice to only test here that any fields in `read_only_fields_for_dependent_phases` - if any - are disabled. If I check here for specific fields, I'd be testing two things in one test. But I agree that the test is more robust if I'm explicit about the fields.

jmsmkn: Ah okay, I think I missed the test for explicit fields. Still, I think it's better for tests to be explicit about the asserts. Generally, I think that something is often off with the test design if the asserts are in for loops or if statements which can lead to the assert not being hit.

app/grandchallenge/evaluation/models.py
amickan: Yes, this is very similar to what we do in `user_algorithms_for_phase`. The lookups are different (`inputs` vs. `algorithm_inputs` etc), so we couldn't directly reuse, but I could write a method to return the annotation and filter dictionaries based on the lookup parameters.

jmsmkn: Okay! Then no, no need to reuse.

app/grandchallenge/evaluation/models.py
amickan: Good idea. I also struggle to find a different wording. I think this is as clear as it can get. People will just need to stand still and think for a moment when enabling this.

jmsmkn (approved):

PR #3333, opened 2024-04-29
Install allauth-mfa and add data migration

amickan:


app/tests/conftest.py
```diff
@@ -480,3 +480,14 @@ def authenticated_staff_user(client):
         user=user,
     )
     return user
+
+
+@pytest.fixture
+def user_with_totp():
+    def _make_user_with_totp(is_staff=False):
```
jmsmkn, line 487: Is there a need to define this as a fixture? It is only used in one test. You could have defined `_make_user_with_totp` just in the one test that it is used in.

app/grandchallenge/core/middleware.py
```diff
@@ -1,33 +1,57 @@
...
+from grandchallenge.subdomains.utils import reverse
 
-class RequireStaffAndSuperuser2FAMiddleware(BaseRequire2FAMiddleware):
-    def require_2fa(self, request):
-        # Staff users and superusers are required to have 2FA.
+
+class RequireStaffAndSuperuser2FAMiddleware(MiddlewareMixin):
+    """Force multi-factor authentication for staff users and superusers."""
+
+    allowed_urls = [
```
jmsmkn, line 18: Should we instead allow any url in the allauth app? Might be easier to maintain. Any risks to that?

app/grandchallenge/profiles/migrations/0019_migrate_mfa.py
```diff
@@ -0,0 +1,50 @@
...
+from allauth.mfa.models import Authenticator
+from django.db import migrations
+from django_otp.plugins.otp_static.models import StaticDevice
+from django_otp.plugins.otp_totp.models import TOTPDevice
+
+
+def migrate_mfa(apps, schema_editor):
+    adapter = get_adapter()
+    authenticators = []
+    for totp in TOTPDevice.objects.filter(confirmed=True).iterator():
```
jmsmkn, line 13: I trust this migration has been tested? e.g. creating a TOTP device with the old code, then migrating, then logging in with the new. Same goes for recovery codes.

app/tests/profiles_tests/test_2fa.py
```diff
@@ -1,175 +1,114 @@
...
-    response = get_view_for_user(
-        viewname="two-factor-setup",
+def test_require_mfa_not_on_allowed_urls(
+    client,
+    url,
+    url_kwargs,
+    login_required,
+    expected_status_code,
+):
+    user = UserFactory(is_staff=True)
```
jmsmkn, line 77: May be clearer to update this variable name to staff_user?

pyproject.toml
```diff
@@ -77,7 +77,7 @@ django-haystack = "*"
 citeproc-py = "*"
 pyswot = "*"
 django-redis = "*"
-django-allauth = {extras = ["socialaccount"], version = "*"}
+django-allauth = {version = "*", extras = ["mfa", "socialaccount"]}
```
jmsmkn, line 80: The lockfile has changed on main so you will need to merge/rebase and generate the lockfile again.

jmsmkn (commented):
Looks good, don't the `allauth_2fa.urls` also need to be removed from the main router? After doing that, have you tested that everything gets set correctly when adding a 2fa device, logging in with that device or recovery code also works? What about testing the flow with social login?

app/tests/conftest.py
amickan: I thought we might want to use this for other tests in the future, but at the moment, it can also just be defined in the one test, yes.

app/grandchallenge/core/middleware.py
amickan: I think it would be ok to do that. It would mean that all authentication related views are always accessible, which I guess is fair. The MFA urls I didn't add are no blockers, they only work after MFA set-up, which is why I didn't add them, but there is no harm in adding them.

app/grandchallenge/profiles/migrations/0019_migrate_mfa.py
amickan: Yes, all tested and works.

amickan:
> don't the `allauth_2fa.urls` also need to be removed from the main router?

I thought that could wait until the next PR when I remove the dependency?

jmsmkn:
> > don't the `allauth_2fa.urls` also need to be removed from the main router?
> 
> I thought that could wait until the next PR when I remove the dependency?

I think you'll need to remove them now - if you don't people could add a TOTP device etc to the old tables after the migration has been run.

app/grandchallenge/core/middleware.py
jmsmkn: I was curious about how to do this and GPT4 told me, might be handy and more robust to the frequent breaking changes in allauth:

```python
from django.urls import URLResolver, URLPattern
from django.utils.module_loading import import_string

def list_url_names(urlconf_name, prefix=''):
    """
    Recursively lists all URLs in a URL configuration and their pattern names.
    """
    # Import the URL configuration module dynamically
    try:
        urls = import_string(urlconf_name)
    except ImportError:
        # This will handle the case where the URL configuration is not a module but an object
        urls = urlconf_name

    # Iterate through all URL patterns
    for pattern in urls.urlpatterns:
        if isinstance(pattern, URLPattern):
            # This is a final URL pattern
            pattern_name = pattern.name
            if pattern_name:
                print(f"{prefix}{pattern.pattern.regex.pattern} -> {pattern_name}")
        elif isinstance(pattern, URLResolver):
            # This pattern is an include, so recurse into the included URLconf
            new_prefix = pattern.pattern.regex.pattern
            if prefix:
                new_prefix = f"{prefix}{new_prefix}"
            list_url_names(pattern.urlconf_name, new_prefix)

# Specify the URL configuration of 'allauth.urls'
list_url_names('allauth.urls')
```

jmsmkn: Maybe we would also want to add a test so that we're sure the set of allowed urls is sensible, we at least then get alerted to upstream changes.

amickan: Ah nice, I would have just done `*[pattern.name for pattern in all_auth_urls]`, importing `from allauth.account.urls import urlpatterns as account_urlpatterns` but what GPT4 proposes looks more robust and adding a test is a good idea as well.

amickan:
> > > don't the `allauth_2fa.urls` also need to be removed from the main router?
> > 
> > 
> > I thought that could wait until the next PR when I remove the dependency?
> 
> I think you'll need to remove them now - if you don't people could add a TOTP device etc to the old tables after the migration has been run.

Ah right, good point. But then I can also remove it from the list of installed apps at the same time, right?

amickan:
> What about testing the flow with social login?

I remember those tests being very difficult to get to work and I see that they were removed in https://github.com/comic/grand-challenge.org/pull/2735. What was the reason for removing them there?

jmsmkn:
> > What about testing the flow with social login?
> 
> I remember those tests being very difficult to get to work and I see that they were removed in #2735. What was the reason for removing them there?

They were broken with Django All Auth 0.52 and I couldn't work out how to fix them. Those were covering functionalty that should be tested in Django All Auth. With my comment today I just meant a sanity check test in your dev environment. I think they have a dummy provider now that should help with development tests https://github.com/pennersr/django-allauth/blob/main/allauth/socialaccount/providers/dummy/provider.py

jmsmkn:
> > > > don't the `allauth_2fa.urls` also need to be removed from the main router?
> > > 
> > > 
> > > I thought that could wait until the next PR when I remove the dependency?
> > 
> > 
> > I think you'll need to remove them now - if you don't people could add a TOTP device etc to the old tables after the migration has been run.
> 
> Ah right, good point. But then I can also remove it from the list of installed apps at the same time, right?

You need to keep it in the installed apps so that the migration will run and the data still appears in the admin for validation.

amickan:
> Those were covering functionalty that should be tested in Django All Auth.

There was one test that checked the 2fa required workflow for staff users with a social login. I think that's the test we would want now as well. Just tried adding that back, but can't get it to work. I'm stuck at mocking the login... the DummyProvider doesn't help with that I think. I must admit that the social login stuff is rather opaque to me, so I'm afraid it will take some time to understand what happens and what needs testing exactly.

amickan:
> taken from the allauth tests



> If I understand it correctly it should, then you wouldn't need any mocks at all, this is taken from the allauth tests:
> 
> ```python
> def test_login(client, db):
>      resp = client.post(reverse("dummy_login"))
>      assert resp.status_code == 302
>      assert resp["location"] == reverse("dummy_authenticate")
>      resp = client.post(
>          reverse("dummy_authenticate"),
>          {"id": "123", "email": "a@b.com", "email_verified": True},
>      )
>      assert resp.status_code == 302
>      assert resp["location"] == settings.LOGIN_REDIRECT_URL
>      get_user_model().objects.filter(email="a@b.com").exists()
> ```

Cool. Hadn't seen that yet. Somehow it can't resolve the dummy_login url though. Adding the dummy provider to the test settings under SOCIALACCOUNT_PROVIDERS doesn't fix it. I will take a look tomorrow.

jmsmkn:
I think you need to add it to INSTALLED_APPS in `tests/settings.py`.

amickan:
> I think you need to add it to INSTALLED_APPS in `tests/settings.py`.

This plus adding it to `SOCIALACOUNT_PROVIDERS` did the trick.

app/tests/settings.py
```diff
@@ -58,4 +58,7 @@
 # Ensure custom models for testing are found
 INSTALLED_APPS += [  # noqa F405
     "tests.core_tests",
+    "allauth.socialaccount.providers.dummy",
```
jmsmkn, line 61: Something for a separate PR - we should check that this provider is not installed in production by using a check (see `check_sagemaker_is_used`). I think that this provider is really dangerous, it allows you to log in as any user by just providing their email address, right?

jmsmkn (approved):
Great work!

jmsmkn (approved):

PR #3340, opened 2024-05-03
Add deployment check that dummy provider is not configured

amickan:


app/grandchallenge/components/__init__.py
```diff
@@ -23,3 +23,21 @@ def check_sagemaker_is_used(app_configs, **kwargs):
         )
 
     return errors
+
+
+@register(deploy=True)
+def check_dummy_provider_is_not_used(app_configs, **kwargs):
```
jmsmkn, line 29: I think it's also good to check INSTALLED_APPS in the same check. Also, a more logical place for this code might be profiles with the rest of the auth stuff or core.

app/grandchallenge/components/__init__.py
```diff
@@ -23,3 +23,21 @@ def check_sagemaker_is_used(app_configs, **kwargs):
...
+    from django.conf import settings
+
+    errors = []
+
+    if "dummy" in settings.SOCIALACCOUNT_PROVIDERS.keys():
+        errors.append(
+            Error(
+                "The dummy social account provider is configured. This provider should only be used for testing purposes and not in production.",
+                hint="Remove the dummy provider from SOCIALACCOUNT_PROVIDERS in your settings.",
+                id="grandchallenge.components.E002",
```
jmsmkn, line 39: If you do move this update the id as it would no longer be a grandchallenge.components check.

PR #3341, opened 2024-05-03
Display set titles

chrisvanrun:
Part of the pitch:
- https://github.com/DIAGNijmegen/rse-roadmap/issues/313#event-12465816350

This adds a `title` field which is guaranteed to be unique for a display set within a reader study.

In addition:
- Reader study copies now also take into account the ordering of display sets
- Deleting list view of selected display set now shows the titles
- Negative display set ordering is now invalid on the form level

Some odd saves were going on concerning `order`. An initial version had a `slug` as a unique counter to the `title`. This `slug` build-up required the correct `order` at the time of saving. While the `slug` is no longer needed in this version: I left the saving changes as I think any data checking should have the correct order involved.

Once this PR passes the first inspection I'll throw out a new one with similar changes for the archive items.

https://github.com/DIAGNijmegen/rse-roadmap/assets/7871310/8f96065c-1412-4955-8e7a-a1718ac6dceb


app/grandchallenge/reader_studies/models.py
```diff
@@ -827,6 +835,13 @@ def assign_permissions(self):
 
     class Meta:
         ordering = ("order", "created")
+        constraints = [
+            models.UniqueConstraint(
```
amickan, line 839: While we're at it, should we also add a uniqueness constraint for `order` and `reader_study`? At the moment, you can have multiple display sets with the same order.

app/grandchallenge/reader_studies/views.py
```diff
@@ -1300,15 +1311,23 @@ def new_interface_url(self):
     def process_data_for_object(self, data):
         """Updates the display set"""
         instance = self.object
-        for ci_slug, new_value in data.items():
-            if ci_slug == "order":
+        non_civ_data_keys = ("order", "title")
+
+        save = False
+        for key in non_civ_data_keys:
+            value = data.get(key, empty)
```
amickan, line 1318: Why not keep with the default (return `None` if the key is not in the data dictionary)?

app/grandchallenge/reader_studies/views.py
```diff
@@ -1410,13 +1429,20 @@ def new_interface_url(self):
 
     def process_data_for_object(self, data):
         """Creates a display set"""
-        instance = DisplaySet.objects.create(reader_study=self.base_object)
-        instance.order = data.pop("order")
-        instance.save()
-        for slug in data:
+        non_civ_data_keys = ("order", "title")
```
amickan, line 1432: This is repeated earlier. Might be good to refactor so we only need to change this in one place in the future.

app/grandchallenge/reader_studies/models.py
chrisvanrun: I think that is not too badly needed. Order uniqueness would require a migration and is out of scope =P.

app/grandchallenge/reader_studies/views.py
chrisvanrun: This is future-proof: it accounts for nullable fields. Generally, whenever doing data retrieval it's better to reply on `empty` to detect it's no-presence rather than `None`.

app/grandchallenge/reader_studies/views.py
chrisvanrun: Good point, I opted to make the retrieval of these fields dynamic in that it's the existing fields minus the interface fields.

app/grandchallenge/components/forms.py
```diff
@@ -106,6 +106,10 @@ class MultipleCIVForm(Form):
         SelectUploadWidget,
     }
 
+    @property
+    def interface_field_names(self):
+        yield from self.base_obj.values_for_interfaces.keys()
```
amickan, line 111: Wouldn't this make more sense as a model property?

chrisvanrun: I get why you'd think that initially. That was my first go-to as well. 

However, I quickly realized that the data filtering is dependent on the form, not on the model. This is showcased by the fact that we are adding custom fields that just HAPPEN to have the same name as the model (i.e. 'order' and 'title'). Would you agree?

chrisvanrun: Also, the interface fields are generated in the form. If we'd ever change the way the fields are named or generated: we should also change the references to the fields there.

chrisvanrun:
Ugh, failing tests. On it!

app/grandchallenge/components/forms.py
amickan: `interface_field_names` returns the slugs for all existing interfaces for a reader study / archive, so those are fixed and do not depend on the form. I think you move this to `ValuesForInterfacesMixin`. The `non_civ_field_names` should indeed remain on the form, I agree with that.

chrisvanrun: Mhh, going to do a bit heavier rework. WIP.

app/grandchallenge/components/forms.py
```diff
@@ -106,6 +106,23 @@ class MultipleCIVForm(Form):
...
+        }
+
+    @property
+    def cleaned_non_civ_data(self):
+        cleaned_civ_data = self.cleaned_civ_data
+        return {
+            k: v
+            for k, v in self.cleaned_data.items()
+            if k not in cleaned_civ_data
+        }
```
amickan, line 124: I'm not sure, but I think this might not work. It could contain new CIVs (those that are dynamically added to the form) and those should go into `cleaned_civ_data`.

amickan: Maybe doing it this way is too complex, and resorting to what you had initially, but then using a variable that you define in one place to define the keys to skip over, is best.

chrisvanrun: Ooooooh, I hadn't thought of new interfaces =/! Yeah, there is no way to keep track of that in a simple matter!

amickan (approved):
LGTM!

PR #3343, opened 2024-05-06
Add indicators for HTMX loading and POSTing

chrisvanrun:
This is a typical Friday afternoon fix.

This addresses a weird issue I had locally. I'd load the update page and click save and nothing would happen... 

By default, the form does a request. HTMX loading is deferred (something which is good for speed), once HTMX passes over the DOM nodes the behaviour of the buttons changes.

This PR disables the HTMX-related buttons until HTMX is initiated. It adds an indicator while HTMX is loading and, for good measure, it now also has an indicator to show HTMX is sending a request.

These attribute-driven changes can readily be added to future interactions.

In addition, I've changed the titles of the Display set list to account for tooltips not working on disabled UI elements.

https://github.com/comic/grand-challenge.org/assets/7871310/e6b953ad-5fb3-44c2-95cb-24a66a3ae2ff


chrisvanrun:
Note that the HTMX.onLoad event is NOT the same as HTMX having processed all DOM nodes or is finished loading. Rather that links to the content being loaded, which is not the same =/.

An alternative solution, that I just now realized, is loading the actual form using HTMX... but that seems very slow.

app/grandchallenge/components/templates/components/civ_set_form.html
```diff
@@ -69,8 +85,10 @@ <h3>
 {% block script %}
     {{ block.super }}
     {{ form_media }}
-    <script type="module" src="{% static 'components/js/create_extra_interfaces.mjs' %}" type="text/javascript"></script>
-    <script type="module" src="{% static 'components/js/autocomplete_htmx.mjs' %}" type="text/javascript"></script>
+    <script type="module" src="{% static 'components/js/create_extra_interfaces.mjs' %}"></script>
+    <script type="module" src="{% static 'components/js/autocomplete_htmx.mjs' %}"></script>
+    <script type="text/javascript" src="{% static "components/js/htmx/enable-me.js" %}"></script>
+    <script type="text/javascript" src="{% static "components/js/htmx/hide-me.js" %}"></script>
```
jmsmkn, line 91: New JS files should be modules (`type="module") and use ".mjs".

app/grandchallenge/core/static/css/core.css
```diff
@@ -218,3 +218,15 @@ a[aria-expanded=true] .fa-chevron-right {
...
+/* htmx */
+.htmx-indicator {
+ display:none;
+}
+.htmx-request .htmx-indicator {
+    display:inline-block;
+}
+.htmx-request.htmx-indicator {
+    display:inline-block;
+}
```
jmsmkn, line 232: Why do you need special class definitions for this? Can't you use `d-block` and `d-inline-block`?

app/grandchallenge/components/templates/components/civ_set_form.html
```diff
@@ -56,11 +57,26 @@ <h3>
...
               hx-target="#new-interfaces"
               hx-swap="beforeend"
-      >Add value for new interface</button>
+              hx-indicator="#hx-add-interface-indicator"
+              hx-enable-me
+              disabled
+      >
+          Add value for new interface
+          <span id="hx-add-interface-indicator" class="htmx-indicator spinner-border spinner-grow-sm" role="status"></span>
+          <span class="spinner-grow spinner-grow-sm" role="status" hx-hide-me></span>
```
jmsmkn, line 66: The spinner grow doesn't communicate loading for me. I think it would be better if it was the actual spinner, that the content tells you "Loading...". Like it is on the bootstrap example page https://getbootstrap.com/docs/4.2/components/spinners/#buttons. The real button text then gets swapped in when it has loaded.

app/grandchallenge/core/templates/grandchallenge/partials/script.html
```diff
@@ -16,4 +16,5 @@
 {% endcompress %}
 
 {# Htmx #}
-<script src="{% static 'vendored/htmx/htmx.min.js' %}" defer></script>
+<script id="htmx-script" src="{% static 'vendored/htmx/htmx.min.js' %}" defer></script>
+<script src="{% static 'components/js/htmx/loaded.js' %}"></script>
```
jmsmkn, line 20: Needs a type and for it to be a JS module.

app/grandchallenge/components/templates/components/civ_set_form.html
chrisvanrun: I had dislike replacing the text completely because you then don't know what you are waiting for... each button has a placeholder.

Just played around a bit and the closest I got is:
image

Which feels way too bulky. So I'll throw in the proverbial towel and go for a `Loading ...` + spinner setup.

jmsmkn: I think doing what is shown on the bootstrap page is fine.

app/grandchallenge/core/static/css/core.css
chrisvanrun: We need to use special classes to select the correct element at the correct time. We could use d-none instead of `htmx-indicator` but that seems hacky.

Reference: https://htmx.org/attributes/hx-indicator/

app/grandchallenge/components/templates/components/civ_set_form.html
chrisvanrun: Now done.

jmsmkn:
The spinners should always be left of the text.

chrisvanrun:
> The spinners should always be left of the text.

Any reason? My rationale for having them on the right side is to keep stability in the UIX. Adding them to the left causes the textual context of the button to jump around.

(Seem I had forgotten to push the latest changes, my bad.)

chrisvanrun:
I've swapped them, it's not that bad having them at the left. The jumping sort gives the effect of a button popping out after being pressed.

amickan (approved):
LGTM!

PR #3345, opened 2024-05-07
Archive Item Titles

chrisvanrun:
Part of the pitch:
- https://github.com/DIAGNijmegen/rse-roadmap/issues/313#event-12465816350

This is a copy of what was done for Display Items. I've opted to abstract the CIV set creation and update form data processing into `CIVSetCreateViewMixin` and `CIVSetCreaViewMixin` respectively as to stay DRY.


chrisvanrun:
The Django test seemed to fail randomly on not being able to find the `PrefixConverter`. I assume it has to do with dependency differences on `main` vs this branch.

Merging main and re-running.

chrisvanrun:
> The Django test seemed to fail randomly on not being able to find the `PrefixConverter`. I assume it has to do with dependency differences on `main` vs this branch.
> 
> Merging main and re-running.

Well, Chris of the past, that worked!

app/grandchallenge/archives/forms.py
```diff
@@ -225,3 +227,44 @@ def save(self, *args, **kwargs):
...
+            raise ValidationError(
+                "An archive item already exists with this title"
+            )
+        return title
+
+    def _title_query(self, title):
+        return ArchiveItem.objects.filter(
+            title=title,
+            archive=self.base_obj,
+        )
```
amickan, line 257: I think this could be refactored as well.

app/grandchallenge/reader_studies/views.py
```diff
@@ -395,7 +396,7 @@ class ReaderStudyDisplaySetList(CivSetListView):
         *CivSetListView.columns,
     ]
 
-    default_sort_column = 3
+    default_sort_column = 2
```
amickan, line 399: Why this change? Now it will be ordered by pk. I think we should keep the default ordering by order.

app/grandchallenge/reader_studies/models.py
```diff
@@ -798,11 +800,16 @@ class DisplaySet(CIVForObjectMixin, UUIDModel):
...
+        result = f"{self.title!r}" if self.title else str(self.pk)
+
+        values = []
+        for value in self.values.all():
+            if value.image:
+                values.append(value.image.name)
+            if value.file:
+                values.append(os.path.basename(value.file.file.name))
+        values = ", ".join(values)
+        return f"{result} ({values})" if values else f"Empty {result}"
```
amickan, line 812: I guess you changed this so that it's the same as for ArchiveItems? Would be worth refactoring as well then. 

I'm not 100% sure about the format though. Do we really want to add all the values in the title? There is no limitation on how many values an item can have, this could get very long. Also, the pk is only added if there is no title. I feel like that pk should always be there, right?

app/tests/archives_tests/test_forms.py
```diff
@@ -201,6 +205,46 @@ def create_archive():
...
+    form = form_class(
+        user=user,
+        instance=instance,
+        base_obj=archive,
+        data={
+            "title": new_title,
+        },
+    )
+
+    assert form.is_valid() is expected_validity
```
amickan, line 245: This test could also be refactored if you refactor the form.

app/tests/archives_tests/test_models.py
```diff
@@ -15,3 +16,40 @@ def create_archive_items_for_images(images, archive):
         civ = ComponentInterfaceValueFactory(image=image)
         ai = ArchiveItemFactory(archive=archive)
         ai.values.add(civ)
+
+
+@pytest.mark.django_db(transaction=True)
+def test_archive_item_set_title():
```
amickan, line 22: This is the same for reader studies, right? Might be nice to combine the tests as well and parametrize them accordingly.

app/grandchallenge/archives/models.py
```diff
@@ -290,15 +291,19 @@ class ArchiveItem(CIVForObjectMixin, UUIDModel):
     values = models.ManyToManyField(
         ComponentInterfaceValue, blank=True, related_name="archive_items"
     )
+    title = models.CharField(max_length=255, default="", blank=True)
```
amickan, line 294: I think theoretically all title related model definitions, the field, the constraint and the string method could be a mixin that you reuse for items and display sets. The only tricky part is getting the base object field name in the constraint definition. I think that would need to be done in the init method. Just a suggestion though

app/tests/archives_tests/test_models.py
jmsmkn: Please do not add any transaction tests, they're hideously slow.

chrisvanrun: I am a bit rusty with atomic interactions. These tests fail on the DB level (Integrity constraints) and the transaction kept hanging as a result. We could use a context manager, maybe (?), but I am not too sure if it's any quicker.

app/grandchallenge/reader_studies/views.py
chrisvanrun: You are right, I neglected that display sets have the addition of order in the mix. It was originally meant to have it equal between Archives and Display Sets. Reverted!

app/grandchallenge/archives/models.py
chrisvanrun: This Needs a weird construct around the unique restraint, which makes it a lot less readable. Not worth the damage to readability imho.

app/tests/archives_tests/test_models.py
chrisvanrun: Fixed with contextmanagering atomic transmissions. (Wow, that is a cool sentence.)

app/tests/archives_tests/test_models.py
```diff
@@ -15,3 +16,40 @@ def create_archive_items_for_images(images, archive):
         civ = ComponentInterfaceValueFactory(image=image)
         ai = ArchiveItemFactory(archive=archive)
         ai.values.add(civ)
+
+
+@pytest.mark.django_db(transaction=True)
```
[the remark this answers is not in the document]
chrisvanrun: Not quite: https://pytest-django.readthedocs.io/en/latest/helpers.html#pytest-mark-django-db-request-database-access

Anyhow, I've removed the tag for now: using contextmanagers reduced local runtime from 17 sec to 11 sec.

app/tests/archives_tests/test_forms.py
chrisvanrun: Done! Or rather, I have created a test for just the refactored Mixin

app/grandchallenge/reader_studies/models.py
chrisvanrun: I've opted to standardize things a bit. Each CIV is now shown as an interface.title + title, this was already the case for the list view but I copied this to the other representations.

civ_set_list_deletion


Furthermore, I've taken the liberty of updating the single civset deletion template it can have all the CIVs shown in a data table format:

single_civ_deletion

app/grandchallenge/archives/forms.py
chrisvanrun: Done!

app/grandchallenge/reader_studies/models.py
chrisvanrun: Note: I've copied the Warning to the list deletion as well.

app/tests/reader_studies_tests/test_models.py
```diff
@@ -697,15 +698,17 @@ def test_display_set_title():
 
     # Duplication attempt via edit
     ds1.title = ds0.title
-    with pytest.raises(IntegrityError):
-        ds1.save()
+    with transaction.atomic():
```
jmsmkn, line 701: Why are atomic transactions needed now?

app/grandchallenge/core/forms.py
```diff
@@ -51,3 +52,45 @@ def __init__(self, *args, **kwargs):
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.fields = {
+            "title": CharField(
+                required=False,
+                initial=self.instance and self.instance.title or "",
+                max_length=self.Meta.model._meta.get_field("title").max_length,
+            ),
+            **self.fields,
```
jmsmkn, line 73: Why are you resetting all fields here, not just the title?

app/grandchallenge/reader_studies/forms.py
```diff
@@ -653,50 +655,37 @@ def save_answers(self):
...
-        non_civ_fields = ("title", "order")
-
-    def clean_title(self):
-        title = self.cleaned_data.get("title")
-        if title and self._title_query(title).exists():
-            raise ValidationError(
-                "A display set already exists with this title"
-            )
-        return title
+            **self.fields,
```
jmsmkn, line 678: Again, strange construction, why?

app/grandchallenge/archives/urls.py
```diff
@@ -10,7 +10,7 @@
     ArchiveItemInterfaceCreate,
     ArchiveItemsList,
     ArchiveItemsToReaderStudyUpdate,
-    ArchiveItemUpdate,
+    ArchiveItemUpdateView,
```
jmsmkn, line 13: `ArchiveItemUpdateView` is a CBV, and has a method `.as_view` that produces a view. So, this should be kept as `ArchiveItemUpdate`.

app/grandchallenge/components/views.py
```diff
@@ -244,6 +244,52 @@ def new_interface_url(self):
...
+        instance = self.create_object(
+            non_civ_data={k: data.get(k) for k in non_civ_data_keys}
+        )
+
+        for key in data:
+            if key in non_civ_data_keys:
+                continue
+            instance.create_civ(
+                ci_slug=key, new_value=data[key], user=self.request.user
+            )
```
jmsmkn, line 289:
```suggestion
            if key not in non_civ_data_keys:
                instance.create_civ(
                    ci_slug=key, new_value=data[key], user=self.request.user
                )

```

app/grandchallenge/components/views.py
```diff
@@ -244,6 +244,52 @@ def new_interface_url(self):
...
+
+        return instance
+
+
+class CIVSetCreateViewMixin(CIVSetFormMixin):
+    def create_object(self, non_civ_data):
```
jmsmkn, line 273: Seems a bit strange to have a special method for object creation here. Cant this be constructed in a general way without the consumer having to override this method?

app/grandchallenge/components/views.py
```diff
@@ -244,6 +244,52 @@ def new_interface_url(self):
...
 
+class CIVSetUpdateViewMixin(CIVSetFormMixin):
+    def process_data_for_object(self, data):
+        """Updates the object"""
+        instance = self.object
+        non_civ_data_keys = self.form_class.Meta.non_civ_fields
+
+        save = False
+        for key in non_civ_data_keys:
+            value = data.get(key, empty)
```
jmsmkn, line 255: What is the value of empty on the first pass of this line?

app/grandchallenge/components/views.py
jmsmkn: This also seems like a duplicate fragment from `CIVSetFormMixin`, why duplicate rather than use inheritance?

app/tests/reader_studies_tests/test_models.py
chrisvanrun: They were covered by the `transaction=True` in the django_db decorator first. Without it, the integrity errors cause havoc on the follow-up tests. Now that I think about it, there is the alternative to split the tests forcibly and have the IntegrityError context manager as the final call on the test.

app/grandchallenge/core/forms.py
chrisvanrun: The order of the form is dependent on the order of the keys in the dict. I've made it more explicit by calling `Form.order_fields`, but in essence it does the same.

app/grandchallenge/archives/urls.py
chrisvanrun: I am confused. The default `View` Django class has the `as_view`. Not calling this a View seems at odds with the Django framework. Moreover, it directly inherits from the Django's `FormView`.

app/grandchallenge/components/views.py
chrisvanrun: The crux is the base_object reference i.e. 'archive' vs 'reader_study'. I'll see if I can cook something alternatively up, using the model directly and passing the linkage information somewhere.

chrisvanrun:
Draft, since I am in the middle of a refactor.

chrisvanrun:
Heavy refactor. I've now pushed the processing of the data into the forms, leaning more heavily on the model's definition of civ_set. I think this is the DRY-est it has been.

For instance, the forms for DisplaySet creating and updating are now an interesting mix of Mixins:
```Python
class DisplaySetCreateForm(
    DisplaySetFormMixin,
    UniqueTitleCreateFormMixin,
    CIVSetCreateFormMixin,
    MultipleCIVForm,
):
    pass


class DisplaySetUpdateForm(
    DisplaySetFormMixin,
    UniqueTitleUpdateFormMixin,
    CIVSetUpdateFormMixin,
    MultipleCIVForm,
):
        if not self.instance.is_editable:
            for _, field in self.fields.items():
                field.disabled = True
```

app/grandchallenge/components/models.py
```diff
@@ -1962,6 +1962,30 @@ def update_size_in_storage(self):
             self.size_in_registry = self.calculate_size_in_registry()
 
 
+class CIVSetContainerMixin:
```
amickan, line 1965: Why "container" in the name here? I find that a bit confusing. Not the catchiest, but `CIVSetStringRepresentationMixin` describes clearly what it does. What do you think?

app/grandchallenge/archives/views.py
```diff
@@ -590,14 +598,11 @@ def new_interface_url(self):
...
-        for slug in data:
-            instance.create_civ(
-                ci_slug=slug, new_value=data[slug], user=self.request.user
-            )
-        return instance
+    def create_object(self, non_civ_data):
+        return ArchiveItem.objects.create(
+            archive=self.base_object,
+            **non_civ_data,
+        )
```
amickan, line 605: Don't you have a method for that on the base object now that you could reuse here?

app/grandchallenge/components/views.py
```diff
@@ -244,6 +244,52 @@ def new_interface_url(self):
...
+    def create_object(self, non_civ_data):
+        raise NotImplementedError
+
+    def process_data_for_object(self, data):
+        """Creates the object, including interfaces"""
+        non_civ_data_keys = self.form_class.Meta.non_civ_fields
+
+        instance = self.create_object(
+            non_civ_data={k: data.get(k) for k in non_civ_data_keys}
+        )
```
amickan: Can't you just call `self.base_object.create_civ_set(data=non_civ_data)`? Then you wouldn't need the extra `create_object` method and would save some duplication.

chrisvanrun: I think you commented on an outdated version. The default is to indeed use that method.

app/grandchallenge/archives/views.py
chrisvanrun: You are right, but in this case this was a left over, good find!

app/grandchallenge/components/models.py
chrisvanrun: Works.

chrisvanrun: I thought it could function as a general spot to put stuff in for CIVSets but I guess that should likely not be a Mixin =/

app/grandchallenge/reader_studies/views.py
```diff
@@ -1427,24 +1408,11 @@ def new_interface_url(self):
...
-        """Creates a display set"""
-        non_civ_data_keys = self.form_class.Meta.non_civ_fields
-
-        instance = DisplaySet.objects.create(
-            reader_study=self.base_object,
-            **{k: data.get(k) for k in non_civ_data_keys},
+    def create_object(self, non_civ_data):
+        return DisplaySet.objects.create(
+            reader_study=self.base_object, **non_civ_data
         )
```
amickan, line 1414: This is not used anywhere now, is it?

app/grandchallenge/archives/forms.py
```diff
@@ -225,3 +232,37 @@ def save(self, *args, **kwargs):
             {"interface_pk": self.cleaned_data["interface"].pk}
         )
         return super().save(*args, **kwargs)
+
+
+class ArchiveItemFormMixin(MultipleCIVForm):
```
amickan, line 237: Does this need to inherit from `MultipleCIVForm`? Now `ArchiveItemCreateForm` basically inherits `MultipleCIVForm` twice, once through `ArchiveItemCreateForm` and once directly. Have you tested that this works as expected?

app/grandchallenge/components/views.py
```diff
@@ -126,15 +126,15 @@ class MultipleCIVProcessingBaseView(
...
 
-    def process_data_for_object(self, data):
-        raise NotImplementedError
-
     @property
     def base_object(self):
         raise NotImplementedError
 
+    def process_data(self, data):
+        raise NotImplementedError
```
amickan, line 134: I don't know if I'm looking at an outdated version again, but I don't see his being implemented on any of the views that subclass this base view, and it's also not used further, so this should go.

app/grandchallenge/reader_studies/views.py
chrisvanrun: Quite right, that other instance should really have made me look for the other one! Thanks!

app/grandchallenge/archives/forms.py
chrisvanrun: Def doesn't need it. Refactor oversight on removing it. The DisplaySet's homologue doesn't use it either.

amickan (dismissed):
LGTM!

app/tests/reader_studies_tests/test_models.py
chrisvanrun: I think splitting the tests would make for a really awkward set of tests. So I think the atomic transactions are OK?

app/grandchallenge/archives/urls.py
chrisvanrun: Not sure if this rename is still needed?

app/tests/reader_studies_tests/test_models.py
jmsmkn: I think that it is better to split the tests out. Remember the three A's (taken from https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/):

1. **Arrange** inputs and targets. Arrange steps should set up the test case. Does the test require any objects or special settings? Does it need to prep a database? Does it need to log into a web app? Handle all of these operations at the start of the test.
2. **Act** on the target behavior. Act steps should cover the main thing to be tested. This could be calling a function or method, calling a REST API, or interacting with a web page. Keep actions focused on the target behavior.
3. **Assert** expected outcomes. Act steps should elicit some sort of response. Assert steps verify the goodness or badness of that response. Sometimes, assertions are as simple as checking numeric or string values. Other times, they may require checking multiple facets of a system. Assertions will ultimately determine if the test passes or fails.

We tend to have tests that do too much, which makes debugging them annoying - and in this case relies on turning transactions on which is slow. Ideally a test tests only one bit of functionality. If you're worried about duplication you can create a scope limited fixture for the tests which encapsulates the set up.

app/grandchallenge/archives/urls.py
jmsmkn: I prefer not to include it for the reasons given before. Most of the other classes imported from this module do not have it, I would rather we stay consistent with that.

PR #3353, opened 2024-05-24
Fix not having Text widget set when creating questions

chrisvanrun:
This PR sets the default of the question answer_type to TEXT on the model level. It was already indirectly the default via the form and being the first option. For the widget choices, this was not clear during the initial render so the options were left empty.

The alternative is to fix it on the Form level but it seems better to make it explicit on the model level.


jmsmkn:
> The alternative is to fix it on the Form level but it seems better to make it explicit on the model level.

Actually, I think this is better to fix on the form level as it is a problem with the form. The only change I think that model field should have is `blank=False`, we want users of the field to be explicit about setting the question type.

chrisvanrun:
Will implement in a different PR.

PR #3354, opened 2024-05-24
Make the answer_type choice explicit for questions

chrisvanrun:
This PR will have the question creation form start with a blank answer.

As a side effect, it fixes a bug with the widget choices being empty at first load. Now them being empty is correct =P

### DIFF
--- a/app/grandchallenge/reader_studies/models.py
+++ b/app/grandchallenge/reader_studies/models.py
@@ -1040,6 +1040,7 @@ class QuestionWidgetKindChoices(models.TextChoices):
 
 
 ANSWER_TYPE_TO_QUESTION_WIDGET = {
+    "": [],  # Blank answer type
     AnswerType.TEXT: [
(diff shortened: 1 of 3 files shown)


app/grandchallenge/reader_studies/models.py
```diff
@@ -1040,6 +1040,7 @@ class QuestionWidgetKindChoices(models.TextChoices):
 
 
 ANSWER_TYPE_TO_QUESTION_WIDGET = {
+    "": [],  # Blank answer type
```
amickan, line 1043: Why is this necessary? If answer type is left blank, it will be `None` and we return `ComponentInterface.objects.none()`, right?

chrisvanrun: I could have sworn this was to resolve a bug I discovered updating from one answer type to the next. However, I cannot reproduce it.

I'll just remove it.

Also: this resolves around widget types not CIs =)

amickan (approved):

PR #3359, opened 2024-05-30
Generalize section rendering for workstation configuration pages

pkcakeout:
## Problem

There currently is huge hand-written template html file to add nice sections to the workstation configuration detail page:

image

However, when clicking "edit", the edit form does not have these sections, even though they are quite handy for such a massive page with many many different configuration options. Also the order of entries differs from the details page. A very confusing situation.

## Solution

- I moved the grouping into the python code, specifically a module called `visual_field_ordering`. This defines a list of names that should be rendered on the different pages (edit form vs details) as well information about the sections 
- The hand-written details template was replaced with a generic one that includes special handling for the different field types using some added template-filters and tags


jmsmkn:
Replaced by #3366

PR #3360, opened 2024-05-30
Algorithm model

amickan:
Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/297 

- Adds a `AlgorithmModel`, and adds it to `Job` and `Evaluation`. 
- Adds views and forms to create and view a model. 
- Updates the checks on submission and job creation


app/grandchallenge/algorithms/views.py
```diff
@@ -557,21 +560,50 @@ def create_upload_session(image_files):
...
+            )
+            .filter(inputs_match_count=input_interface_count)
+            .exists()
+        ):
+            form.add_error(
+                None,
+                "A result for these inputs with the current image "
+                "and model already exists.",
+            )
+            return self.form_invalid(form)
```
amickan, line 589: This is not the way to do it, but it requires a more extensive refactoring of CIV creation and validation for jobs to fix. That should be a seperate PR.

app/grandchallenge/algorithms/models.py
```diff
@@ -484,6 +502,16 @@ def public_test_case(self):
         except AttributeError:
             return False
 
+    def form_field_label(self):
+        title = f"{self.title}"
+        if self.active_image:
+            title += f" (Active image: {self.active_image.pk})"
+        if self.active_model:
+            title += f" (Active model: {self.active_model.pk})"
```
jmsmkn, line 510: Should we also include the user provided comment about the algorithm image and/or model?

app/grandchallenge/algorithms/models.py
```diff
@@ -619,12 +647,148 @@ def spent_credits(self, user):
...
+        validators=[validate_gzip_mimetype],
+    )
+    model = models.FileField(
+        blank=True,
+        upload_to=algorithm_models_path,
+        validators=[ExtensionValidator(allowed_extensions=(".tar.gz",))],
+        help_text=(
+            ".tar.gz file of the algorithm model that will be extracted to /opt/ml/model/ during inference"
+        ),
+        storage=private_s3_storage,
```
jmsmkn, line 687: This should be on the protected storage as we will need to serve them later, see https://github.com/DIAGNijmegen/rse-roadmap/issues/301.

app/grandchallenge/algorithms/models.py
```diff
@@ -619,12 +647,148 @@ def spent_credits(self, user):
...
+
+        if adding:
+            self.assign_permissions()
+
+    def assign_permissions(self):
+        # Editors can view this algorithm model
+        assign_perm(
+            f"view_{self._meta.model_name}", self.algorithm.editors_group, self
+        )
+        # Editors can change this algorithm model
```
jmsmkn, line 715: Why are change permissions needed?

app/grandchallenge/algorithms/models.py
```diff
@@ -619,12 +647,148 @@ def spent_credits(self, user):
...
+        if not self.import_status == ImportStatusChoices.FAILED:
+            raise RuntimeError("Cannot delete model from completed upload.")
+
+        s3_client = boto3.client(
+            "s3",
+            endpoint_url=settings.AWS_S3_ENDPOINT_URL,
+        )
+        s3_client.delete_object(
+            Bucket=self.model.storage.bucket_name, Key=self.model.name
+        )
```
jmsmkn, line 746:
```suggestion
        self.model.delete()
```

app/grandchallenge/algorithms/models.py
amickan: To be able to change the active model later. I haven't added that view yet as I thought that could easily be a seperate PR, but I thought the change permissions would be good to add here already.

app/grandchallenge/algorithms/models.py
amickan: Good idea. Before the image and model pks would probably be best.

app/tests/evaluation_tests/test_forms.py
```diff
@@ -222,12 +223,61 @@ def test_algorithm_with_permission(self):
...
+        p = Phase.objects.get(pk=p.pk)
+
+        ai = AlgorithmImageFactory(
+            is_manifest_valid=True,
+            is_in_registry=True,
+            is_desired_version=True,
+            algorithm=alg,
+        )
+        am = AlgorithmModelFactory(algorithm=alg, is_desired_version=True)
+        AlgorithmJobFactory(algorithm_image=ai, status=4)
```
jmsmkn, line 270: It is better not to use the choice values as 4 is not informative to a future reader.

app/tests/components_tests/test_amazon_sagemaker_training_backend.py
```diff
@@ -196,6 +197,7 @@ def test_execute(settings):
                     "PYTHONUNBUFFERED": "1",
                     "no_proxy": "amazonaws.com",
                     "GRAND_CHALLENGE_COMPONENT_WRITABLE_DIRECTORIES": "/opt/ml/output/data:/opt/ml/model:/opt/ml/checkpoints:/tmp",
+                    "GRAND_CHALLENGE_COMPONENT_POST_CLEAN_DIRECTORIES": "/opt/ml/output/data:/opt/ml/model",
```
jmsmkn, line 200: Note to self - before deploying we will need to upgrade all of the active images to use the latest version of SageMaker Shim.

app/grandchallenge/evaluation/forms.py
```diff
@@ -342,19 +349,24 @@ def __init__(self, *args, user, phase: Phase, **kwargs):  # noqa: C901
...
+                has_active_image=True
+            ).order_by("title")
             if self._phase.parent:
                 qs = qs.filter(
-                    submission__phase=self._phase.parent,
-                    submission__evaluation__status=Evaluation.SUCCESS,
-                    job__status=Job.SUCCESS,
+                    algorithm_container_images__submission__phase=self._phase.parent,
+                    algorithm_container_images__submission__evaluation__status=Evaluation.SUCCESS,
+                    algorithm_container_images__job__status=Job.SUCCESS,
```
jmsmkn, line 359: For the submission to the child phases do we also take into account the algorithm image AND model to the parent phase?

app/grandchallenge/algorithms/tasks.py
```diff
@@ -478,3 +484,80 @@ def set_credits_per_job():
             algorithm.credits_per_job = default_credits_per_job
 
         algorithm.save(update_fields=("credits_per_job",))
+
+
+@shared_task(**settings.CELERY_TASK_DECORATOR_KWARGS["acks-late-micro-short"])
```
jmsmkn, line 489: More likely this needs to go on the large queue as I'm not confident it will always fit into 5 minutes.

app/grandchallenge/algorithms/tasks.py
```diff
@@ -478,3 +484,80 @@ def set_credits_per_job():
...
+        .exclude(pk=current_model.pk)
+        .exists()
+    ):
+        current_model.import_status = ImportStatusChoices.FAILED
+        current_model.status = (
+            "Algorithm model with this sha256 already exists."
+        )
+        current_model.save()
+
+        current_model.delete_model_file()
```
jmsmkn, line 532: Not to self - we will need to add delete permissions before deploying.

app/grandchallenge/algorithms/views.py
jmsmkn: Does the same need to be done for the job creation API view?

jmsmkn (commented):
Looks good!

app/grandchallenge/evaluation/forms.py
amickan: Ah good catch. No we don't yet. Just for my own understanding: we want to only allow algorithms that have a successful submission to the parent phase with the currently active image and model (if applicable). That's a more complex query than what we currently have.

jmsmkn: Indeed - that's correct, see roadmap 301 for the motivation.

app/grandchallenge/algorithms/views.py
amickan: Yes, good catch.

amickan: Actually, it turns out that on the job serializer we always create new civs, so this check will always pass. Hence, not needed. We might want to update the serializer to not always create new civs, but that's for another PR.

app/grandchallenge/evaluation/forms.py
amickan: This is one huge query now, but it's as efficient as it will get if we want to check all this. Something that I noticed just now when checking queries though is that we do two additional lookups per algorithm in order to display the active image pk and active model pk and their comments in the field label. I should annotate those on the queryset instead. I might not get to pushing those changes until Thursday, depends on Loïc

jmsmkn (approved):

app/grandchallenge/algorithms/views.py
jmsmkn: Could you create a separate issue for this?

amickan: Done!

amickan:
I will merge this PR so that I can continue working on the pitch, but I will refrain from deploying GC until the two admin issues have been completed.

PR #3361, opened 2024-05-31
Make archive items editable again

amickan:
Fixes bug introduced in https://github.com/comic/grand-challenge.org/pull/3345 which made archive items uneditable.


jmsmkn (approved):

chrisvanrun:
Thanks! I had neglected to consider its usage in templates at the time of removing it.

PR #3364, opened 2024-05-31
Rework Component Interface Value representations

chrisvanrun:
Currently, we have several slightly different not-so-nice, representations of component interface values (CIVs).

This PR contains a rework of these representations in the job detail pages for evaluation and algorithms. The ultimate goal is to extend its usage to the ArchiveItem detail page and ArchiveItem/DisplaySet list pages.

## Template tags
~Two template tags are introduced: `civ` and `civ_inline`. Each tag takes a component interface value as the only argument and renders an HTML representation of the CIV. The inline variant is meant for list-like representations where horizontal space is limited and relies on a modal—the regular one where there is plenty of horizontal space to render the CIV proper.~

A new template civ.html is introduced that accepts the arguments `civ` and `display_inline`. The `display_inline` is a switch that can be set to True and is meant for list-like representations where horizontal space is limited and relies on a modal—the regular one where there is plenty of horizontal space to render the CIV proper

## Vega Lazy loading
In addition, this PR adds lazy loading for vega charts which is very much needed if viewing a list of CIVSets that output charts.

## New Fixtures
CIVs are a crucial part of development now, and I think adding a bunch of test CIVs by hand is a lot of hassle. This, I think, plays a part in these representations flying under the radar. As such, this PR also introduces the `scripts/component_interface_value_fixtures.py`. This repeatable script creates an algorithm job that has an overabundance of different CIVs for quick development.

The screenshots are made based on this fixture.


chrisvanrun:
A possibly important change I forgot to mention is the following:

CIVs' rendering on a detail page was selective: some types would show, but some would not. With this PR that changes and they are all shown. For instance, output images will have a download button. This did mean that the BOOL/STRING/FLOAT/INT(?) types used to be gathered in a single table:

long_format_old_render

Whereas they are now shown as individual items:
Screenshot 2024-05-30 172449

IMHO this is fine as not many algorithms output multiple bool/strings.

app/grandchallenge/core/templatetags/json.py
amickan, line 1: Don't we have this tag already?

amickan (dismissed):
LGTM!

scripts/component_interface_value_fixtures.py
amickan, line 1: Maybe add this the makefile so that it's just a matter of calling `make component_interface_value_fixtures`

app/grandchallenge/core/templatetags/json.py
chrisvanrun: We did, but it was part of the evaluation_extra template_tag module; this was already used in several other non-evaluation locations. So I opted to pull it out of that module and create a dedicated json module.

amickan: But it's now in the codebase twice, isn't it? Or did I miss the deletion?

app/grandchallenge/charts/static/js/charts/render_charts_lazy.mjs
```diff
@@ -0,0 +1,44 @@
+// This module will lazily embed vega-lite charts
```
jmsmkn, line 1: Why add a new file for this? Why not use the same (existing) rendering method for everything?

app/grandchallenge/core/templates/grandchallenge/partials/component_interface_values/fallback.html
```diff
@@ -0,0 +1,3 @@
+<div class="badge badge-warning">
+    {{ civ.interface.title }}: cannot display
```
jmsmkn, line 2:
```suggestion
    {{ civ.interface.title }}: cannot be displayed
```

app/grandchallenge/core/templates/grandchallenge/partials/component_interface_values/inline/fallback.html
```diff
@@ -0,0 +1 @@
+{% include "../fallback.html" %}
```
jmsmkn, line 1: Please use absolute references to templates, not relative.

app/grandchallenge/core/templates/grandchallenge/partials/component_interface_values/inline/file.html
```diff
@@ -0,0 +1,8 @@
+{% extends "../file.html" %}
```
jmsmkn, line 1: Please use absolute references to templates, not relative.

jmsmkn: There are a lot of these, so won't repeat this.

app/tests/core_tests/test_templatetags.py
```diff
@@ -0,0 +1,58 @@
...
+        ),
+    ),
+)
+@pytest.mark.django_db
+def test_civ_render_tag(tag, component_interface_value):
+    if component_interface_value.interface.kind != InterfaceKindChoices.IMAGE:
+        component_interface_value.image = ImageFactory()
+        ImageFileFactory(image=component_interface_value.image)
+
+    assert tag(component_interface_value)
```
jmsmkn, line 58: This doesn't really test anything, only that the method runs, not the expected value, it could return junk.

app/grandchallenge/core/templatetags/civ.py
```diff
@@ -0,0 +1,101 @@
...
+            InterfaceKindChoices.THUMBNAIL_PNG,
+        ):
+            return "file_thumbnail.html"
+
+        return "file.html"
+
+    if component_interface_value.image:
+        return "image.html"
+
+    return "fallback.html"
```
jmsmkn, line 101: Using these three tags to render templates really breaks the discoverability of what is being used, still. This could all be done with a single template and switch statements to include partials. That way, the discoverability of templates is still integrated with IDEs, etc.

app/grandchallenge/core/templatetags/civ.py
```diff
@@ -0,0 +1,101 @@
...
+    return render_to_string(
+        template_name=str(CIV_PARTIALS / "inline" / template_name),
+        context={"civ": component_interface_value},
+    )
+
+
+def _get_civ_template(component_interface_value) -> str:
+    interface = component_interface_value.interface
+
+    if component_interface_value.value is not None:
```
jmsmkn, line 83: You need to inspect the super kind of the image, `None` is a valid value for some CIVs.

app/tests/core_tests/test_templatetags.py
chrisvanrun: It is a smoke test. What it tests is that it can run through the render. I think adding a whole 'snapshot' of the render is a bit too much but perhaps adding a snippet to check on is OK?

app/grandchallenge/core/templatetags/civ.py
chrisvanrun: Good point. I copied this from an earlier template switch. I'll clean it up.

app/tests/core_tests/test_templatetags.py
chrisvanrun: I've added a new test with the template format.

app/grandchallenge/core/templatetags/civ.py
chrisvanrun: I did test a `FLOAT` kind in both the `store_in_database=True` (i.e. `VALUE` super_kind) and the same `FLOAT` kind for `store_in_database=False` (i.e. `FILE` super_kind).

AFAIK these JSON types are the only ones that can switch between object_store and database_store.

I have only considered valid CIVs in the rendering but have a fallback if there is a civ with an unset `.image` or unset `.file`. To be clear, under normal circumstances, that should never occur.

app/grandchallenge/core/templatetags/json.py
chrisvanrun: Ah, I see. You are 100% correct! Good catch!

app/grandchallenge/components/templatetags/civ.py
```diff
@@ -0,0 +1,59 @@
...
+    return component_interface_value.interface.kind in [
+        InterfaceKindChoices.BOOL,
+        InterfaceKindChoices.FLOAT,
+        InterfaceKindChoices.INTEGER,
+        InterfaceKindChoices.STRING,
+    ]
+
+
+@register.filter
+def is_thumbnail_kind(component_interface_value):
```
jmsmkn, line 55: The kind definitions [are done on the class](https://github.com/comic/grand-challenge.org/blob/main/app/grandchallenge/components/models.py#L748), could these be moved there to prevent any split logic?

app/tests/components_tests/test_partial_templates.py
```diff
@@ -0,0 +1,121 @@
...
+            "thumbnail.jpg",
+            ContentFile(b"<bh:ff><bh:d8><bh:ff><bh:e0><bh:00><bh:10>JFIF"),
+        )
+
+    if ci.kind == InterfaceKindChoices.IMAGE:
+        civ.image = ImageFactory()
+        ImageFileFactory(image=civ.image)
+
+    # Actually create the CIV
+    if ci.kind != InterfaceKindChoices.SEGMENTATION:  # Intentionally broken
```
jmsmkn, line 116: Just for my information, why?

app/grandchallenge/components/templates/components/partials/civ_inline.html
```diff
@@ -0,0 +1,30 @@
...
+        {% endif %}
+    {% elif interface.super_kind == interface.SuperKind.IMAGE %}
+        {% if not civ.image %}
+            {% include "components/partials/civ_inline/fallback.html" %}
+        {% else %}
+            {% include "components/partials/civ_inline/image.html" %}
+        {% endif %}
+    {% endif %}
+
+{% endwith %}
```
jmsmkn, line 30: This seems to be duplicate logic, or there are some subtle differences that are hard to follow. Rather than having two files and a `{% block badge_content %}`, maybe it would be better to have one file with an if statement and to pass the variable `display_civ_inline=True/False`?

app/grandchallenge/components/templates/components/partials/civ_inline/file_thumbnail.html
```diff
@@ -0,0 +1,5 @@
+{% extends "components/partials/civ_inline/modal_base.html" %}
+
+{% block content %}
+    {% include "components/partials/civ//file_thumbnail.html" %}
```
jmsmkn, line 4:
```suggestion
    {% include "components/partials/civ/file_thumbnail.html" %}
```

app/grandchallenge/components/templates/components/partials/civ_inline.html
jmsmkn: Sorry, the PR overall looks good, just getting a bit lost as everything is spread out over multiple interlinked files.

app/grandchallenge/components/templatetags/civ.py
chrisvanrun: I intentionally left them here since this is display/template logic which does not always align with model logic.

app/tests/components_tests/test_partial_templates.py
chrisvanrun: To test the fallback which represents a graceful fail route. This stems from that we create the CIV generally by first binding the interface and later bind the image/value/file. I am not 100% convinced we do or if it's even possible to do these atomic.

Perhaps 'incomplete' is a better term.

app/grandchallenge/components/templates/components/partials/civ_inline.html
chrisvanrun: > Sorry, the PR overall looks good

Thanks ;-).

Now that I look at it with a bit more scrutiny I think I can push the small difference into the actual template.

I could have a single file with a variable, but then we'd lose the requested IDE compact since the base path would be dynamic of the templates. Furthermore, does that not swim into the 'do one thing well' philosophy?

jmsmkn: That's my point really, I don't think we need two templates for each interface type in different paths - but one template for each interface type with an if statement in there which would clearly show the differences between the two implementations.

app/grandchallenge/components/templatetags/civ.py
jmsmkn: I still think that it's better to put it on the class, no matter where it's used.

app/grandchallenge/components/templates/components/partials/civ_inline.html
chrisvanrun: Ah, I see your point! Yes. That would put the two different renditions of a CIV nicely together.

jmsmkn (approved):
Looks good!

PR #3367, opened 2024-06-05
Add bounding box configuration for point annotations

pkcakeout:
Pitch https://github.com/DIAGNijmegen/rse-roadmap/issues/319

Open question: Does this need tests. It feels kind of silly to me to add tests for this, but maybe?


app/grandchallenge/workstation_configs/models.py
```diff
@@ -416,6 +416,14 @@ class ImageInterpolationType(models.TextChoices):
         "outputs or the first overlay segment when viewing a reader study",
     )
 
+    point_bounding_box_size_mm = models.FloatField(
```
jmsmkn, line 419: Are you sure you want a float field? They can be weird, maybe a Decimal Field would suit better. See `default_zoom_scale` for an example.

You probably also want to set at least a minimum value validator, maybe a maximum too.

The value is also missing from the serializer.

I would at least add the value to a serializer test (they will be an existing one).

jmsmkn (approved):

app/grandchallenge/workstation_configs/models.py
jmsmkn: Yes exactly that.

jmsmkn (changes requested):
Actually the tests need fixing. As `blank=False,` a value for this field needs to be explicitly sent in the data field for the two broken tests.

pkcakeout:
Thanks James!

I was on it already, but did not know why or how to get it working! I was seeing this error in the logs:

```
postgres-1                  | 2024-06-05 12:55:04.636 UTC [2547] ERROR:  new row for relation "workstation_configs_windowpreset" violates check constraint "workstation_configs_windowpreset_either_fixed_or_percentile"
postgres-1                  | 2024-06-05 12:55:04.636 UTC [2547] DETAIL:  Failing row contains (64, foo, null, foo, 1, null, null, null).
postgres-1                  | 2024-06-05 12:55:04.636 UTC [2547] STATEMENT:  INSERT INTO "workstation_configs_windowpreset" ("title", "description", "slug", "width", "center", "lower_percentile", "upper_percentile") VALUES ('foo', NULL, 'foo', 1, NULL, NULL, NULL) RETURNING "workstation_configs_windowpreset"."id"
```

but maybe this was a red herring?

Will try your suggestion!

jmsmkn:
Adding `"point_bounding_box_size_mm": "0",` to the posted data in the two tests or setting `blank=True,` (and ensuring you handle `None`s alongside `0`) would solve it.

jmsmkn (approved):

PR #3369, opened 2024-06-06
Enable algorithm model version control

amickan:
This adds update and activate views for algorithm models. 
It also adds the algorithm models to the challenge costs. 

Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/297


app/grandchallenge/core/storage.py
```diff
@@ -165,7 +165,9 @@ def copy_s3_object(
     target_key = to_field.storage.get_available_name(
         name=target_key, max_length=to_field.field.max_length
     )
-    extra_args = {"ContentType": mimetype, "ChecksumAlgorithm": "SHA256"}
+    extra_args = {
+        "ContentType": mimetype,
```
jmsmkn, line 169: Why was the checksum removed? We need a test for this as we know it is functionality from S3 we rely on but cannot test with minio. We should make sure this argument is sent. See the use of the boto3 stubber tests for the SageMaker API interactions in the components module.

amickan: By accident! I had removed it for some testing and forgot to put it back. I will add a test.

jmsmkn: No problem! Thats what I thought.

jmsmkn (approved):
LGTM!

PR #3370, opened 2024-06-06
Add test for model sha256 being set

jmsmkn:


amickan (approved):
Thanks for figuring this out!

PR #3378, opened 2024-06-13
Archive Item and related Jobs pages

chrisvanrun:
Part of the pitch:
- https://github.com/DIAGNijmegen/rse-roadmap/issues/324

This PR uses the CIV representations introduced in an earlier PR to show details of Archive Items and Display Sets.

It also contains a page that shows all Jobs that have the same inputs as the Archive Item has as content/values.

Draft: I still need to fix an algorithm list test and provide some up-to-date screenshots. Ran out of time today but any initial feedback is welcome.

## Updated Job List

result_list_new

## Updated Archive Item list

archive_item_list

## New Archive Item Detail list
archive_item

## New Archive Item Job list
Shows the same content as the Job list, but with an ID and optional title displayed on top.

archive_item_related_jobs


chrisvanrun:
The exact query is still something I am not 100% sure about.

app/grandchallenge/archives/models.py
```diff
@@ -359,6 +359,13 @@ def delete_url(self):
             kwargs={"slug": self.base_object.slug, "pk": self.pk},
         )
 
+    @property
+    def detail_url(self):
```
amickan, line 363: This should be called `get_absolute_url`, see https://docs.djangoproject.com/en/5.0/ref/models/instances/#get-absolute-url

app/grandchallenge/reader_studies/models.py
```diff
@@ -898,6 +898,13 @@ def update_url(self):
             kwargs={"slug": self.base_object.slug, "pk": self.pk},
         )
 
+    @property
+    def detail_url(self):
```
amickan, line 902: This should also be `get_absolute_url`.

app/grandchallenge/archives/urls.py
```diff
@@ -59,6 +61,16 @@
         name="permission-request-update",
     ),
     path("<slug>/items/", ArchiveItemsList.as_view(), name="items-list"),
+    path(
+        "<slug>/items/<uuid:pk>/",
+        ArchiveItemDetailView.as_view(),
+        name="item-detail",
+    ),
+    path(
+        "<slug>/items/<uuid:pk>/jobs",
```
amickan, line 70:
```suggestion
        "<slug>/items/<uuid:pk>/jobs/",
```

app/grandchallenge/reader_studies/urls.py
```diff
@@ -93,6 +94,11 @@
         DisplaySetCreate.as_view(),
         name="display-set-create",
     ),
+    path(
+        "<slug>/display-sets/<uuid:pk>",
```
amickan, line 98:
```suggestion
        "<slug>/display-sets/<uuid:pk>/",
```

app/grandchallenge/algorithms/views.py
```diff
@@ -640,6 +639,22 @@ class JobsList(PaginatedTableListView):
...
+        Column(title="Created", sort_field="created"),
+        Column(title="Creator", sort_field="creator__username"),
+        Column(title="Status", sort_field="status"),
+        Column(title="Visibility", sort_field="public"),
+        Column(
+            title="Comment",
+            sort_field="comment",
+            optional_condition=lambda obj: bool(obj.comment),
+        ),
+        Column(title="Results", sort_field="outputs"),
```
amickan, line 654: Have you tried this? I don't think ordering on a many to many field like this works.

app/grandchallenge/archives/views.py
```diff
@@ -558,6 +564,115 @@ def get_serializer_class(self):
             return ArchiveItemSerializer
 
 
+class ArchiveItemJobListView(
+    LoginRequiredMixin,
+    ObjectPermissionRequiredMixin,
+    JobsList,
+):
+    model = Job
```
amickan, line 572: This can go, right? It's already set on `JobsList`

app/grandchallenge/archives/views.py
```diff
@@ -558,6 +564,115 @@ def get_serializer_class(self):
             return ArchiveItemSerializer
 
 
+class ArchiveItemJobListView(
+    LoginRequiredMixin,
+    ObjectPermissionRequiredMixin,
+    JobsList,
```
amickan, line 570: What are you inheriting this for? The only thing that is common between the JobList and this archive item job list is the column definition, right? Everything else is overwritten. I think it would be better to inherit from `PaginatedTableListView` directly.

app/grandchallenge/archives/views.py
```diff
@@ -558,6 +564,115 @@ def get_serializer_class(self):
...
+    login_url = reverse_lazy("account_login")
+
+    @cached_property
+    def archive_item(self):
+        return get_object_or_404(ArchiveItem, pk=self.kwargs["pk"])
+
+    def get_permission_object(self):
+        return self.archive_item
+
+    def get_queryset(self):
```
amickan, line 601: This should include a call to `super()`. Currently that will fail because you inherited from `JobList` which expects an algorithm slug, not an archive slug, but as I argue above, I would inherit directly from `PaginatedListVIew` instead.

app/grandchallenge/archives/views.py
```diff
@@ -558,6 +564,115 @@ def get_serializer_class(self):
...
+            {
+                "algorithm": object_.algorithm_image.algorithm,
+            }
+        )
+        return context
+
+    def get_context_data(self, *args, **kwargs):
+        context = PaginatedTableListView.get_context_data(
+            self, *args, **kwargs
+        )
```
amickan, line 653: This should be `super().get_context_data(*args, **kwargs)`

app/grandchallenge/components/views.py
```diff
@@ -270,6 +276,29 @@ def get_context_data(self, *args, **kwargs):
...
+):
+    model = None
+    permission_required = None
+    raise_exception = True
+    login_url = reverse_lazy("account_login")
+    template_name = "components/civ_set_detail.html"
+
+    @cached_property
+    def base_object(self):
+        return NotImplementedError
```
amickan, line 292: Since this is the same for both views, you could save yourself overriding this property on the subclassed views. What do you think? 

```suggestion
        return self.object.base_object

```

app/grandchallenge/components/views.py
```diff
@@ -270,6 +276,29 @@ def get_context_data(self, *args, **kwargs):
...
+
+    def get_context_data(self, *args, **kwargs):
+        context = super().get_context_data(*args, **kwargs)
+
+        context.update(dict(base_object=self.base_object))
```
amickan, line 297: Using a literal here is more in line with what we do in the rest of the codebase. Did you have a specific reason for using the `dict()` constructor instead? 
```suggestion
        context.update({"base_object": self.base_object})

```

app/grandchallenge/archives/views.py
amickan: You can disregard this comment given the other comments about inheriting directly from `PaginatedListView`

app/grandchallenge/archives/views.py
```diff
@@ -558,6 +564,115 @@ def get_serializer_class(self):
...
+        return self.archive_item
+
+    def get_queryset(self):
+        input_civs = self.archive_item.values.all()
+        num_civs = len(input_civs)
+
+        jobs_subquery = Job.objects.filter(inputs__in=input_civs).distinct()
+
+        query_set = (
+            Job.objects.filter(pk__in=jobs_subquery)
```
amickan, line 608: Why do the jobs_subquery separately? Isn't it more efficient (or at least not less efficient) to integrate this into the main query below?

app/grandchallenge/reader_studies/views.py
```diff
@@ -395,7 +397,7 @@ class ReaderStudyDisplaySetList(CivSetListView):
         *CivSetListView.columns,
     ]
 
-    default_sort_column = 3
+    default_sort_column = 2
```
amickan, line 400: Shouldn't this be 4 now?

app/grandchallenge/components/views.py
amickan, line 297: You don't need to add this to the context, do you? Since this is a detail view, there will always be an object for which you can access the base_object in the template by simply doing `object.base_object`.

app/tests/archives_tests/test_views.py
```diff
@@ -999,3 +1001,177 @@ def test_archive_item_bulk_delete_permissions(client):
...
+def test_archive_item_detail_permissions(client):
+    _test_archive_item_permissions(
+        viewname="archives:item-detail", client=client
+    )
+
+
+def _test_archive_item_permissions(viewname, client):
+    archive = ArchiveFactory()
+    archive.add_editor(editor := UserFactory())
+    archive.add_uploader(uploader := UserFactory())
```
amickan, line 1146: Don't you also need an archive user instance?

app/tests/archives_tests/test_views.py
```diff
@@ -999,3 +1001,177 @@ def test_archive_item_bulk_delete_permissions(client):
...
+
+@pytest.mark.parametrize(
+    "job_creator,expected_num_matches",
+    (
+        (lambda: AlgorithmJobFactory.create_batch(3, public=True), 3),
+        (lambda: AlgorithmJobFactory.create_batch(3, public=False), 0),
+    ),
+)
+@pytest.mark.django_db
+def test_archive_item_job_list_permissions_filter(
```
amickan, line 1087: The permissions depend on the `view_job` permission, right? Testing private vs public jobs is then not the most meaningful or direct test, right? For editors it shouldn't matter whether the jobs are private or public as far as I know.

app/tests/archives_tests/test_views.py
```diff
@@ -999,3 +1001,177 @@ def test_archive_item_bulk_delete_permissions(client):
...
+    )
+
+
+@pytest.mark.django_db
+def test_archive_item_detail_permissions(client):
+    _test_archive_item_permissions(
+        viewname="archives:item-detail", client=client
+    )
+
+
```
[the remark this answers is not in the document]
chrisvanrun: Would it? I generally take the parameterize as _'Test this in different ways',_ whereas this is an _'In this way, test these different things'_.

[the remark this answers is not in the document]
amickan: I definitely also use it as  _'In this way, test these different things'._ I think we have quite a few examples of passing different viewnames to a test via parametrize.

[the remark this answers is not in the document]
amickan: But I don't have a very strong opinion on this, it just struck me as unusual in our codebase.

app/grandchallenge/components/views.py
chrisvanrun: This is in-line with the other CIV template styles.

app/grandchallenge/reader_studies/models.py
chrisvanrun: Thanks!

app/grandchallenge/components/views.py
amickan: In those other views, it is needed though, because you don't have a single object to get the base_object from (yet) (i.e. on the listview and the formview). The delete view also does this unnecessarily, I just saw, that could be refactored (we also have a duplicate template there for some reason, but that can be fixed in a seperate PR).

app/grandchallenge/reader_studies/views.py
chrisvanrun: Oh yeah, order makes a lot more sense to keep instead of ID.

app/grandchallenge/archives/views.py
amickan: Somehow, whenever I select a row to comment on, it also selects a few rows before, making the comment less clear. Sorry about that! This refers to the `get_queryset` method, not `get_permission_object`

app/grandchallenge/algorithms/views.py
chrisvanrun: Good point. Not much use having this here.

chrisvanrun:
Added a link to the algorithm:
image

app/tests/archives_tests/test_views.py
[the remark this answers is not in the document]
chrisvanrun: Go with the flow I guess! Updated.

app/grandchallenge/components/views.py
chrisvanrun: Updated!

app/grandchallenge/archives/views.py
chrisvanrun: I've incorporated it with a subquery now. Which passes the tests, but I am unsure about performance improvements.

app/grandchallenge/components/views.py
chrisvanrun: I like the way the IDE's style the `dict()` constructor more than having `"argument"` everywhere. No longer applicable though, since this was refactored away.

app/grandchallenge/archives/views.py
chrisvanrun: Done!

app/grandchallenge/archives/views.py
chrisvanrun: Opted to directly inherit PaginatedListView

app/grandchallenge/algorithms/views.py
chrisvanrun: It doesn't! A such I have removed it!

app/grandchallenge/archives/views.py
```diff
@@ -558,6 +564,128 @@ def get_serializer_class(self):
...
+        num_civs = len(input_civs)
+
+        queryset = (
+            queryset.filter(
+                pk__in=Subquery(
+                    Job.objects.filter(
+                        inputs__archive_items=self.archive_item
+                    ).values("pk"),
+                )
+            )
```
amickan, line 628: For my understanding: why do you need a subquery here? Can't you just do something like `queryset.filter(inputs__archive_items=self.archive_item)`? And why is it faster than doing `inputs__in=[input_civs]`?

app/grandchallenge/archives/models.py
```diff
@@ -359,6 +359,13 @@ def delete_url(self):
             kwargs={"slug": self.base_object.slug, "pk": self.pk},
         )
 
+    @property
```
jmsmkn, line 362: `get_absolute_url` should not be a property. https://docs.djangoproject.com/en/5.0/ref/models/instances/#get-absolute-url

app/tests/archives_tests/test_views.py
```diff
@@ -999,3 +1001,183 @@ def test_archive_item_bulk_delete_permissions(client):
         list(response.context["form"].fields["civ_sets_to_delete"].queryset)
         == []
     )
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize(
+    "distractor_generator",
```
jmsmkn, line 1008: What is a distractor?

app/tests/archives_tests/test_views.py
```diff
@@ -999,3 +1001,183 @@ def test_archive_item_bulk_delete_permissions(client):
...
+    assert ArchiveItemJobListView.columns[detail_column].title == "Detail"
+
+    assert str(aj1.pk) in resp["data"][0][detail_column]
+    assert str(aj2.pk) in resp["data"][1][detail_column]
+
+
+@pytest.mark.django_db
+def test_archive_item_job_list_permissions_filter(client):
+    archive = ArchiveFactory()
+    archive.add_editor(editor := UserFactory())
```
jmsmkn, line 1082: That was incredibly confusing. There is no point at all to using a walrus argument here.

```suggestion
    editor = UserFactory()
    archive.add_editor(editor)
```

app/tests/reader_studies_tests/test_views.py
```diff
@@ -384,6 +384,39 @@ def test_display_set_update_permissions(client):
     assert response.status_code == 200
 
 
+@pytest.mark.django_db
+def test_display_set_detail_permissions(client):
+    rs = ReaderStudyFactory()
+    rs.add_editor(editor := UserFactory())
+    rs.add_reader(reader := UserFactory())
```
jmsmkn, line 391: A line costs nothing, and it is incredibly weird to have what looks like a function call have side effects in the parent frame. This is also subtly not using kwargs. Please read https://peps.python.org/pep-0572/#the-importance-of-real-code, and especially "Exceptional cases", for appropriate uses of this operator.

app/tests/archives_tests/test_views.py
```diff
@@ -999,3 +1001,183 @@ def test_archive_item_bulk_delete_permissions(client):
...
+        lambda civs: [ComponentInterfaceValueFactory()],
+        # Partial CIVs in distractor
+        lambda civs: [civs[0]],
+        # Extra CIVs in distractor
+        lambda civs: [*civs, ComponentInterfaceValueFactory()],
+    ),
+)
+def test_archive_item_job_list_filter(distractor_generator, client):
+    archive = ArchiveFactory()
+    archive.add_editor(editor := UserFactory())
```
jmsmkn, line 1020: And here.

app/tests/archives_tests/test_views.py
```diff
@@ -999,3 +1001,183 @@ def test_archive_item_bulk_delete_permissions(client):
...
+
+@pytest.mark.django_db
+@pytest.mark.parametrize(
+    "viewname", ("archives:item-job-list", "archives:item-detail")
+)
+def test_archive_item_permissions_detail_and_list(viewname, client):
+    archive = ArchiveFactory()
+    archive.add_editor(editor := UserFactory())
+    archive.add_uploader(uploader := UserFactory())
+    archive.add_user(archive_user := UserFactory())
```
jmsmkn, line 1150: And here.

app/tests/algorithms_tests/test_views.py
```diff
@@ -312,7 +313,7 @@ def test_algorithm_jobs_list_view(client):
             "length": 50,
             "draw": 1,
             "search[value]": job.creator.username,
-            "order[0][column]": 0,
+            "order[0][column]": JobsList.default_sort_column,
```
jmsmkn, line 316: You've gone from an explicit test, to one depending on your code. What if the content of `JobsList.default_sort_column` does not match the API of what jquery tables expects? That is what this test was testing.

app/grandchallenge/reader_studies/models.py
```diff
@@ -898,6 +898,13 @@ def update_url(self):
             kwargs={"slug": self.base_object.slug, "pk": self.pk},
         )
 
+    @property
```
jmsmkn, line 901: `get_absolute_url` should not be a property.

app/grandchallenge/archives/views.py
```diff
@@ -558,6 +564,128 @@ def get_serializer_class(self):
...
+        "inputs__image__name",
+        "inputs__image__files__file",
+        # Outputs
+        "outputs__interface__title",
+        "outputs__interface__description",
+        # Algorithm
+        "algorithm_image__algorithm__title",
+        "algorithm_image__algorithm__description",
+        "algorithm_image__comment",
+    ]
```
jmsmkn, line 590: Having so many search fields with so many joins will make the search impossible to use.

app/grandchallenge/datatables/views.py
```diff
@@ -28,17 +28,27 @@ def get_context_data(self, *, object_list=None, **kwargs):
...
+        page_context = self.get_context_data()
+        rows = []
+        for o in object_list:
+            row_context = self.get_row_context(
+                object_=o, page_context=page_context
+            )
+            rows.append(
+                self.render_row(row_context=row_context),
+            )
+        return rows
```
jmsmkn, line 51: Why was this file changed? The API now does not match its superclass.

app/grandchallenge/components/templates/components/civ_set_detail.html
```diff
@@ -0,0 +1,91 @@
...
+            {% elif update_perm not in object_perms %}
+                title="Not allowed to edit {{ object|verbose_name }}"
+            {% endif %}
+        >
+            <a class="btn btn-primary {% if not object.is_editable or update_perm not in object_perms %}disabled{% endif %}"
+               href="{{ object.update_url }}"
+            >
+                <i class="fas fa-edit"></i> Edit
+            </a>
+        </span>
```
jmsmkn, line 50: Why show a button to users who cannot edit?

app/grandchallenge/components/templates/components/civ_set_detail.html
```diff
@@ -0,0 +1,91 @@
...
+        <h4>{{ object.title }}</h4>
+    {% endif %}
+    <dl>
+        {% for civ in object.values.all|sort_civs %}
+            <dt class="mt-3">{{ civ.interface.title }}</dt>
+            <dd>
+                {% include 'components/partials/civ.html' with object=civ only %}
+            </dd>
+        {% empty %}
+            <span class="text-muted">Empty or processing</span>
```
jmsmkn, line 82: "Empty or processing" is not super informative. Processing what? Are you sure?

app/grandchallenge/archives/views.py
chrisvanrun: I do get your confusion! The need for this rolled out of the query and filter tests I wrote that this is needed.  I think the following is happening:

`queryset.filter(inputs__archive_items=self.archive_item)` would result in a join that contains rows of Jobs and CIVs. Each Job is represented as many times as it has CIVs **that match the archvie_item**. However, it would not have a representation of a CIV that IS an input but not part of the archive item. Which would then result in incorrect detection by the `WHERE`/`annotate() + filter()` and might include subsets.

Rather what we need is all the jobs that _contain_  an output that is also in the archive item. Hence the subquery and selection on `pk`. 

There might be a different approach, but this passed the tests and seemed performant via an Index Scan in the query planner + practical local tests.

app/grandchallenge/archives/views.py
chrisvanrun: Good to know. I wasn't too sure about the practical effects. I do think we should:
- to open up the search for diversity in outputs 
- make it easy to find specific algorithms

How about:
```Python
search_fields = [
        "pk",
        # Outputs
        "outputs__interface__title",
        # Algorithm
        "algorithm_image__algorithm__title",
        "algorithm_image__algorithm__description",
        "algorithm_image__comment",
    ]
```

The `algorithm_image__comment` one would allow to search for a specific version of an algorithm.

app/tests/archives_tests/test_views.py
chrisvanrun: The general setup of the test is to ensure that a similar-looking set of inputs to a Job are NOT searched upon. Hence I minted them 'Distractors'. I could abstract the creation of the non-distractors into the params definition, but that seemed to me to be overcomplicating things.

app/tests/reader_studies_tests/test_views.py
chrisvanrun: Noted. Looking back at this introduction I think I know why I found the single line attractive, despite now agreeing it is too confusing to read:

```Python
editor = UserFactory()
rs.add_editor(editor)
```

On the first line, it isn't an editor yet...

app/grandchallenge/archives/views.py
jmsmkn: Joins and free text are expensive to search, I would get rid of both the description and the comment.

app/grandchallenge/components/templates/components/civ_set_detail.html
```diff
@@ -0,0 +1,91 @@
...
+                title="Not allowed to delete {{ object|verbose_name }}"
+            {% endif %}
+            class="float-right"
+            >
+            <a class="btn btn-danger {% if not object.is_editable or delete_perm not in object_perms %}disabled{% endif %}"
+            href="{{ object.delete_url }}"
+            >
+                <i class="fas fa-trash"></i> Delete
+            </a>
+        </span>
```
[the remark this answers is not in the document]
chrisvanrun: Good find! for the civ_row I am leaving the switches in as it makes sense to have them there as disabled buttons, even if you don't have the permissions.

app/tests/algorithms_tests/test_views.py
chrisvanrun: I think I see what you mean with the `"desc'` vs `"asc"`, those should likely be explicit.

The other ones: I don't see why those can't directly pull from the code. Since these are meant to mimic the initial or subsequent queries from the page once loaded.

Currently, if the `JobList.default_sort_column` is set to a non-valid value, the tests will fail whereas previously you'd only know if one would load the page.

So unless the test explicitly tests the ordering I'd argue it is fine to short-circuit the definitions to the mock request.

app/grandchallenge/datatables/views.py
chrisvanrun: I originally saw the need to add a row context and manipulate that in subclasses.

However, with 'voortschrijdend inzicht' I see that I can work around that problem by referencing the algorithm via the job object itself.

app/grandchallenge/components/templates/components/civ_set_detail.html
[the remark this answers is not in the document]
jmsmkn: I disagree, no point showing anything for something that will never enable. That is how we do it on the rest of the platform, please follow existing implementations.

app/tests/archives_tests/test_views.py
jmsmkn: Maybe name it differently? That term is non-standard and non-obvious. Better to choose something that tells the future reader exactly what is going on.

chrisvanrun: Ok, I've updated it to make more sense: I hope.

chrisvanrun:
While double-checking somethings I found:
- https://github.com/comic/grand-challenge.org/issues/3388

Going to fix that via a follow PR.

jmsmkn (approved):

PR #3380, opened 2024-06-14
Check for active method on submission form

### LINKED ISSUE #3374: Submission unfindable
> I accidentally submitted that algorithm before I uploaded the evaluation container, and now it’s stuck as a “ghost submission”. It doesn’t show up as a submission anywhere, but I also cannot submit it again.

The submission is visible on `/evaluation/submissions/`, but the admin does not have a way of getting to the submission detail page. The title "Submissions and Evaluations" of `/evaluation/<slug>/admin/` is misleading, as it's only the evaluations. Why is admin also in the url?

amickan:
Closes #3374

### DIFF
--- a/app/tests/evaluation_tests/test_forms.py
+++ b/app/tests/evaluation_tests/test_forms.py
@@ -247,31 +247,23 @@ class TestSubmissionForm:
 
         assert form.errors["algorithm"] == ["This field is required."]
 
-    def test_no_submission_possible_without_method(self):
-        csv_phase = PhaseFactory(submission_kind=SubmissionKindChoices.CSV)
-        alg_phase = PhaseFactory(
-            submission_kind=SubmissionKindChoices.ALGORITHM
-        )
-
+    @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


jmsmkn (dismissed):

app/tests/evaluation_tests/test_forms.py
```diff
@@ -247,6 +247,57 @@ def test_no_algorithm_selection(self):
...
+            is_desired_version=True,
+        )
+        del csv_phase.active_image
+
+        form = SubmissionForm(
+            user=UserFactory(),
+            phase=csv_phase,
+            data={"creator": UserFactory()},
+        )
+        form.is_valid()
```
jmsmkn, line 294: Missing assert? In general, this test probably could have been done with parametrize.

amickan: Yes, and it's not actually valid (missing the algorithm and a whole bunch of other things to be valid, but it's not needed for this test in my opinion, just want to show that this specific error disappears if you add a method).

Issue #3385, opened 2024-06-17
Unexpected hash in URL after navigation


chrisvanrun:
**Describe the bug**

After some navigation, a user ends up with the following URL: 'https://grand-challenge.org/archives/demo-challenge/#ttps://grand-challenge.org/archives/demo-challenge/items/'

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to an archive page
2. Click on 'Items'
3. Go back using the browser navigation

**Expected behaviour**

A regular, boring URL. However:

<img width="628" alt="image" src="https://github.com/comic/grand-challenge.org/assets/7871310/54f34b1b-04d0-4771-8291-adab5ce99723">

jmsmkn:
Probably caused by https://github.com/comic/grand-challenge.org/blob/b1343c0da46d7f3b610d065e82f3833dc73615d8/app/grandchallenge/core/static/js/refresh_sidebar.mjs#L25-L27. Have no idea how to get the back button working with the side panel otherwise.

PR #3386, opened 2024-06-17
Fix sometimes having an partial URL as hash

chrisvanrun:
Closes: #3385 

The `href` attribute was never parsed as a URL and incorrectly assumed to always be a hash-only string.


chrisvanrun:
Cleaner solution to have an allow list in the first place!

PR #3389, opened 2024-06-20
Make HX indicators robust against validation failures

chrisvanrun:
When adding a title to a display set / archive item that already exists, the validation error causes a partial update of the DOM, messing the the states of the 'Save' button.

This then shows the permanent state of 'Loading...':
image

The indicators handlers now:
- Listen  (also) to: [htmx:load](https://htmx.org/events/#htmx:load)
- Cleanup after themselves to make them robust for multiple subsequent calls

### DIFF
--- a/app/grandchallenge/components/static/components/js/htmx/display-me.mjs
+++ b/app/grandchallenge/components/static/components/js/htmx/display-me.mjs
@@ -7,5 +7,4 @@ function displayElements() {
 }
 
 document.addEventListener('htmx:load', displayElements);
-
-displayElements();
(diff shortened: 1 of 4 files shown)


miriam-groeneveld:
image

Isn't there a filter that will correct that? Or remove the "An" altogether?

chrisvanrun:
> image
> 
> Isn't there a filter that will correct that? Or remove the "An" altogether?

Fixed! An oversight since I was focussing on Archive Items at the time =P

jmsmkn (approved):

PR #3390, opened 2024-06-21
Add a Save and add another button to CIV forms

chrisvanrun:
Inspired by the admin interface, if a user needs to add a few display sets / archive items in a sequence this cuts the clicks by 50%.

image

Ofc, this isn't a problem if you add, say 5 ... but at 10/20 this is a nice addition while using the GC-api might be a step too far.

### DIFF
--- a/app/grandchallenge/archives/views.py
+++ b/app/grandchallenge/archives/views.py
@@ -598,9 +598,6 @@ class ArchiveItemCreateView(
             kwargs={"slug": self.base_object.slug},
         )
 
-    def get_success_url(self):
-        return self.return_url
-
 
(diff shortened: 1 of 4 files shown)


jmsmkn:
I find that cluster of buttons unintuitive. You have three buttons (Add value - does that save something? Save, and Save and add another) that seem to me to do something similar, and they have two different colours. Does back mean the changes will be discarded? Why do we need a back button? Why can't the user just use the back button in their browser? By "saving 50% of clicks" you mean 1 right? Given people seriously setting up a reader study or archive will be using the API I do not think this is worth it.

chrisvanrun:
If it is confusing, it is not worth adding!

PR #3395, opened 2024-06-24
Add external evaluators group to challenges

amickan:
First tiny PR to add the `external_evaluators_group` to the challenge model. This needs to happen in two steps: first, add the field as nullable and do a data migration to create this group for all existing challenges. That's what this PR does. In the next PR (together with all the other model changes for the pitch), we make the field non-nullable. 


jmsmkn:
When migrations are applied I think that the entire table is locked, so that should be a non-issue.

jmsmkn (approved):

PR #3399, opened 2024-06-27
Fix CIVSet POST process serializers not picking up uploads correctly

chrisvanrun:
It seems that ~both `"user_upload"` and~ `"upload_session"` were not picked up correctly for the API calls.

Furthermore, no tests existed for this use case.

Direct cause of: 
- https://github.com/DIAGNijmegen/rse-grand-challenge-forge/issues/44


app/grandchallenge/components/models.py
```diff
@@ -2080,15 +2080,23 @@ def create_civ_for_image(self, *, ci, current_civ, new_value, user):
             if created:
                 civ.full_clean()
             self.values.add(civ)
-        elif isinstance(new_value, QuerySet):
+        elif isinstance(
+            new_value, (QuerySet, UserUpload, RawImageUploadSession)
```
amickan, line 2084: For my understanding, when would we get a `user_upload` for image type interfaces?

app/tests/archives_tests/test_views.py
```diff
@@ -444,6 +445,105 @@ def test_api_archive_item_add_and_update_non_image_file(
...
+    # Add
+    assert item.values.count() == 0
+    set_civ_via_user_upload()
+    assert item.values.count() == 1
+    item.values.clear()
+
+    # Update
+    assert item.values.count() == 0
+    set_civ_via_user_upload()
+    assert item.values.count() == 1
```
amickan, line 544: What is the difference between add and update here? You're calling the exact same function and clear all values before, so same starting point and same api call, right?

app/grandchallenge/components/models.py
amickan: I thought you only get a `user_upload` for file type interfaces. If that is currently the case, I think we should not accomodate for a user_upload kwarg for image type interfaces and instead have the error raised (that you added, thanks!) so that we get notified of the change in the API.

app/grandchallenge/components/models.py
```diff
@@ -2099,6 +2107,8 @@ def create_civ_for_image(self, *, ci, current_civ, new_value, user):
                     immutable=True,
                 )
             )
+        else:
+            raise RuntimeError("Unexpected value presented")
```
amickan, line 2111: If new_value is an image and it it the same as the current value, we do nothing. That needs to be accounted for here, or else we'd get errors in those cases.

app/tests/archives_tests/test_views.py
chrisvanrun: Ah, ehm. Good point! It SHOULD update. I now see it does not because I am clearing things out. That should only happen at the start.

app/grandchallenge/components/models.py
chrisvanrun: Ok, I thought some flexibility might be nice but makes sense to keep things simple.

chrisvanrun:
I've updated the test and logic such that only Image / QuerySet / RawImageUploadSession is allowed.

amickan (approved):
LGTM! Thanks for fixing!

PR #3401, opened 2024-06-28
Enable image and model downloading

amickan:
This enables serving algorithm images and models to users with download permission. 

This also requires some minor changes to our terraform repo, a cloudfront distribution for the private S3 bucket and the following environment variables:
`PRIVATE_S3_STORAGE_CLOUDFRONT_DOMAIN `
`PRIVATE_S3_STORAGE_USE_CLOUDFRONT`


app/grandchallenge/algorithms/models.py
```diff
@@ -567,6 +570,13 @@ def get_absolute_url(self):
             kwargs={"slug": self.algorithm.slug, "pk": self.pk},
         )
 
+    @property
+    def download_url(self):
```
amickan, line 574: This link (and the corresponding one for the model below) are not used yet, but will be once the remaining external evaluation changes are implemented, hence I added them already.

app/config/settings.py
```diff
@@ -190,7 +190,13 @@
 PRIVATE_S3_STORAGE_KWARGS = {
     "bucket_name": os.environ.get(
         "PRIVATE_S3_STORAGE_BUCKET_NAME", "grand-challenge-private"
-    )
+    ),
+    # This is the domain where people will be able to go to download data
+    # from this bucket. Usually we would use reverse to find this out,
+    # but this needs to be defined before the database is populated
```
jmsmkn, line 196: Can't you use `reverse_lazy`?

app/config/settings.py
```diff
@@ -210,6 +216,12 @@
 PROTECTED_S3_STORAGE_CLOUDFRONT_DOMAIN = os.environ.get(
     "PROTECTED_S3_STORAGE_CLOUDFRONT_DOMAIN_NAME", ""
 )
+PRIVATE_S3_STORAGE_CLOUDFRONT_DOMAIN = os.environ.get(
+    "PRIVATE_S3_STORAGE_CLOUDFRONT_DOMAIN_NAME", ""
+)
+PRIVATE_S3_STORAGE_USE_CLOUDFRONT = strtobool(
+    os.environ.get("PRIVATE_S3_STORAGE_USE_CLOUDFRONT", "False")
+)
```
jmsmkn, line 224: We won't want to use CloudFront for the private storage.

jmsmkn:
I think the approach here isn't quite right. We will not want to add CloudFront to the private storage. All that needs to happen is that a signed S3 url is generated to download the object. I'll see if I can migrate the algorithm images to the protected bucket quickly.

amickan:
> I think the approach here isn't quite right. We will not want to add CloudFront to the private storage. All that needs to happen is that a signed S3 url is generated to download the object. I'll see if I can migrate the algorithm images to the protected bucket quickly so this might not be needed at all.

Alright. I was unsure about using CloudFront. Could you elaborate why we shouldn't use CloudFront for this use case?

amickan:
> Could you elaborate why we shouldn't use CloudFront for this use case?

Ah, I misunderstood your comment and though we didn't want to use CloudFront for algorithm image downloading, but you mean for the private bucket in general. That does make sense actually.

amickan:
> > Could you elaborate why we shouldn't use CloudFront for this use case?
> 
> Ah, I misunderstood your comment and thought you meant that we didn't want to use CloudFront for algorithm image downloading, but you mean for the private bucket in general. That does make sense actually.

amickan:
Closed in favour of https://github.com/comic/grand-challenge.org/pull/3409

Issue #3406, opened 2024-07-02
Evaluation methods and ground truths use the same icon


jmsmkn:
We should separate them a little. Maybe be consistent and use the docker logo for the containers as we do with the algorithm images.

PR #3408, opened 2024-07-03
Delete objects from s3 when object is deleted from DB

amickan:
Part of https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/268 
Closes https://github.com/comic/grand-challenge.org/issues/3406


app/grandchallenge/components/signals.py
```diff
@@ -0,0 +1,13 @@
...
+from grandchallenge.algorithms.models import AlgorithmImage, AlgorithmModel
+from grandchallenge.evaluation.models import EvaluationGroundTruth, Method
+
+
+@receiver(pre_delete, sender=AlgorithmImage)
+@receiver(pre_delete, sender=AlgorithmModel)
+@receiver(pre_delete, sender=EvaluationGroundTruth)
+@receiver(pre_delete, sender=Method)
+def delete_file_from_s3(instance, **_):
+    instance._client.delete_object(Bucket=instance.bucket, Key=instance.key)
```
jmsmkn, line 13: Why go through the client and not call `instance.model.delete()`?

amickan: Oh I didn't know that worked. That's much simpler.

app/grandchallenge/evaluation/models.py
```diff
@@ -1048,6 +1048,10 @@ def get_absolute_url(self):
     def get_peer_images(self):
         return Method.objects.filter(phase=self.phase)
 
+    @property
+    def linked_file(self):
+        raise self.image
```
jmsmkn, line 1053: This only needs to be set on the abstract `ComponentImage` as it is the same for `AlgorithmImage`s, `WorkstationImage`s and `Method`s, right?

Does the deletion work when using the bulk delete in the admin?

app/grandchallenge/components/signals.py
```diff
@@ -0,0 +1,13 @@
...
+from django.dispatch import receiver
+
+from grandchallenge.algorithms.models import AlgorithmImage, AlgorithmModel
+from grandchallenge.evaluation.models import EvaluationGroundTruth, Method
+
+
+@receiver(pre_delete, sender=AlgorithmImage)
+@receiver(pre_delete, sender=AlgorithmModel)
+@receiver(pre_delete, sender=EvaluationGroundTruth)
+@receiver(pre_delete, sender=Method)
```
jmsmkn, line 11: `WorkstationImage` is missing. Shouldn't the sender be the two abstract classes instead so that we do not have to keep this up to date?

amickan: The sender can't be an abstract model, as far as I know. Since abstract models do not correspond to actual database tables, they cannot directly be the sender for signals tied to database actions like pre_delete. I will add the WorkstationImage here though

app/grandchallenge/evaluation/models.py
amickan: Yes, this works with the bulk delete in the admin. And you're right, the property can be set on the abstract class

app/grandchallenge/components/signals.py
amickan: True, that works! Thanks. However, I realized that deleting the ComponentImage requires a change to the save method, where we currently check if the image has changed, and if so throw RunTimeError. I'm not sure if we want to change this because it would then open up the possibility that the image gets deleted when the instance does not. So maybe we should go through the client after all?

jmsmkn: What about setting `save=False` and/or using the `post_delete` signal? https://docs.djangoproject.com/en/5.0/ref/models/fields/#django.db.models.fields.files.FieldFile.delete

amickan: Ah nice, that does the trick

jmsmkn (approved):

PR #3420, opened 2024-07-04
External evaluation phases

amickan:
This implements all the necessary model changes and adds API views for listing claimable evaluations, claiming an evaluation and updating a claimed evaluation. 


app/grandchallenge/evaluation/models.py
```diff
@@ -1164,13 +1185,32 @@ def save(self, *args, **kwargs):
                     actor_only=False,
                     send_action=False,
                 )
-            e = create_evaluation.signature(
-                kwargs={"submission_pk": self.pk}, immutable=True
-            )
-            on_commit(e.apply_async)
+            if self.phase.external_evaluation:
+                Evaluation.objects.create(submission=self)
```
jmsmkn, line 1189: Even though it might not be necessary speed-wise it might be better to keep it all in the `create_evaluation` task so that it is logically handled in one place, that way future developers will only need to look in one place to find the business logic.

app/grandchallenge/evaluation/models.py
```diff
@@ -1315,6 +1358,14 @@ def assign_permissions(self):
         assign_perm("view_evaluation", admins_group, self)
         assign_perm("change_evaluation", admins_group, self)
 
+        if self.submission.phase.external_evaluation:
+            external_evaluators = (
+                self.submission.phase.challenge.external_evaluators_group
+            )
+            assign_perm("view_evaluation", external_evaluators, self)
+            assign_perm("change_evaluation", external_evaluators, self)
```
jmsmkn, line 1366: We probably should keep the "change_evaluation" permissions separate as this allows different existing functionality. We want to minimise the scope of what the external evaluator can do as these will be service accounts.

app/grandchallenge/evaluation/templates/evaluation/submission_form.html
```diff
@@ -131,7 +131,7 @@ <h3>Create a new submission</h3>
                 {% endif %}
                 to this phase.
 
-                {% if phase.submission_kind == phase.SubmissionKindChoices.ALGORITHM %}
+                {% if phase.submission_kind == phase.SubmissionKindChoices.ALGORITHM and not phase.external_evaluation %}
```
jmsmkn, line 134: The participants need to know that their algorithm and model will be shared with the Challenge admins. They need to know that their algorithms and models will leave Grand Challenge, and we will have no control or insight into how often those containers are used, who by, and with what data. It needs explicit opt-in on the submission form with a checkbox.

app/grandchallenge/evaluation/views/api.py
```diff
@@ -23,3 +45,88 @@ class EvaluationViewSet(ReadOnlyModelViewSet):
         *api_settings.DEFAULT_RENDERER_CLASSES,
         PaginatedCSVRenderer,
     )
+
+    @action(detail=True, permission_classes=[CanClaimEvaluation])
```
jmsmkn, line 49: What methods does this use? Maybe better to be explicit about it here. It should not be GET as GET requests should not alter the objects. It should be a POST or PATCH request.

app/grandchallenge/evaluation/views/api.py
```diff
@@ -23,3 +45,88 @@ class EvaluationViewSet(ReadOnlyModelViewSet):
...
+
+    @action(
+        detail=True,
+        methods=["PATCH"],
+        permission_classes=[CanClaimEvaluation],
+        serializer_class=ExternalEvaluationUpdateSerializer,
+    )
+    def update_external_evaluation(self, request, *args, **kwargs):
+        evaluation = self.get_object()
+        if evaluation.status != Evaluation.EXECUTING:
```
jmsmkn, line 86: 1. How do we differentiate between evaluations that GC is executing and evaluations that are being evaluated externally? This is kindof important for the CloudWatch metrics where we track how busy the system is, so that logic will need to change too.

2. Do you check to ensure that only the externally evaluated algorithms can be patched here? We need to ensure that things we are evaluating cannot be change, and that the entity who is evaluating the submission is the only one who can update.

3. Do we track who claimed the job? Maybe this is needed to ensure the correct entity is updating it again.

4. Is there a timeout on the claim? We should probably track this and mark it as failed if there do not return in time.

amickan: > 1. How do we differentiate between evaluations that GC is executing and evaluations that are being evaluated externally?

Ok, in that case, I suggest adding a status `CLAIMED` to the option of status choices and using that (set when evaluation is claimed).

amickan: > 2. Do you check to ensure that only the externally evaluated algorithms can be patched here?

Yes, the `CanClaimEvaluation` object permission checks that the requesting user has the `claim_evaluation` permission for the given evaluation. And that permission is granted when an evaluation is created for an external phase, never for internal evaluations.

amickan: > 3. Do we track who claimed the job? Maybe this is needed to ensure the correct entity is updating it again.

I need to add that indeed, or else we have no way of checking that the same entity is updating the evaluation.

amickan: > 4\. Is there a timeout on the claim?

Yes, that is still missing. I was thinking of adding a period task that marks evals as Cancelled if they have been in the claimed state for longer than X hours. We can start with a timeout of 1 day, and change that if necessary.

app/grandchallenge/evaluation/views/api.py
amickan: Ah good point, it should be patch then IMO.

app/grandchallenge/evaluation/views/api.py
jmsmkn: RE: 3 and 4

This is why I proposed using the `nonce` approach like we do with email validation. Sets a token that has a timeout and ensures that it is invalidated once it is used, and has a timeout. Double checks that the correct evaluation is being updated by the correct entity that requested it. If you issue an `nonce` token then you would not need to encode who claimed the evaluation, but can be sure that the evaluation that is being changed is the correct one.

amickan: I have added a nonce token that is created when an evaluation is claimed, stored on the evaluation instance and checked on update.

amickan: I have added a `claimed_by` field that is set when the evaluation is claimed and checked, together with the `started_at` time, on update. For now, I set the timeout to 1 day.

amickan: Wait, just realized that I forgot to check that the requesting user on update doesn't have any other claimed evals.

app/grandchallenge/evaluation/tasks.py
```diff
@@ -537,3 +548,21 @@ def assign_submission_permissions(*, phase_pk: uuid.UUID):
     )
     for sub in Submission.objects.filter(phase__id=phase_pk):
         sub.assign_permissions()
+
+
+@acks_late_micro_short_task
+def cancel_external_evaluations_past_timeout():
```
jmsmkn, line 554:
```suggestion
@transaction.atomic
def cancel_external_evaluations_past_timeout():
```

app/grandchallenge/evaluation/tasks.py
```diff
@@ -537,3 +548,21 @@ def assign_submission_permissions(*, phase_pk: uuid.UUID):
...
+    Evaluation = apps.get_model(  # noqa: N806
+        app_label="evaluation", model_name="Evaluation"
+    )
+    timeout_threshold = datetime.now() - timedelta(
```
jmsmkn, line 558: This should be `django.utils.timezone.now`. Likely we should add `datetime.now` to banned modules in a separate PR.

app/grandchallenge/evaluation/tasks.py
```diff
@@ -537,3 +548,21 @@ def assign_submission_permissions(*, phase_pk: uuid.UUID):
...
+        seconds=settings.EXTERNAL_EVALUATION_TIMEOUT_IN_SECONDS
+    )
+
+    Evaluation.objects.filter(
```
jmsmkn, line 562: De we need to notify the followers of the evaluation?

app/grandchallenge/evaluation/views/api.py
```diff
@@ -23,3 +46,118 @@ class EvaluationViewSet(ReadOnlyModelViewSet):
...
+                },
+                status=403,
+            )
+
+        if (
+            evaluation.started_at - now()
+        ).seconds > settings.EXTERNAL_EVALUATION_TIMEOUT_IN_SECONDS:
+            evaluation.status = Evaluation.CANCELLED
+            evaluation.error_message = "External evaluation timed out."
+            evaluation.save()
```
jmsmkn, line 117: We shouldn't modify the object and return a status of 400. Again we have the problem of the notifications. I think it's best to call the `cancel_external_evaluations_past_timeout` task here instead and return 400.

app/grandchallenge/evaluation/views/api.py
```diff
@@ -23,3 +46,118 @@ class EvaluationViewSet(ReadOnlyModelViewSet):
...
+
+            civ = ComponentInterfaceValue(interface=interface, value=metrics)
+            try:
+                civ.full_clean()
+                civ.save()
+                evaluation.outputs.add(civ)
+                evaluation.status = Evaluation.SUCCESS
+            except ValidationError as e:
+                evaluation.status = Evaluation.FAILURE
+                evaluation.error_message = str(e)
```
jmsmkn, line 149: Does it make sense to put `ValidationError`s on the evaluation? Those are for the submitter, whereas the `ValidationError`s are for the evaluator. I don't think it makes sense to save the object and mark as Failed here as the evaluator could potentially fix the issue, and again it would be a case of returning 4xx but changing the object.

app/config/settings.py
```diff
@@ -1220,6 +1224,10 @@ def sentry_before_send(event, hint):
         "task": "grandchallenge.emails.tasks.cleanup_sent_raw_emails",
         "schedule": crontab(hour=6, minute=0),
     },
+    "cancel_external_evaluations_past_timeout": {
+        "task": "grandchallenge.evaluation.tasks.cancel_external_evaluations_past_timeout",
+        "schedule": crontab(hour=6, minute=15),
```
jmsmkn, line 1229: Is once a day enough? This could potentially leave evaluations running for 47:59:59 hours.

app/grandchallenge/evaluation/views/api.py
```diff
@@ -23,3 +46,118 @@ class EvaluationViewSet(ReadOnlyModelViewSet):
...
+                {"status": "You can only claim one evaluation at a time."},
+                status=400,
+            )
+
+        evaluation.status = Evaluation.CLAIMED
+        evaluation.claimed_by = request.user
+        evaluation.started_at = now()
+        evaluation.save()
+
+        serializer = ExternalEvaluationSerializer(
```
jmsmkn, line 73: The serializer is not checked to see if it is valid, and is also not saved. I would expect to see something like:

```python
        object = self.get_object()

        ...

        serializer = self.get_serializer(
            instance=object, data=request.data, partial=True
        )

        if serializer.is_valid():
            serializer.save()
            return Response(data=serializer.data)
        else:
            return Response(
                data=serializer.errors, status=HTTP_400_BAD_REQUEST
            )
```

app/grandchallenge/evaluation/views/api.py
```diff
@@ -23,3 +46,118 @@ class EvaluationViewSet(ReadOnlyModelViewSet):
...
+                evaluation.outputs.add(civ)
+                evaluation.status = Evaluation.SUCCESS
+            except ValidationError as e:
+                evaluation.status = Evaluation.FAILURE
+                evaluation.error_message = str(e)
+                evaluation.save()
+                raise DRFValidationError(e)
+        else:
+            evaluation.status = Evaluation.FAILURE
+            evaluation.error_message = request.data.get("error_message", None)
```
jmsmkn, line 154: Same here.

app/grandchallenge/evaluation/views/api.py
```diff
@@ -23,3 +46,118 @@ class EvaluationViewSet(ReadOnlyModelViewSet):
...
+
+        evaluation.completed_at = now()
+        evaluation.save()
```
jmsmkn, line 157: Shouldn't the `serializer` be saved and not the object?

app/grandchallenge/evaluation/views/api.py
jmsmkn: Think this is all solved now! Looks good, just some minor stuff on approach.

jmsmkn:
Another thing - do the compute costs for the evaluation get set correctly, i.e. they should be zero.

app/grandchallenge/evaluation/views/api.py
```diff
@@ -23,3 +46,118 @@ class EvaluationViewSet(ReadOnlyModelViewSet):
...
+
+        serializer = ExternalEvaluationSerializer(
+            instance=evaluation, context={"request": request}
+        )
+        return Response(serializer.data)
+
+    @action(detail=False, methods=["GET"])
+    def claimable_evaluations(self, request, *args, **kwargs):
+        queryset = get_objects_for_user(
```
jmsmkn, line 80: You can probably reuse `permission_classes=[CanClaimEvaluation]`

app/grandchallenge/evaluation/views/api.py
```diff
@@ -23,3 +46,118 @@ class EvaluationViewSet(ReadOnlyModelViewSet):
...
+            evaluation.error_message = "External evaluation timed out."
+            evaluation.save()
+            return Response(
+                {
+                    "status": "You can only update an evaluation within 24 hours."
+                },
+                status=400,
+            )
+
+        serialized_evaluation = ExternalEvaluationUpdateSerializer(
```
jmsmkn, line 125: This should probably use `get_serializer`.

app/grandchallenge/evaluation/views/api.py
amickan: Don't we normally put `ValidationErrors` on the evaluation as well? I do agree though that it doesn't make sense to share evaluation failure with the submitter, but it's not clear to me what to do instead. Marking it as cancelled doesn't make sense either, does it? Also the evaluator could not fix it then since cancelled evaluations can't be re-claimed. If we want the latter, we should leave the status at pending and just send a notification to the evaluator.

app/grandchallenge/evaluation/tasks.py
amickan: Currently, evaluations don't have followers, it's the phases that have followers (namely the challenge admins and the submission creators). For this case, I think we would only want to notify the external evaluator that claimed the evaluation, right?

app/grandchallenge/evaluation/views/api.py
amickan: No, I would need to write a custom filter, but since we only need this here, I think doing it inside the view is clearer.

app/grandchallenge/evaluation/views/api.py
amickan: Ok, what is the best place to set `claimed_by` and `started_at` and `status` then? In the `update` method on the serializer?

app/grandchallenge/evaluation/views/api.py
jmsmkn: > Don't we normally put ValidationErrors on the evaluation as well? 

Only `ComponentError`s should be put there. `ComponentError` means this is information for the participant and the error message has been reviewed. We should not be putting `ValidationError`s there unless they are for the participant. We do not need to communicate with the participant here as the `ValidationError` comes from a call from the evaluator, who needs to fix the call they make.

The evaluator should be a user that has an unusable password so does not log in to the web interface.

The evaluator will see the error when they do:

```python
response = requests.post(...)
response.raise_for_status()
```

They will then hopefully get notified by their internal systems. They can then fix it and re-play the `post` command. They would not need to claim the job again. If they do not do this then we will mark it as timed out.

app/grandchallenge/evaluation/views/api.py
jmsmkn: You can use then permission class then filter afterwards, right?

app/grandchallenge/evaluation/views/api.py
amickan: Ok, so we just raise the error, that's it.

app/grandchallenge/evaluation/views/api.py
jmsmkn: Yes - serializers are the DRF equivalent of forms, so any modification of the objects should be done there.

jmsmkn:
Screenshot 2024-07-15 at 14 19 21

I get this error displayed if I create an external algorithm phase, but users can already submit and validation is no problem. I think the warning can just be removed?

jmsmkn:
Screenshot 2024-07-15 at 14 21 46

Should this info box be with the algorithm selection? Looks a bit weird under the form.

app/grandchallenge/evaluation/views/api.py
jmsmkn: I think so! Double check the DRF docs.

amickan:
> Another thing - do the compute costs for the evaluation get set correctly, i.e. they should be zero.

Hm, how would I best do that? I see that the compute_cost_euro_millicents get set on the executor. We wouldn't want to change that. Would it be best to set it to 0 when updating the evaluation object?

app/grandchallenge/evaluation/views/api.py
amickan: Setting the permission_class doesn't filter the queryset by permission. Normally on a list view, the `ObjectPermissionsFilter` does that, so I'd need a custom version of that filter to achieve this, but that's not more efficient than what I'm doing now, right?

jmsmkn:
The executor shouldn't need to know about our internal database codes. They should be able to do:

```python
response = requests.patch(
            f"{GRAND_CHALLENGE_API_URL}/evaluations/{evaluation['pk']}/update_external_evaluation/",
            json={"metrics": metrics, "status": "Succeeded"},
            headers={"Authorization": f"BEARER {API_BEARER_TOKEN}"},
        )
```

Currently that returns a validation error `{'status': ['"Succeeded" is not a valid choice.']}`. I have to use `{"metrics": metrics, "status": 4}` instead, but that is far too cryptic.

jmsmkn:
The evaluation objects do not contain any information about what phase they belong to, nor can you filter `claimable_evaluations` by the phase. `/evaluations/` are filterable by `submission__phase`, where a phase `pk` is provided, that should be sufficient. Serialising the Phase pk on the evaluation would be sufficient, we do not need a new Phase API endpoint.

app/grandchallenge/evaluation/serializers.py
```diff
@@ -74,3 +90,113 @@ class Meta:
...
+            )
+            civ = ComponentInterfaceValue(
+                interface=interface, value=validated_data["metrics"]
+            )
+            try:
+                civ.full_clean()
+                civ.save()
+                instance.outputs.add(civ)
+            except ValidationError as e:
+                raise DRFValidationError(e)
```
jmsmkn, line 189: Shouldn't the CIV validation be done in `validate` and then only saved in update? This is how it should be in Django, not sure how DRF differs w.r.t. handling validation errors here.

app/grandchallenge/evaluation/views/api.py
```diff
@@ -23,3 +57,96 @@ class EvaluationViewSet(ReadOnlyModelViewSet):
...
+                status=403,
+            )
+
+        if (
+            evaluation.started_at - now()
+        ).seconds > settings.EXTERNAL_EVALUATION_TIMEOUT_IN_SECONDS:
+            cancel_external_evaluations_past_timeout.apply_async()
+            return Response(
+                {
+                    "status": "You can only update an evaluation within 24 hours."
```
jmsmkn, line 137: Hardcoded timeout, you can make the message more general in case we change it in settings (and the error message should not change with our settings in case the client uses it for matching).

app/grandchallenge/notifications/models.py
```diff
@@ -354,28 +354,32 @@ def print_notification(self, user):  # noqa: C901
             self.type
             == NotificationType.NotificationTypeChoices.EVALUATION_STATUS
             and self.actor != user
-            and self.message == "failed"
+            and self.message != "succeeded"
         ):
+            if self.action_object.error_message:
+                error_message = format_html(
```
jmsmkn, line 360: Please use the `kwargs` style of `format_html`, I think it makes it easier to parse, e.g. https://github.com/comic/grand-challenge.org/blob/58ec92ab2b3b1f4aecaeebf672729600f0f8b421/app/grandchallenge/challenges/admin.py#L69-L76

Also update any usages you happen to touch.

app/grandchallenge/evaluation/serializers.py
amickan: Hm, true, I had put this in update() because we're also updating the evaluation object there at the same time, but I see now that this makes more in validate().

amickan:
> I think that this is still missing:

The filter is there, it's just not shown in the UI for some reason. I think this is an issue with the swagger UI. I couldn't figure out how to fix it and thought as long as filtering is possible through the url, it's fine, but if we do want this in the UI as well, then maybe making "claimable" a filter option would be a good alternative implementation. I don't immediately know how we would add the permission check to that filter, but I can find out. 

Do you want me to prioritize this, or do you want to pick this up and I work on getting the job creation pitch into a shareable state?

jmsmkn:
> > I think that this is still missing:
> 
> The filter is there, it's just not shown in the UI for some reason. I think this is an issue with the swagger UI. I couldn't figure out how to fix it and thought as long as filtering is possible through the url, it's fine, but if we do want this in the UI as well, then maybe making "claimable" a filter option would be a good alternative implementation. I don't immediately know how we would add the permission check to that filter, but I can find out.
> 
> Do you want me to prioritize this, or do you want to pick this up and I work on getting the job creation pitch into a shareable state?

Ah okay, is there a test for that? I would prioritise the Job creation, this is best left until after your vacation anyway.

amickan:
> > > I think that this is still missing:
> > 
> > 
> > The filter is there, it's just not shown in the UI for some reason. I think this is an issue with the swagger UI. I couldn't figure out how to fix it and thought as long as filtering is possible through the url, it's fine, but if we do want this in the UI as well, then maybe making "claimable" a filter option would be a good alternative implementation. I don't immediately know how we would add the permission check to that filter, but I can find out.
> > Do you want me to prioritize this, or do you want to pick this up and I work on getting the job creation pitch into a shareable state?
> 
> Ah okay, is there a test for that? I would prioritise the Job creation, this is best left until after your vacation anyway.

I have not added a test for the filter yet. I will work on Job creation for the rest of the day, and then return to this pitch after my vacation.

amickan:
Adding the phase_pk filter to the extra router action was actually easy to with a schema extension. I've also added a test for the filter. The PR is now ready for a final round of review, or merging.

jmsmkn:
@amickan could you solve the conflicting files?

amickan:
I solved the conflicts, and also saw that I added a test with `TestCase`, which I will refactor to use a pytest fixture instead. I will remove my assignment once I'm done with that.

amickan:
Took a bit longer than expected due to me not knowing that the order in which settings are overridden in tests makes a difference. The PR is now ready for review again.

jmsmkn (approved):

PR #3421, opened 2024-07-04
Remove jquery formset, use standard crispy implementation

### LINKED ISSUE #3403: Question choice options do not get stored on initial save
To reproduce:

- Create a reader study
- Add a question, set the type to Choice, Add an option
- Click save
- Click on edit question, notice that the options are gone. Adding them here works.

jmsmkn:
I couldn't find a good way to disable the delete button when the question is not-editable, so I added a validation error instead.

Closes #3403


jmsmkn:
Note that there is a slight change here - I decided not to delete all of the choices if you switch question type. That felt a bit risky to me. This way, if they change the question type back again then their choices are already there. It could be that we serialize the choices now for questions that do not need or use them, maybe that should be disabled but also maybe it doesn't matter as CIRRUS should just ignore them.

app/grandchallenge/core/static/js/formset.mjs
```diff
@@ -1,10 +1,17 @@
...
-    addCssClass: "btn btn-primary my-2",
-    deleteCssClass: "btn btn-danger delete-row mb-2",
-    hideLastAddForm: true,
+    addButton.addEventListener("click", addFormRow);
+
+    function addFormRow(e) {
+        e.preventDefault();
+
+        const formNum = parseInt(totalForms.value) + 1;
+        const newForm = emptyForm.replace(/__prefix__/g, totalForms.value);
```
jmsmkn, line 13: This JS isn't tested, but we have playwrite tests now so maybe it should be added? Otherwise, just checking that the template is in the response and that `__prefix__` is still used by Django is enough. Honestly though, I felt that Django is stable enough that the API on this bit wouldn't change...

jmsmkn: I also have not tested this with multiple formsets per page. This is the only place that it is used so I think it's okay. Didn't love working with formsets so unsure if we would repeat this pattern. One place where it would be useful is in the Scoring options of the phase where we have people provide awkward JSON objects.

jmsmkn:
Also when reviewing note that the implementation is much smaller - the line count is high as I'm replacing a load of untested JS with test fixtures!

app/grandchallenge/reader_studies/forms.py
```diff
@@ -504,9 +570,18 @@ class Meta:
 
 
 class CategoricalOptionForm(ModelForm):
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, is_editable, **kwargs):
         super().__init__(*args, **kwargs)
+
         self.fields["title"].label = False
+        self.fields["title"].required = True
```
amickan, line 577: This doesn't seem to work. I can add an empty option.

jmsmkn: Are you sure? There is an empty option by default. Does it actually end up in the db?

jmsmkn: I cannot replicate this, how did you do it?

jmsmkn: It's controlled by the `extra` argument to the formset factory https://github.com/comic/grand-challenge.org/blob/00171b53cd1b23b3249be18bf355f57b4122b070/app/grandchallenge/reader_studies/forms.py#L596. I did set that to zero but turned it back to 1 as then you get an empty form to start with the save the user a click. The `hideLastAddForm` option then deleted it if there was no prior data but I didn't add back that functionality. Probably the easiest thing is to set `extra` to zero and then see if you prefer that.

app/grandchallenge/core/static/js/formset.mjs
amickan: I don't have a very strong opinion. I guess a playwright test would be nice to make sure we get notified when things change, and it would be a pretty simple test as well. But I do agree that it is unlikely to change and so also fine if it doesn't have a test. There are other js-snippets in our code base that would benefit more from E2E tests.

app/grandchallenge/reader_studies/forms.py
amickan: Not a strong preference for either or: when there already are options, I prefer if the empty option is not there. When there are no options to begin with, I do prefer the empty one because it makes it more obvious that you need to fill something in there. So yeah, leave it as is.

amickan (dismissed):
LGTM! Maybe @HarmvZ wants to take a look at the JS stuff. Everything else looks good to me. I also tested it out with and without answers, switching questions types, removing, adding and whatnot and it all seems to work like a charm!

app/grandchallenge/core/static/js/formset.mjs
jmsmkn: In the interests of available time I will leave it. I think this is fine for now.

app/grandchallenge/reader_studies/forms.py
jmsmkn: I have made a PR that addresses this (#3453), please test it out extensively to see if it still performs as expected! I have targeted this branch so can be merged here if it's okay.

HarmvZ:
The js looks good to me. I do have some suggestions that would improve the UI and usability.

- Put the Default and Delete checkboxes on the same line as the text box.
- Make the Default checkbox work like a radio button, allowing only a single one being selected at a time.
- Do not use a checkbox for Delete but a button that removes the form. (This is maybe too big of an overhaul)

HarmvZ (approved):
My suggestions are minor, we can merge this PR already in the meantime

jmsmkn:
> The js looks good to me. I do have some suggestions that would improve the UI and usability.
> 
> * Put the Default and Delete checkboxes on the same line as the text box.
> * Make the Default checkbox work like a radio button, allowing only a single one being selected at a time.
> * Do not use a checkbox for Delete but a button that removes the form. (This is maybe too big of an overhaul)

I agree and looked at it when preparing this PR. However, how to get that done using Crispy Forms was not obvious and it turned into too much of a rabbit hole. For an admin form I think this is fine.

PR #3422, opened 2024-07-05
Reduce queries starting jobs

chrisvanrun:
Part of the pitch:
- https://github.com/DIAGNijmegen/rse-roadmap/issues/331

Collection of small optimisations, effectively reducing the number of database queries. Since all queries we've recorded for a single job creation are ~2ms, there is only the number of queries to reduce during Job creation.

`Main` number of queries => PR number of queries
1 job: 
- 104 => 81 (-22%)

32 jobs (batch limit)
- Main: 2402 => 1910 (-20%)

In theory, this should push things to ~35 jobs/min on production (up from ~30 jobs/min).

### DIFF
--- a/app/grandchallenge/evaluation/tasks.py
+++ b/app/grandchallenge/evaluation/tasks.py
@@ -174,7 +174,11 @@ def create_algorithm_jobs_for_evaluation(
     Evaluation = apps.get_model(  # noqa: N806
         app_label="evaluation", model_name="Evaluation"
     )
-    evaluation = Evaluation.objects.get(pk=evaluation_pk)
+    evaluation = Evaluation.objects.select_related(
+        "submission__phase__challenge",
+        "submission__algorithm_image__algorithm",
+        "submission__phase__archive",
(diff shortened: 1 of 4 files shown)


app/grandchallenge/evaluation/tasks.py
```diff
@@ -218,6 +222,12 @@ def retry_with_delay():
 
     evaluation.update_status(status=Evaluation.EXECUTING_PREREQUISITES)
 
+    civ_sets = [
```
chrisvanrun, line 225: This might be a strange optimisation, but this increases the speed at which the `cache.lock` releases and hence indirectly the job-creation throughput overall.

jmsmkn:
This is good but a lot of this adds complexity (through no fault of our own, why do `add_perm` and `remove_perm` have different APIs?). This gives a 15% speed up, but really we need a 10000% speed up to do what we want. That is clearly not going to come from optimising the current SQL queries, it needs a different design altogether. I will close this PR so that the code stays cleanish for future reworking.

PR #3427, opened 2024-07-09
Add registration vector field component interfaces

pkcakeout:
Implement displacement and deformation field component interfaces that include special image format checks.

Pitch https://github.com/DIAGNijmegen/rse-roadmap/issues/329

Question: Is there a blueprint for creating an image-upload test that verifies that the import (and the image shape) work correctly? I know that I could do a `Image.create(shape=[123, 123, 123])` and feed that into a component interface, but that makes the test rather silly. I much rather would like to write a test that uploads a known-good and a known-bad displacement field and check that the verifications run correctly. However, that requires that one waits on celery queues etc, I think?!

I am sure that I could figure this out myself, but wanted to ask first before wasting 4+ hours on finding all the right endpoints... is there a nice blueprint for such a test?


app/grandchallenge/workstations/apps.py
```diff
@@ -11,7 +11,7 @@ def init_default_workstation(*_, **__):
 
     try:
         Workstation.objects.get(slug=settings.DEFAULT_WORKSTATION_SLUG)
-    except ObjectDoesNotExist:
+    except (ObjectDoesNotExist, Workstation.DoesNotExist):
```
jmsmkn, line 14: This should be unnecessary.

app/grandchallenge/components/models.py
```diff
@@ -112,6 +112,8 @@ class InterfaceKindChoices(models.TextChoices):
     IMAGE = "IMG", _("Image")
     SEGMENTATION = "SEG", _("Segmentation")
     HEAT_MAP = "HMAP", _("Heat Map")
+    DISPLACEMENT_FIELD = "DSPF", _("Displacement field")
+    DEFORMATION_FIELD = "DEFF", _("Deformation field")
```
jmsmkn, line 116: What is the difference between the two? They seem to share all of the same validation methods.

app/grandchallenge/components/models.py
```diff
@@ -1108,7 +1127,7 @@ def _clean_store_in_database(self):
         object_store_required = self.kind in {
             *InterfaceKind.interface_type_image(),
             *InterfaceKind.interface_type_file(),
-            # These values can be large, so for any new interfaces of this
+            # These values can beValidatio large, so for any new interfaces of this
```
jmsmkn, line 1130: Typo

jmsmkn:
> Question: Is there a blueprint for creating an image-upload test that verifies that the import (and the image shape) work correctly? 

You can tell celery to working inline and use the `create_raw_upload_image_session` helper See https://github.com/comic/grand-challenge.org/blob/494b92be405c97369419a8963120784df9e563c7/app/tests/cases_tests/test_background_tasks.py#L222

app/grandchallenge/components/models.py
pkcakeout: The values in the vector-field are different, I cannot quite quote the definition that Alessa and people from FME give because I always found it a bit confusing, but I can write it down using math:

Let's assume we want to map a position $\overrightarrow{v_A}$ to $\overrightarrow{v_B}$ using one of the two types of vector fields. This works like this:

- $D$ as a deformation field: $\overrightarrow{v_B} = D\( \overrightarrow{v_A} \)$
- $D$ as a displacement field: $\overrightarrow{v_B} = \overrightarrow{v_A} + D\( \overrightarrow{v_A} \)$

And that is all. So the semantics of the two are different, but the two cases can easily be implemented in one go in CIRRUS, so I added both.

app/grandchallenge/workstations/apps.py
pkcakeout: This was an accidental change stemming from debugging... will undo it!

app/grandchallenge/components/models.py
pkcakeout: Looks like a butterfinger-paste :-)

app/grandchallenge/components/models.py
jmsmkn: If we want different types in GC we need to be able to have some rule that distinguishes between the two, otherwise they should be the same type as far as GC is concerned. What would that rule be?

pkcakeout: > rule that distinguishes between the two

Not _quite sure_ what you mean here. The difference between a deformation field and a displacement field is the same as an MRI and an CT scan, I think. The data format they use for storing is the same (3d vector fields) but the semantics are different, which I thought we encode in the ComponentInterface. It is like `cardiac-ct` vs `cardiac-mri`, but this time also adding a different _kind_ to allow CIRRUS to lock on to the different semantics in a structured way.

Would you suggest another kind of property that we add to a component interface? That would work, too. Ultimately, a user needs to tag if something is a deformation of displacement field. There is no way for us to know what the thing they upload is - just like with MRI vs CT scans (ignoring the existence of potential related DICOM tags for second here).

jmsmkn: For example, what is a heat map and what is a segmentation? A Heat map has continuous values, a Segmentation has discrete values. When we get something new we need to be able to distinguish between the two, and we do that on validation. If we have two new types we need to be able to distinguish between them somehow. Otherwise it should be one type.

jmsmkn: Comes down to - if a user provides a DISPLACEMENT_FIELD when a DEFORMATION_FIELD is expected it should be rejected at import time, not at CIRRUS run time.

pkcakeout: > Comes down to - if a user provides a DISPLACEMENT_FIELD when a DEFORMATION_FIELD is expected it should be rejected at import time, not at CIRRUS run time.

Ok, I see. Problem with this is that the two _types_ of vector fields are _fully compatible_. A displacement field looks _exactly_ like a deformation field. Only the offsets are different leading to an incorrect interpretation when attempting to use a deformation field as a displacement field and vice versa. CIRRUS cannot detect this either when shown, the only thing that would happen is that the mapped coordinates are all wrong.

To decide if a displacement field is "correct" one needs to evaluate if it actually "does the coordinate mapping correctly" which is something we cannot compute. (I would not even algorithmically know how to do this in a simple and reliable way...)

Also, right now it is possible to upload a heat map as a mri image, right? And one could upload a displacement field as a mri image which then would be interpreted as a 3d MRI image with multiple time points (namely 3). Therefore, I would argue that we already make semantic decisions based on the interface-kind. I can see that displacement and deformation fields are more closely related than, for example, an 4d image vs displacement fields and it might be sub-optimal to distinguish between these types on the interface-kind level. But we need the user to tell us what they actually uploaded.

From the little maths-section I show above, you can derive that deformation fields are equivalent to displacement fields as long as image $A$ is available. Having image $A$ allows one to add or remove the offset $\overrightarrow{v_A}$ from $D$ to convert one into the other type. In CIRRUS this can be done because we have image $A$. But also here the user has to tell us in which direction this conversion has to be applied to make sense.

Only solutions I see here are:
- Drop support for deformation fields to avoid this confusion
    - Alessa told me that this mostly would work, but I know that it would be somewhat unfortunate because it requires existing registrations to be recomputed. Sort of a bike-shedding thing that we would do here. There are pretty much no complication of _applying_ the conversion in CIRRUS so this would be an unfortunate miss, imo.
- Collect the deformation/displacement semantics in some other way during the upload process - happy to hear suggestions here. As I outlined, there is unfortunately no way for us to automatically latch on to the type of vector field that is uploaded because they use the exact same format.
- (Keep this implementation... of course this is an option, too)

jmsmkn: > Drop support for deformation fields to avoid this confusion

Do this, that way it fits in with the existing design of the application.

pkcakeout:
@jmsmkn Can I add simpletik to the test-dependencies to generate mha test data on-the-fly or should I prefer adding a valid displacement field in a mha-file as binary blob (~1 kB, yes they are typically _tiny_, especially for 2d. The sample that I am using for testing CIRRUS is only 800 bytes).

I tend to prefer the binary blob?

pkcakeout:
... `SimpleITK` ... I am a bit confused. It is actually there, right? At least from what I read. It seems to be a transitive dependency because we import some other package, but then SimpleITK is used directly in the codebase. But it is not mentioned in the `pyproject.toml` file, which derailed me a bit.

Is this an oversight? I assume if the codebase uses SimpleITK directly, it should specify that it uses it, right?

jmsmkn:
Binary blob. Simpleitk is installed.

jmsmkn:
If you can add tests that do not depend directly on simple itk then that is better.

app/grandchallenge/components/models.py
pkcakeout: Done!

pkcakeout:
... any idea why there is a failing test? I might have expected the `pull_request` tests to fail if there was a bug after merging. Is this a flaky test? `works_on_my_machine() == true`...

app/tests/utils.py
```diff
@@ -230,3 +233,30 @@ def recurse_callbacks(callbacks, django_capture_on_commit_callbacks):
...
+
+
+def create_raw_upload_image_session(
+    *,
+    django_capture_on_commit_callbacks,
+    image_paths: list[str],
+    delete_files: list[str] = None,
+    user=None,
+    linked_task=None,
+) -> tuple[RawImageUploadSession, dict[str, UserUpload]]:
```
pkcakeout, line 245: I made this function more generic and promoted it to an official "utils" function for it to be available across tests. I hope that is alright?

jmsmkn:
I have re-run the tests. The notification ones are sometimes flaky. Looks irrelevant.

app/tests/utils.py
jmsmkn: Would prefer refactoring to be separate from feature PRs, but it's fine.

jmsmkn (approved):

PR #3430, opened 2024-07-10
Fix acquisition of locks for rank calculations

jmsmkn:


app/grandchallenge/components/tasks.py
```diff
@@ -1170,18 +1171,21 @@ def assign_tarball_from_upload(
...
+            TarballModel.objects.filter(
+                pk=tarball_pk,
+                import_status=TarballModel.ImportStatusChoices.INITIALIZED,
+            )
             .select_for_update(nowait=True)
             .get()
         )
-        peer_tarballs = current_tarball.get_peer_tarballs().select_for_update(
-            nowait=True
+        peer_tarballs = list(
```
jmsmkn, line 1184: @amickan just FYI - to acquire locks for a queryset it needs to be evaluated, so casting this to a list as it is anyway done later.

PR #3440, opened 2024-07-12
Remove usage of datetime.now

jmsmkn:
This doesn't catch usages of `datetime.now` due to an existing bug in `flake8-tidy-imports` but I might try to fix that.

### DIFF
--- a/app/config/settings.py
+++ b/app/config/settings.py
@@ -1,7 +1,7 @@
 import os
 import re
 import socket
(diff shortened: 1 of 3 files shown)


amickan (approved):

jmsmkn:
I tried to fix the bug but it is really difficult. Learnt a lot about the `ast` module though so not all wasted.

PR #3447, opened 2024-07-17
Fix evaluation filters to correctly include model or not

### LINKED ISSUE #3438: Potential issues with model submission
I think there is a potential bug in `clean_algorithm` in `SubmissionForm`. I think the assumption is that a particular image would either always be submitted in isolation, or always submitted with a model. But of course an algorithm image can be submitted with or without a model. I think if `algorithm.active_model` is `False` we should add:
 
```python
            extra_submission_filter = {
                "algorithm_model__isnull": True
            }
            extra_evaluation_filter = {
                "submission__algorithm_model__isnull": True
            }
```

to https://github.com/comic/grand-challenge.org/blob/b085188e6f13cde718cf8fe4b3b1912c9c592da7/app/grandchallenge/evaluation/forms.py#L473.
 
Then if so, I notice there is a comment warning about `set_evaluation_inputs`. I think there we also need a filter for the active model or None in two places:

https://github.com/comic/grand-challenge.org/blob/28296b2689744798e309352536f4c7005891601a/app/grandchallenge/evaluation/tasks.py#L277 

and

https://github.com/comic/grand-challenge.org/blob/28296b2689744798e309352536f4c7005891601a/app/grandchallenge/evaluation/tasks.py#L319 
 
If so, there might also be something missing from `create_algorithm_jobs`, but I'm not sure there.

amickan:
Closes #3438

### DIFF
--- a/app/grandchallenge/evaluation/tasks.py
+++ b/app/grandchallenge/evaluation/tasks.py
@@ -278,7 +278,7 @@ def set_evaluation_inputs(*, evaluation_pk):
             "algorithm_model__submission__evaluation": evaluation_pk
         }
     else:
-        pending_jobs_extra_filter = {}
+        pending_jobs_extra_filter = {"algorithm_model__isnull": True}
 
     has_pending_jobs = (
         Job.objects.active()
@@ -328,7 +328,7 @@ def set_evaluation_inputs(*, evaluation_pk):
             "algorithm_model": evaluation.submission.algorithm_model
         }
     else:
-        extra_filter = {}
+        extra_filter = {"algorithm_model__isnull": True}
 
     successful_jobs = (
         Job.objects.filter(
--- a/app/grandchallenge/evaluation/tasks.py
+++ b/app/grandchallenge/evaluation/tasks.py
@@ -9,7 +9,7 @@ from django.db.models import Count, Q
 from django.db.transaction import on_commit
 
 from grandchallenge.algorithms.exceptions import TooManyJobsScheduled
-from grandchallenge.algorithms.models import Job
+from grandchallenge.algorithms.models import AlgorithmModel, Job
 from grandchallenge.algorithms.tasks import create_algorithm_jobs
 from grandchallenge.components.models import (
     ComponentInterface,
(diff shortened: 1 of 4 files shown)


app/grandchallenge/evaluation/tasks.py
```diff
@@ -313,10 +323,18 @@ def set_evaluation_inputs(*, evaluation_pk):
         .prefetch_related("values")
     }
 
+    if evaluation.submission.algorithm_model:
+        extra_filter = {
+            "algorithm_model": evaluation.submission.algorithm_model
+        }
+    else:
+        extra_filter = {}
```
jmsmkn, line 331: Don't you want to set `"algorithm_model__isnull": True` here? Same above for `pending_jobs_extra_filter`

jmsmkn:
I just merged #3449 but I think this also needs to be merged before next deploying otherwise it could lead to problems downstream.

app/grandchallenge/evaluation/tasks.py
amickan: Yes, clearly my tests need to be expanded! Working on that now.

amickan: Suffering from brain fog this morning, but I think the tests now cover all situations. I think the tests could be improved, they are hard to read at the moment, but I couldn't find a nice way to parametrize the different situations...

jmsmkn: I spent ages yesterday fighting a test that was far too long. The ones in test_tasks are short and clear, and no logic. Perfect. Easy to understand, easy to debug later. If anything `test_eval_exists_for_image_and_model` is too smart - two if statements which really shouldn't be in tests. A better approach may be to use two tests with module scoped fixtures. But, whatever.

jmsmkn (approved):

jmsmkn:
I'll leave you to merge as I'm not sure whether you're complete with the other methods you mentioned in the issue.

amickan:
> I'll leave you to merge as I'm not sure whether you're complete with the other methods you mentioned in the issue.

Let's merge, I will create a separate PR for the other places.

PR #3448, opened 2024-07-17
Take active ground truth into account for submission re-evaluation button

amickan:
The "Re-evaluate submission" would only show when the method was updated, not when the ground truth was updated.

### DIFF
--- a/app/grandchallenge/evaluation/models.py
+++ b/app/grandchallenge/evaluation/models.py
@@ -1177,16 +1177,11 @@ class Submission(UUIDModel):
         active_image = self.phase.active_image
         active_ground_truth = self.phase.active_ground_truth
         if active_image:
-            if active_ground_truth:
-                return Evaluation.objects.filter(
-                    submission=self,
-                    method=active_image,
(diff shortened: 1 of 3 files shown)


app/grandchallenge/evaluation/models.py
```diff
@@ -1173,12 +1173,20 @@ class Meta:
...
+            if active_ground_truth:
+                return Evaluation.objects.filter(
+                    submission=self,
+                    method=active_image,
+                    ground_truth=active_ground_truth,
+                ).exists()
+            else:
+                return Evaluation.objects.filter(
+                    submission=self, method=active_image
+                ).exists()
```
jmsmkn, line 1189: Doesn't this work?

```suggestion
            return Evaluation.objects.filter(
                submission=self,
                method=active_image,
                ground_truth=active_ground_truth,
            ).exists()
```

amickan: Ah, yes, that works and is necessary, or else we match too many evaluations

jmsmkn (approved):

PR #3450, opened 2024-07-17
Fix failing rendering of archive item algorithm jobs page

### LINKED ISSUE #3434: Archive Item jobs page does not render
In `/archives/{slug}/items/{pk}/jobs/`

AttributeError
'str' object has no attribute 'slug'

Where `workstation=""` in method `workstation_session_control_data`.

ammar257ammar:
The issue was in a defined variable that was mistakingly put within double quotes. That lead, in a downstream call, to raise an error when attempting to access an attribute on that "string". 

closes #3434 

The issue manifested in that datatables (JS dependency) was not able to render a table of the jobs related to an archive item. The spinner appears on the screen with an error message (browser alert) as shown below:

Screenshot 2024-07-17 at 20 48 02

After the fix, the table renders correctly, as shown below:

Screenshot 2024-07-17 at 20 50 17


jmsmkn:
Whenever you approach to fix something you first need to write a test that fails as it is clear that our tests do not cover this functionality. Calling the api view and checking for status 200 would be a minimum, you could even check for a piece of the html that is missing from the response. Please add a test for this.

Now that you've found the problem, does it exist anywhere else? People often make the same mistake and it could cause a bug elsewhere. In this case, it does. See if you can find it.

jmsmkn:
Please also see https://cbea.ms/git-commit/ and update your PR title.

jmsmkn:
This still needs a test.

app/tests/algorithms_tests/test_views.py
```diff
@@ -1367,3 +1369,20 @@ def test_algorithm_model_version_management(settings, client):
...
+
+@pytest.mark.django_db
+def test_job_list_row_template_renders(
+    client, settings, django_capture_on_commit_callbacks
+):
+    job = AlgorithmJobFactory()
+
+    assert job.algorithm_image
+
+    job.status = Job.SUCCESS
```
amickan, line 1381: You can add the status directly to the factory. Also, in most cases, you will need to save the object again after modifying it. In this case, it doesn't matter, but in most other cases it does.

app/tests/algorithms_tests/test_views.py
```diff
@@ -1367,3 +1369,20 @@ def test_algorithm_model_version_management(settings, client):
...
+    
+    response = render_to_string(
+        template_name="algorithms/job_list_row.html",
+        context={"object": job},
+    )
+
+    assert response.find("View algorithm result") != -1
```
amickan, line 1388: I think you would wanna check for the actual link here as well not just the column header.

ammar257ammar: The string is not the column header, but the button title which will not be there if the link is not generated

app/tests/algorithms_tests/test_views.py
ammar257ammar: Sure, I will keep that in mind. Should I pass the status to the factory and save the object for this PR?

app/tests/algorithms_tests/test_views.py
amickan: Oh sorry! But I think it still makes sense to check that the button has the expected link set. You could also go one step further and test that all other columns are filled as expected to have a full test. You call the test `test_job_list_row_template_renders`, so I would expect it to test the full template, not just one of the buttons.

app/tests/algorithms_tests/test_views.py
```diff
@@ -1367,3 +1368,21 @@ def test_algorithm_model_version_management(settings, client):
     assert not m2.is_desired_version
     del alg.active_model
     assert not alg.active_model
+
+
+@pytest.mark.django_db
+def test_job_list_row_template_renders(
+    client, settings, django_capture_on_commit_callbacks
```
amickan, line 1375: settings and django_capture_on_commit_callbacks can go here and below

app/tests/archives_tests/test_views.py
```diff
@@ -1343,3 +1345,21 @@ def make_request():
...
+def test_job_list_row_template_renders(
+    client, settings, django_capture_on_commit_callbacks
+):
+    job = AlgorithmJobFactory()
+
+    assert job.algorithm_image
+
+    job.status = Job.SUCCESS
+
+    response = render_to_string(
```
jmsmkn, line 1360: These two tests only test that they're rendering here, not that they're rendering correctly in our application. You cannot be sure that these templates will be used by the view that was broken. You cannot be sure that the template was correctly parametrised. To test a view you need to use the client to make a call.

ammar257ammar: The case here is that the call to render the table rows in the view is asyncronous (made as an AJAX call by JQuery/datatables). So, when testing the view with the client call, it will not fail, since the actual problematic rendering happens in the AJAX call. Therefore, I thought of testing the rendering of the row template with an job object in the context and check for the presence of the button (i.e. rendered successfully) in the output string. 

How do you propose testing for dynamic Javascript using Django view client calls?

jmsmkn: Make the same call that the AJAX request makes.

app/tests/algorithms_tests/test_views.py
```diff
@@ -1367,3 +1370,104 @@ def test_algorithm_model_version_management(settings, client):
...
+            "columns[3][name]": "",
+            "columns[3][searchable]": "true",
+            "columns[3][orderable]": "true",
+            "columns[3][search][value]": "",
+            "columns[3][search][regex]": "false",
+            "columns[4][data]": "4",
+            "columns[4][name]": "",
+            "columns[4][searchable]": "true",
+            "columns[4][orderable]": "true",
+            "columns[4][search][value]": "",
```
amickan, line 1434: I don't think you need all this data to run the test, do you? Restrict it to the minimum of what's necessary, or else it wont be clear what you're testing here.

ammar257ammar: Indeed, I updated the tests with minimum parameter needed for the view test

amickan (approved):
LGTM!

PR #3452, opened 2024-07-18
Add missing algorithm model filters

amickan:
Part of #3438

### DIFF
--- a/app/grandchallenge/algorithms/tasks.py
+++ b/app/grandchallenge/algorithms/tasks.py
@@ -253,7 +253,7 @@ def create_algorithm_jobs(
 
 
 def filter_civs_for_algorithm(
-    *, civ_sets, algorithm_image, algorithm_model=None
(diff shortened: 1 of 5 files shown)


app/grandchallenge/algorithms/tasks.py
```diff
@@ -250,7 +252,9 @@ def create_algorithm_jobs(
     return jobs
 
 
-def filter_civs_for_algorithm(*, civ_sets, algorithm_image):
+def filter_civs_for_algorithm(
+    *, civ_sets, algorithm_image, algorithm_model=None
```
jmsmkn, line 256: I think we should be explicit to prevent bugs upstream:

```suggestion
    *, civ_sets, algorithm_image, algorithm_model
```

jmsmkn (dismissed):

jmsmkn (dismissed):

jmsmkn (approved):

PR #3453, opened 2024-07-18
Address PR feedback for CategoricalOptions form

jmsmkn:
Quick implementation to address the PR feedback before I go. Not extensively tested but see what you think! This:

- only adds an extra field on create
- when not fully editable: disables the add button and removes the delete checkboxes

### DIFF
--- a/app/grandchallenge/core/layout.py
+++ b/app/grandchallenge/core/layout.py
@@ -5,8 +5,16 @@ from django.template.loader import render_to_string
 class Formset(LayoutObject):
     template = "layout/formset.html"
 
-    def __init__(self, formset):
-        self.formset = formset
+    def __init__(self, *args, formset, can_add_another, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._formset = formset
+        self._can_add_another = can_add_another
 
     def render(self, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


HarmvZ:
There is some strange behavior here. Recipe:

1. Create a new question. Add a question text and select "choice" type.
1. Click "add another" while not filling in any option.
2. Try to remove any of the options by checking delete and click save.
3. This gives a validation error: "At least one option should be supplied" as expected, but it also shows "This field is required" on the option that had the "Delete" option checked. 

image

amickan:
> There is some strange behavior here. Recipe:
> 
> 1. Create a new question. Add a question text and select "choice" type.
> 2. Click "add another" while not filling in any option.
> 3. Try to remove any of the options by checking delete and click save.
> 4. This gives a validation error: "At least one option should be supplied" as expected, but it also shows "This field is required" on the option that had the "Delete" option checked.
> 
> image

This is not specific to this branch/PR and is also the case on the main PR. I also see that this only happens when you add no options for a Choice or MultipleChoice type. The general error message that choice questions require at least one option is present at the top of the page. It's a bit strange that the to-be-deleted option is marked as required, but this only happens in this very specific case where the user needs to go back and fix the options anyway, so I don't see this as blocking and will merge it.

PR #3454, opened 2024-07-22
Update UI running algorithm card to reflect average job duration

### LINKED ISSUE #3429: Algorithm duration predictions are wildly optimistic
Container pulling time, provisioning and parsing times are not taken into account:

Screenshot 2024-07-10 at 11 26 53

ammar257ammar:
Since running a job involves different stages other than executing the algorithm on the selected image(s) (e.g. container pulling time, provisioning and parsing times), the time viewed in the card is very optimistic. 

Therefore, it would be better (James suggestion) not to say how long is remaining, rather how long a successful job has taken on average.

This small UI update PR changes the message in the algorithm running card based on the status of the job. If the status is one of ("started", "provisioning", "provisioned"), it will show a message that the job is being provisioned.

Screenshot 2024-07-25 at 09 41 00

Otherwise, if the job has a status as one of ("executing", "executed", "parsing outputs"), it will show a message  reflecting the average job duration and informing the user the the job is already being executed. 

Screenshot 2024-07-25 at 09 42 13

closes #3429

### DIFF
--- a/app/grandchallenge/algorithms/static/algorithms/js/job_session.mjs
+++ b/app/grandchallenge/algorithms/static/algorithms/js/job_session.mjs
@@ -28,23 +28,21 @@ function handleJobStatus(job) {
     if (jobStatus === "succeeded") {
         setCardCompleteMessage(cards.job, "View Results");
     } else if (["started", "provisioning", "provisioned"].includes(jobStatus)) {
-        setCardActiveMessage(cards.job, `Job is queued for execution`);
-        setTimeout(function () {
-            getJobStatus(job.api_url)
-        }, Math.floor(Math.random() * timeout) + 100);
+        setCardActiveMessage(cards.job, `Job is being provisioned`);
     } else if (["executing", "executed", "parsing outputs"].includes(jobStatus)) {
         setCardActiveMessage(cards.job, `Job is being executed <br> Average job duration: ${moment.duration(averageJobDuration).humanize()}`);
-        setTimeout(function () {
-            getJobStatus(job.api_url)
-        }, Math.floor(Math.random() * timeout) + 100);
     } else if (jobStatus === "queued" || jobStatus === "re-queued") {
         setCardAwaitingMessage(cards.job, "Queued");
-        setTimeout(function () {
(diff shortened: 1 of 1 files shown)


ammar257ammar:
I put a screenshot to depict the update. I will not repeat the description of the issue anymore.

app/grandchallenge/algorithms/static/algorithms/js/job_session.mjs
```diff
@@ -25,12 +25,15 @@ function handleJobStatus(job) {
...
-
     if (jobStatus === "succeeded") {
         setCardCompleteMessage(cards.job, "View Results");
-    } else if (["started", "provisioning", "provisioned", "executing", "executed", "parsing outputs"].includes(jobStatus)) {
-        setCardActiveMessage(cards.job, `Started, ${moment.duration(estimatedRemainingTime).humanize()} remaining`);
+    } else if (["started", "provisioning", "provisioned"].includes(jobStatus)) {
+        setCardActiveMessage(cards.job, `Job is queued for execution`);
+        setTimeout(function () {
+            getJobStatus(job.api_url)
+        }, Math.floor(Math.random() * timeout) + 100);
```
HarmvZ, line 34: I know these lines were already repeated below, but while you're in here it would be nice to refactor this by removing the duplicate lines.

HarmvZ (changes requested):
Nice job on this Ammar! I have one refactoring comment below. 

Also I'm wondering if the text is clear enough. Now, it will show "Queued", then "Job is queued for execution" and then "Job is being executed". I wonder if the first 2 are too similar? As a user I don't think it would be clear to me what the difference is. Maybe change the second one to "Job is being provisioned" or something?

app/grandchallenge/algorithms/static/algorithms/js/job_session.mjs
ammar257ammar: Thanks for the comment Harm!
I agree on both points. I removed the duplication and changed the card text.
Testing now.

HarmvZ (approved):

PR #3457, opened 2024-07-24
Show box-shadow for card on stretched link focus

### LINKED ISSUE #3456: Unclear focus for cards when navigating by keyboard
When navigating grand challenge exclusively with a keyboard you want to be able to tab through all clickable elements and see what is currently selected. This works fine for most elements on the page, except for the cards used for challenges/readerstudies/algorithms/archives/etc. Those cards contain an anchor tag with `stretched-link` class that makes the whole card clickable. But the anchor tag itself is empty so not visible. That means it's unclear which card is currently in focus when tabbing through the elements. 

I think it would be nice if the card shows a dropped shadow when in focus, just like it shows on hover. In [bootstrap examples](https://getbootstrap.com/docs/4.3/utilities/stretched-link/) of the stretched link they always show a button or link on the card that uses the stretched link so that would make it clear what is in focus. That's also an option.

Example:

Here I'm tabbing back- and forward through all elements. There is a moment when the card is focused but this is not visible to the user (after the button above and before the info model button inside the card is focused).

https://github.com/user-attachments/assets/7a03bdf7-3118-44b5-8247-2519700d3f27

HarmvZ:
Closes #3456 

Demo
https://github.com/user-attachments/assets/c12d5c58-f56c-48ac-aaca-a136bd64e1cf

### DIFF
--- a/app/grandchallenge/core/static/css/core.css
+++ b/app/grandchallenge/core/static/css/core.css
@@ -178,8 +178,13 @@ ul.socialaccount_providers > li {
     border-radius: 0.9rem;
 }
 
-.gc-card:hover {
+.gc-card:hover, .gc-card:focus-within {
     box-shadow: 0 0 1rem 0 rgba(0, 0, 0, 0.2);
+}
+
+/* Reset card shadow when above-stretched-link elements are focused */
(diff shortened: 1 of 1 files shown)


amickan (approved):
LGTM!

PR #3459, opened 2024-07-25
Add a check for evaluations with rank 0 and missing metrics and notify admin

### LINKED ISSUE #3346: Explain why an evaluation does not show on the leaderboard
Sometimes visible evaluations have a rank of 0 as they are missing defined metrics. Add a message to the evaluation detail page in this case to inform the user which metrics are missing from their metrics.json.

ammar257ammar:
Implementing a solution to check for missing metrics and notify the admin on the Evaluation detail page. Missing metrics lead to keep the evaluation rank as zero and thus not showing in the leaderboard even when it is finished with success and published.

A property is added to the Evaluation mode to check for missing metrics in metrics.json. Moreover, a message is added to the details view notifying the admin on which metrics are missing. Model and View tests are provided for both functionalities.

Screenshot 2024-07-29 at 11 44 48

closes #3346

### DIFF
--- a/app/grandchallenge/evaluation/models.py
+++ b/app/grandchallenge/evaluation/models.py
@@ -56,10 +56,14 @@ from grandchallenge.evaluation.tasks import (
     create_evaluation,
     update_combined_leaderboard,
 )
+from grandchallenge.evaluation.templatetags.evaluation_extras import (
+    get_jsonpath,
+)
 from grandchallenge.evaluation.utils import (
     Metric,
     StatusChoices,
     SubmissionKindChoices,
+    get,
 )
 from grandchallenge.hanging_protocols.models import HangingProtocolMixin
 from grandchallenge.notifications.models import Notification, NotificationType
@@ -1402,6 +1406,35 @@ class Evaluation(UUIDModel, ComponentJob):
         for output in self.outputs.all():
             if output.interface.slug == "metrics-json-file":
                 return output.value
+
+    @property
+    def missing_metrics_json_file(self):
+        if self.status == self.SUCCESS:
+
+            output_values = [
+                o.value
+                for o in self.outputs.all()
+                if o.interface.slug == "metrics-json-file"
+            ]
+
+            if len(output_values) > 0:
+
+                if self.rank == 0:
+
+                    valid_metrics = self.submission.phase.valid_metrics
+                    missing_metrics = []
+
+                    for metric in valid_metrics:
+
+                        if get_jsonpath(get(output_values), metric.path) in [
+                            "",
+                            None,
+                        ]:
+                            missing_metrics.append(metric.path)
+
+                    return missing_metrics
+
+        return None
 
     def clean(self):
         if self.submission.phase != self.method.phase:
--- a/app/grandchallenge/evaluation/models.py
+++ b/app/grandchallenge/evaluation/models.py
@@ -1427,7 +1427,10 @@ class Evaluation(UUIDModel, ComponentJob):
                 ]:
                     missing_metrics.append(metric.path)
 
-            return missing_metrics
+            if len(missing_metrics) > 0:
+                return missing_metrics
+            else:
+                return None
         return None
 
     def clean(self):
--- a/app/grandchallenge/evaluation/models.py
+++ b/app/grandchallenge/evaluation/models.py
@@ -1408,8 +1408,8 @@ class Evaluation(UUIDModel, ComponentJob):
                 return output.value
 
     @property
-    def missing_metrics_json_file(self):
-        if self.status == self.SUCCESS:
+    def missing_metrics(self):
+        if self.status == self.SUCCESS and self.rank == 0:
 
             output_values = [
                 o.value
@@ -1417,23 +1417,17 @@ class Evaluation(UUIDModel, ComponentJob):
                 if o.interface.slug == "metrics-json-file"
             ]
 
-            if len(output_values) > 0:
+            valid_metrics = self.submission.phase.valid_metrics
+            missing_metrics = []
 
-                if self.rank == 0:
+            for metric in valid_metrics:
+                if get_jsonpath(get(output_values), metric.path) in [
+                    "",
+                    None,
+                ]:
+                    missing_metrics.append(metric.path)
 
-                    valid_metrics = self.submission.phase.valid_metrics
-                    missing_metrics = []
-
-                    for metric in valid_metrics:
-
-                        if get_jsonpath(get(output_values), metric.path) in [
-                            "",
-                            None,
-                        ]:
-                            missing_metrics.append(metric.path)
-
-                    return missing_metrics
-
+            return missing_metrics
         return None
 
     def clean(self):
--- a/app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
+++ b/app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
@@ -310,10 +310,13 @@
                         </form>
                     {% endif %}
 
-                    {% if object.rank == 0 %}
+                    {% if object.missing_metrics_json_file %}
                         <br>
-                        <div class="alert alert-warning mb-0">This submission has a rank 0 due to missing metric(s) from metrics.json <br>
-                        Therefore, it is not visible in the leaderboard.</div>
+                        <div class="alert alert-warning mb-0">
+                            This submission has a rank 0 due to missing metric(s) from metrics.json {{ object.missing_metrics_json_file }}
+                            <br>
+                            Therefore, it is not visible in the leaderboard.
+                        </div>
                     {% endif %}
                 </div>
             {% endif %}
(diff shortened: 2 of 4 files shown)


app/grandchallenge/evaluation/models.py
```diff
@@ -1403,6 +1407,35 @@ def metrics_json_file(self):
             if output.interface.slug == "metrics-json-file":
                 return output.value
 
+    @property
+    def missing_metrics_json_file(self):
+        if self.status == self.SUCCESS:
```
amickan, line 1412: I think you would want to check `self.rank` here already. No need to fetch the outputs if the rank is not `0`

app/grandchallenge/evaluation/models.py
```diff
@@ -1403,6 +1407,35 @@ def metrics_json_file(self):
...
+
+            output_values = [
+                o.value
+                for o in self.outputs.all()
```
amickan, line 1416: Wouldn't it be more efficient to filter in the database (doing `self.outputs.filter(interface__slug="metrics-json-file")` rather than retrieving all and then filtering in python?

amickan: If you do this, you would need to catch the case where there are mutliple outputs of this kind. Then you don't need to use the custom `get()` method later on.

app/grandchallenge/evaluation/models.py
```diff
@@ -1403,6 +1407,35 @@ def metrics_json_file(self):
             if output.interface.slug == "metrics-json-file":
                 return output.value
 
+    @property
+    def missing_metrics_json_file(self):
```
amickan, line 1411: To me this property name suggests that the metrics.json file is missing. Maybe rename to just `missing_metrics`? or if you want it to be more descriptive `metrics_missing_from_metrics_file` or something?

app/grandchallenge/evaluation/models.py
amickan: Does the evaluation need to be public as well for this to be an issue? The issue description would suggest so. In that case, I would check that here as well.

app/grandchallenge/evaluation/models.py
```diff
@@ -1403,6 +1407,35 @@ def metrics_json_file(self):
...
+        if self.status == self.SUCCESS:
+
+            output_values = [
+                o.value
+                for o in self.outputs.all()
+                if o.interface.slug == "metrics-json-file"
+            ]
+
+            if len(output_values) > 0:
```
amickan, line 1420: This will be redundant if you do the filtering in the database that I suggest above.

amickan:
A general comment: looking at the screenshot, it still says "This result is published on the leaderboard", even though it is not, right? That message should probably be replaced with the text you added.

app/grandchallenge/evaluation/models.py
ammar257ammar: Indeed, I now check for self.rank before. The publish status would indeed affect the message, but I think the admin should still see a message about the missing metrics, and if published, an extended message that due to missing metrics the evaluation won't appear on the leaderboard.

app/grandchallenge/evaluation/models.py
ammar257ammar: I see your point. I meant to emphasize that missing metrics here are for json file component interface only. But probably better to rename it to missing_metrics.

app/grandchallenge/evaluation/models.py
ammar257ammar: Agree, fixed

ammar257ammar:
I updated the screenshots showing the difference in the message with published/unpublished evalautions.

amickan:
I think now you still show the button to publish the result if there are missing metrics, right? That button should not be there.

ammar257ammar:
ah, I thought we need to provide the option of publishing anyway. Okay, will change it.

amickan:
> ah, I thought we need to provide the option of publishing anyway. Okay, will change it.

All that the view that this button links to does, is update the `published` field on the evaluation to `True`. So that button is only necessary if `published=False`

app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
```diff
@@ -309,6 +309,17 @@ <h3 class="card-title">Visibility</h3>
                             </button>
                         </form>
                     {% endif %}
+
+                    {% if object.missing_metrics %}
+                        <br>
+                        <div class="alert alert-warning mb-0">
+                            This evaluation has missing metric(s) from metrics.json {{ object.missing_metrics }}
```
amickan, line 316: I think I would prefer something like `The metrics.json output file for this evaluation is missing the following metrics: ... `

ammar257ammar:
> > ah, I thought we need to provide the option of publishing anyway. Okay, will change it.
> 
> All that the view that this button links to does, is update the `published` field on the evaluation to `True`. So that button is only necessary if `published=False`

So, we need to now show this button if there are missing metrics. However, if the phase settings is set to publish the evaluation then it will be published anyway and then we need to show the unpublish button if there are missing metrics.

Does that make sense?

app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
```diff
@@ -309,6 +309,17 @@ <h3 class="card-title">Visibility</h3>
...
+                            {% if object.published %}
+                                <br>
+                                Eventhough this evaluation is published, it is not visible in the leaderboard, due to the missing metrics and thus it has a rank 0
```
amickan, line 319: I think this is redundant if you change the visibility information above to `Does not appear on leaderboard`.

amickan:
> So, we need to not show this button if there are missing metrics. However, if the phase settings is set to publish the evaluation then it will be published anyway and then we need to show the unpublish button if there are missing metrics.

Try it out and see. As far as I know published evaluations with a rank of 0 (which is the case when there are missing metrics) do not appear on the leaderboard.

ammar257ammar:
Okay, I think it is not helpful to show the unpublish button anyway if the rank is 0. So, if there are missing metric, the publish/unpublish option will not appear, as below:

Do you agree?

Screenshot 2024-07-26 at 13 06 31

amickan:
> Do you agree?

You still want to show the crossed out eye with the text "This results is not published on the leaderboard", I think.

ammar257ammar:
> > Do you agree?
> 
> You still want to show the crossed out eye with the text "This results is not published on the leaderboard", I think.

You mean when the evaluation is not published, right?

Because if the phase settings is to publish evaluations, then it will be published anyway.

amickan:
> Because if the phase settings is to publish evaluations, then it will be published anyway.

Are you sure? I thought that evaluations with rank==0 do not appear on the leadboard, even if they have published set to True.

ammar257ammar:
> > Because if the phase settings is to publish evaluations, then it will be published anyway.
> 
> Are you sure? I thought that evaluations with rank==0 do not appear on the leadboard, even if they have published set to True.

I mean they will have the value of published=True, not they will visible.

app/tests/evaluation_tests/test_models.py
```diff
@@ -688,6 +688,75 @@ def test_combined_leaderboard_ranks(combined_ranks, expected_ranks):
...
+        (
+            Evaluation.SUCCESS,
+            0,
+            {"cosine": {"std": 0.1, "mean": 0.0}},
+            ["acc.mean", "dice.mean"],
+        ),
+    ),
+)
+@pytest.mark.django_db
+def test_evaluation_missing_metrics(
```
amickan, line 725: I think you are missing the case where the status is `Evaluation.SUCCESS` and the rank is 0 and no metrics are missing.

app/tests/evaluation_tests/test_views.py
```diff
@@ -1455,3 +1459,54 @@ def test_ground_truth_version_management(settings, client):
     assert not gt2.is_desired_version
     del phase.active_ground_truth
     assert not phase.active_ground_truth
+
+
+@pytest.mark.django_db
+def test_evaluation_details_zero_rank_message(client):
+
+    participant = UserFactory()
```
amickan, line 1467: I think it is a bit misleading to call this user `participant` when they are the admin / creator of the challenge.

app/tests/evaluation_tests/test_views.py
```diff
@@ -1455,3 +1459,54 @@ def test_ground_truth_version_management(settings, client):
...
+
+    challenge = ChallengeFactory(hidden=False, creator=participant)
+
+    phase = PhaseFactory(
+        challenge__hidden=False,
```
amickan, line 1472: Why don't you pass the challenge that you created above here?

app/tests/evaluation_tests/test_views.py
```diff
@@ -1455,3 +1459,54 @@ def test_ground_truth_version_management(settings, client):
...
+        public=True,
```
amickan, line 1473: `public=True` is the default right?

amickan (approved):

jmsmkn:
Please see the simplified implementation in #3464. Why does this PR mention notifying the admin? No notifications about this are sent (and rightly so).

PR #3463, opened 2024-08-01
Update dependencies

jmsmkn:
Fix sphinx rendering


jmsmkn:
Note that:

- `django-rest-knox` is upgraded. This involved a major breaking change which invalidates all existing tokens. See https://jazzband.github.io/django-rest-knox/changelog/. But this might just be a duplicate warning. Have asked to check.

Issue #3325, opened 2024-04-23
Jobs remain in queued state when input validation fails


jmsmkn:
Celery errors are uncaught in the job creation process. Now there is a celery task with no task name, and just:

```
{"exc_type": "ChordError", "exc_message": ["ValidationError([\"{'__all__': [ValidationError(['Image segments could not be determined, ensure the voxel values are integers and that it contains no more than 64 segments. Ensure the image has the minimum and maximum voxel values set as tags if the image is a TIFF file.'])]}\"])"], "exc_module": "celery.exceptions"}
```

Issue #3139, opened 2023-12-15
Validation errors on job creation are not handled


amickan:
https://grand-challenge.sentry.io/issues/4667838707/?project=303639&query=%21logger%3Acsp+is%3Aunresolved&referrer=issue-stream&statsPeriod=14d&stream_index=0

jmsmkn:
We are seeing this a lot, important to fix.

amickan:
PR to fix this for archive items is about ready, but for algorithm jobs I was hoping to get the open PR from before Christmas merged first because it is contingent on it. I can also make a quick fix if necessary in the meantime, but it won't use the same code as display sets and archive items (yet).

jmsmkn:
OK, no need to do something separate, I'll review #3125 as soon as I can.

jmsmkn:
I have a big PR that touches on algorithm job creation right now, so best to hold off a few days on a fix for this.

amickan:
Good that say that. I was just going to start looking at this.

Issue #3368, opened 2024-06-06
Allow scheduling jobs with existing CIVs through API


amickan:
At the moment, when scheduling a job through the API, new CIVs are created for all inputs, even if an existing image is provided as input. The behavior when scheduling a job should match what we do on the "Try out algorithm" page, where we reuse CIVs when possible. 

This relates to https://github.com/comic/grand-challenge.org/issues/3139 - both require a heavy refactoring of civ creation and validation around algorithm jobs.

amickan:
I might give this a try this week, although it is a short week with DIAG day on Wednesday and the AWS workshop on Thursday, so might not get to making a PR.

jmsmkn:
Yes I will do it.

amickan:
Closing this in favour of: https://github.com/comic/grand-challenge.org/issues/3576 which is a more concise description of what is still missing for the API views.

PR #3465, opened 2024-08-02
Refactor job creation code

amickan:
This PR refactors the job creation code. Jobs, archive items and display sets now all use the same creation and validation logic. 

Errors from asynchronous input validation now get caught and are reported back to the user as error message on the Job object. The job is then cancelled and not executed. 

This PR also enables selecting existing images and files for jobs through the UI and enables re-using CIVs for existing values if possible. The same holds partially for the API view as well. It is not yet possible to create a job through the API with an existing non-image file. This was out of scope for this PR (https://github.com/DIAGNijmegen/rse-roadmap/issues/335#issuecomment-2304484574)

Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/335 

Closes https://github.com/comic/grand-challenge.org/issues/3139
Closes https://github.com/comic/grand-challenge.org/issues/3325
Closes https://github.com/comic/grand-challenge.org/issues/3565
Partially addresses https://github.com/comic/grand-challenge.org/issues/3368


app/grandchallenge/components/utils.py
```diff
@@ -0,0 +1,74 @@
+from django.core.exceptions import ObjectDoesNotExist
```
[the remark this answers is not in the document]
jmsmkn: I won't have time to review this today but this is the first thing that sticks out.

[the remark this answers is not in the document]
amickan: Thinking about where to place `reformat_inputs` - I could just move this to the JobPostSerializer (which is the only place this is currently needed), but what this is doing is reformatting the inputs, so  another solution would be to create a new ComponentInterfaceValue serializer that inherits from the `ComponentInterfaceValuePostSerializer` but overrides the `to-representation` method and directly outputs the inputs in the desired format. The JobPostSerializer would then use that serializer instead. That way we could more easily reuse this elsewhere if ever needed. What do you think?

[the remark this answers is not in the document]
jmsmkn: If it's not needed anywhere else then it's not worth abstracting, it would be premature. I always think it is best to:

- Use the code once somewhere
- Then, if you have a similar thing, implement it again, possibly duplicating code
- Then, when you have a third thing that needs it start on abstraction.
 
That way you have a good idea of the differing requirements between the three use cases and can more reliably abstract.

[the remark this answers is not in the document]
amickan: Ok, that makes sense. I will move it to the JobPostSerializer then.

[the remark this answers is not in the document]
jmsmkn: I think this is called the "use-use-reuse" model, can't find good references.

app/grandchallenge/evaluation/models.py
```diff
@@ -1420,6 +1424,70 @@ def container(self):
...
+
+        successful_jobs = (
+            Job.objects.filter(
+                algorithm_image=self.submission.algorithm_image,
+                status=Job.SUCCESS,
+                **extra_filter,
+            )
+            .annotate(
+                inputs_match_count=Count(
+                    "inputs",
```
amickan, line 1460: Just fyi: this is not new, I just moved this from `set_evaluation_inputs`

app/grandchallenge/components/views.py
```diff
@@ -453,3 +428,28 @@ def form_valid(self, form):
...
+                    name=self.interface.slug,
+                    value=None,
+                    attrs={
+                        "id": self.interface.slug,
+                        "help_text": clean(self.interface.description),
+                    },
+                )["widget"],
+            },
+        )
+        return HttpResponse(html_content)
```
amickan, line 455: For this new view, which only return a rendered file upload field (the actual upload is handled in the form that this field is added to), I'm unsure if we need to check permissions. This view is used in the archive item and display set forms and also in the job create form when someone chooses to upload a new noon-image file rather than selecting an existing one. Since this is only returning a form field, which is loaded on a page that is already locked down by permissions and since we check the permissions around the upload later on in the form, I think it is fine as is.

app/grandchallenge/algorithms/forms.py
```diff
@@ -93,16 +94,27 @@ class JobCreateForm(SaveFormInitMixin, Form):
     algorithm_model = ModelChoiceField(
         queryset=None, disabled=True, required=False, widget=HiddenInput
     )
+    creator = ModelChoiceField(
+        queryset=None, disabled=True, required=False, widget=HiddenInput
+    )
+    time_limit = IntegerField(
+        disabled=True, required=False, widget=HiddenInput
+    )
```
jmsmkn, line 102: Is there a test to ensure that the `time_limit` cannot be altered by the user? There is no other guard other than using `HiddenInput`.

app/grandchallenge/algorithms/models.py
```diff
@@ -629,6 +649,118 @@ def create(
...
+        extra_logs_viewer_groups=None,
+    ):
+        civ_data = {
+            k: v for k, v in data.items() if k not in self.non_interface_fields
+        }
+        non_civ_data = {
+            k: v for k, v in data.items() if k in self.non_interface_fields
+        }
+        job = Job.objects.create(
+            **non_civ_data,
```
jmsmkn, line 741: There could be a bug here. What if someone introduces a CI with the slug the same as one of the private field for the job? It is safer to be explicit about what is being accepted by doing a direct lookup of the fields that are accepted here.

app/grandchallenge/algorithms/models.py
```diff
@@ -629,6 +649,118 @@ def create(
...
+                )
+            )
+            .filter(inputs_match_count=input_interface_count)
+        )
+        return existing_jobs
+
+    def create_job_and_process_inputs(
+        self,
+        *,
+        data,
```
jmsmkn, line 730: Again data is a bit loose.

app/grandchallenge/algorithms/models.py
```diff
@@ -629,6 +649,118 @@ def create(
...
+        linked_task = execute_algorithm_job_for_inputs.signature(
+            kwargs={"job_pk": job.pk}, immutable=True
+        )
+
+        for key, value in civ_data.items():
+            job.create_civ(
+                ci_slug=key,
+                new_value=value,
+                user=data["creator"],
+                linked_task=linked_task,
```
jmsmkn, line 759: Can this task handle being called multiple times? It might be better to set this up as a group of tasks with one at the end.

app/grandchallenge/algorithms/models.py
```diff
@@ -770,6 +902,20 @@ def container(self):
...
+    def inputs_complete(self):
+        # check if all inputs are present and if they all have a value
+        input_interfaces = self.algorithm_image.algorithm.inputs.all()
+        existing_inputs = [
+            civ.interface for civ in self.inputs.all() if civ.has_value
+        ]
+        missing_inputs = [
+            interface
+            for interface in input_interfaces
+            if interface not in existing_inputs
```
jmsmkn, line 915: Might be clearer to have two sets of interfaces and then just check for equality. I think:

```python
return {civ.interface for civ in self.inputs.all() if civ.has_value} == {*self.algorithm_image.algorithm.inputs.all()}
```

app/grandchallenge/algorithms/serializers.py
```diff
@@ -219,63 +217,65 @@ def validate(self, data):
...
+            "user_upload",
+            "upload_session",
+        ]
+        data = {}
+        for civ in serialized_civs:
+            interface_slug = civ["interface"].slug
+            for key in possible_keys:
+                if key in civ:
+                    data[interface_slug] = civ[key]
+                    break
```
jmsmkn, line 280: The user should be informed with a 400 error should occur if there is no match, or multiple matches, right?

app/grandchallenge/algorithms/serializers.py
```diff
@@ -219,63 +217,65 @@ def validate(self, data):
...
-            time_limit=validated_data["algorithm_image"].algorithm.time_limit,
-        )
-        job.sort_inputs_and_execute(
-            upload_session_pks=upload_session_pks,
-            user_upload_pks=user_upload_pks,
         )
-
         return job
+
+    def reformat_inputs(self, *, serialized_civs):
```
jmsmkn, line 257: `self` doesn't appear to be used, likely is a static method.

app/grandchallenge/algorithms/serializers.py
```diff
@@ -174,16 +171,17 @@ def __init__(self, *args, **kwargs):
...
+        if not self._algorithm.active_image:
             raise serializers.ValidationError(
                 "Algorithm image is not ready to be used"
             )
         data["creator"] = user
-        data["algorithm_image"] = alg.active_image
-        data["algorithm_model"] = alg.active_model
+        data["algorithm_image"] = self._algorithm.active_image
+        data["algorithm_model"] = self._algorithm.active_model
+        data["time_limit"] = self._algorithm.time_limit
```
jmsmkn, line 184: Do we test to ensure that these are all fixed?

app/grandchallenge/algorithms/tasks.py
```diff
@@ -95,26 +36,19 @@ def execute_algorithm_job_for_inputs(*, job_pk):
...
-        job.update_status(
-            status=job.CANCELLED,
-            error_message=(
-                f"Job can't be started, input is missing for "
-                f"{oxford_comma([c.interface.title for c in missing_inputs])}"
-            ),
-        )
-        on_commit(linked_task.apply_async)
-    else:
+    if job.inputs_complete:
```
jmsmkn, line 39: Do we need to lock the job?

app/grandchallenge/algorithms/tasks.py
```diff
@@ -95,26 +36,19 @@ def execute_algorithm_job_for_inputs(*, job_pk):
...
         job.task_on_success = linked_task
         job.save()
         on_commit(
             execute_algorithm_job.signature(
                 kwargs={"job_pk": job_pk}, immutable=True
             ).apply_async
         )
+    else:
```
jmsmkn, line 47: What happens if the any of the preceding tasks fail, is the job left in a strange state?

app/grandchallenge/components/models.py
```diff
@@ -2087,57 +2094,95 @@ def assign_permissions(self):
...
+                getattr(self, self.base_object.civ_set_lookup).remove(
+                    current_civ
+                )
+                getattr(self, self.base_object.civ_set_lookup).add(civ)
             except ValidationError as e:
                 if new_value:
-                    raise e
+                    if hasattr(self, "error_message"):
+                        self.status = self.CANCELLED
+                        self.error_message = format_validation_error_message(e)
```
jmsmkn, line 2157: Are these attributes guaranteed to exist and be db fields? This is only a mixin, does it just happen to work for all users of the mixin? Why are you not using the update status method?

app/grandchallenge/components/tasks.py
```diff
@@ -603,19 +606,30 @@ def get_model_instance(*, app_label, model_name, **kwargs):
     return model.objects.get(**kwargs)
 
 
-@acks_late_2xlarge_task
+@acks_late_2xlarge_task(retry_on=(LockNotAcquiredException,))
+@transaction.atomic
```
jmsmkn, line 610: I think that the `raise` in L651 will cause the transaction to be cancelled and the job left in an undefined state.

app/grandchallenge/components/tasks.py
```diff
@@ -640,7 +654,8 @@ def provision_job(
         on_commit(execute_job.signature(**job.signature_kwargs).apply_async)
 
 
-@acks_late_micro_short_task(retry_on=(RetryStep,))
+@acks_late_micro_short_task(retry_on=(RetryStep, LockNotAcquiredException))
+@transaction.atomic
```
jmsmkn, line 658: Again, see my previous comments. Was this relevant to this PR?

app/grandchallenge/components/tasks.py
```diff
@@ -822,17 +844,28 @@ def handle_event(*, event, backend):  # noqa: C901
         )
 
 
-@acks_late_2xlarge_task
+@acks_late_2xlarge_task(retry_on=(LockNotAcquiredException,))
+@transaction.atomic
```
jmsmkn, line 848: Again, see my previous comments. Was this relevant to this PR?

app/grandchallenge/components/tasks.py
```diff
@@ -822,17 +844,28 @@ def handle_event(*, event, backend):  # noqa: C901
...
+    try:
+        job = job_queryset.get()
+    except OperationalError as error:
+        raise LockNotAcquiredException from error
+
     executor = job.get_executor(backend=backend)
 
     if job.status == job.EXECUTED and not job.outputs.exists():
         job.update_status(status=job.PARSING)
+    elif job.status in (job.SUCCESS, job.FAILURE, job.CANCELLED):
```
jmsmkn, line 866: Why was this check added, does the same thing as the existing else as far as I can tell. Again, don't think these tasks needed to be addressed.

app/grandchallenge/components/tasks.py
```diff
@@ -847,7 +880,7 @@ def parse_job_outputs(
             error_message=str(e),
             detailed_error_message=e.message_details,
         )
-        raise PriorStepFailed("Could not parse outputs")
+        return
```
jmsmkn, line 883: Why return here? Something has gone wrong.

app/grandchallenge/components/tasks.py
```diff
@@ -1064,38 +1051,68 @@ def validate_voxel_values(*, civ_pk):
...
+def update_uploadsession_and_object_status(
+    *, error_message, upload_session, object
+):
+    formatted_error = format_validation_error_message(error=error_message)
+    upload_session.status = RawImageUploadSession.FAILURE
+    upload_session.error_message = formatted_error
+    upload_session.save()
+    if hasattr(object, "error_message"):
+        object.status = object.CANCELLED
+        object.error_message = formatted_error
```
jmsmkn, line 1063: Are you sure that these attributes exist and are database fields? Why are you not using the update status method?

app/grandchallenge/components/tasks.py
```diff
@@ -1064,38 +1051,68 @@ def validate_voxel_values(*, civ_pk):
...
+    )
+    upload_queryset = RawImageUploadSession.objects.filter(
+        pk=upload_session_pk
+    ).select_for_update(nowait=True)
+
+    try:
+        object = object_queryset.get()
+        upload_session = upload_queryset.get()
+    except OperationalError as error:
+        raise LockNotAcquiredException from error
```
jmsmkn, line 1096: What is the guard to ensure that this task is only executed once?

app/grandchallenge/components/tasks.py
```diff
@@ -1123,46 +1141,65 @@ def add_file_to_object(
...
+    model = apps.get_model(app_label=app_label, model_name=model_name)
+    object_queryset = model.objects.filter(pk=object_pk).select_for_update(
+        nowait=True
     )
     interface = ComponentInterface.objects.get(pk=interface_pk)
+
+    try:
+        object = object_queryset.get()
+    except OperationalError as error:
+        raise LockNotAcquiredException from error
```
jmsmkn, line 1161: What is the guard to ensure that this task is only executed once?

app/grandchallenge/components/tasks.py
```diff
@@ -1123,46 +1141,65 @@ def add_file_to_object(
...
-                f"File for interface {interface.title} "
-                f"failed validation:\n{error}."
-            ),
+        error_description = (
+            f"File for interface {interface.title} failed validation:{error}."
         )
+        if hasattr(object, "error_message"):
+            object.error_message = error_description
+            object.status = object.CANCELLED
+            object.save()
```
jmsmkn, line 1190: See previous comments.

app/tests/algorithms_tests/test_forms.py
```diff
@@ -661,3 +665,77 @@ def test_model_version_control_form():
     assert "Model updating already in progress." in str(
         form.errors["algorithm_model"]
     )
+
+
+@pytest.mark.django_db
+class TestJobCreateForm(TestCase):
```
jmsmkn, line 671: Indeed, there shouldn't be a need to introduce new `TestCase` classes.

app/grandchallenge/components/tasks.py
amickan: The pitch mentions that all celery tasks involved in job creation should be reviewed and adjusted to fit our new design principles. I can undo the last commit I made and get rid of all those changes. Can indeed be done in a separate PR.

amickan: And you're right, that the transaction would prevent the intermediate status change from ever taking effect. I think that is ok, to be honest, but maybe better to leave as is for now and make the celery task changes separately.

app/grandchallenge/algorithms/forms.py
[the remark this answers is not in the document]
amickan: The field is `disabled`. If I understand the docs correctly (https://docs.djangoproject.com/en/5.1/ref/forms/fields/#django.forms.Field.disabled) that should be enough to ensure it won't be tampered with. I have added a test to confirm. 

Or would an explicit check in the clean method be better?

app/grandchallenge/algorithms/models.py
```diff
@@ -629,6 +649,118 @@ def create(
...
+        return ["algorithm_image", "algorithm_model", "creator", "time_limit"]
+
+    def retrieve_existing_civs(self, *, civ_data):
+        """
+        Checks if there are existing CIVs for the provided data and returns those.
+
+        Parameters
+        ----------
+        civ_data
+            A dictionary with interface slugs as keys and CIV values as values.
```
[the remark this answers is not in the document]
amickan: Thinking about how to improve this:
We already have a standardized way to represent CIV data through the ComponentInterfaceValuePostSerializer. I could rewrite this method and the other one below to use serialized CIV data. The `reformat_inputs` method on the serializer is then no longer needed. However, we would need the opposite of that method on the JobCreateForm to turn the cleaned form data into the serialized format. In any case, creating a new data class for it seems unnecessary. Would you agree?

app/grandchallenge/algorithms/models.py
```diff
@@ -629,6 +649,118 @@ def create(
 
         return obj
 
+    @property
+    def non_interface_fields(self):
+        return ["algorithm_image", "algorithm_model", "creator", "time_limit"]
```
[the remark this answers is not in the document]
amickan: Unsure. It would need to be added in two places, the `JobCreateForm` and the `JobPostSerializer`. So it seemed better to lift this up to the model level so as to only define it once, but I agree that this hides it from future developers. The only way to make sure it gets updated in the future would be to check on the model that the provided data doesn't contain any keys other than interface slugs and what is defined in `non_interface_fields`. Or do you see a different way?

app/grandchallenge/algorithms/models.py
amickan: Ok, good point. I will change this to use direct lookups for the non civ fields. And in order to tell the interface value from the private field value in the case you describe, the callers of this method here should provide `civ_data` and `non_civ_data` rather than splitting it here.

app/grandchallenge/algorithms/models.py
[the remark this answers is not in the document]
amickan: Forget what I suggested, we do need the civ data in the format as it is in a form's clean method ("interface_slug": "value"), or else more places need updating. So I will create a new data class.

app/grandchallenge/components/tasks.py
jmsmkn: Indeed, job creation, not job execution. Should have been clearer. I know that part of the codebase is tricky and problematic but we really shouldn't touch it given the demand in production right now.

app/grandchallenge/algorithms/models.py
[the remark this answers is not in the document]
jmsmkn: I'd say yes as they're properties of two independent forms. It may even be better to auto generate this somehow for each form so that when a new field is added it is updated. Or, at the very least cover it by tests.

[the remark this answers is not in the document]
amickan: I will add a test for them for now. Auto-generation would probably take me too much to implement right now.

app/grandchallenge/algorithms/models.py
```diff
@@ -629,6 +649,118 @@ def create(
...
+                        interface=ci, value=value
+                    ).get()
+                    existing_civs.append(civ)
+                except ObjectDoesNotExist:
+                    continue
+
+        return existing_civs
+
+    def get_jobs_with_same_inputs(
+        self, *, data, algorithm_image, algorithm_model
```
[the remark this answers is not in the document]
amickan: It should be called `inputs` or `input_civ_data` or something. It is now a list of CIVData objects. We don't enforce type hints, so just changing the variable name is enough here, right?

app/grandchallenge/algorithms/models.py
amickan: To prevent the issue you describe above, we actually need to prepend the interface fields in the form with something to clearly mark them as separate from the object fields. I will add an underscore.

app/grandchallenge/algorithms/models.py
amickan: I have removed this method. It's used in 2 places and half of what used to be done here is now done in the respective forms anyway, so no need for this.

app/grandchallenge/algorithms/serializers.py
amickan: That type of validation is already done by the ComponentInterfaceValuePostSerializer, so the data we get here is guaranteed to have one and only one of the possible keys.

app/tests/algorithms_tests/test_forms.py
amickan: Just for my understanding: why is TestCase not good to use?

app/grandchallenge/algorithms/serializers.py
amickan: I have added a test. These are not fields on the serializer, so they will be ignored in the user's input.

app/grandchallenge/components/models.py
amickan: Setting the status will only work for Jobs, not archive items and display sets. This is why I checked for the existing of the error_message attribute, but on second thought I think being explicit and checking if `isinstance(self, Job)` is better and I will also use `update_status`.

app/tests/algorithms_tests/test_forms.py
jmsmkn: It's from a test framework that we don't use. We use pytest instead.

app/grandchallenge/algorithms/models.py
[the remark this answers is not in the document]
jmsmkn: That sounds stronger.

app/grandchallenge/components/models.py
jmsmkn: Maybe using `hasattr(self, "update_status")` rather than checking for a particular class? Or add it to the base class and leave it as `pass` by default?

app/grandchallenge/algorithms/tasks.py
amickan: By fail, do you mean unexpectedly fail with an error that we don't catch in the task or System failure? If any of the previous tasks fail due to unexpected errors, the follow-up tasks won't be called and the job will indeed be left dangling. The solution to that would be to catch any exception in the task and update the object and re-raise the exception, right?

jmsmkn: Yes I think so.

amickan: There must a better way. If a task fails, you would want some task_on_failure to run that updates the job objects status, right? Looking at the celery docs, that should be possible using a chord with an error callback, but I think we've been having lots of issues with celery chords or any advanced canvas stuff, so probably not the way to go.

app/grandchallenge/algorithms/tasks.py
amickan: Yes, we can lock the job but we also need to retry the task on OperationalError then since this task will be called multiple times in the process of CIV creation.

app/grandchallenge/algorithms/models.py
amickan: Yes, I have implemented it in such a way that it is meant to be called multiple times. Down the line it ends up calling `provision_inputs`, where we check the job status, and that's where things will block if an earlier chain of events got there (because provision inputs has object locking in place and updates the status). Am I overseeing something? 

> group of tasks with one at the end.

I wonder how that would work. Not all CIV validation and creation happens in async tasks, so for each CIV you can have 0, 1 or 2 chained async tasks and for a job you want to run all of these possible chains in parallel with one final task after all others have completed successfully. We'd end up with a pretty complex canvas. I thought we wanted avoid those given how many issues we've had with Celery in the past?

app/grandchallenge/algorithms/models.py
```diff
@@ -629,6 +649,118 @@ def create(
...
+        else:
+            unique_kwargs["algorithm_model__isnull"] = True
+        existing_jobs = (
+            Job.objects.filter(**unique_kwargs)
+            .annotate(
+                inputs_match_count=Count(
+                    "inputs", filter=Q(inputs__in=existing_civs)
+                )
+            )
+            .filter(inputs_match_count=input_interface_count)
```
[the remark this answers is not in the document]
amickan: I have added a test for this (TestGetJobsWithSameInputs) where I tried to test all possible scenarios. So yes, I think this works. But would be good to have a second pair of eyes check if all scenarios are covered by the test.

app/grandchallenge/algorithms/tasks.py
jmsmkn: The celery stuff is unreliable. You first proposal is exactly how we solve it in other tasks.

app/grandchallenge/components/tasks.py
amickan: I don't think there is a way to check if it ran before (for archive items and display sets don't have a status we can check and there is no job status that would be appropriate to update here once this task has run). The only thing we can do is remove the `current_value`. That makes sure that we don't end up with duplicate CIVs on the object and that is ultimately what we want to prevent from happening.

app/grandchallenge/components/tasks.py
amickan: Same as above. All we can do, as far as I can see, is make sure that we don't end up with duplicate civs on the object and so we remove the `current_value` if it exists.

app/grandchallenge/algorithms/serializers.py
jmsmkn: You are relying on how things are organised today. Every method should be robust. You can future proof yourself by explicitly handling the unexpected cases. Maybe you don't need to send something to the user here but at the very least raise an exception when your assumption breaks down.

app/tests/algorithms_tests/test_api.py
```diff
@@ -104,3 +87,347 @@ def test_job_list_view_num_queries(
         # Sanity checks
         assert response.status_code == 200
         assert len(response.json()["results"]) == num_jobs
+
+
+class TestJobCreationThroughAPI(TestCase):
+
+    def setUp(self):
```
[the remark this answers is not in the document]
amickan: I've reworked all tests to not use `TestCase`

app/grandchallenge/components/views.py
amickan: I have added object permission checks. I check permissions for the base_object since this view is used both in create and update views, so the common denominator for object permission checking is the algorithm/reader_study/archive, not the Job/DisplaySet/ArchiveItem.

app/grandchallenge/components/tasks.py
amickan: I was working on catching exceptions in this task and the `add_image_to_object` task. The pattern we'd want is to catch all exceptions, then update the job object status (with an informative error or with something like "An unknown error occurred") an then reraise the exception (unless it is an expected exception that we know how to handle). This pattern however doesn't work if we want to wrap the entire task in a transaction. The job status would never get updated if we raise the exception. So then either, we are ok with not catching all exceptions, or we don't wrap these tasks in transactions. I'm unsure which of these is the way to go. I tend towards the first, only catch what we know how to handle, update the job accordingly, and let all other exceptions get raised without updating the job object and then we can handle them as they come in.

amickan: Celery has this feature where you specify a task to be executed on task failure (`link_error`), that could be a solution that would allow wrapping the tasks in a transaction but also mark the job as failed if an unexpected error occurs. I don't see it in use anywhere in the codebase though, so maybe another case of "unreliable celery feature"?

jmsmkn: If you don't want to raise then you do not need to. You should be able to do `logger.error("foo", exc_info=True)` then it would still be reported in the logs and to Sentry.

amickan:
I finished some more refactoring that I had started on Friday afternoon and that just needed a bit more testing. This PR is now also ready for another review. As per our cycle-work rules, I will now prioritize my pitch for this cycle.

jmsmkn:
@amickan could you solve the conflicting files?

amickan:
> @amickan could you solve the conflicting files?

Yes, I was just going to post here that I will solve the conflicts and take a fresh look at the PR now that my pitch is done.

app/grandchallenge/components/tasks.py
```diff
@@ -632,8 +635,16 @@ def provision_job(
...
+        logger.warning("Nothing to do: inputs are not complete.")
+        return
+
     if job.status not in [job.PENDING, job.RETRY]:
-        raise RuntimeError("Job is not ready for provisioning")
+        logger.error(
+            "Job is not ready for provisioning, or has already been provisioned",
+            exc_info=True,
+        )
+        return
```
jmsmkn, line 647: Why is this change? This should raise an error here - nothing has been altered in the task, so there is nothing to rollback. If it does need to change the `exc_info=True` is superfluous as there is no unhandled exception here.

app/grandchallenge/components/tasks.py
```diff
@@ -632,8 +635,16 @@ def provision_job(
     )
     executor = job.get_executor(backend=backend)
 
+    if not job.inputs_complete:
+        logger.warning("Nothing to do: inputs are not complete.")
```
jmsmkn, line 639: We expect this state often right? Nothing special is going wrong or needs to be highlighted:

```suggestion
        logger.info("Nothing to do: inputs are not complete.")
```

amickan: True. Info logging level makes more sense.

app/grandchallenge/components/tasks.py
amickan: Hm, I don't think we want to raise an error here. This task can be called multiple times with complete inputs. The first time it's called, the job is locked, inputs are provisioned and the state is updated. The second time it is called the state will no longer be pending (or retry), but we don't want to raise an error in that case. It is expected, we just want to exit and move on. So no need to raise an exception and the logger level should be info. Would you agree?

amickan: We could check `job.inputs_complete` again here and only raise an error if that is not the case. Then we'd catch the unexpected cases.

amickan: Ah no that makes no sense, since the code won't reach that point if the inputs are not complete.

amickan: I still think that we don't need to raise the exception here. Can you explain when that would be necessary?

jmsmkn: Because the task should not be called unless the job is in the correct state. If the job is not in the correct state and this task has been called it means that something has gone wrong in our application and we must know about it.

amickan: Since we don't want to use Celery chains and groups anymore, I think the only solution to guarantee that we only call `provision_job` once is to have 1 celery task that does all input creation and validation (rather than individual celery tasks per input). This task would need to run on the large queue. We could theoretically run into timeout issues for jobs with lots of very large image inputs, but I guess this would be very rare.

amickan: After discussing this synchronously in our meeting today, I did not move to using 1 single task for all inputs, but instead added a new status and the necessary checks in the task that runs before `provision_inputs` to ensure it runs only once.

amickan:
PR is ready for another review now. 

It somehow updated 2 css files with white spaces. Not sure how that happened and couldn't immediately find a way to revert that change, but that shouldn't stand in the way of the review. I will figure out how to remove the added white spaces on Wednesday.

jmsmkn:
There is a fix for biome changing the css on main 46fde956d1abb0353631e5f53c54ab976703a43b

You can just checkout the affected files with `git checkout main <path>`.

app/grandchallenge/components/form_fields.py
```diff
@@ -43,75 +48,130 @@ def __init__(
...
+        ]
+        extra_help = ""
+        return field_type(
+            help_text=_join_with_br(self.help_text, extra_help), **self.kwargs
+        )
+
+    def get_file_field(self):
+        key = f"value_type_{self.instance.slug}"
+        # on JobCreateForm interfaces are prepended with underscore
+        alt_key = f"value_type__{self.instance.slug}"
```
jmsmkn, line 139: We should be consistent, no?

jmsmkn (commented):
That was very difficult to review, far too large of a PR. 2 hours and I haven't looked at the tests.

app/grandchallenge/algorithms/models.py
```diff
@@ -414,6 +424,19 @@ def add_user(self, user):
...
+    def file_upload_field_url_path(self):
+        return "algorithms:file-upload"
+
+    @property
+    def civ_set_lookup(self):
+        return "inputs"
+
+    @property
+    def civ_sets_related_manager(self):
+        # should return all jobs for this algorithm (not just for the active image)
```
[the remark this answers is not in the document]
amickan: Because we use this to determine whether there are existing files for a user to choose from, and for that we should take into account files linked to all past jobs of an algorithm, not just jobs linked to the current active image.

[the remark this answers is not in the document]
jmsmkn: But then why limit it just to the one algorithm?

app/grandchallenge/reader_studies/views.py
```diff
@@ -1419,3 +1403,17 @@ def get_queryset(self, *args, **kwargs):
         return self.base_object.civ_sets_related_manager.filter(
             pk__in=[ds.pk for ds in qs if ds.is_editable]
         )
+
+
+class ReaderStudyFileUploadFieldView(FileUploadFormFieldBaseView):
+    permission_required = (
+        f"{ReaderStudy._meta.app_label}.change_{ReaderStudy._meta.model_name}"
```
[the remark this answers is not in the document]
amickan: Yes, because we want to be able to use this both on create and on update, and on create the display doesn't exist yet.

app/grandchallenge/components/views.py
```diff
@@ -449,3 +424,48 @@ def form_valid(self, form):
...
+            ComponentInterface, slug=self.kwargs["interface_slug"]
+        )
+
+    @property
+    def widget_name(self):
+        raise NotImplementedError
+
+    @cached_property
+    def base_object(self):
+        raise NotImplementedError
```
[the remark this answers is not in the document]
amickan: How would you implement that while still using `ObjectPermissionRequiredMixin`? We'd need to override `get_permission` and `get_permission_object` and have an awkward if statement in both of these to decide which permission to check for what, right? 

It seemed cleaner and much more clear to me to separate these out given that the permission and the object class to check the permission on is different in all these views.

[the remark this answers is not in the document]
amickan: I agree though that the `base_object` property can be removed, and we can have the subclasses directly provide `get_permission_object` instead.

app/grandchallenge/cases/tasks.py
```diff
@@ -114,44 +122,77 @@ def build_images(*, upload_session_pk):
...
+        if object:
+            object.handle_error(
+                notification_type=NotificationType.NotificationTypeChoices.IMAGE_IMPORT_STATUS,
+                upload_session=upload_session,
+                error_message="An unexpected error occurred",
+            )
+        else:
+            upload_session.error_message = "An unexpected error occurred"
+            upload_session.status = upload_session.FAILURE
+            upload_session.save()
```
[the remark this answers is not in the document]
amickan: I agree. I had started working on moving this to a method on the upload session, but all the ways I could think of at the time included even more changes in places that I thought I should better not touch to not explode this PR even more, so I stopped, but your suggestion is very straight-forward and should work. Don't know why that didn't occur to me on Monday.

app/grandchallenge/cases/tasks.py
```diff
@@ -114,44 +122,77 @@ def build_images(*, upload_session_pk):
...
-    with transaction.atomic():
-        upload_session = session_queryset.get()
-        upload_session.status = upload_session.STARTED
-        upload_session.save()
-
     try:
         with transaction.atomic():
-            # Acquire locks
-            upload_session = session_queryset.get()
+            if linked_object_pk:
```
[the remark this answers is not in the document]
amickan: That should indeed be possible if I remove the `raise` at the end of the last block and use `logger.error` there instead.

app/grandchallenge/algorithms/models.py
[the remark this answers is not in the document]
amickan: Good point. We can return all jobs the user has `view_job` permission for. Or should it be locked down to `change_job` permission. Difference would be that the latter would only allow re-using files they uploaded themselves (as creator of a job), where the former would allow a challenge participant to use files from a challenge phase archive if the phases is an open logs phase.  I think `view_job` permission is fine.

app/grandchallenge/algorithms/serializers.py
```diff
@@ -219,63 +223,88 @@ def validate(self, data):
...
         )
 
+        linked_task = execute_algorithm_job_for_inputs.signature(
+            kwargs={"job_pk": job.pk}, immutable=True
+        )
+
+        for civ in self.inputs:
+            job.create_civ(
+                ci_slug=civ.interface_slug,
+                new_value=civ.value,
```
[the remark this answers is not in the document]
amickan: I don't think so. `new_value` can only be large for interfaces of json kind that are not stored in the DB, and those are not handled asynchronously. For interfaces stored in the DB, `new_value` will be the pk of the corresponding `UserUpload` object.

app/grandchallenge/components/form_fields.py
amickan: Yes, I agree that we should be consistent, but at the time when I added the underscore my thought process was the following: adding an underscore to all interface fields everywhere would be consistent but touches a lot more of the codebase than is strictly necessary for this PR, and so it seemed better to only add the underscore where it is really necessary (i.e. on JobCreation where we have other fields that could name clash with interface names - this is not (yet) an issue for archive items or display sets since they only have interface fields).

amickan: Having said that, I do agree that only adding the underscore for job creation is weird and adds awkward work-around code snippets in a few places. Moreover, it does prevent future issues for archive items and display sets if we ever add fields to those forms where potential name clashes could happen. Maybe the way to go about this is to add the underscore everywhere in a separate PR.

app/grandchallenge/components/views.py
```diff
@@ -449,3 +424,48 @@ def form_valid(self, form):
...
+
+    def get_permission_object(self):
+        return self.base_object
```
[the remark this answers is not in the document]
amickan: It's used both in creating and updating. Maybe we need to have separate views for updating and creating, but then we'd end up with 6 different views, which feels a bit overkill for a view that anyway only returns a template (actual updating is handled in the update view where permission are checked on the display set / archive item).

app/grandchallenge/components/models.py
```diff
@@ -1451,6 +1455,7 @@ class ComponentJob(models.Model):
     EXECUTED = 9
     PARSING = 10
     EXECUTING_PREREQUISITES = 11
+    INPUTS_VALIDATED = 13
```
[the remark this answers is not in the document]
amickan: Victim of an earlier implementation where I had added 2 new status, and then removed one but forgot to update the number ;-)

app/grandchallenge/components/models.py
```diff
@@ -2014,6 +2024,22 @@ def update_size_in_storage(self):
...
+    interface_slug: str
+    value: (
+        str
+        | bool
+        | int
+        | float
+        | dict
+        | RawImageUploadSession
+        | UserUpload
+        | QuerySet
```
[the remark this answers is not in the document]
amickan: The `UserUploadMultipleWidget` (used by the FlexibleImageField) returns a queryset of UserUpload objects.

app/grandchallenge/components/models.py
```diff
@@ -2062,58 +2088,146 @@ def assign_permissions(self):
...
             )
         else:
             NotImplementedError(f"CIV creation for {ci} not handled.")
 
-    def create_civ_for_value(self, *, ci, current_civ, new_value):
+    def create_civ_for_value(self, *, ci, current_civ, new_value, linked_task):
         current_value = current_civ.value if current_civ else None
-        civ = ComponentInterfaceValue(interface=ci, value=new_value)
+
+        if new_value is not None:
```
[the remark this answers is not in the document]
amickan: We can have a CIV with value `None` if that passes validation. 

Not sure why I added this if-statement here  though. It seems unnecessary since creating a CIV without passing a value is the same as creating one with value=None. Will remove!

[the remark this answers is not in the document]
amickan: Ah, I did that to avoid saving a new CIV to the database when the new value was None, but doesn't matter that much, and it would anyway be reused, so it would at most mean creating 1 CIV with value=None per interface.

app/grandchallenge/components/form_fields.py
jmsmkn: Why not use a proper prefix rather than just an underscore?

amickan: What do you mean by proper prefix? How would that be different from just an underscore?

jmsmkn: I suggested `interface_form_field_prefix__` in another comment - then it is obvious to future developers what this is there for.

app/grandchallenge/components/models.py
```diff
@@ -2062,58 +2088,146 @@ def assign_permissions(self):
...
+                    self.handle_error(
+                        error_message=format_validation_error_message(e),
+                    )
                     raise e
-                else:
-                    self.values.remove(current_civ)
 
-    def create_civ_for_image(self, *, ci, current_civ, new_value, user):
+            self.remove_civ(current_civ)
+            if new_value:
```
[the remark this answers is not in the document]
amickan: Good point! I'm now checking `new_value` against `ci.default_field.emoty_values` (which covers the cases you mentioned.

app/grandchallenge/components/models.py
```diff
@@ -2156,14 +2278,43 @@ def create_civ_for_file(self, *, ci, current_civ, new_value):
...
+        raise NotImplementedError
+
+    def get_current_value_for_interface(
+        self, *, interface, user_upload=None, upload_session=None
+    ):
+        try:
+            return self.get_civ_for_interface(interface=interface)
+        except ObjectDoesNotExist:
+            return None
+        except MultipleObjectsReturned as e:
```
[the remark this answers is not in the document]
amickan: If `MultipleObjectsReturned` is raised, we already get a sentry error (I'm re-raising the error further below). `ObjectDoesNotExist` can technically happen for display sets and archive items because the user can delete those (I think we've had sentry errors related to that in the past). For jobs it should indeed not happen, I can add a check for the model  name here and raise the error if it happens for Jobs.

app/grandchallenge/algorithms/serializers.py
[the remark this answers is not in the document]
jmsmkn: If these are different things wouldn't it make more sense to pass them along as different variables?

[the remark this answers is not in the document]
amickan: I have refactored the civ creation code to be clearer earlier on. It made most sense to me to move the logic to the CIVData class with a custom create method. However, it's not fool-proof, since there is nothing that prevents a future developer from instantiating a CIVData instance directly rather than through that create method. If that happens the validation wouldn't run. Unsure how big of an issue that is. What do you think? Implementation is in the following commit: https://github.com/comic/grand-challenge.org/pull/3465/commits/d2bc25fd39252fd392b2b8498d688d8a1e29e09e

app/grandchallenge/components/form_fields.py
amickan: Ok, I'll use that. But do we want to stay with the current implementation and only add this prefix on the JobCreateForm, or should I go ahead and use this everywhere (with all the necessary code changes in other places)?

app/grandchallenge/components/models.py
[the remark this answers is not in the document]
amickan: This should be clearer now from the variable name with the latest changes I made.

app/grandchallenge/components/views.py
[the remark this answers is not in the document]
amickan: I still don't see how we would make do with just 1 view (or seeing your comment about not using `get_permission_object` on update views, 1 for updating and 1 for creating), for the reasons outlined above. Could you comment / explain how you would do this? Else I think the way forward would be 6 views: update & create for Jobs / ArchiveItems/ DisplaySets respectively.

app/grandchallenge/algorithms/serializers.py
```diff
@@ -157,6 +154,12 @@ class JobPostSerializer(JobSerializer):
     class Meta:
         model = Job
         fields = ["pk", "algorithm", "inputs", "status"]
+        non_interface_fields = [
+            "algorithm_image",
+            "algorithm_model",
+            "creator",
+            "time_limit",
+        ]
```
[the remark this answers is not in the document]
amickan: Ok, I will just give this a try and see how much of the code base this touches if I add this prefix to all interface form fields. It solves a few issues that you commented on in this PR, so it might be worth it if it's not too big a changeset.

app/grandchallenge/components/views.py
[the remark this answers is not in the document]
jmsmkn: Why would you still need `ObjectPermissionRequiredMixin`? This is not doing anything with the object right?

app/grandchallenge/algorithms/serializers.py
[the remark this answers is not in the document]
jmsmkn: I have no idea, sorry, I find what is implemented here far too complicated and do not understand it enough to be able to suggest how to make it robust.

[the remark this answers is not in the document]
jmsmkn: But think of basic design methods. Why have a separate create? Why not do the initialisation in __init__? If you don't want people to mess with variables make them protected and use properties for access. Why even use a NamedTuple there?

app/grandchallenge/components/views.py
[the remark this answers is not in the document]
jmsmkn: I really don't see the need for 6 different views. See the upload model, we have one endpoint for anyone who wants to upload stuff to the site, or the downloads model where we have one endpoint for anyone who wants to download stuff from the site.

app/grandchallenge/reader_studies/views.py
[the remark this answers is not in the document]
amickan: Removed in favour of just 1 view where we check if the user has change permissions for any archive, reader study or algorithm on the platform.

app/grandchallenge/components/views.py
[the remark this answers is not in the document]
amickan: Refactored as discussed.

app/grandchallenge/archives/views.py
```diff
@@ -746,3 +747,17 @@ class ArchiveItemBulkDelete(CIVSetBulkDelete):
     @property
     def base_object(self):
         return Archive.objects.get(slug=self.kwargs["slug"])
+
+
+class ArchiveFileUploadFieldView(FileUploadFormFieldBaseView):
+    permission_required = (
+        f"{Archive._meta.app_label}.change_{Archive._meta.model_name}"
```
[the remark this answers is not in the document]
amickan: NA

app/grandchallenge/components/views.py
```diff
@@ -449,3 +424,48 @@ def form_valid(self, form):
...
+
+    @cached_property
+    def interface(self):
+        return get_object_or_404(
+            ComponentInterface, slug=self.kwargs["interface_slug"]
+        )
+
+    @property
+    def widget_name(self):
+        raise NotImplementedError
```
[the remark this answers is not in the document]
amickan: It is now ;-)

app/grandchallenge/components/views.py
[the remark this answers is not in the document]
amickan: Refactored

app/grandchallenge/algorithms/models.py
```diff
@@ -414,6 +424,19 @@ def add_user(self, user):
     def remove_user(self, user):
         return user.groups.remove(self.users_group)
 
+    @property
+    def file_upload_field_url_path(self):
```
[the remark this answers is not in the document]
amickan: NA

app/grandchallenge/algorithms/tasks.py
```diff
@@ -14,122 +13,61 @@
...
-    if Job.objects.active().count() >= settings.ALGORITHMS_MAX_ACTIVE_JOBS:
+    job = Job.objects.get(pk=job_pk)
+    if not job.status == Job.INPUTS_VALIDATED:
+        raise RuntimeError("Job is not ready for execution")
+    elif Job.objects.active().count() >= settings.ALGORITHMS_MAX_ACTIVE_JOBS:
         raise TooManyJobsScheduled
     else:
-        job = Job.objects.get(pk=job_pk)
+        job.status = Job.PENDING
+        job.save()
```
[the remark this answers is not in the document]
amickan: Ok, I followed the pattern we use for evaluations, where we also re-use the PENDING state after executing algorithms is done, but I get your point that that state should be unique. What you suggest makes sense and means that a job would be in "validating inputs" state from creation onwards, unless we add yet another new state to distinguish pending from pending for execution. In other words, I see two ways of state progression after job creation:

1) VALIDATING_INPUTS -> PENDING -> PROVISIONING -> PROVISIONED etc. 
2) PENDING -> VALIDATING_INPUTS -> PENDING_FOR_EXECUTION -> PROVISIONING -> PROVISIONED etc. 

I think option 1 is good enough since validating inputs starts right away (i.e "create_civ" is called synchronously as soon as we create the job, so there is little point in having an initial "PENDING" state). I will proceed with implementing that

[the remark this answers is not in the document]
amickan: > Is raising a RuntimeError correct here? We're definitely only expecting this job to be called once?

This task should only be called once, when inputs are complete. We make sure it is only scheduled once by checking that the status is `job.VALIDATING_INPUTS` before changing it to `INPUTS_VALIDATED` and calling this function in `execute_algorithm_job_for_inputs.` A job can only go from validating inputs to inputs validated once and so this task will only be called once, or else we raise an error. Does that make sense? 

Manually requeuing a job in the admin will not call this method, but will directly call the execute method on the job.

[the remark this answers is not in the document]
amickan: Having looked at this again, I actually don't see the need for this separate `execute_algorithm_job` task, it can all be done in `execute_algorithm_job_for_inputs` I think. Then we also don't need the `INPUTS_VALIDATED` status, and can make do with just `VALIDATING_INPUTS` similar to how we handle evaluations. A job now goes from `VALIDATING_INPUTS` to `PENDING` to `PROVISIONING` etc.

app/grandchallenge/algorithms/views.py
```diff
@@ -503,109 +498,37 @@ def get_context_data(self, *args, **kwargs):
...
-            )
-            job.sort_inputs_and_execute(upload_session_pks=upload_session_pks)
+    def form_valid(self, form):
+        inputs = form.cleaned_data.pop("inputs")
+        self.object = Job.objects.create(
+            **form.cleaned_data,
+            extra_logs_viewer_groups=[self.algorithm.editors_group],
+        )
+        # local import to avoid circular dependency
+        from grandchallenge.algorithms.tasks import (
```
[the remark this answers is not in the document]
amickan: NA

app/grandchallenge/algorithms/forms.py
```diff
@@ -93,16 +95,35 @@ class JobCreateForm(SaveFormInitMixin, Form):
...
+
+    class Meta:
+        non_interface_fields = [
```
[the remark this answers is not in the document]
amickan: NA

app/grandchallenge/algorithms/serializers.py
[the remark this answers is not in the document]
amickan: Done. I'm using the prefix everywhere now and removed `non_interface_fields`.

app/grandchallenge/algorithms/models.py
```diff
@@ -881,24 +981,18 @@ def add_viewer(self, user):
...
-                kwargs={
-                    "job_pk": self.pk,
-                    "upload_session_pks": upload_session_pks,
-                    "user_upload_pks": user_upload_pks,
-                },
-                immutable=True,
-            ).apply_async
-        )
+    def remove_civ(self, civ):
+        return self.inputs.remove(civ)
```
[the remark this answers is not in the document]
amickan: True. I have added a check for that, should also be checked for display sets and archive items (i.e. if they're editable or not)

app/grandchallenge/algorithms/models.py
```diff
@@ -943,6 +1037,24 @@ def get_or_create_display_set(self, *, reader_study):
 
         return display_set
 
+    def handle_error(
```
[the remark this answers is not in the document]
amickan: I have refactored the error handling such that this method is no longer necessary.

app/grandchallenge/algorithms/views.py
```diff
@@ -1098,3 +1021,15 @@ def get_success_message(self, cleaned_data):
...
+class AlgorithmFileUploadFieldView(FileUploadFormFieldBaseView):
+    permission_required = "algorithms.execute_algorithm"
+
+    @cached_property
+    def base_object(self):
+        return get_object_or_404(Algorithm, slug=self.kwargs["slug"])
+
+    @property
+    def widget_name(self):
+        return f"_{self.interface.slug}"
```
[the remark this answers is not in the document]
amickan: NA

app/grandchallenge/algorithms/models.py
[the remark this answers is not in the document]
amickan: Actually, I'm not sure. Conceptually I agree that adding and removing should only be possible when the job is in `VALIDATING_INPUTS` state. When there are multiple inputs, however, and validation of one of them fails, the job will be in `CANCELLED` state when possibly subsequent inputs are validated. So then, raising an error in `add_civ` and `remove_civ` would mean that the these subsequent, possibly fully valid inputs would not get created, which to a user would suggest that they too failed validation when they did not. So I'm not 100% sold on adding a status check to these methods.

amickan:
This PR is ready for another round of review.

jmsmkn:
Sorry I've not reviewed this - been ill and not been able to concentrate on it properly so just doing easy work (MarkDown conversion).

amickan:
> Sorry I've not reviewed this - been ill and not been able to concentrate on it properly so just doing easy work (MarkDown conversion).

Get well soon, James!

pkcakeout:
Beterschap James!

app/grandchallenge/algorithms/models.py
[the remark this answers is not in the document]
jmsmkn: Agreed - it's fine for now.

app/grandchallenge/profiles/models.py
```diff
@@ -244,6 +245,41 @@ def dispatch_unread_direct_messages_email(
...
+    def file_civs_user_has_permission_to_use(self):
+        # local import to avoid circular dependency
+        from grandchallenge.components.models import (
+            ComponentInterfaceValue,
+            InterfaceKind,
+        )
+
+        job_query = get_objects_for_user(
+            self.user, "algorithms.view_job"
+        ).filter(inputs__pk__in=OuterRef("pk"))
```
jmsmkn, line 258: They should also be able to use job outputs too right? I think evaluation outputs is also missing. This is quite similar to the logic in `serve_component_interface_value`, maybe they should be unified. We have to be extremely careful here as things like say, evaluation inputs, should not be downloadable or viewable by people with the `view_evaluation` permission. Maybe that permission should be changed but that is a separate issue.

app/grandchallenge/profiles/models.py
```diff
@@ -244,6 +245,41 @@ def dispatch_unread_direct_messages_email(
             subscription_type=EmailSubscriptionTypes.NOTIFICATION,
         )
 
+    @cached_property
+    def file_civs_user_has_permission_to_use(self):
```
jmsmkn, line 249: Is the users profile the logical place for this? It's only used by `InterfaceFormField`, and only depends on attributes stored on those instances, so could be a `cached_property` there, right?

app/grandchallenge/cases/models.py
```diff
@@ -119,7 +121,44 @@ def save(self, *args, **kwargs):
...
+        if (
+            self.error_message
+            and linked_object
+            and hasattr(linked_object, "update_status")
+        ):
+            linked_object.update_status(
+                status=linked_object.CANCELLED, error_message=error_message
+            )
+
+    def process_images(
```
jmsmkn, line 155: Could be `kwargs` only here?

app/grandchallenge/components/forms.py
```diff
@@ -118,7 +113,7 @@ def __init__(self, *args, instance, base_obj, user, **kwargs):
 
         # add fields for all interfaces that already exist on
         # other display sets / archive items
-        for slug, values in base_obj.values_for_interfaces.items():
+        for slug, _ in base_obj.values_for_interfaces.items():
```
jmsmkn, line 116: Tip: if you're just discarding the values you can drop the `.items()` to just iterate over the keys.

app/grandchallenge/components/form_fields.py
```diff
@@ -26,6 +32,9 @@ def _join_with_br(a, b):
         return b
 
 
+INTERFACE_FORM_FIELD_PREFIX = "interface_field_"
```
jmsmkn, line 35: Would it also work if we threw in an underscore or two? Just a bit of extra safety:

```suggestion
INTERFACE_FORM_FIELD_PREFIX = "__INTERFACE_FIELD__"
```

app/grandchallenge/components/tasks.py
```diff
@@ -1249,17 +1209,35 @@ def add_image_to_object(
...
+        # and the job status was updated to CANCELLED
+        logger.error(e, exc_info=True)
+        return
 
-@acks_late_micro_short_task
+    if linked_task is not None:
+        on_commit(signature(linked_task).apply_async)
+
+
+@acks_late_micro_short_task(retry_on=(LockNotAcquiredException,))
```
jmsmkn, line 1240: I'm still a little bit concerned about competing locks for jobs with multiple file inputs for both `add_file_to_object` and `add_image_to_object`. Should we set `delayed_retry=False`?

jmsmkn (commented):
Getting much easier to follow now, just some small points.

app/grandchallenge/components/tasks.py
amickan: I had to read up on that argument first, but yes, I think you're right that setting `delayed_retry=False` is the right thing to do here. 

If I understood things correctly then this means that a task will be retried on the same queue (not the delayed queue) and will be immediately re-queued rather than with a delay of 5 minutes (as it would on the delayed queue). I was a bit worried that retrying immediately would be pointless if there is a blocking task, but by the time the re-queued task gets picked up, the blocking tasks will most likely have finished already anyway, and if not, it can be retried again, we allow plenty of retries.

app/grandchallenge/profiles/models.py
amickan: That is a fair point. I thought of it as a property of the user, but it is indeed only used on the InterfaceFormField, and I also don't foresee it being used anywhere else, so putting it there makes sense.

app/grandchallenge/components/tasks.py
amickan: Clear, thanks for explaining.

app/grandchallenge/profiles/models.py
amickan: Then again, this is only about file CIVs, and the standard evaluation output is a metrics file that is stored in the DB (so a value CIV). So that wouldn't be included here anyway.

amickan: > It would need to be tested with something roughly the size of production

So something like 2 million CIVs spready over algorithm inputs, outputs, display sets and archive items. I'll do some testing

jmsmkn: The users profile is for metadata about the user. This is about authorisation, which is a very different thing. It would be best to move this functionality to an independent function that takes the user. If the consumer is calling it many times it can cache it separately.

> A challenge admin can add outputs they don't want the participants to see.

We should be absolutely clear that we do not offer that functionality and we never have. We have row based permissions, there is no difference between any of the outputs, if someone can view the evaluation, they can view all of the outputs. This has always been the case. Now, you have the situation where someone could download the output, then re-upload it. We want to avoid that.

> Then again, this is only about file CIVs, and the standard evaluation output is a metrics file that is stored in the DB (so a value CIV).

We have many evaluations that output many things. The point here is about being consistent throughout the platform so that we do not have the same, but subtly different, business logic in different places in the codebase. We have to aim for consistency here, especially if we find bugs they need to be fixed in one place, and we shouldn't have to survey the entire codebase for these cases where we have multiple implementations of what should be the same thing. This is especially important for things as critical as authorisation as data security is our number 1 priority on the platform.

amickan: > It would be best to move this functionality to an independent function that takes the user.

Ok, makes sense!

> We have row based permissions, there is no difference between any of the outputs, if someone can view the evaluation, they can view all of the outputs.

Yes, I know that. What I was thinking of when I wrote is the filtering of the metrics in metrics.json that you implemented a while ago, but this is not on the level of permissions of course. 

> Now, you have the situation where someone could download the output, then re-upload it. We want to avoid that.

I'm not sure I understand this point. So then evaluation outputs should not be downloadable, right? 

> The point here is about being consistent throughout the platform so that we do not have the same, but subtly different, business logic in different places in the codebase.

I completely agree. This needs to be unified in one place.

jmsmkn: > I'm not sure I understand this point. So then evaluation outputs should not be downloadable, right?

Evaluation outputs should be downloadable. If you did not include them in the 2nd permission implementation then people could download then upload, we want to avoid that pattern. This is why you're implementing this right?

jmsmkn: Ah, you're correct. Sorry! They were removed in fa5ae89f9e40ae2c0bbb83e48c9e072e2b7d5a68. Lets not change it.

app/grandchallenge/serving/views.py
```diff
@@ -137,25 +139,10 @@ def serve_component_interface_value(
...
-                user=user,
-                perms=perm,
-            )
-            .filter(**{lookup: civ})
-            .exists()
-        ):
-            return protected_storage_redirect(
-                name=civ.file.name, creator=user, component_interface_value=civ
-            )
+    if civ in get_component_interface_values_for_user(user=user):
```
jmsmkn, line 142: This loads all of the objects into memory and does a filter client side, will be a lot for users with a lot of files. This will offload everything to the db:

```suggestion
    if get_component_interface_values_for_user(user=user).filter(pk=civ.pk).exists():
```

app/grandchallenge/serving/models.py
```diff
@@ -43,3 +45,36 @@ class Download(models.Model):
...
+        user=user, perms="archives.view_archiveitem"
+    ).filter(values__pk__in=OuterRef("pk"))
+
+    return (
+        ComponentInterfaceValue.objects.annotate(
+            has_view_job_inputs_perm=Exists(job_inputs_query)
+        )
+        .annotate(has_view_job_outputs_perm=Exists(job_outputs_query))
+        .annotate(has_view_ds_perm=Exists(display_set_query))
+        .annotate(has_view_ai_perm=Exists(archive_item_query))
```
jmsmkn, line 73: For my understanding: is there a reason for separate annotation clauses? Do we need to consider things like https://docs.djangoproject.com/en/5.1/topics/db/aggregation/#combining-multiple-aggregations ? We have to make sure this bit of code is rock solid. Are there comprehensive tests to cover the cases mentioned in the Django documentation?

app/grandchallenge/serving/views.py
amickan: Good point. Thanks!

app/grandchallenge/serving/models.py
jmsmkn: I have done some testing in production. The new method is just as performant as the old one, so that's great! It is still a little slow. If you put everything into one aggregation block rather than 4 it is consistently faster, so would be good to know the motivation for using separate ones.

I checked with 4 users: the anonymous user, a superuser, a user with many permissions who does not have access, a user who has access.

amickan: I did separate annotations because we once had issues with combining multiple annotations (see the link you posted). 
But I think we might be able to simplify it by doing:

```
ComponentInterfaceValue.objects.annotate(
            has_permissions=Exists(job_inputs_query)
                            | Exists(job_outputs_query)
                            | Exists(display_set_query)
                            | Exists(archive_item_query)
        )
        .filter(has_permissions=True)
```

Need to test this out to make sure this works fully though.

jmsmkn: That is cleaner, maybe be a bit more specific on the name of the annotation (has_view_permission?) performance wise there's no difference and returns the correct result with the 4 user test I did.

amickan: Thanks for checking! I have updated the method and also did some tests and can confirm that it returns the correct files. How slow is it in production, James?

jmsmkn: ~50 ms for normal users, 1.6s for the user with lots of perms which is not great but at least not a regression.

jmsmkn (approved):
Epic work, well done!

amickan:
> Epic work, well done!

It only took 219 comments... Thanks for the help and review!

jmsmkn:
Indeed, this is the 7th largest PR we've had, and the largest one that doesn't add/remove any vendored JS:

1. Vendor cdnjs files: 173767 lines changed
2. Consolidate management of vendored JS: 74266 lines changed
3. Vendor uppy: 41626 lines changed
4. Remove isolated viewers: 40311 lines changed
5. Replace jsdelivr and datatables CDN : 33030 lines changed
6. Add Cornerstone3D viewer for GPU enabled clientside viewing of volumes: 5116 lines changed
**7. Refactor job creation code: 5028 lines changed**
8. Remove the Retina API: 4291 lines changed
9. Workstation annotation style options: 4260 lines changed
10. Adds global-bundle.pem: 3054 lines changed
11. Remove unused apps: 2871 lines changed
12. Remove products app: 2837 lines changed
13. Add Direct Messages: 2607 lines changed
14. Deprecate images: 2445 lines changed
15. Introduce TEXT-type answers: 2095 lines changed
16. Update dependencies: 2082 lines changed
17. Add ellipse drawing: 1929 lines changed
18. Add prettier to unify formatting of js files: 1903 lines changed
19. External evaluation phases: 1890 lines changed
20. Add 3 point angle annotation type: 1888 lines changed
21. Add SageMaker Training Backend: 1861 lines changed
22. Groundtruth seperate from method container: 1859 lines changed
23. Remove multi and single line text question types: 1848 lines changed
24. Add Amazon SageMaker Batch backend: 1819 lines changed
25. Remove inline JS: 1778 lines changed

PR #3462, opened 2024-07-29
Add visibility indicator to phase/submission/leaderboard for Admins

ammar257ammar:
Adding indicators for visibility status for challenge admins to make it easier to see which phase, phase submissions, and phase leaderboards are public/not public.

Screenshot 2024-08-08 at 11 54 19

Screenshot 2024-08-08 at 11 54 27

Screenshot 2024-08-08 at 11 54 35

closes #3020

PR #3475, opened 2024-08-06
Add examples for component interface values

### LINKED ISSUE #2919: Schema of json files from algorithm upload page is unclear
As an algorithm creator, what should these json files look like?

Screenshot 2023-07-06 at 10 47 00

ammar257ammar:
Example values can be added to interfaces of type json. They are validated against any schema and are displayed in the component list.

Screenshot 2024-08-28 at 15 26 33

Closes #2919


ammar257ammar (commented):
This test is only for invalid values of schema examples where it tests for raised errors on validation. Having both valid and invalid cases in teh same test would require using an "if statement" which I would prefer to avoid. I would prefer having another independent test for valid values. What are your thoughts on this?

jmsmkn:
> The model test is only for invalid values of schema examples where it tests for raised errors on clean() where validation happens. Having both valid and invalid cases in the same test would require using an "if statement" which I prefer to avoid. I would prefer having another independent test for valid values. What are your thoughts on this?

You can parameterise the test and pass a `nullcontext()` rather than `pytest.raises(ValidationError)`. See other examples in the tests.

app/grandchallenge/components/models.py
```diff
@@ -881,6 +881,13 @@ class ComponentInterface(OverlaySegmentsMixin):
         ),
         validators=[JSONSchemaValidator()],
     )
+    example_value = models.JSONField(
+        blank=True,
+        null=True,
+        default=None,
+        help_text="Example JSON that complies with the additional JSON schema",
+        validators=[JSONSchemaValidator()],
```
jmsmkn, line 889: Why are you validating that this is a schema? This is just a JSON value.

app/grandchallenge/components/models.py
```diff
@@ -1143,6 +1151,15 @@ def _clean_store_in_database(self):
                 f"Interface {self.kind} objects cannot be stored in the database"
             )
 
+    def _validate_example_value(self):
```
jmsmkn, line 1154: Please be consistent with the rest of the codebase. Method name should be `_clean_example_value`.

app/grandchallenge/components/models.py
```diff
@@ -1143,6 +1151,15 @@ def _clean_store_in_database(self):
...
+        if self.example_value:
+            if self.schema:
```
jmsmkn, line 1156: We shouldn't only validate against the schema. The same validation chain that all values would need to go through needs to be followed. [I stated this here](https://github.com/comic/grand-challenge.org/issues/2919#issuecomment-2263508206).

app/grandchallenge/components/models.py
```diff
@@ -1143,6 +1151,15 @@ def _clean_store_in_database(self):
...
+                JSONValidator(schema=self.schema)(value=self.example_value)
+            else:
+                raise ValidationError(
+                    "Schema is not provided to validate the example_value against"
```
jmsmkn, line 1160: Setting a schema is not necessary for value validation.

app/grandchallenge/components/templates/components/componentinterface_list.html
```diff
@@ -67,6 +68,10 @@ <h2>{{ object_type|title }} {{ list_type|title }} Options</h2>
                                 <code>/input/{{ interface.relative_path }}{% if interface.relative_path %}/{% endif %}&lt;uuid&gt;.tif</code>
                             {% else %}
                                 <code>/input/{{ interface.relative_path }}</code>
+                                {% if interface.example_value %}
+                                    <br>
+                                    <a download="json_schema_example.json" href="data:application/octet-stream;charset=utf-8;base64,{{ interface.example_value | b64encode_json }}">Download schema example</a>
+                                {% endif %}
```
jmsmkn, line 74: This is putting extra data in the same column of a table. It has nothing to do with _where_ the user needs to write the output. It is about _what_ should be written there. So, put this in a new column.

app/grandchallenge/components/templates/components/componentinterface_list.html
```diff
@@ -77,6 +82,10 @@ <h2>{{ object_type|title }} {{ list_type|title }} Options</h2>
                                 <code>/output/{{ interface.relative_path }}{% if interface.relative_path %}/{% endif %}&lt;uuid&gt;.tif</code>
                             {% else %}
                                 <code>/output/{{ interface.relative_path }}</code>
+                                {% if interface.example_value %}
+                                    <br>
+                                    <a download="json_schema_example.json" href="data:application/octet-stream;charset=utf-8;base64,{{ interface.example_value | b64encode_json }}">Download schema example</a>
+                                {% endif %}
```
jmsmkn, line 88: The example value doesn't need to vary with input and output, so this code block should not be duplicated.

app/grandchallenge/components/templates/components/componentinterface_list.html
```diff
@@ -67,6 +68,10 @@ <h2>{{ object_type|title }} {{ list_type|title }} Options</h2>
                                 <code>/input/{{ interface.relative_path }}{% if interface.relative_path %}/{% endif %}&lt;uuid&gt;.tif</code>
                             {% else %}
                                 <code>/input/{{ interface.relative_path }}</code>
+                                {% if interface.example_value %}
+                                    <br>
+                                    <a download="json_schema_example.json" href="data:application/octet-stream;charset=utf-8;base64,{{ interface.example_value | b64encode_json }}">Download schema example</a>
```
jmsmkn, line 73: Downloading a file to see the content is awkward. It is better to display the value. If the value does not fit into the table then put it in a modal.

app/grandchallenge/components/templatetags/base64.py
```diff
@@ -0,0 +1,11 @@
+import json
+from base64 import b64encode
+
+from django import template
+
+register = template.Library()
+
+
+@register.filter
+def b64encode_json(value):
```
jmsmkn, line 10: With the data put into the table or a model this filter is unnecessary.

app/grandchallenge/components/models.py
ammar257ammar: Please correct me if I am worng. We want to valiadte the example_value against the custom schema before saving the model and not just checking that the example_value is a valid JSON, right? This is what this call is doing right now, and in the GC Admin you see a response like this:

Screenshot 2024-08-14 at 14 29 49

app/grandchallenge/components/models.py
ammar257ammar: Done

app/grandchallenge/components/models.py
ammar257ammar: Fixed, my mistake.

app/grandchallenge/components/models.py
ammar257ammar: Please bear with me. I saw you statement but I am not sure I understand what to call straightforward from another model `InterfaceValue` to this one.

Screenshot 2024-08-14 at 14 33 56

The `_validate_example_value` is called within `clean()` which calls `super().clean()`. None of the other function within `clean()` make external calls to `InterfaceValue`.

I checked ChatGPT and it gave me the following example:

Did you mean by calling the validation chain, calling `full_clean()` on model `save()` ?

```
from django.db import models
from django.core.exceptions import ValidationError
from django.utils.translation import gettext_lazy as _

# Step 1: Define a custom validator
def validate_even(value):
    if value % 2 != 0:
        raise ValidationError(
            _('%(value)s is not an even number'),
            params={'value': value},
        )

class Product(models.Model):
    name = models.CharField(max_length=100)
    price = models.DecimalField(max_digits=10, decimal_places=2)
    quantity = models.IntegerField(validators=[validate_even])  # Custom validator added here

    # Step 2: Add clean method for field-level validation
    def clean(self):
        # Example: Ensure price is not negative
        if self.price < 0:
            raise ValidationError({'price': _('Price cannot be negative.')})

    # Step 3: Add clean method for model-level validation
    def clean(self):
        super().clean()
        if self.price < 0:
            raise ValidationError({'price': _('Price cannot be negative.')})

        if self.quantity < 0:
            raise ValidationError({'quantity': _('Quantity cannot be negative.')})

        # Example: Ensure the product has either a name or a price
        if not self.name and self.price <= 0:
            raise ValidationError(_('Product must have either a name or a price greater than zero.'))

    # Step 4: Override save method to trigger validation
        self.full_clean()  # Runs all validations including custom and model-level validation
        super(Product, self).save(*args, **kwargs)

    def __str__(self):
        return self.name
```

app/grandchallenge/components/templates/components/componentinterface_list.html
ammar257ammar: Done

app/grandchallenge/components/templates/components/componentinterface_list.html
ammar257ammar: I rewrote the part of the view eliminating duplicate code

app/grandchallenge/components/templates/components/componentinterface_list.html
ammar257ammar: I made it view in modal instead

app/grandchallenge/components/templatetags/base64.py
ammar257ammar: Deleted

app/grandchallenge/components/models.py
ammar257ammar: @jmsmkn I know you are busy, but if you had time to comment on this before your day off, that would be great. Then I can finish this issue this week.
Thanks.

jmsmkn: When you create a CIV you are validating the value of the CIV against the schema of the interface, plus all of the base validation of that interface. So, the value of the example value should also be a valid value for the CIV given the interface.

app/grandchallenge/components/models.py
jmsmkn: Yes. But setting the schema is not required.

app/grandchallenge/components/templates/components/componentinterface_list.html
jmsmkn: The title of that column and the objects are still not correct. You are confusing concepts here. This is an example value. It still has nothing to do with schemas.

ammar257ammar: Got it. A quick discussion with Chris helped a lot clarifying the confusion. I fixed it.

app/grandchallenge/components/migrations/0022_alter_componentinterface_example_value.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 4.2.14 on 2024-08-14 08:06
```
jmsmkn, line 1: PRs should only introduce one migration per app. Delete both of these and regenrate them.

app/grandchallenge/components/models.py
```diff
@@ -1143,6 +1150,14 @@ def _clean_store_in_database(self):
                 f"Interface {self.kind} objects cannot be stored in the database"
             )
 
+    def _clean_example_value(self):
+
+        if self.example_value:
+            civ = ComponentInterfaceValue(
+                interface=self, value=self.example_value
+            )
+            civ.clean()
```
jmsmkn, line 1159: Why only `clean` and not `full_clean`? You're skipping a bunch of validations here.

app/grandchallenge/components/templates/components/partials/ci_example_value.html
```diff
@@ -0,0 +1,36 @@
+{% load json%}
```
jmsmkn, line 1: Please take care with the work you ask for review, I have had to correct this many times

```suggestion
{% load json %}
```

app/grandchallenge/components/templates/components/partials/ci_example_value.html
```diff
@@ -0,0 +1,36 @@
...
+    data-toggle="modal"
+    data-target="#ciModal-{{ interface.pk }}"
+    class="text-decoration-none"
+    role="button"
+    title="View {{ interface.title }}"
+>
+    <div class="badge badge-primary">
+        <i
+                class="fas fa-fw fa-eye"
+        ></i>
```
jmsmkn, line 13: What is all this whitespace?

```suggestion
        <i class="fas fa-fw fa-eye"></i>
```

app/grandchallenge/components/templates/components/partials/ci_example_value.html
```diff
@@ -0,0 +1,36 @@
+{% load json%}
+
+<a href="#ciModal-{{ interface.pk }}"
```
jmsmkn, line 3: Why mix all the cases? Be consistent!! `example-value-modal-{{ interface.pk }}`

app/tests/components_tests/test_models.py
```diff
@@ -1466,3 +1466,133 @@ def test_displacement_field_validation(
...
+    ci = ComponentInterfaceFactory(
+        kind=InterfaceKindChoices.TWO_D_BOUNDING_BOX,
+        schema=sc,
+        relative_path="example_value.json",
+        example_value=example_value,
+    )
+
+    with context as ctx:
+        ci.clean()
+        assert getattr(ctx, "value", None) == expected_value
```
jmsmkn, line 1598: Putting your assert here doesn't test anything! `ci.clean()` will raise an exception.

app/tests/components_tests/test_models.py
```diff
@@ -1466,3 +1466,133 @@ def test_displacement_field_validation(
...
+                ],
+            },
+            "probability": {"type": "number"},
+        },
+        "required": ["version", "type", "name", "corners", "probability"],
+    }
+
+    ci = ComponentInterfaceFactory(
+        kind=InterfaceKindChoices.TWO_D_BOUNDING_BOX,
+        schema=sc,
```
jmsmkn, line 1591: As far as I can tell this test doesn't test your functionality at all. You've set the kind to `InterfaceKindChoices.TWO_D_BOUNDING_BOX` then replicating that schema?! The schema should add extra restrictions on top of the underlying schema. In this case it should be for extra restrictions on top of 2d bounding boxes.

app/tests/components_tests/test_views.py
```diff
@@ -442,3 +444,51 @@ def test_display_set_bulk_delete(
...
+                "type": "string",
+                "description": "The person's last name.",
+            },
+            "age": {
+                "description": "Age in years which must be equal to or greater than zero.",
+                "type": "integer",
+                "minimum": 1,
+            },
+        },
+    }
```
jmsmkn, line 472: Why bother with a schema here? It is not used by your test at all.

app/grandchallenge/components/templates/components/componentinterface_list.html
```diff
@@ -59,27 +60,23 @@ <h2>{{ object_type|title }} {{ list_type|title }} Options</h2>
...
-                                <code>/input/{{ interface.relative_path }}{% if interface.relative_path %}/{% endif %}&lt;uuid&gt;.mha</code>
-                                or
-                                <code>/input/{{ interface.relative_path }}{% if interface.relative_path %}/{% endif %}&lt;uuid&gt;.tif</code>
-                            {% else %}
-                                <code>/input/{{ interface.relative_path }}</code>
-                            {% endif %}
-                        </td>
-                    {% elif list_type == list_type_options.OUTPUT %}
+
+                    {% with list_type|lower as url_segment %}
```
jmsmkn, line 64: Undo this refactoring. At this stage, only touch code that is relevant to your functionality.

app/grandchallenge/components/templates/components/componentinterface_list.html
```diff
@@ -59,27 +60,23 @@ <h2>{{ object_type|title }} {{ list_type|title }} Options</h2>
...
                             {% else %}
-                                <code>/output/{{ interface.relative_path }}</code>
+                                <code>/{{ url_segment }}/{{ interface.relative_path }}</code>
                             {% endif %}
                         </td>
-                    {% endif %}
+                    {% endwith %}
+                    <td>
+                        {% if interface.example_value %}
+                            {% include "components/partials/ci_example_value.html" %}
```
jmsmkn, line 77: Why make this a partial? It is not being reused anywhere.

app/grandchallenge/components/templates/components/partials/ci_example_value.html
```diff
@@ -0,0 +1,36 @@
...
+>
+    <div class="badge badge-primary">
+        <i
+                class="fas fa-fw fa-eye"
+        ></i>
+        View Example
+    </div>
+</a>
+
+<div class="modal" id="ciModal-{{ interface.pk }}" tabindex="-1" role="dialog"
```
jmsmkn, line 18: Why are you creating one model per interface? One per page is plenty.

app/grandchallenge/components/models.py
```diff
@@ -1143,6 +1150,14 @@ def _clean_store_in_database(self):
                 f"Interface {self.kind} objects cannot be stored in the database"
             )
 
+    def _clean_example_value(self):
+
+        if self.example_value:
+            civ = ComponentInterfaceValue(
+                interface=self, value=self.example_value
```
jmsmkn, line 1157: I don't think that this works with types that are not stored in the database.

app/grandchallenge/components/models.py
```diff
@@ -1143,6 +1150,14 @@ def _clean_store_in_database(self):
                 f"Interface {self.kind} objects cannot be stored in the database"
             )
 
+    def _clean_example_value(self):
```
jmsmkn, line 1153: Example values should only be set when the kind is `is_json_kind`. Tests should be added for that.

app/grandchallenge/components/templates/components/partials/ci_example_value.html
ammar257ammar: This is strange. It happened a couple of times before. I did not write it like this. I suspect something wrong is happening on pre-commit. Is there a specific config should be set there? I will check again the changes after pre-commit before pushing.

jmsmkn: Pre commit would never do this.

ammar257ammar: oh, sorry. I just realized the original template `modal_base.html` where I got the code already has it like this. I thought you already approved the available code and that is why I try to use as much as possible. I will check more carefully next time.

app/grandchallenge/components/templates/components/partials/ci_example_value.html
```diff
@@ -0,0 +1,36 @@
...
+                <h5 class="modal-title">
+                    {{ interface.title }} example
+                </h5>
+                <button type="button" class="close" data-dismiss="modal"
+                        aria-label="Close">
+                    <span aria-hidden="true">&times;</span>
+                </button>
+            </div>
+            <div class="text-left modal-body">
+                <pre>{{ interface.example_value|json_dumps:4 }}</pre>
```
jmsmkn, line 32: If the `example_value` is `None`, and the `kind` is one of the JSON kinds then the default example for that interface should be provided. For now, duplicate the examples from the documentation in code.

app/tests/components_tests/test_models.py
jmsmkn: This test would be much clearer to test this with a Number kind and then limit that number with a custom schema to greater than zero.

app/grandchallenge/components/templates/components/partials/ci_example_value.html
jmsmkn: I am having to go through your code much more closely as what you give me to review has a lot of problems including breaking existing functionality that I had to revert (ab18aed), work where you didn't address PR feedback and I have to step in to complete it (#3462), and work that I had to re-do due to an overly-complex implementation (#3464). Please critically read the code you request review for before asking me.

app/grandchallenge/components/migrations/0022_alter_componentinterface_example_value.py
ammar257ammar: Done

app/grandchallenge/components/models.py
ammar257ammar: Done

app/grandchallenge/components/templates/components/partials/ci_example_value.html
ammar257ammar: Done

app/grandchallenge/components/templates/components/partials/ci_example_value.html
ammar257ammar: Done

app/grandchallenge/components/models.py
ammar257ammar: Do you mean it should raise a validation error if the example was provided with non json kind interfaces?
If so, then no need for extra tests for it, correct?

jmsmkn: Yes. Why do you think that it wouldn't need a test?

app/grandchallenge/components/templates/components/partials/ci_example_value.html
ammar257ammar: I see 20 examples in the code documentation for JSON kinds. Do you mean here duplicating them with an if statement in the template?

jmsmkn: It should not be in the template, do it in code. The template should look up a property on the model.

app/grandchallenge/components/models.py
ammar257ammar: ah, I misintepreted it, my bad. Of course, it needs a test that fails on interfaces with non-json kinds when having an example value (and tests for correct cases).

app/tests/components_tests/test_models.py
ammar257ammar: Done

app/tests/components_tests/test_models.py
ammar257ammar: Done

app/tests/components_tests/test_views.py
ammar257ammar: Done

app/grandchallenge/components/templates/components/componentinterface_list.html
ammar257ammar: Done

app/grandchallenge/components/templates/components/componentinterface_list.html
ammar257ammar: Done

app/grandchallenge/components/models.py
ammar257ammar: Checked and updated

app/grandchallenge/components/models.py
ammar257ammar: Done

app/grandchallenge/components/templates/components/partials/ci_example_value.html
ammar257ammar: Done

ammar257ammar:
This PR is ready for review. It was in draft status until yesterday, but I requested a review before that. So, I am not sure if you get notified if the PR is a draft.

jmsmkn:
> This PR is ready for review. It was in draft status until yesterday, but I requested a review before that. So, I am not sure if you get notified if the PR is a draft.

I know. This is a large PR and has lower priority than cycle, management or operations work.

jmsmkn:
There is a bug in this which means that the example value cannot be created at the same time as the interface. Needs to be split up.

jmsmkn:
Also needed to undo changes related to a misconception about optionally storing in the database vs requiring it.

jmsmkn (approved):

PR #3480, opened 2024-08-07
Fix image responsiveness in blog posts

### LINKED ISSUE #3095: Blog post can overflow into sidebar
Blog post can overflow into the sidebar:

image

Easiest fix: `overflow: hidden`

ammar257ammar:
Make sure that HTML images in a blog post Markdown are responsive by adding  the class "img-fluid" to their tags.

The current problem occurs when there is an html img tag within the Markdow syntax. The reason is that "markdown" function extracts the HTML tags prior to processing the Markdown and then inject it back after processing. The explicit HTML does not go through the TreeProcessors the same way Markdown do, and thus, no changes to the class occur. 
The solution was to add a function that parse the extracted HTML blocks and check for raw html tags to add the desired class if missing.

closes #3095

### DIFF
--- a/app/grandchallenge/core/utils/markdown.py
+++ b/app/grandchallenge/core/utils/markdown.py
@@ -54,7 +54,7 @@ class HtmlTagsExtension(Extension):
 
 
 class HtmlTagsPostprocessor(Postprocessor):
-    def run(self, lines):
+    def run(self, text):
         for i in range(len(self.md.htmlStash.rawHtmlBlocks)):
             bs4block = BeautifulSoup(
                 self.md.htmlStash.rawHtmlBlocks[i], "html.parser"
@@ -68,4 +68,4 @@ class HtmlTagsPostprocessor(Postprocessor):
                     img["class"].append("img-fluid")
 
                 self.md.htmlStash.rawHtmlBlocks[i] = str(bs4block)
-        return lines
+        return text
--- a/app/grandchallenge/core/utils/markdown.py
+++ b/app/grandchallenge/core/utils/markdown.py
@@ -38,7 +38,6 @@ class BS4Treeprocessor(Treeprocessor):
 
     @staticmethod
     def set_css_class(*, element, class_name):
-
         if isinstance(element, ElementTree.Element):
             current_class = element.attrib.get("class", "")
 
@@ -55,6 +54,7 @@ class BS4Treeprocessor(Treeprocessor):
             for name in class_name.split(" "):
                 if class_name not in current_class:
                     current_class.append(name)
+
         else:
             raise TypeError("Unsupported element")
 
--- a/app/grandchallenge/core/utils/markdown.py
+++ b/app/grandchallenge/core/utils/markdown.py
@@ -59,13 +59,13 @@ class HtmlTagsPostprocessor(Postprocessor):
             bs4block = BeautifulSoup(
                 self.md.htmlStash.rawHtmlBlocks[i], "html.parser"
             )
-
-            for img in bs4block.find_all("img"):
+            img = bs4block.find("img")
+            if img:
                 if "class" not in img.attrs:
                     img.attrs["class"] = []
 
                 if "img-fluid" not in img.attrs["class"]:
                     img["class"].append("img-fluid")
 
-            self.md.htmlStash.rawHtmlBlocks[i] = str(bs4block)
+                self.md.htmlStash.rawHtmlBlocks[i] = str(bs4block)
         return lines
--- a/app/grandchallenge/core/utils/markdown.py
+++ b/app/grandchallenge/core/utils/markdown.py
@@ -1,6 +1,6 @@
 from xml.etree import ElementTree
 
-from bs4 import BeautifulSoup
+from bs4 import BeautifulSoup, Tag
 from markdown import Extension
 from markdown.treeprocessors import Treeprocessor
 
@@ -26,22 +26,19 @@ class BS4Treeprocessor(Treeprocessor):
         for el in root.iter():
 
             if el.tag in el_class_dict:
-                BS4Treeprocessor.set_css_class(el, el_class_dict[el.tag])
+                self.set_css_class(el, el_class_dict[el.tag])
 
-        for i in range(len(self.md.htmlStash.rawHtmlBlocks)):
+        for i, html_block in enumerate(self.md.htmlStash.rawHtmlBlocks):
 
-            bs4block = BeautifulSoup(
-                self.md.htmlStash.rawHtmlBlocks[i], "html.parser"
-            )
+            bs4block = BeautifulSoup(html_block, "html.parser")
 
-            el = bs4block.find()
-
-            if el and el.name in el_class_dict:
-                BS4Treeprocessor.set_css_class(el, el_class_dict[el.name])
-                self.md.htmlStash.rawHtmlBlocks[i] = str(bs4block)
+            for tag, tag_class in el_class_dict.items():
+                for el in bs4block.find_all(tag):
+                    self.set_css_class(el, tag_class)
+                    self.md.htmlStash.rawHtmlBlocks[i] = str(bs4block)
 
     @staticmethod
-    def set_css_class(element, class_name: str):
+    def set_css_class(element, class_name):
 
         if isinstance(element, ElementTree.Element):
 
@@ -51,7 +48,7 @@ class BS4Treeprocessor(Treeprocessor):
                 new_class = f"{current_class} {class_name}".strip()
                 element.set("class", new_class)
 
-        elif isinstance(element, BeautifulSoup().new_tag("").__class__):
+        elif isinstance(element, Tag):
 
             if "class" not in element.attrs:
                 element.attrs["class"] = []
@@ -62,14 +59,7 @@ class BS4Treeprocessor(Treeprocessor):
                 if class_name not in current_class:
                     current_class.append(name)
         else:
-            raise TypeError(
-                "element can either be of type {}.{} or {}.{}".format(
-                    BeautifulSoup().new_tag("").__class__.__module__,
-                    BeautifulSoup().new_tag("").__class__.__qualname__,
-                    ElementTree.Element.__module__,
-                    ElementTree.Element.__qualname__,
-                )
-            )
+            raise TypeError("Unsupported element")
 
 
 class LinkBlankTargetExtension(Extension):
--- a/app/grandchallenge/core/utils/markdown.py
+++ b/app/grandchallenge/core/utils/markdown.py
@@ -38,7 +38,7 @@ class BS4Treeprocessor(Treeprocessor):
                     self.md.htmlStash.rawHtmlBlocks[i] = str(bs4block)
 
     @staticmethod
-    def set_css_class(element, class_name):
+    def set_css_class(*, element, class_name):
 
         if isinstance(element, ElementTree.Element):
 
--- a/app/grandchallenge/core/utils/markdown.py
+++ b/app/grandchallenge/core/utils/markdown.py
@@ -26,7 +26,9 @@ class BS4Treeprocessor(Treeprocessor):
         for el in root.iter():
 
             if el.tag in el_class_dict:
-                self.set_css_class(el, el_class_dict[el.tag])
+                self.set_css_class(
+                    element=el, class_name=el_class_dict[el.tag]
+                )
 
         for i, html_block in enumerate(self.md.htmlStash.rawHtmlBlocks):
 
@@ -34,7 +36,7 @@ class BS4Treeprocessor(Treeprocessor):
 
             for tag, tag_class in el_class_dict.items():
                 for el in bs4block.find_all(tag):
-                    self.set_css_class(el, tag_class)
+                    self.set_css_class(element=el, class_name=tag_class)
                     self.md.htmlStash.rawHtmlBlocks[i] = str(bs4block)
 
     @staticmethod
--- a/app/grandchallenge/core/utils/markdown.py
+++ b/app/grandchallenge/core/utils/markdown.py
@@ -1,7 +1,37 @@
+from xml.etree import ElementTree
+
 from bs4 import BeautifulSoup
 from markdown import Extension
-from markdown.postprocessors import Postprocessor
 from markdown.treeprocessors import Treeprocessor
+
+
+class HtmlElementWrapper:
+    def __init__(self, element):
+        self.element = element
+
+    def set_css_class(self, class_name: str):
+
+        if isinstance(self.element, ElementTree.Element):
+
+            current_class = self.element.attrib.get("class", "")
+
+            if class_name not in current_class:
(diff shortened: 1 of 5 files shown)


jmsmkn:
I guess this fixes it but it is a very minimal solution, only covering this specific case. So technically this works, but we need to solve things sustainably and not just for specific cases. Please think about the wider picture when implementing your solutions.

Images can be included in more places than blog posts. This needs to work everywhere that a user provides markdown. Additionally, what happens when someone sets an explicit width on the img?

pkcakeout:
What would be your "ideal" solution here then? A post-processor for markdown that transcribes the `` tags?

ammar257ammar:
> what happens when someone sets an explicit width on the img?
Do you want to forbid users form specifying attributes on images like width and height or customize any element in general?

ammar257ammar:
> images can be included in more places than blog posts

Good points, I will expand the solution

jmsmkn:
> > what happens when someone sets an explicit width on the img?
> 
> Do you want to forbid users form specifying attributes on images like width and height or customize any element in general?

We already do that with bleach. We allow setting width. We should continue to allow setting the width as that is how images work on Pages.

jmsmkn:
> What would be your "ideal" solution here then? A post-processor for markdown that transcribes the `` tags?

As pointed out in #3095 we already have a post processor for markdown.

ammar257ammar:
This PR is also done, Please review when possible.

jmsmkn:
Please un-assign yourself when you're done.

app/grandchallenge/core/utils/markdown.py
```diff
@@ -41,3 +43,29 @@ def run(self, root):
             if el.tag == "a":
                 el.set("target", "_blank")
                 el.set("rel", "noopener")
+
+
+class HtmlTagsExtension(Extension):
+    def extendMarkdown(self, md):  # noqa: N802
+        md.registerExtension(self)
+        md.postprocessors.register(
+            HtmlTagsPostprocessor(md), "htmltags_extension", 35
```
jmsmkn, line 52:
```suggestion
            HtmlTagsPostprocessor(md), "html_tags_extension", 35
```

Missing underscore. What does 35 mean?

app/grandchallenge/core/utils/markdown.py
```diff
@@ -41,3 +43,29 @@ def run(self, root):
...
+        )
+
+
+class HtmlTagsPostprocessor(Postprocessor):
+    def run(self, lines):
+        for i in range(len(self.md.htmlStash.rawHtmlBlocks)):
+            bs4block = BeautifulSoup(
+                self.md.htmlStash.rawHtmlBlocks[i], "html.parser"
+            )
+            img = bs4block.find("img")
```
jmsmkn, line 62: Does this work for multiple images? Or does the processor receive one block at a time?

jmsmkn: It's going block by block it seems. Initialising Beautiful soup for every block in the passed html seems inefficient. Why do you use a `PostProcessor` and not a `TreeProcessor`? Why do you iterate over the blocks and not look at lines? It seems we now have two processors that do the same thing (HtmlTagsPostprocessor and BS4Treeprocessor), why is their functionality not combined?

app/grandchallenge/core/utils/markdown.py
ammar257ammar: The number is the priority of the processor (order of execution, higher priority is earlier execution). In the package they use three levels of priority with steps of 10, so 30 is the highest, and smaller increments by 5 for in between. In the postprocessor level, the [RawHtmlPostprocessor](https://python-markdown.github.io/extensions/api/#usages_4) is used to inject back the HTML blocks after processing the markdow, so we want our processor (adding the classes to the HTML tag) to execute before the blocks are injected back into the output text, so I added an increment of 5 to the highest priority 30, so it is 35.

app/grandchallenge/core/utils/markdown.py
ammar257ammar: The HTML is extracted before processing the markdown and injected later after the processing, so it does not go throught the TreeProcessor. I thought of either using regex to get the img with classes, or to use BeautifoulSoup to parse the HTML. 
If I print the tags in the BS4Treeprocessor, I don't see the tags coming from the explicit HTML and that is why the class is not added to those tags in the first place.

ammar257ammar: And  indeed, it is two processors now, one for the Markdown elements (through the TreeProcessor) and one for the raw HTML elements (through the PostProcessor). Any suggestions for an alternative?

ammar257ammar: You can see [here](https://python-markdown.github.io/extensions/api/#usages)
Screenshot 2024-08-14 at 11 53 37
So the raw HTML is removed before processing the Markdown

jmsmkn: My question is why do we have two bits of logic doing the same thing, you can see that you have duplicated functionality. Why is that logic not being reused?

app/grandchallenge/core/utils/markdown.py
jmsmkn: This needs a comment, especially if ordering is important. What happens when the priorities upstream change? If this needs to be done before anything else then the priority needs to be automatically assigned from the things it needs to run before.

ammar257ammar: I changed the logic and moved the htmlStash processing to the BS4Treeprocessor which is executed always before the postprocessors, hence, no need for specific priority checkup. But still, the package does handle the raw html the same way as the converted markdown to html and thus they still need to be handles separately. I implemented a solution, check the comment on the review below.

app/grandchallenge/core/utils/markdown.py
ammar257ammar: I see you point. I agree, the logic should be combined. However, as said earlier, the package does not pass the raw html through the same treeprocessor as the html from converted markdown. So, we still need two for loops to handle them, but now I made this processing in the same phase (TreeProcessor) to not worry about the execution order in the postprocessor phase, and combined the logic for setting the css class to remove the duplicated functionality. I made a commit.

app/grandchallenge/core/utils/markdown.py
```diff
@@ -1,7 +1,39 @@
...
+            if "class" not in self.element.attrs:
+                self.element.attrs["class"] = []
+
+            current_class = self.element["class"]
+
+            for name in class_name.split(" "):
+                if class_name not in current_class:
+                    current_class.append(name)
+
+    def get_element(self):
```
jmsmkn, line 33: You shouldn't need a `get_` method for public attributes. Why is this thing a class at all? Seems like it should just be a function `set_css_class` which takes an element and class name and returns an element.

jmsmkn: A `staticmethod` in `BS4Treeprocessor` would make the most sense. Other things:

- it should be kwargs only
- The fallthrough in the switch statement needs to be explicitly handled

ammar257ammar: Do you mean by "kwargs only" that the method has a signature like 
```
@staticmethod
set_css_class(**kwargs)
``` 
or that it uses input from arguments only?
or to imply that `self` is not needed in the static method?

ammar257ammar: I made a commit, not sure if it is exactly as what you have in mind.

app/grandchallenge/core/utils/markdown.py
```diff
@@ -9,22 +12,64 @@ def extendMarkdown(self, md):  # noqa: N802
...
+            )
+
+            el = bs4block.find()
+
+            if el and el.name in el_class_dict:
+                BS4Treeprocessor.set_css_class(el, el_class_dict[el.name])
+                self.md.htmlStash.rawHtmlBlocks[i] = str(bs4block)
+
+    @staticmethod
+    def set_css_class(element, class_name: str):
```
jmsmkn, line 44:
```suggestion
    def set_css_class(*, element, class_name):
```

See https://peps.python.org/pep-3102/

No type annotations should be added as we're not checking them.

app/grandchallenge/core/utils/markdown.py
```diff
@@ -9,22 +12,64 @@ def extendMarkdown(self, md):  # noqa: N802
...
+                    current_class.append(name)
+        else:
+            raise TypeError(
+                "element can either be of type {}.{} or {}.{}".format(
+                    BeautifulSoup().new_tag("").__class__.__module__,
+                    BeautifulSoup().new_tag("").__class__.__qualname__,
+                    ElementTree.Element.__module__,
+                    ElementTree.Element.__qualname__,
+                )
+            )
```
jmsmkn, line 72:
```suggestion
            raise TypeError("Unsupported element")
```

Code style - don't use `format`, use `f` strings instead. All the detail is unnecessary.

app/grandchallenge/core/utils/markdown.py
```diff
@@ -9,22 +12,64 @@ def extendMarkdown(self, md):  # noqa: N802
...
+
+        if isinstance(element, ElementTree.Element):
+
+            current_class = element.attrib.get("class", "")
+
+            if class_name not in current_class:
+                new_class = f"{current_class} {class_name}".strip()
+                element.set("class", new_class)
+
+        elif isinstance(element, BeautifulSoup().new_tag("").__class__):
```
jmsmkn, line 54: Why are you initialising BeautifulSoup, creating a new tag, then getting the class? Why doesn't this work:

```suggestion
from bs4 import Tag
...
        elif isinstance(element, Tag):
```

app/grandchallenge/core/utils/markdown.py
```diff
@@ -9,22 +12,64 @@ def extendMarkdown(self, md):  # noqa: N802
...
+        }
+
         for el in root.iter():
-            if el.tag == "img":
-                el.set("class", "img-fluid")
 
-            elif el.tag == "blockquote":
-                el.set("class", "blockquote")
+            if el.tag in el_class_dict:
+                BS4Treeprocessor.set_css_class(el, el_class_dict[el.tag])
```
jmsmkn, line 29: Why hardcode the class name?
```suggestion
                self.set_css_class(el, el_class_dict[el.tag])
```

app/grandchallenge/core/utils/markdown.py
```diff
@@ -9,22 +12,64 @@ def extendMarkdown(self, md):  # noqa: N802
...
+        for i in range(len(self.md.htmlStash.rawHtmlBlocks)):
+
+            bs4block = BeautifulSoup(
+                self.md.htmlStash.rawHtmlBlocks[i], "html.parser"
+            )
+
+            el = bs4block.find()
+
+            if el and el.name in el_class_dict:
+                BS4Treeprocessor.set_css_class(el, el_class_dict[el.name])
```
jmsmkn, line 40: Again...

```suggestion
                self.set_css_class(el, el_class_dict[el.name])
```

app/grandchallenge/core/utils/markdown.py
```diff
@@ -9,22 +12,64 @@ def extendMarkdown(self, md):  # noqa: N802
...
         for el in root.iter():
-            if el.tag == "img":
-                el.set("class", "img-fluid")
 
-            elif el.tag == "blockquote":
-                el.set("class", "blockquote")
+            if el.tag in el_class_dict:
+                BS4Treeprocessor.set_css_class(el, el_class_dict[el.tag])
+
+        for i in range(len(self.md.htmlStash.rawHtmlBlocks)):
```
jmsmkn, line 31: Why are you getting the length of the blocks and then looking them up? Why not iterate over them directly?

app/grandchallenge/core/utils/markdown.py
```diff
@@ -9,22 +12,64 @@ def extendMarkdown(self, md):  # noqa: N802
...
+
+            bs4block = BeautifulSoup(
+                self.md.htmlStash.rawHtmlBlocks[i], "html.parser"
+            )
+
+            el = bs4block.find()
```
jmsmkn, line 37: What is this finding? Everything? Why not only the relevant things you're looking for?

app/grandchallenge/core/utils/markdown.py
```diff
@@ -9,22 +12,64 @@ def extendMarkdown(self, md):  # noqa: N802
...
+
     def run(self, root):
+
+        el_class_dict = {
+            "img": "img-fluid",
+            "blockquote": "blockquote",
+            "table": "table table-hover table-borderless",
+            "thead": "thead-light",
+            "code": "codehilite",
+        }
```
jmsmkn, line 24: Now you're modifying all of these tags, it all needs tests which combine all of these blocks set either through markdown syntax or html directly.

ammar257ammar: There is already a test for those tags/classes in markdown format. Do you recommend making a new test for the HTML case or modify the existing test? that test is not parametrized, so I need to modify parts of it. You told me before not to change other things, so I am checking beforehand.

jmsmkn: It is relevant to this changeset so change it.

app/grandchallenge/core/utils/markdown.py
ammar257ammar: Done

ammar257ammar: Done

app/grandchallenge/core/utils/markdown.py
ammar257ammar: Done

app/grandchallenge/core/utils/markdown.py
ammar257ammar: Done

app/grandchallenge/core/utils/markdown.py
ammar257ammar: Done

app/grandchallenge/core/utils/markdown.py
ammar257ammar: Done

app/grandchallenge/core/utils/markdown.py
ammar257ammar: We have a dictionary of elements to add classes two, and a list of rawHtmlBlocks. So either loop over the elements in the HtmlBlock and check if the element is in the dictionary, or, loop over the dictionary and `find_all` the element in the HtmlBlock. The dictionary is not long (for now) and also the HtmlBlocks not expected to be long (since the editor is a markdown editor and raw html is not the norm there), so I don't have a hard opinion here. However, I followed your comment and checked for each element in the dictionary with `find_all()`.

app/grandchallenge/core/utils/markdown.py
ammar257ammar: Done

jmsmkn (approved):
Well done!

jmsmkn (approved):

ammar257ammar:
Thanks!!

PR #3484, opened 2024-08-08
Fix chart rendering issue on pages with BS tabs

### LINKED ISSUE #3446: Algorithm usage chart does not render
Screenshot 2024-07-16 at 16 17 40

Resizing the screen makes it render. This is on Safari. I think Chris worked on such an issue in the past.

ammar257ammar:
The issue relates to the rendering of vega-lite charts with responsive specs `width: 'container'` on pages that contains bootstrap tabs with dynamic content. The solution was to utilize the IntersectionObserve in place to run the rendering function only when there is intersection between the chart element and its root. Also, the issue arises on pages with JQuery Datatables in which the vega-chart HTML elements are fetched through an AJAX request, and thus they are not available in DOM on page load to be registered to the IntersectionObserve. The solution to this is to use a MutationObserver that obsereves the ajax table container for added nodes and triggers the vega chart rendering whenever a new chart element obtained.

I tested the fix with pages that contains tabs and charts:
- Algorithm details page (information tab)
- Job details page (logs tab)
- Evaluation details page (logs tab)
- Archive item details page (civ chart value, if available) 
- Reader study display set page (civ chart value, if available) 

Also, test the fix with pages that contains ajax tables were a civ chart may exist in the table:
- Job list/algorithm results page
- Archive items page
- Reader study display sets page

Also, tested that the fix does not affect charts in pages without tabs, namely:
- Challenge statistics page
- General GC statistics page

closes #3446


jmsmkn (commented):
Much better.

Now the question is why we have two methods for rendering charts. Right now, future developers need to be aware of this bug and use one of two classes that achieve the same result. Why are there two classes - `vega-lite-chart` and `vega-lite-chart-lazy`? Why can't developers just use `vega-lite-chart` and then have it handle the rendering?

ammar257ammar:
The only difference I can see is that the ones without -lazy will just render on page loaded event. The lazy loading is happening through observing the element and triggering the rendering if it is not already rendered and satisfy the interaction observer condition. So basically, the lazy loading works for both scenarios. I already tested it for the general stastistics page and challenge statistics pages and it works.

The only place where I see a choice between the two classes is here. This is in component interface value (of type chart) to choose if it should be views in a modal or in the page (modal needs lazy-loading since the content is dynamically changing) but even then as with other cases like the tab, the lazy-loading works with both cases. However, I would like to test it with civ (chart kind) is not affected but I don't have any example on how the chart civ look like (no example in development fixtures), and I can't add one myself (see screenshot below). Any hint on how to do it?

Screenshot 2024-08-08 at 16 19 55



Do you want to adopt only this way of chart rendering (with observer) and renaming all classes to vega-lite-chart (without -lazy) to have a unified way of chart rendering?

chrisvanrun:
Rendering the charts only when they are scrolled into view can be quite confusing. Originally, the lazy rendering was only introduced to allow having 50+ charts defined on the same page. I don't see why it would not work having that as default, but I would suggest double-checking to see if the statistic page isn't going ballistic.

I _would_ suggest adding a spinner so users know something is there. If you make it a child of the 'vega-lite-chart' classed element, it will be removed when the chart is rendered:

```HTML
<div class="spinner-border" role="status">
   <span class="sr-only">Loading...</span>
</div>
```

jmsmkn:
I don't _think_ a spinner is necessary - I wouldn't imagine that it would make much difference with the stats page given everything is at once at page load immediately right now, and there is no obvious delay on my mac.

ammar257ammar:
Okay, I am working on using one way of rendering the chart (with observers) and getting rid of the `-lazy `class name by renaming all classes to `vega-lite-chart`.

I discovered that for the component interface values with type (CHART), in several places, the views does not load vega js files so the chart does not render. For example, in archive item details page, if one of the CIVs is a chart, the view is like this:

Screenshot 2024-08-09 at 09 33 55

To solve this, vega js files need to be imported in [civ_set_detail.html](https://github.com/comic/grand-challenge.org/blob/874793808e154d4af2f6cdc490ac7c57e75b0b41/app/grandchallenge/components/templates/components/civ_set_detail.html). I made a commit for it.

The result is this now:

Screenshot 2024-08-09 at 09 47 07


There is another issue in the job list details view, where 
- First, the vega js files are not loaded
- Second, the civ value is actually fetched (included in the whole table row) through an AJAX request. That means the DOM of the vega-lite-chart container is not in the page on load, so even with vega js files imported, the element is not in the DOM to be registered with the observer, and thus render when opening the modal. 

Screenshot 2024-08-09 at 09 53 19

In this case, the event of registering the fetched vega-lite element in the observer has to be triggered manually after the successful AJAX call. I suggest creating an event (other than DOMContentLoaded to avoid dispatching it with all other actions linked to it) and dispatch that event to register the ajax-fetched vega element.

What do you think?

btw, there are still a couple of other places where civs are involved (evaluation details, job details) where the CHART civ need to be tested. I am working on this too.

app/grandchallenge/charts/static/js/charts/render_charts.mjs
```diff
@@ -41,9 +34,35 @@ const observer = new IntersectionObserver(handleInterSection, {
...
+                for(const element of addedNode.getElementsByClassName("vega-lite-chart")){
+                    if (!element.getAttribute(handledAttribute)) {
+                        renderVegaLiteChart(element);
+                        element.setAttribute(handledAttribute, "");
+                    }
+                }
+            }
+        })
+    })
+}
```
jmsmkn, line 61: We do not need all of this whitespace.

```suggestion
function mutationObserverCallback(mutationList, observer) {
    mutationList.forEach((mutation) => {
        mutation.addedNodes.forEach((addedNode) => {
            if (addedNode.nodeType !== Node.TEXT_NODE) {
                for(const element of addedNode.getElementsByClassName("vega-lite-chart")){
                    if (!element.getAttribute(handledAttribute)) {
                        renderVegaLiteChart(element);
                        element.setAttribute(handledAttribute, "");
                    }
                }
            }
        })
    })
}
```

app/grandchallenge/core/static/js/refresh_sidebar.mjs
```diff
@@ -1,4 +1,5 @@
 $(document).ready(function() {
+
```
jmsmkn, line 2:
```suggestion
```

app/grandchallenge/charts/static/js/charts/render_charts.mjs
```diff
@@ -41,9 +34,35 @@ const observer = new IntersectionObserver(handleInterSection, {
...
+
+const mutationObserver = new MutationObserver(mutationObserverCallback);
+
+if(document.getElementById("ajaxDataTable")){
+    mutationObserver.observe(document.getElementById("ajaxDataTable"),
+                        {childList: true, subtree: true,});
```
jmsmkn, line 67:
```suggestion
    mutationObserver.observe(
        document.getElementById("ajaxDataTable"),
        {childList: true, subtree: true,}
    );
```

app/grandchallenge/charts/static/js/charts/render_charts.mjs
```diff
@@ -41,9 +34,33 @@ const observer = new IntersectionObserver(handleInterSection, {
...
         observer.observe(element, );
     }
 });
+
+
+function mutationObserverCallback(mutationList, observer) {
+    mutationList.forEach((mutation) => {
+        mutation.addedNodes.forEach((addedNode) => {
+            if (addedNode.nodeType !== Node.TEXT_NODE) {
+                for(const element of addedNode.getElementsByClassName("vega-lite-chart")){
```
jmsmkn, line 48:
```suggestion
                for (const element of addedNode.getElementsByClassName("vega-lite-chart")) {
```

app/grandchallenge/charts/static/js/charts/render_charts.mjs
```diff
@@ -41,9 +34,33 @@ const observer = new IntersectionObserver(handleInterSection, {
...
+                    }
+                }
+            }
+        })
+    })
+}
+
+const mutationObserver = new MutationObserver(mutationObserverCallback);
+
+if(document.getElementById("ajaxDataTable")){
```
jmsmkn, line 61:
```suggestion
if (document.getElementById("ajaxDataTable")) {
```

app/grandchallenge/charts/static/js/charts/render_charts.mjs
```diff
@@ -41,9 +34,33 @@ const observer = new IntersectionObserver(handleInterSection, {
...
+                    }
+                }
+            }
+        })
+    })
+}
+
+const mutationObserver = new MutationObserver(mutationObserverCallback);
+
+if (document.getElementById("ajaxDataTable")) {
```
jmsmkn, line 61: `render_charts.js` now needs to know about ajax data tables. This also won't work if the table has a different id, or there are multiple tables in the DOM. Can't the mutationObserver be exported and the tables that contain charts be registered?

ammar257ammar: Since the render_charts.mjs is loaded on the page as a JS module, the variables defined inside it have a module scope and not global scope, so they are not accessible from other scripts. That was the reason I asked the question on Teams chat since I was trying to make the Datatable register itself to the observer and I am not sure if render_charts.mjs is actually a module since its logic now need to know about things happening in the page and not entirely self-contained, maybe I am wrong and I need more reading on this topic.

One suggestion is to make the observer observe the body and not specifically to an element so it will detect any chart whenever added to the screen. Also, since the view are not heavily dynamic (i.e. with many JS node additions/deletions), I don't expect that to affect performance and the observer only acts on node additions and on a specific class. What do you think?

ammar257ammar: Thanks! I am implementing it. A new thing I learned which is that `import` statement can not be used outside a module. So the file `list.js` should be loaded as a module in the page. I am testing it now.

jmsmkn: Yes, in general we should refactor JS files to modules, but each one needs to be checked carefully as it doesn't always work.

app/grandchallenge/datatables/templates/datatables/list_base.html
```diff
@@ -13,4 +13,9 @@
     {{ default_sort_order|json_script:"defaultSortOrder" }}
 
     <script type="text/javascript" src="{% static 'js/datatables/list.js' %}"></script>
+
+    <script src="{% static 'vendored/vega/vega.min.js' %}"></script>
+    <script src="{% static 'vendored/vega-lite/vega-lite.min.js' %}"></script>
+    <script src="{% static 'vendored/vega-embed/vega-embed.min.js' %}"></script>
+    <script type="module" src="{% static 'js/charts/render_charts.mjs' %}"></script>
```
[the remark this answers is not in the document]
ammar257ammar: I removed it from list_base and loaded the scripts where the charts are, namely, job list page, reader studies display sets list page, and archive items list page.

app/grandchallenge/components/templates/components/civ_set_list.html
```diff
@@ -51,4 +51,8 @@ <h2>{{ base_object.title }} {{ base_object.civ_set_model|verbose_name_plural|tit
 {% block script %}
     {{ block.super }}
     {% include 'workstations/partials/session-control.html' %}
+    <script src="{% static 'vendored/vega/vega.min.js' %}"></script>
+    <script src="{% static 'vendored/vega-lite/vega-lite.min.js' %}"></script>
+    <script src="{% static 'vendored/vega-embed/vega-embed.min.js' %}"></script>
+    <script type="module" src="{% static 'js/charts/render_charts.mjs' %}"></script>
```
jmsmkn, line 57:
```suggestion
    {{ block.super }}
    
    {% include 'workstations/partials/session-control.html' %}
    
    <script src="{% static 'vendored/vega/vega.min.js' %}"></script>
    <script src="{% static 'vendored/vega-lite/vega-lite.min.js' %}"></script>
    <script src="{% static 'vendored/vega-embed/vega-embed.min.js' %}"></script>
    <script type="module" src="{% static 'js/charts/render_charts.mjs' %}"></script>
```

app/grandchallenge/datatables/templates/datatables/list_base.html
```diff
@@ -12,5 +12,5 @@
     {{ text_align|json_script:"textAlign" }}
     {{ default_sort_order|json_script:"defaultSortOrder" }}
 
-    <script type="text/javascript" src="{% static 'js/datatables/list.js' %}"></script>
+    <script type="module" src="{% static 'js/datatables/list.js' %}"></script>
```
jmsmkn, line 15: Now it's a module the extension should be `.mjs`

app/grandchallenge/charts/static/js/charts/render_charts.mjs
```diff
@@ -41,9 +34,28 @@ const observer = new IntersectionObserver(handleInterSection, {
...
+                        renderVegaLiteChart(element);
+                        element.setAttribute(handledAttribute, "");
+                    }
+                }
+            }
+        })
+    })
+}
+
+const mutationObserver = new MutationObserver(mutationObserverCallback);
```
jmsmkn, line 59: `mutationObserver` and `mutationObserverCallback` should have more specific names. What do these functions do? Maybe `mutationObserverCallback` is `renderVegaChartsInAddedNodes`? And the observer is `renderVegaChartsObserver`?

jmsmkn (changes requested):
Pre-commit is failing. @HarmvZ can you take a look at the JS changes?

app/grandchallenge/charts/static/js/charts/render_charts.mjs
ammar257ammar: I agree. Done

HarmvZ (approved):
LGTM.

There definitely is something to gain if we use a linter and/or formatter to unify all the different coding styles in the js files in this repo.

jmsmkn (approved):

PR #3488, opened 2024-08-08
Add endpoint for lowest supported gcapi version

HarmvZ:
It is set to 0.12.0 since this will probably be the version in which the version check will be included.

I'm not sure about the long URL, open to suggestions.

### DIFF
--- a/app/config/settings.py
+++ b/app/config/settings.py
@@ -150,6 +150,8 @@ CHALLENGE_NUM_SUPPORT_YEARS = int(
     os.environ.get("CHALLENGE_NUM_SUPPORT_YEARS", 5)
 )
 
+GCAPI_LOWEST_SUPPORTED_VERSION = "0.12.0"
+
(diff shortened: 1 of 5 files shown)


jmsmkn:
Looks good but lets think about future proofing as we may want to extend this with more information (specifying broken versions, versions per feature, max supported version etc). Could you make this a general API endpoint, call it gcapi, then return a JSON object with:

```
{
  "lowest_supported_version": "0.13.0",
}
```

May be easiest to use a DRF serializer and view even though it is static data. Version 0.13.0 as 0.12.0 is out already without this check.

HarmvZ:
Ok, will do.

> as 0.12.0 is out already without this check.

Ah, I see now that there is a mistake in the gcapi changelog, it says unreleased for 0.12.0 while it was released already in February last year.

jmsmkn (approved):

PR #3503, opened 2024-08-16
Add prettier to unify formatting of js files

HarmvZ:
I've turned it off for other filetypes like CSS, HTML and json. We could enable it for that but it will probably create a huge changeset and I'm not sure if we want that.

Sidenote: Adding a linter like ESlint would be another improvement to catch linting errors like using `==` instead of `===` or using `let` or `var` instead of `const`.


jmsmkn (approved):
Thanks!

Issue #3495, opened 2024-08-14
Result text no longer visible in job list view


jmsmkn:
Found in #3484. Counts for any job list.

PR #3506, opened 2024-08-19
Show result text in job list

chrisvanrun:
Closes: #3495

Re-adds the rendered result text for direct viewing in the job lists. This used to be cramped with the status and results. It is now in its own column. 

The alternative was adding it directly to the results list, but this makes it clearer that it is a singular representation of the result.

Extra CSS was needed to ensure the forcefully added `<p>` in the template rendering does not line out weirdly:

# Pre-CSS
image

# Post-CSS
image

### DIFF
--- a/app/grandchallenge/core/static/css/core.css
+++ b/app/grandchallenge/core/static/css/core.css
@@ -10,7 +10,7 @@ p {
     line-height: 1.6;
 }
 
-.child-mb-0>* {
+.child-p-mb-0 > p {
     margin-bottom: 0;
 }
 
--- a/app/grandchallenge/core/static/css/core.css
+++ b/app/grandchallenge/core/static/css/core.css
@@ -10,10 +10,6 @@ p {
     line-height: 1.6;
 }
 
-.child-mb-0>* {
-    margin-bottom: 0;
-}
-
 .jumbotron {
     text-shadow: 0.1rem 0.1rem 0.2rem #000000;
 }
@@ -29,8 +25,7 @@ p {
 }
 
 /* fix modal display with overflow-y: scroll on html */
-.modal,
-.modal-open {
+.modal, .modal-open {
     padding-right: 0 !important;
 }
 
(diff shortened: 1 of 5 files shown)


app/grandchallenge/core/static/css/core.css
```diff
@@ -25,7 +29,8 @@ p {
 }
 
 /* fix modal display with overflow-y: scroll on html */
-.modal, .modal-open {
+.modal,
```
jmsmkn, line 32: Please undo all of the css formatting changes.

app/grandchallenge/core/static/css/core.css
```diff
@@ -10,6 +10,10 @@ p {
     line-height: 1.6;
 }
 
+.child-mb-0>* {
```
jmsmkn, line 13: Seems a bit too non-specific? Or is this a problem in many places?

chrisvanrun: I thought to use the bootstrap approach and describe the class with what it does. An alternative is to use a very specific 'rendered-result-text' class, but that seemed totally out of place here.

I for see we might want to do this in other places in the future.

app/grandchallenge/core/static/css/core.css
chrisvanrun: Mea Culpa, I did not notice. 

It seems like the basic CSS-language-support for VScode has formatting-on-save turned on by default.

app/grandchallenge/core/static/css/core.css
jmsmkn: Doesn't MarkDown wrap everything in `<p>` so we only need to select that?

chrisvanrun: It does. We can make it a bit more specific with `.child-p-md-0>p`.

app/grandchallenge/core/static/css/core.css
jmsmkn: Okay we can do this with pre-commit now.

jmsmkn (approved):

PR #3508, opened 2024-08-19
Fix using algorithm model instructions instead of ground truth

chrisvanrun:
Quick fix: reported by a challenge org.

image


jmsmkn (approved):
Thanks, copy paste fail

PR #3510, opened 2024-08-22
Replace prettier with biome for js formatting

HarmvZ:
I tried installing ESLint at first but this is not possible without requiring node in the dev environment.

So this removes prettier and adds [biome](https://biomejs.dev/). This is a linter and formatter in one. It's a replacement for prettier + ESLint.


HarmvZ:
> app/grandchallenge/core/static/js/markdownx.js shouldn't be touched. Can the prettier configurations .prettierrc and .prettierignore also be removed?

Fixed.

> Does biome format css also?

Not yet: https://biomejs.dev/internals/language-support/

> Do we need to pin to precisely 1.8.3?

Not precisely indeed, I've changed it to ^1.8.3.

jmsmkn (approved):

PR #3514, opened 2024-08-28
Auto process pending permission requests upon verification (accept verified users only)

### LINKED ISSUE #2461: Refresh access for users that apply or lose verification after placing access request
I see this for algorithms that auto-accept verified users now. Users that place an access request for an algorithm first and then apply for verification do not get automatically accepted after their verification is successful.

If a user passes or loses verification, all their access requests can be looped through and updated depending on the request handling rule per algorithm/challenge/reader study.

ammar257ammar:
This fix will add a step to the Verification model save method which will check all pending requests and accept them if the user is verified and the object (which the permission is requested for) has a request handling setting of accepting verified users only. 

Screenshot 2024-09-09 at 12 27 40

closes #2461

### DIFF
--- a/app/grandchallenge/core/models.py
+++ b/app/grandchallenge/core/models.py
@@ -81,9 +81,7 @@ class RequestBase(models.Model):
                 and self.base_object.access_request_handling
                 == utils.access_requests.AccessRequestHandlingOptions.ACCEPT_VERIFIED_USERS
             ):
-                status = (
-                    "You need to verify your account in order to get access"
-                )
+                status += ", is awaiting review. Your request will be automatically accepted if you verify your account."
             else:
                 status += ", is awaiting review"
 
--- a/app/grandchallenge/core/models.py
+++ b/app/grandchallenge/core/models.py
@@ -9,6 +9,8 @@ from django_extensions.db.models import (
     TitleSlugDescriptionModel as BaseTitleSlugDescriptionModel,
 )
 
+from grandchallenge.core import utils
+
 
 class TitleSlugDescriptionModel(BaseTitleSlugDescriptionModel):
     # Fix issue in upstream where description can be null
@@ -63,20 +65,27 @@ class RequestBase(models.Model):
         super().save(*args, **kwargs)
 
     def status_to_string(self):
+
         status = (
             f"Your request to join {self.object_name}, "
             f"sent {self.format_date(self.created)}"
         )
         if self.status == self.PENDING:
-            status += ", is awaiting review"
-
             try:
                 user_is_verified = self.user.verification.is_verified
             except ObjectDoesNotExist:
                 user_is_verified = False
 
-            if not user_is_verified:
-                status += ". You need to verify your account in order to request access permissions"
+            if (
+                not user_is_verified
+                and self.base_object.access_request_handling
+                == utils.access_requests.AccessRequestHandlingOptions.ACCEPT_VERIFIED_USERS
+            ):
+                status = (
+                    "You need to verify your account in order to get access"
+                )
+            else:
+                status += ", is awaiting review"
 
         elif self.status == self.ACCEPTED:
             status += ", was accepted at " + self.format_date(self.changed)
--- a/app/grandchallenge/core/models.py
+++ b/app/grandchallenge/core/models.py
@@ -69,10 +69,6 @@ class RequestBase(models.Model):
         )
         if self.status == self.PENDING:
             status += ", is awaiting review"
-        elif self.status == self.ACCEPTED:
-            status += ", was accepted at " + self.format_date(self.changed)
-        elif self.status == self.REJECTED:
-            status += ", was rejected at " + self.format_date(self.changed)
 
             try:
                 user_is_verified = self.user.verification.is_verified
@@ -81,6 +77,12 @@ class RequestBase(models.Model):
 
             if not user_is_verified:
                 status += ". You need to verify your account in order to request access permissions"
+
+        elif self.status == self.ACCEPTED:
+            status += ", was accepted at " + self.format_date(self.changed)
+        elif self.status == self.REJECTED:
+            status += ", was rejected at " + self.format_date(self.changed)
+
         return status
 
     @staticmethod
--- a/app/grandchallenge/core/models.py
+++ b/app/grandchallenge/core/models.py
@@ -1,6 +1,7 @@
 import uuid
 
 from django.conf import settings
+from django.core.exceptions import ObjectDoesNotExist
 from django.db import models
 from django.utils.timezone import localtime
 from django.utils.translation import gettext_lazy as _
@@ -72,6 +73,14 @@ class RequestBase(models.Model):
             status += ", was accepted at " + self.format_date(self.changed)
         elif self.status == self.REJECTED:
             status += ", was rejected at " + self.format_date(self.changed)
+
+            try:
+                user_is_verified = self.user.verification.is_verified
+            except ObjectDoesNotExist:
+                user_is_verified = False
+
+            if not user_is_verified:
+                status += ". You need to verify your account in order to request access permissions"
         return status
 
     @staticmethod
--- a/app/grandchallenge/core/utils/access_requests.py
+++ b/app/grandchallenge/core/utils/access_requests.py
@@ -40,10 +40,13 @@ def process_access_request(request_object):
             if request_object.user.verification.is_verified:
                 # immediately allow access, no need for a notification
                 request_object.status = request_object.ACCEPTED
-                request_object.save()
-                return
+            else:
+                request_object.status = request_object.REJECTED
+            request_object.save()
+            return
         except ObjectDoesNotExist:
-            pass
+            request_object.status = request_object.REJECTED
+            request_object.save()
 
     follow(
         user=request_object.user,
--- a/app/grandchallenge/core/utils/access_requests.py
+++ b/app/grandchallenge/core/utils/access_requests.py
@@ -40,13 +40,10 @@ def process_access_request(request_object):
             if request_object.user.verification.is_verified:
                 # immediately allow access, no need for a notification
                 request_object.status = request_object.ACCEPTED
-            else:
-                request_object.status = request_object.REJECTED
-            request_object.save()
-            return
+                request_object.save()
+                return
         except ObjectDoesNotExist:
-            request_object.status = request_object.REJECTED
-            request_object.save()
+            pass
 
     follow(
         user=request_object.user,
--- a/app/grandchallenge/verifications/models.py
+++ b/app/grandchallenge/verifications/models.py
@@ -2,6 +2,7 @@ from datetime import timedelta
 
 from allauth.account.models import EmailAddress
 from allauth.account.signals import email_confirmed
+from django.apps import apps
 from django.conf import settings
 from django.contrib.auth import get_user_model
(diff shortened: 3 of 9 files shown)


app/grandchallenge/verifications/signals.py
```diff
@@ -93,3 +100,22 @@ def auto_disable_user_in_verification_user_set(
                         vus_link=verification_user_set.get_absolute_url(),
                     ),
                 )
+
+
+@receiver(post_save, sender=Verification)
```
jmsmkn, line 105: Signals should only be used when absolutely necessary. There is no reason to use a signal here. This should be done in the save method of the verification, and only be invoked if `is_verified` has changed to `True`, see usages of `FieldChangeMixin` to help you with that.

app/grandchallenge/core/utils/access_requests.py
```diff
@@ -40,10 +40,13 @@ def process_access_request(request_object):
...
-                request_object.save()
-                return
+            else:
+                request_object.status = request_object.REJECTED
+            request_object.save()
+            return
         except ObjectDoesNotExist:
-            pass
+            request_object.status = request_object.REJECTED
+            request_object.save()
```
jmsmkn, line 49: They should not be automatically rejected. People cannot make a request again in this case.

app/grandchallenge/core/models.py
```diff
@@ -72,6 +73,14 @@ def status_to_string(self):
...
         elif self.status == self.REJECTED:
             status += ", was rejected at " + self.format_date(self.changed)
+
+            try:
+                user_is_verified = self.user.verification.is_verified
+            except ObjectDoesNotExist:
+                user_is_verified = False
+
+            if not user_is_verified:
+                status += ". You need to verify your account in order to request access permissions"
```
jmsmkn, line 83: They end up permanently rejected. They should be left in a pending state. Then, if verified accounts are auto accepted for this thing they should be informed to verify their account.

ammar257ammar: Done. Now the message is updated for the requests with pending status only.

app/grandchallenge/core/models.py
```diff
@@ -1,13 +1,16 @@
...
 from django.conf import settings
+from django.core.exceptions import ObjectDoesNotExist
 from django.db import models
 from django.utils.timezone import localtime
 from django.utils.translation import gettext_lazy as _
 from django_extensions.db.models import (
     TitleSlugDescriptionModel as BaseTitleSlugDescriptionModel,
 )
 
+from grandchallenge.core import utils
```
jmsmkn, line 12: Why is this line being imported?

app/grandchallenge/core/models.py
```diff
@@ -62,16 +65,33 @@ def save(self, *args, **kwargs):
...
+                user_is_verified = False
+
+            if (
+                not user_is_verified
+                and self.base_object.access_request_handling
+                == utils.access_requests.AccessRequestHandlingOptions.ACCEPT_VERIFIED_USERS
+            ):
+                status = (
+                    "You need to verify your account in order to get access"
+                )
```
jmsmkn, line 86: This status isn't necessarily correct. People _could_ still be accepted, despite not being verified. 

```suggestion
                status += ", is awaiting review. Your request will be automatically accepted if you verify your account."
```

app/grandchallenge/verifications/models.py
```diff
@@ -80,6 +83,29 @@ def save(self, *args, **kwargs):
...
+                apps.get_model(
+                    app_label="reader_studies",
+                    model_name="ReaderStudyPermissionRequest",
+                ),
+            ]
+
+            for request_class in permission_request_classes:
+                pending_requests = request_class.objects.filter(
+                    user=self.user, status=request_class.PENDING
+                )
```
jmsmkn, line 104: This includes all requests. We only need ones where the `access_request_handling` is `ACCEPT_VERIFIED_USERS`.

app/grandchallenge/verifications/models.py
```diff
@@ -80,6 +83,29 @@ def save(self, *args, **kwargs):
...
+
+                for request in pending_requests:
+                    process_access_request(request_object=request)
```
jmsmkn, line 107: Why do you need to process each request? Why not batch update them all to accepted?

app/grandchallenge/verifications/models.py
jmsmkn: There is not a test to ensure that only requests with `access_request_handling == ACCEPT_VERIFIED_USERS` are updated.

app/grandchallenge/verifications/models.py
```diff
@@ -80,6 +83,29 @@ def save(self, *args, **kwargs):
 
         super().save(*args, **kwargs)
 
+        if self.has_changed("is_verified") and self.is_verified:
+            permission_request_classes = [
```
jmsmkn, line 87: The contents of block should be in its own method. See L109-L110.

app/grandchallenge/verifications/models.py
ammar257ammar: Do you mean update the model `status` value to True through ORM?
The `process_access_request` function does two other things besides status update, a call to `follow()`and a notification `send()`
Are those not needed in this case?

jmsmkn: You should be able to answer that yourself by reading `process_access_request`.

ammar257ammar: I did read the function. The follow function allows the object activities to appear in the user's stream and this is applied to every request in general. If we opt for changing only the status value in this case, that means some requests are not being handled in the same way across the app.
My question still, do you want to keep those function calls for this case or not?

app/grandchallenge/verifications/models.py
ammar257ammar: Done

app/grandchallenge/verifications/models.py
ammar257ammar: Done

ammar257ammar: btw, process_access_request already checks if the request handling is to accept verified users only ([here](https://github.com/comic/grand-challenge.org/blob/093c5980e57b71f664cdd139997685a84ee07529/app/grandchallenge/core/utils/access_requests.py#L37)), even if we did not filter for them in the verifications model save(). I added the filter anyway to the mentioned code (line 103).

Should I test process_access_request in this case, or the verification model?

app/grandchallenge/verifications/models.py
jmsmkn: Please read the function again because you clearly did not read it closely enough.

No notifications are generated if `request_object.base_object.access_request_handling == AccessRequestHandlingOptions.ACCEPT_VERIFIED_USERS` and `request_object.user.verification.is_verified`.

Notifications are generated for other things, but you should not be touching those requests in the first place, change it!

app/grandchallenge/verifications/models.py
jmsmkn: You should only be modifying requests with `access_request_handling == ACCEPT_VERIFIED_USERS`. You need a filter for that. You need a test for that.

app/grandchallenge/verifications/models.py
jmsmkn: Also - WHO received those requests? It's the admins of the challenge/reader study/archive. Why do they need to get notified AGAIN that someone request access when their verification status changed? It has nothing to do with them, they've already been notified and their verification status is irrelevant.

app/grandchallenge/verifications/models.py
jmsmkn: No it isn't.

app/grandchallenge/verifications/models.py
ammar257ammar: Understood. Since the call of this function is within save(), I tested it by making two permission requests (one for object with manual review and the other for accept_verified_users_only). Then checked the status after verification to make sure the request for object with manual review is still pending.

app/grandchallenge/verifications/models.py
ammar257ammar: Now I see the bigger picture. Thank you for the clear reasoning and explanation 👍

app/grandchallenge/verifications/models.py
```diff
@@ -131,6 +139,30 @@ def send_verification_email(self):
             user_email_override={self.user: self.email},
         )
 
+    def handle_pending_requests(self):
```
jmsmkn, line 142: This method name is misleading and should be changed. What is it doing? It is accepting pending requests for verified users. It should probably double check that the user is verified (or move that check to within the function call) before doing anything else in case a future developer calls it on the wrong object.

app/grandchallenge/verifications/models.py
```diff
@@ -131,6 +139,30 @@ def send_verification_email(self):
...
+            "archive": apps.get_model(
+                app_label="archives", model_name="ArchivePermissionRequest"
+            ),
+            "reader_study": apps.get_model(
+                app_label="reader_studies",
+                model_name="ReaderStudyPermissionRequest",
+            ),
+        }
+
+        for object_name, request_class in permission_request_classes.items():
```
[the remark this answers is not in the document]
ammar257ammar: I was trying to figure out how challenge requests fit into this issue. After spending some time, I realised you probably meant `RegistrationRequest` from the `participants` module, right?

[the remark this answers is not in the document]
ammar257ammar: Then, I tried adding registration requests form admin, but I don't find that model in admin. Also, there is no admin.py for the `participants` module. Is that on purpose?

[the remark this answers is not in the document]
jmsmkn: You need to look at everything that inherits the base class, the admin should not have anything to do with it.

app/grandchallenge/verifications/models.py
```diff
@@ -131,6 +139,35 @@ def send_verification_email(self):
             user_email_override={self.user: self.email},
         )
 
+    def accept_pending_requests_for_verified_users(self):
+
+        if self.is_verified:
```
jmsmkn, line 144: It is often clearer to escape early rather than nesting

```suggestion
        if not self.is_verified:
            raise RuntimeError("Refusing to accept verifications for unverified user")
            
        permission_request_classes = ...
```

app/grandchallenge/verifications/models.py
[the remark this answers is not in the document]
ammar257ammar: I understand, and that what I am doing. You told me to do manual testing and try things from admin beside the automated tests. I was trying to do that.

app/tests/core_tests/test_views.py
```diff
@@ -96,3 +103,71 @@ def test_product_redirect(client):
...
+    [
+        (
+            AccessRequestHandlingOptions.ACCEPT_VERIFIED_USERS,
+            "Your request will be automatically accepted if you verify your account",
+        ),
+        (AccessRequestHandlingOptions.MANUAL_REVIEW, "is awaiting review"),
+    ],
+)
+@pytest.mark.django_db
+def test_registration_request_status_msg(
```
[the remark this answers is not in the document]
ammar257ammar: retracted and updated the parameter on the original test.

app/grandchallenge/verifications/models.py
ammar257ammar: Done

PR #3515, opened 2024-08-29
Disable delete all display sets button when answers exist

### LINKED ISSUE #3502: Delete all displaysets warning is unclear
**Describe the bug**

**To Reproduce**
1. Go to a reader study where all cases are answered
2. Click on Cases
3. Click on `Delete all displaysets`
4. The text `No editable display sets selected.` is displayed. This does not tell the user what they *should*  do to delete all display sets.

**Expected behavior**
The text should tell the user what should be done (first delete all answers).

**Screenshots**

**Additional context**

ammar257ammar:
The message was extended, asking the user to delete all answers before trying to delete the display sets.

closes #3502

### DIFF
--- a/app/grandchallenge/components/templates/components/civ_set_list.html
+++ b/app/grandchallenge/components/templates/components/civ_set_list.html
@@ -39,6 +39,10 @@
                 hx-push-url="{{ base_object.bulk_delete_url }}"
                 hx-vals='{"delete-all": "True"}'
                 hx-target="body"
+                {% if delete_all_disabled_message %}
+                    disabled
+                    title="{{ delete_all_disabled_message }}"
+                {% endif %}
         > Delete all {{ base_object.civ_set_model|verbose_name_plural }}</button>
     </p>
 
--- a/app/grandchallenge/reader_studies/views.py
+++ b/app/grandchallenge/reader_studies/views.py
@@ -404,6 +404,18 @@ class ReaderStudyDisplaySetList(CivSetListView):
     def base_object(self):
         return get_object_or_404(ReaderStudy, slug=self.kwargs["slug"])
 
+    def get_context_data(self, *args, **kwargs):
+        context = super().get_context_data(*args, **kwargs)
+
+        if Answer.objects.filter(
+            question__reader_study=self.base_object
+        ).exists():
+            context["delete_all_disabled_message"] = (
+                "Cannot delete all display sets: first you need to delete all of the answers for this reader study"
+            )
+
+        return context
+
     def get_queryset(self):
         queryset = super().get_queryset()
         return (
--- a/app/tests/reader_studies_tests/test_views.py
+++ b/app/tests/reader_studies_tests/test_views.py
@@ -791,25 +791,11 @@ def test_display_set_bulk_delete_permissions(client):
 
 @pytest.mark.django_db
 def test_display_set_delete_all_button_disabled(client):
-
     editor = UserFactory()
     rs = ReaderStudyFactory()
     rs.add_editor(editor)
 
-    ds = DisplaySetFactory(reader_study=rs)
-
-    q = QuestionFactory(
-        reader_study=rs,
-        question_text="question_text",
-        answer_type=Question.AnswerType.TEXT,
-    )
-
-    AnswerFactory(
-        creator=editor,
-        question=q,
-        answer="question_answer",
-        display_set=ds,
-    )
+    AnswerFactory(question__reader_study=rs)
 
     response = get_view_for_user(
         client=client,
--- a/app/tests/reader_studies_tests/test_views.py
+++ b/app/tests/reader_studies_tests/test_views.py
@@ -787,3 +787,52 @@ def test_display_set_bulk_delete_permissions(client):
         list(response.context["form"].fields["civ_sets_to_delete"].queryset)
         == []
     )
+
+
+@pytest.mark.django_db
+def test_display_set_delete_all_button_disabled(client):
+
+    editor = UserFactory()
+    rs = ReaderStudyFactory()
+    rs.add_editor(editor)
+
+    ds = DisplaySetFactory(reader_study=rs)
+
+    response = get_view_for_user(
+        client=client,
+        viewname="reader-studies:display_sets",
+        reverse_kwargs={"slug": rs.slug},
+        user=editor,
+    )
+
+    assert response.status_code == 200
+    assert (
+        "Cannot delete all display sets: first you need to delete all of the answers for this reader study"
+        not in str(response.rendered_content)
+    )
+
+    q = QuestionFactory(
+        reader_study=rs,
+        question_text="question_text",
+        answer_type=Question.AnswerType.TEXT,
+    )
+
+    AnswerFactory(
+        creator=editor,
+        question=q,
+        answer="question_answer",
+        display_set=ds,
+    )
+
+    response = get_view_for_user(
+        client=client,
+        viewname="reader-studies:display_sets",
+        reverse_kwargs={"slug": rs.slug},
+        user=editor,
+    )
+
+    assert response.status_code == 200
+    assert (
+        "Cannot delete all display sets: first you need to delete all of the answers for this reader study"
+        in str(response.rendered_content)
+    )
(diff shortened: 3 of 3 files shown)


app/grandchallenge/components/templates/components/civ_set_delete_confirm.html
```diff
@@ -13,7 +13,11 @@
 {% block content %}
     <h1>Delete {{ base_object.title }} {{ base_object.civ_set_model|verbose_name_plural|title }}</h1>
     {% if delete_count == 0 %}
-        <p>No editable {{ base_object.civ_set_model|verbose_name_plural }} selected.</p>
+        <p>No editable {{ base_object.civ_set_model|verbose_name_plural }} selected.
+        {% if base_object.civ_set_model|verbose_name_plural == "display sets" %}
```
jmsmkn, line 17: The verbose name should not be used for matching as that is easily changed. Does this error message even make sense? How does the user get here? Shouldn't the delete all button be disabled if it cannot be used?

ammar257ammar: The original issue is that it is not clear what needs to be done to delete all display sets (after clicking the delete all button and seeing the message "No editable display sets are selected". 
I think disabling the delete all button while also the display sets are non-selectable would make it even more vague to the user on what should be done.

jmsmkn: Why not disable the button and add a tooltip telling them what they need to do to be able to click the button? https://getbootstrap.com/docs/4.1/components/tooltips/#disabled-elements

ammar257ammar: Done

app/tests/archives_tests/test_views.py
```diff
@@ -1335,7 +1335,7 @@ def make_request():
 
     make_request()  # set up caches
 
-    expected_queries = 32
+    expected_queries = 38
```
jmsmkn, line 1338: This test catches inefficient implementations, the change needs to be reverted.

ammar257ammar: I reverted the update of the failed test and optimized the implementation according to your suggestions.

app/grandchallenge/components/views.py
```diff
@@ -373,6 +373,13 @@ def get_context_data(self, *args, object_list=None, **kwargs):
                 "base_object": self.base_object,
                 "base_model_options": BaseModelOptions,
                 "request": self.request,
+                "editable_objects_count": len(
+                    [
+                        ds
+                        for ds in self.get_queryset().all()
+                        if ds.is_editable is True
+                    ]
+                ),
```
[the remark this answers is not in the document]
ammar257ammar: I rewrote the implementation according to your suggestions.

app/grandchallenge/components/views.py
```diff
@@ -330,6 +330,7 @@ class CivSetListView(
 ):
     model = None
     permission_required = None
+    delete_all_enabled = True
```
jmsmkn, line 333: Why do you need to set this as class variable?

app/grandchallenge/components/templates/components/civ_set_list.html
```diff
@@ -39,6 +39,10 @@ <h2>{{ base_object.title }} {{ base_object.civ_set_model|verbose_name_plural|tit
                 hx-push-url="{{ base_object.bulk_delete_url }}"
                 hx-vals='{"delete-all": "True"}'
                 hx-target="body"
+                {% if delete_all_enabled is not True %}
```
jmsmkn, line 42: `delete_all_enabled` does not exist in the context for non-reader studies...

app/grandchallenge/reader_studies/views.py
```diff
@@ -404,6 +404,21 @@ def columns(self):
     def base_object(self):
         return get_object_or_404(ReaderStudy, slug=self.kwargs["slug"])
 
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        self.delete_all_enabled = all(
+            not ds.answers.exists() for ds in self.get_queryset()
+        )
```
jmsmkn, line 412: Why do all of that work in Python? You can do this directly with `not self.base_object.answers.exists()` or `not Answer.objects.filter(reader_study=self.base_object).exists()`?

app/grandchallenge/components/templates/components/civ_set_list.html
jmsmkn: For any of the usages of this file. This PR would again break existing implementations. Again, test everything that you touch.

app/grandchallenge/components/views.py
ammar257ammar: I saw it as a static configuration (always True) on the base view class level `CivSetListView` so I made as a class variable. Looking at again, and knowing that its override will be dynamic and involves a database query (in `ReaderStudyDisplaySetList`) I think it should be a property (see below). What do you think?

Screenshot 2024-09-03 at 13 33 58

What I am not sure about is, should we check for the delete_all_enabled and add the message to the context only when it will be viewed, or it is okay to add anyway and just do the check in the template?

app/grandchallenge/components/templates/components/civ_set_list.html
ammar257ammar: Done

app/grandchallenge/reader_studies/views.py
ammar257ammar: Nice way with ORM alone! I did not think of it this way (reaching the reader study from the question).
I will get better with using those connections through object relations.

app/grandchallenge/components/views.py
jmsmkn: You're making this far too complicated. Most straightforward I can think of is `app/grandchallenge/reader_studies/views.py`:

```python
def get_context_data(self, *args, **kwargs):
    context = super().get_context_data(*args, **kwargs)
    
    if Answer.objects.filter(question__reader_study=self.base_object).exists():
        context["delete_all_disabled_message"] = "Cannot delete all display sets: first you need to delete all of the answers for this reader study"

    return context
```

in `app/grandchallenge/components/templates/components/civ_set_list.html`:

```
{% if delete_all_disabled_message %}
   disabled
   title="{{ delete_all_disabled_message }}"
{% endif %}
```

ammar257ammar:
To test the status of the `Disable all` button, I wrote a view test that checks the view of display sets before and after creating an answer to the reader_study question. The button disabled message should not appear if no answers exist for the reader study and should appear otherwise. 
@jmsmkn Is that enough to test this functionality change? [commit](https://github.com/comic/grand-challenge.org/pull/3515/commits/abcfaebbb7d388dddc27adbb0c1d83c109446fb4)

app/tests/reader_studies_tests/test_views.py
```diff
@@ -787,3 +787,52 @@ def test_display_set_bulk_delete_permissions(client):
...
+        question_text="question_text",
+        answer_type=Question.AnswerType.TEXT,
+    )
+
+    AnswerFactory(
+        creator=editor,
+        question=q,
+        answer="question_answer",
+        display_set=ds,
+    )
```
jmsmkn, line 825: We only need to test if answers are created for the reader study:

```suggestion
    AnswerFactory(question__reader_study=rs)
```

app/tests/reader_studies_tests/test_views.py
```diff
@@ -787,3 +787,52 @@ def test_display_set_bulk_delete_permissions(client):
...
+    response = get_view_for_user(
+        client=client,
+        viewname="reader-studies:display_sets",
+        reverse_kwargs={"slug": rs.slug},
+        user=editor,
+    )
+
+    assert response.status_code == 200
+    assert (
+        "Cannot delete all display sets: first you need to delete all of the answers for this reader study"
```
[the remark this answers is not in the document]
ammar257ammar: In the first request, it checks that the button is not disabled (by checking that the disabled message does not exist), but I agree with your point where it is easy to forget changing this message.

Should I test for the non-existence of the `disabled` attribute of the button then? or drop this check from the test entirely?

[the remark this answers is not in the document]
ammar257ammar: Done

app/tests/reader_studies_tests/test_views.py
ammar257ammar: Done

jmsmkn (approved):

PR #3526, opened 2024-09-05
Challenge-Participant Registration Questions - 1/2

chrisvanrun:
This PR is part of the pitch:
- https://github.com/DIAGNijmegen/rse-roadmap/issues/344

It introduces the creation, updating and deletion of registration questions. It also has some validation on the level of models for when answering is introduced.

To make things reviewable it does NOT contain the actual form changes that will show the `RegistrationQuestion` and create the `RegistrationQuestionAnswer` during sign-up to a challenge. It also does not show the answers anywhere. That will be a follow-up PR.

## A quick preview:

Note: the second deletion had associated answers with it.

https://github.com/user-attachments/assets/77a6b790-09c3-427e-b02b-6043011c4167


app/grandchallenge/participants/models.py
```diff
@@ -54,3 +56,88 @@ def delete(self, *args, **kwargs):
...
+    question_help_text = models.TextField(blank=True)
+
+    schema = models.JSONField(
+        default=string_type_schema,
+        help_text="A JSON schema definition against which an answer is validated. See https://json-schema.org/."
+        "Only Draft 7, 6, 4 or 3 are supported.",
+        validators=[JSONSchemaValidator()],
+    )
+
+    required = models.BooleanField(default=True)
```
jmsmkn, line 81: Maybe `question_text`, `question_help_text` and `required` should have help text?

app/grandchallenge/participants/models.py
```diff
@@ -54,3 +56,88 @@ def delete(self, *args, **kwargs):
...
+    challenge = models.ForeignKey(
+        Challenge,
+        on_delete=models.CASCADE,
+    )
+
+    question_text = models.TextField(blank=False)
+    question_help_text = models.TextField(blank=True)
+
+    schema = models.JSONField(
+        default=string_type_schema,
```
jmsmkn, line 75: Does a default of string make sense? Will we have to coerse answers later on?

app/grandchallenge/participants/models.py
```diff
@@ -54,3 +56,88 @@ def delete(self, *args, **kwargs):
...
+        if (
+            type(self)
+            .objects.filter(
+                challenge=self.challenge,
+                question_text=self.question_text,
+            )
+            .exclude(pk=self.pk)
+            .exists()
+        ):
+            raise ValidationError(
```
jmsmkn, line 104: Isn't this anyway raised in `full_clean` by the `unique_together` constraint?

app/grandchallenge/participants/models.py
```diff
@@ -54,3 +56,88 @@ def delete(self, *args, **kwargs):
...
+        unique_together = (("question_text", "challenge"),)
+
+    def clean(self):
+        super().clean()
+
+        if (
+            type(self)
+            .objects.filter(
+                challenge=self.challenge,
+                question_text=self.question_text,
```
jmsmkn, line 99: This is going to be case sensitive, is that sufficient?

app/grandchallenge/participants/models.py
```diff
@@ -54,3 +56,88 @@ def delete(self, *args, **kwargs):
...
+
+
+def string_type_schema():
+    return {"type": "string"}
+
+
+class RegistrationQuestion(UUIDModel):
+    challenge = models.ForeignKey(
+        Challenge,
+        on_delete=models.CASCADE,
```
jmsmkn, line 68: Maybe set `related_name="registration_questions"`? `registration_questions=...` is a nicer interface than `registrationquestion_set=...`.

app/grandchallenge/participants/models.py
```diff
@@ -54,3 +56,88 @@ def delete(self, *args, **kwargs):
...
+
+class RegistrationQuestionAnswer(models.Model):
+    registration_request = models.ForeignKey(
+        RegistrationRequest,
+        on_delete=models.CASCADE,
+    )
+    question = models.ForeignKey(
+        RegistrationQuestion,
+        on_delete=models.CASCADE,
+    )
```
jmsmkn, line 117: Both of these could have `related_name`s. I don't know if `answers` is too generic or not, probably `registration_request_answers` to distinguish it from reader studies.

app/grandchallenge/participants/models.py
```diff
@@ -54,3 +56,88 @@ def delete(self, *args, **kwargs):
...
+
+    answer = models.JSONField(blank=True, editable=False)
```
jmsmkn, line 119: Is making this non-editable correct, you want the user to provide an answer so it should be editable (e.g. usable in forms), no?

app/grandchallenge/participants/models.py
```diff
@@ -54,3 +56,88 @@ def delete(self, *args, **kwargs):
...
+
+    class Meta:
+        unique_together = (("registration_request", "question"),)
+
+    @property
+    def answered(self):
+        return not (isinstance(self.answer, str) and len(self.answer) == 0)
```
jmsmkn, line 126: Shouldn't the existence of this object be enough to determine if it's answered or not?

app/grandchallenge/participants/forms.py
```diff
@@ -0,0 +1,30 @@
...
+            "schema": JSONEditorWidget(),
+        }
+
+    def __init__(self, *args, challenge, **kwargs):
+        super().__init__(*args, **kwargs)
+        if (
+            not hasattr(self.instance, "challenge")
+            or not self.instance.challenge
+        ):
+            self.instance.challenge = challenge
```
jmsmkn, line 30: Seems a bit of a strange construction. What if the challenges differ?

app/grandchallenge/participants/models.py
```diff
@@ -54,3 +56,88 @@ def delete(self, *args, **kwargs):
...
+            )
+            .exclude(pk=self.pk)
+            .exists()
+        ):
+            raise ValidationError(
+                f"There is already an existing {self._meta.model._meta.verbose_name} with this question text"
+            )
+
+
+class RegistrationQuestionAnswer(models.Model):
```
jmsmkn, line 109: We don't have any answering views yet. Does it make sense to include this model in this PR? There is also no admin view for this. Given it is not used I would say to wait until you see how the integration goes.

app/grandchallenge/participants/urls.py
```diff
@@ -26,4 +30,24 @@
...
+        RegistrationQuestionCreate.as_view(),
+        name="registration-question-create",
+    ),
+    path(
+        "registration/questions/<uuid:pk>/update",
+        RegistrationQuestionUpdate.as_view(),
+        name="registration-question-update",
+    ),
+    path(
+        "registration/questions/<uuid:pk>/delete",
```
jmsmkn, line 49: Missing trailing slashes for 3 views.

app/grandchallenge/participants/views.py
```diff
@@ -107,3 +111,74 @@ def get_success_url(self):
...
+    permission_required = "change_challenge"
+    raise_exception = True
+    login_url = reverse_lazy("account_login")
+
+    def get_permission_object(self):
+        return self.request.challenge
+
+    def get_queryset(self):
+        queryset = super().get_queryset()
+        queryset = queryset.filter(Q(challenge=self.request.challenge))
```
jmsmkn, line 129: Using `Q` here is unnecessary.

app/grandchallenge/participants/views.py
```diff
@@ -107,3 +111,74 @@ def get_success_url(self):
             "participants:registration-list",
             kwargs={"challenge_short_name": self.object.challenge.short_name},
         )
+
+
+class RegistrationQuestionList(
+    LoginRequiredMixin, ObjectPermissionRequiredMixin, ListView
+):
+    model = RegistrationQuestion
+    permission_required = "change_challenge"
```
jmsmkn, line 120: It is safer to use the existing `view_registrationquestion` object permission, and assign this to the challenge editors group when the question is created.

app/grandchallenge/participants/views.py
```diff
@@ -107,3 +111,74 @@ def get_success_url(self):
...
+
+
+class RegistrationQuestionMixin(
+    LoginRequiredMixin,
+    SuccessMessageMixin,
+    ObjectPermissionRequiredMixin,
+):
+    model = RegistrationQuestion
+
+    permission_required = "change_challenge"
```
jmsmkn, line 140: You have a security bug here, another challenge admin could Update or Delete questions from other challenges if they knew the UUIDs as you're only inspecting the request and not the objects that you're editing. 

It is safer to:

- On the challenge: create an `add_registration_question` object permission on the challenge, assign that to the editors group.
- For the update and delete views use the existing `change_registrationquestion` and `delete_registrationquestion` object based permissions. Creating new tables from `UserObjectPermissionBase` and `GroupObjectPermissionBase` would be required.

Honestly I do not think a mixin is worth it here.

app/grandchallenge/pages/templates/pages/challenge_settings_base.html
```diff
@@ -29,6 +29,11 @@
                     href="{% url 'challenge-update' challenge_short_name=challenge.short_name %}">
                       <i class="fas fa-cog fa-fw"></i> General Settings</a>
               </li>
+              <li class="nav-item">
+                <a class="nav-link px-4 py-1 mb-1 {% if request.resolver_match.view_name == 'participants:registration-question-list' or request.resolver_match.view_name == 'participants:registration-question-create' or request.resolver_match.view_name == 'participants:registration-question-delete'  %}active{% endif %}"
+                  href="{% url 'participants:registration-question-list' challenge_short_name=challenge.short_name %}">
+                    <i class="fas fa-question-circle fa-fw"></i> Registration Questions </a>
+              </li>
```
jmsmkn, line 36: Maybe use `{% comment %}` for now to hide it whilst the feature is incomplete?

app/grandchallenge/participants/forms.py
chrisvanrun: It is strange. Reusing the form for both creating and updating. I've it to be more clearly a check to only update the challenge if it is being created.

If they differ: the model validation should pick that up.

app/grandchallenge/participants/models.py
chrisvanrun: Requirements state that the validation should default to a string by default. I thought the cleanest way was to add a string schema.

I am unsure what you mean with answer cooercion later on. The actual HTML inputs (and hence POST) will contain only strings and hence will need to be coerced into pythonic objects before fed to this database field.

app/grandchallenge/participants/models.py
chrisvanrun: I think so, it's primarily to prevent people from differentiating one question to the next. Having only a case difference is hard to spot, but still different enough for this goal.

app/grandchallenge/participants/models.py
chrisvanrun: Huh, good point. I remember going expecting that but my practical tests showed: 
image

So I thought it was more of a 'you better not try this via shell' sort of thing.

Turns out that the form saving does not care about any uniqueness constraints for non-included fields. One way to address this, just tested, would be to add a hidden challenge field and have it hidden. Also cleans up the weird instance challenge setting. So I'll follow that route.

jmsmkn: Ah, yes, hidden challenge field is definitely the way to do it.

app/grandchallenge/participants/models.py
chrisvanrun: You are right, this is a missnomen. Should be a check for an empty answer.

app/grandchallenge/participants/models.py
chrisvanrun: Not sure, we have a single request per user+challenge that can either be rejected or accepted. I sort of think that it should not be editable, but I can also forsee a future where they are. There isn't any path planned in this pitch to allow for it to be editable. Might as well save us a migration down the road. I'll make it editable.

app/grandchallenge/pages/templates/pages/challenge_settings_base.html
chrisvanrun: Good idea! Sort of an ad hoc, but very easy to remove feature flag.

app/grandchallenge/participants/models.py
jmsmkn: Unresolving this as I think you misunderstand a concept in Django still. Please check the meaning of "editable" in Django https://docs.djangoproject.com/en/5.1/ref/models/fields/#editable

> If False, the field will not be displayed in the admin or any other ModelForm. They are also skipped during model validation. Default is True.

It means that you cannot use it in forms, even the creation one, so the user _would not_ be able to provide an answer.

jmsmkn: btw, this I think is the danger of including new models in a PR that are not fully hooked up yet.

app/grandchallenge/participants/views.py
chrisvanrun: I sort of like the usage of:
```Python
    def get_permission_object(self):
        return self.request.challenge
```

When checking `"challenges.add_registration_question"` permission against the challenge itself in the creation view. Since you propose to remove (or ban) the use of the function. I agree that it smells. What would be an acceptable alternative? Implement the check ourselves in the `dispatch()`, directly? As follows:

```Python
def dispatch(self, request, *args, **kwargs):
        if not self.request.user.has_perm(
            "challenges.add_registration_question",
            self.request.challenge,
        ):
            raise PermissionDenied
        return super().dispatch(request, *args, **kwargs)
```

chrisvanrun: Maybe the better approach would be to ban the usage of the function `get_permission_object` with an exception for `CreateView` derivatives.

chrisvanrun:
I've removed the Answer model. It will make rebasing the follow-up PR a bit more difficult 🤷🏻. I've opted not to add any `'disabled'` additions to the create/edit/remove controls in the list view since in practice they should never be disabled. (Plus, the views they link to are safeguarded by permissions.)

Currently, waiting for the tests to run through before the re-review request.

chrisvanrun:
Grmbl, this is irritating. Local tests seem to pass, but CI does not agree. Fun! Scrubbing docker volumes and re-running tests locally.

app/grandchallenge/participants/views.py
jmsmkn: Yes it is necessary for the create view as you need an object to test against, in this case returning the challenge is correct but the permission should be "add_registration_question".

app/grandchallenge/participants/forms.py
```diff
@@ -0,0 +1,30 @@
...
+            "question_text": TextInput,
+            "question_help_text": TextInput,
+            "schema": JSONEditorWidget(),
+        }
+
+    def __init__(self, *args, challenge=None, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        if challenge:
+            self.fields["challenge"].initial = challenge
```
amickan, line 30: `Challenge` is required for a `RegistrationQuestion`, so why allow it be `None` here?

app/grandchallenge/participants/views.py
```diff
@@ -115,3 +126,103 @@ def get_success_url(self):
...
+
+        if not self.request.user.has_perm(
+            "challenges.change_challenge",
+            self.request.challenge,
+        ):
+            # View contains admin side-panel
+            raise PermissionDenied
+
+        return super().dispatch(request, *args, **kwargs)
+
```
amickan, line 154: You could also use `UserPassesTestMixin` to enforce the extra permission check. Feels more django-y to me.

app/grandchallenge/participants/views.py
```diff
@@ -115,3 +129,103 @@ def get_success_url(self):
...
+    def dispatch(self, request, *args, **kwargs):
+
+        if not self.request.user.has_perm(
+            "challenges.change_challenge",
+            self.request.challenge,
+        ):
+            # View contains admin side-panel
+            raise PermissionDenied
+
+        return super().dispatch(request, *args, **kwargs)
```
jmsmkn, line 156: I think you're getting into a bit of a mess here. Overriding the dispatch method is dangerous, especially with so many mixins. If there is something in the page that requires a different permission then that should be checked in the template.

app/grandchallenge/participants/views.py
```diff
@@ -115,3 +129,103 @@ def get_success_url(self):
...
+):
+    model = RegistrationQuestion
+    form_class = RegistrationQuestionCreateForm
+
+    success_message = "Question successfully created"
+
+    permission_required = "challenges.add_registration_question"
+
+    raise_exception = True
+    login_url = reverse_lazy("account_login")
```
jmsmkn, line 173: Code style: 3 new views, 3 different ways of spacing the attributes, 3 different ways of ordering them. Please be consistent.

app/grandchallenge/participants/views.py
```diff
@@ -115,3 +126,103 @@ def get_success_url(self):
...
+
+
+class RegistrationQuestionDelete(
+    LoginRequiredMixin,
+    SuccessMessageMixin,
+    ObjectPermissionRequiredMixin,
+    DeleteView,
+):
+    model = RegistrationQuestion
+    fields = ()
```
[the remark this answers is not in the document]
jmsmkn: Still unresolved.

app/grandchallenge/participants/templates/participants/registrationquestion_confirm_delete.html
```diff
@@ -0,0 +1,40 @@
...
+
+    <form action="" method="post" disabled>
+        {% csrf_token %}
+        <p>Are you sure you want to delete this question:</p>
+
+        <h5 class="font-weight-bolder pb-4">'{{ object.question_text }}'</h5>
+
+
+
+
```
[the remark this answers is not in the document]
chrisvanrun: Ah yes, removed some Answer related things. Didn't clean up the lines correctly.

app/grandchallenge/participants/models.py
```diff
@@ -54,3 +57,72 @@ def delete(self, *args, **kwargs):
...
+        unique_together = (("question_text", "challenge"),)
+
+    def save(self, *args, **kwargs):
+        adding = self._state.adding
+
+        result = super().save(*args, **kwargs)
+
+        if adding:
+            self.assign_permissions()
+        return result
```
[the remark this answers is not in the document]
chrisvanrun: Good find! Confused this for a moment with the ModelForm.save()

app/grandchallenge/participants/views.py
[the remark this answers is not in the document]
chrisvanrun: `ModelFormMixin` (from `UpdateView`/`CreateView`), which I now realize was because of a copy+paste error when rewriting the views. I had not given the delete the correct base View but it gave me an error. The `DeleteView` has `FormMixin` as a base which does not require the `fields` defined. Removed!

app/grandchallenge/participants/views.py
chrisvanrun: I agree with it being a weird mixin versus custom dispatch overriding. I was hesitant to push it into the template because it broadens the impact of this PR. In short, the challenge settings sidebar will not show unless someone has `"change_challenge"` permissions (see the challenge_settings_base.html template).

If the user `readerstudy` looks at the page `http://demo.gc.localhost:8000/participants/registration/questions/` they will see the following:
image

jmsmkn: That looks fine, so long as the items are filtered. Probably we should solve that they can load the page in the first place systematically for all of the list views but would require double permission handling => another issue.

chrisvanrun: I can imagine that if we add more subtle permissions, sections of the settings would be hidden. Indeed another issue. First this pitch PR!

jmsmkn (approved):

chrisvanrun:
Thanks for the review James! I always appreciate your thoroughness.

jmsmkn:
> Thanks for the review James! I always appreciate your thoroughness.

PR #3528, opened 2024-09-09
Challenge-Participant Registration Questions - 2/2 

chrisvanrun:
This PR is linked to the pitch:
- https://github.com/DIAGNijmegen/rse-roadmap/issues/344

It hooks in showing and getting answers to the questions from:
- https://github.com/comic/grand-challenge.org/pull/3526

Note that `""` is not technically a JSON (`""""` is the empty string in JSON notation). However, having users input everything with double quotes seems very, very silly. As such the `JSONField` is wrapped to correctly parse strings. 

Tests are extensive because of my preference for TDD, but also because registration is a cornerstone of the challenges and should be robust.

PR also does a minor style update on accepting/rejecting participant requests. Making it (IMHO) a bit clearer what the buttons do.

#### Participant POV

Questions

#### Admin POV

list_requests

list_requests_with_modal


chrisvanrun:
Currently working on rebasing and updating this based on the changes in the base PR.

app/grandchallenge/participants/static/participants/js/registrationquestion_list.mjs
```diff
@@ -2,5 +2,12 @@ $(document).ready(() => {
     $("#registrationQuestionsTable").DataTable({
         order: [[0, "asc"]],
         pageLength: 10,
+        columnDefs: [
+            {
+                targets: [3, 4],
+                searchable: false,
```
chrisvanrun, line 8: The list view still had the Edit / Delete searchable

app/grandchallenge/participants/templates/participants/registrationquestion_list.html
```diff
@@ -51,7 +51,7 @@ <h2>Registration Questions for {{ challenge.short_name }}</h2>
                     <td>
                         {{ question.question_text }}
                     </td>
-                    <td>
+                    <td data-order="{{ question.required }}">
```
chrisvanrun, line 54: Could not sort the question list on the required/non-required.

app/grandchallenge/participants/admin.py
```diff
@@ -31,3 +33,10 @@ class RegistrationQuestionAdmin(GuardedModelAdmin):
 admin.site.register(
     RegistrationQuestionGroupObjectPermission, GroupObjectPermissionAdmin
 )
+
+admin.site.register(
```
jmsmkn, line 37: Is `RegistrationQuestionAnswer` missing from the admin?

app/grandchallenge/participants/migrations/0004_alter_registrationquestion_options.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("participants", "0003_registrationquestionanswer_and_more"),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
```
jmsmkn, line 13: Delete both migrations and compress down to one.

app/grandchallenge/participants/models.py
```diff
@@ -123,3 +125,81 @@ class RegistrationQuestionGroupObjectPermission(GroupObjectPermissionBase):
...
+            raise ValidationError(
+                f"The question {self.question.question_text!r} requires an answer"
+            )
+
+        if not self.empty_answer and self.question.schema:
+            try:
+                JSONValidator(schema=self.question.schema)(value=self.answer)
+            except ValidationError as e:
+                raise ValidationError(
+                    {"answer": f"Incorrect format: {e.__cause__}"}
```
jmsmkn, line 163: Seems a strange thing to access. What is the issue with just using `f"... {e}"`? That should be human readable.

app/grandchallenge/participants/models.py
```diff
@@ -123,3 +125,81 @@ class RegistrationQuestionGroupObjectPermission(GroupObjectPermissionBase):
...
+                )
+
+    def validate_constraints(self, exclude=None):
```
jmsmkn, line 166: May be better to use `*args, **kwargs` incase the function signature changes upstream.

app/grandchallenge/participants/views.py
```diff
@@ -1,21 +1,26 @@
+from collections import defaultdict
+from typing import Any
```
jmsmkn, line 2: We do not run mypy on GC so no typing please.

app/grandchallenge/participants/views.py
```diff
@@ -94,8 +98,48 @@ def get_queryset(self):
...
+
+        answers = self._get_registration_questions_answers(questions=questions)
+
+        answer_lookup = defaultdict(dict)
+        for answer in answers:
+            answer_lookup[answer.registration_request.pk][
+                answer.question.pk
+            ] = answer
+
+        context_data["registration_questions_answer_lookup"] = answer_lookup
```
jmsmkn, line 118: This feels like the wrong approach. Why isn't a `prefetch_related` being used and the answers then being accessed via `RegistrationRequest.questions.answers`?

app/grandchallenge/participants/views.py
```diff
@@ -94,8 +98,48 @@ def get_queryset(self):
...
+            registration_request__in=self.object_list,
+            question__in=questions,
+        ).select_related("registration_request", "question")
+
+        return filter_by_permission(
+            queryset=queryset,
+            user=self.request.user,
+            codename="view_registrationquestionanswer",
+            accept_user_perms=False,
+        )
```
jmsmkn, line 141: This feels like a little bit of overkill. If the user can see the request they should be able to see the questions and answers anyway right?

app/grandchallenge/participants/models.py
```diff
@@ -123,3 +125,81 @@ class RegistrationQuestionGroupObjectPermission(GroupObjectPermissionBase):
...
+    def save(self, *args, **kwargs):
+        adding = self._state.adding
+        super().save(*args, **kwargs)
+        if adding:
+            self.assign_permissions()
+
+    def assign_permissions(self):
+        # Admins can view an answer
+        assign_perm(
+            "view_registrationquestionanswer",
```
jmsmkn, line 188: Unsure if this is totally necessary as this is never in a separate view.

app/grandchallenge/participants/views.py
```diff
@@ -43,7 +48,8 @@ class RegistrationRequestCreate(
     LoginRequiredMixin, SuccessMessageMixin, CreateView
 ):
     model = RegistrationRequest
-    fields = ()
+    form_class = RegistrationRequestForm
+
```
jmsmkn, line 52:
```suggestion
```

app/grandchallenge/participants/forms.py
```diff
@@ -1,8 +1,86 @@
...
+        self.fields["user"].disabled = True
+        self.fields["challenge"].initial = challenge
+        self.fields["challenge"].disabled = True
+
+        self._registration_questions = RegistrationQuestion.objects.filter(
+            challenge=challenge
+        )
+
+        for q in self._registration_questions:
+            self.fields[str(q.pk)] = RegistrationQuestionField(
```
jmsmkn, line 40: Should we have a better field name than just the pk? `registration_question_{q.pk}` maybe? Could also make it a model property for re-use if necessary.

app/grandchallenge/participants/forms.py
```diff
@@ -1,8 +1,86 @@
...
+                answer.full_clean()
+                answer.save()
+        except Exception as e:
+            self.instance.delete()  # Also deletes already created answers
+            raise e
+
+    def save(self, *args, **kwargs):
+        result = super().save(self, *args, **kwargs)
+        on_commit(self._save_questions)
+        return result
```
jmsmkn, line 83: This construction is a bit awkward, you are using one form to create multiple objects. Unsure on the solution, but I think Django has something built in for this: [formsets](https://docs.djangoproject.com/en/5.1/topics/forms/formsets/). You would then have one form per answer which would be individually validated.

app/grandchallenge/participants/form_fields.py
```diff
@@ -0,0 +1,36 @@
...
+        super().__init__(widget=widget, **kwargs)
+
+    def to_python(self, value):
+        if value is None or value == "":
+            return self.empty_value
+        try:
+            return super().to_python(value)
+        except ValidationError:
+            # We assume it is a string
+            return value
```
jmsmkn, line 36: I don't really understand this block.

app/tests/participants_tests/test_form_fields.py
```diff
@@ -0,0 +1,45 @@
...
+def test_registration_field():
+    rq = RegistrationQuestion(
+        question_text="Foo",
+        question_help_text="<script></script>",
+        required=False,
+    )
+    field = RegistrationQuestionField(registration_question=rq)
+
+    assert isinstance(
+        field.help_text, SafeString
```
jmsmkn, line 17: Checking that it is safe is insufficient, it should be escaped!

app/grandchallenge/participants/views.py
jmsmkn: See https://mypy.readthedocs.io/en/stable/existing_code.html

> If your codebase is large, pick a subset of your codebase (say, 5,000 to 50,000 lines) and get mypy to run successfully only on this subset at first, _before adding annotations_.

app/grandchallenge/participants/admin.py
chrisvanrun: It is, do we need to have access to it in the admin? I don't forsee us ever having to access it.

If we do need it, how is access to the RSEs handled? Is that done live in production?

jmsmkn: It's good practise to register every model. Access is then defined by group permissions.

app/grandchallenge/participants/models.py
chrisvanrun: Currently, the constructs are such that `'change_challenge'`, `'add_registration_question'` and `'view_registrationquestionanswer'` perfectly overlap with the `challenge.admins_group`.

Since the `reader_studies.models.Answer` follows the same pattern and `'change_challenge'` vs`'add_registration_question'` is something we want, I don't think it is overkill to cleanly add the permissions here and now.

jmsmkn: See the above comment about the usage of these things. They are entirely intertwined, are never used independently, so this is unnecessary.

app/grandchallenge/participants/form_fields.py
chrisvanrun: Yeah, it had my head in a twist for a bit as well. Most (all?) input widgets produce a pure string value.

If one would want to input the empty string you'd have to type:
`""` into the HTML input. A regular string would be `"Hello world"` instead of `Hello world`.

That is:
`json.loads("Hello world")` vs `json.loads("\"Hello world\"")`.

This block allows the field to interpret `Hello world` as a valid string input.

app/grandchallenge/participants/models.py
chrisvanrun: Agreed!

app/grandchallenge/participants/models.py
chrisvanrun: Since this isn't quite as core as the `save()` I am not sure: should a `return result` also be added? I've added it: better be save (hehe).

app/grandchallenge/participants/views.py
chrisvanrun: Check! I think this was a lazy copy-pasta, sorry!

app/grandchallenge/participants/form_fields.py
jmsmkn: What ValidationError would we expect to catch here? Is there a code attached to it? Should we check that the value is a string rather than assuming it?

chrisvanrun: Looking at the problem at hand closer, I also added some tests which had me in turn make the to_python more complex to handle double-quotation marks in different numbers, locations etc.

app/grandchallenge/participants/models.py
chrisvanrun: Removed and simplified.

app/grandchallenge/participants/forms.py
chrisvanrun: It is awkward; I get the sense that generally the practice is to have a Form per object/model. I'll see if I can introduce formsets to make that happen. It did take a non-trivial amount of tinkering to get current form to behave as it should, we'll see.

jmsmkn: Yeah, I'm not sure if it is the solution. We only have them in one other place in the codebase, and there you are adding multiple of the same objects to one other, here the number would be fixed. Good to research this.

jmsmkn: See `QuestionForm` and the `option_form_set`. Happy to walk you through that.

chrisvanrun: From what I understand, the main objection was that a single `Form` created multiple objects. 

The way that the `QuestionForm` goes about it is that it is still within the `Form`, creating multiple objects. The only difference with the current construct in this PR is that you'd put a `FormSet -> Form` as an extra layer. 

I am currently exploring the option of having the `View` hold both a form and a modelformset, which comes with slightly different set 'tee' functions where you call things both the form and the modelformset.

Thoughts?

jmsmkn: I think you misunderstand the concept.

Now you are validating and saving Answer objects using a Question form.

Formsets move validation of individual objects to their own forms. So, you would have a formset of answer forms, each answer form is responsible for validating and saving its answer.

You still need to hookup the validation and saving of the formset from the parent form - this is where the link to the registration request comes in of course - but it means the validation and saving of answers is pushed down to an answer form which is where it should be.

jmsmkn: Views shouldn't mess about with forms, views should be extremely simple - does the user have permission? Then hook them up with a form and template. Redirect them.

chrisvanrun: Ah. So it's more about having a dedicated `Form` perform the saving per object. Sure, adding some boilerplate that pushes the field and validation into a Form works, and using a formal formset is reasonably trivial.

chrisvanrun: I refactored the form to use the fromsets. Using them is a reasonable fit. Had to make two 'fixes':
- Force the use of the required attribute, this only plays a part for formats when you don't want to create objects for empty forms. Having a 'required' attribute in the input would prevent posting empty forms.
- Force saving of empty forms, empty answers should be explicit and not implicit by missing answers.

app/grandchallenge/participants/forms.py
chrisvanrun: N/A

app/grandchallenge/participants/views.py
chrisvanrun: Done!

app/grandchallenge/participants/form_fields.py
chrisvanrun: Resolved, I think. Leaving it up to you James.

app/grandchallenge/participants/forms.py
```diff
@@ -1,8 +1,110 @@
...
+            self.add_error(field=None, error=error)
+
+    def save(self, *args, **kwargs):
+        registration_request = super().save(*args, **kwargs)
+
+        # By default, formsets only save changed or non-empty
+        # forms. Short-circuit saving to ensure it saves empty
+        # answers
+        for form in self.answer_formset.forms:
+            self.answer_formset.save_new(form)
```
jmsmkn, line 83: This seems a bit iffy. Doesn't Django skip validation for empty forms? Can you force that validation and for the formset to be considered? Can you set `empty_permitted` on the form instead?

app/grandchallenge/participants/forms.py
```diff
@@ -1,8 +1,110 @@
...
+    class Meta:
+        model = RegistrationQuestionAnswer
+        fields = ("answer",)
+
+    def __init__(self, *args, initial, **kwargs):
+
+        # Formsets disabled the use of the attribute by default,
+        # since we don't allow for 'extra' forms
+        # we can forcefully allow it.
+        kwargs["use_required_attribute"] = True
```
jmsmkn, line 99: I think this then assumes that this form will only be used in a formset. Can you pass it as a kwarg only where necessary? See `form_kwargs` in:

```python
self.options_form_set = options_form_set_factory(
            instance=self.instance,
            data=kwargs.get("data"),
            form_kwargs={"is_editable": self.instance.is_fully_editable},
        )
```

app/tests/participants_tests/test_forms.py
```diff
@@ -0,0 +1,173 @@
...
+    return {
+        "registration_question_answers-TOTAL_FORMS": str(n),
+        "registration_question_answers-INITIAL_FORMS": "0",
+        "registration_question_answers-MIN_NUM_FORMS": str(n),
+        "registration_question_answers-MAX_NUM_FORMS": str(n),
+        **{
+            f"registration_question_answers-{i}-answer": a
+            for i, a in enumerate(args)
+        },
+    }
```
jmsmkn, line 27: This is a bit too smart for a test. Easier for future programmers to work with explicit data.

chrisvanrun: I think we are in different camps with this. I'm more of the: single place to change the interface to the form. But I'll bend to the common denom in the code base.

app/grandchallenge/participants/forms.py
chrisvanrun: Good suggestion! I ignored the form_kwargs when I saw this:

https://github.com/django/django/blob/1857b6663b664847a426e8c79b8aec42994660ba/django/forms/formsets.py#L259-L270

But then I realized it was actually done here, and allows the kwarg forms to overwrite.

https://github.com/django/django/blob/1857b6663b664847a426e8c79b8aec42994660ba/django/forms/formsets.py#L219-L246

Caution for the future: the `form_kwargs` route only works when you do not allow new empty forms.

app/tests/participants_tests/test_forms.py
chrisvanrun: Done!

app/grandchallenge/participants/forms.py
chrisvanrun: Only if `empty_permitted=True` is set, are empty forms not validated. The `BaseForm` has this in its `full_clean`:
```Python
if self.empty_permitted and not self.has_changed():
   return

self._clean_fields()
self._clean_form()
self._post_clean()
```

This is the construct that is being short-circuited, which is called from the formset' `save()`:

```Python
def save_new_objects(self, commit=True):
    self.new_objects = []
    for form in self.extra_forms:
        if not form.has_changed():
            continue
        # If someone has marked an add form for deletion, don't save the
        # object.
        if self.can_delete and self._should_delete_form(form):
            continue
        self.new_objects.append(self.save_new(form, commit=commit))
        if not commit:
            self.saved_forms.append(form)
    return self.new_objects
```

The only other entry to this would be to:
- Short circuit `has_changed()` in the form.
- Not commit
  - Save the `saved_forms`
  - Save the `extra_forms` not in `saved_forms`, via `save_new()`

The first is a clear hack. The latter is equivalent to the current: there are only extra forms.

app/tests/participants_tests/test_forms.py
jmsmkn: I would agree if this were application code, but test code is different. It is much easier to debug if you can see exactly what is going on. If you have some centralised utility in a test, if you modify that then you need to check all of the tests where it is used in case you change any of the assumptions about what the test is testing in the first place. Same reason that a test should have no-to-little logic: setup, mutate, assert.

chrisvanrun: > If you have some centralised utility in a test, if you modify that then you need to check all of the tests where it is used in case you change any of the assumptions about what the test is testing in the first place.

That is a good argument. A lot of 'out of scope' tests might still work when you change them, but you might have accidentally broken a test somewhere. Had not thought about that danger. Thanks!

I'll try and grog it a bit more next time!

chrisvanrun:
Huh, does the CI re-run tests when `main` is updated? Would make sense if it does.... I think it broke a test on this branch. Fixin!

chrisvanrun:
Test fixed. 

I could have sworn that all the `core` tests were locally passing Friday morning. Bit of history of why permission request creation for several apps are tested in the `core` app.

jmsmkn:
They can be refactored to a separate app, definitely some things I would do differently with those now.

jmsmkn:
The modals do not work in Firefox:

Screenshot 2024-09-23 at 14 24 23


Suggestion to add the admin and not mess around with the instance in the form:

```diff
diff --git a/app/grandchallenge/participants/admin.py b/app/grandchallenge/participants/admin.py
index c8dacd586..212438fc7 100644
--- a/app/grandchallenge/participants/admin.py
+++ b/app/grandchallenge/participants/admin.py
@@ -9,9 +9,13 @@ from grandchallenge.participants.models import (
     RegistrationQuestion,
     RegistrationQuestionAnswer,
     RegistrationQuestionGroupObjectPermission,
-    RegistrationQuestionUserObjectPermission,
+    RegistrationQuestionUserObjectPermission, RegistrationRequest,
 )
 
+@admin.register(RegistrationRequest)
+class RegistrationRequestAdmin(GuardedModelAdmin):
+    readonly_fields = ("user", "challenge")
+
 
 @admin.register(RegistrationQuestion)
 class RegistrationQuestionAdmin(GuardedModelAdmin):
diff --git a/app/grandchallenge/participants/forms.py b/app/grandchallenge/participants/forms.py
index c55861470..c5ceed12c 100644
--- a/app/grandchallenge/participants/forms.py
+++ b/app/grandchallenge/participants/forms.py
@@ -95,15 +95,21 @@ class RegistrationQuestionAnswerForm(ModelForm):
 
     class Meta:
         model = RegistrationQuestionAnswer
-        fields = ("answer",)
+        fields = ("answer", "question", )
+        widgets = {
+            "question": HiddenInput(),
+        }
 
     def __init__(self, *args, initial, **kwargs):
         super().__init__(*args, initial=initial, **kwargs)
 
-        self.instance.question = initial["question"]
+        question = initial["question"]
+
+        self.fields["question"].initial = question
+        self.fields["question"].disabled = True
 
         self.fields["answer"] = RegistrationQuestionAnswerField(
-            registration_question=self.instance.question
+            registration_question=question
         )
 
 

```

chrisvanrun:
Huh? It's a pretty standard modal usage. Seems to work fine on my version of Firefox (Windows 11, Firefox 130.0.1):

image

Is there something else that might be causing this?

chrisvanrun:
Thanks for the review James!

PR #3529, opened 2024-09-09
Preload algorithms when reader studies are switched

jmsmkn:
See https://github.com/DIAGNijmegen/rse-roadmap/issues/340


pkcakeout:
I am curious about how this works. Normally, when switching reader studies, a javascript message is fired off in the frontend and only when the message was not received by CIRRUS we would go back to the backend. So I would have assumed that I would at least see some javascript? Something that interacts with this piece of code?:

https://github.com/comic/grand-challenge.org/blob/f42b01a25780c9226547282f9eb6207fd8b0ed6d/app/grandchallenge/workstations/static/workstations/js/session-control.mjs#L64-L70

jmsmkn:
Yes, when typing up my daily summary I realised the drawback of this initial approach, so [suggested moving to a registration based one instead](https://github.com/DIAGNijmegen/rse-roadmap/issues/340#issuecomment-2338389245) which saves messing around with JS calls. Copied here for reference:

Add a reference to the Sessions from the Reader Study and re-work the task to be periodic, keeping the lambda functions warm every 5 minutes. Use the cache to ensure that the preloading does not happen too often.

PR #3533, opened 2024-09-11
Enable defining interactive algorithm on question

amickan:
This PR enables users with `add_interactive_algorithm_to_question` permission to define an interactive algorithm for reader study questions. The selection is restricted and determined by question type. 
It also adds the algorithm to the question serializer.

This is part of https://github.com/DIAGNijmegen/rse-roadmap/issues/340


app/grandchallenge/reader_studies/forms.py
```diff
@@ -345,6 +354,14 @@ def __init__(self, *args, reader_study, **kwargs):
                     f"hx-include='[id=id_answer_type]'>"
                     f"</div>"
                 ),
+                HTML(
+                    f"<div "
+                    f"hx-get={reverse_lazy('reader-studies:question-interactive-algorithms', kwargs={'slug': reader_study.slug})!r} "
```
jmsmkn, line 359: We should be using `format_html` whenever we're templating out HTML, never using f-string substitution.

app/grandchallenge/reader_studies/forms.py
```diff
@@ -412,9 +436,33 @@ def widget_choices(self):
             ) from error
         return choices
 
+    def interactive_algorithm_choices(self):
+        answer_type = self.__get_answer_type()
```
jmsmkn, line 440: Why is this using a double underscored method? Shouldn't this just be a property?

amickan: Not sure why it was implemented this way. I just reused what was there to stay within the same code pattern as for the other dynamic fields on the form, but I agree that making this a property `answer_type` on the form would make sense. Should I go ahead and refactor this?

jmsmkn (commented):
LGTM, just a couple of questions that look like they come from the existing design.

app/grandchallenge/reader_studies/forms.py
jmsmkn: Yes please, no need to use `__get` for a property.

jmsmkn (approved):

PR #3535, opened 2024-09-12
Allow setting minimum credits per job

### LINKED ISSUE #3263: Save used credits on job model, use maximum costs to determine credits and set maximum number of jobs
At the moment the credits per job are set only by the average runtime, `compute_cost_euro_millicents` should be used instead. We should also allow the user to set a maximum number of jobs per month per verified user for each algorithm.

jmsmkn:
This allows the user to set a minimum number of credits per job. The credits per job now takes into account the maximum runtime of the algorithm plus the hardware it is configured for. The number of credits is updated whenever the relevant parameters change.

TODO:
- [ ] There are too many variables involved with how many credits a particular job consumes. It can change based on whether the user selects a GPU or not. Like the complimentary jobs we need to track how many credits are consumed on each job object and sum that when working out the spent credits, and have it as a property of the algorithm job that is added at runtime instead.

Closes #3263


jmsmkn:
Another approach is needed.

PR #3536, opened 2024-09-13
Move celery cleanup task to large queue

amickan:
Related to https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/339


amickan:
> LGTM, let's hope this does not have the same problems.

The task already ran successfully on the queue with the increased memory limit half an hour ago, so moving it to the large queue (which has the same increased memory limit by default) is the way to go.

PR #3541, opened 2024-09-16
Split challenge migration step

chrisvanrun:
Incorrectly assumed that these operations could be combined in a single migration. They cannot: the correct Permissions seem to not exist within the step.


jmsmkn (approved):

chrisvanrun:
Running some ad hoc tests to make sure this handles existing challenges correctly.

jmsmkn:
Same problem, I deleted the migration and will use run assign_permissions against all existing challenges.

chrisvanrun:
> Might not work... somehow the permission does not get created. =/

nvm, rebuilding things locally suddenly had the permission created. /shrug/

chrisvanrun:
Note for future permissions work, untested:

The actual permission creation is being done by a Django build-in `post_migrate` signal. AFAIK that is only called at the end of `migrate`. Whenever we're doing permission migrations we should likely explicitly call the creation of permissions(`create_permissions`, with the app_config via `apps.get_app_config("challenges")`). 

https://docs.djangoproject.com/en/5.1/topics/auth/customizing/#custom-permissions

jmsmkn:
It's [always been dodgy](https://stackoverflow.com/questions/49100963/migrating-to-django-guardian-permissions-with-custom-user-class), much easier to run it as a separate command.

Issue #3537, opened 2024-09-13
Challenge admin: unclear parent form saving


chrisvanrun:
A challenge admin (Topcow24) could not figure out why they could not set a parent phase.

From the report it seems like the 'The parent phase needs to have at least 1 valid archive item.' does not show.

To be investigated.

chrisvanrun:
The error **is** presented but there is something else going on. When failing the validation, it already shows the general warning (danger) for when it is already hooked up. Very confusing.


![image](https://github.com/user-attachments/assets/4c95b234-0021-4796-9bce-6fbbfd4569ba)

chrisvanrun:
Note: this was the second challenge admin that mentioned the `parent` setting was initially confusing for them. When stating out a challenge the error/warning list is long and form validation errors are easily missed.

PR #3543, opened 2024-09-16
Phase parenting warnings

chrisvanrun:
Closes: #3537 

This PR makes the phase parenting warnings a bit less intrusive and the related errors easier to see.

### Before
image

### Updated

- No longer shows the info blob when validation fails for the parent field, incorrectly giving the suggestion it succeded in updating
- Shows the parent-enabled information as info, instead of error (danger!)
- ~Binds the `parent` errors to the field!~

phase_parent_info

### DIFF
--- a/app/grandchallenge/core/templatetags/negate.py
+++ b/app/grandchallenge/core/templatetags/negate.py
@@ -1,8 +1,0 @@
-from django import template
-
-register = template.Library()
-
-
-@register.filter
-def negate(value):
-    return not value
--- a/app/grandchallenge/core/templatetags/negate.py
+++ b/app/grandchallenge/core/templatetags/negate.py
@@ -1,0 +1,8 @@
+from django import template
+
+register = template.Library()
+
+
+@register.filter
+def negate(value):
+    return not value
--- a/app/grandchallenge/evaluation/models.py
+++ b/app/grandchallenge/evaluation/models.py
@@ -624,7 +624,10 @@ class Phase(FieldChangeMixin, HangingProtocolMixin, UUIDModel):
         super().clean()
         self._clean_algorithm_submission_settings()
         self._clean_submission_limits()
-        self._clean_parent_phase()
+        try:
+            self._clean_parent_phase()
+        except ValidationError as e:
+            raise ValidationError({"parent": e})
 
     def _clean_algorithm_submission_settings(self):
         if self.submission_kind == SubmissionKindChoices.ALGORITHM:
--- a/app/grandchallenge/evaluation/models.py
+++ b/app/grandchallenge/evaluation/models.py
@@ -624,10 +624,7 @@ class Phase(FieldChangeMixin, HangingProtocolMixin, UUIDModel):
         super().clean()
         self._clean_algorithm_submission_settings()
         self._clean_submission_limits()
-        try:
-            self._clean_parent_phase()
-        except ValidationError as e:
-            raise ValidationError({"parent": e})
+        self._clean_parent_phase()
 
     def _clean_algorithm_submission_settings(self):
         if self.submission_kind == SubmissionKindChoices.ALGORITHM:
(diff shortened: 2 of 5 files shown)


chrisvanrun:
Quick end-of-the-day fix for some spit and polish.

app/grandchallenge/core/templatetags/negate.py
```diff
@@ -0,0 +1,8 @@
+from django import template
+
+register = template.Library()
+
+
+@register.filter
+def negate(value):
```
jmsmkn, line 7: Why do we need this? I would have thought that doing `not value` in the template would work? Or if you need it as a variable using `with`?

app/grandchallenge/evaluation/templates/evaluation/partials/phase_admin_warnings.html
```diff
@@ -22,9 +22,9 @@
     </div>
 {% endif %}
 
-{% if phase.parent %}
-    <div class="alert alert-danger" role="alert">
-        <i class="fa fa-exclamation-triangle" aria-hidden="true"></i>
+{% if phase.parent and not hide_phase_parent_info %}
```
jmsmkn, line 25: Why not have it as `{% if phase.parent and show_phase_parent_info %}` to avoid the double negatives?

app/grandchallenge/core/templatetags/negate.py
chrisvanrun: I tried both, and there does not seem to be a construct that works. So, I resorted to this.

app/grandchallenge/evaluation/templates/evaluation/partials/phase_admin_warnings.html
chrisvanrun: The double negative allowed it to be shown by default since non-existing context variables are false. IIRC the partial template is in use by several (5-7?) other templates. Between the double negative versus repeating a `show_phase_parent_info` many times, I opted for the first.

However, I just found the filter `default_if_none` which makes for a slightly cleaner approach to this.

app/grandchallenge/core/templatetags/negate.py
chrisvanrun: Quickly double checked and yeah, none of these work:

`{% with hide=not form.initial.parent %}`
`{% with not form.initial.parent as hide %}`

I think (but we no longe need it) we could have used `|default:True` instead. Which isn't much easier to read imho.

app/grandchallenge/evaluation/models.py
```diff
@@ -624,7 +624,10 @@ def clean(self):
         super().clean()
         self._clean_algorithm_submission_settings()
         self._clean_submission_limits()
-        self._clean_parent_phase()
+        try:
+            self._clean_parent_phase()
+        except ValidationError as e:
+            raise ValidationError({"parent": e})
```
jmsmkn, line 630: I think this is too fragile. It relies on the fact that the option is on the first tab and that the tab is displayed by default. For these settings all validation errors go to the top otherwise they're hidden. This is not just a problem for parent selection.

What are the options here? The form tabs are necessary otherwise you'd get a gigantic form that you'd have to check every field for. We have to have a place for errors that involve more than one field, so it should be fine to use that, so that should be fixed if it really is a problem rather than just addressing this single field.

chrisvanrun: > I think this is too fragile. It relies on the fact that the option is on the first tab and that the tab is displayed by default. For these settings all validation errors go to the top otherwise they're hidden. This is not just a problem for parent selection.

Crispy forms have tab focus at the time of errors... but having typed that I realize there isn't any indication of multiple errors across multiple so it's still not a sufficient replacement. I agree.

We want to display the problems that might exist on a per-phase basis. When a challenge organizer starts with the challenge there are a heap of warnings, and errors, ranging from submissions not being open to phase family links.

I'll create an issue (maybe later a pitch) but I'll remove the try except here for now.

jmsmkn (approved):

Issue #3294, opened 2024-03-29
Summernote editor doesn't justify text properly


amickan:
**Describe the bug**
Our summernote editor for challenge pages doesn't correctly align text. 

**To Reproduce**
1. Go to a challenge page and edit it, e.g.: https://demo.grand-challenge.org/demo/update/
2. Select to justify the text with "justify full" (paragraph --> justify full)
3. Safe the page 
4. The text is not justified.

**Expected behavior**
The text should be justifed. 

**Additional context**
It turns out that if you use the editor for justify full <p align="justify“> is added to the html code, which is obsolete. If you change it to <p style="text-align: justify;“> everything works fine.

It looks like django-summernote is no longer maintained. Maybe it would make sense to switch to our markdown editor for challenge pages instead?

jmsmkn:
Using MarkDown for pages has been on my mind for a while. I think it is nicer and you can use full html. However, taking the existing pages and putting them into the MarkDown editor will make them really difficult to modify. We could try adding pandoc to convert the HTML to MarkDown and seeing what the result is like, but for some pages it might be manual work (which given we have more than 2000 may be a problem).

I think if we were to modify this we would need to have an opt-in to MarkDown for existing pages, but new ones could be created with that setting being true.

jmsmkn:
The markdown editor for pages should:

- [ ] Be as large as possible
- [ ] Have the preview and editor side by side and independently scrollable
- [ ] Have the markdown cheatsheet at the top
- [ ] Present the editor in monospace font

PR #3552, opened 2024-09-19
Add a pandoc view for testing markdown pages

jmsmkn:
This PR adds a view that uses `pandoc` to convert the existing html pages into markdown but only in the view. It allows staff users to test conversion to see if challenge pages can be easily migrated.

See #3294

### DIFF
--- a/app/grandchallenge/pages/views.py
+++ b/app/grandchallenge/pages/views.py
@@ -20,7 +20,6 @@ from guardian.mixins import LoginRequiredMixin
 from grandchallenge.challenges.views import ActiveChallengeRequiredMixin
 from grandchallenge.charts.specs import stacked_bar, world_map
 from grandchallenge.core.guardian import ObjectPermissionRequiredMixin
-from grandchallenge.core.templatetags.bleach import md2html
 from grandchallenge.evaluation.models import Evaluation, Submission
 from grandchallenge.pages.forms import PageCreateForm, PageUpdateForm
 from grandchallenge.pages.models import Page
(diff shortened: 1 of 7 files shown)


app/grandchallenge/pages/views.py
```diff
@@ -86,6 +88,48 @@ def test_func(self):
...
+    login_url = reverse_lazy("account_login")
+
+    raw = None
+
+    def test_func(self):
+        return self.request.user.is_staff
+
+    def get_object(self, *args, **kwargs):
+        if not self.kwargs["format"].startswith("markdown"):
+            raise PermissionDenied()
```
amickan, line 106: Why have a `format` variable if we only allow one format at the moment?

app/grandchallenge/pages/views.py
```diff
@@ -86,6 +88,48 @@ def test_func(self):
...
+        return super().get_object(*args, **kwargs)
+
+    def get_template_names(self):
+        if self.raw:
+            return ["pages/page_pandoc_detail.html"]
+        else:
+            return super().get_template_names()
```
amickan, line 113: Why do we need a separate raw view? It shows the same converted content, just in isolation, right?

app/grandchallenge/pages/views.py
jmsmkn: Because there are many flavours of MarkDown that I would like to try: https://pandoc.org/chunkedhtml-demo/8.22-markdown-variants.html

app/grandchallenge/pages/views.py
jmsmkn: Yes, this is what the user will be editing, so using different Markdown formats will result in different sources, some of which will be easier to work with than others. A big one is how tables are represented.

app/grandchallenge/pages/views.py
amickan: Ah I misread this check. This is already allowing multiple formats, so long as they start with `markdown`.

app/grandchallenge/pages/views.py
amickan: Sorry, I don't really follow. The user is editing their page in the update view. They won't even see this converted version for now, will they?

jmsmkn: No, but they will once we migrate to using markdown, I want to compare what it will look like to them without having to guess, then migrate, then find out I made a bad choice afterwards.

PR #3553, opened 2024-09-19
Allow filtering metrics on API views

jmsmkn:
We have an option "Display all metrics" that controlled whether to show all of the metrics in the evaluation detail view. This was just to keep things tidy, but a challenge organiser would like to generate metrics that only they can view, and not the participants. This extends the option to also count for the API view, where the metrics will be filtered based on whether the phase allows displaying all metrics, plus whether the user is an admin for that challenge.

Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/348


amickan (dismissed):
LGTM!

PR #3554, opened 2024-09-19
Fix viewer in viewer button link

### LINKED ISSUE #3525: `View in viewer` button in archive item detail page does not work
**Describe the bug**
`View in viewer` button in archive item detail page does not work

**To Reproduce**
1. Go to the items page in an archive
2. Click on `Details`
3. In the details page, click on `View in viewer`
4. Nothing happens

**Expected behavior**
I expect the image to be opened in Cirrus

**Screenshots**

**Additional context**

ammar257ammar:
a context variable in the CIV details view was missing

closes #3525


ammar257ammar:
Usually, I write a view test to check if a string or url exists in the rendered content, but I did not ask before if this is the best way to do it. Should every fix like the one in this PR have a view test? knowing that the views does not have tests to check the urls, buttons, rows ..etc (at least the parts of GC that I dealt with so far)

app/tests/archives_tests/test_views.py
```diff
@@ -1259,7 +1264,7 @@ def get_resp():
     "viewname", ("archives:item-job-list", "archives:item-detail")
 )
 def test_archive_item_permissions_detail_and_list(viewname, client):
-    archive = ArchiveFactory()
+    archive = ArchiveFactory(workstation=WorkstationFactory())
```
jmsmkn, line 1267: The view should work without a workstation being set.

ammar257ammar: I got an error when creating the session url (for the civ details view), so when calling `workstation.slug`, workstation is None

ammar257ammar: After this fix, the test worked

jmsmkn: Nowhere in this test is `archive.workstation` used. It should not be necessary.

ammar257ammar: I tried again without workstation and I got this error (which is the same one I had before)

Screenshot 2024-09-19 at 16 57 23

The test calls `get_view_for_user` for viewname `archives:item-detail`, which in turn calls `ArchiveItemDetailView` which uses the `template archives/archive_item_detail.html` which inherits `components/civ_set_detail.htm` which calls this line ([here](https://github.com/comic/grand-challenge.org/blob/9584c6b8217bec7225f6c0c1f994dd0a81dc05c7/app/grandchallenge/components/templates/components/civ_set_detail.html#L34)):

`{% workstation_session_control_data workstation=object.base_object.workstation context_object=object archive_item=object config=object.base_object.workstation_config %} `

which calls the `slug` of `workstation` [here](https://github.com/comic/grand-challenge.org/blob/9584c6b8217bec7225f6c0c1f994dd0a81dc05c7/app/grandchallenge/workstations/templatetags/workstations.py#L84)

Am I wrong in tracing this issue?

jmsmkn: Think of the fundamental problem. A workstation is not required for an archive. You have a bit of code that is assuming that a workstation is required for an archive. You need to fix that code rather than changing the requirement so that all archives must define a workstation.

app/grandchallenge/workstations/templatetags/workstations.py
```diff
@@ -79,6 +79,9 @@ def get_workstation_path_and_query_string(  # noqa: C901
 def workstation_session_control_data(
     *, workstation, context_object, timeout=False, **kwargs
 ):
+    if workstation is None:
+        return ""
```
jmsmkn, line 83: Seriously? So just keep the workstation button broken unless a workstation is defined?

jmsmkn (changes requested):
Test things before you ask me to review them.

app/grandchallenge/workstations/templatetags/workstations.py
ammar257ammar: that is a silly solution, my bad. I made a new commit where the button only shows if the workstation is not none, and tested it. It looks like this now when there is no workstation.

Screenshot 2024-09-24 at 11 54 25

jmsmkn: Really? Yet again, not a solution. The user needs to be able to view the archive item.

ammar257ammar: It looks like I don't understand what is a workstation and how does it relate to an archive item.
To my understanding, a workstation is the medical image viewer (in this case, Cirrus). Then, the "view in viewer" button should open the medical image linked to the archive item (as CIV) in the assigned workstation to the archive.
If there is no workstation assigned to the archive, that means any image linked to any archive item can not be viewed using "view in viewer" button. However, the archive item details view should still be accessible (as shown in the screenshot above) but without showing the "view in viewer" button.

Can you please give me feedback on how I understand it?

jmsmkn: Specifying a workstation on any object just means use _this_ workstation. Otherwise, use the default.

ammar257ammar:
I have been looking into this more in depth and I have some questions. I will lay down my thought chain.

**1-** I implemented a fix by checking if a workstation is not None, then show the "View in Viewer" button in the archive item details view (commit [https://github.com/comic/grand-challenge.org/pull/3554/commits/9bd3dad43d69cb7d9f7724c31989e16bb112cebd](https://github.com/comic/grand-challenge.org/pull/3554/commits/9bd3dad43d69cb7d9f7724c31989e16bb112cebd)). That was not accepted as a solution. However, I see the same logic implemented for the list view ([here](https://github.com/comic/grand-challenge.org/blob/9bd3dad43d69cb7d9f7724c31989e16bb112cebd/app/grandchallenge/components/templates/components/civ_set_row.html#L57))

where the "View in Viewer" button is shown if workstation is not none (screenshot below)

Screenshot 2024-09-25 at 09 22 00

and not shown when a workstation is None, as in the screenshot below

Screenshot 2024-09-25 at 09 21 41

In this case, using the same feedback on my fix, the behaviour above is also not correct and should be fixed, right?

**2-** I learned from Anne, that what you meant by using the default workstation if not set, that the default workstation is set by an environment variable in settings `settings.DEFAULT_WORKSTATION_SLUG`

Things started to be confusing when I looked at the [ArchiveForm](https://github.com/comic/grand-challenge.org/blob/9bd3dad43d69cb7d9f7724c31989e16bb112cebd/app/grandchallenge/archives/forms.py#L51) where the workstation is a "required" filed and can not be left empty/None. 

This means, there no way through the GC frontend, for an archive not to have a workstation.

Also, looking at the other place where a workstation is used, Reader Studies, the workstation is required on the model level ([here](https://github.com/comic/grand-challenge.org/blob/9bd3dad43d69cb7d9f7724c31989e16bb112cebd/app/grandchallenge/reader_studies/models.py#L205))

So, why mandating a workstation when adding an archive from GC forms while allowing it to be None in the model and then using the default workstation in the view in the case it is None? knowing that it is enforced on the model level in reader studies.

**3-** I could think of 2 ways to approach this

- Check if the workstation is not None, only then show the "View in Viewer" button. That one you did not accept, which probably mean the list view should also be updated.
     

- use the default workstation from `settings` when it is None.
    
But then I am not sure where to check for that. 

Should I add a model property `default_workstation` to the `Archive` model (and use it when workstation in none), that does something like this:

```python
 @cached_property
 def default_workstation(self):
     w, created = Workstation.objects.get_or_create(
            slug=settings.DEFAULT_WORKSTATION_SLUG
     )
     return w
```

**OR**, add a check to [workstation_session_control_data](https://github.com/comic/grand-challenge.org/blob/9bd3dad43d69cb7d9f7724c31989e16bb112cebd/app/grandchallenge/workstations/templatetags/workstations.py#L79). In this case, it will affect the algorithm views that are using the "View in Viewer" button. btw, the workstation is also mandatory in the `Algorithm` model.

```python
if workstation is None:
    Workstation = apps.get_module('workstation', 'Workstation')
    workstation, _ = Workstation.objects.get_or_create(
        slug=settings.DEFAULT_WORKSTATION_SLUG
    )
```

**OR**, do the check in the view (list and details) and add the workstation to the context.

Probably there are other ways too.

What do you think?

**4-** For my understanding, what is the rational behind not making the workstation not nullable on the Archive model level?
I understand that the archive could have archive items that does not hold any image date (i.e. other types of CIVs like JSON .etc). But then, isn't enforcing the default workstation in the view and not allowing the null workstation in the Archive form, mimicking similar behaviour to non-null workstation in the model?

jmsmkn:
I will just do it again.

PR #3557, opened 2024-09-20
Fix inconsistent HTML titles for reader study and statistics pages

### LINKED ISSUE #3556: Page title inconsistencies
Some pages have correct titles:

- https://grand-challenge.org/challenges/ "Challenges - Grand Challenge". 
- https://grand-challenge.org/reader-studies/ "Reader Studies - Grand Challenge"

But some are inconsistent or missing detail:

- https://grand-challenge.org/reader-studies/test-reader-study-harm/ "Grand Challenge"
- https://grand-challenge.org/reader-studies/test-reader-study-harm/progress/ "User Progress - Grand Challenge" (missing reader study name?)
- https://grand-challenge.org/stats/ "Statistics" (missing Grand Challenge?)

This is inconsistent and adds some guesswork when having multiple tabs open or navigating through history.

ammar257ammar:


HarmvZ:
These were actually just examples that I mentioned. I think the issue persists throughout the codebase. I wonder if there is a more structural fix we can apply for this.

HarmvZ:
I think the breadcrumbs are present on most pages. If we could use those to generate a title that would already be much better.

ammar257ammar:
I agree, I will look into it

jmsmkn:
I do not think that there is a good way to generate the title from the breadcrumbs in the same way that there is not a good way to generate the breadcrumbs from the object. I think this is manual work where every page would need to be inspected. But should be fairly easy to find as you only need to lookup where breadcrumbs are defined and look at the corresponding title block.

app/grandchallenge/algorithms/templates/algorithms/algorithmimage_form.html
ammar257ammar, line 1: This title is a bit too long, is it still okay?

ammar257ammar:
I modified a couple of titles from different page types (form, update, details, list, activate) to check with you if this is how you want titles to look like. Please give me feedback to apply it to the rest of the titles.

ammar257ammar:
Any feedback here would be appreciated.

app/grandchallenge/algorithms/templates/algorithms/algorithmimage_form.html
chrisvanrun: I always find this very hard to judge from code alone: a screenshot would help to judge this.

Having '"Algorithm Container Image"` and `"Algorithm"` right behind it is superfluous, but maybe that is the default style?

ammar257ammar: I see the point, it does not seem useful to add the object name and type in this case and the title becomes long.
Those are HTML titles that appear in the browser tab, so not sure if a screenshot help much, check below:

Screenshot 2024-09-30 at 10 52 45

And for long titles, the full title can't be seen anyway

Screenshot 2024-09-30 at 10 56 42

jmsmkn: When thinking about how things are done I often look at GitHub. Take this PR:

Screenshot 2024-09-30 at 11 00 52

It has the object title (or the pk for things that do not have titles, but only if that pk is a UUID) and the thing it belongs too. I think we should follow something similar. In our case I think it is still useful to have Grand Challenge as the last thing. A maximum of 3 things in the title seems sensible - the object, the thing that the object belongs to, the site. New things only have two:

Screenshot 2024-09-30 at 11 03 39

ammar257ammar:
After discussing this issue with James, and in the light of what he had suggested before:

Screenshot 2024-09-30 at 17 24 25


The way to go is to be consistent with the breadcrumb and use two segments from it (the object, the thing that the object belongs to), and then the site name as the third segment.

ammar257ammar:
Also, we agreed to do it one app at a time. So now I pushed a commit [https://github.com/comic/grand-challenge.org/pull/3557/commits/6f4b951740561fb59b898eb90f8441bcfcd3cd45](https://github.com/comic/grand-challenge.org/pull/3557/commits/6f4b951740561fb59b898eb90f8441bcfcd3cd45) fixing the titles for the Algorithms app, which I need feedback for @amickan @chrisvanrun

app/grandchallenge/algorithms/templates/algorithms/algorithmpermissionrequest_form.html
```diff
@@ -3,6 +3,9 @@
 {% load clean from bleach %}
 {% load guardian_tags %}
 
+{% block title %}
+    {% if "change_algorithm" in algorithm_perms %}Review access request{% else %}Request access{% endif %} - {{ algorithm.title }} - {{ block.super }}
```
chrisvanrun, line 7: Not sure if this works. the `get_obj_perms` are block scoped. As such the `algorithm_perms` will likely not exist here?

chrisvanrun (changes requested):
Thanks for tackling this Ammar! Some comments and one general remark since we are aiming for consistency: could you kindly double-check the correct usage of Title Case everywhere?

app/grandchallenge/algorithms/templates/algorithms/algorithmmodel_form.html
```diff
@@ -3,7 +3,7 @@
 {% load url %}
 
 {% block title %}
-    {% if object %}Update{% else %}Create{% endif %} An Algorithm Model - {{ block.super }}
+    Create model - {{ algorithm.title }} - {{ block.super }}
```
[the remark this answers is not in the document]
ammar257ammar: Indeed, it will be cleaner to use that. I am trying to stick with how it is done in the breadcrumbs. I prefer to. be consistent across all titles and use the `|yesno` construct. I will change and see later what James would say.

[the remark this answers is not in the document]
ammar257ammar: Harmonzing all the breadcrumbs should follow (in a separate issue) after finishing the titles.

app/grandchallenge/algorithms/templates/algorithms/algorithm_form.html
```diff
@@ -3,7 +3,7 @@
 {% load url %}
 
 {% block title %}
-        {{ object|yesno:"Update,Create"}} Algorithm {% if request.resolver_match.view_name == 'algorithms:update' %} Settings {% elif request.resolver_match.view_name == 'algorithms:description-update' %} Description {% elif request.resolver_match.view_name == 'evaluation:phase-algorithm-create' %} for Challenge {% endif %}- {{ block.super }}
+    {{ object|yesno:"Update,Create"}} - {{ object }} - {{ block.super }}
```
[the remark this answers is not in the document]
ammar257ammar: Good point, I will fix it.

app/grandchallenge/algorithms/templates/algorithms/display_set_from_job_form.html
```diff
@@ -3,7 +3,7 @@
 {% load url %}
 
 {% block title %}
-    Create Display Set - {{ block.super }}
+    Create Display Set - {{ object.pk }} - {{ block.super }}
```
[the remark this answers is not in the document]
ammar257ammar: mmm I see your point. But since James did not mention to truncate the PK, I will leave it as is and ask again when he reviews it. sounds ok?

app/grandchallenge/algorithms/templates/algorithms/algorithmpermissionrequest_form.html
ammar257ammar: Thanks for mentioning this. I did not know that. I will double check.

app/grandchallenge/algorithms/templates/algorithms/algorithmimage_form.html
```diff
@@ -3,7 +3,7 @@
 {% load url %}
 
 {% block title %}
-    Create An Algorithm Container Image - {{ block.super }}
+    Create Image - {{ algorithm.title }} - {{ block.super }}
```
amickan, line 6: Isn't this template used on update as well?

app/grandchallenge/algorithms/templates/algorithms/algorithmmodel_form.html
```diff
@@ -3,7 +3,7 @@
 {% load url %}
 
 {% block title %}
-    {% if object %}Update{% else %}Create{% endif %} An Algorithm Model - {{ block.super }}
+    Create Model - {{ algorithm.title }} - {{ block.super }}
 {% endblock %}
```
amickan, line 7: Same as above: this template is used both on creation and on updating, right? Why did you remove that from the title?

app/grandchallenge/algorithms/templates/algorithms/display_set_from_job_form.html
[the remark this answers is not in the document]
amickan: I would print the full pk.

app/grandchallenge/algorithms/templates/algorithms/algorithmimage_form.html
ammar257ammar: Yes indeed, when I discussed it with James, he said to follow how it is implemented in the breadcrumbs of the page, even if they are not accurate. For this one, the breadcrumbs has a fixed "Create image" for the page even though it should be dynamic create/update. That is why I left it as is.

app/grandchallenge/algorithms/templates/algorithms/algorithmmodel_form.html
ammar257ammar: Check my comment above. It doesn't feel well to do it like this.

app/grandchallenge/algorithms/templates/algorithms/image_activate.html
```diff
@@ -2,7 +2,7 @@
 {% load crispy_forms_tags %}
 
 {% block title %}
-    Activate Algorithm Image - {{ block.super }}
+    Activate Algorithm Image - {{ object.title }} - {{ block.super }}
```
amickan, line 5: Does this follow the pattern "the object, the thing that the object belongs to, the site" ? Shouldn't this be "Activate Algorithm Image <pk> - Algorithm XYZ - GC?

app/grandchallenge/algorithms/templates/algorithms/job_detail.html
```diff
@@ -12,7 +12,7 @@
 {% load civ %}
 
 {% block title %}
-    Algorithm Result - {{ block.super }}
+    {{ object.pk }} - {{ object.algorithm_image.algorithm.title }} - {{ block.super }}
```
amickan, line 15: I think it would be clearer if you prepend either Job or Algorithm Result to the pk. Just doing `{{ object }}` might be the simplest, that will return "Job 1234"

app/grandchallenge/algorithms/templates/algorithms/job_form_update.html
```diff
@@ -3,7 +3,7 @@
 {% load url %}
 
 {% block title %}
-    Update Algorithm Result - {{ block.super }}
+    Update - {{ object.pk }} - {{ block.super }}
```
amickan, line 6: Again, I think this doesn't follow the discussed pattern, right? "Update Job 1234 - Algorithm Foobar - GC" is what I would expect.

app/grandchallenge/algorithms/templates/algorithms/image_activate.html
ammar257ammar: yeah, "the thing that the object belongs to," is the algorithm image since we are activating it, right?
Also, the breadcrumbs is using {{object.title}}, even though it is preceded by "Algorithms".

app/grandchallenge/algorithms/templates/algorithms/job_progress_detail.html
```diff
@@ -2,7 +2,7 @@
 {% load static %}
 
 {% block title %}
-    Algorithm Job Progress - {{ block.super }}
+    Progress - {{ object.pk }} - {{ block.super }}
```
amickan, line 5: Again, not sure if it is clear anymore what progress your looking at if you just print the pk. I think `Progress - {{ object }} - {{ block.super }}` would be better.

app/grandchallenge/algorithms/templates/algorithms/algorithmpermissionrequest_form.html
chrisvanrun: I assume it checks out? =)

ammar257ammar:
An update was posted on the main issue #3556  where it should be addressed with independent PR for each app.

I will close this PR and follow up with other PRs, starting with this one #3577

PR #3573, opened 2024-10-02
Export of registration requests

chrisvanrun:
With the addition of registration questions, I noticed that aggregating the answers can sometimes be hard.

To allow for maximum flexibility, this PR adds an export button:

### DIFF
--- a/app/grandchallenge/participants/templates/participants/registrationrequest_list.html
+++ b/app/grandchallenge/participants/templates/participants/registrationrequest_list.html
@@ -20,6 +20,14 @@
 
     <h2>Participation Requests for {{ challenge.short_name }}</h2>
 
+    {% if viewable_registration_questions %}
+        <a class="btn btn-primary mb-4"
+            href="{% url 'participants:registration-question-answer-list-export' challenge_short_name=challenge.short_name %}"
+        >
+            <i class="fas fa-download pr-1"></i> Answers to questions
+        </a>
+    {% endif %}
+
     <div class="table-responsive">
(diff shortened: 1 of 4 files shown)


jmsmkn:
Good idea but please discuss these things in issues first as it would have saved you some time.

I think you might have pulled this out from Reader Studies, but we already know from experience that the approach there is not a good one. It's better for us if we don't make the same mistakes again (only new ones please).

The approach here is not aligned with Django, you have a function that mixes in everything - permissions, serialisation, and rendering - and it's something completely custom which makes maintenance and extension more onerous.

The way to tackle this that is aligned with Django is to use Django Rest Framework instead. That way you have a proper separation of concerns, you benefit from the existing rendering options, and it fits in with the existing operation and maintenance of the platform.

chrisvanrun:
> Good idea but please discuss these things in issues first as it would have saved you some time.

I made my round via Teams to query: I will use an issue next time.

Minor loss of time: already had half of an API-route implementation finished before switching it to this direct `get()` approach. Felt a lot less complex. I'll revive the other API route.

jmsmkn:
Sorry, I honestly did not see that Teams message at all - far too easy to miss things there.

I agree that DRF has you touching more locations, but this gives the advantage of separation of concerns and makes features easier to maintain. Security too - one thing I didn't mention above was pagination too, you will need that for lots of answers with lots of sign ups otherwise you can DOS the site. There is an implementation of this sort of thing in the evaluation app for downloading leaderboards, I think that is a better reference for this.

PR #3577, opened 2024-10-07
Make consistent html titles with breadcrumbs for algorithms app

ammar257ammar:
This is part of Task 1 under issue #3556  aiming to fix inconsistent HTML titles in GC.

Each app will have a PR to make sure titles are there for all pages that have a breadcrumbs and that titles matches the breadcrumbs as described in issue #3556

### DIFF
--- a/app/grandchallenge/algorithms/templates/algorithms/algorithmpermissionrequest_form.html
+++ b/app/grandchallenge/algorithms/templates/algorithms/algorithmpermissionrequest_form.html
@@ -3,6 +3,10 @@
 {% load clean from bleach %}
 {% load guardian_tags %}
 
+{% block title %}
+    {% get_obj_perms request.user for algorithm as "algorithm_perms" %}
+    {% if "change_algorithm" in algorithm_perms %}Review Access Request{% else %}Request Access{% endif %} - {{ algorithm.title }} - {{ block.super }}
+{% endblock %}
 
 {% block breadcrumbs %}
     {% get_obj_perms request.user for algorithm as "algorithm_perms" %}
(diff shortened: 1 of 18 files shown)


app/grandchallenge/algorithms/templates/algorithms/algorithmpermissionrequest_form.html
```diff
@@ -3,6 +3,10 @@
 {% load clean from bleach %}
 {% load guardian_tags %}
 
+{% block title %}
+    {% get_obj_perms request.user for algorithm as "algorithm_perms" %}
+    {% if "change_algorithm" in algorithm_perms %}Review Access Request{% else %}Request Access{% endif %} - {{ algorithm.title }} - {{ block.super }}
```
amickan, line 8: `get_obj_perms` is block-scoped, so you indeed need to call it twice if you want to use it in both the title and the main content block. It does seem rather inefficient to do that though, so I'm wondering whether it's really necessary, or whether a more generic title "Access Request" would suffice. I understand that the idea of this PR is to make the title identical to the breadcrumbs, so fine for now. But for round 2 of the overarching PR, I think it's worth thinking about this. You could check the perms once and then add a context variable that you check instead.

amickan (approved):

app/grandchallenge/algorithms/templates/algorithms/algorithmpermissionrequest_form.html
ammar257ammar: I agree with that. Since this part of the task is to make sure the titles are consistent with breadcrumbs, it should be fine for this PR, but definitely to be address in the title fixing PR. I will copy this note to my notes to be address later. Is that okay?

PR #3578, opened 2024-10-07
Add export for participant requests

chrisvanrun:
## Note

This is in a first reviewable state. However, I am pausing development on this and switching to cycle work. So any work following reviews will be delayed. Feel free to delay the review as you see fit.

## Export of participant requests
This PR adds an export via API for participants' requests.

image

The JSON format:
image

The CSV format has the following headers:

- challenge
- changed
- created
- registration_question_answers.0.answer
- registration_question_answers.0.question.question_help_text
- registration_question_answers.0.question.question_text
- registration_question_answers.0.question.required	
- registration_status	
- user.department	
- user.institution	
- user.location	
- user.mugshot	
- user.user.username	
- user.website

The default CSV renderer dumps any JSON under the top level. I have intentionally switched to using the 'regular' `PaginatedCSVRender` with a special application for `registration_question_answers`. This is to address any strange rendering of actual JSONField values.

### DIFF
--- a/app/grandchallenge/participants/api.py
+++ b/app/grandchallenge/participants/api.py
@@ -1,5 +1,15 @@
 import csv
+
 from django.http import HttpResponse
+from django_filters.rest_framework import DjangoFilterBackend
+from rest_framework import mixins, viewsets
+from rest_framework.filters import BaseFilterBackend
+from rest_framework.permissions import DjangoObjectPermissions
+from rest_framework.settings import api_settings
+from rest_framework.views import APIView
+from rest_framework.viewsets import GenericViewSet, ReadOnlyModelViewSet
+from rest_framework_guardian.filters import ObjectPermissionsFilter
+
 from grandchallenge.api.permissions import IsAuthenticated
 from grandchallenge.challenges.models import Challenge
 from grandchallenge.core.guardian import filter_by_permission
@@ -11,19 +21,9 @@ from grandchallenge.participants.models import (
     RegistrationQuestion,
     RegistrationQuestionAnswer,
 )
-
 from grandchallenge.participants.serializers import (
     RegistrationQuestionAnswersSerializer,
 )
-from rest_framework.permissions import DjangoObjectPermissions
-from django_filters.rest_framework import DjangoFilterBackend
-from rest_framework_guardian.filters import ObjectPermissionsFilter
-from rest_framework.settings import api_settings
-from rest_framework import mixins, viewsets
-from rest_framework.views import APIView
-from rest_framework.viewsets import GenericViewSet, ReadOnlyModelViewSet
-
-from rest_framework.filters import BaseFilterBackend
 
 
 class CanViewRegistrationQuestionFilter(BaseFilterBackend):
--- a/app/grandchallenge/participants/api.py
+++ b/app/grandchallenge/participants/api.py
@@ -1,0 +1,56 @@
+import csv
+from django.http import HttpResponse
+from grandchallenge.api.permissions import IsAuthenticated
+from grandchallenge.challenges.models import Challenge
+from grandchallenge.core.guardian import filter_by_permission
+from grandchallenge.core.renderers import PaginatedCSVRenderer
+from grandchallenge.participants.filters import (
+    RegistrationQuestionAnswerFilter,
+)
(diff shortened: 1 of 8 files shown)


app/grandchallenge/participants/api.py
```diff
@@ -0,0 +1,68 @@
...
+    RegistrationRequestCSVRenderer,
+)
+from grandchallenge.participants.serializers import (
+    RegistrationRequestSerializer,
+)
+
+
+class RegistrationRequestViewSet(mixins.ListModelMixin, GenericViewSet):
+    serializer_class = RegistrationRequestSerializer
+    permission_classes = [IsAuthenticated]
```
jmsmkn, line 25: This should have object based permissions, otherwise this is all public information.

app/grandchallenge/participants/renderers.py
```diff
@@ -0,0 +1,32 @@
+import json
+
+from rest_framework.settings import api_settings
+from rest_framework.utils.encoders import JSONEncoder
+from rest_framework_csv.renderers import PaginatedCSVRenderer
+
+
+class RegistrationRequestCSVRenderer(PaginatedCSVRenderer):
```
jmsmkn, line 8: Rather than flattening stuff with a special renderer why not serialize the answers instead?

app/grandchallenge/participants/api.py
chrisvanrun: There are permission checks in the get_queryset. Currently, we don't have any `"view_registrationrequest"` permissions being assigned to the challenge.admins_groups. Should we? If so, I'll create a PR + manual call to have them added first, before doing anything that relies on them.

app/grandchallenge/participants/renderers.py
chrisvanrun: It was 50/50. I had the answers as a `Charfield` initially, but it felt really strange to have a JSON string in a JSON object for the regular API.

chrisvanrun:
Side note: is the swagger view supposed to work in local development?

app/grandchallenge/participants/renderers.py
jmsmkn: That's what it is though - it is a JSON field so in the future may not necessarily be a string. I think it is best to take the same approach as the evaluation serialisation and serialise things on the answer level, rather than the registration request. Permissions are already implemented for that so is less impact. You would then get some duplicate data that is serialised along with each answer (e.g. the username, the request status, the question text) but that is okay for now, it's not much.

chrisvanrun: Do you mean to have a CSV row per answer? 

If so, what about registrees that did not answer any questions? IMHO they should be included in any challenge registration overview. Plus, there are no permissions for the `RegistrationQuestionAnswer`s.

> Permissions are already implemented for that so is less impact.

Are you referring to the permissions for the  `RegistrationQuestionAnswer`s? Those were deemed to be overkill and were removed from the original implementation.

jmsmkn:
> Side note: is the swagger view supposed to work in local development?

Set the environment variable `CSP_REPORT_ONLY=True`

app/grandchallenge/participants/renderers.py
jmsmkn: Empty answers are saved for users that don't answer, right? The permissions weren't added because they weren't needed. Now they are needed if you want to add this view in line with the existing framework. We should do that rather than trying to fight it with custom renderers specific to one particular field.

chrisvanrun: Empty answers are indeed saved (by design), but if there are no questions at the time of sign-up, the registree would now show up in an answer-level serialization.

jmsmkn: Then add another viewset, serialiser and permissions for the registration requests and link to those from the answers.

chrisvanrun: Ok. The current plan (somewhere in the next cooldown week) is to have two routes:

- Link for downloading the registration requests.
- Link for downloading the answers to question registrations (with a crosslink to the registration requests).

chrisvanrun:
Making this a draft to clearly indicate that this is WIP and been put on the backburner for now.

PR #3580, opened 2024-10-07
Make consistent html titles with breadcrumbs for archives app

ammar257ammar:
This is part of Task 1 under issue https://github.com/comic/grand-challenge.org/issues/3556 aiming to fix inconsistent HTML titles in GC.

Each app will have a PR to make sure titles are there for all pages that have a breadcrumbs and that titles matche the breadcrumbs as described in issue https://github.com/comic/grand-challenge.org/issues/3556

### DIFF
--- a/app/grandchallenge/archives/templates/archives/archive_user_groups_form.html
+++ b/app/grandchallenge/archives/templates/archives/archive_user_groups_form.html
@@ -1,6 +1,10 @@
 {% extends "groups/groups_form_base.html" %}
 {% load url %}
 
+{% block title %}
+    Add user - {{ object.title }} - {{ block.super }}
+{% endblock %}
+
 {% block breadcrumbs %}
     <ol class="breadcrumb">
         <li class="breadcrumb-item"><a href="{% url 'archives:list' %}">Archives</a></li>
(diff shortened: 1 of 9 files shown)


app/grandchallenge/archives/templates/archives/archive_user_groups_form.html
```diff
@@ -1,6 +1,10 @@
 {% extends "groups/groups_form_base.html" %}
 {% load url %}
 
+{% block title %}
+    Add user - {{ object.title }} - {{ block.super }}
```
amickan, line 5: This is fine since this is only about making the title and breadcrumbs consistent, but for the next step, I think we should watch for capitalization. I notice that sometimes it's not capitalized, like  here, whereas in most places we do use caps.

amickan (approved):

app/grandchallenge/archives/templates/archives/archive_user_groups_form.html
ammar257ammar: Absolutely, I have it in my notes. There are several places to fix capitalization.

PR #3586, opened 2024-10-07
Keep track of the gpu and memory requirements at creation time

jmsmkn:
This PR ensures that the gpu type, memory limit and time limit do not change during the execution of a job or evaluation. This ensures that we can set the credits a-priori, and ensure that the correct GPU types are being used during submission in the future.

See #3263


app/grandchallenge/algorithms/forms.py
```diff
@@ -100,9 +100,6 @@ class JobCreateForm(SaveFormInitMixin, Form):
     creator = ModelChoiceField(
         queryset=None, disabled=True, required=False, widget=HiddenInput
     )
-    time_limit = IntegerField(
```
jmsmkn, line 103: Removed as this didn't need to be on the form, we still have some hardcoded things in the `form_valid` method. I think this should be a create view that manipulated the object in the forms `save` method if necessary.

app/grandchallenge/components/backends/base.py
```diff
@@ -48,8 +48,7 @@ def __init__(
         exec_image_repo_tag: str,
         memory_limit: int,
         time_limit: int,
-        requires_gpu: bool,
-        desired_gpu_type: GPUTypeChoices,
+        requires_gpu_type: GPUTypeChoices,
```
jmsmkn, line 51: Simplification of the API, we just allow an empty string (or anything False-y) to indicate not to attach a GPU.

app/grandchallenge/evaluation/forms.py
```diff
@@ -629,9 +617,32 @@ def check_submission_limit_avoidance(self, *, creator):
...
+        if self._phase.submission_kind == SubmissionKindChoices.ALGORITHM:
+            self.instance.algorithm_requires_gpu_type = self.cleaned_data[
+                "algorithm_image"
+            ].requires_gpu_type
+            self.instance.algorithm_requires_memory_gb = self.cleaned_data[
+                "algorithm_image"
+            ].requires_memory_gb
+        else:
+            self.instance.algorithm_requires_gpu_type = ""
+            self.instance.algorithm_requires_memory_gb = 0
```
jmsmkn, line 630: Not sure that this is the best but these fields are non-editable so cannot be included in the form.

app/grandchallenge/algorithms/models.py
```diff
@@ -926,6 +931,10 @@ def init_is_complimentary(self):
             > 0
         )
 
+    def init_credits_consumed(self):
+        # TODO implementation
```
jmsmkn, line 935: Doing this in another PR.

jmsmkn:
Do not merge this until #3595 is added.

app/grandchallenge/evaluation/forms.py
amickan: I think this is the way to go given that the fields cannot be added to the form.

app/grandchallenge/components/backends/base.py
amickan: Ok, but isn't `requires_gpu_type` defined to be non-nullable on the model level now? So how would "" pass validation for a ComponentJob instance?

jmsmkn: It should pass as `""` is not `None`.

jmsmkn: We could set `blank=True`, but that is meaningless as the field is non-editable.

jmsmkn: Confirmed that `""` does pass validation.

amickan: Ok, thanks. I always forget how CharFields handle "", and if it's not None, I would have thought that "" would only be allowed if it's included in the GPUTypeChoices. But anyway, thanks for checking!

jmsmkn: Good blog post on this topic (of course it's Adam Johnson): https://adamj.eu/tech/2020/01/22/djangos-field-choices-dont-constrain-your-data/

app/grandchallenge/components/models.py
```diff
@@ -1876,6 +1884,13 @@ def can_execute(self):
             .exists()
         )
 
+    @property
+    def requires_gpu_type(self):
+        if self.requires_gpu:
+            return self.desired_gpu_type
+        else:
+            return ""
+
```
amickan, line 1893: Is the plan to also migrate the `ComponentImage` model to only have a `requires_gpu_type` CharField and deprecate the `requires_gpu` and `desired_gpu` fields?

jmsmkn: Precisely, makes GPU selection much more intuitive when we have more than one option.

amickan (approved):
LGTM!

PR #3601, opened 2024-10-09
Add option to set the minimum number of credits per job

### LINKED ISSUE #3263: Save used credits on job model, use maximum costs to determine credits and set maximum number of jobs

jmsmkn:
This PR allows the user to set the minimum number of credits per job. We now consider the time limit of the job rather than the average duration in determining the number of credits used. The number of credits is now calculated before the job starts.

See #3263


amickan (approved):
LGTM! Just out of curiosity: why would users want to set a non-default minimum_credits_per_job? All they can use this for is to make their algorithm more expensive to run than it would be just considering the max run time, right?

jmsmkn:
Yes that's the use case. We used to allow users to set the credits directly but they didn't take account the computation costs, so we switched, this brings the original feature back. Low runtime, no gpu, and low memory requirements can lead to low credits per job, so users could run hundreds of job per month in that case. Request was that algorithm editors wanted to limit it further.

amickan:
> Yes that's the use case. We used to allow users to set the credits directly but they didn't take account the computation costs, so we switched, this brings the original feature back. Low runtime, no gpu, and low memory requirements can lead to low credits per job, so users could run hundreds of job per month in that case. Request was that algorithm editors wanted to limit it further.

Clear, thanks for explaining!

PR #3604, opened 2024-10-09
Make consistent html titles with breadcrumbs for participants app

ammar257ammar:
This is part of Task 1 under issue https://github.com/comic/grand-challenge.org/issues/3556 aiming to fix inconsistent HTML titles in GC.

Each app will have a PR to make sure titles are there for all pages that have a breadcrumbs and that titles match the breadcrumbs as described in issue https://github.com/comic/grand-challenge.org/issues/3556

### DIFF
--- a/app/grandchallenge/participants/templates/participants/participants_list.html
+++ b/app/grandchallenge/participants/templates/participants/participants_list.html
@@ -1,6 +1,10 @@
 {% extends 'auth/user_list.html' %}
 {% load url %}
 
+{% block title %}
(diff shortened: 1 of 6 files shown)


ammar257ammar:
**Note for Task 2**
Titles for this app need to be carefully checked in Task 2 of the issue. Several of the templates in this app extends pages/challenge_settings_base.html which already has a title block (which may not be necessary to have since it is a base template and mainly used by extending it). Therefore, using {{block.super}} will add the entire title to participants templates.

amickan (approved):

PR #3616, opened 2024-10-10
Consolidate component job requirement fields

jmsmkn:
This PR consolidates `requires_gpu` and `desired_gpu_type` into one choice field `requires_gpu_type` and sets this on the algorithm and phase rather than on the individual container images along with `requires_memory_gb`. This is because they rarely change between container images for the same algorithm and users more often forget to update. 

I removed the option for challenge admins to set a gpu for an evaluation as this has not been used in 3 years. We can still do that from the admin.

See https://github.com/DIAGNijmegen/rse-roadmap/issues/346


app/grandchallenge/algorithms/migrations/0059_auto_20241010_1159.py
```diff
@@ -0,0 +1,36 @@
+# Generated by Django 4.2.16 on 2024-10-10 11:59
+
+from django.db import migrations
+
+
+def set_requirements_from_active_image(apps, schema_editor):
```
jmsmkn, line 6: Note that this is an active migration.

app/grandchallenge/evaluation/migrations/0066_auto_20241010_1200.py
```diff
@@ -0,0 +1,32 @@
+# Generated by Django 4.2.16 on 2024-10-10 12:00
+
+from django.db import migrations
+
+
+def set_requirements_from_active_method(apps, schema_editor):
```
jmsmkn, line 6: Note that this is an active migration.

jmsmkn:
1. Because the choices are only done in validation and do not change the SQL generated by the migration file - see https://adamj.eu/tech/2020/01/22/djangos-field-choices-dont-constrain-your-data/
2. It is, but in this case we require stopping all workers anyway in order to gain the locks on all of these objects so the old workers with the old code will not be running.

amickan:
> 1. Because the choices are only done in validation and do not change the SQL generated by the migration file - see https://adamj.eu/tech/2020/01/22/djangos-field-choices-dont-constrain-your-data/

Ah right, thanks!

> 2. It is, but in this case we require stopping all workers anyway in order to gain the locks on all of these objects so the old workers with the old code will not be running.

Oh that is a good point. So then the site will be down for the duration of the migration, right?

amickan (approved):

jmsmkn:
> So then the site will be down for the duration of the migration, right?

Only the celery workers.

PR #3619, opened 2024-10-14
Correct html titles for algorithms app

ammar257ammar:
This is part of Task 2 under issue https://github.com/comic/grand-challenge.org/issues/3556 aiming to make sure the titles are clear and correct.

Each app will have a PR to make sure titles are clear and correct for all pages that have breadcrumbs as described in issue https://github.com/comic/grand-challenge.org/issues/3556

### DIFF
--- a/app/grandchallenge/algorithms/templates/algorithms/display_set_from_job_form.html
+++ b/app/grandchallenge/algorithms/templates/algorithms/display_set_from_job_form.html
@@ -3,7 +3,7 @@
 {% load url %}
 
 {% block title %}
-    Create a Display Set - {{ object.pk }} - {{ block.super }}
+    Create a Display Set from Result {{ object.pk }} - {{ object.algorithm_image.algorithm.title }} - {{ block.super }}
 {% endblock %}
 
 {% block breadcrumbs %}
--- a/app/grandchallenge/algorithms/templates/algorithms/display_set_from_job_form.html
+++ b/app/grandchallenge/algorithms/templates/algorithms/display_set_from_job_form.html
@@ -3,7 +3,7 @@
 {% load url %}
 
 {% block title %}
-    Create a Display Set from Result {{ object.pk }} - {{ object.algorithm_image.algorithm.title }} - {{ block.super }}
+    Create a Display Set - {{ object.pk }} - {{ block.super }}
 {% endblock %}
 
 {% block breadcrumbs %}
--- a/app/grandchallenge/algorithms/templates/algorithms/display_set_from_job_form.html
+++ b/app/grandchallenge/algorithms/templates/algorithms/display_set_from_job_form.html
@@ -3,7 +3,7 @@
 {% load url %}
 
 {% block title %}
-    Create a Display Set - {{ object.pk }} - {{ block.super }}
+    Create a Display Set from Result - {{ object.pk }} - {{ block.super }}
 {% endblock %}
 
 {% block breadcrumbs %}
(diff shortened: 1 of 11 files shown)


app/grandchallenge/algorithms/templates/algorithms/display_set_from_job_form.html
```diff
@@ -3,7 +3,7 @@
 {% load url %}
 
 {% block title %}
-    Create a Display Set - {{ object.pk }} - {{ block.super }}
+    Create a Display Set from Result {{ object.pk }} - {{ object.algorithm_image.algorithm.title }} - {{ block.super }}
```
amickan, line 6: Here the title is different from the breadcrumbs, right? "Create Display Set" vs "Create Display Set from Result"

app/grandchallenge/algorithms/templates/algorithms/job_detail.html
```diff
@@ -12,7 +12,7 @@
 {% load civ %}
 
 {% block title %}
-    {{ object.pk }} - Results - {{ block.super }}
+    Result {{ object.pk }} - {{ object.algorithm_image.algorithm.title }} - {{ block.super }}
```
amickan, line 15: This is introducing inconsistencies between the title and breadcrumbs again. I thought we wanted them to be the same? I see that with a maximum depth of 3, you wouldn't be able to include the algorithm title in the title here, but then maybe do: `{{ object }} - {{ object.algorithm_image.algorithm.title }} - {{ block.super }}` and change the last breadcrumb to `{{ object }}` too? That will print (if I'm not mistaken, do double check) "Job 1234".

app/grandchallenge/algorithms/templates/algorithms/job_form_update.html
```diff
@@ -3,7 +3,7 @@
 {% load url %}
 
 {% block title %}
-    Update - {{ object.pk }} - {{ block.super }}
+    Update Result {{ object.pk }} - {{ object.algorithm_image.algorithm.title }} - {{ block.super }}
```
amickan, line 6: See comment above.

app/grandchallenge/algorithms/templates/algorithms/display_set_from_job_form.html
ammar257ammar: True, but then we won't be able to make the titles clear as proposed in the issue earlier [here](https://github.com/comic/grand-challenge.org/issues/3556#issuecomment-2390943068)

Then, I will only fix the incorrect things and casing, this applies to the other two comments too.

app/grandchallenge/algorithms/templates/algorithms/job_detail.html
ammar257ammar: Here, {{ object }} will show Job instead of results which was pointed before as inconsistent.
See comment above. I will stick to fixing the casing and incorrect things to remain consistent with the breadcrumbs.

app/grandchallenge/algorithms/templates/algorithms/display_set_from_job_form.html
amickan: Ok, but why not change it to "Display Set from Result" in both the title and the breadcrumb?

app/grandchallenge/algorithms/templates/algorithms/job_detail.html
amickan: Ok, but on the Job Progress template you do use the term "Job" instead of result, so that needs to be changed then too, I think.

ammar257ammar: Indeed, I will delete it, "Progress" will do fine I think.

app/grandchallenge/algorithms/templates/algorithms/display_set_from_job_form.html
ammar257ammar: That is also fine. I will change it.

PR #3620, opened 2024-10-14
Make all inputs required on job creation

amickan:
Addresses the job stuck in `VALIADTING_INPUTS` from this morning.

### DIFF
--- a/app/grandchallenge/algorithms/forms.py
+++ b/app/grandchallenge/algorithms/forms.py
@@ -148,7 +148,7 @@ class JobCreateForm(SaveFormInitMixin, Form):
                 instance=inp,
                 initial=initial if initial else inp.default_value,
                 user=self._user,
-                required=inp.value_required,
+                required=True,
                 help_text=clean(inp.description) if inp.description else "",
                 form_data=self.data,
             ).field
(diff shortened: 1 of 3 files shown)


jmsmkn (dismissed):

jmsmkn:
It looks like the API test needs to be updated too.

amickan:
> It looks like the API test needs to be updated too.

I didn't touch the API code. All inputs are required there already. The test that failed passes locally for me, so I'm rerunning the tests here

jmsmkn:
Sorry, thought it was an API test but I was mistaken. It is a test which touches this view, probably a wrong assumption there. Very strange that it passes locally as it has failed again.

amickan:
> It is a test which touches this view, probably a wrong assumption there. Very strange that it passes locally as it has failed again.

Strange indeed. I've tried a few times, but can't get it to fail locally. I have to go now, will revisit this on Wednesday.

jmsmkn:
Root cause: the test `test_job_gpu_type_set_with_api` that I added in #3586 did not actually use the api view but the template view, which was changed here. Now fixed. Can't explain the differences between CI and your machine, but it also failed on mine. Please double check your setup.

PR #3629, opened 2024-10-16
Fix error reporting for CIVs

### LINKED ISSUE #3618: Some notification messages are too long to be stored in a CharField
See https://grand-challenge.sentry.io/issues/5985194691/?project=303639&query=%21logger%3Acsp%20is%3Aunresolved&referrer=issue-stream&statsPeriod=7d&stream_index=5

We have validation errors that are too long to be stored in a `CharField` (with `max_length` set to 255). I don't think we should truncate them, instead I suggest migrating the `message` field on the `Notification` model to a `TextField`.

### LINKED ISSUE #3628: `create_civ` calls inexistent `handle_error` method
See sentry; https://grand-challenge.sentry.io/issues/5993062423/?project=303639&query=%21logger%3Acsp%20is%3Aunresolved&referrer=issue-stream&statsPeriod=7d&stream_index=2

amickan:
Closes #3628 
Closes #3618


app/grandchallenge/components/models.py
```diff
@@ -2288,13 +2290,31 @@ def create_civ_for_value(
...
+                    civ.delete()
+
                 if new_value in ci.default_field.empty_values:
                     self.remove_civ(civ=current_civ)
                 else:
-                    self.handle_error(
-                        error_message=format_validation_error_message(e),
-                    )
-                    raise e
+                    if hasattr(self, "update_status"):
```
jmsmkn, line 2299: It seems like this function might be doing too much, sometimes sending an update status, sometimes sending a notification. I don't really understand when `self` has an `update_status` method or not, and as it is this feels like it would be fragile. Can an `update_status` method be added to the cases where it currently doesn't exist where the notification is sent?

amickan: It currently does not exist for display sets and archive items. I can add that method there, albeit it would be a bit of a misnomer since these models don't have a `status` to update. I could also call it `handle_error`. For algorithms, all that this `handle_error` function would then do is call the existing `update_status` method that we have for component jobs, fine I guess

jmsmkn: Yes that sounds good.

amickan: Ok, that took longer than anticipated. I went back and forth between one method vs, separate model methods; and also wasn't sure whether or not to pass along the upload_session.

I ultimately added a new `handle_civ_validation_error` method that gets attached to all models that inherit from `CIVForObjectMixin`. This method either calls each models own `handle_civ_error` method (doing model specific stuff) or it calls the `update_status` method on the passed upload_session instance. 

I opted to pass along the upload_session so that we can call this new method everywhere in the CIV validation code. The update_status method repeats some of the logic around updating the job status, but also updates the upload session object. Refactoring that method wasn't possible since there are other places that rely on it that are not connected to CIV validation. The alternative would be to call the update_status method directly where that is possible, but then we don't have 1 way of handling errors for CIV validation anymore.....

amickan: While I was at it, I also changed the job updating so that we now add errors to the `detailed_error_message` rather than overriding it every time. that makes sure that users will see all error messages, not just the latest;
image

app/grandchallenge/components/models.py
```diff
@@ -1504,7 +1504,7 @@ class ComponentJob(FieldChangeMixin, UUIDModel):
     runtime_metrics = models.JSONField(default=dict, editable=False)
     error_message = models.CharField(max_length=1024, default="")
     detailed_error_message = models.JSONField(
-        blank=True, null=True, default=None
+        blank=True, null=True, default=dict
```
amickan, line 1507: Changing this to an empty dictionary by default since that is how it is used everywhere in the codebase, and if it were ever used with a list or, things would break, so making this explicit here seems the way to go.

jmsmkn: Do the existing values need to be updated? Should we disallow `null` so that there are not 2 different representations of empty data?

jmsmkn: If the default is the only thing being changed then new migrations are not necessary. The default is only used when the field is added so the new migrations will not have any SQL changes. If this is the case then the new migrations can be deleted and the latest ones that touches these fields can be modified.

app/grandchallenge/algorithms/models.py
```diff
@@ -1060,6 +1060,15 @@ def validate_inputs_and_execute(self, *, inputs):
                     linked_task=linked_task,
                 )
 
+    def handle_civ_error(self, interface, error_message, user=None):
+        detailed_error_message = self.detailed_error_message
+        detailed_error_message[interface.title] = error_message
```
jmsmkn, line 1065: This construction will modify the original `self.detailed_error_message` due to how dictionaries are referenced in Python, not sure that is what you want.

app/grandchallenge/cases/models.py
```diff
@@ -148,8 +170,14 @@ def update_status(self, *, status, error_message=None, linked_object=None):
             and linked_object
             and hasattr(linked_object, "update_status")
         ):
+            detailed_error_message_dict = linked_object.detailed_error_message
+            for key, val in detailed_error_message.items():
+                detailed_error_message_dict[key] = val
```
jmsmkn, line 175: Same here, I think.

app/grandchallenge/cases/tasks.py
```diff
@@ -159,20 +165,43 @@ def build_images(  # noqa:C901
...
+            error_message=(
+                "One or more of the inputs failed validation."
+                if linked_object
+                else "An unexpected error occurred"
+            ),
+            detailed_error_message=(
+                {ci.title: "An unexpected error occurred"}
+                if linked_object
+                else None
+            ),
```
jmsmkn, line 204: Could these repeated `if`/`else` statements be handled in `upload_session.update_status` instead. They only seem to depend on `linked_object` which `upload_session.update_status` has access to.

app/grandchallenge/components/models.py
jmsmkn: We have to be careful about linking too many things together here, it is quite difficult to follow. Maybe one method of handling errors is not possible, but there should only be one API. Concretely that would mean that we standardise the API of an error handler, a specific error handler for the situation is instantiated when needed (factory method) and then it is called when used. Example of this in the codebase would be the swappable component backends.

For another PR but that representation of error messages could be made nicer, a dictionary to a description list should be straightforward.

app/grandchallenge/components/models.py
jmsmkn: Okay then that would require a new migration so no need to edit the old ones.

app/grandchallenge/components/models.py
amickan: Ok, I will give that a try.

app/grandchallenge/algorithms/models.py
amickan: True, it doesn't matter too much since we're updating `self.detailed_error_message` right after this anyway, but it's probably still better to copy the dictionary contents into a new variable.

app/grandchallenge/components/models.py
amickan: I have pushed a suggestion. Let me know what you think

amickan: and changing the display to a description list is such a minor change that it doesn't warrant a separate PR, so I have made that change here as well:

image

app/grandchallenge/core/error_handlers.py
```diff
@@ -0,0 +1,98 @@
+from grandchallenge.notifications.models import Notification, NotificationType
+
+
+class ErrorHandler:
+    def handle_error(self, *, error_message, user=None, interface=None):
+        raise NotImplementedError("Subclasses must implement this method")
```
jmsmkn, line 6: You could use ABC here, this way everything is checked at import time rather than runtime. 

```python
class ErrorHandler(ABC):
    @abstractmethod
    def handle_error(self, *, error_message, user=None, interface=None):
        pass
```

app/grandchallenge/core/error_handlers.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+class ErrorHandler:
+    def handle_error(self, *, error_message, user=None, interface=None):
+        raise NotImplementedError("Subclasses must implement this method")
+
+
+class JobCIVErrorHandler(ErrorHandler):
+
+    def __init__(self, job):
```
jmsmkn, line 11: All of the `__init__` definitions in this file are missing `*args`, `**kwargs` and calls to `super()`. We can future-proof the codebase a little by adding that, which would also add out stylistic preference for kwargs only methods.
```suggestion
    def __init__(self, *args, job, **kwargs):
        super().__init__(*args, **kwargs)

```

app/grandchallenge/core/error_handlers.py
```diff
@@ -0,0 +1,98 @@
...
+
+        if not job or not isinstance(job, Job):
+            raise RuntimeError(
+                "You need to provide a Job instance to this error handler."
+            )
+
+        self._job = job
+
+    def handle_error(self, *, error_message, interface, user=None):
+        detailed_error_message = self._job.detailed_error_message.copy()
```
jmsmkn, line 22: `.copy()` is a shallow copy, you likely want to use `copy.deepcopy(self._job.detailed_error_message)`.

app/grandchallenge/core/error_handlers.py
jmsmkn: 3 more instances of this in this file.

app/grandchallenge/cases/models.py
```diff
@@ -129,27 +134,72 @@ def default_error_message(self):
...
+        from grandchallenge.algorithms.models import Job
 
         if (
             self.error_message
             and linked_object
-            and hasattr(linked_object, "update_status")
+            and isinstance(linked_object, Job)
         ):
+            detailed_error_message_dict = (
+                linked_object.detailed_error_message.copy()
```
jmsmkn, line 194: `deepcopy` is maybe also needed here.

app/grandchallenge/core/error_handlers.py
```diff
@@ -0,0 +1,98 @@
+from grandchallenge.notifications.models import Notification, NotificationType
+
+
+class ErrorHandler:
+    def handle_error(self, *, error_message, user=None, interface=None):
+        raise NotImplementedError("Subclasses must implement this method")
+
+
+class JobCIVErrorHandler(ErrorHandler):
```
jmsmkn, line 9: Some of the names are still unclear to me, could you add some docstrings about which situations each error handler should be used?

app/grandchallenge/cases/tasks.py
```diff
@@ -125,23 +129,43 @@ def build_images(  # noqa:C901
...
+                app_label=linked_app_label,
+                model_name=linked_model_name,
+                pk=linked_object_pk,
+            )
+        except ObjectDoesNotExist as e:
+            if linked_model_name in [
+                ArchiveItem._meta.model_name,
+                DisplaySet._meta.model_name,
+            ]:
+                # users can delete archive items and display sets before this task runs
```
jmsmkn, line 158: Why would we continue in this case? There's nothing to add the image to so why continue to import it?

app/grandchallenge/components/models.py
```diff
@@ -2397,25 +2421,38 @@ def create_civ_for_file(
...
+                linked_job=self if isinstance(self, Job) else None,
+            )
+        elif isinstance(self, Job):
+            return JobCIVErrorHandler(job=self)
+        elif linked_object and isinstance(linked_object, UserUpload):
+            return UserUploadCIVErrorHandler(
+                user_upload=linked_object,
+            )
+        else:
+            return SystemCIVErrorHandler()
```
jmsmkn, line 2454: We don't have a fallthrough for new/future situations that the developer may need to implement their own specific handler for. Can we already determine the cases where `SystemCIVErrorHandler` is appropriate and raise an error in else?

app/grandchallenge/components/models.py
```diff
@@ -2219,18 +2223,25 @@ def remove_civ(self, *, civ):
...
         if not self.is_editable:
-            raise RuntimeError(
-                f"{self} is not editable. CIVs cannot be added or removed from it."
+            # This can happen for Jobs with multiple inputs,
+            # if another input has already failed validation
+            logger.error(
+                f"{self} is not editable. CIVs cannot be added or removed from it.",
+                exc_info=True,
             )
+            return
```
jmsmkn, line 2232: For my understanding, what was the motivation for this change? We're definitely okay to continue here?

app/grandchallenge/components/models.py
jmsmkn: I think that the tasks are much cleaner now!

app/grandchallenge/cases/tasks.py
amickan: Good point! We shouldn't continue here.

app/grandchallenge/components/models.py
amickan: This was a bug in the implementation that we just have not hit yet in production. You can run into this If a job has multiple inputs, say 2 image inputs, and the user selects existing images for both where the first fails validation (voxel values don't match, for example). The job is then marked as CANCELLED before the second input is validated. 

This is a corner case that hardly ever happens because people are probably more likely to upload new images then to search for an existing image. For values, where validation is also instantaneous (not async), this is not an issue because they are validated client side already.

amickan: Another way of circumventing this would be to subset the choice of existing images by interface (like we do for files now), but the lookup is a bit more complex for images and I'm not sure if we want to restrict it in that way.

jmsmkn: I don't think we want to add that restriction. I'm just not sure if not raising the error here is and just returning the right thing we want.

amickan: We're still reporting the error to sentry here, so we will still get informed (also good because if this happens for display sets or archive items, we need to investigate). If we raise the error, the user will get an ugly error screen. They should just see that the Job has been cancelled. 

What would you do instead?

jmsmkn: I'm not sure which is why I'm asking. Should it raise an exception and be handled in the caller? I would expect a function called `create_civ` to do exactly that, and something must have gone wrong if it didn't. I think the risk is here that the caller function carries on having expected a civ to be created but it wasn't.

app/grandchallenge/components/models.py
```diff
@@ -2221,15 +2221,35 @@ def remove_civ(self, *, civ):
...
+            )
+        else:
+            for civ_data in values:
+                try:
+                    self.create_civ(
+                        civ_data=civ_data,
+                        user=user,
+                        linked_task=linked_task,
+                    )
+                except RuntimeError:
```
jmsmkn, line 2239: Why do we catch this here? The same sort of error is raised above so callers are expected to handle this I think.

Additionally, catching `RuntimeError` might not be specific enough. It is being caught without being inspected, so could hide other `RuntimeError`s. Maybe a specific `CIVNotEditableException` should be created and used?

amickan: A more specific exception is a good idea indeed. 

And as for catching it here: the editable state can change during the for-loop, so while the object is still editable when this function is called and also still for the first civ that is being created, it might no longer be when calling `create_civ` for the second time (because the first civ failed validation), hence why we catch it here.

jmsmkn: Indeed, but if the caller of `validate_values_and_execute_linked_task` is expected to handle `RuntimeError/CIVNotEditableException` which is raised on L2228 then we can let `self.create_civ` raise it too, so shouldn't need to explicitly handle it here, right? It also seems like `self.is_editable` is checked in `self.create_civ`, so the extra check here is unnecessary? This function would then reduce down to:

```python
    def validate_values_and_execute_linked_task(
        self, *, values, user, linked_task=None
    ):
            for civ_data in values:
                self.create_civ(
                    civ_data=civ_data,
                    user=user,
                    linked_task=linked_task,
                )
```

app/grandchallenge/components/backends/exceptions.py
```diff
@@ -25,3 +25,12 @@ class TaskCancelled(ComponentBaseException):
...
     """Raised if the process did not exit cleanly"""
+
+
+class CIVNotEditableException(Exception):
+    """
+    Raised during attempt to update an archive item,
+    display set or algorithm job when they were not editable
+    """
+
+    pass
```
jmsmkn, line 36:
```suggestion
class CIVNotEditableException(ComponentBaseException):
    """
    Raised during attempt to update an archive item,
    display set or algorithm job when they were not editable
    """
```

app/grandchallenge/components/models.py
jmsmkn: Much clearer!

jmsmkn (approved):

PR #3630, opened 2024-10-16
Correct html titles for evaluation app

ammar257ammar:
This is part of Task 2 under issue https://github.com/comic/grand-challenge.org/issues/3556 aiming to make sure the titles are clear and correct.

Each app will have a PR to make sure titles are clear and correct for all pages that have breadcrumbs as described in issue https://github.com/comic/grand-challenge.org/issues/3556

### DIFF
--- a/app/grandchallenge/evaluation/templates/evaluation/combinedleaderboard_detail.html
+++ b/app/grandchallenge/evaluation/templates/evaluation/combinedleaderboard_detail.html
@@ -6,7 +6,7 @@
 {% load profiles %}
 
 {% block title %}
-    {{ object.title|title }} - Combined Leaderboards - {{ block.super }}
+    {{ object.title|title }} - Leaderboards - {{ block.super }}
 {% endblock %}
 
 {% block breadcrumbs %}
@@ -17,7 +17,7 @@
         <li class="breadcrumb-item"><a
                 href="{{ challenge.get_absolute_url }}">{% firstof challenge.title challenge.short_name %}</a></li>
         <li class="breadcrumb-item"><a
-                href="{% url 'evaluation:leaderboard' challenge_short_name=challenge.short_name slug=challenge.phase_set.first.slug %}">Combined Leaderboards</a></li>
+                href="{% url 'evaluation:leaderboard' challenge_short_name=challenge.short_name slug=challenge.phase_set.first.slug %}">Leaderboards</a></li>
         <li class="breadcrumb-item active"
             aria-current="page">{{ object.title|title }}</li>
     </ol>
--- a/app/grandchallenge/evaluation/templates/evaluation/combinedleaderboard_detail.html
+++ b/app/grandchallenge/evaluation/templates/evaluation/combinedleaderboard_detail.html
@@ -6,7 +6,7 @@
 {% load profiles %}
 
 {% block title %}
-    {{ object.title|title }} - Leaderboards - {{ block.super }}
+    {{ object.title|title }} - Combined Leaderboards - {{ block.super }}
 {% endblock %}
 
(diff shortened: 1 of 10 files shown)


app/grandchallenge/evaluation/templates/evaluation/combinedleaderboard_detail.html
```diff
@@ -17,7 +17,7 @@
         <li class="breadcrumb-item"><a
                 href="{{ challenge.get_absolute_url }}">{% firstof challenge.title challenge.short_name %}</a></li>
         <li class="breadcrumb-item"><a
-                href="{% url 'evaluation:leaderboard' challenge_short_name=challenge.short_name slug=challenge.phase_set.first.slug %}">Leaderboards</a></li>
+                href="{% url 'evaluation:leaderboard' challenge_short_name=challenge.short_name slug=challenge.phase_set.first.slug %}">Combined Leaderboards</a></li>
```
amickan, line 20: I think this should stay at Leaderboards because it links to the leaderboards not only the combined leaderboards.

app/grandchallenge/evaluation/templates/evaluation/combinedleaderboard_form.html
```diff
@@ -5,7 +5,7 @@
 {% load crispy_forms_tags %}
 
 {% block title %}
-    {{ object|yesno:"Update,Create" }} -{% if object %} {{ object.title|title }} -{% endif %} {{ block.super }}
+    {{ object|yesno:"Update,Create Combined Leaderboard" }} {% if object %} - {{ object.title|title }} {% else %} - Combined Leaderboards {% endif %} - {{ block.super }}
```
amickan, line 8: Same here. This should "Leaderboards" and that also needs to be fixed in the breadcrumbs, I think.

app/grandchallenge/evaluation/templates/evaluation/combinedleaderboard_detail.html
ammar257ammar: I checked and there is a template for leaderboards ([here](https://github.com/comic/grand-challenge.org/blob/6954da87bc61d47ae23f7c30469b3d8cf05629cf/app/grandchallenge/evaluation/templates/evaluation/leaderboard_detail.html)) other than combined leaderboards.

For example, in the leaderboard tab, both phase leaderboards and combined leaderboards will have the word "Leaderboards" in the breadcrumb and title, as in below:

Screenshot 2024-10-18 at 13 41 47

Screenshot 2024-10-18 at 13 41 59

While the "combined leaderboard" settings and delete confirmation views, they will have "Combined Leaderboards" in the breadcrumb and title, as in below:

Screenshot 2024-10-18 at 13 42 06

Do I get it right?

amickan: Yes, you are right, but I think that actually does not make sense. The breadcrumbs that read "Leaderboards" and those that read "Combined leaderboards" all lead to the same view (i.e. the first challenge leaderboard), so I think the "Combined Leaderboards" breadcrumb is misleading and should be replaced with just "Leaderboards" everywhere

ammar257ammar: okay, I made it "Leaderboards" in the three templates for combinedleaderboard

app/grandchallenge/evaluation/templates/evaluation/combinedleaderboard_form.html
ammar257ammar: Done for the three templates for combinedleaderboard

amickan (approved):

PR #3634, opened 2024-10-16
Migrate to per algorithm credits

jmsmkn:
This PR introduces `AlgorithmUserCredit`, a new model that allows algorithm editors to grant specific users additional credits for their algorithm. This is useful for:

- Research collaborations where specific users need more compute time
- Paid access to algorithms where users purchase additional credits
- Testing periods where selected users get extra credits

## How it works

Each `AlgorithmUserCredit` record contains:
- User: Who gets the credits
- Algorithm: Which algorithm the credits are for
- Credits: How many credits they get
- Valid period: When the credits can be used (from/until dates)
- Comment: Documentation of why these credits were granted

When a user tries to run an algorithm, the system:
1. Checks for any active algorithm-specific credits
2. If none exist, falls back to the default monthly credits
3. Calculates remaining jobs based on the algorithm's credits-per-job cost

Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/361


app/grandchallenge/algorithms/models.py
```diff
@@ -584,15 +663,68 @@ def get_remaining_complimentary_jobs(self, *, user):
...
+            return 0
+
+        user_credits = settings.ALGORITHMS_GENERAL_CREDITS_PER_MONTH_PER_USER
+
+        user_algorithms_with_active_credits = (
+            AlgorithmUserCredit.objects.active_credits()
+            .filter(
+                user=user,
+            )
+            .select_related("algorithm")
```
amickan, line 706: Not sure how much time it really saves, but I think it's be more efficient if you just returned the algorithm pk here: 
```suggestion
            .values_list("algorithm__pk", flat=True)
```

and then in the next query filter on those:
`.exclude(algorithm_image__algorithm__pk__in=user_algorithms_with_active_credits)`

amickan (approved):
Looks good! Very nice to now have a specific way of granting extra credits to users that makes sure that they are time bound and for a specific algorithm, rather than across the board and for all time like they used to be.

app/grandchallenge/algorithms/models.py
```diff
@@ -584,15 +663,68 @@ def get_remaining_complimentary_jobs(self, *, user):
...
+                creator=user,
+                is_complimentary=False,
+                created__gte=timezone.now() - relativedelta(months=1),
+            )
+            .exclude(
+                algorithm_image__algorithm__in=[
+                    credits.algorithm
+                    for credits in user_algorithms_with_active_credits
+                ]
+            )
```
jmsmkn, line 720:
```suggestion
            .exclude(
                algorithm_image__algorithm__pk__in=user_algorithms_with_active_credits
            )
```

PR #3642, opened 2024-10-18
Add algorithm template forge JSON

chrisvanrun:
MVP for using future algorithm template forging in `grand-challenge-forge.`

Part of the pitch:

It adds:
- An algorithm-template context generator for future forging
- A site admin readable field for `Algorithm` for debugging integration with forge
- Adds example values to the context of the challenge json for the forge

### DIFF
--- a/app/grandchallenge/core/utils/grand_challenge_forge.py
+++ b/app/grandchallenge/core/utils/grand_challenge_forge.py
@@ -66,6 +66,7 @@ def get_forge_algorithm_template_context(algorithm):
     return {
         "algorithm": {
             "title": algorithm.title,
+            "slug": algorithm.title,
             "url": algorithm.get_absolute_url(),
             "inputs": [
                 _process_component_interface(ci)
--- a/app/grandchallenge/core/utils/grand_challenge_forge.py
+++ b/app/grandchallenge/core/utils/grand_challenge_forge.py
@@ -1,8 +1,22 @@
 from grandchallenge.evaluation.utils import SubmissionKindChoices
-from grandchallenge.subdomains.utils import reverse
 
 
-def get_forge_json_description(challenge, phase_pks=None):
+def _process_component_interface(component_interface):
+    result = {
+        "slug": component_interface.slug,
+        "kind": component_interface.get_kind_display(),
+        "super_kind": component_interface.super_kind.label,
+        "relative_path": component_interface.relative_path,
+        "example_value": None,
+    }
+
+    if component_interface.is_json_kind:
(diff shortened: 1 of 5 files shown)


app/grandchallenge/core/utils/grand_challenge_forge.py
```diff
@@ -21,35 +35,45 @@ def get_forge_json_description(challenge, phase_pks=None):
...
+            "inputs": [
+                _process_component_interface(ci)
+                for ci in algorithm.inputs.all()
+            ],
+            "outputs": [
+                _process_component_interface(ci)
+                for ci in algorithm.outputs.all()
+            ],
+        }
+    }
```
amickan, line 79: Don't you also want the `slug` in here?

chrisvanrun: What do you imagine we'd need the `slug` for? We won't actually be interfacing with the algorithm on GC. 

I had the idea in my head we might re-slug the title since we are working locally anyway. That way the slug keeps pointing to the algorithm. There is the trouble with duplicates then, but I honestly think the chances of that are so tiny that it's acceptable to keep things recognizable locally.

amickan: I didn't have anything in particular in mind, I just figured that since the slug is the unique, human readable identifier of the algorithm it probably should be in there (just like you do for the phases and archives), but I trust your intuition here if you think you don't need it.

amickan: But just for my understanding: why re-slug the title if you could just pass along the slug? What do you need the title for?

chrisvanrun: Slug is only created at creation time and can never be updated. The title is the most current 'human' name that the algorithm carries. We see that a lot in challenge phases, I think it also applies to algorithms.

app/grandchallenge/core/utils/grand_challenge_forge.py
```diff
@@ -21,35 +35,46 @@ def get_forge_json_description(challenge, phase_pks=None):
...
             "archives": [process_archive(a) for a in archives],
         }
     }
+
+
+def get_forge_algorithm_template_context(algorithm):
+    return {
+        "algorithm": {
+            "title": algorithm.title,
+            "slug": algorithm.title,
```
jmsmkn, line 69: Sneaky 😉

```suggestion
            "slug": algorithm.slug,
```

amickan (approved):
LGTM!

PR #3653, opened 2024-10-24
Use full page markdown editor for page editing view

koopmant:
This PR adds the usage of the full page markdown editor for page editing and thereby splits up the page editing into a metadata edit and a page content edit. 

related to https://github.com/DIAGNijmegen/rse-roadmap/issues/351

### DIFF
--- a/app/grandchallenge/core/widgets.py
+++ b/app/grandchallenge/core/widgets.py
@@ -102,7 +102,7 @@ class MarkdownEditorFullPageWidget(MarkdownEditorInlineWidget):
     template_name = "markdownx/full_page_widget.html"
 
     def __init__(self, attrs=None):
-        default_attrs = {"rows": "40"}
+        default_attrs = {"rows": "30"}
         if attrs:
             default_attrs.update(attrs)
         super().__init__(default_attrs)
--- a/app/grandchallenge/pages/templates/pages/page_content_update.html
+++ b/app/grandchallenge/pages/templates/pages/page_content_update.html
@@ -1,4 +1,4 @@
-{% extends "pages/challenge_settings_base.html" %}
+{% extends "base.html" %}
 {% load crispy from crispy_forms_tags %}
 {% load static %}
 {% load url %}
@@ -28,7 +28,9 @@
     </ol>
 {% endblock %}
 
-{% block content %}
+{% block container %}container-fluid{% endblock %}
+
+{% block outer_content %}
 
     <h2>{% if object %}Update{% else %}Create{% endif %} Page</h2>
 
--- a/app/grandchallenge/pages/templates/pages/page_content_update.html
+++ b/app/grandchallenge/pages/templates/pages/page_content_update.html
@@ -4,7 +4,7 @@
 {% load url %}
 
 {% block title %}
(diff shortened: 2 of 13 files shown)


app/grandchallenge/pages/views.py
```diff
@@ -56,6 +60,15 @@ def form_valid(self, form):
...
         return super().form_valid(form)
 
+    def get_success_url(self):
+        return reverse(
+            "pages:update",
+            kwargs={
+                "challenge_short_name": self.object.challenge.short_name,
+                "slug": self.object.slug,
+            },
+        )
```
jmsmkn, line 70:
```suggestion
        return page.get_absolute_url()
```

app/grandchallenge/core/widgets.py
```diff
@@ -102,7 +102,7 @@ class MarkdownEditorFullPageWidget(MarkdownEditorInlineWidget):
     template_name = "markdownx/full_page_widget.html"
 
     def __init__(self, attrs=None):
-        default_attrs = {"rows": "40"}
+        default_attrs = {"rows": "30"}
```
jmsmkn, line 105: Does this need to be set or will the widget take up as much space as possible?

app/grandchallenge/pages/templates/pages/page_content_update.html
```diff
@@ -0,0 +1,43 @@
+{% extends "base.html" %}
+{% load crispy from crispy_forms_tags %}
+{% load static %}
+{% load url %}
+
+{% block title %}
+    {% if object %}Update{% else %}Create{% endif %} -{% if object %} {{ object|title }} -{% endif %} {{ block.super }}
```
jmsmkn, line 7: Do the title and breadcrumbs make sense for this template?

app/grandchallenge/pages/views.py
```diff
@@ -116,6 +130,23 @@ def form_valid(self, form):
...
+):
+    model = Page
+    form_class = PageContentUpdateForm
+    template_name_suffix = "_content_update"
+    permission_required = "change_challenge"
+    raise_exception = True
+    login_url = reverse_lazy("account_login")
+
+    def get_permission_object(self):
+        return self.request.challenge
```
jmsmkn, line 147: I think that it is safer to use `self.object.challenge`.

app/grandchallenge/core/widgets.py
koopmant: It will not take up as much space as possible. I set this as an initial size, but the user can resize the widget. The reason the widget has a fixed size is so that the user can have two independent scroll areas to line up the markdown with the preview content. (Scrolling is not linked btw.) 
If we let the widget take up the space it needs, the user can only use the scroll bar of the page. This means that if the markdown does not line up with the preview content, they still need to scroll up and down to edit and preview.

app/grandchallenge/pages/templates/pages/page_content_update.html
koopmant: Thank you. Removed.

jmsmkn (approved):

PR #3657, opened 2024-10-25
Handle empty pages in datatables

chrisvanrun:
A slight race condition can occur when searching for a term and requesting the next page on any datatable. 

See sentry issue: https://grand-challenge.sentry.io/issues/6016634627/?project=303639&query=%21logger%3Acsp%20is%3Aunresolved&referrer=issue-stream&statsPeriod=24h&stream_index=0

This fixes it by simply forcing the showing of the last page when that happens.

### DIFF
--- a/app/grandchallenge/datatables/views.py
+++ b/app/grandchallenge/datatables/views.py
@@ -2,6 +2,7 @@ from dataclasses import dataclass
 from functools import reduce
 from operator import or_
 
+from django.core.paginator import EmptyPage
 from django.db.models import Q
 from django.http import JsonResponse
 from django.template.loader import render_to_string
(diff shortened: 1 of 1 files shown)


app/grandchallenge/datatables/views.py
```diff
@@ -54,7 +55,12 @@ def draw_response(self, *, form_data):
         order_by = f"{'-' if order_dir == 'desc' else ''}{order_by}"
         data = self.filter_queryset(self.object_list, search, order_by)
         paginator = self.get_paginator(queryset=data, per_page=page_size)
-        objects = paginator.page(page)
+
+        try:
+            objects = paginator.page(page)
+        except EmptyPage:
+            # If the page is out of range, show the last page
+            objects = paginator.page(paginator.num_pages)
```
jmsmkn, line 63: Not too sure if this is the solution. Is there a risk the page indicator could get out of sync?

chrisvanrun: I don't think so: the page number doesn't exist anymore anyway in the page indicator ui: there simply aren't that many pages for that search result. Not sure how the JS handles the indicator internally, but I haven't been able to catch it going out of sync.

jmsmkn (approved):

app/grandchallenge/datatables/views.py
jmsmkn: Okay, doesn't feel great but lets see!

PR #3659, opened 2024-10-27
Fix auto accepting pending requests upon verification

ammar257ammar:
This fix relates to issue #2461 and PR #3514. The issue is that the logic of auto accepting pending requests upon verification was implemented in the save() method of the model. However, in Admin, the actions are performed as an update on the queryset which does not trigger the save() method, and similar behaviour affected the proper handling of permissions since the queryset update did not trigger the pre_save signals.

### DIFF
--- a/app/grandchallenge/verifications/admin.py
+++ b/app/grandchallenge/verifications/admin.py
@@ -17,7 +17,9 @@ from grandchallenge.verifications.models import (
     permissions=("change",),
 )
 def mark_verified(modeladmin, request, queryset):
-    for verification in queryset:
+    verifications = queryset.filter(email_is_verified=True)
+
+    for verification in verifications:
         verification.is_verified = True
         verification.verified_at = now()
         verification.save()
--- a/app/grandchallenge/verifications/admin.py
+++ b/app/grandchallenge/verifications/admin.py
@@ -17,9 +17,7 @@ from grandchallenge.verifications.models import (
     permissions=("change",),
 )
 def mark_verified(modeladmin, request, queryset):
-    verifications = queryset.filter(email_is_verified=True)
-
-    for verification in verifications:
+    for verification in queryset.filter(email_is_verified=True):
         verification.is_verified = True
         verification.verified_at = now()
         verification.save()
--- a/app/grandchallenge/verifications/admin.py
+++ b/app/grandchallenge/verifications/admin.py
@@ -17,9 +17,10 @@ from grandchallenge.verifications.models import (
     permissions=("change",),
 )
 def mark_verified(modeladmin, request, queryset):
-    queryset.filter(email_is_verified=True).update(
-        is_verified=True, verified_at=now()
-    )
+    for verfication_request in queryset:
+        verfication_request.is_verified = True
+        verfication_request.verified_at = now()
+        verfication_request.save()
 
 
 @admin.action(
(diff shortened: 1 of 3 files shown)


app/tests/verification_tests/test_admin.py
```diff
@@ -52,3 +78,71 @@ def test_deactivate_users(django_capture_on_commit_callbacks, settings):
...
+    usr = UserFactory()
+
+    t = entity_factory(access_request_handling=access_request_handling)
+    pr = perm_request_factory(**{"user": usr, perm_request_entity_attr: t})
+
+    VerificationFactory(user=usr, email_is_verified=True, is_verified=False)
+
+    assert pr.status == expected_request_status_without_verification
+
+    with django_capture_on_commit_callbacks(execute=True):
```
jmsmkn, line 139: Why is `django_capture_on_commit_callbacks` necessary here?

app/grandchallenge/verifications/admin.py
```diff
@@ -17,9 +17,10 @@
     permissions=("change",),
 )
 def mark_verified(modeladmin, request, queryset):
-    queryset.filter(email_is_verified=True).update(
-        is_verified=True, verified_at=now()
-    )
+    for verfication_request in queryset:
```
jmsmkn, line 20: Variable names should follow their class names. So this should be `verification`, not `verification_request`.

ammar257ammar: Noted, and fixed.

app/tests/verification_tests/test_admin.py
ammar257ammar: Looking at it again, it is not necessary. There is no `on_commit` callbacks in the underlying function to justify its use.
Fixed it.

jmsmkn: Did you copy this test from somewhere? Is it also redundant there?

ammar257ammar: I made tests earlier on the model level to test the save() of the model after adding the implementation ([test_models.py](https://github.com/comic/grand-challenge.org/blob/bc13992a12eef6985c2d3ca77025f3891703f755/app/tests/verification_tests/test_models.py#L75)). I used the same parametrized tests here with the only difference of calling mark_verified() from admin.py (the changed function) to test it here.
Should there be only a test in either in test_models.py or test_admin.py?

app/grandchallenge/verifications/admin.py
```diff
@@ -17,9 +17,10 @@
     permissions=("change",),
 )
 def mark_verified(modeladmin, request, queryset):
-    queryset.filter(email_is_verified=True).update(
-        is_verified=True, verified_at=now()
-    )
+    for verfication in queryset:
+        verfication.is_verified = True
+        verfication.verified_at = now()
+        verfication.save()
```
amickan, line 23: There is a typo here. `verfication` --> `verification`

ammar257ammar: Fixed

app/grandchallenge/verifications/admin.py
```diff
@@ -17,9 +17,10 @@
...
```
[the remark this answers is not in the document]
ammar257ammar: oh, you are right!
I will change it.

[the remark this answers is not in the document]
ammar257ammar: Fixed it

amickan (approved):

jmsmkn:
@ammar257ammar what do you want to do with this? If it is ready for review then you need to mark it as ready for review.

PR #3660, opened 2024-10-27
Add help text with interface list link to CIV create forms

### LINKED ISSUE #3555: Link to interface list when updating displaysets
see title.

ammar257ammar:
closes #3555

### DIFF
--- a/app/grandchallenge/archives/views.py
+++ b/app/grandchallenge/archives/views.py
@@ -716,6 +716,9 @@ class ArchiveItemInterfaceCreate(InterfacesCreateBaseView):
     def base_object(self):
         return Archive.objects.get(slug=self.kwargs["slug"])
 
+    def get_interface_viewname(self):
+        return "components:component-interface-list-archives"
+
     def get_htmx_url(self):
         if self.kwargs.get("pk") is not None:
             return reverse_lazy(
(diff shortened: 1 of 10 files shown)


app/grandchallenge/archives/views.py
```diff
@@ -716,6 +716,9 @@ def object(self):
     def base_object(self):
         return Archive.objects.get(slug=self.kwargs["slug"])
 
+    def get_interface_viewname(self):
+        return "components:component-interface-list-archives"
```
amickan, line 720: I think this would make more sense as a model property on the base model. Then it you also don't need to pass it to the form, but can just just call `base_object.interface_viewname` there.

app/tests/archives_tests/test_views.py
```diff
@@ -1458,3 +1458,24 @@ def test_item_job_list_row_template_ajax_renders(client):
...
+        method=client.get,
+        reverse_kwargs={
+            "slug": ar.slug,
+        },
+        user=u,
+    )
+    assert (
+        reverse("components:component-interface-list-archives")
+        in response.rendered_content
+    )
```
amickan, line 1481: Any specific reason why you didn't parametrize this test (one test for both reader studies and archives)?

ammar257ammar: ah, good point. I was thinking of keeping each test in its own app. Should I then move this test to the components app and cover both cases in a parametrized test?

amickan: Yes, that sounds like a good place. I think we have a bunch of parametrized tests for archive items and display sets there as well.

app/grandchallenge/archives/views.py
ammar257ammar: I agree, done.

amickan (approved):

PR #3663, opened 2024-10-28
Independent Algorithm Templates

chrisvanrun:
This PR hooks grand-challenge-forge into grand-challenge for the first time.

Part of the pitch:

It adds a tab and button to download a ZIP archive that contain the algorithm template files:

Locally, the creation + downloading of the ZIP seems to run under 90ms and generates:


jmsmkn:
The poetry.lock file got messed up in this PR, I cannot work out how to fix it so I cherry-picked the relevant commits into #3665.

PR #3665, opened 2024-10-29
Independent Algorithm Templates

jmsmkn:
This PR hooks grand-challenge-forge into grand-challenge for the first time.

Part of the pitch:

It adds a tab and button to download a ZIP archive that contain the algorithm template files:

Locally, the creation + downloading of the ZIP seems to run under 90ms and generates:

### DIFF
--- a/app/grandchallenge/algorithms/views.py
+++ b/app/grandchallenge/algorithms/views.py
@@ -1033,8 +1033,8 @@ class AlgorithmImageTemplate(ObjectPermissionRequiredMixin, DetailView):
     permission_required = "algorithms.change_algorithm"
     raise_exception = True
     queryset = Algorithm.objects.prefetch_related(
-        "inputs__json_kind_example",
-        "outputs__json_kind_example",
+        "inputs",
+        "outputs",
     )
 
     def get(self, *_, **__):
--- a/app/grandchallenge/algorithms/views.py
+++ b/app/grandchallenge/algorithms/views.py
@@ -1,9 +1,6 @@
-import io
 import logging
-import os
 import tempfile
 from pathlib import Path
-from zipfile import ZipFile
 
 from django.conf import settings
 from django.contrib import messages
@@ -13,11 +10,7 @@ from django.contrib.auth.mixins import (
 )
 from django.contrib.messages.views import SuccessMessageMixin
 from django.core.cache import cache
-from django.core.exceptions import (
-    PermissionDenied,
-    SuspiciousFileOperation,
-    ValidationError,
-)
+from django.core.exceptions import PermissionDenied, ValidationError
 from django.db.models import Window
 from django.db.models.functions import Rank
 from django.forms.utils import ErrorList
@@ -93,6 +86,7 @@ from grandchallenge.core.templatetags.random_encode import random_encode
 from grandchallenge.core.utils.grand_challenge_forge import (
     get_forge_algorithm_template_context,
 )
+from grandchallenge.core.utils.zip import zip_memory_buffer
 from grandchallenge.core.views import PermissionRequestUpdate
 from grandchallenge.datatables.views import Column, PaginatedTableListView
 from grandchallenge.evaluation.models import Evaluation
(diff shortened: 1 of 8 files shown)


app/grandchallenge/algorithms/views.py
```diff
@@ -1013,3 +1022,52 @@ def get_success_message(self, cleaned_data):
...
+    permission_required = "algorithms.change_algorithm"
+    raise_exception = True
+    queryset = Algorithm.objects.prefetch_related("inputs", "outputs")
+
+    def get(self, *_, **__):
+        algorithm = self.get_object()
+
+        forge_context = get_forge_algorithm_template_context(algorithm)
+
+        # Step 1: Create a temporary directory for generating content
```
jmsmkn, line 1038: The LLM comments can be removed. L1039 tells a developer that a temporary directory is being generated. If they don't understand it then they can ask an LLM to explain it.

app/grandchallenge/algorithms/views.py
```diff
@@ -1013,3 +1022,52 @@ def get_success_message(self, cleaned_data):
...
+                output_path=Path(temp_dir),
+            )
+
+            # Step 3: Create an in-memory buffer to store the ZIP file
+            zip_buffer = io.BytesIO()
+
+            # Step 4: Create the ZIP archive in memory
+            with ZipFile(zip_buffer, "w") as zipf:
+                for foldername, _, filenames in os.walk(
+                    algorithm_template_path
```
jmsmkn, line 1052: How are we certain that the `algorithm_template_path` is safe? We should only package files that are in `temp_dir`. We should double check that the file that is being included are actually in `temp_dir`.

app/grandchallenge/algorithms/views.py
```diff
@@ -1013,3 +1022,52 @@ def get_success_message(self, cleaned_data):
...
+                ):
+                    for filename in filenames:
+                        file_path = os.path.join(foldername, filename)
```
jmsmkn, line 1055: Django has a function called `safe_join` that you can use in order to ensure that a file is within a certain directory.

app/tests/algorithms_tests/test_views.py
```diff
@@ -2018,3 +2020,57 @@ def test_update_view_limits_gpu_choice(client):
     )
 
     assert response.status_code == 302
+
+
+@pytest.mark.django_db
+def test_algorithm_template_download(client):
```
jmsmkn, line 2026: Think AAA. Probably better to split this into two tests - one for permissions, one for functionality.

app/grandchallenge/algorithms/views.py
chrisvanrun: safe_join is under django.utils._os ... not sure if the _ is indicative of internal use. Given the documentation incorrectly mentions a raised ValueError I've taken the precaution of implementing the check myself but have taken over the exception it raises. Imho it's applicable.

jmsmkn: Yes fine for us to use it, we've been using it for years and I think the API is pretty stable.

app/grandchallenge/algorithms/views.py
```diff
@@ -1013,3 +1026,56 @@ def get_success_message(self, cleaned_data):
...
+        buffer = io.BytesIO()
+        with ZipFile(buffer, "w") as zipf:
+            for foldername, _, filenames in os.walk(
+                source_path, followlinks=True
+            ):
+                for filename in filenames:
+                    file_path = Path(foldername) / filename
+                    file_path = file_path.resolve()
+
+                    if allowed_base_path not in file_path.parents:
```
jmsmkn, line 1072: I would be more comfortable with using Django's implementation rather than implementing this ourselves.

app/tests/algorithms_tests/test_views.py
chrisvanrun: Looked at it afresh: turns out there was already a collection of 'check permission vs views' in the algorithms. I've added the permission checking to that.

app/grandchallenge/algorithms/views.py
chrisvanrun: Took a bit of head scratching but I decided to sidestep the issue all together: it now packs all the content of the `temp_dir`. This entails an extra subdirectory in the ZIP but simplifies the trust issue.

I kep the checks on the resolved path to catch relative symlinks which might be a potential vector for unintended additions to the ZIP.

chrisvanrun:
Mhh, pre-commit is acting up. Giving a nudge.

jmsmkn (commented):
Can't approve this as I opened the PR but LGMT

Issue #3639, opened 2024-10-17
Users get confused by spinning status


chrisvanrun:
An observation form the case study: 'A workshop dry-run'. They seem to call it a problem on GC, and refreshing sometimes helps.

Perhaps we should remove the spinner, or actually make it auto-update?

amickan:
Which spinner are you referring to here?

jmsmkn:
We have a way of doing this now - make the component job status button a HTMX component. Would need the correct permissions etc. @koopmant I think this would be a nice thing for you to do for Algorithm and Evaluation tasks after you are done with the new page edit views.

chrisvanrun:
> Which spinner are you referring to here?

Several. Non-exhaustive list on the top of my head:
- Several job states
- Container Image states

User in this case was referring to the job states.

amickan:
That would be very nice to get before the workshop on November 7.

jmsmkn:
We discussed splitting this into several PRs:

1. Get the existing algorithm job status buttons rendered with HTMX when the page loads/when they are visible. Add a detail view to return this from a template and ensure the permissions are tested. Check where the current badges are rendered, we think that they're an include, if not, they should be.
2. Update the button to auto-refresh if the job is in a non-final state (success, failed, cancelled)
3. Repeat the same for evaluation jobs, using the fact these are both component jobs.
4. Do 1+2 for container image imports.
5. Make the pre-requisite jobs list an auto-refreashing htmx table.

koopmant:
I believe this is complete now. We have auto-refresh badges using the following new views:

Job status:
JobStatusDetail
EvaluationStatusDetail

Build status:
AlgorithmImageBuildStatusDetail

Import status:
AlgorithmImageImportStatusDetail
AlgorithmModelImportStatusDetail
WorkstationImageImportStatusDetail
MethodImportStatusDetail
EvaluationGroundTruthImportStatusDetail

The badges reload every 30s until in a final state. 

The pre-requisite jobs table in the evaluation detail now also automatically refreshes every 3 minutes. When there are no jobs left, the table disappears.

jmsmkn:
Indeed, @chrisvanrun and @amickan please check that everything is working as expected ASAP.

amickan:
I tried out uploading an algorithm image just now and running a job - status spinners update nicely! 

[...]

PR #3667, opened 2024-10-29
Move job status badge to separate view and load with htmx

koopmant:
Get the existing algorithm job status buttons rendered with HTMX when the page loads/when they are visible. Add a detail view to return this from a template and ensure the permissions are tested. Check where the current badges are rendered, we think that they're an include, if not, they should be.

part of #3639

### DIFF
--- a/app/grandchallenge/algorithms/templates/algorithms/job_list_row.html
+++ b/app/grandchallenge/algorithms/templates/algorithms/job_list_row.html
@@ -27,7 +27,7 @@
 
 <split></split>
 
-<div hx-get="{% url 'algorithms:job-status-badge-detail' slug=object.algorithm_image.algorithm.slug pk=object.pk %}" hx-trigger="revealed" hx-swap="outerHTML"></div>
+{% include "algorithms/job_status_badge_detail.html" %}
 <split></split>
 
 {% if object.public %}
--- a/app/grandchallenge/algorithms/templates/algorithms/job_list_row.html
+++ b/app/grandchallenge/algorithms/templates/algorithms/job_list_row.html
@@ -27,7 +27,7 @@
 
 <split></split>
 
-{% include "algorithms/job_status_badge_detail.html" %}
+<div hx-get="{% url 'algorithms:job-status-badge-detail' slug=object.algorithm_image.algorithm.slug pk=object.pk %}" hx-trigger="revealed" hx-swap="outerHTML"></div>
 <split></split>
 
 {% if object.public %}
--- a/app/grandchallenge/algorithms/templates/algorithms/job_list_row.html
+++ b/app/grandchallenge/algorithms/templates/algorithms/job_list_row.html
@@ -27,13 +27,7 @@
 
 <split></split>
 
-<span class="badge badge-{{ object.status_context }}">
-    {% if object.animate %}
-        <span class="spinner-border spinner-border-sm" role="status" aria-hidden="true"></span>
-    {% endif %}
-    {{ object.get_status_display }}{% if object.status == object.SUCCESS and object.stderr %},
-        with warnings{% endif %}
-</span>
+{% include "algorithms/job_status_badge_detail.html" %}
 <split></split>
 
 {% if object.public %}
--- a/app/grandchallenge/algorithms/templates/algorithms/job_status_badge_detail.html
+++ b/app/grandchallenge/algorithms/templates/algorithms/job_status_badge_detail.html
@@ -1,4 +1,4 @@
-<span class="badge badge-{{ object.status_context }}" {% if not object.finished %}hx-get="{% url 'algorithms:job-status-badge-detail' slug=object.algorithm_image.algorithm.slug pk=object.pk %}" hx-trigger="every {{ delay }}s" hx-swap="outerHTML"{% endif %}>
+<span class="badge badge-{{ object.status_context }}" {% if not object.finished %}hx-get="{% url 'algorithms:job-status-badge-detail' slug=object.algorithm_image.algorithm.slug pk=object.pk %}" hx-trigger="every {{ delay|default:"5" }}s" hx-swap="outerHTML"{% endif %}>
     {% if object.animate %}
         <span class="spinner-border spinner-border-sm" role="status" aria-hidden="true"></span>
     {% endif %}
--- a/app/grandchallenge/algorithms/templates/algorithms/job_status_badge_detail.html
+++ b/app/grandchallenge/algorithms/templates/algorithms/job_status_badge_detail.html
@@ -1,9 +1,8 @@
-{% with delay_options="20 21 22 23 24 25 26 27 28 29 30 31 32 33 34 35 36 37 38 39 40" %}
-    <span class="badge badge-{{ object.status_context }}" {% if not object.finished %}hx-get="{% url 'algorithms:job-status-badge-detail' slug=object.algorithm_image.algorithm.slug pk=object.pk %}" hx-trigger="every {{ delay_options.split|random }}s" hx-swap="outerHTML"{% endif %}>
-        {% if object.animate %}
-            <span class="spinner-border spinner-border-sm" role="status" aria-hidden="true"></span>
-        {% endif %}
-        {{ object.get_status_display }}{% if object.status == object.SUCCESS and object.stderr %},
-            with warnings{% endif %}
-    </span>
-{% endwith %}
+{% load job_status_badge %}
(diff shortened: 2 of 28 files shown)


jmsmkn:
This looks good! Remember to mark when it is ready for review. There is at least one other usage of this button in `evaluation_detail.html`. Did you test this out with multiple pages of results in datatables?

koopmant:
> This looks good! Remember to mark when it is ready for review. There is at least one other usage of this button in `evaluation_detail.html`. Did you test this out with multiple pages of results in datatables?

Tested with pages in datatables this morning. I have now hooked to the draw event using the drawCallback init option. see `datatables/list.mjs`. 

I found 21 similar usages when looking for `class="spinner`. Maybe it's best if a tackle those together in a new PR?

jmsmkn:
I think it requires a lot of manual inspection as we have spinners for Evaluations, Algorithm Image, Methods, Models and Ground Truth imports at least, probably others. Just forus on the ones that you're 100% sure are algorithm jobs for now.

app/grandchallenge/algorithms/templates/algorithms/job_list_row.html
```diff
@@ -27,13 +27,7 @@
...
 <split></split>
 
-<span class="badge badge-{{ object.status_context }}">
-    {% if object.animate %}
-        <span class="spinner-border spinner-border-sm" role="status" aria-hidden="true"></span>
-    {% endif %}
-    {{ object.get_status_display }}{% if object.status == object.SUCCESS and object.stderr %},
-        with warnings{% endif %}
-</span>
+<div hx-get="{% url 'algorithms:job-status-badge-detail' slug=object.algorithm_image.algorithm.slug pk=object.pk %}" hx-trigger="revealed" hx-swap="outerHTML"></div>
```
jmsmkn, line 30: Is there a way to do this without htmx on load? Other wise we will hammer the server with these requests.

app/grandchallenge/algorithms/urls.py
```diff
@@ -104,6 +105,11 @@
     path("<slug>/jobs/", JobsList.as_view(), name="job-list"),
     path("<slug>/jobs/create/", JobCreate.as_view(), name="job-create"),
     path("<slug>/jobs/<uuid:pk>/", JobDetail.as_view(), name="job-detail"),
+    path(
+        "<slug>/jobs/<uuid:pk>/status-badge",
```
jmsmkn, line 109: URLs should always end with a `/`

```suggestion
        "<slug>/jobs/<uuid:pk>/status-badge/",
```

app/grandchallenge/algorithms/templates/algorithms/job_list_row.html
koopmant: I'm not sure yet. I did it this way to get the random delay in the context generated by the view. One way I could do it is by generating the random delay in the template itself, which is also possible (although not very pretty). 
That would be the only way to avoid multiple requests I think...

jmsmkn: Why can't you use `{% include "algorithms/job_status_badge_detail.html" %}` here? That way you would avoid multiple requests for completed jobs.

Then, for generating a random delay rather than setting the context on the view you could add a [custom template tag](https://docs.djangoproject.com/en/5.1/howto/custom-template-tags/#writing-custom-template-tags) for `randint` so that you could do:

```
hx-trigger="every {% randint 20 40 %}s"
```

in your template.

jmsmkn: Updated the example to use delay as the outerHTML is being replaced here, see https://htmx.org/attributes/hx-trigger/. Would that still work with the new badges brought in?

jmsmkn: Also, interestingly in the HTMX docs there is a queue option, so we might not need to jitter the requests at all, could you test with a straight up 30s delay with multiple buttons on a page and check the timing diagram?

koopmant: Thanks for the pointer to custom template tags. That is much nicer than what I had. 

I replaced `every` with a `delay` modifier on the load event. `hx-trigger="load delay:{% randint 20 40 %}"`
(This does not work with the "revealed" event btw. If I use that, delay is ignored and the request triggers constantly.)

koopmant: > Also, interestingly in the HTMX docs there is a queue option, so we might not need to jitter the requests at all, could you test with a straight up 30s delay with multiple buttons on a page and check the timing diagram?

I read about the queue modifier in hx-trigger and hx-sync. Is that the option you mean? 

With ten badges with a delay of 30s I see the following:

image

image

jmsmkn: Ah perfect. HTMX is great and we do not need to jitter then, super!

app/grandchallenge/evaluation/templates/evaluation/evaluation_list.html
```diff
@@ -71,12 +71,7 @@ <h2>Submissions and Evaluations for {{ phase.title }}</h2>
                         </td>
                     {% endif %}
                     <td>
-                    <span class="badge badge-{{ evaluation.status_context }}">
-                        {% if evaluation.animate %}
-                            <span class="spinner-border spinner-border-sm" role="status" aria-hidden="true"></span>
-                        {% endif %}
-                        {{ evaluation.get_status_display }}
-                    </span>
+                        {% include "algorithms/job_status_badge_detail.html" with object=evaluation %}
```
jmsmkn, line 74: Note this is for the evaluation status and not the job status, you have not implemented this yet and should follow with another PR.

app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
```diff
@@ -53,13 +53,7 @@ <h2>Evaluation</h2>
...
         <dt>Status</dt>
         <dd>
-            <span class="badge badge-{{ evaluation.status_context }}">
-                {% if evaluation.animate %}
-                    <span class="spinner-border spinner-border-sm" role="status"
-                          aria-hidden="true"></span>
-                {% endif %}
-                {{ evaluation.get_status_display }}
-            </span>
+            {% include "algorithms/job_status_badge_detail.html" with object=evaluation %}
```
jmsmkn, line 56: Note this is for the evaluation status and not the job status, you have not implemented this yet and should follow with another PR.

app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
```diff
@@ -238,13 +232,7 @@ <h3 class="card-title">Prerequisite Jobs</h3>
...
                                 <td>{{ job.created }}</td>
                                 <td>
-                                    <span class="badge badge-{{ job.status_context }}">
-                                    {% if job.animate %}
-                                        <span class="spinner-border spinner-border-sm" role="status"
-                                            aria-hidden="true"></span>
-                                    {% endif %}
-                                    {{ job.get_status_display }}
-                                    </span>
+                                    {% include "algorithms/job_status_badge_detail.html" with object=job %}
```
jmsmkn, line 235: These are algorithm jobs though, so can stay in this PR.

app/grandchallenge/algorithms/templates/algorithms/job_status_badge_detail.html
```diff
@@ -0,0 +1,7 @@
+<span class="badge badge-{{ object.status_context }}" {% if not object.finished %}hx-get="{% url 'algorithms:job-status-badge-detail' slug=object.algorithm_image.algorithm.slug pk=object.pk %}" hx-trigger="load delay:30s" hx-swap="outerHTML"{% endif %}>
```
jmsmkn, line 1: Due to using subdomains we had to re-implement the url tag. We should discuss that part of the code base in our next 1-1 as it affects a lot of things.

```suggestion
{% load url %}

<span class="badge badge-{{ object.status_context }}" {% if not object.finished %}hx-get="{% url 'algorithms:job-status-badge-detail' slug=object.algorithm_image.algorithm.slug pk=object.pk %}" hx-trigger="load delay:30s" hx-swap="outerHTML"{% endif %}>
```

app/grandchallenge/evaluation/templates/evaluation/evaluation_list.html
koopmant: Yes, correct. I was looking to see if it worked because this also inherits from `ComponentJob`. (I was a little surprised that it actually works, but maybe I did not test it thoroughly enough.)

Will revert this.

app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
koopmant: reverted

jmsmkn (approved):

PR #3668, opened 2024-10-29
Add workshop fixtures

amickan:
Adds a script to create all demo challenges for the workshop. 
The script can be run with `python manage.py runscript workshop_fixtures`


scripts/workshop_fixtures.py
```diff
@@ -0,0 +1,201 @@
...
+    "Nova-Nexus",
+    "Omega-Orbit",
+]
+
+WORKSHOP_HOST_USERNAMES = [
+    "amickan",
+    "chris.vanrun.diag",
+    "ammar.ammar",
+    "koopman.t",
+]
```
amickan, line 47: Unsure if this should be hardcoded, or an environment variables that James sets in the shell and that we lookup in the script....

scripts/workshop_fixtures.py
```diff
@@ -0,0 +1,201 @@
...
+    "Gamma-Grid",
+    "Hydra-Haven",
+    "Ion-Infinity",
+    "Jade-Journey",
+    "Kinetic-Key",
+    "Lunar-Link",
+    "Mystic-Matrix",
+    "Nova-Nexus",
+    "Omega-Orbit",
+]
```
amickan, line 40: Challenge title suggestion by Chris :-) This could also be passed as a variable instead of being hard-coded. But updating this script every time we host a new edition is also not that much work. We could even keep the names the same every time and append a year or edition number suffix.

jmsmkn:
This is just a one-off command, right? If so, why do we need it in the codebase? For one-off things we use admin issues.

amickan:
> This is just a one-off command, right? If so, why do we need it in the codebase? For one-off things we use admin issues.

We are likely to run this workshop again, so the codebase seemed like a good place to keep this script.

jmsmkn:
Okay, then the script needs to be updated to that it can be run more than once and is not tied to one particular instance (including in the logos), does not have hardcoded challenge, users, or interfaces.

amickan:
Ok. Thinking about the best way to pass the variable information in: are environment variables the way to go here? 

For the logo and the challenge host usernames, we could source those from the workshop website, and hence pass in the slug of the workshop website challenge. We will always have a workshop website, so that should be future-proof. 

Unsure how to handle the demo fundus image. If we upload it somewhere on the platform (which for the current workshop, we have done), we can reference it through an image pk. For future workshops it would require adding the image manually to some demo archive before running this script. Does that make sense or is there a better way? 

Then the script would require the following arguments / environment variables: `workshop_challenge_slug`, `input_slugs`, `output_slugs`, `challenge_titles` and `demo_image_pk`.

jmsmkn:
IMO the best way to go is to run this as a one off, in which case no PR is required and only an admin issue is. This is completely specific to this instance of GC and this workshop so shouldn't be in the codebase for the platform. Without being able to run scripts locally they cannot be reliably tested. We have to maintain each script, which is a lot of manual work. I think it is unlikely that you will want to run exactly the same script in a years time, there are bound to be changes.

amickan:
Closes in favour of an admin issue: https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/380

PR #3672, opened 2024-10-31
Prevent duplicate notifications for failed image imports

amickan:
I noticed while doing some more local testing for the workshop that failed image uploads (uploading an image that none of our importers can read) results in 2 notifications being sent to users, one from `build_images` and one in `add_image_to_object`. 

This PR adds a check on the upload session status at the beginning of `add_image_to_object`, making sure that we exit this task if the upload session failed. We also check in this task that there is an image for the upload session and we could in theory exit the task there (rather than sending a notification, like we do now), but that seems too indirect of a check to me, since there might be other reasons that the image does not exist (anymore), independent of the upload session status. 

While working on this, I noticed that we mark uploads as successful even if 0 images could be imported. That doesn't really make sense. I checked in the codebase for dependencies on that assumption, but didn't find any, so I'm proposing to mark those as failed.

### DIFF
--- a/app/grandchallenge/cases/tasks.py
+++ b/app/grandchallenge/cases/tasks.py
@@ -180,7 +180,7 @@ def build_images(  # noqa:C901
                 upload_session=upload_session,
             )
 
-        if len(upload_session.import_result["consumed_files"]) == 0:
+        if upload_session.image_set.count() == 0:
             error_handler.handle_error(
                 interface=ci,
                 error_message=upload_session.default_error_message,
--- a/app/grandchallenge/cases/tasks.py
+++ b/app/grandchallenge/cases/tasks.py
@@ -170,25 +170,16 @@ def build_images(  # noqa:C901
         with TemporaryDirectory() as tmp_dir:
             tmp_dir = Path(tmp_dir).resolve()
             _populate_tmp_dir(tmp_dir, upload_session)
-            _handle_raw_image_files(
-                tmp_dir=tmp_dir,
+            importer_result = import_images(
+                input_directory=tmp_dir, origin=upload_session
+            )
+            _handle_raw_files(
+                consumed_files=importer_result.consumed_files,
+                file_errors=importer_result.file_errors,
+                base_directory=tmp_dir,
                 upload_session=upload_session,
             )
-
-        if (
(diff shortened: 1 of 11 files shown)


app/grandchallenge/cases/tasks.py
```diff
@@ -208,27 +208,25 @@ def build_images(  # noqa:C901
...
-        upload_session=upload_session,
-    )
-
-    _delete_session_files(upload_session=upload_session)
+    finally:
+        upload_session.user_uploads.all().delete()
+
+    if (
+        len(upload_session.import_result["file_errors"]) != 0
+        and len(upload_session.import_result["consumed_files"]) == 0
```
jmsmkn, line 221: Why do we need the check on both? I would have thought that `len(upload_session.import_result["consumed_files"]) == 0` would be enough?

jmsmkn:
Good change, I added the success path out of the parts that should be handled with the exception blocks and removed some repeated calls to remove the upload session files.

app/grandchallenge/cases/tasks.py
amickan: True, I wasn't sure, and went the save route in checking both, but I think you're right that there will always be file errors if there are 0 consumed files, so just checking for the consumed files should be enough.

app/grandchallenge/components/tasks.py
```diff
@@ -1192,6 +1192,10 @@ def add_image_to_object(  # noqa: C901
     )
     error_handler = object.get_error_handler(linked_object=upload_session)
 
+    if upload_session.status == upload_session.FAILURE:
+        logger.info("Nothing to do: upload session failed.")
+        return
```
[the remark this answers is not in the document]
jmsmkn: Agreed

amickan:
Thanks for fixing the tests. I was trying to understand why the import_result didn't persist on the upload_session in the finally block...

jmsmkn:
It didn't exist so was using the default of None, so moved the block back, otherwise would have had to add return statements everywhere. The `try` block is really big now but okay, lets move on.

amickan:
I'm looking at the failed tests.

jmsmkn (approved):

PR #3678, opened 2024-10-31
Remove banner from page editing view

### LINKED ISSUE #3674: Banner ungainly large when editing pages
Banner takes up 50% of the screen, I think the banner might need some restrictions on size:

image

koopmant:
closes #3674

### DIFF
--- a/app/grandchallenge/pages/templates/pages/page_content_update.html
+++ b/app/grandchallenge/pages/templates/pages/page_content_update.html
@@ -30,6 +30,8 @@
 
 {% block container %}container-fluid{% endblock %}
 
+{% block messages %}{% endblock %}
+
 {% block outer_content %}
 
     <h2>Update Page</h2>


koopmant:
I overrode the messages block, so messages and debug statements are removed from this view as well. 

in `base.html`:
```python
            {% block messages %}
                {% include 'challenges/challenge_banner.html' %}
                <div id="messages">
                    {% include "grandchallenge/partials/messages.html" %}
                </div>
                {% include "grandchallenge/partials/debug_statement.html" %}
            {% endblock %}

```

jmsmkn (changes requested):
The messages are still necessary as they can contain form validation information. Probably safest to add a new `challenge_banner` block in `challenge_banner.html` so that you can override that instead.

koopmant:
> The messages are still necessary as they can contain form validation information. Probably safest to add a new `challenge_banner` block in `challenge_banner.html` so that you can override that instead.

I've added the block in `base.html` instead.

jmsmkn (approved):

PR #3679, opened 2024-10-31
Create auto refresh badge for import and build status

koopmant:
Use auto refresh status badge for container imports and algorithm build status. 

part of https://github.com/comic/grand-challenge.org/issues/3639


jmsmkn:
Build status should be done sometime along with Models and Ground Truth to make this complete but it is not required right now. To test the Workstation detail in development you can use the workstation user - see `development_fixtures.py` for a list of the created users in your development environment.

koopmant:
> To test the Workstation detail in development you can use the workstation user - see `development_fixtures.py` for a list of the created users in your development environment.

Thanks, tested it. It's working as expected.

app/grandchallenge/components/models.py
```diff
@@ -2030,7 +2030,20 @@ class Meta:
 
     @property
     def animate(self):
-        return self.import_status == self.ImportStatusChoices.STARTED
+        return self.import_status in {
```
jmsmkn, line 2033: Why change the states when animation is done? Given how the import task works I think the previous was correct.

app/grandchallenge/components/models.py
```diff
@@ -2592,5 +2605,18 @@ def import_status_context(self):
             return "info"
 
     @property
-    def import_in_progress(self):
-        return self.import_status == ImportStatusChoices.INITIALIZED
+    def animate(self):
```
jmsmkn, line 2608: Why change the states when animation is done? Given how the import task works I think the previous was correct.

app/grandchallenge/evaluation/views/__init__.py
```diff
@@ -280,6 +280,19 @@ class MethodDetail(
     login_url = reverse_lazy("account_login")
 
 
+class MethodImportStatusDetail(
+    LoginRequiredMixin,
+    ObjectPermissionRequiredMixin,
+    CachedPhaseMixin,
```
jmsmkn, line 286: Why is `CachedPhaseMixin` needed?

app/grandchallenge/evaluation/views/__init__.py
```diff
@@ -1105,6 +1118,17 @@ def get_context_data(self, **kwargs):
...
 
+class EvaluationGroundTruthImportStatusDetail(
+    LoginRequiredMixin,
+    ObjectPermissionRequiredMixin,
+    DetailView,
+):
+    model = EvaluationGroundTruth
+    permission_required = "evaluation.view_evaluationgroundtruth"
+    template_name_suffix = "_import_status_detail"
+    raise_exception = True
```
jmsmkn, line 1129: You might be missing `login_url = reverse_lazy("account_login")`

jmsmkn (commented):
Looks good!

app/grandchallenge/evaluation/views/__init__.py
koopmant: I think it's not needed; but I'm not sure. I saw that it was used for the MethodDetail view and assumed I needed it here as well. But when I look at the commit in which it was added, I don't understand why it is needed for the MethodDetail view at all (and I think it might not be needed there either.)

jmsmkn: Take a read at what the Mixin is doing - it is adding a context variable and form kwargs. You don't have a form in your view so there is no need for the latter, and I don't think that you're using the `phase` variable in your template, so, I do not think that it is necessary.

app/grandchallenge/components/models.py
koopmant: reverted

app/grandchallenge/evaluation/views/__init__.py
koopmant: Thanks for the explanation .I think it's not needed in the MethodDetail view either; but I left it there since it's not part of this PR.

jmsmkn (approved):
Nice work!

PR #3681, opened 2024-11-02
Remember FlexibleImageWidget choice on form redisplay

### LINKED ISSUE #3558: FlexibleImageWidget resets when form is re-rendered due to validation error on another form field
**Describe the bug**
On job creation (or display set / archive item creation / updating), if a user selects an image or uploads a new one and some other interface field in the form fails validation, the image choice isn't remembered and the user has to reupload or reselect the image on the returned form. 

**To Reproduce**
1. Create an algorithm with at least two inputs: one image input and, e.g. a number input. 
2. Upload an image or select an existing image for the image interface. Enter invalid input for the second interface. 
3. Click 'Submit'
4. The form will be re-rendered, displaying the validation error for the second input. The FlexibleImageWidget will now be empty, and you will need to re-select an image or re-upload an image for that interface before being able to re-submit. 

**Expected behavior**
The FlexibleImageWidget should remember the user's choice and should be prepopulated accordingly.

ammar257ammar:
closes #3558


app/grandchallenge/cases/views.py
```diff
@@ -142,11 +144,32 @@ def get(self, request, *args, **kwargs):
...
+        ):
+            # this can happen on the display set update view or redisplay of
+            # form upon validation, where one of the options is the current
+            # image, this enables switching back from one of the above widgets
+            # to the chosen image. In the case of form redisplay on validation
+            # error, current_value will be set in the flexible image widget
+            # select dropdown element with name starting with "WidgetChoice-"
+            # which will not be captured on resubmission, because the
+            # current_value is looked up from the form data element with
+            # interface name (in JobCreateForm and MultipleCIVForm). This make
```
amickan, line 159: What you describe is no longer the case with your changes, right?

app/grandchallenge/components/form_fields.py
```diff
@@ -102,10 +103,19 @@ def get_initial_value(self):
...
 
     def get_image_field(self):
+        current_value = None
+
+        if isinstance(self.initial, ComponentInterfaceValue):
+            current_value = self.initial.image
+        elif Image.objects.filter(pk=self.initial).exists():
+            current_value = Image.objects.filter(pk=self.initial).first()
+        elif UserUpload.objects.filter(pk=self.initial).exists():
+            current_value = UserUpload.objects.filter(pk=self.initial).first()
```
amickan, line 113: We should handle cases other than these 3. 

And since you're checking for the existence of Image and UserUpload, why don't you just use get()?

app/grandchallenge/cases/templates/cases/flexible_image_widget.html
```diff
@@ -5,11 +5,11 @@
         id="widgetSelect-{{ widget.name }}"
         hx-get="{% url 'cases:select-image-widget' %}"
         hx-target="#area-{{ widget.name }}"
-        hx-trigger="widgetSelected, change"
+        hx-trigger="widgetSelected, change, load"
         hx-include="[id='values-{{ widget.name }}']"
 >
     {% if widget.attrs.current_value %}
-        <option value="{{ widget.attrs.current_value.image.pk }}">{{ widget.attrs.current_value }}</option>
+        <option value="{{ widget.attrs.current_value.pk }}">{{ widget.attrs.current_value.title }}</option>
```
amickan, line 12: For my understanding: we don't actually use the `value` of this anywhere, do we? Could it then just be left empty or set to one of the two widget choice types below (depending on the type)?

app/grandchallenge/cases/views.py
ammar257ammar: it is still the case. On redisplay we need a form variable with the prefixed interface slug name to be sent to avoid checking for the "WidgetChoice-" variable.

app/grandchallenge/cases/templates/cases/flexible_image_widget.html
ammar257ammar: I made the change in [Add WidgetChoice for selected image for semantic consistency](https://github.com/comic/grand-challenge.org/pull/3681/commits/833c47a80a86a529aa15b0f9d6f0981670d6e7fe)
It is not much change but semantically more sense instead of printing the pk value without using it anywhere.

app/grandchallenge/components/form_fields.py
```diff
@@ -111,6 +111,8 @@ def get_image_field(self):
             current_value = Image.objects.get(pk=self.initial)
         elif UserUpload.objects.filter(pk=self.initial).exists():
             current_value = UserUpload.objects.get(pk=self.initial)
+        elif self.initial:
```
ammar257ammar, line 114: There is always the case where `self.initial` is not set (None), for example, in create views, and therefore the last conditions should check that the value is there with `elif` instead of `else`

app/grandchallenge/cases/views.py
ammar257ammar: Done

app/grandchallenge/components/form_fields.py
ammar257ammar: Done

app/grandchallenge/cases/views.py
```diff
@@ -144,9 +144,13 @@ def get(self, request, *args, **kwargs):
                 },
             )
             return HttpResponse(html_content)
-        elif current_value and (
-            Image.objects.filter(pk=current_value).exists()
-            or UserUpload.objects.filter(pk=current_value).exists()
+        elif (
+            widget_name == WidgetChoices.IMAGE_SELECTED.name
```
ammar257ammar, line 148: This part of the condition will work anyway without checking for `widget_name` value, but it is more robust to check for it.

app/grandchallenge/components/form_fields.py
ammar257ammar: outdated now

app/grandchallenge/components/form_fields.py
```diff
@@ -102,10 +105,27 @@ def get_initial_value(self):
...
+        current_value = None
+
+        if isinstance(self.initial, ComponentInterfaceValue):
+            current_value = self.initial.image
+        elif self.initial:
+            try:
+                uuid = UUID(self.initial, version=4)
+            except ValueError:
+                uuid = None
+
```
amickan, line 117: Why do you need this? Can't you just check `Image.objects.filter(pk=self.initial).exists()` and `UserUpload.objects.filter(pk=self.initial).exists()` directly? Seems like an unnecessary conversion step here.

app/grandchallenge/components/form_fields.py
```diff
@@ -102,10 +105,27 @@ def get_initial_value(self):
...
+            if uuid and Image.objects.filter(pk=uuid).exists():
+                current_value = Image.objects.get(pk=uuid)
+            elif uuid and UserUpload.objects.filter(pk=uuid).exists():
+                current_value = UserUpload.objects.get(pk=uuid)
+            else:
+                raise RuntimeError(f"Unknown image pk: {self.initial}")
```
amickan, line 123: This is a bit misleading, since `self.initial` does not need to be an image pk. I would just say something along the lines of "Unknown type for initial value"

app/grandchallenge/components/form_fields.py
```diff
@@ -102,10 +105,27 @@ def get_initial_value(self):
             return self.initial
 
     def get_image_field(self):
+        current_value = None
+
+        if isinstance(self.initial, ComponentInterfaceValue):
+            current_value = self.initial.image
+        elif self.initial:
```
amickan, line 112: I think it makes more sense to move this check up. So first check if `self.initial` is not None and then do the type checks or raise a Runtimeerror.

app/tests/cases_tests/test_widget.py
```diff
@@ -148,6 +149,22 @@ def test_flexible_image_widget(client):
...
+        data={
+            f"WidgetChoice-{ci.slug}": WidgetChoices.IMAGE_SELECTED.name,
+            "interface_slug": ci.slug,
+            "current_value": image.pk,
+        },
+    )
+    assert format_html(
+        '<input type="hidden" name="{}" value="{}">', ci.slug, image.pk
+    ) in str(response4.content)
+
```
amickan, line 167: Aren't you missing a test with a user upload pk here as well for completeness?

app/tests/components_tests/test_views.py
```diff
@@ -539,3 +543,74 @@ def test_interfaces_list_link_in_new_interface_form(
...
+        (
+            ArchiveFactory,
+            "archive",
+            ArchiveItemFactory,
+            "archives:item-edit",
+        ),
+    ),
+)
+@pytest.mark.django_db
+def test_image_widget_populated_value_on_update_view_validation_error(
```
amickan, line 566: I think this is only a partial test, right? It should work both on create and on update views, so both types of views need testing.

app/tests/components_tests/test_views.py
```diff
@@ -539,3 +543,74 @@ def test_interfaces_list_link_in_new_interface_form(
...
+        "current_value": image2.pk,
+    }
+    data.update(
+        **get_interface_form_data(interface_slug=image_ci.slug, data=image2.pk)
+    )
+    data.update(
+        **get_interface_form_data(
+            interface_slug=annotation_ci.slug, data='{"1":1}'
+        )
+    )
```
amickan, line 602: Why update the dictionary here and not just construct it in one go?

app/tests/components_tests/test_views.py
```diff
@@ -539,3 +543,74 @@ def test_interfaces_list_link_in_new_interface_form(
...
+        reverse_kwargs={"slug": base_obj.slug, "pk": object.pk},
+        user=editor,
+        method=client.post,
+        follow=True,
+        data=data,
+    )
+    assert response.status_code == 200
+    assert f'<option value="IMAGE_SELECTED">{image2.title}</option>' in str(
+        response.rendered_content
+    )
```
amickan, line 616: Don't you also want to check that this is a hidden field?

app/grandchallenge/components/form_fields.py
ammar257ammar: before I implemented the check `if self.initial`, there was the case when submitting without choosing an image from the image search widget that causes a 500 error failing on `Image.objects.filter(pk=self.initial).exists()` because the submitted value is not a valid UUID (empty string), which also means if any other value submitted (like the screenshot below) will cause a 500 error. Currently, after the check `if self.initial` that case does not happen anymore, but still possible by tampering the form variables.
That is why I thought of checking for a valid UUID first before doing `filter()`.
Is that over protective? should I remove it?

Screenshot 2024-11-06 at 09 22 00

app/grandchallenge/components/form_fields.py
ammar257ammar: fixed

app/grandchallenge/components/form_fields.py
amickan: Ah ok. How was it possible to submit without a chosen image for the ImageSelect widget? I would expect there to be a user error "This field is required" that prevent the user from submitting without a value. 

And as for the possibility of tinkering with the form data, ok, but in that case all we want is to be made aware of this happening, which will happen either way (whether you convert to uuid first and then raise a RuntimeError or whether you let it fail at the query), so I think it is not needed. 

And jut to make sure, if an error is raised here, the form won't be submitted, right? I'm asking because we don't want algorithm jobs stuck in strange states.

app/tests/components_tests/test_views.py
ammar257ammar: the function `get_interface_form_data` already returns a dictionary and it is called twice for each field, so I used update() to append the two dictionaries.

app/grandchallenge/components/form_fields.py
ammar257ammar: Done

app/tests/components_tests/test_views.py
amickan: You're unpacking that dictionary. So you can just do 
```suggestion
    data = {
        "interface_slug": image_ci.slug,
        "current_value": image2.pk,
        **get_interface_form_data(interface_slug=image_ci.slug, data=image2.pk)
        **get_interface_form_data( interface_slug=annotation_ci.slug, data='{"1":1}')
    }
```

ammar257ammar: much cleaner, thanks!

app/tests/components_tests/test_views.py
ammar257ammar: The hidden field already tested in test_widgets.py. Here it tests that the form value is set correctly in the image widget drop down upon validation error. Also, the hidden field is asynchronously populated with htmx so does not appear in the view test response.rendered_content.

app/tests/components_tests/test_views.py
ammar257ammar: Done

app/grandchallenge/components/form_fields.py
ammar257ammar: There was no check for initial value since the value was not captured on form redisplay (that is the main issue), and thus, current_value was None by default.
I fixed it as proposed and it works as intended now.

amickan: Yes, indeed, I understand how the current_value was None on resubmission, but that should still have triggered a ValidationError on the form and prevent submitting. This is what happens on main when I try it out, so I'm trying to understand what happened in your case.

ammar257ammar: That is my point, because `current_value` was None by default and not set to the value of the form element on submission, which in the case of not choosing an image, will be an empty string sent with the form and leading to that exception when trying to check if it is a `pk` for `Image` or `UserUpload`. So now, we need to check that initial is not empty or None before trying to check if it is a `pk` for an `Image` or `UserUpload`.

amickan: Ok, clear.

app/tests/components_tests/test_views.py
amickan: Maybe I'm missing something, but if all you're testing is the widget, it would be more straight forward to test the form directly, rather than go through the view

app/tests/components_tests/test_views.py
```diff
@@ -539,3 +544,103 @@ def test_interfaces_list_link_in_new_interface_form(
...
+    base_obj = base_object_factory()
+    base_obj.add_editor(editor)
+
+    object = object_factory(**{base_obj_lookup: base_obj})
+    object.values.set([image_civ, annotation_civ])
+
+    wt_object = widget_type_object_factory()
+    data = {
+        "interface_slug": image_ci.slug,
+        "current_value": wt_object.pk,
```
amickan, line 621: These two are unnecessary for what you're testing here, I think

ammar257ammar: I will test it and remove them if not necessary.

app/tests/components_tests/test_views.py
ammar257ammar: What I am testing is that the widget is correctly populated with values on form redisplay upon submission in create/update views. So, I should delete this test?

app/tests/components_tests/test_views.py
ammar257ammar: outdated

app/tests/components_tests/test_views.py
amickan: Discussed in synchronous meeting

app/tests/components_tests/test_forms.py
```diff
@@ -24,3 +44,145 @@ def test_interface_form_field_image_queryset_filter():
...
+    user_upload = UserUploadFactory(creator=user)
+    user_upload.status = UserUpload.StatusChoices.COMPLETED
+    user_upload.save()
+
+    form1 = form_class(
+        user=user,
+        instance=None,
+        base_obj=base_object_factory(),
+        data={
+            "order": 1,
```
amickan, line 81: order is only necessary for display sets. It will just be ignored on archive item forms, but it's cleaner to add another variable to parametrize `extra_form_kwargs` and to pass that to data `**extra_form_kwargs`.

ammar257ammar: Done

amickan (approved):

PR #3684, opened 2024-11-04
Create auto-refreshing incomplete jobs table

koopmant:
Create auto-refreshing incomplete jobs table for pre-requisite jobs during evaluations. 

The table will refresh after 3 minutes if there are incomplete jobs when it loads. It will show the latest list of incomplete jobs when it refreshes or it will vanish if there are no incomplete jobs left. 

part of https://github.com/comic/grand-challenge.org/issues/3639

### DIFF
--- a/app/grandchallenge/evaluation/views/__init__.py
+++ b/app/grandchallenge/evaluation/views/__init__.py
@@ -610,7 +610,7 @@ class EvaluationStatusBadgeDetail(ObjectPermissionRequiredMixin, DetailView):
 class EvaluationIncompleteJobsDetail(
     ObjectPermissionRequiredMixin, DetailView
 ):
-    permission_required = "view_evaluation"
+    permission_required = "change_evaluation"
     template_name_suffix = "_incomplete_jobs_detail"
     model = Evaluation
     raise_exception = True
--- a/app/grandchallenge/evaluation/views/__init__.py
+++ b/app/grandchallenge/evaluation/views/__init__.py
@@ -535,7 +535,36 @@ class EvaluationAdminList(
         return context
 
 
-class EvaluationDetail(ObjectPermissionRequiredMixin, DetailView):
+class EvaluationIncompleteJobsMixin(ObjectPermissionRequiredMixin, DetailView):
+    model = Evaluation
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        incomplete_jobs = filter_by_permission(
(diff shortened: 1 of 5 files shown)


app/grandchallenge/evaluation/views/__init__.py
```diff
@@ -607,6 +607,38 @@ class EvaluationStatusBadgeDetail(ObjectPermissionRequiredMixin, DetailView):
     raise_exception = True
 
 
+class EvaluationIncompleteJobsDetail(
+    ObjectPermissionRequiredMixin, DetailView
+):
+    permission_required = "view_evaluation"
```
jmsmkn, line 613: In the template this data was only available to people with the `change_challenge` perm, so we should keep that. As we discussed, using cross object permissions is a bad idea. Fortunately, we assign the `change_evaluation` permission to the challenge admins group so can use that instead.

app/grandchallenge/evaluation/views/__init__.py
```diff
@@ -607,6 +607,38 @@ class EvaluationStatusBadgeDetail(ObjectPermissionRequiredMixin, DetailView):
...
+            codename="view_job",
+        )
+
+        context.update(
+            {
+                "incomplete_jobs": incomplete_jobs,
+            }
+        )
+
+        return context
```
jmsmkn, line 639: Was this copied from somewhere? If so, would be better to unify these and create an `IncompleteJobsMixin` instead.

app/tests/evaluation_tests/test_views.py
```diff
@@ -221,6 +221,12 @@ def test_group_permission_required_views(self, client):
                 "view_evaluation",
                 e,
             ),
+            (
+                "evaluation-incomplete-jobs-detail",
+                {"pk": e.pk},
+                "view_evaluation",
```
jmsmkn, line 227: Should be `change_evaluation`

app/grandchallenge/evaluation/views/__init__.py
```diff
@@ -535,7 +535,36 @@ def get_context_data(self, *args, **kwargs):
         return context
 
 
-class EvaluationDetail(ObjectPermissionRequiredMixin, DetailView):
+class EvaluationIncompleteJobsMixin(ObjectPermissionRequiredMixin, DetailView):
+    model = Evaluation
```
jmsmkn, line 539: The Mixin can be leaner, otherwise looks good to go!

```suggestion
class EvaluationIncompleteJobsMixin:
```

jmsmkn (approved):

PR #3695, opened 2024-11-13
Add Newick tree support

chrisvanrun:
Part of the pitch:
- https://github.com/DIAGNijmegen/rse-roadmap/issues/354

This PR adds support for Newick files. 

- Validates content by attempting to parse the trees.
- It forces `.newick` suffix in the relative path.
- Uploader (Uppy) allows selection of:
   - `*.nwk`, used by the test files in the parser library.
   - `*.tree`, defined by Wikipedia.
   - `*.newick`, provided in the example files.
 
 At this point, it runs on `acks-late-micro-short`. As the Newick are not very hard to parse or very big.


app/grandchallenge/components/models.py
```diff
@@ -565,9 +570,32 @@ def file_mimetypes(self):
...
+    def file_extension(self):
+        if self.is_json_kind:
+            return ".json"
+
+        if self.is_file_kind:
+            if self.kind == InterfaceKind.InterfaceKindChoices.NEWICK:
+                return ".newick"
+            return f".{self.kind.lower()}"
+
+        raise RuntimeError(f"Unknown kind {self.kind}")
```
jmsmkn, line 597: These functions are not clear to me. What do these properties do? Why do the file extensions differ between the two methods? Maybe the naming can be changed or a docstring added to help out other developers.

jmsmkn, line 597: These functions are not clear to me. What do these properties do? Why do the file extensions differ between the two methods? Maybe the naming can be changed or a docstring added to help out other developers.

app/grandchallenge/components/models.py
```diff
@@ -565,9 +570,32 @@ def file_mimetypes(self):
...
+            return ("text/x-nh", "application/octet-stream")
         else:
             raise RuntimeError(f"Unknown kind {self.kind}")
 
+    @property
+    def file_extensions(self):
+        exts = [self.file_extension]
+
+        if self.kind == InterfaceKind.InterfaceKindChoices.NEWICK:
+            exts.extend([".nwk", ".tree"])
```
jmsmkn, line 583: Does that mean we store data with multiple different extensions? When they're used in a component job is only one file extension used?

app/grandchallenge/components/models.py
```diff
@@ -534,7 +539,7 @@ def default_field(self):
             return forms.JSONField
 
     @property
-    def file_mimetypes(self):
+    def file_mimetypes(self):  # noqa:C901
```
jmsmkn, line 542: Can be refactored to use a map, e.g. `INTERFACE_KIND_TO_MIMETYPE`, rather than lots of if statements.

app/grandchallenge/components/validators.py
```diff
@@ -20,3 +21,17 @@ def validate_safe_path(value):
...
+def validate_newick_tree_format(tree):
+    """Validates a Newick tree by passing it through a validator"""
+    parser = Phylo.NewickIO.Parser.from_string(tree)
+
+    try:
+        has_tree = False
+        for _ in parser.parse():
+            has_tree = True
+        if not has_tree:
+            raise ValueError("No tree found")
```
jmsmkn, line 35: Shouldn't this be a `ValidationError`?

app/grandchallenge/components/models.py
```diff
@@ -565,9 +570,32 @@ def file_mimetypes(self):
...
+
+    @property
+    def file_extension(self):
+        if self.is_json_kind:
+            return ".json"
+
+        if self.is_file_kind:
+            if self.kind == InterfaceKind.InterfaceKindChoices.NEWICK:
+                return ".newick"
+            return f".{self.kind.lower()}"
```
jmsmkn, line 595: Using blocks rather than fall-throughs is preferred, please add this to the style guide:

```suggestion
        elif self.is_file_kind:
            if self.kind == InterfaceKind.InterfaceKindChoices.NEWICK:
                return ".newick"
            else:
                return f".{self.kind.lower()}"
```

jmsmkn (commented):
`acks-late-micro-short` really doesn't have much memory available, so don't be surprised when we get memory errors from that validation.

chrisvanrun:
> `acks-late-micro-short` really doesn't have much memory available, so don't be surprised when we get memory errors from that validation.

Would you agree that for a first implementation, it is OK? They should be around the same size as the multiple annotation JSON types.  I'll be looking at re-routing file-to-object task to larger workers when implementing BIOM.

app/grandchallenge/uploads/widgets.py
```diff
@@ -5,17 +5,24 @@ class UserUploadWidgetMixin:
...
+        self.allowed_file_extensions = allowed_file_extensions or []
 
     def get_context(self, *args, **kwargs):
         context = super().get_context(*args, **kwargs)
         widget_id = f'X_{context["widget"]["attrs"]["id"]}'
         context["widget"]["attrs"]["id"] = widget_id
         context["widget"]["allowed_file_types"] = {
             "id": f"{widget_id}AllowedFileTypes",
-            "value": self.allowed_file_types,
+            "value": [*self.allowed_file_types, *self.allowed_file_extensions],
```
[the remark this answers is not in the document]
chrisvanrun: It's a property of Uppy that file extensions and mime types are combined. As such, I think the correct place to mix them is in the widget: just before they get passed to Uppy.

app/grandchallenge/components/validators.py
chrisvanrun: It'll eventually be a validation error because of the try-catch. I can make it a `ValidationError` from a `ValidationError`, but that seemed odd.

app/grandchallenge/components/models.py
chrisvanrun: One is the primary, used to validate the relative path of interfaces. The multiple is the one for uploading. I think it's best to rename the `file_extensions` to `upload_file_extensions` and add some docstrings to explain their function.

app/grandchallenge/components/models.py
chrisvanrun: One is the primary, used to validate the relative path of interfaces. The multiple is the one for uploading. I think it's best to rename the file_extensions to upload_file_extensions and add some docstrings to explain their function.

jmsmkn:
Yes totally fine, was just a heads up.

app/grandchallenge/components/models.py
chrisvanrun: Indeed, only a single extension in the relative paths of the NEWICK kind. The additional upload extensions should only result in additional extensions in the original filename in the serialization of the CIV. AFAIK that is only applicable when the original uploaders are curious about what input was used to generate component job outputs: and it sort of make sense to keep the original extension there.

app/grandchallenge/components/models.py
chrisvanrun: I have replaced this with hardcorded lookups. I did add it to the style guide.

chrisvanrun:
Summary of updates:

The mime-types were not being really used as MIME-types but rather as input to Uppy. At the same time the need to support `.newick` as file extension and uploads caused several other properties. This added complexity.

Since we don't need to actually map to mime types I've refactored things to a `allowed_file_types` map and lookup, combining both MIME and the other file type definitions Uppy supports: https://uppy.io/docs/uppy/#restrictions.

Newick validator now also handles time-out errors, memory errors. It also no longer exposes generic exceptions.

jmsmkn (approved):

chrisvanrun:
Thanks for the review James!

PR #3696, opened 2024-11-13
Update challenge cost calculation

amickan:
This updates the cost calculation in line with our new policy and removes the cost calculation from the request form. 
The cost estimate in the detail view has been updated to include the warnings and extra calculations that were previously only displayed in the request form. 

Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/356

The updated cost estimate on the detail view (only visible to reviewers) now looks as follows:

image


amickan:
CI/CD is failing with a timeout error on trying to pull the postgres image. I will wait and rerun them later.

app/grandchallenge/challenges/models.py
```diff
@@ -1229,36 +1229,60 @@ def base_cost_euros(self):
                 exempt_from_base_costs=True, members_group__user=self.creator
             ).exists()
         ):
-            return 0
+            return settings.CHALLENGE_MINIMAL_COMPUTE_AND_STORAGE_IN_EURO
```
jmsmkn, line 1232: These changes mean the property name is no longer correct.

jmsmkn (commented):
Rest looks good. Whilst we're here - invoices use compute cost and storage cost. When creating them we manually have to add up the values across the phases. Could those two totals be displayed in the table so that we can put those values directly into the invoice form?

app/grandchallenge/challenges/models.py
jmsmkn: Or maybe it is better to create a new property as now this property does two things. The "base_costs" calculation is still valid in the new case.

amickan: What would you call the second property? `minimal_challenge_cost`?

amickan:
> Rest looks good. Whilst we're here - invoices use compute cost and storage cost. When creating them we manually have to add up the values across the phases. Could those two totals be displayed in the table so that we can put those values directly into the invoice form?

That reminds me, in the admin, what do we want to note down as base cost from now on? The 5000 or 0 as usual?

amickan:
And how do we want to split compute and storage across the minimal 1000 Euros? Say if a challenge comes down to 100 storage and 500 compute, how do we want to split the remaining 400 of the minimal costs across the two? Evenly or proportionately or all on compute?

app/grandchallenge/challenges/models.py
jmsmkn: Yes sounds fine.

app/grandchallenge/challenges/models.py
```diff
@@ -1234,31 +1234,110 @@ def base_cost_euros(self):
...
+            self.total_data_and_docker_storage_bytes
+        )
+
+    @property
+    def total_compute_costs_euros(self):
+        return self.get_compute_costs_euros(self.total_compute_time)
+
+    def calculate_invoiced_amount(self, cost, ratio):
+        if self.storage_and_compute_cost_surplus <= 0:
+            # Below minimum prize, add proportional shortfall
```
jmsmkn, line 1279:
```suggestion
            # Below minimum price, add proportional shortfall
```

app/grandchallenge/challenges/models.py
```diff
@@ -1234,31 +1234,110 @@ def base_cost_euros(self):
...
+    def total_storage_costs_euros(self):
+        return self.get_data_storage_costs_euros(
+            self.total_data_and_docker_storage_bytes
+        )
+
+    @property
+    def total_compute_costs_euros(self):
+        return self.get_compute_costs_euros(self.total_compute_time)
+
+    def calculate_invoiced_amount(self, cost, ratio):
```
jmsmkn, line 1277:
```suggestion
    def calculate_invoiced_amount(self, *, cost, ratio):
```

app/grandchallenge/challenges/models.py
```diff
@@ -1234,31 +1234,110 @@ def base_cost_euros(self):
...
+        if self.storage_and_compute_cost_surplus <= 0:
+            # Below minimum prize, add proportional shortfall
+            return round(
+                cost + ratio * abs(self.storage_and_compute_cost_surplus)
+            )
+        else:
+            # Above minimum prize, allocate surplus proportionally
```
jmsmkn, line 1284:
```suggestion
            # Above minimum price, allocate surplus proportionally
```

jmsmkn (approved):

amickan:
Looking at this again with fresh eyes, I noticed rounding inconsistencies due to that fact that I recalculated `total_storage_costs_euros` and `total_compute_costs_euros` from scratch (using the helper methods and the actual compute time and stored bytes) rather than summing the already calculated (and rounded to the nearest 10 Euros) totals for the phases. Fixed that now.

jmsmkn (approved):

PR #3697, opened 2024-11-13
Dynamic examples for view content field help text

ammar257ammar:
Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/362

### DIFF
--- a/app/grandchallenge/hanging_protocols/forms.py
+++ b/app/grandchallenge/hanging_protocols/forms.py
@@ -135,11 +135,11 @@ class ViewContentExampleMixin:
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         if self.instance:
-            interface_slugs = self.instance.interfaces.values_list(
-                "slug", flat=True
-            )
+            interface_slugs = [
+                interface.slug for interface in self.instance.interfaces
+            ]
 
-            if interface_slugs.count() > 0:
+            if len(interface_slugs) > 0:
                 self.fields[
                     "view_content"
                 ].help_text += f"The following interfaces are used in your {self.instance._meta.verbose_name}: {oxford_comma(interface_slugs)}. "


app/tests/hanging_protocols_tests/test_forms.py
```diff
@@ -294,3 +302,100 @@ def test_hanging_protocol_clientside():
...
+@pytest.mark.parametrize(
+    "overlay_type",
+    (
+        InterfaceKindChoices.SEGMENTATION,
+        InterfaceKindChoices.HEAT_MAP,
+        InterfaceKindChoices.DISPLACEMENT_FIELD,
+    ),
+)
+@pytest.mark.django_db
+def test_forms_view_content_help_text(
```
ammar257ammar, line 341: I put this form test (still not final) under hanging_protocols since it is related to the view_content field of the HangingProtocolMixin, and the logic of assessing the forms (Algorithm, Archive, ReaderStudy, Phase) is similar but not identical. Therefore, the test will be long. Now, I put the test for Algorithm and Archive forms and will discuss first if it is better to put the test all here, or split them over multiple tests or put each test in its relevant app.

app/grandchallenge/archives/forms.py
```diff
@@ -68,16 +70,33 @@ def __init__(self, *args, **kwargs):
...
-                self.instance.items.exclude(values__isnull=True)
-                .values_list("values__interface__slug", flat=True)
+            interfaces = (
+                ComponentInterface.objects.filter(
+                    componentinterfacevalue__in=self.instance.items.exclude(
+                        values__isnull=True
+                    ).values_list("values", flat=True)
+                )
                 .order_by()
                 .distinct()
```
amickan, line 80: Why not just do `self.instance.items.exclude(values__isnull=True).distinct()` to get the interfaces like you do for algorithms, for example? 

And is the `order_by` necessary? And if so, would it make sense to do that also in the other places? 

In general it looks like you're doing the same thing in multiple places. If so, I think this should be a mixin - reduces the logic to one place, makes it easier to maintain.

app/grandchallenge/components/utils.py
```diff
@@ -0,0 +1,49 @@
...
+            break
+
+        view_content_example[port] = [images.pop(0)]
+        for _ in range(overlays_per_image):
+            view_content_example[port].append(overlays.pop(0))
+        if remaining_overlays > 0:
+            view_content_example[port].append(overlays.pop(0))
+            remaining_overlays -= 1
+
+    JSONValidator(schema=VIEW_CONTENT_SCHEMA)(value=view_content_example)
```
amickan, line 47: I don't think checking the example for adherence to the schema here makes sense. It should always be valid. If it's not valid, this will raise an error and will make the entire form unusable for the user. I don't think that's what we want. 

Ultimately, what we need is to make sure that the generated example adheres to the schema and that it continues to do so even when the schema changes. To achieve that, I think a test that calls this function and validates it against the schema is the way to go. Future developers will then be notified if their changes to the view content schema break this function and can update the code accordingly.

app/grandchallenge/components/utils.py
```diff
@@ -0,0 +1,49 @@
...
+
+def generate_view_content_example(interfaces):
+    images = list(
+        interfaces.filter(kind=InterfaceKindChoices.IMAGE)
+        .order_by("slug")
+        .values_list("slug", flat=True)
+    )
+
+    if not images:
+        return None
```
amickan, line 19: Would it make sense to inform the user that there are no image interfaces and that a viewport needs at least 1 image assigned? You don't need to do that here, and instead check that in the form and add to the helptext accordingly.

app/grandchallenge/components/utils.py
```diff
@@ -0,0 +1,49 @@
...
+        .order_by("slug")
+        .values_list("slug", flat=True)
+    )
+    view_content_example = {}
+    overlays_per_image = len(overlays) // len(images)
+    remaining_overlays = len(overlays) % len(images)
+
+    for port in ViewportNames.values:
+        if len(images) == 0:
+            break
```
amickan, line 38: You are already checking if images is empty above, so no need to repeat that here, right?

ammar257ammar: This is for the for loop, I am looping over the viewport names which can be more than the number of available images and poping the image from the list. And we need to finish when image_ci list is empty, that is what this check does.

amickan: Ah of course, thanks for clarifying!

app/grandchallenge/archives/forms.py
amickan: And the following is probably more efficient:

```suggestion
                interface_slugs = self.instance.items.exclude(values__isnul=True).values_list("values__slug", flat=True)
                ComponentInterface.objects.filter(
                    slugs__in=interface_slugs
                )
```

ammar257ammar: the `order_by` is not necessary.
Agree with the more efficient query.
Done.

app/grandchallenge/components/utils.py
```diff
@@ -0,0 +1,49 @@
...
+
+from grandchallenge.components.models import InterfaceKindChoices
+from grandchallenge.core.validators import JSONValidator
+from grandchallenge.hanging_protocols.models import (
+    VIEW_CONTENT_SCHEMA,
+    ViewportNames,
+)
+
+
+def generate_view_content_example(interfaces):
```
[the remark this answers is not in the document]
ammar257ammar: A FormMixin was created where the logic is reduced to one place, and thus, the utils is not needed anymore.
Done.

app/grandchallenge/components/utils.py
jmsmkn: A test is necessary but I don't think that it is possible to cover all situations a-priori. In that case, you still need a runtime check. Having an un-rendered view is okay, but it is useful to catch the `ValidationError` here and raise it as a `RuntimeError` so that future developers know from Sentry that this is not actually a data validation problem, but that the logic in this method that would need to be fixed.

ammar257ammar: Actually, after confirming that all interfaces need to be present in the example, I think the generated example should go through the same validation path as the `view_content` field in the [HangingProtocolMixin](https://github.com/comic/grand-challenge.org/blob/9c7930c1a7d2c28af4753dbaac249931eae855a3/app/grandchallenge/hanging_protocols/models.py#L305). That is, checking that `isolated_interfaces` are solely present in a viewport, only one `image` in a view port, and no `undisplayable_interfaces` present in any view port.

However, since the `HangingProtocolMixin` is an Abstract class, it **can not** be instantiated. So, I **can not** do:
```
hp = HangingProtocolMixin(view_content=view_content_example)
hp.full_clean()

```
I am exploring alternatives.

app/tests/hanging_protocols_tests/test_forms.py
ammar257ammar: outdated

app/grandchallenge/components/utils.py
ammar257ammar: Done

app/tests/hanging_protocols_tests/test_forms.py
```diff
@@ -294,3 +312,223 @@ def test_hanging_protocol_clientside():
...
+)
+@pytest.mark.parametrize(
+    "isolated_interface_type",
+    (
+        InterfaceKind.InterfaceKindChoices.CHART,
+        InterfaceKind.InterfaceKindChoices.PDF,
+    ),
+)
+@pytest.mark.django_db
+def test_forms_view_content_help_text(
```
ammar257ammar, line 424: This is a comprehensive test due to the many cases to cover (images, overlays, isolated_interfaces, undisplayable_interfaces) but maybe it is better to split. 
To discuss

app/grandchallenge/core/validators.py
```diff
@@ -188,3 +190,72 @@ def __eq__(self, other):
 
     def __ne__(self, other):
         return not (self == other)
+
+
+@deconstructible
+class ViewContentValidator:
+    """Validates view_content JSON against governing rules."""
+
+    def __call__(self, value):
```
amickan, line 199: I think a custom validator is not the way to go here. It spreads the validation logic related to the view_content over multiple places (validator and the model's clean method), making it harder to maintain, I think. You now have a dangling `check_consistent_viewports` check that this validator does not (and cannot) do, but which you want to check as well when validating your example. 

I think it would be better to leave all validation (except the JSON schema, of course) on the model level. You could add a `clean_view_content` method to the `HangingProtocolMixin` that calls all the necessary sub-clean-functions, and then call that method in the `ViewContentExampleMixin` with `self.instance.clean_view_content()`

amickan: Making sure to have set the example on the instance before, of course `self.instance.view_content = example_view_content`

app/tests/hanging_protocols_tests/test_forms.py
amickan: Very complex indeed. You can also parametrize the object (Algorithm, Archive etc), which would reduce the actual assert part of the test to just a few lines. 

And you also want to have a test for an example that fails validation to make sure that your validation step is working.

ammar257ammar: I parametrized the object but still there is a difference in the implementation between (Algorithm, Phase) and (Archive, ReaderStudy) so I had to use an `if` statement. It seemed a reasonable tradeoff compared to having two tests with lengthy case coverage.

ammar257ammar: One more point, by trying to cover as much as possible of interface types, the test now unfold to 128 tests taking ~15 seconds to finish. We could for example do the tests with only one example of mandatory_isolated_interfaces and one of the overlays instead of currently testing with two of each. That will bring the number of tests down to 32 tests taking ~5 sec.

app/grandchallenge/core/validators.py
ammar257ammar: I did it as suggested, like
```
self.instance.view_content = example_view_content
self.instance.clean_view_content()
```

However, we are setting the example as the value of the actual field `view_content`. In update forms that is already overwriting the original `view_content` value (as I can see in the debugger). Fortunately, the JSON widget is not using that instance value, and therefore, the original value is correctly kept in the field on the rendered form.
However, it does not feel ideal/neat to validate this way. What do you think?

amickan: Ah yes, that is a good point. I'm actually not sure what the best way is. Maybe @jmsmkn can comment?

jmsmkn: Why not provide the view content as an argument to the validation method rather than acting on an attribute?

ammar257ammar: Thanks! I provided the view content as an argument to the validation method.
Done.

app/grandchallenge/archives/models.py
```diff
@@ -267,6 +268,14 @@ def interface_viewname(self):
...
         return self.items
 
+    @property
+    def interfaces(self):
+        return ComponentInterface.objects.filter(
+            slug__in=self.items.exclude(values__isnull=True).values_list(
+                "values__interface__slug", flat=True
+            )
+        ).distinct()
+
```
amickan, line 278: This query could also be: `ComponentInterface.objects.filter(componentinterfacevalue__archive_items__in=self.items).exclude(componentinterfacevalue__isnull=True)`

I'm not sure if that is more efficient, or if the difference is significant. 

In your proposed query, I would switch to filtering by `pk` because that is more efficient than filtering by `slug`. 

And you can move this property to the `ValuesForInterfacesMixin`, saves a few lines of code ;-)

app/grandchallenge/evaluation/models.py
```diff
@@ -814,6 +814,10 @@ def set_default_interfaces(self):
             [ComponentInterface.objects.get(slug="metrics-json-file")]
         )
 
+    @property
+    def interfaces(self):
+        return (self.inputs.all() | self.outputs.all()).distinct()
```
amickan, line 819: This should be 
```suggestion
    def interfaces(self):
        return (self.algorithm_inputs.all() | self.algorithm_outputs.all()).distinct()
```

app/grandchallenge/hanging_protocols/forms.py
```diff
@@ -116,3 +129,86 @@ def _validate_slice_plane_indicator(self, *, viewport, viewport_names):
...
+            view_content_example = self.generate_view_content_example()
+
+            if view_content_example:
+                self.fields[
+                    "view_content"
+                ].help_text += f"Example usage: {view_content_example}. "
+            else:
+                self.fields[
+                    "view_content"
+                ].help_text += "No interfaces of type (image, chart, pdf, mp4, thumbnail_jpg, thumbnail_png) are used. At least one interface of those types is needed to config the viewer. "
```
amickan, line 156:
```suggestion
                ].help_text += "No interfaces of type image, chart, pdf, mp4, thumbnail_jpg or thumbnail_png are used. At least one interface of those types is needed to configure the viewer. "

```

app/grandchallenge/hanging_protocols/forms.py
```diff
@@ -116,3 +129,86 @@ def _validate_slice_plane_indicator(self, *, viewport, viewport_names):
...
+        images = list(
+            self.instance.interfaces.filter(
+                kind=InterfaceKindChoices.IMAGE
+            ).values_list("slug", flat=True)
+        )
+        mandatory_isolation_interfaces = list(
+            self.instance.interfaces.filter(
+                kind__in=InterfaceKind.interface_type_mandatory_isolation()
+            ).values_list("slug", flat=True)
+        )
```
amickan, line 173: I think a simple list comprehension would be clearer here (it probably won't make a real difference if `interfaces` is a cached property, but it's easier to read):  

```suggestion
        images = [
           int.slug for int in self.instance.interfaces.filter(kind=InterfaceKindChoices.IMAGE)
        ]

        mandatory_isolation_interfaces = [
            int.slug for in in 
            self.instance.interfaces.filter(
                kind__in=InterfaceKind.interface_type_mandatory_isolation()
            )
        ]

```

app/grandchallenge/hanging_protocols/forms.py
```diff
@@ -116,3 +129,86 @@ def _validate_slice_plane_indicator(self, *, viewport, viewport_names):
...
+
+        try:
+            JSONValidator(schema=VIEW_CONTENT_SCHEMA)(
+                value=view_content_example
+            )
+            self.instance.clean_view_content(view_content_example)
+        except ValidationError as e:
+            raise RuntimeError(
+                f"view_content example is not valid. Validation error: {e.messages}"
+            )
```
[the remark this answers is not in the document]
jmsmkn: The way to do this so that context is not lost is

```python
try:
    ...
except InterestingError as error:
    raise NewError("New message") from error
```

See https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/

app/grandchallenge/hanging_protocols/forms.py
```diff
@@ -116,3 +129,86 @@ def _validate_slice_plane_indicator(self, *, viewport, viewport_names):
...
+        if not images and not mandatory_isolation_interfaces:
+            return None
+
+        overlays = list(
+            self.instance.interfaces.exclude(
+                kind__in=InterfaceKind.interface_type_undisplayable()
+                | InterfaceKind.interface_type_mandatory_isolation()
+                | {InterfaceKindChoices.IMAGE}
+            ).values_list("slug", flat=True)
+        )
```
amickan, line 184: See my comment above. 

And also: does this really only return overlays? What about strings, integers etc., should those be considered overlays and are they displayable?  I guess it's fine, or else they would be listed in `interface_type_undisplayable`, but just wondering.

app/grandchallenge/reader_studies/models.py
```diff
@@ -748,6 +748,14 @@ def list_url(self):
     def civ_sets_related_manager(self):
         return self.display_sets
 
+    @property
+    def interfaces(self):
+        return ComponentInterface.objects.filter(
+            slug__in=self.display_sets.exclude(
+                values__isnull=True
+            ).values_list("values__interface__slug", flat=True)
+        ).distinct()
```
amickan, line 757: Same comment as for the archive property

app/grandchallenge/hanging_protocols/forms.py
jmsmkn: You're right here Anne. All of these should anyway be list comprehensions to save going to the database multiple times, this data is small enough to filter client side so is more efficient that way.

app/tests/hanging_protocols_tests/test_forms.py
```diff
@@ -294,3 +315,351 @@ def test_hanging_protocol_clientside():
...
+            interface=ci,
+        )
+        ci_list.append(ci)
+        civ_list.append(civ)
+
+    base_obj = base_object_factory()
+
+    if object_factory:
+        object = object_factory(
+            **{base_object_lookup: base_obj},
```
amickan, line 513: The base_object doesn't matter for this test, right? You can just create an ArchiveItem or DisplaySet and then reference the automatically created base object in the form with `object.base_object`. You also don't need to explicitly create a challenge for the Phase, that is done by the factory for you.

app/tests/hanging_protocols_tests/test_forms.py
amickan: I think here the only thing you want to test is that the helptext is added to the field. The actualy content (i.e. the example generation) is tested in the next test, right? So then you can just test the case where there are no interfaces, or none of the necessary kind and one case where the view content is generated. Or did I misunderstand?

app/grandchallenge/hanging_protocols/forms.py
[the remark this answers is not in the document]
amickan: Thanks, James!

app/grandchallenge/hanging_protocols/forms.py
amickan: If they become `cached_properties` like I suggested somewhere else, I think hitting the database is not a concern, since the filtering would be happening on the cached queryset, but regardless a list comprehension is cleaner.

app/grandchallenge/hanging_protocols/models.py
```diff
@@ -326,10 +328,9 @@ class HangingProtocolMixin(models.Model):
     def clean(self):
         super().clean()
 
-        self.check_consistent_viewports()
-        self.check_all_interfaces_in_view_content_exist()
+        self.clean_view_content(self.view_content)
 
-    def check_consistent_viewports(self):
+    def check_consistent_viewports(self, view_content):
```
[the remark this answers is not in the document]
amickan: You mean making all the clean functions separate functions?

app/grandchallenge/hanging_protocols/forms.py
jmsmkn: The queryset would be cached but not necessarily evaluated. Each time `filter` is called on it and the queryset evaluated (in this case by putting it in a list comprehension) you go back to the database. The suggestion here would result in two database hits, by doing a pure client side filtering you do it in one.

app/grandchallenge/hanging_protocols/models.py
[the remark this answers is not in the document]
jmsmkn: Only the ones that depend on `view_content`.

app/grandchallenge/evaluation/models.py
ammar257ammar: Done

app/grandchallenge/hanging_protocols/forms.py

app/grandchallenge/archives/models.py
ammar257ammar: In our discussion about this last week, we agreed not to go through `componentinterfacevalue__` since that table is huge and thus the query not efficient. I prefer to stick with this one, and agree with using `pk` instead of the `slug`.
Also, moved the property to `ValuesForInterfacesMixin`

Done.

app/grandchallenge/algorithms/models.py
```diff
@@ -443,6 +443,10 @@ def add_user(self, user):
     def remove_user(self, user):
         return user.groups.remove(self.users_group)
 
+    @property
+    def interfaces(self):
+        return (self.inputs.all() | self.outputs.all()).distinct()
```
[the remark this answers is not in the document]
ammar257ammar: Done

app/grandchallenge/hanging_protocols/forms.py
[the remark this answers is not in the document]
ammar257ammar: Thanks for the feedback!
Points taken.
Done.

app/grandchallenge/hanging_protocols/forms.py
ammar257ammar: According the discussion on the pitch, strings, bools and integers can be used in combination with Images and shown as overlays in the viewport.

from Miriam

> because for example you can add CI's with text, which will be shown as labels on the viewport

from James

> Also can't some non-image interfaces also be assigned to view items: pdf files and charts in isolation, floats and bools in combination with images, etc?

list comprehension point is taken.
Done.

app/grandchallenge/reader_studies/models.py
ammar257ammar: Done

app/grandchallenge/hanging_protocols/forms.py
ammar257ammar: Changed to list comprehension with client side filtering.
Done.

app/tests/hanging_protocols_tests/test_forms.py
ammar257ammar: For the first point about `base_object` that is true, but then we need to move the form and base_obj to inside the if statement (and duplicate the form part) to deal with each case separately. I was trying to minimize the difference as much as possible. So, currently it is:

```python
    base_obj = base_object_factory()

    if object_factory:
        object = object_factory(
            **{base_object_lookup: base_obj},
        )
        object.values.set(civ_list)
    else:
        base_obj.inputs.set(ci_list)
        base_obj.outputs.set([])

    form = form_class(
        user=UserFactory(), instance=base_obj, **form_extra_kwargs
    )
```

After the suggestion, I have this in mind now:

```python
    if object_factory:
        object = object_factory()
        object.values.set(civ_list)

        form = form_class(
            user=UserFactory(), instance=object.base_object,
        )
    else:
        base_obj = base_object_factory()

        base_obj.inputs.set(ci_list)
        base_obj.outputs.set([])

        form = form_class(
            user=UserFactory(), instance=base_obj, **form_extra_kwargs
        )
```

 Am I missing something? and which one you prefer?

ammar257ammar: Also realized, since we need to use `algorithm_inputs` to set the ci_list to the `Phase` (based on another comment you made), then the property name `inputs`/`algorithm_inputs` should also be passed as a parameter to the test.

Would it be better to move the Phase case to its own test?

app/tests/hanging_protocols_tests/test_forms.py
ammar257ammar: That is correct, the next test does the example check. For this one, we need to test the following cases:
- no interfaces at all
- no interfaces of type Image or isolated, but others like overlays or undisplayable (so we cover the case where we have both messages "you have the following interfaces .." and "No interfaces of type".
- one case where the view content is generated

Done

app/tests/hanging_protocols_tests/test_forms.py
amickan: Ok, given the complexity, I think it would be better to create separate tests after all. 
- One test for Archives and ReaderStudies (you can parametrize that test without the need for if statements)
- One test for Algorithms
- One test for Phases

It will introduce 3 very similar tests, but they will each be easier to read.

ammar257ammar: Thanks! I agree.
Done

app/grandchallenge/hanging_protocols/forms.py
```diff
@@ -116,3 +129,91 @@ def _validate_slice_plane_indicator(self, *, viewport, viewport_names):
...
+
+        self.fields["view_content"].help_text += format_lazy(
+            'Refer to the <a href="{}">documentation</a> for more information',
+            reverse("documentation:detail", args=["viewer-content"]),
+        )
+
+    def generate_view_content_example(self):
+        images = [
+            interface.slug
+            for interface in self.instance.interfaces.all()
```
amickan, line 166: The `.all()` should not be necessary here, right?

app/grandchallenge/hanging_protocols/models.py
```diff
@@ -326,24 +328,25 @@ class HangingProtocolMixin(models.Model):
...
-        self.check_consistent_viewports()
-        self.check_all_interfaces_in_view_content_exist()
+        self.clean_view_content(self.view_content, self.hanging_protocol)
 
-    def check_consistent_viewports(self):
-        if self.view_content and self.hanging_protocol:
-            if set(self.view_content.keys()) != {
-                x["viewport_name"] for x in self.hanging_protocol.json
+    @staticmethod
+    def check_consistent_viewports(view_content, hanging_protocol):
```
amickan, line 334:
```suggestion
    def check_consistent_viewports(*, view_content, hanging_protocol):
```

app/grandchallenge/hanging_protocols/models.py
```diff
@@ -326,24 +328,25 @@ class HangingProtocolMixin(models.Model):
...
             }:
                 raise ValidationError(
                     "Image ports in view_content do not match "
                     "those in the selected hanging protocol."
                 )
 
-    def check_all_interfaces_in_view_content_exist(self):
-        if not hasattr(self.view_content, "items"):
+    @staticmethod
+    def check_all_interfaces_in_view_content_exist(view_content):
```
amickan, line 345:
```suggestion
    def check_all_interfaces_in_view_content_exist(*, view_content):
```

app/grandchallenge/hanging_protocols/models.py
```diff
@@ -395,5 +398,9 @@ def check_all_interfaces_in_view_content_exist(self):
                     f"{', '.join(i.slug for i in undisplayable_interfaces)}"
                 )
 
+    def clean_view_content(self, view_content, hanging_protocol):
```
amickan, line 401:
```suggestion
    @staticmethod
    def clean_view_content(*, view_content, hanging_protocol):
```

app/tests/hanging_protocols_tests/test_forms.py
```diff
@@ -294,3 +312,415 @@ def test_hanging_protocol_clientside():
...
+        )
+        civ_list.append(civ)
+
+    for i in range(number_of_undisplayable_interfaces):
+        civ = ComponentInterfaceValueFactory(
+            interface=ComponentInterfaceFactory(
+                kind=InterfaceKind.InterfaceKindChoices.ZIP,
+                title=f"test-ci-undisplayable-{i}",
+            ),
+        )
```
amickan, line 406: This could be a helper function that you re-use in all 3 tests.

app/grandchallenge/hanging_protocols/models.py
ammar257ammar: I kept the wrapping function `clean_view_content` as a method, so it can be called with `self.instance` in the form and still the validation functions inside it can still be called with `self.`
Otherwise it would be:

```python
@staticmethod
def clean_view_content(*, view_content, hanging_protocol):
       HangingProtocolMixin.check_consistent_viewports(view_content, hanging_protocol)
       HangingProtocolMixin.check_all_interfaces_in_view_content_exist(view_content)

```
And the call in the form will be:
`HangingProtocolMixin.clean_view_content(...
`

Is that what we want here?

app/grandchallenge/hanging_protocols/forms.py

app/grandchallenge/hanging_protocols/models.py
ammar257ammar: yeah true.
I will stick to it.
Done.

app/tests/hanging_protocols_tests/test_forms.py
ammar257ammar: ah, great idea!
Actually, I made two fixtures, one to create a list of component interfaces `ci_list` (needed for Algorithm and Phase) and the second to create a list of component interface values `civ_list` (needed for Archive and ReaderStudy). That made the tests much cleaner and concise :)
The last function `test_generate_view_content_example` does not create a list but create the actual objects and the fixture is not needed there.

Done.

app/tests/hanging_protocols_tests/test_forms.py
```diff
@@ -294,3 +312,391 @@ def test_hanging_protocol_clientside():
...
+    for i in range(number_of_images):
+        ComponentInterfaceFactory(
+            kind=InterfaceKindChoices.IMAGE, title=f"test-ci-image-{i}"
+        )
+
+    for i in range(number_of_overlays):
+        ComponentInterfaceFactory(
+            kind=InterfaceKindChoices.SEGMENTATION,
+            title=f"test-ci-overlay-{i}",
+        )
```
amickan, line 687: Can't you use the `ci_list` fixture here as well?

app/tests/hanging_protocols_tests/test_forms.py
```diff
@@ -294,3 +312,391 @@ def test_hanging_protocol_clientside():
...
+    civ_list = []
+
+    for i in range(number_of_isolated_interfaces):
+        civ = ComponentInterfaceValueFactory(
+            interface=ComponentInterfaceFactory(
+                kind=InterfaceKindChoices.CHART,
+                title=f"test-ci-isolated-{i}",
+            ),
+        )
+        civ_list.append(civ)
```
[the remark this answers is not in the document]
amickan: Discussed this synchronously: we decided to go with a helper function instead of fixtures. This suits the current test cases the best.

app/tests/hanging_protocols_tests/test_forms.py
ammar257ammar: Upon synchronous discussion, a helper function is made and used here as well.
Done.

app/tests/hanging_protocols_tests/test_forms.py
```diff
@@ -294,3 +311,358 @@ def test_hanging_protocol_clientside():
         }
     )
     assert form.is_valid()
+
+
+def make_ci_list(
+    *,
```
ammar257ammar, line 317: I used keyworded arguments here to make sure the number of ci by type is correctly passed, to avoid mistakes since the four arguments are integers.

amickan (approved):

PR #3699, opened 2024-11-14
Adjustable gpu types

koopmant:
Make the selectable gpu type and maximum memory adjustable for evaluations. 


app/grandchallenge/evaluation/forms.py
```diff
@@ -157,12 +161,19 @@ def __init__(self, *args, **kwargs):
...
             MinValueValidator(settings.ALGORITHMS_MIN_MEMORY_GB),
-            MaxValueValidator(settings.ALGORITHMS_MAX_MEMORY_GB),
+            MaxValueValidator(self.instance.maximum_settable_memory_gb),
         ]
+        self.fields["evaluation_requires_gpu_type"].widget = forms.Select(
+            choices=[
+                (c.value, c.label)
+                for c in self.instance.get_selectable_gpu_types()
+            ]
+        )
```
koopmant, line 171: I'm not sure if this is the way to do it. I wanted to dynamically adjust the choices on the field itself; but  that does not appear to be possible?

amickan: The following does not work? 

```python
self.fields["evaluation_requires_gpu_type"].choices = [(c.value, c.label) for c in self.instance.get_selectable_gpu_types()]
```

app/grandchallenge/evaluation/models.py
```diff
@@ -743,6 +784,37 @@ def _clean_external_evaluation(self):
...
+        if (
+            self.evaluation_requires_memory_gb
+            > self.maximum_settable_memory_gb
+        ):
+            raise ValidationError(
+                f"Ensure the value for Evaluation requires memory gb (currently "
+                f"{self.evaluation_requires_memory_gb}) is less than or equal "
+                f"to the maximum settable (currently "
+                f"{self.maximum_settable_memory_gb})."
+            )
```
amickan, line 816: Aren't you mixing evaluation settings with algorithm settings here? This would keep a challenge admin from using more memory or a more powerful GPU for their evaluations than their participants algorithms, right?

app/grandchallenge/evaluation/forms.py
koopmant: This is embarrassing, yes that works. (I really thought I had tried that...) Thanks!

app/grandchallenge/evaluation/models.py
koopmant: Yes that's correct. The 'maximum' and 'selectable' settings will be used for both algorithms and evaluations. I believe that was the intension, but I can easily add another setting for algorithms in the next PR.

amickan:
> The original plan was to change the method form. But because we have recently adjusted the settings for algorithms to be set on the algorithm rather than the image, it doesn't feel right to do it differently for methods. Instead, I've adjusted the phase update form to include the settings for `evaluation_requires_gpu_type` and `evaluation_requires_memory_gb`. I've added a new tab 'Evaluation' for this because there was no tab for which the settings seemed logical, but there is not enough room for this tab... Please advice. image

I think we don't want `evaluation_requires_gpu_type` to be settable by the challenge admin. This should only be adjustable by site admins. But anyway, don't you need to expose `selectable_gpu_types` and `maximum_settable_memory_gb` to challenge admins somewhere in the form? Probably best under the Algorithm tab.

app/grandchallenge/evaluation/models.py
amickan: Ah ok, fine in that case.

jmsmkn: Actually - we may need to split this. I can think of a case where the challenge admin would no longer be able to update the phase form:

- They ask support for large instances, so we set 64GB evaluations
- They want to edit their phase form, validation would fail
- They don't want participants to select more than 32GB

jmsmkn:
> I've added a new tab 'Evaluation' for this because there was no tab for which the settings seemed logical, but there is not enough room for this tab... Please advice.

Submission or scoring section I guess?

koopmant:
> * `evaluation_requires_memory_gb` this is currently in the scoring tab, `evaluation_requires_gpu_type` should also be added there. Sorry I forgot about that refactoring.

Okay. I initially created a new tab because I think "Scoring" should not really include evaluations settings. But without creating something new, I guess the scoring tab is currently the best place for it. 

> * The selectable memory and gpu types should be separated for algorithms and evaluations as there are instances where you want to decouple the two and you would otherwise end up with a non-submittable phase form (we currently have that now for SynthRAD2025).

Thanks for the clarification. I have renamed the settings and will create new ones for algorithms in the next PR

app/grandchallenge/evaluation/models.py
```diff
@@ -616,6 +641,15 @@ class Meta:
             ("configure_algorithm_phase", "Configure Algorithm Phase"),
         )
 
+    def get_selectable_gpu_type_choices_evaluation(self):
+        choices = {GPUTypeChoices.NO_GPU, GPUTypeChoices.T4}
```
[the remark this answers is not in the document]
jmsmkn: If you did that in `settings.py` you would need to import from the application, which won't work (or you will need to refactor for circular dependencies). I think it's okay there for now, otherwise you can turn it into a module constant, e.g.

```python
DEFAULT_GPU_CHOICES = [GPUTypeChoices.NO_GPU, GPUTypeChoices.T4]
```

Changed to a list rather than a set here as ordering is often important.

app/grandchallenge/evaluation/models.py
```diff
@@ -616,6 +641,15 @@ class Meta:
...
+        if self.selectable_gpu_type_choices_evaluation and isinstance(
+            self.selectable_gpu_type_choices_evaluation, list
+        ):
+            for choice in self.selectable_gpu_type_choices_evaluation:
+                choices.add(GPUTypeChoices[choice])
+        return choices
```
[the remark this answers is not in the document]
jmsmkn: Hmm, good question. Maybe rather than combining sets they are two independent lists?

app/grandchallenge/evaluation/models.py
```diff
@@ -134,6 +134,14 @@
     },
 }
 
+SELECTABLE_GPU_TYPES_SCHEMA = {
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "type": "array",
+    "title": "The Selectable GPU Types Schema",
+    "items": {"type": "string", "required": [], "properties": {}},
+    "uniqueItems": True,
+}
```
jmsmkn, line 143: We can make the schema stronger by integrating the options (this is untested! so we should add a test to make sure that providing a wrong selection is caught):

```suggestion
SELECTABLE_GPU_TYPES_SCHEMA = {
    "$schema": "http://json-schema.org/draft-07/schema",
    "type": "array",
    "title": "The Selectable GPU Types Schema",
    "items": {
        "type": "string",
        "enum": [choice.value for choice in GPUTypeChoices]
    },
    "uniqueItems": True
}
```

app/grandchallenge/evaluation/models.py
```diff
@@ -475,6 +483,25 @@ class Phase(FieldChangeMixin, HangingProtocolMixin, UUIDModel):
         blank=True,
         help_text="The output interfaces that the algorithms for this phase must use",
     )
+    selectable_gpu_type_choices_evaluation: list[str] = models.JSONField(
```
jmsmkn, line 486: Please do not add type annotations, they are not checked in this repo and the recommendation from mypy for integration is to always start without type hints:

```suggestion
    selectable_gpu_type_choices_evaluation = models.JSONField(
```

app/grandchallenge/evaluation/models.py
```diff
@@ -475,6 +483,25 @@ class Phase(FieldChangeMixin, HangingProtocolMixin, UUIDModel):
...
+        blank=True,
+        null=True,
```
jmsmkn, line 488: `null` should not be allowed as that would clash with the empty list which represents no choices.

app/grandchallenge/evaluation/models.py
```diff
@@ -475,6 +483,25 @@ class Phase(FieldChangeMixin, HangingProtocolMixin, UUIDModel):
...
+        default=list,
```
jmsmkn, line 489: Maybe the default choices should be `lambda: [GPUTypeChoices.NO_GPU, GPUTypeChoices.T4]` as discussed in the other comment?

app/grandchallenge/evaluation/models.py
```diff
@@ -475,6 +483,25 @@ class Phase(FieldChangeMixin, HangingProtocolMixin, UUIDModel):
         blank=True,
         help_text="The output interfaces that the algorithms for this phase must use",
     )
+    selectable_gpu_type_choices_evaluation: list[str] = models.JSONField(
+        blank=True,
```
jmsmkn, line 487: I'm not sure if we want to allow `blank`, does `[]` count as `blank`? If so, then we want to keep it, if not, we want to deny it.

app/grandchallenge/evaluation/models.py
```diff
@@ -475,6 +483,25 @@ class Phase(FieldChangeMixin, HangingProtocolMixin, UUIDModel):
...
+        null=True,
+        default=list,
+        help_text=(
+            "The GPU type choices that challenge admins will be able to set for the "
+            f"evaluation method. Options are {GPUTypeChoices.names}.".replace(
```
jmsmkn, line 492: I think this should be the values, not the names, as we want to match how these are represented in the database.

app/grandchallenge/evaluation/models.py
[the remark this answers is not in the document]
jmsmkn: Looking at this further I think it makes most sense to define this as the default to the field.

app/grandchallenge/evaluation/models.py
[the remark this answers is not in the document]
jmsmkn: Looking at this further I think that if we define it as the default we then only have one list, and can take the value of `selectable_gpu_type_choices_evaluation` directly rather than having to combine anything.

app/grandchallenge/evaluation/models.py
```diff
@@ -743,6 +778,37 @@ def _clean_external_evaluation(self):
                     "An external evaluation phase must have a parent phase."
                 )
 
+    def _clean_evaluation_requirements(self):
```
jmsmkn, line 781: This would need a bit of rework if you take on the review suggestions but it looks good!

jmsmkn (commented):
Looks good! I think the main thing is that we should be consistent with storing the choice values rather than the name - that way it is consistent with how the choices would work with the text field. The rest would need a little rework to work with that change.

app/grandchallenge/evaluation/models.py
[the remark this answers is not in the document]
koopmant: Settings are already imported here through `from django.conf import settings`, so circular dependency is not an issue.

> Changed to a list rather than a set here as ordering is often important.

I was looking into OrderedSets, but it requires another package. 

> Looking at this further I think it makes most sense to define this as the default to the field.  

I'll go with this for now!

app/grandchallenge/evaluation/models.py
koopmant: `[]` counts as blank. But I'm denying it, because we only have one list now; otherwise there are no options.

jmsmkn: What about CPU only: `[""]`?

app/grandchallenge/evaluation/models.py
jmsmkn: Ah, good to know!

app/grandchallenge/evaluation/models.py
jmsmkn: The values should never be changed as that would involve a data migration. It's still there for historical reasons. Having it as an empty string is fine - only site admins will be modifying this. If we ever open it up to others we will anyway need to rework the widget as we do not want people editing JSON fields directly, has proven to be difficult for users.

koopmant: Ah of course!

app/grandchallenge/evaluation/models.py
koopmant: Excellent! Already tested manually and this works as expected so I remove the manual validation  in the clean method. I'll write a test for it now.

Issue #3650, opened 2024-10-22
Exception when patching display set order


jmsmkn:
**Describe the bug**

When a user patches a display set and only sets the order a AttributeError is raised: This QueryDict instance is immutable on https://github.com/comic/grand-challenge.org/blob/01ebde194b3b5d6732e260c17aff02cd8e82aaff/app/grandchallenge/reader_studies/views.py#L987

**To Reproduce**

Make a patch request with only order set.

**Expected behavior**

The order is updated.

amickan:
I tried to replicate this with a test to start out with, but wasn't able to replicate it (yet). The test calls: 

```
response = get_view_for_user(
       viewname="api:reader-studies-display-set-detail",
       reverse_kwargs={"pk": ds.pk},
       user=editor,
       client=client,
       method=client.patch,
       content_type="application/json",
       data={
           "order": 20
       },
)
```

No error, and the order gets updated as expected. 

It's the end of the day, maybe I misunderstood the issue, will continue tomorrow.

jmsmkn:
Maybe speak with Daan Geijs about what he was doing, I assume the calls were him.

amickan:
I've contacted him

amickan:
Got a response from Daan that he was able to update the order through the API following your instructions, James. 

But then he wanted to remove some display sets and did the following, quote from Daan:

> Since the deletion endpoint is gone, my idea was to set the order of these display sets to a value that i can use the sort function and then batch delete on the website. I tried to update them all to order 999999.
> That did not work, then i tried to only set the title to "remove" for all, and that also did not work I remember. 

Maybe one of these caused the error. I still need to investigate. Setting all display sets to the same title and order shouldn't work, but it should not error out like it did....

amickan:
Going back to the roots and just looking at the error we got, it suggests that `request.data` was immutable, which it shouldn't be if the request were sent as `application/json`. Only if the request is sent with content type `application/x-www-form-urlencoded`, `request.data` will be a `QueryDict` instance (which is indeed immutable), see DRF docs here: https://www.django-rest-framework.org/api-guide/parsers/#jsonparser

I tried the test with the form content type and was finally able to replicate the error. 

Taking a quick look at gcapi, it looks like all requests through our api client are sent with content type `application/json`. Am I missing something?

jmsmkn:
Clients don't have to use GCAPI so it is easy for them send other data along. Good that you can now reproduce it.

amickan:
[...]

PR #3703, opened 2024-11-19
[WIP] Fix PUT for ArchiveItem/DisplaySet

### LINKED ISSUE #3701: Errors on updating (PUT) `ArchiveItem` and `DisplaySet`
I was playing around with the rse-gcapi and exploring what is needed, and when for updating (PUT) a CIVSet (i.e. DisplaySet/ArchiveItem).

A few observations:

## PUT `ArchiveItem`
```
{
  "values": [
    {
      "interface": "integer",
      "value": 13
    }
  ]
}
```

Requires an `"archive"` to be set, this is already an existing archive item, and should source from the current instance. Worse, this needs to be an api_url.

## PUT `DisplaySet`
Same body as the above put, however:
```
AssertionError: The `.update()` method does not support writable nested fields by default.
Write an explicit `.update()` method for serializer `grandchallenge.reader_studies.serializers.DisplaySetPostSerializer`, or set `read_only=True` on nested serializer fields.
```

chrisvanrun:
Fixes #3701


chrisvanrun:
Might also fix #3650

app/grandchallenge/components/serializers.py
```diff
@@ -172,3 +181,64 @@ class HyperlinkedComponentInterfaceValueSerializer(
...
+
+    def create(self, validated_data):
+        if "values" in validated_data and validated_data.pop("values") != []:
+            raise DRFValidationError("Values can only be added via update")
+        return super().create(validated_data)
+
+    def update(self, instance, validated_data):
+        request = self.context["request"]
+
+        values = validated_data.pop("values")
```
chrisvanrun, line 205: I thought the tests involved empty values, better double-check.

chrisvanrun: Yeah, missed a super call here.

app/grandchallenge/reader_studies/serializers.py
```diff
@@ -163,6 +165,12 @@ def __init__(self, *args, **kwargs):
                 codename="change_readerstudy",
             )
 
+    # def update(self, instance, validated_data):
+    #     assert (
+    #         not instance.is_editable
+    #     ), "This display set cannot be changed, as answers for it already exist."
+    #     return super().update(instance, validated_data)
```
amickan, line 172: This should be moved to the mixin, see my earlier comment.

amickan (commented):
LGTM! Just some minor comments

app/grandchallenge/components/serializers.py
```diff
@@ -172,3 +181,72 @@ class HyperlinkedComponentInterfaceValueSerializer(
...
+
+        if not self.partial:
+            instance.refresh_from_db()
+            current_civs = {
+                civ.interface.slug: civ for civ in instance.values.all()
+            }
+            current_interfaces = set(current_civs.keys())
+            updated_interfaces = {v["interface"].slug for v in values}
+            for interface in current_interfaces - updated_interfaces:
+                self.instance.remove_civ(civ=current_civs[interface])
```
amickan, line 250: If this happens for image interfaces, the user will loose permission to use the image. I remember a discussion with James that if we allow removal of images from display sets or archive items, that they need to be moved to a different AI or DS first, so that the user doesn't loose access, see [here](https://github.com/comic/grand-challenge.org/pull/3156#issuecomment-1884507499). I'm not sure if this is still a requirement, but thought I'd mention it. It feels a bit like a strange side-effect to create a new DS/AI for the removed image, especially given that users can then remove the AI/DS and lose access anyway. Whichever way we go, I think the consequences need to be clear to the user (either losing access or creating a new AI/DS), I think.

chrisvanrun: Huh? The idea here is that replacing an image-linked CIV within the same CIVSet works because the new CIV is first created and the old one is only later removed. This should keep the linked image bound to the CIVSet. Much what James is warning against.

From your description, it seems you suggest to move it to a completely separate CIVSet? That seems to go against the idea of PUT (or REST at that). 

If users decide to run a (somewhat weird) update workflow against a reader-study or archive where they locally store a reference to an image, then PUT completely new values to a CIVSet and then decide to create a new CIV off of the image ref only to find it's no longer accessible... that is really their problem. We can only guarantee things to stay cohesive in the scope of a single request.






with an image linked gets used by a different interface (with the same CIVSet) the new CIV

jmsmkn: This is a problem but I think the solution of creating a new archive item to hold is was not the correct one (rather, the `view_image` permission should have been attached through the upload session too). I don't think we need to maintain that solution if we still have it - I thought that we removed it?

amickan: Ah true, we no longer do that. When someone uploads a new image to an archive item, we don't create a new item with the previous image, same for DS. We used to do that, at least for display sets, but that was removed indeed.

chrisvanrun: Not 100% sure, does this PR pivot on the view_image permissions that still need to be added via user upload?

amickan: Not sure either, here is how it works right now: We do not attach view permissions to upload sessions at the moment. Image viewing permissions are set through the objects the images are attached to (job, display sets, archive items, rs answers). The piece of code that handles the logic is `update_viewer_groups_permissions`. This gets called when an archive item or display set get edited, for example. 

I didn't mean to suggest that creating a new CIV set is the right/best solution, I just mentioned this because that is how we used to deal with the issue of losing access . Attaching permissions to the upload session would be better - these persist forever, so we don't delete upload session, like we do with user uploads, right? 

But I think that change requires quite some changes in the codebase, so not something we would do in a cooldown week PR, IMO. Is this a blocker for this PR? I actually don't think so since we anyway already removed the creation of new CIVsets, so users are already losing access to their images if they update the DS or AI... but maybe James can comment?

jmsmkn: It wouldn't be that much of a migration to be honest - Images already have user "view_image" permissions for answer creators. It gets in to an area where we have to think about behaviour though: is the user who uploads the image really the owner of the data? Presumably the user is currently working for an organisation, and they're uploading the organisations data, in which case they should probably loose access to the data if they leave their job. Likewise - you could argue this is the case for any other object in GC (Challenges, Archives, Reader Studies) - so the usual way to handle this is that they forfeit their account when they leave the organisation, just like they forfeit their email address.

So, I don't know what to do here. Users already lose access when replacing, no one has actually complained about it AFAIK, so I don't see a user-need to change the behaviour.

amickan: Agreed. So then this can be merged.

amickan (approved):

PR #3705, opened 2024-11-19
Add default ordering for `Notification`s

### LINKED ISSUE #3704: Sporadic failure on `test_permission_request_notifications_flow_for_manual_review`
See run: https://github.com/comic/grand-challenge.org/actions/runs/11909994990/job/33188499636

I think it is quickly fixable by ordering the notifications.

chrisvanrun:
Fixes #3704

Likely the cause of the sporadic test failure.


jmsmkn:
Rather than adding default ordering the test should be refactored to not rely on ordering. This is because that ordering is not free, so should only be done where it is required.

PR #3713, opened 2024-11-21
Process Newick files on `acks-late-2xlarge`

chrisvanrun:
This tiny PR introduces an (optional) custom upload processing queue per component interface. Not sure if we should also add this to the processing of images. I suggest we tackle that when we want to define a custom queue for super_kind images.

# TODO
- [x] re-define tests so this is actual tested.

### DIFF
--- a/app/grandchallenge/components/models.py
+++ b/app/grandchallenge/components/models.py
@@ -555,13 +555,6 @@ class ComponentInterface(OverlaySegmentsMixin):
         except KeyError as e:
             raise RuntimeError(f"Unknown kind {self.kind}") from e
 
-    @property
-    def custom_upload_processing_queue(self):
-        try:
-            return INTERFACE_KIND_TO_CUSTOM_QUEUE[self.kind]
-        except KeyError:
-            return False  # No custom queue
-
     def create_instance(self, *, image=None, value=None, fileobj=None):
         civ = ComponentInterfaceValue.objects.create(interface=self)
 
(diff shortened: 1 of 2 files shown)


jmsmkn:
> Not sure if we should also add this to the processing of images.

That should already be done, it doesn't go through the add file to object route as they are images. Validation of those is done in process outputs.

app/grandchallenge/components/models.py
```diff
@@ -554,6 +554,13 @@ def file_extension(self):
         except KeyError as e:
             raise RuntimeError(f"Unknown kind {self.kind}") from e
 
+    @property
+    def custom_upload_processing_queue(self):
+        try:
+            return INTERFACE_KIND_TO_CUSTOM_QUEUE[self.kind]
+        except KeyError:
+            return None  # No custom queue: use default
```
jmsmkn, line 562: Probably doesn't even need a method!

```suggestion
        return INTERFACE_KIND_TO_CUSTOM_QUEUE.get(self.kind)
```

chrisvanrun:
James' hunch was right: passing `None` to the signature defaults to the "celery" queue. I've added a test to check the custom queue integration.

app/tests/components_tests/test_models.py
```diff
@@ -1634,3 +1636,47 @@ def test_component_interface_value_manager():
...
+    )
+    assert mock_task_signature.called_once()
+
+    kwargs = mock_task_signature.call_args.kwargs
+    if not expected_queue:
+        # Ensure it's not passed since setting to None would
+        # have Celery use the Celery-scope default ("celerly")
+        assert "queue" not in kwargs
+    else:
+        assert kwargs["queue"] == expected_queue
```
jmsmkn, line 1682: Reduces test logic:

```suggestion
    assert kwargs.get("queue") == expected_queue
```

The comment could then be moved to the parametrisation

jmsmkn (approved):

PR #3714, opened 2024-11-22
Adjustable gpu types for algorithms

koopmant:
Make the selectable gpu type and maximum memory setting for algorithms adjustable using Phase settings. relevant phases are filtered when updating or creating an algorithm, this is based on inputs and outputs match, participant match, the phase submissions kind must be algorithm, and the phase must be public. 
The new Phase settings are also used to validate algorithm gpu and memory settings when it is submitted to the phase. 


app/grandchallenge/evaluation/forms.py
```diff
@@ -633,11 +664,11 @@ def check_submission_limit_avoidance(self, *, creator):
     def save(self, *args, **kwargs):
         if self._phase.submission_kind == SubmissionKindChoices.ALGORITHM:
             self.instance.algorithm_requires_gpu_type = self.cleaned_data[
-                "algorithm_image"
-            ].algorithm.job_requires_gpu_type
+                "algorithm"
```
jmsmkn, line 667: Why the change here to look at the algorithm rather than the algorithm image?

app/grandchallenge/evaluation/forms.py
```diff
@@ -542,6 +542,37 @@ def clean_algorithm(self):
...
+        if (
+            algorithm.job_requires_gpu_type
+            not in phase.algorithm_selectable_gpu_type_choices
+        ):
+            job_requirement_errors.append(
+                ValidationError(
+                    "The requested GPU type for this algorithm "
+                    f"({GPUTypeChoices(algorithm.job_requires_gpu_type).label}) is "
+                    "not allowed for this phase. Options are "
+                    f"{[GPUTypeChoices(c).label for c in phase.algorithm_selectable_gpu_type_choices]}. "
```
jmsmkn, line 569: Is using `.label` necessary in these places?

app/grandchallenge/algorithms/forms.py
```diff
@@ -365,21 +365,77 @@ class Meta:
...
+        ).filter(
+            submission_kind=SubmissionKindChoices.ALGORITHM,
+            public=True,
+            total_algorithm_input_count=len(inputs),
+            total_algorithm_output_count=len(outputs),
+            relevant_algorithm_input_count=len(inputs),
+            relevant_algorithm_output_count=len(outputs),
+        )
+
+    def get_selectable_gpu_type_choices(self):
```
jmsmkn, line 416: Anytime you have a `def get...` think if it can be replaced by an `@property`.

app/grandchallenge/algorithms/forms.py
```diff
@@ -365,21 +365,77 @@ class Meta:
...
+                .values_list(
+                    "algorithm_selectable_gpu_type_choices", flat=True
+                )
+                .distinct()
+            )
+        )
+        choices = list(dict.fromkeys(choices))
+        return [(choice, GPUTypeChoices(choice).label) for choice in choices]
+
+    def get_maximum_settable_memory_gb(self):
```
jmsmkn, line 430: Same here.

app/grandchallenge/algorithms/forms.py
```diff
@@ -365,21 +365,77 @@ class Meta:
...
+        self.fields["job_requires_gpu_type"].choices = (
+            self.get_selectable_gpu_type_choices()
+        )
         self.fields["job_requires_memory_gb"].validators = [
             MinValueValidator(settings.ALGORITHMS_MIN_MEMORY_GB),
-            MaxValueValidator(settings.ALGORITHMS_MAX_MEMORY_GB),
+            MaxValueValidator(self.get_maximum_settable_memory_gb()),
         ]
 
+    @cached_property
```
jmsmkn, line 383: I don't think that caching will help here as it is returning an unevaluated queryset.

app/grandchallenge/algorithms/forms.py
```diff
@@ -365,21 +365,77 @@ class Meta:
...
+        choices = list(dict.fromkeys(choices))
+        return [(choice, GPUTypeChoices(choice).label) for choice in choices]
+
+    def get_maximum_settable_memory_gb(self):
+        value = settings.ALGORITHMS_MAX_MEMORY_GB
+        maxima_in_phases = self.relevant_phases.order_by(
+            "-algorithm_maximum_settable_memory_gb"
+        ).values_list("algorithm_maximum_settable_memory_gb", flat=True)
+        if maxima_in_phases:
+            value = max(value, maxima_in_phases[0])
```
jmsmkn, line 436: Not sure if this will work but you can use database filtering here rather than fetching all records:

```suggestion
        ).values_list("algorithm_maximum_settable_memory_gb", flat=True).first()
        if maxima_in_phases is not None:
            value = max(value, maxima_in_phases)
```

app/grandchallenge/algorithms/forms.py
```diff
@@ -365,21 +365,77 @@ class Meta:
...
+            self.get_selectable_gpu_type_choices()
+        )
         self.fields["job_requires_memory_gb"].validators = [
             MinValueValidator(settings.ALGORITHMS_MIN_MEMORY_GB),
-            MaxValueValidator(settings.ALGORITHMS_MAX_MEMORY_GB),
+            MaxValueValidator(self.get_maximum_settable_memory_gb()),
         ]
 
+    @cached_property
+    def relevant_phases(self):
```
jmsmkn, line 384: This seems relatively familiar, was it duplicated from somewhere?

app/grandchallenge/evaluation/forms.py
koopmant: It's completely unrelated to this PR, sorry. I stumbled upon this and wondered why we look at the algorithm image to get the algorithm when we can get the algorithm directly. I changed it to test if something would break.

app/grandchallenge/algorithms/forms.py
koopmant: Good to know. Ironically, I actually prefer properties but have been holding myself back from using them, because I wasn't sure if others agree.

app/grandchallenge/algorithms/forms.py
koopmant: Yes, I copied it from the AlgorithmsForPhaseMixin, [this part (L396-436)](https://github.com/comic/grand-challenge.org/pull/3714/files/f8560de9d07fba4a3bdd505a37d298cd9a890cf1#diff-800ed244caee41acb2694b83ed490082bcfe999937aa38f5e3cf1f4fdf0ef8ceL396-L436)

jmsmkn: Is there a way to reuse this rather than duplicating it?

app/grandchallenge/algorithms/forms.py
```diff
@@ -365,21 +365,77 @@ class Meta:
...
+            chain.from_iterable(
+                self.relevant_phases.order_by()
+                .values_list(
+                    "algorithm_selectable_gpu_type_choices", flat=True
+                )
+                .distinct()
+            )
+        )
+        choices = list(dict.fromkeys(choices))
+        return [(choice, GPUTypeChoices(choice).label) for choice in choices]
```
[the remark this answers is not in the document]
koopmant: Ah good point. I was assuming all choices I would get from the queryset would be still be in order but that's not the case of course.

app/grandchallenge/algorithms/forms.py
koopmant: Sorry, I didn't mean it is an exact copy. I just copied the way the queryset uses annotations and then filters them. The code there is actually doing the opposite (filtering for algorithms based on a phase, instead of filtering for phases based on an algorithm). I don't see a way of generalising this.

app/grandchallenge/evaluation/forms.py
koopmant: I thought it would be nicest at first, because this is the representation that the user can know from the dropdown menu. But the labels are quite long to use them like this. I've changed it to using the name instead.

app/grandchallenge/algorithms/forms.py
jmsmkn: Okay that's fine!

app/grandchallenge/algorithms/forms.py
```diff
@@ -365,21 +365,77 @@ class Meta:
...
+            total_algorithm_output_count=len(outputs),
+            relevant_algorithm_input_count=len(inputs),
+            relevant_algorithm_output_count=len(outputs),
+        )
+
+    def get_selectable_gpu_type_choices(self):
+        choices = [GPUTypeChoices.NO_GPU, GPUTypeChoices.T4]
+        choices.extend(
+            chain.from_iterable(
+                self.relevant_phases.order_by()
```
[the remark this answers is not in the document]
koopmant: Not using it anymore now, since I replaced it with ArrayAgg, but just for future reference:
I ran into the issue where distinct() was still returning duplicate values. According to the documentation: 
> If you order by fields from a related model, those fields will be added to the selected columns and they may make otherwise duplicate rows appear to be distinct. 

Adding the `order_by()` call without any arguments removes any ordering (including the default ordering). 

This solved the issue of still getting duplicate values. The SQL query changed from this: 
```sql
SELECT DISTINCT "evaluation_phase"."algorithm_selectable_gpu_type_choices",
       "challenges_challenge"."id",
       "evaluation_phase"."submissions_open_at",
       "evaluation_phase"."created"
  FROM "evaluation_phase"
  LEFT OUTER JOIN "evaluation_phase_algorithm_inputs"
    ON ("evaluation_phase"."id" = "evaluation_phase_algorithm_inputs"."phase_id")
  LEFT OUTER JOIN "evaluation_phase_algorithm_outputs"
    ON ("evaluation_phase"."id" = "evaluation_phase_algorithm_outputs"."phase_id")
 INNER JOIN "challenges_challenge"
    ON ("evaluation_phase"."challenge_id" = "challenges_challenge"."id")
 WHERE ("evaluation_phase"."public" AND "evaluation_phase"."submission_kind" = 3)
 GROUP BY "evaluation_phase"."id",
          "challenges_challenge"."id"
HAVING (COUNT(DISTINCT "evaluation_phase_algorithm_inputs"."componentinterface_id") FILTER (WHERE ("evaluation_phase_algorithm_inputs"."componentinterface_id" IN (SELECT U0."id" FROM "components_componentinterface" U0 INNER JOIN "algorithms_algorithm_inputs" U1 ON (U0."id" = U1."componentinterface_id") WHERE U1."algorithm_id" = '48d17a78e3c8433ca2b14f4932803cfe'::uuid))) = 0 AND COUNT(DISTINCT "evaluation_phase_algorithm_outputs"."componentinterface_id") FILTER (WHERE ("evaluation_phase_algorithm_outputs"."componentinterface_id" IN (SELECT U0."id" FROM "components_componentinterface" U0 INNER JOIN "algorithms_algorithm_outputs" U1 ON (U0."id" = U1."componentinterface_id") WHERE U1."algorithm_id" = '48d17a78e3c8433ca2b14f4932803cfe'::uuid))) = 0 AND COUNT(DISTINCT "evaluation_phase_algorithm_inputs"."componentinterface_id") = 0 AND COUNT(DISTINCT "evaluation_phase_algorithm_outputs"."componentinterface_id") = 0)
```

to this:
```sql
SELECT DISTINCT "evaluation_phase"."algorithm_selectable_gpu_type_choices"
  FROM "evaluation_phase"
  LEFT OUTER JOIN "evaluation_phase_algorithm_inputs"
    ON ("evaluation_phase"."id" = "evaluation_phase_algorithm_inputs"."phase_id")
  LEFT OUTER JOIN "evaluation_phase_algorithm_outputs"
    ON ("evaluation_phase"."id" = "evaluation_phase_algorithm_outputs"."phase_id")
 WHERE ("evaluation_phase"."public" AND "evaluation_phase"."submission_kind" = 3)
 GROUP BY "evaluation_phase"."id"
HAVING (COUNT(DISTINCT "evaluation_phase_algorithm_inputs"."componentinterface_id") FILTER (WHERE ("evaluation_phase_algorithm_inputs"."componentinterface_id" IN (SELECT U0."id" FROM "components_componentinterface" U0 INNER JOIN "algorithms_algorithm_inputs" U1 ON (U0."id" = U1."componentinterface_id") WHERE U1."algorithm_id" = 'c9eecb91a8fc49d3b7fa8a701ac3984e'::uuid))) = 0 AND COUNT(DISTINCT "evaluation_phase_algorithm_outputs"."componentinterface_id") FILTER (WHERE ("evaluation_phase_algorithm_outputs"."componentinterface_id" IN (SELECT U0."id" FROM "components_componentinterface" U0 INNER JOIN "algorithms_algorithm_outputs" U1 ON (U0."id" = U1."componentinterface_id") WHERE U1."algorithm_id" = 'c9eecb91a8fc49d3b7fa8a701ac3984e'::uuid))) = 0 AND COUNT(DISTINCT "evaluation_phase_algorithm_inputs"."componentinterface_id") = 0 AND COUNT(DISTINCT "evaluation_phase_algorithm_outputs"."componentinterface_id") = 0)
```

app/grandchallenge/algorithms/forms.py
koopmant: That works. I also tested Max aggregation, but that appears to be slightly slower. So I went with this method.

app/grandchallenge/algorithms/forms.py
koopmant: I now removed cached from the decorator for this reason. I also looked into whether we could combine the two queries into a single one using annotations. But that ultimately ended up being slower, since I would have to loop through the resulting annotated queryset in python. The current code with two separate queries is also much easier to read. So I left it with two separate queries.

app/grandchallenge/algorithms/forms.py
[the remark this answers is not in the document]
jmsmkn: Good to know!

app/grandchallenge/algorithms/forms.py
```diff
@@ -365,21 +366,87 @@ class Meta:
...
+    @property
+    def maximum_settable_memory_gb(self):
+        value = settings.ALGORITHMS_MAX_MEMORY_GB
+        maximum_in_phases = (
+            self.relevant_phases.order_by(
+                "-algorithm_maximum_settable_memory_gb"
+            )
+            .values_list("algorithm_maximum_settable_memory_gb", flat=True)
+            .first()
+        )
```
jmsmkn, line 445: This will still cause the new queries to go to the database twice. Once for the max memory, once for the selectable gpus. Did you consider doing both in a single aggregation, for instance using the `Max` method? See https://docs.djangoproject.com/en/5.1/topics/db/aggregation/#generating-aggregates-over-a-queryset

If that works, you could move the aggregation to `relevant_phases` (probably change the name as it would then return a map with two keys, max memory and selectable gpus), make that a `cached_property`, then have the two properties extract and clean the values returned from the database.

If that is unclear I can meet at 9am EU time on Thursday to explain what I mean, let me know!

jmsmkn: Cool use of `ArrayAgg` and `distinct`, didn't know about that either.

koopmant: I think I understand, and I tried that. But what I ended up with was something like this and when I timed it, it was slower:
```python
query = list(Phase.objects.annotate(
    max_memory=Max('algorithm_maximum_settable_memory_gb'),
    choices=ArrayAgg("algorithm_selectable_gpu_type_choices", distinct=True)
    ).values_list('max_memory', 'choices'))

max_value = query[0][0]  # The maximum value is constant across rows
choices = [row[1][0] for row in query]  # Extract array_field values

set(chain.from_iterable(choices))
```

koopmant: Hold on, I think I get my issue. I should not use annotate here, but aggregate instead.

koopmant: I think I got it now. Could you look at it again?

jmsmkn: Yes that is what I meant!

app/grandchallenge/algorithms/forms.py
```diff
@@ -365,20 +366,87 @@ class Meta:
...
+
+    @property
+    def selectable_gpu_type_choices(self):
+        choices = [GPUTypeChoices.NO_GPU, GPUTypeChoices.T4]
+        choices.extend(
+            chain.from_iterable(
+                self.job_requirement_properties_from_phases["gpu_type_choices"]
+            )
+        )
+        choices_set = set(choices)
```
jmsmkn, line 434: GitHub messes with the formatting but this might be more straightforward:

```suggestion
        choices_set = {
            GPUTypeChoices.NO_GPU, 
            GPUTypeChoices.T4, 
            *chain.from_iterable(self.job_requirement_properties_from_phases["gpu_type_choices"])
        }
```

koopmant: Yes, that's much cleaner! Thank you

app/grandchallenge/algorithms/forms.py
amickan: This is still a `cached_property` and would it be faster to first filter the queryset by `public` and `submission_kind` before annotating the counts?

app/tests/algorithms_tests/test_forms.py
```diff
@@ -811,6 +812,50 @@ def test_algorithm_form_gpu_limited_choices():
...
 
 
+@pytest.mark.django_db
+def test_algorithm_form_gpu_choices_from_phases():
+    user = UserFactory()
+    p1 = PhaseFactory(
+        submission_kind=SubmissionKindChoices.ALGORITHM,
+        algorithm_selectable_gpu_type_choices=[
+            GPUTypeChoices.NO_GPU,
+            GPUTypeChoices.T4,
```
amickan, line 822: I think this and the other tests only test the logic partially. They are missing inputs and outputs.

app/grandchallenge/algorithms/forms.py
koopmant: When I inspect the executed SQL, they are the same regardless of where I place the filter for `public` and `submission_kind`. Apparently there is some optimization going on under the hood.

app/tests/algorithms_tests/test_forms.py
koopmant: Thanks, you are right. (of course :angel:) 
Updated them

amickan: Looks good! For a complete test, I think you would also want a phase that has, for example, 2 inputs (`ci_in` and another one) to make sure the `distinct` in the `Count` annotation does its job.

koopmant: Good point. Added it, thanks!

PR #3716, opened 2024-11-22
Add example to overlay_segments help text

ammar257ammar:
This is related to Part 2 "Overlay segments" of pitch https://github.com/DIAGNijmegen/rse-roadmap/issues/362.

The overlay_segments is used in three places:
- It is used in [OverlaySegmentsMixin](https://github.com/comic/grand-challenge.org/blob/cdf9725db3e6b39c1ae69dfb0becb19938011d2e/app/grandchallenge/components/models.py#L311) which is used in:
   - [ComponentInterface](https://github.com/comic/grand-challenge.org/blob/cdf9725db3e6b39c1ae69dfb0becb19938011d2e/app/grandchallenge/components/models.py#L387) model
   - [Question](https://github.com/comic/grand-challenge.org/blob/cdf9725db3e6b39c1ae69dfb0becb19938011d2e/app/grandchallenge/reader_studies/models.py#L1233) model (reader studies)
- It is used in the [WorkstationConfig](https://github.com/comic/grand-challenge.org/blob/cdf9725db3e6b39c1ae69dfb0becb19938011d2e/app/grandchallenge/workstation_configs/models.py#L131) model.

This PR adds a default JSON example and a link to documentation to the overlay_segments field.


ammar257ammar:
Sphinx issue fixed

ammar257ammar:
Since the example added to the overlay_segments form field is fixed (hardcoded), is a form test still recommended to check it is there in the help_text?

app/grandchallenge/workstation_configs/forms.py
```diff
@@ -84,6 +86,18 @@ def __init__(self, *args, read_only=False, **kwargs):
...
         )
 
+        self.fields["overlay_segments"].help_text += format_lazy(
+            ". If a categorical overlay is shown, "
+            "it is possible to show toggles to change the visibility of the different overlay categories. "
+            "To do so, configure the categories that should be displayed. Data from the "
+            "algorithm's output.json can be added as an extra label to each "
+            "toggle using jinja templating. "
+            'Example usage: [{{"name": "background", "visible": true, "voxel_value": 0}},'
+            '{{"name": "tissue", "visible": true, "voxel_value": 1}}]. '
```
amickan, line 96: The example no longer contains the jinja templating example. Is that on purpose?

app/grandchallenge/components/models.py
```diff
@@ -314,7 +314,10 @@ class OverlaySegmentsMixin(models.Model):
         default=list,
         help_text=(
             "The schema that defines how categories of values in the overlay "
-            "images are differentiated."
+            "images are differentiated. "
+            "Example usage: "
```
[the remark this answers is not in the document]
ammar257ammar: Done

app/grandchallenge/workstation_configs/forms.py
ammar257ammar: Good point. Since this help_text already have an example, maybe it does not need to be changed to the default one which also keep the jinja example. I will ask Miriam on the pitch.

amickan:
> Since the example added to the overlay_segments form field is fixed (hardcoded), is a form test still recommended to check it is there in the help_text?

No, we don't normally test static helptext.

app/grandchallenge/workstation_configs/forms.py
amickan: Yes, or incorporate the jinja example into the default help_text? It is relevant in all places, or is it not?

app/grandchallenge/workstation_configs/forms.py
```diff
@@ -84,6 +86,11 @@ def __init__(self, *args, read_only=False, **kwargs):
             ),
         )
 
+        self.fields["overlay_segments"].help_text += format_lazy(
+            'Refer to the <a href="{}#segmentation-masks">documentation</a> for more information',
+            reverse("documentation:detail", args=["interfaces"]),
```
amickan, line 91: I'm only seeing this now, but does this work? Don't you need to pass the `slug` along as a `kwargs`? Also, it's better to create an environment variable for the slug so that it's configurable and can be changed if necessary. See `DOCUMENTATION_HELP_FORUM_SLUG` as example.

ammar257ammar: "interfaces" in the slug that is passed in args.
This will create the URL: https://GRAND_CHALLENGE_DOMAIN/documentation/interfaces
locally, it will create: https://gc.localhost/documentation/interfaces
And should create the 
I will add an environment variable to enable changing it if necessary.

amickan: Still better to use key word arguments, makes it explicit what you're passing to the resolver.

ammar257ammar: Done

amickan (approved):

PR #3719, opened 2024-11-27
Add AlgorithmInterface model

amickan:
First PR for https://github.com/DIAGNijmegen/rse-roadmap/issues/153

This adds the `AlgorithmInterface` model, as well as the necessary through table, and removes the inputs and outputs fields from the algorithm form. 

The following restrictions apply:
- AlgorithmInterfaces cannot be deleted (protection on model level and admin form level)
- AlgorithmInterfaces cannot be updated (protection on admin form level)
- AlgorithmInterfaces are unique (i.e. input and output combinations are unique, we check for that in the form)
- An algorithm can only have 1 default algorithm interface
- There can only be one entry per algorithm and interface in the through table.


app/grandchallenge/algorithms/migrations/0063_algorithminterface_algorithminterfacethroughtable_and_more.py
```diff
@@ -0,0 +1,137 @@
...
+                input_count=Count("inputs", distinct=True),
+                output_count=Count("outputs", distinct=True),
+            ).get(
+                inputs__in=inputs,
+                outputs__in=outputs,
+                input_count=len(inputs),
+                output_count=len(outputs),
+            )
+        except ObjectDoesNotExist:
+            io = AlgorithmInterface.objects.create()
```
amickan, line 39: We cannot use the model manager method that I added here because migrations work with historic model instances, and don't have access to custom manager methods or model methods. Is that ok given that this is a one off migration? Or would it be better if I created a function to do this that I import here and in all other places where I need it (i.e. the forms)? A manager method seemed better than a separate utility function to me, but I'm open for other suggestions.

amickan:
Somehow it didn't let me select the feature branch as merge target. And the migration check passes locally as well. Need to look into what's going wrong.

amickan:
Creating a new PR with the correct target worked. I guess I just overlooked the option when I created this one. Closing this in favour of https://github.com/comic/grand-challenge.org/pull/3720

PR #3720, opened 2024-11-27
Add AlgorithmInterface model

amickan:
First PR for https://github.com/DIAGNijmegen/rse-roadmap/issues/153

This adds the `AlgorithmInterface` model, as well as the necessary through table, and removes the inputs and outputs fields from the algorithm form. 

The following restrictions apply:
- AlgorithmInterfaces cannot be deleted (protection on model level and admin form level)
- AlgorithmInterfaces cannot be updated (protection on admin form level)
- AlgorithmInterfaces are unique (i.e. input and output combinations are unique, we check for that in the form)
- An algorithm can only have 1 default algorithm interface
- There can only be one entry per algorithm and interface in the through table.


app/grandchallenge/algorithms/migrations/0063_algorithminterface_algorithminterfacethroughtable_and_more.py
```diff
@@ -0,0 +1,137 @@
...
+        outputs = ComponentInterface.objects.filter(
+            pk__in=algorithm.outputs.values_list("pk", flat=True)
+        ).all()
+
+        try:
+            io = AlgorithmInterface.objects.annotate(
+                input_count=Count("inputs", distinct=True),
+                output_count=Count("outputs", distinct=True),
+            ).get(
+                inputs__in=inputs,
```
amickan, line 33: We cannot use the model manager method here because migrations work with historic model instances, and don't have access to custom manager methods or model methods. Is that ok given that this is a one off migration? Or would it be better if I created a function to do this that I import here and in all other places where I need it (i.e. the forms)? A manager method seemed better than a separate utility function to me, but I'm open for other suggestions.

app/grandchallenge/algorithms/models.py
```diff
@@ -69,6 +69,38 @@
...
+                inputs__in=inputs,
+                outputs__in=outputs,
+                input_count=len(inputs),
+                output_count=len(outputs),
+            )
+        except ObjectDoesNotExist:
+            return None
+
+
+class AlgorithmInterface(models.Model):
```
jmsmkn, line 90: We probably should use a `UUIDModel` as we will likely need to serialize these at some point.

app/grandchallenge/algorithms/models.py
```diff
@@ -69,6 +69,38 @@
...
+    inputs = models.ManyToManyField(
+        to=ComponentInterface, related_name="inputs"
+    )
+    outputs = models.ManyToManyField(
+        to=ComponentInterface, related_name="outputs"
+    )
```
jmsmkn, line 96: All `ManyToManyField`s should implement a through table. I think this is in our code guidelines, if not, please add it.

app/grandchallenge/algorithms/models.py
```diff
@@ -69,6 +69,38 @@
...
+
+    objects = AlgorithmInterfaceManager()
+
+    def delete(self, *args, **kwargs):
+        raise ValidationError("AlgorithmInterfaces cannot be deleted.")
```
jmsmkn, line 101: What about if a bulk delete is used? Can we catch if the inputs or outputs are changed?

app/grandchallenge/algorithms/models.py
```diff
@@ -521,6 +558,28 @@ def form_field_label(self):
...
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["algorithm"],
+                condition=Q(is_default=True),
+                name="unique_default_interface_per_algorithm",
+            ),
+            models.UniqueConstraint(
+                fields=["algorithm", "interface"],
```
jmsmkn, line 574: Is this necessary? I thought that `ManyToMany` would handle this?

app/grandchallenge/algorithms/models.py
```diff
@@ -69,6 +69,38 @@
 JINJA_ENGINE = sandbox.ImmutableSandboxedEnvironment()
 
 
+class AlgorithmInterfaceManager(models.Manager):
+    def get_existing_interface_for_inputs_and_outputs(
+        self, *, inputs, outputs
+    ):
+        try:
+            return self.annotate(
```
jmsmkn, line 77: There might be a bug here that I think we faced in algorithm input matching before. I think the current query here would match when searching for `inputs=[A, B]` if the inputs on the object were `[A, C]`.

app/grandchallenge/algorithms/migrations/0063_algorithminterface_algorithminterfacethroughtable_and_more.py
jmsmkn: Maybe create a method that the manager uses and can also be imported here. Important to reuse and test as I think there is the same bug as I just mentioned in this query too.

app/grandchallenge/algorithms/migrations/0063_algorithminterface_algorithminterfacethroughtable_and_more.py
```diff
@@ -0,0 +1,137 @@
...
+        "components", "ComponentInterface"
+    )
+
+    for algorithm in Algorithm.objects.all():
+        inputs = ComponentInterface.objects.filter(
+            pk__in=algorithm.inputs.values_list("pk", flat=True)
+        ).all()
+        outputs = ComponentInterface.objects.filter(
+            pk__in=algorithm.outputs.values_list("pk", flat=True)
+        ).all()
```
jmsmkn, line 26: Why not `Algorithm.objects.prefetch_related("inputs", "outputs")` and `inputs = algorithm.inputs.all()` and `outputs = algorithm.outputs.all()`?

app/grandchallenge/algorithms/admin.py
```diff
@@ -54,6 +60,7 @@ class AlgorithmAdmin(GuardedModelAdmin):
         "title",
         "created",
         "public",
+        "default_io",
```
jmsmkn, line 63: Aren't the inputs and outputs still editable in the algorithm admin?

app/grandchallenge/algorithms/models.py
amickan: It wasn't in the guidelines yet. I have added it to our guidelines.

app/grandchallenge/algorithms/models.py
amickan: To prevent bulk deletion, I can override the `delete` method on the model manager. Catching if inputs and outputs have changed is not possible, I'm afraid. I had mentioned this on the pitch before. The only thing we can do is make them ineditable in the admin (this is what `def has_change_permission(self, request, obj=None): return False` does) and by not allowing these fields to be updated by the user later on through the form (we will have a GetOrCreate view and then updating will be restricted to changing the `is_default` on the through table entry, I'd say. Is that sufficient?

app/grandchallenge/algorithms/models.py
amickan: Hadn't thought of that, thanks!

app/grandchallenge/algorithms/migrations/0063_algorithminterface_algorithminterfacethroughtable_and_more.py
amickan: Model methods have the same issue. i.e. can't be called in a migration. The only alternative I can think of then is to write a standalone function that does this and call that in all places where we need it.

app/grandchallenge/algorithms/models.py
amickan: Yes, this is necessary, or else the through table can contain multiple rows for the same interface and algorithm. Why would a ManyToMany handle this be default?

app/grandchallenge/algorithms/models.py
amickan: You are completely right. I will update the annotation and filtering accordingly.

app/grandchallenge/algorithms/admin.py
amickan: True. I made them read only now.

app/grandchallenge/algorithms/migrations/0063_algorithminterface_algorithminterfacethroughtable_and_more.py
amickan: I vaguely remember that this didn't work (related to the fact that migrations work with historical models), but I can't replicate it now, so your version it is.

app/grandchallenge/algorithms/migrations/0063_algorithminterface_algorithminterfacethroughtable_and_more.py
amickan: Actually, since migrations only work with historical models, I will need to also pass the model to the function. I'm leaving that as an optional argument to the function with a default at the current `AlgorithmInterface` model.

app/grandchallenge/algorithms/models.py
jmsmkn: Because methods like `algorithm.inputs.add(foo)` are idempotent.

app/grandchallenge/algorithms/migrations/0063_algorithminterface_algorithminterfacethroughtable_and_more.py
jmsmkn: Does the data migration also need to go into its own file? I remember seeing problems with combined table and data modifications in the past.

app/grandchallenge/algorithms/models.py
```diff
@@ -69,6 +69,38 @@
 JINJA_ENGINE = sandbox.ImmutableSandboxedEnvironment()
 
 
+class AlgorithmInterfaceManager(models.Manager):
```
jmsmkn, line 72: Should the create method also be overridden?

app/grandchallenge/algorithms/models.py
jmsmkn: Yes probably

app/grandchallenge/algorithms/migrations/0063_algorithminterface_algorithminterfacethroughtable_and_more.py
amickan: In my tests (with an admittedly small sample of algorithms to migrate) it worked in one go. I can separate them out just in case. But out of curiosity: what issues did we encounter in the past with combined table and data modifications?

jmsmkn: From what I recall the database would refuse to apply it at runtime. This would have been >7 years ago so my memory is hazy. I think it is best to separate them just in case it is a difference in configuration between development and prod, the Django docs recommend keeping them separate:

> Migrations that alter data are usually called “data migrations”; they’re best written as separate migrations, sitting alongside your schema migrations.

https://docs.djangoproject.com/en/5.1/topics/migrations/#data-migrations

app/grandchallenge/algorithms/models.py
amickan: Ah ok, I added the constraint based on the documentation (see the first info box [here](https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.ManyToManyField.through)), but maybe I misunderstood that comment? Testing it out in the shell, it looks like the add and remove methods are still idempotent even when using a custom through table.... 

In any case, though, the idempotency only holds for the `add` and `remove` methods. You can still create duplicates by directly using the through table to create them (if you don't have a constraint in place) `AlgorithmInterfaceThroughTable.objects.create(interface=..., algorithm-alg)`. So I would say it's safer to add the constraint, right?

amickan: That also leads me to a secondary question: is it then preferred to add interfaces to algorithm using the `add` method (with a `through_defaults` specified to populate the `is_default` field), rather than doing what I do now in the data migration: `AlgorithmInterfaceThroughTable.objects.create(algorithm=algorithm, interface=io, is_default=True)`?

jmsmkn: > So I would say it's safer to add the constraint, right?

Yes definitely, I didn't notice that box. We should add them for all existing through tables.

> Is it then preferred to add interfaces to algorithm using the add method

I would say so, for normal m2m fields you do not have access to the through table, so here you are making the assumption that the class is the correct through table (which is pretty safe but you never know!). I think it is best to be consistent. Generally, we shouldn't have to access through tables directly.

jmsmkn: One more question: is the naming of the new through tables consistent with the existing ones in the codebase?

app/grandchallenge/algorithms/models.py
amickan: Good idea. then we can handle setting inputs and outputs in the custom create method, and don't need to repeat the same 3 lines everywhere we create a new interface. I can't use that in the migration, but that's ok.

app/grandchallenge/algorithms/models.py
amickan: > One more question: is the naming of the new through tables consistent with the existing ones in the codebase?

No, the "ThroughTable" part should go if we want to be consistent with the rest of the codebase. I had added that bit to clearly distinguish it from the `AlgorithmInterface` model (and to avoid a name clash). We could call it `InterfaceAlgorithm` (which implies that it's something else to me). What would you suggest for the name? 

The other two that I added today are more straightforward: `AlgorithmInterfaceInput` and `AlgorithmInterfaceOutput`

jmsmkn: `AlgorithmInputInterface` (with `algorithm` and `input_interface` defined) and `AlgorithmOutputInterface` (with `algorithm` and `output_interface` defined) wouldn't clash and would be consistent, right?

Or, maybe this is too confusing but we could pre-empt our renaming to `AlgorithmInputSlot` (with `algorithm` and `input_slot` defined) and `AlgorithmOutputSlot` (with `algorithm` and `output_slot` defined)?

amickan: I think I wasn't clear enough earlier. There are 3 through tables that need names. 

Normally, we follow `Model1Model2` naming convention for through tables. So then for the through table with `algorithm_interface` and `algorithm` defined, we'd end up with either `AlgorithmInterfaceAlgorithm` or `AlgorithmAlgorithmInterface` - both sound terrible to me. Pre-empting our naming change would not affect the name because the `AlgorithmInterface` model would still be called that after the naming change. 

For the other 2 (linking `algorithm_interface` and `input` / `output` respectively), we could already adopt the slot terminology, but it might be a bit confusing if the naming migration is not taken up in the near future... The names I suggested above `AlgorithmInterfaceInput` and `AlgorithmInterfaceOutput` do not clash with anything and would be an option for the time being.

jmsmkn: Okay then it's probably fine.

amickan:
This is ready for another review.

app/grandchallenge/algorithms/models.py
```diff
@@ -69,6 +69,76 @@
...
+    def create(
+        self,
+        *,
+        inputs,
+        outputs,
+        **kwargs,
+    ):
+        obj = super().create(**kwargs)
+        obj.inputs.set(inputs)
+        obj.outputs.set(outputs)
```
jmsmkn, line 83: This is not idempotent right? This should implement what you have in the data migration I think:

```suggestion
        obj = get_existing_interface_for_inputs_and_outputs(
            model=AlgorithmInterface, inputs=inputs, outputs=outputs
        )
        if not obj:
            obj = super().create(**kwargs)
            obj.inputs.set(inputs)
            obj.outputs.set(outputs)
```

jmsmkn (approved):
Rest LGTM!

app/grandchallenge/algorithms/models.py
amickan: I'm not sure I understand. This is what we do in the migration as well. Calling this method twice with the same inputs and outputs would only result in 1 interface because of the call to `get_existing_interface_for_inputs_and_outputs`, so it is idempotent. What am I missing?

jmsmkn: In the implementation that is on the branch right now doing:

```python
inputs = [...]
outputs = [...]
AlgorithmInterface.objects.create(inputs=inputs, outputs=outputs)
AlgorithmInterface.objects.create(inputs=inputs, outputs=outputs)
```

will create two `AlgorithmInterface`s with the same inputs and outputs, right? I think that the suggestion will mean that only one is created and re-used on subsequent calls.

amickan: Oh wow, I read your suggestion as my code. I think I need another coffee. Yes, you are right, of course. Sorry!

jmsmkn: Haha, it's okay, roll on the weekend!

jmsmkn: I don't know if passing the model to `get_existing_interface_for_inputs_and_outputs` here is right as that hardcodes it. You might be able to do `self.model` instead.

amickan: You don't actually need it because it defaults to `AlgorithmInterface`. I just added the option to pass the model so that we can use this function in the migration where we need to work with historical models. 

Or is setting a default in `get_existing_interface_for_inputs_and_outputs` not a good idea ?

app/grandchallenge/algorithms/models.py
```diff
@@ -69,6 +69,80 @@
...
+
+    def create(
+        self,
+        *,
+        inputs,
+        outputs,
+        **kwargs,
+    ):
+        obj = get_existing_interface_for_inputs_and_outputs(
+            model=AlgorithmInterface, inputs=inputs, outputs=outputs
```
jmsmkn, line 82:
```suggestion
            inputs=inputs, outputs=outputs
```

PR #3722, opened 2024-11-28
Include hanging protocols check in view_content_example generation

### LINKED ISSUE #3721: Reader study with invalid view_content can't be edited
**Describe the bug**
Apparently I have a reader study with an invalid view_content. I can't edit the reader study, as I get a 
ServerError.
From the sentry log:
`RuntimeError: view_content example is not valid.` and
`ValidationError: Image ports in view_content do not match those in the selected hanging protocol.`

**To Reproduce**
1. Go to https://grand-challenge.org/reader-studies/reader-study-demo-2021-pathology
2. Click on 'Update Settings'
3. See error

**Expected behavior**
I should be able to correct the issue

**Screenshots**

**Additional context**

ammar257ammar:
closes #3721 

Basically, example generation should create example with viewports equal to the viewports in the hanging protocol json (if the hanging protocol is defined). Otherwise, the viewports can be up to the number of values in ViewportNames

### DIFF
--- a/app/grandchallenge/hanging_protocols/forms.py
+++ b/app/grandchallenge/hanging_protocols/forms.py
@@ -187,7 +187,9 @@ class ViewContentExampleMixin:
     def _get_viewports(self):
         if self.instance.hanging_protocol:
             return {
-                x["viewport_name"] for x in self.instance.hanging_protocol.json
+                x["viewport_name"]
+                for x in self.instance.hanging_protocol.json
+                if "viewport_name" in x
             }
         return ViewportNames.values
 
--- a/app/grandchallenge/hanging_protocols/forms.py
+++ b/app/grandchallenge/hanging_protocols/forms.py
@@ -206,12 +206,27 @@ class ViewContentExampleMixin:
     ):
         view_content_example = {}
 
-        if images and self.instance.hanging_protocol:
+        if self.instance.hanging_protocol:
             viewport_count = len(viewports)
-            images = (
-                images * (viewport_count // len(images))
-                + images[: (viewport_count % len(images))]
-            )
+
+            if mandatory_isolation_interfaces:
+                mandatory_isolation_interfaces = (
+                    mandatory_isolation_interfaces
+                    * (viewport_count // len(mandatory_isolation_interfaces))
+                    + mandatory_isolation_interfaces[
+                        : (
+                            viewport_count
+                            % len(mandatory_isolation_interfaces)
+                        )
+                    ]
+                )
+                viewport_count -= len(mandatory_isolation_interfaces)
+
+            if images:
+                images = (
+                    images * (viewport_count // len(images))
+                    + images[: (viewport_count % len(images))]
+                )
 
         overlays_per_image = len(overlays) // len(images) if images else 0
         remaining_overlays = len(overlays) % len(images) if images else 0
--- a/app/grandchallenge/hanging_protocols/forms.py
+++ b/app/grandchallenge/hanging_protocols/forms.py
@@ -160,7 +160,7 @@ class ViewContentExampleMixin:
             reverse("documentation:detail", args=["viewer-content"]),
         )
 
-    def generate_view_content_example(self):
+    def _get_interface_lists(self):
         images = [
             interface.slug
             for interface in self.instance.interfaces
@@ -172,10 +172,6 @@ class ViewContentExampleMixin:
             if interface.kind
             in InterfaceKind.interface_type_mandatory_isolation()
         ]
-
-        if not images and not mandatory_isolation_interfaces:
-            return None
-
         overlays = [
             interface.slug
             for interface in self.instance.interfaces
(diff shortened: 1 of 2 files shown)


app/grandchallenge/hanging_protocols/forms.py
```diff
@@ -160,7 +160,7 @@ def __init__(self, *args, **kwargs):
             reverse("documentation:detail", args=["viewer-content"]),
         )
 
-    def generate_view_content_example(self):
```
ammar257ammar, line 163: The function became too long and triggered flake C901 error 
'ViewContentExampleMixin.generate_view_content_example' is too complex
So, I split it into smaller functions with specific tasks.

app/tests/hanging_protocols_tests/test_forms.py
```diff
@@ -666,3 +669,65 @@ def test_generate_view_content_example(
     )
 
     assert view_content_example_json == expected_example_json
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize(
+    "hanging_protocol_json",
```
ammar257ammar, line 676: The test function creates a civ_list with 4 component interfaces that can be set in a view port (images+isolated). The parameterized test covers the following cases:
- No viewports in the hanging protocol json
- The hanging protocol viewports are less than the number of CIs (images+isolated)
- The hanging protocol viewports are equal to the number of CIs (images+isolated) 
- The hanging protocol viewports are larger than the number of CIs (images+isolated)

ammar257ammar: outdated

app/tests/hanging_protocols_tests/test_forms.py
```diff
@@ -666,3 +669,78 @@ def test_generate_view_content_example(
     )
 
     assert view_content_example_json == expected_example_json
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize(
+    "hanging_protocol_json,view_content_example_json_expected_keys",
```
ammar257ammar, line 676: The test function creates a civ_list with 4 component interfaces that can be set in a view port (images+isolated). The parameterized test covers the following cases:

- No hanging protocol is provided (None)
- No viewports in the hanging protocol json
- The hanging protocol viewports are less than the number of CIs (images+isolated)
- The hanging protocol viewports are equal to the number of CIs (images+isolated)

Regarding this case:
- The hanging protocol viewports are larger than the number of CIs (images+isolated)
This will never pass since the validation condition is that the keys of the view_content example json are the same as the viewports of the hanging protocol json.
When the hanging protocol viewports are larger than the number of CIs, the generated example will not have the same keys and will always raise RuntimeError.
What do you suggest for this case? we can't add more CIs to the example if none are there, so I am not sure about covering this case.

Also, regarding the case:
- No viewports in the hanging protocol json
I found a bug where the user can provide the following value for the hanging protocol json "[{}]" and it will pass the validation but it does not have the key "viewport_name" in the json and that will lead to an exception.
This issue is not related to the content_view example, and I will approach it later once I am done with my pitch.

app/grandchallenge/hanging_protocols/forms.py
```diff
@@ -186,14 +182,25 @@ def generate_view_content_example(self):
                 InterfaceKindChoices.IMAGE,
             )
         ]
-
-        if images:
-            overlays_per_image = len(overlays) // len(images)
-            remaining_overlays = len(overlays) % len(images)
-
+        return images, mandatory_isolation_interfaces, overlays
```
amickan, line 185: Please use a NamedTuple when returning multiple values from a function.

app/tests/hanging_protocols_tests/test_forms.py
amickan: > * When the hanging protocol viewports are larger than the number of CIs, the generated example will not have the same keys and will always raise RuntimeError.

Hm, I guess the only option in that case is to return None as example then, right? @miriam-groeneveld, could you comment?

amickan: Or is it possible to add an empty viewport to the `view_content`?

miriam-groeneveld: What about re-using the CI's? I think that's more helpful. For example image A with overlay X on main, image A with overlay Y on secondary would be an example for this use-case.

amickan: Ah of course, that makes a lot more sense! Thanks, Miriam!

app/grandchallenge/hanging_protocols/forms.py

app/grandchallenge/hanging_protocols/forms.py
```diff
@@ -160,7 +161,12 @@ def __init__(self, *args, **kwargs):
             reverse("documentation:detail", args=["viewer-content"]),
         )
 
-    def generate_view_content_example(self):
+    def _get_interface_lists(self):
+        interface_lists = namedtuple(
+            "interface_lists",
+            ["images", "mandatory_isolation_interfaces", "overlays"],
+        )
```
amickan, line 168: Is there a reason you used `collections.namedtuple` instead of `typing.NamedTuple` like we do in the majority of places in the codebase?

ammar257ammar: I changed it to `typing.NamedTuple`. It is safer to do it with type checking. 
Done

amickan (approved):

PR #3725, opened 2024-12-02
Job requirement options from organizations

koopmant:
Selectable gpu type and maximum memory setting for algorithms can now be adjusted with Organization settings. 

In addition, the tests for the options from Phase settings have been adjusted to better cover the complex query. 

See https://github.com/DIAGNijmegen/rse-roadmap/issues/346


app/grandchallenge/evaluation/migrations/0069_phase_maximum_settable_memory_gb_evaluation_and_more.py
```diff
@@ -24,7 +24,7 @@ class Migration(migrations.Migration):
             model_name="phase",
             name="evaluation_selectable_gpu_type_choices",
             field=models.JSONField(
-                default=grandchallenge.evaluation.models.get_default_gpu_type_choices,
+                default=grandchallenge.components.utils.get_default_gpu_type_choices,
```
koopmant, line 27: When I run `makemigrations`, the move of this function is not detected. 

According to the [Django docs](https://docs.djangoproject.com/en/5.1/topics/migrations/#:~:text=References%20to%20functions%20in%20field%20options%20such%20as%20upload_to%20and%20limit_choices_to%20and%20model%20manager%20declarations%20with%20managers%20having%20use_in_migrations%20%3D%20True%20are%20serialized%20in%20migrations%2C%20so%20the%20functions%20and%20classes%20will%20need%20to%20be%20kept%20around%20for%20as%20long%20as%20there%20is%20a%20migration%20referencing%20them.): 
> References to functions in field options [...] are serialized in migrations, so the functions and classes will need to be kept around for as long as there is a migration referencing them. 

I therefore edited this manually. However, I think this might lead to some issues?

app/grandchallenge/evaluation/migrations/0070_phase_algorithm_maximum_settable_memory_gb_and_more.py
```diff
@@ -27,7 +27,7 @@ class Migration(migrations.Migration):
             model_name="phase",
             name="algorithm_selectable_gpu_type_choices",
             field=models.JSONField(
-                default=grandchallenge.evaluation.models.get_default_gpu_type_choices,
+                default=grandchallenge.components.utils.get_default_gpu_type_choices,
```
koopmant, line 30: Same as comment for migration 0069. 
These two are the only old migrations where this function is referenced.

app/tests/evaluation_tests/test_forms.py
```diff
@@ -127,7 +125,7 @@ def test_algorithm_queryset(self):
 
         assert alg1 in form.fields["algorithm"].queryset
         assert alg2 not in form.fields["algorithm"].queryset
-        assert alg2 not in form.fields["algorithm"].queryset
+        assert alg3 not in form.fields["algorithm"].queryset
```
koopmant, line 128: Not relevant for this PR, but I noticed this typo in this test.

jmsmkn: Thanks for highlighting it!

app/grandchallenge/components/utils/__init__.py
jmsmkn, line 1: It is best not to use "utils", this can become a collection of unorganised stuff. I think that it can go into schemas, `SELECTABLE_GPU_TYPES_SCHEMA` definitely belongs there, `GPUTypeChoices` probably not but it is required there so that would maybe go into it's own file, but that's not worth it.

jmsmkn (commented):
Rest looks good!

app/grandchallenge/components/utils/__init__.py
koopmant: Alright, moved it there.

jmsmkn (approved):

PR #3727, opened 2024-12-03
Fix Hanging Protocols json field raised exception

### LINKED ISSUE #3724: Hanging Protocols JSON field fails to validate properly and an exception is raised
**Describe the bug**

When creating a new hanging protocol, any value that is an array with at least an object structure inside of it will raise an exception without going through proper validation, even though a schema check is in place.
The bug was discovered while working on https://github.com/DIAGNijmegen/rse-roadmap/issues/362

**To Reproduce**
1. Go to Hanging Protocol Create form
2. Enter a title and a value in the json field `[{}] `or `[{"test":1}]` (both of them are invalid according to the schema)
3. Click save

**Expected behavior**
A redisplayed form with validation error on the json field, as in the screenshot below.

Screenshot 2024-12-02 at 15 54 31

**Screenshots**
However, I get the following error

Screenshot 2024-12-02 at 15 56 08

 - OS: [ iOS]
 - Browser [e.g. chrome]

ammar257ammar:
closes #3724

### DIFF
--- a/app/grandchallenge/hanging_protocols/forms.py
+++ b/app/grandchallenge/hanging_protocols/forms.py
@@ -26,19 +26,19 @@ from grandchallenge.subdomains.utils import reverse
 
 
 class HangingProtocolForm(SaveFormInitMixin, forms.ModelForm):
-    json = forms.JSONField(
-        validators=[JSONValidator(schema=HANGING_PROTOCOL_SCHEMA)],
-        widget=JSONEditorWidget(schema=HANGING_PROTOCOL_SCHEMA),
-        help_text="To display a single image in full size, define the "
-        "protocol as follows: "
-        '[{"viewport_name": "main", "x": 0,"y": 0,"w": 1,"h": 1,'
-        '"fullsizable": true,"draggable": false,"selectable": true,'
-        '"order": 0}]',
-    )
-
     class Meta:
         model = HangingProtocol
         fields = ("title", "description", "json")
+        widgets = {"json": JSONEditorWidget(schema=HANGING_PROTOCOL_SCHEMA)}
+        help_texts = {
+            "json": (
+                "To display a single image in full size, define the "
+                "protocol as follows: "
+                '[{"viewport_name": "main", "x": 0,"y": 0,"w": 1,"h": 1,'
+                '"fullsizable": true,"draggable": false,"selectable": true,'
+                '"order": 0}]'
+            )
+        }
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -59,8 +59,14 @@ class HangingProtocolForm(SaveFormInitMixin, forms.ModelForm):
 
     def clean_json(self):
         hanging_protocol_json = self.cleaned_data["json"]
+
+        if not isinstance(hanging_protocol_json, list):
+            return hanging_protocol_json
+
         viewport_names = [
-            viewport.get("viewport_name") for viewport in hanging_protocol_json
+            viewport["viewport_name"]
+            for viewport in hanging_protocol_json
+            if "viewport_name" in viewport
         ]
 
         self._validate_viewport_uniqueness(viewport_names=viewport_names)
--- a/app/grandchallenge/hanging_protocols/forms.py
+++ b/app/grandchallenge/hanging_protocols/forms.py
@@ -60,17 +60,14 @@ class HangingProtocolForm(SaveFormInitMixin, forms.ModelForm):
     def clean_json(self):
         hanging_protocol_json = self.cleaned_data["json"]
 
-        if not isinstance(hanging_protocol_json, list):
-            return hanging_protocol_json
-
-        viewport_names = [
-            viewport["viewport_name"]
-            for viewport in hanging_protocol_json
-            if isinstance(viewport, dict) and "viewport_name" in viewport
-        ]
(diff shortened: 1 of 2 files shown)


app/grandchallenge/hanging_protocols/forms.py
```diff
@@ -59,7 +59,9 @@ def __init__(self, *args, **kwargs):
 
     def clean_json(self):
         json = self.cleaned_data["json"]
-        viewport_names = [x["viewport_name"] for x in json]
+        viewport_names = [
+            x["viewport_name"] for x in json if "viewport_name" in x
```
jmsmkn, line 63: `json` should not be used as a variable name as it shadows the built in module which is imported on L1. `x` is not a very clear variable name, `viewport` might be a better one but I am not sure what this "json" contains (that also could be better named). If a key only possibly exists in a dictionary you can use `foo.get("viewport_name")` which will return `None` by default. See https://realpython.com/python-lbyl-vs-eafp/. If an error found in production is being fixed a test should be included to ensure that it does not happen again.

amickan:
I think some of the other validation functions have the same issue and need to be fixed as well. 

Also: what if someone enters `"test"` (or anything that's not an array of dictionaries), I think the `get()` would fail, right?

ammar257ammar:
> I think some of the other validation functions have the same issue and need to be fixed as well.
> 
> Also: what if someone enters `"test"` (or anything that's not an array of dictionaries), I think the `get()` would fail, right?

Entering strings or numbers is not allowed by the JSON widget (can't submit with such a value), but maybe other data structures are possible. I will check other possibilities.

amickan:
> Entering strings or numbers is not allowed by the JSON widget (can't submit with such a value)

I just tried and I'm able to submit.

koopmant:
See https://github.com/comic/grand-challenge.org/issues/3724#issuecomment-2528293670 for an alternative solution.

ammar257ammar:
> See [#3724 (comment)](https://github.com/comic/grand-challenge.org/issues/3724#issuecomment-2528293670) for an alternative solution.

Thanks a lot Thomas! I will try that alternative.

jmsmkn:
> > > Entering strings or numbers is not allowed by the JSON widget (can't submit with such a value)
> > 
> > 
> > I just tried and I'm able to submit.
> 
> Oh, to correct for reference, with strings I can't submit (screenshot below), but with numbers it is possible. Also, with dictionary structure that is not in an array it is possible. All cases should be handled on the server side. working on it.


It does not matter what the client says server side as that can always be bypassed, it has to be handled server side.

ammar257ammar:
I tried @koopmant suggestion mentioned above (applied in commit https://github.com/comic/grand-challenge.org/pull/3727/commits/579a9bdf5310f254dda74fb372533e832c989735) and it works neatly where the schema validation is performed before the field cleaning methods. 

I still have a question @jmsmkn. Is it recommended in this case to not update further the clean field function (clean_json) to handle all possible cases of json field inputs?

Also, do you recommend, in general, defining a form field with the same validators as the model property in cases where there is a clean field method defined, to make sure the validation run before the cleaning and avoid exceptions in the clean methods? (as a best practice to follow from now on)

jmsmkn:
> Is it recommended in this case to not update further the clean field function (clean_json) to handle possible cases of invalid json field inputs?

If you want to perform validation on a form field then that is where it needs to be implemented.

> Also, do you recommend, in general, defining a form field with the same validators as the model property in cases where there is a clean field method defined, to make sure the validation run before the cleaning and avoid exceptions in the clean methods?

This duplicates code in multiple places, so I do not think that this is a good workaround as developers will need to remember to update both. An alternative solution to the problematic line would have been:

```python
try:
    viewport_names = [
        viewport["viewport_name"] for viewport in hanging_protocol_json
    ]
except (KeyError, TypeError):
    raise ValidationError("Hanging protocol definition is invalid")
```

ammar257ammar:
I ended up doing more checks in the cleaning function rather that raising validation error as suggested earlier. That exception would be raised in many situations, some of them need a more specific validation message. 
For example, the value `[{"test":"main"}] `would raise an exception with message `"Hanging protocol definition is invalid"` and that is not informative enough, while the validation should display `"JSON does not fulfill schema: instance 'viewport_name' is a required property"`.


Another possible solution is to apply JSONValidator in the function before selecting the viewport_names and that will raise validation errors with appropriate messages. An implementation would be: 

```python
def clean_json(self):
    hanging_protocol_json = self.cleaned_data["json"]

    JSONValidator(schema=HANGING_PROTOCOL_SCHEMA)(value=hanging_protocol_json)

    viewport_names = [
        viewport["viewport_name"]
        for viewport in hanging_protocol_json
    ]
    .....
```

amickan:
> For example, the value `[{"test":"main"}] `would raise an exception with message `"Hanging protocol definition is invalid"` and that is not informative enough, while the validation message would better be `"JSON does not fulfill schema: instance 'viewport_name' is a required property"`.

Why not do what james suggested but make the error more informative? 

```python

try:
    viewport_names = [
        viewport["viewport_name"] for viewport in hanging_protocol_json
    ]
except (KeyError, TypeError):
    raise ValidationError("JSON does not fulfill schema: instance 'viewport_name' is a required property")

```

amickan (approved):

PR #3728, opened 2024-12-03
Updates to `JSONEditorWidget`

### LINKED ISSUE #3698: Switching view of JSON editor fails with empty content
**Describe the bug**

When on the try-out page of an Algorithm, the JSON editor can crash. That leads to a partially broken form where you do not have feedback on what might be missing. The form does still work if you re-fill all the values.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to https://grand-challenge.org/algorithms/pi-cai-pubpriv-pimed/jobs/create/
2. In the JSON editor, swap to code mode
3. Remove the null content
4. Switch back to tree
5. Try to save.

**Expected behaviour**

To have several 'is required' validation errors.

**Screenshots**

image

**Additional context**

image

chrisvanrun:
This PR has a few changes bundled with respect to the JS-dep JSONEditorWidget.

Minor: it sets the default mode to "code". I despise the tree mode and always switch out immediately for the code mode. I am sure I am not the only one.

## Amass

- Update `amass.lock`, this was outdated with respect to the `pyproject.toml`, a lot of no-longer-defined dependencies are present
- Update JSONEditor to the latest version (`10.1.1`)

## JsonEditor invocation

Fixes #3698

The crux of the original issue lies with `setCustomValidty` on a `d-none` classed `<textarea>`. When saving, it has the client-side validity crash since it cannot focus on the element.

To solve this the validity error is now set on the textarea of the ace editor:

image

Another (hidden) error would occur when trying to switch from an empty "code" mode to "tree" mode. With the latest version of JSONEditor this now shows as an informative (albeit ugly) prompt:
image


chrisvanrun:
image
Wowa, going to set to draft to address that.

chrisvanrun:
> It could be right - the changes are mostly to vendored files. Did you use `amass install` to update the files?

Yes I did, had to jump into an onboarding meeting so decided to quickly 'park' this PR.

chrisvanrun:
James is correct, the number of changes is 99% from the updated JS files for JSONEditor. I used `amass install` to install them. Albeit I had to kick it a few times because it ran into rate-limiting when running it too often, hitting the sri-is-not-equal error while it likely got an HTML page back telling me to stop requesting stuff so often.

app/grandchallenge/core/static/js/jsoneditor_widget.mjs
```diff
@@ -11,26 +11,41 @@ function initialize_jsoneditor_widget(jsoneditorWidgetID) {
         jsoneditorWidget.classList.remove("d-none");
     } else {
         const options = {
-            mode: "tree",
+            mode: "code",
```
jmsmkn, line 14: Lets keep it defaulting to tree mode right now. You are a developer so might switch straight away, but the users are not and are more often just looking at the values of the field. The tree mode offers a nicer interface for that.

```suggestion
            mode: "tree",
```

jmsmkn (approved):

PR #3729, opened 2024-12-04
Replace responsive DataTables behaviour with floating scrolling bar

### LINKED ISSUE #3028: Users confused by responsive wide table information hiding
It's behind the blue cross. Both leaderboard and admin panels. Maybe limit wide tables to 10 rows and then use just the bootstrap responsive. Also allow word breaks?

chrisvanrun:
Fixes #3028

This PR adds a new JS dependency, `floating-scroll`. This tiny JS package allows a floating scrollbar at the bottom of a viewport instead of the bottom of the element (which is possibly hidden way below).

Floating-scroll is applied after the Datatables are initiated. This is done via a generic listener to `init.dt` on the document level: in the same breath, it sets the `scrollX: true` to allow the inner element to overflow and spawn a scrollbar in the first place.

Datatables come in two flavours:
- Dynamically, initialized via the partial template `datatable.html` (and some Python, JS et cetera)
- Statically, by initializing it on a `<table>` directly

The new overflow behaviour applies to both. _For the static route_ this entails the scrollbar is properly under the table now instead of under the paging UI. The same holds for the _dynamic route_ but this also has the `responsive` component explicitly removed.

### Current
 
https://github.com/user-attachments/assets/1479346b-69a2-4e51-903a-73e48a87ea7c

https://github.com/user-attachments/assets/393d12f0-1513-4800-977e-e70802eaddfb

### New

https://github.com/user-attachments/assets/189a9cdd-f42d-4d39-9328-2fb94999d465

https://github.com/user-attachments/assets/393a8f48-68d8-4e28-bdc8-581917f0c3c4

### Side note

Technically, the bootstrap class `"table-responsive"` is now redundant: the element containing the datatable should never overflow—unless perhaps there are a gazillion pages(?). However, since it's generally applied as a wrapping element, removing it here would result in an unneedingly bloated changeset.

I first had handy-scroll, but that has a component-only version 2 now... which does not work with dynamic datatables.


app/grandchallenge/datatables/static/js/datatables/list.mjs
```diff
@@ -53,12 +40,3 @@ $(document).ready(() => {
         },
     });
 });
-
-$("#ajaxDataTable").on("init.dt", () => {
-    // This is a work-around to get the table to resize properly on extra-large Bootstrap viewport
-    setTimeout($("#ajaxDataTable").DataTable().columns.adjust, 1000);
-});
-
-$(window).resize(() => {
```
chrisvanrun, line 62: No need to account for resizing anymore: this was to help the datatable responsive components pack in more of the columns in the dropdown.

app/grandchallenge/core/static/js/datatables.defaults.mjs
```diff
@@ -0,0 +1,11 @@
+$.extend(true, DataTable.defaults, {
+    scrollX: true,
+});
+
+$(document).on("init.dt", event => {
+    // This is a work-around to get the table to resize properly on extra-large Bootstrap viewport
+    setTimeout($(event.target).DataTable().columns.adjust, 1000);
+
+    // // Set up floating scroll, note that the target class only shows up when scrollX is set to true
```
jmsmkn, line 9:
```suggestion
    // Set up floating scroll, note that the target class only shows up when scrollX is set to true
```

app/grandchallenge/core/static/js/datatables.defaults.mjs
```diff
@@ -0,0 +1,11 @@
+$.extend(true, DataTable.defaults, {
+    scrollX: true,
+});
+
+$(document).on("init.dt", event => {
+    // This is a work-around to get the table to resize properly on extra-large Bootstrap viewport
+    setTimeout($(event.target).DataTable().columns.adjust, 1000);
```
jmsmkn, line 7: There must be a better way. Why does this occur? Is it caused by an interaction with `table-responsive`? Does the event need to happen on init or once the table is ready? Is there another signal this could be attached to?

app/grandchallenge/core/templates/grandchallenge/partials/style.html
```diff
@@ -10,6 +10,8 @@
     <link href="{% static 'vendored/datatables.net-bs4/css/dataTables.bootstrap4.min.css' %}" rel="stylesheet"/>
     <link href="{% static 'vendored/datatables.net-buttons-bs4/css/buttons.bootstrap4.min.css' %}" rel="stylesheet"/>
     <link href="{% static 'vendored/datatables.net-responsive-bs4/css/responsive.bootstrap4.min.css' %}" rel="stylesheet"/>
+
```
jmsmkn, line 13: Do we still need `responsive.bootstrap4.min.css`?

app/grandchallenge/core/templates/grandchallenge/partials/script.html
```diff
@@ -15,5 +19,8 @@
     <script src="{% static 'vendored/datatables.net-responsive-bs4/js/responsive.bootstrap4.min.js' %}"></script>
 {% endcompress %}
 
+{# Set global defaults for datatables #}
```
jmsmkn, line 22: Do we still need `responsive.bootstrap4.min.js`?

chrisvanrun:
[WIP] I did a search on the `.DataTable(` init call and I am checked (nearly) every thus initialized table to see if it would load and behave correctly with respect to overflow.

Skipped:
- statistics table, 
-verification-userset table

pyproject.toml
```diff
@@ -140,6 +140,7 @@ jquery = { version = "==3.7.0", include = ["jquery(.min)?(.js|.map)?$"] }
 font-awesome = { version = "==5.15.4", include = ["css/all(.min)?.css", "webfonts/*"] }
 bootstrap = { version = "==4.6.2", include = ["js/bootstrap.bundle(.min)?.js(.map)?", "css/bootstrap(.min)?.css(.map)?", "scss/*"] }
 bootswatch = { version = "==5.3.0", provider = "unpkg", include = ["dist/flatly/.*.scss"] }
+floating-scroll = { version = "==3.2.0", provider = "unpkg", include = ["dist/jquery.floatingscroll.min.js", "dist/jquery.floatingscroll.css"] }
 "@github/markdown-toolbar-element" = { version = "==1.2.0", provider = "unpkg", include = ["dist/index.umd.js$"] }
 # Use the builder at https://datatables.net/download/
 "datatables.net" = { version = "==1.11.3", provider = "unpkg", include = ["js/jquery.dataTables(.min)?.js"]}
```
jmsmkn, line 146: Do we still need `datatables.net-responsive` and `datatables.net-responsive-bs4`?

app/grandchallenge/core/static/js/datatables.defaults.mjs
chrisvanrun: This was a leftover I just grandfathered in. Quite possibly it's not needed anymore and was only applicable for the responsive dt. Not even sure how to reproduce it.

jmsmkn:
It probably is a good idea to replace the class `table-responsive` with `table`, but only for tables where this feature applies. We probably have other non-datatables tables that can take advantage of `table-responsive`.

app/grandchallenge/core/static/js/datatables.defaults.mjs
jmsmkn: It was added in #2160 - there is a video in that. If the behaviour can be replicated now without this line then it can be removed. Unless @amickan can remember exactly why it was added?

amickan: I think it had to do with inline images (thumbnails), and them loading slowly and the resize event firing before the image was fully loaded. It was a race condition that only occurred sometimes. I also see that I added a max width and max height to the thumbnails right after the linked PR, which might already solve this, unsure.

chrisvanrun: Ah. Ok. I think it is no longer applicable since the responsiveness is gone anyway and all results are now shown with a badge + modal.

chrisvanrun:
Checklist, all tables that have `.DataTable(` applied to them:
- [ ] http://gc.localhost:8000/settings/api-tokens/

- [ ] http://gc.localhost:8000/challenges/costs/

- [ ] http://gc.localhost:8000/challenges/requests/

- [ ] http://gc.localhost:8000/components/interfaces/inputs/

- [ ] http://gc.localhost:8000/components/interfaces/outputs/

- [ ] http://demo.gc.localhost:8000/participants/registration/

- [ ] http://gc.localhost:8000/algorithms/test-algorithm/jobs/

- [ ] http://demo.gc.localhost:8000/evaluation/combined-leaderboards/overall/

- [ ] http://demo.gc.localhost:8000/evaluation/phase-1/

- [ ] http://demo.gc.localhost:8000/evaluation/phase-1/ground-truths/

- [ ] http://demo.gc.localhost:8000/evaluation/phase-1/leaderboard/

- [ ] http://demo.gc.localhost:8000/statistics/

- [ ] http://demo.gc.localhost:8000/participants/

- [ ] http://demo.gc.localhost:8000/participants/registration/

- [ ] http://demo.gc.localhost:8000/participants/registration/questions/

- [ ] http://demo.gc.localhost:8000/pages/

- [ ] http://gc.localhost:8000/reader-studies/reader-study/statistics/

- [ ] http://gc.localhost:8000/verifications/user-sets/1/

pyproject.toml
chrisvanrun: I had not realized it was actually an extension instead of a build-in. Removed!

chrisvanrun:
All codebase occurrences that instantiate Datatables to a `<table>` have been checked and seem to render OK. 

I also noticed that there is a DataTables v2, a quick install suggested that it seems like it is a near drop-in replacement. Might explore it in the cooldown week.

chrisvanrun:
Jeej: 
image

jmsmkn (approved):
Please also check this on a diverse range of tables in production once deployed.

chrisvanrun:
Production check, checking same tables as dev:


- [ ] http://gc.localhost:8000/settings/api-tokens/

- [ ] http://gc.localhost:8000/challenges/costs/

- [ ] http://gc.localhost:8000/challenges/requests/

- [ ] http://gc.localhost:8000/components/interfaces/inputs/

- [ ] http://gc.localhost:8000/components/interfaces/outputs/

- [ ] http://demo.gc.localhost:8000/participants/registration/

- [ ] http://gc.localhost:8000/algorithms/test-algorithm/jobs/

- [ ] http://demo.gc.localhost:8000/evaluation/combined-leaderboards/overall/

- [ ] http://demo.gc.localhost:8000/evaluation/phase-1/

- [ ] http://demo.gc.localhost:8000/evaluation/phase-1/ground-truths/

- [ ] http://demo.gc.localhost:8000/evaluation/phase-1/leaderboard/

- [ ] http://demo.gc.localhost:8000/statistics/

- [ ] http://demo.gc.localhost:8000/participants/

- [ ] http://demo.gc.localhost:8000/participants/registration/

- [ ] http://demo.gc.localhost:8000/participants/registration/questions/

- [ ] http://demo.gc.localhost:8000/pages/

- [ ] http://gc.localhost:8000/reader-studies/reader-study/statistics/

- [ ] http://gc.localhost:8000/verifications/user-sets/1/

PR #3731, opened 2024-12-05
Update budget based on gpu type and memory

koopmant:
GPU type choices and maximum memory for algorithm inference jobs are now included in challenge requests. This information is used in the budget calculation and also passed down to the challenge phases when they are converted to Algorithm submission phases. 


app/grandchallenge/challenges/models.py
```diff
@@ -906,6 +911,22 @@ class ChallengeRequestStatusChoices(models.TextChoices):
             MaxValueValidator(limit_value=60),
```
koopmant, line 911: @amickan @jmsmkn Are `blank=True` and `null=True` correct for this field? On the form we set the related field to required, so I'm a bit confused why blank is allowed here.

jmsmkn: I do not think that is correct. I think that is the same for all of these fields:

- inference_time_limit_in_minutes
- average_size_of_test_image_in_mb
- phase_1_number_of_submissions_per_team
- phase_2_number_of_submissions_per_team
- phase_1_number_of_test_images
- phase_2_number_of_test_images
- budget_for_hosting_challenge
- long_term_commitment (boolean, so unsure)
- data_license (boolean, so unsure)
- algorithm_inputs
- algorithm_outputs
- challenge_fee_agreement (boolean, so unsure)

Maybe Anne can confirm?

amickan: They should indeed all be required. Back when we first implemented this form, users could choose between Type 1 and Type 2 challenges and for the former the budget fields were not required. When we removed the type choice, looks like we didn't migrate the schema, but only made the fields required always on the form.

koopmant: Thanks for the clarification! Will update this in this PR.

amickan: The phase 2 number of submissions and number of test images need to allow 0s (see this example: https://grand-challenge.org/documentation/create-your-own-challenge/#example-of-a-challenge-with-more-than-2-phases)

app/grandchallenge/challenges/models.py
```diff
@@ -898,17 +903,31 @@ class ChallengeRequestStatusChoices(models.TextChoices):
...
+    )
+    algorithm_maximum_settable_memory_gb = models.PositiveSmallIntegerField(
+        default=settings.ALGORITHMS_MAX_MEMORY_GB,
+        help_text=(
+            "Maximum amount of memory that participants will be allowed to "
+            "assign to algorithm inference jobs for submission."
+        ),
+    )
     average_size_of_test_image_in_mb = models.PositiveIntegerField(
         null=True,
```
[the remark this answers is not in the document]
jmsmkn: You can use option 2 as we forced users to fill in the field anyway.

app/tests/components_tests/resources/backends.py
```diff
@@ -75,7 +76,12 @@ def duration(self):
 
     @property
     def usd_cents_per_hour(self):
-        return 100
+        return {
+            GPUTypeChoices.NO_GPU: 50,
+            GPUTypeChoices.T4: 100,
+            GPUTypeChoices.A10G: 150,
+            GPUTypeChoices.V100: 400,
+        }.get(self._requires_gpu_type, 100)
```
koopmant, line 84: I excluded A100 and K80 here because I also did not get anything back when I tested these with 'grandchallenge.components.backends.amazon_sagemaker_training.AmazonSageMakerTrainingExecutor'.

app/grandchallenge/challenges/forms.py
```diff
@@ -189,9 +150,23 @@ def clean(self):
...
-):
+class ChallengeRequestForm(forms.ModelForm):
+    algorithm_selectable_gpu_type_choices = forms.MultipleChoiceField(
+        initial=get_default_gpu_type_choices(),
+        choices=[
+            (choice.value, choice.label)
+            for choice in [
+                GPUTypeChoices.NO_GPU,
+                GPUTypeChoices.T4,
+                GPUTypeChoices.A10G,
```
amickan, line 161: Why only these 3?

app/grandchallenge/challenges/forms.py
```diff
@@ -604,21 +574,51 @@ def clean_status(self):
         return status
 
 
-class ChallengeRequestBudgetUpdateForm(
-    ChallengeRequestBudgetFieldValidationMixin, forms.ModelForm
-):
+class ChallengeRequestBudgetUpdateForm(forms.ModelForm):
+    algorithm_selectable_gpu_type_choices = forms.MultipleChoiceField(
+        initial=get_default_gpu_type_choices(),
```
amickan, line 579: Shouldn't this have the current choices pre-selected, rather than the default choices?

app/tests/components_tests/resources/backends.py
jmsmkn: The changes to this file should be reverted, you can change the executor in your test by adding the `settings` fixture and setting `settings.COMPONENTS_DEFAULT_BACKEND = "grandchallenge.components.backends.amazon_sagemaker_training.AmazonSageMakerTrainingExecutor"`.

The missing A100 makes sense as that is only available in an 8 GPU configuration, but the K80 should be possible as this is available as a single GPY instance (`ml.p2.xlarge`).

However, K80s are quite outdated so will not be used by customers and V100s are also overkill, so users should only be able to select None, T4 and A10G in the challenge request form.

app/grandchallenge/challenges/forms.py
jmsmkn: These are the only ones we want to enable right now. K80s - too old, V100s - overkill for the types of challenges we run right now, A100s - overkill and only available as multi-GPU instances.

amickan:
> When the budget update form is loaded, the option is still selected on the form. However, when the update form is saved, the data has the NO_GPU option removed. (`'T4']`)

Sounds like an HTMX issue / bug to me where empty values are not sent along with POST requests, or did you already find the reason / a fix for it?

koopmant:
> > When the budget update form is loaded, the option is still selected on the form. However, when the update form is saved, the data has the NO_GPU option removed. (`'T4']`)
> 
> Sounds like an HTMX issue / bug to me where empty values are not sent along with POST requests, or did you already find the reason / a fix for it?

Feels like it indeed. Haven't found a fix yet.

koopmant:
> > When the budget update form is loaded, the option is still selected on the form. However, when the update form is saved, the data has the NO_GPU option removed. (`'T4']`)
> 
> Sounds like an HTMX issue / bug to me where empty values are not sent along with POST requests, or did you already find the reason / a fix for it?

It indeed appears to be an issue with HTMX. Specifically, it appears HTMX is excluding the empty string checkbox value when other values in the same group exist, treating it as redundant. 
The POST data after a create request includes empty string: 
`algorithm_selectable_gpu_type_choices: `
`algorithm_selectable_gpu_type_choices: T4`
But the POST data after an update request with HTMX does not:
`algorithm_selectable_gpu_type_choices: T4`
Unless it is the only checkbox that is checked:
`algorithm_selectable_gpu_type_choices: `
If have made a workaround for it and it's working. I've changed the "" value in the choices with "no_gpu" and manually swap these during instantiation and in cleanup. I'm happy for suggestions if this can be done in a better way.

amickan:
> If have made a workaround for it and it's working. I've changed the "" value in the choices with "no_gpu" and manually swap these during instantiation and in cleanup. I'm happy for suggestions if this can be done in a better way.

Is there a reason that the `NO_GPU` option in the enum is `""`? If not, I think updating the enum would be the way to go. Then you don't need a special work around.

app/grandchallenge/challenges/models.py
koopmant: Okay, just learned that the PositiveIntegerField allows zero by default. So nothing needed there.

app/tests/challenges_tests/test_models.py
```diff
@@ -120,18 +121,31 @@ def test_total_challenge_cost():
...
-    assert request2.total_challenge_cost == 6000
+        assert request3.storage_and_compute_cost_surplus == 1380
+        assert request3.total_challenge_cost == 7500
 
-    assert request3.storage_and_compute_cost_surplus == 1290
-    assert request3.total_challenge_cost == 7500
+        assert request4.storage_and_compute_cost_surplus == 2580
+        assert request4.total_challenge_cost == 9000
+    finally:
+        settings.COMPONENTS_DEFAULT_BACKEND = original_value
```
jmsmkn, line 148: You do not need to keep the original value and then set it back, the fixture handles rollback automatically. The `try`/`finally` can also go.

app/grandchallenge/challenges/forms.py
```diff
@@ -604,24 +574,58 @@ def clean_status(self):
...
 
 
-class ChallengeRequestBudgetUpdateForm(
-    ChallengeRequestBudgetFieldValidationMixin, forms.ModelForm
-):
+class ChallengeRequestBudgetUpdateForm(forms.ModelForm):
+    algorithm_selectable_gpu_type_choices = forms.MultipleChoiceField(
+        # replace NO_GPU value ("") in choices to avoid HTMX POST issue
+        choices=[
+            ("no_gpu", GPUTypeChoices.NO_GPU.label),
```
jmsmkn, line 581: `"no_gpu"` seems to be repeated in a few places. Maybe create a variable:

```python
HTMX_BLANK_CHOICE_KEY = "__HTMX_BLANK_CHOICE_KEY__"
```

and then use that in this class and the tests.

jmsmkn:
This doesn't implement the setting of these values when phases are converted to Algorithm phases, I think? Is that intentional and coming in another PR?

koopmant:
> This doesn't implement the setting of these values when phases are converted to Algorithm phases, I think? Is that intentional and coming in another PR?

That was intended for another PR indeed. But it's a very small thing. I could add it here if you don't mind.

app/grandchallenge/challenges/forms.py
koopmant: I like that. Thanks!

jmsmkn:
> > This doesn't implement the setting of these values when phases are converted to Algorithm phases, I think? Is that intentional and coming in another PR?
> 
> That was intended for another PR indeed. But it's a very small thing. I could add it here if you don't mind.

Completely up to you.

app/tests/challenges_tests/test_models.py
koopmant: Yes, I was a little confused by this. It should not be necessary, but without it I get a lot of failures in the tests. The current job is failing again.

koopmant:
Ok, I've included setting the values on the phases when they are converted.

app/tests/challenges_tests/test_models.py
koopmant: I see you marked this as resolved. I had reverted these changes. Do you want me to bring them back again?

jmsmkn:
Great, can be merge once the tests are fixed.

app/tests/challenges_tests/test_models.py
jmsmkn: Sorry, you had only pushed the one commit removing the try finally and had closed the other conversation without pushing anything, I thought you had clicked the wrong one. The test failures are strange, that shouldn't happen.

jmsmkn: It is because `from django.conf import settings` is being used in the test. This should never be used in tests, the fixture should be used instead.

jmsmkn: I have pushed a fix, the test still fails with:

```
>       assert request1.storage_and_compute_cost_surplus == -270
E       assert 410 == -270
```

I am not sure what the correct thing to do there is. But now you shouldn't get unrelated errors in other tests.

jmsmkn: It appears to be caused by db9db17, which seems odd.

koopmant: > It is because `from django.conf import settings` is being used in the test. This should never be used in tests, the fixture should be used instead.

*Slaps forehead*
Thank you!

jmsmkn: Ah, it is because the challenge request factory was modified.

jmsmkn (approved):

app/tests/challenges_tests/test_models.py
koopmant: > I have pushed a fix, the test still fails with:
> 
> ```
> >       assert request1.storage_and_compute_cost_surplus == -270
> E       assert 410 == -270
> ```
> 
> I am not sure what the correct thing to do there is. But now you shouldn't get unrelated errors in other tests.

That test failure was related to a later change. Just fixed it.

PR #3735, opened 2024-12-09
Add create and list views for algorithm interfaces

amickan:

This adds a create and list view for algorithm interfaces that is accessible to algorithm editors with the global `add_algorithm` permission. 

Create view: 
image

List view: 
image


app/grandchallenge/algorithms/views.py
```diff
@@ -1091,3 +1095,61 @@ def get(self, *_, **__):
...
+        else:
+            return self.handle_no_permission()
+
+
+class AlgorithmInterfaceForAlgorithmCreate(
+    AlgorithmInterfacePermissionMixin, CreateView
+):
+    model = AlgorithmInterface
+    form_class = AlgorithmInterfaceGetOrCreateForm
+    success_message = "Algorithm interface successfully added"
```
amickan, line 1121: I was unsure which `model` to use for this view. This view either creates or retrieves an existing `AlgorithmInterface` and then adds it to an algorithm, i.e. adds a row to the custom through table `AlgorithmAlgorithmInterface`. So at the end of the day, it creates an entry in the through model, but it feels wrong to have a `CreateView` for a through model, since that is not normally accessible. So I ultimately went with a `CreateView` for the `AlgorithmInterface`, with the name reflecting the side effect that it also adds an entry to the through table. 

Using the `AlgorithmInterface` as base model for the view also interfaces better with the form I'm using. For the form, I have a clear preference for using a `AlgorithmInterface` `ModelForm` (rather than a `ModelForm` for the through model).  Much more straight-forward to implement. 

I also briefly considered making this an `UpdateView` for the `Algorithm` (mainly for the permission check), but I abandoned that pretty quickly.

jmsmkn: My gut feeling is that `AlgorithmInterface` Create is correct here, or maybe a `FormView` with a custom save method on the form would be more accurate, but I don't think that really matters. Although we have access to the through table as we are using a custom one Django always prefers you not to interact with it directly.

amickan: I opted to stick to an `AlgorithmInterface` Create.

app/grandchallenge/algorithms/views.py
```diff
@@ -1091,3 +1095,61 @@ def get(self, *_, **__):
                 filename=f"{algorithm.slug}-template.zip",
                 content_type="application/zip",
             )
+
+
+class AlgorithmInterfacePermissionMixin(
+    VerificationRequiredMixin, AccessMixin
```
jmsmkn, line 1101: This is stacking `UserPassesTestMixin` on top of `AccessMixin` which might lead to trouble. Why is verification required here?

app/grandchallenge/algorithms/forms.py
```diff
@@ -1377,3 +1393,60 @@ def clean(self):
...
+    def save(self):
+        io = (
+            self.cleaned_data["existing_io"]
+            if self.cleaned_data["existing_io"]
+            else super().save()
+        )
+
+        if self.cleaned_data["set_as_default"]:
+            AlgorithmAlgorithmInterface.objects.filter(
+                algorithm=self._algorithm
```
jmsmkn, line 1442: Could this match multiple objects?

app/grandchallenge/algorithms/forms.py
```diff
@@ -1377,3 +1393,60 @@ def clean(self):
...
+                "existing interface on the interface list."
+            )
+
+        return cleaned_data
+
+    def save(self):
+        io = (
+            self.cleaned_data["existing_io"]
+            if self.cleaned_data["existing_io"]
+            else super().save()
```
jmsmkn, line 1437: Is passing around the `existing_io` object necessary? I thought that with the new `create` method the call to `get_existing_interface_for_inputs_and_outputs` would be unnecessary?

app/grandchallenge/algorithms/views.py
amickan: We check for verification on all other algorithm views as well, hence the inclusion here. We do already have a similar stacking of `UserPassesTestMixin` on top of `AccessMixin` in other places (e.g. `MethodCreate` where we use the `VerificationRequredMixin` and `ActiveChallengeRequiredMixin`). My tests suggest this works and is not a problem, but I could also include the verificaiton check in the dispatch method directly to be sure, or we drop the verification requirement, but that would not be in line with the rest of the algorithm create and update logic.

app/grandchallenge/algorithms/forms.py
amickan: That is true, but we're doing the check for the existing interface in the base form (that is used in the admin and in this view). Calling `create` in the base form is not what we want because in the admin, we want to throw an error if there is an existing interface rather than create a new one. So then `existing_io` is already there and it would be weird not to use it, right? I could also get rid of the base form and have two seperate, independent forms for the admin and the user-facing view, but there is also the filter on the CI queryset and the check for duplicate interfaces in the base form, so it made sense to have a common base form. What do you think?

jmsmkn: > in the admin, we want to throw an error if there is an existing interface rather than create a new one. 

Why do we want to do that? I would have thought that the same logic would apply everywhere, that the existing interface is returned if it exists. Do we even need separate admin functionality for this?

I think using the `create` method on the queryset is more aligned with Django, therefore less surprising to future developers. Right now on main we have two forms and a mixin to do common work, but I am unsure why we need different behaviour.

app/grandchallenge/algorithms/views.py
jmsmkn: Verification should only be required for container uploads, which is why it is in `MethodCreate`. Changing the algorithm settings like this should not require verification. Verification is only there to protect our resources somewhat. Looking through the history:

- AlgorithmImageCreate: Should stay to prevent spam containers (consumes storage)
- JobCreate: Should stay to prevent spam jobs (consumes storage and compute)
- AlgorithmRepositoryUpdate: Should stay to prevent spam containers (consumes storage)
- AlgorithmModelCreate: Should stay to prevent spam models (consumes storage)

However:

- AlgorithmCreate (#1979): This was added when we allowed all challenge users to create algorithms, we now have other limits in place via the phase algorithm create so I do not think that this is necessary anymore.
- AlgorithmUpdate (#2030): Added for GitHub checking, those settings are no longer there and are now in AlgorithmRepositoryUpdate. I don't think there is anything in there that now requires verification.
- AlgorithmDescriptionUpdate (#2171): Copied from #2030 when the settings were split, again I don't think this is necessary.
- Also in AlgorithmInterfacePermissionMixin: I don't see a reason to limit it to verified users.

app/grandchallenge/algorithms/forms.py
amickan: I thought it would make sense to raise an error in the admin to make the staff user aware of the fact that there already is an interface with the chosen inputs and outputs, but you're right that we can also just return the existing interface. 

Would you agree that the check for an existing interface on the algorithm that I added to the `clean` function of `AlgorithmInterfaceGetOrCreateForm` is useful? We can also just not create another entry if that is the case, but I thought it would be useful to alert the user that the create form is not meant for updating the "default" interface.

jmsmkn: > Would you agree that the check for an existing interface on the algorithm that I added to the clean function of AlgorithmInterfaceGetOrCreateForm is useful?

I think it would be more user friendly to handle it here rather than sending them elsewhere.

amickan: Ok. Then we will end up with two ways of updating the `is_default` property of an algorithm interface (CreateView and  UpdateView), which felt strange to me, but I don't have a strong opinion on this to be honest and am also fine with handling it here.

app/grandchallenge/algorithms/views.py
amickan: Ok, I just wanted to stay within the existing pattern/logic, but your argumentation is convincing. I will remove the verification check.

app/grandchallenge/algorithms/forms.py
jmsmkn: I think that's okay - because these objects are immutable and globally unique the create view and update view are going to be the same thing, right?

amickan: I have not thought much about the update view. All it should allow a user to do is update the `is_default` property on the through table entry. So I was envisioning a separate form and view for that. Possibly using htmx so that it can be done from the list view, or else a button on a detail view (like the activate algorithm image view).

amickan: I'm working on using the custom `create()` method rather than the usual `save()` on the `ModelForm`. I think the only way to do that is to override the form's `save` method since we don't want the usual saving to happen: 

```python
        return AlgorithmInterface.objects.create(
            inputs=self.cleaned_data["inputs"], outputs=self.cleaned_data["outputs"]
        )
```

That works fine for the form in the user-facing view, but it does not work in the admin. Django complains that it needs a `save_m2m` method (which is usually attached to the form in the `save` method when `commit=False`). I don't quite understand why it is throwing this error but it looks like the admin view handles things a little different and breaks when you override the save method. Is there a better way to achieve this?

amickan: I think in the admin, the model's `save` method is called directly, which explains the above error, but overriding the save method on the model does not work either (and is also not recommended in the Django docs). Maybe I'm overlooking a simple way of doing this, but if not, maybe raising an error like I did before in the admin is the way to go after all.

jmsmkn: Maybe lets save ourself some time - do we really need to do this in the admin? We could treat this as a no go. When would we need to create interfaces without going through an algorithm? Any algorithm editor can do it themselves if they need to (add algorithm perms), and it would be done automatically for challenge participants. The only other place I can think of doing this would be for phases but again a standard form will do. Best to pick our battles.

amickan: Ok. The only use case I see is someone without the add_algorithm permission wanting to host an alogirhtm outside of a challenge. They pay us and so far, I think, we did the basic set-up for them. That would no longer work if we can't create interfaces in the admin. But it would probably be better anyway to grant this person add_algorithm permission through a special group membership?

amickan: Forget what I say above: we could create the algorithm and its interfaces for them through the UI, of course. Duh.

amickan: I'll remove the add ability from the admin views then

amickan:
Ready for re-review

app/grandchallenge/algorithms/forms.py
```diff
@@ -1369,11 +1396,31 @@ def clean(self):
...
+            AlgorithmAlgorithmInterface.objects.filter(
+                algorithm=self._algorithm, interface=interface
+            ).update(is_default=self.cleaned_data["set_as_default"])
+        else:
+            self._algorithm.interfaces.add(
+                interface,
+                through_defaults={
+                    "is_default": self.cleaned_data["set_as_default"]
+                },
+            )
```
jmsmkn, line 1424: `update` returns the number of changed objects, so I think you can do this in one call (swapping a LBYL for an EAFP):

```suggestion
        affected_rows = AlgorithmAlgorithmInterface.objects.filter(
                algorithm=self._algorithm, interface=interface
            ).update(is_default=self.cleaned_data["set_as_default"])
        
        if affected_rows == 0:
            self._algorithm.interfaces.add(
                interface,
                through_defaults={
                    "is_default": self.cleaned_data["set_as_default"]
                },
            )
        elif affected_rows > 1:
            raise RuntimeError("This _really_ shouldn't happen")
```

jmsmkn: Actually - not sure about that - does `affected_rows` return the number of objects changed or the number of objects in the filter?

jmsmkn: I asked ChatGPT and it suggests that it does behave in the same way that I originally suspected so this should work - the `affected_rows` is the number of rows in the filter regardless of whether `is_default` was changed. But, good to double check this.

app/grandchallenge/algorithms/views.py
jmsmkn: Added #3744 for cleanup of the existing views.

app/grandchallenge/algorithms/forms.py
amickan: As per the documentation: 

> The update() method is applied instantly and returns the number of rows matched by the query (which may not be equal to the number of rows updated if some rows already have the new value).

But that doesn't matter. Your code still works since there should always be either 1 or 0 matching rows.

amickan: It just semantically is no longer correct to call them `affected_rows`, maybe `rows_to_update` is better?

jmsmkn: I don't think `rows_to_update` works as the tense is misleading - the updating has already happened by the time the value is assigned. I suggested affected rows as that is what the Django docs calls them, but maybe `updated_rows` or `number_of_updated_interfaces` is clearer.

amickan: But that suggests that it reflects the number of rows that were actually updated, which is not what update returns. Going by what the Django documentation calls them, `matched_rows` is closest to describing what `update` returns.

app/grandchallenge/algorithms/forms.py
```diff
@@ -1369,11 +1396,31 @@ def clean(self):
...
+
+        if matched_rows == 0:
+            self._algorithm.interfaces.add(
+                interface,
+                through_defaults={
+                    "is_default": self.cleaned_data["set_as_default"]
+                },
+            )
+        elif matched_rows > 1:
+            raise RuntimeError("This _really_ shouldn't happen")
```
jmsmkn, line 1424:
```suggestion
            raise RuntimeError(
                "This algorithm and interface are associated "
                "with each other more than once."
            )
```

jmsmkn (approved):

PR #3751, opened 2024-12-13
Removes oninput for challenge request form

### LINKED ISSUE #3748: Challenge request clarification questions show up when they should not
**Describe the bug**
When making a challenge request, user now have to fill in the extra clarification questions ("Why are you not willing/able to support this challenge long-term?" and "Why are you not willing/able to use a CC-BY license for your data?") even when they check the corresponding checkboxes. I think this is related to a recent change in the default of the checkbox fields (them being opt-in now when they apparently were opt-out before). 

**To Reproduce**
1. Go to https://gc.localhost/challenges/requests/create/
2. Scroll down to see the "We agree to support this challenge for up to 5 years." checkbox question. It is not checked initially and because of that you see the "Why are you not willing/able to support this challenge long-term? " textfield underneath. 
3. Check the "We agree to support this challenge for up to 5 years." checkbox. The extra question should disappear, but it does not. 

**Expected behavior**
The textfield to explain why they do not want to support the challenge for 5 years should only be visible when the checkbox is unchecked. 

The same holds for the data license checkbox further below. 

image

image

jmsmkn:
The old functionality was to drop invalid values, this does not seem user friendly. Now, we use the default behaviour of the browser forcing the user to fill in a valid value when they try to submit the form instead.

This also removes the budget default of zero which prevents the requester overlooking the field.

Closes #3748

### DIFF
--- a/app/grandchallenge/challenges/forms.py
+++ b/app/grandchallenge/challenges/forms.py
@@ -193,33 +193,6 @@ class ChallengeRequestForm(forms.ModelForm):
         widgets = {
             "start_date": forms.TextInput(attrs={"type": "date"}),
             "end_date": forms.TextInput(attrs={"type": "date"}),
-            "expected_number_of_teams": forms.NumberInput(
-                attrs={"oninput": "validity.valid||(value='');"}
-            ),
-            "number_of_tasks": forms.NumberInput(
-                attrs={"oninput": "validity.valid||(value='');"}
-            ),
-            "average_size_of_test_image_in_mb": forms.NumberInput(
-                attrs={"oninput": "validity.valid||(value='');"}
-            ),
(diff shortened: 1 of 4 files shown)


amickan (dismissed):
LGTM! I can't remember why (presumably) I added this strange `oninput` validation, but it indeed seems redundant, user unfriendly and also not in line with the Django framework.

jmsmkn:
My guess is for making it easier for the budget response from HTMX, but I don't think that is done anymore.

amickan:
> My guess is for making it easier for the budget response from HTMX, but I don't think that is done anymore.

Ah that's very possible, but that was indeed removed.

amickan (approved):

Issue #3623, opened 2024-10-15
Quoted text not clear


jmsmkn:
Quoted text on GitHub is obvious:

> This is a quote

On GC that is not so. The styling should be updated.

PR #3752, opened 2024-12-13
Adjust quote styling

koopmant:
Adjust blockquote style to make quotes more obvious. 

closes https://github.com/comic/grand-challenge.org/issues/3623

### DIFF
--- a/app/grandchallenge/core/static/css/core.css
+++ b/app/grandchallenge/core/static/css/core.css
@@ -242,3 +242,10 @@ a[aria-expanded="false"] .fa-chevron-down {
 .htmx-request.htmx-indicator {
     display: inline-block;
 }
+
+blockquote {
+    padding: 0 1em;
+    font-size: inherit !important;
+    border-left: .25em solid var(--dark);
+    color: #595959;
+}
--- a/app/grandchallenge/core/static/css/core.css
+++ b/app/grandchallenge/core/static/css/core.css
@@ -242,10 +242,3 @@ a[aria-expanded="false"] .fa-chevron-down {
 .htmx-request.htmx-indicator {
     display: inline-block;
 }
-
-blockquote {
-    padding: 0 1em;
-    font-size: inherit !important;
-    border-left: .25em solid var(--dark);
-    color: #595959;
-}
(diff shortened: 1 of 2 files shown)


app/grandchallenge/core/static/css/core.css
```diff
@@ -242,3 +242,10 @@ a[aria-expanded="false"] .fa-chevron-down {
 .htmx-request.htmx-indicator {
     display: inline-block;
 }
+
+blockquote {
+    padding: 0 1em;
+    font-size: inherit !important;
+    border-left: .25em solid var(--dark);
+    color: #595959;
```
koopmant, line 250: Could also use `var(--primary)` here, but opted for a lighter gray to mute the text a bit more, while still having enough contrast for accessibility (according to this website: https://webaim.org/resources/contrastchecker/)

I only found the colors defined in bootstrap. We don't have more colors defined somewhere do we?

jmsmkn: Is it possible to do this in base.scss and use the bootstrap spacing and colour variables? I think that is important for consistency.

koopmant: Yes, thank you for pointing me in that direction.

koopmant:
Preview:
image

I've opted for the primary color now, because the bootstrap gray colors are overridden in our flatly theme and that does not leave me with the dark gray that I'd like to use. (Not sure why Flatly chose to do it this way; I think it's a bit odd because this leaves a gap in between gray-700 and gray-800, see below.)
I thought about resetting the gray colors back to the bootstrap original values, but then I'd have to apply fixes for those changes as well (such as setting the secondary color from gray-600 to gray-500), so I decided to not do that. 

Flatly (left), bootstrap (right). Gray 500, 600, and 700 are overridden. 
image   image

jmsmkn (approved):

jmsmkn:
I get this when trying to merge, do not know what is going on:

Screenshot 2024-12-16 at 16 00 44

koopmant:
I see the same thing when I try it. I have not experienced this before.

jmsmkn:
It fixed itself.

PR #3753, opened 2024-12-16
Enable interface selection for algorithm job

amickan:
This enables selecting an interface when trying out an algorithm. Interface selection happens in a separate form and view, but is unnecessary if an algorithm has only 1 interface. 

image

### DIFF
--- a/app/grandchallenge/algorithms/migrations/0065_create_interfaces_for_jobs.py
+++ b/app/grandchallenge/algorithms/migrations/0065_create_interfaces_for_jobs.py
@@ -1,0 +1,28 @@
+from django.db import migrations
+
+
+def add_algorithm_interfaces_to_jobs(apps, _schema_editor):
+    Algorithm = apps.get_model("algorithms", "Algorithm")  # noqa: N806
+    Job = apps.get_model("algorithms", "Job")  # noqa: N806
+
+    for algorithm in Algorithm.objects.prefetch_related("interfaces").all():
+        default_interface = algorithm.interfaces.get(
+            algorithmalgorithminterface__is_default=True
+        )
+        jobs = Job.objects.filter(algorithm_image__algorithm=algorithm)
+        for job in jobs:
+            job.algorithm_interface = default_interface
+        jobs.bulk_update(jobs, ["algorithm_interface"])
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        (
+            "algorithms",
+            "0064_create_algorithm_interfaces",
+        ),
+    ]
+
+    operations = [
+        migrations.RunPython(add_algorithm_interfaces_to_jobs, elidable=True),
+    ]
--- a/app/grandchallenge/algorithms/migrations/0065_create_interfaces_for_jobs.py
+++ b/app/grandchallenge/algorithms/migrations/0065_create_interfaces_for_jobs.py
@@ -1,18 +1,29 @@
 from django.db import migrations
 
+from grandchallenge.algorithms.models import (
+    get_existing_interface_for_inputs_and_outputs,
+)
+
 
 def add_algorithm_interfaces_to_jobs(apps, _schema_editor):
-    Algorithm = apps.get_model("algorithms", "Algorithm")  # noqa: N806
+    AlgorithmInterface = apps.get_model(  # noqa: N806
+        "algorithms", "AlgorithmInterface"
+    )
     Job = apps.get_model("algorithms", "Job")  # noqa: N806
 
-    for algorithm in Algorithm.objects.prefetch_related("interfaces").all():
-        default_interface = algorithm.interfaces.get(
-            algorithmalgorithminterface__is_default=True
+    jobs = Job.objects.select_related("algorithm_image__algorithm").all()
+    for job in jobs:
+        interface = get_existing_interface_for_inputs_and_outputs(
+            inputs=job.inputs.all(), outputs=job.outputs.all()
         )
-        jobs = Job.objects.filter(algorithm_image__algorithm=algorithm)
-        for job in jobs:
-            job.algorithm_interface = default_interface
-        jobs.bulk_update(jobs, ["algorithm_interface"])
+        if not interface:
+            interface = AlgorithmInterface.objects.create()
+            interface.inputs.set(job.inputs.all())
+            interface.outputs.set(job.outputs.all())
+
+        job.algorithm_interface = interface
+        job.algorithm_image.algorithm.interfaces.add(interface)
+    jobs.bulk_update(jobs, ["algorithm_interface"])
 
 
 class Migration(migrations.Migration):
--- a/app/grandchallenge/algorithms/migrations/0065_create_interfaces_for_jobs.py
+++ b/app/grandchallenge/algorithms/migrations/0065_create_interfaces_for_jobs.py
@@ -1,37 +1,17 @@
 from django.db import migrations
 
-from grandchallenge.algorithms.models import (
-    get_existing_interface_for_inputs_and_outputs,
-)
-
 
 def add_algorithm_interfaces_to_jobs(apps, _schema_editor):
-    AlgorithmInterface = apps.get_model(  # noqa: N806
-        "algorithms", "AlgorithmInterface"
-    )
+    Algorithm = apps.get_model("algorithms", "Algorithm")  # noqa: N806
     Job = apps.get_model("algorithms", "Job")  # noqa: N806
 
-    jobs = (
-        Job.objects.select_related("algorithm_image__algorithm")
-        .prefetch_related("inputs", "outputs")
-        .all()
-    )
-    for job in jobs:
-        inputs = [input.interface for input in job.inputs.all()]
-        outputs = [output.interface for output in job.outputs.all()]
-
-        interface = get_existing_interface_for_inputs_and_outputs(
-            model=AlgorithmInterface, inputs=inputs, outputs=outputs
+    for algorithm in Algorithm.objects.prefetch_related("interfaces").all():
+        default_interface = algorithm.interfaces.get(
+            algorithmalgorithminterface__is_default=True
         )
-        if not interface:
-            interface = AlgorithmInterface.objects.create()
-            interface.inputs.set(inputs)
-            interface.outputs.set(outputs)
-
-        job.algorithm_interface = interface
-        job.algorithm_image.algorithm.interfaces.add(interface)
-
-    jobs.bulk_update(jobs, ["algorithm_interface"])
+        Job.objects.filter(algorithm_image__algorithm=algorithm).update(
+            algorithm_interface=default_interface
+        )
 
 
 class Migration(migrations.Migration):
--- a/app/grandchallenge/algorithms/migrations/0065_create_interfaces_for_jobs.py
+++ b/app/grandchallenge/algorithms/migrations/0065_create_interfaces_for_jobs.py
@@ -11,18 +11,26 @@ def add_algorithm_interfaces_to_jobs(apps, _schema_editor):
     )
     Job = apps.get_model("algorithms", "Job")  # noqa: N806
 
-    jobs = Job.objects.select_related("algorithm_image__algorithm").all()
+    jobs = (
+        Job.objects.select_related("algorithm_image__algorithm")
+        .prefetch_related("inputs", "outputs")
+        .all()
+    )
     for job in jobs:
+        inputs = [input.interface for input in job.inputs.all()]
+        outputs = [output.interface for output in job.outputs.all()]
+
         interface = get_existing_interface_for_inputs_and_outputs(
-            inputs=job.inputs.all(), outputs=job.outputs.all()
+            model=AlgorithmInterface, inputs=inputs, outputs=outputs
         )
         if not interface:
             interface = AlgorithmInterface.objects.create()
-            interface.inputs.set(job.inputs.all())
-            interface.outputs.set(job.outputs.all())
+            interface.inputs.set(inputs)
+            interface.outputs.set(outputs)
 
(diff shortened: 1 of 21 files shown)


app/grandchallenge/algorithms/models.py
```diff
@@ -1072,6 +1087,9 @@ class Job(CIVForObjectMixin, ComponentJob):
     algorithm_model = models.ForeignKey(
         AlgorithmModel, on_delete=models.PROTECT, null=True, blank=True
     )
+    algorithm_interface = models.ForeignKey(
+        AlgorithmInterface, on_delete=models.PROTECT, null=True, blank=True
+    )
```
amickan, line 1092: Eventually this will need to made `blank=False` and `null=False`

app/grandchallenge/algorithms/urls.py
```diff
@@ -133,7 +134,16 @@
...
+    path(
+        "<slug>/jobs/interface-select/",
+        JobInterfaceSelect.as_view(),
+        name="job-interface-select",
+    ),
+    path(
+        "<slug>/<interface>/jobs/create/",
+        JobCreate.as_view(),
+        name="job-create",
+    ),
```
amickan, line 146: Unsure if we want to expose the interface pks like this? But the only alternative I could come up with was to pass the interface as a URL parameter, but that is less elegant and also still exposes the pk.

app/grandchallenge/algorithms/views.py
```diff
@@ -478,15 +479,11 @@ def get_success_url(self):
         return self.algorithm.get_absolute_url()
 
 
-class JobCreate(
+class JobCreatePermissionMixin(
```
amickan, line 482: Unsure about the permissions implementation here. I used what we currently have on the `JobCreateView` and put it into a mixin for re-use across both views. But it uses `get_permission_object`, so we probably want to change the way we check permissions. We could use the `AccessMixin`, but then we have the issue that the `VerificationRequiredMixin` already subclasses that, which might be problematic. Is the best way forward then to drop the verification mixin and include the verification check in the dispatch method of this new mixin here?

app/grandchallenge/algorithms/urls.py
jmsmkn: The `pk` is a uuid so it is fine to expose right?

app/grandchallenge/algorithms/views.py
jmsmkn: `VerificationRequiredMixin` uses `UserPassesTestMixin` which cannot be mixed with other `UserPassesTestMixin`. That is due to `test_func` only being able to be defined once as the base raises `NotImplementedError` so `super()` cannot be reliably used, meaning multiple `test_func`s would be skipped.

If `VerificationRequiredMixin` used `AccessMixin` it can be mixed with other `AccessMixin`s. That is because the `dispatch` method that it overrides calls `super()`.

app/grandchallenge/algorithms/models.py
```diff
@@ -507,11 +515,14 @@ def default_workstation(self):
...
-        except ObjectDoesNotExist:
-            return None
+            return self.interfaces.get()
+        except MultipleObjectsReturned:
+            try:
+                return self.interfaces.get(
+                    algorithmalgorithminterface__is_default=True
+                )
+            except ObjectDoesNotExist:
+                return None
```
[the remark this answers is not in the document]
jmsmkn: It is a requirement but you may have the problem where Algorithms need to exist before  interfaces can be attached to them.

app/grandchallenge/algorithms/urls.py
amickan: True, it's a uuid. Ok, thanks for confirming!

app/grandchallenge/algorithms/views.py
amickan: Alright, then I will change the `VerificationRequiredMixin`. I will do that in a separate PR to main.

app/grandchallenge/algorithms/models.py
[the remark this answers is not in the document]
amickan: Ok, so it is a requirement that every algorithm has one (and exactly one) default interface. Then I will add a check to the form to ensure that.

app/grandchallenge/algorithms/urls.py
jmsmkn: I think the specification that it is a uuid is missing from the url, same with slug, it should be: `"<slug:slug>/<uuid:interface>/jobs/create/"` https://docs.djangoproject.com/en/5.1/topics/http/urls/#path-converters

amickan: Good point. I'll fix that.

app/grandchallenge/algorithms/migrations/0065_create_interfaces_for_jobs.py
```diff
@@ -0,0 +1,28 @@
...
+
+def add_algorithm_interfaces_to_jobs(apps, _schema_editor):
+    Algorithm = apps.get_model("algorithms", "Algorithm")  # noqa: N806
+    Job = apps.get_model("algorithms", "Job")  # noqa: N806
+
+    for algorithm in Algorithm.objects.prefetch_related("interfaces").all():
+        default_interface = algorithm.interfaces.get(
+            algorithmalgorithminterface__is_default=True
+        )
+        jobs = Job.objects.filter(algorithm_image__algorithm=algorithm)
```
amickan, line 12: Note to self: using the current `default_interface` of the algorithm for all its past jobs might be incorrect for some jobs. Editors might have changed the interfaces between jobs. So I guess the better approach would be to determine the job's interface based on its inputs. When this piece of code runs, each algorithm will have only 1 interface though. So then we should also add the job's interface to the list of interfaces for the algorithm if it's not listed there yet.

app/grandchallenge/algorithms/models.py
```diff
@@ -967,13 +971,13 @@ def retrieve_existing_civs(*, civ_data):
...
     def get_jobs_with_same_inputs(
-        self, *, inputs, algorithm_image, algorithm_model
+        self, *, inputs, interface, algorithm_image, algorithm_model
     ):
         existing_civs = self.retrieve_existing_civs(civ_data=inputs)
         unique_kwargs = {
             "algorithm_image": algorithm_image,
         }
-        input_interface_count = algorithm_image.algorithm.inputs.count()
+        input_interface_count = interface.inputs.count()
```
amickan, line 980: I'm unsure about this one. I'm wondering now if it would be enough to count the provided inputs and first filter on jobs with that number of inputs, and then do the matching inputs annotation. Then we'd be checking for jobs with exactly the inputs provided, right?

jmsmkn: That check is already being done on L990 with `.filter(inputs_match_count=input_interface_count)` right?

app/grandchallenge/algorithms/migrations/0065_create_interfaces_for_jobs.py
```diff
@@ -0,0 +1,47 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        (
+            "algorithms",
+            "0064_create_algorithm_interfaces",
+        ),
+    ]
+
+    operations = [
+        migrations.RunPython(add_algorithm_interfaces_to_jobs, elidable=True),
```
jmsmkn, line 46: Can this be consolidated into the existing data migration 0064? I think you can have multiple RunPython operations.

app/grandchallenge/algorithms/migrations/0065_create_interfaces_for_jobs.py
```diff
@@ -0,0 +1,47 @@
...
+        "algorithms", "AlgorithmInterface"
+    )
+    Job = apps.get_model("algorithms", "Job")  # noqa: N806
+
+    jobs = (
+        Job.objects.select_related("algorithm_image__algorithm")
+        .prefetch_related("inputs", "outputs")
+        .all()
+    )
+    for job in jobs:
```
jmsmkn, line 19: This is going to be a huge migration, we have about a million jobs. They will not fit into memory nor will you be able to do a bulk update by operating on each one. You need to think about the strategy here.

You know that at this point each algorithm only has one default interface. So, you know that in all likelyhood only that interface was used for all jobs for that algorithm. So you could loop over all algorithms, find the interface, then set that for all of that algorithms jobs with one update (`Job.objects.filter(...).update(algorithm_interface=...)`).

This will not work for algorithms whose interfaces changed throughout time, but that should be minor. If that information is lost or inconsistent does it really matter?

I don't think that we should go back and update the algorithms interface based on what was fed to the job - that can be incomplete and come from very old data that are no longer supported by the current version of the algorithm.

Another question is if we need this migration at all if all jobs and evaluations are in a final state.

app/grandchallenge/algorithms/urls.py
```diff
@@ -133,7 +134,16 @@
...
     ),
     path("<slug>/jobs/", JobsList.as_view(), name="job-list"),
-    path("<slug>/jobs/create/", JobCreate.as_view(), name="job-create"),
+    path(
+        "<slug:slug>/jobs/interface-select/",
+        JobInterfaceSelect.as_view(),
+        name="job-interface-select",
+    ),
+    path(
+        "<slug:slug>/<uuid:interface>/jobs/create/",
```
jmsmkn, line 143: I think this would help clarity as you're passing a pk around and not an instance:

```suggestion
        "<slug:slug>/<uuid:interface_pk>/jobs/create/",
```

app/grandchallenge/algorithms/views.py
```diff
@@ -478,28 +479,84 @@ def get_success_url(self):
...
+        context.update(
+            {
+                "algorithm": self.algorithm,
+                "editors_job_limit": settings.ALGORITHM_IMAGES_COMPLIMENTARY_EDITOR_JOBS,
+            }
+        )
+        return context
+
+    def form_valid(self, form):
+        self.selected_interface = form.cleaned_data.pop("algorithm_interface")
```
jmsmkn, line 528: Is there a need to modify the data here rather than just accessing?

app/grandchallenge/algorithms/templates/algorithms/job_form_create.html
```diff
@@ -36,59 +36,64 @@
 {% block content %}
     <h2>Try-out Algorithm</h2>
 
-    {{ algorithm.job_create_page_markdown|md2html }}
+    {% if not algorithm.interfaces.all %}
+        <p>Your algorithm does not have any interfaces yet. You need to define at least one interface (i.e. input - output combination) before you can try it out.</p>
```
jmsmkn, line 40: Shouldn't non-algorithm editors should get a different message?

jmsmkn (commented):
Looks good! I think the migration is the only thing that needs some more work due to the large amount of existing data. You might need to take another look at the other data migrations in the feature branch and double check that they're likely to run okay.

app/grandchallenge/algorithms/migrations/0065_create_interfaces_for_jobs.py
amickan: Indeed. What you suggest is actually what I had first implemented. I figured we'd end up with inconsistent data, but I think you're right that a job that is no longer consistent with the current interfaces will also not be re-run, and so it should not matter.

app/grandchallenge/algorithms/views.py
amickan: Good point. Just accessing should be fine.

app/grandchallenge/algorithms/migrations/0065_create_interfaces_for_jobs.py
amickan: I commented on this earlier. It probably can, but this migration is dependent on 0064, and I couldn't find anything in the docs that convinced me that the RunPython commands run in order, so I figured it was safer to separate them. Did I miss something?

app/grandchallenge/algorithms/urls.py
amickan: Good point. And I actually think that `slug` should be changed to `algorithm_slug` for clarity. Otherwise it's unclear what object's slug this refers to.

app/grandchallenge/algorithms/models.py
amickan: On that line we're currently checking if the number of matching inputs match those in the interface that we pass to this function. But I'm wondering if we need to pass the interface to this function at all. All we want to know if there are any jobs with the provided inputs, so it should be enough to check whether the `inputs_match_count=len(inputs)`, right?

app/grandchallenge/algorithms/migrations/0065_create_interfaces_for_jobs.py
jmsmkn: I think they run in order otherwise the creation of fields and then their foreign keys wouldn't work, but yeah, we can keep it as is for safety, doesn't really matter.

app/grandchallenge/algorithms/urls.py
amickan: On second thought I will leave `slug` as is, because changing it to `algorithm_slug` here would require doing it for all other urls in the file - not worth it, and it's also actually pretty clear that the slug refers to the algorithm. Don't know why I was confused before.

app/grandchallenge/algorithms/models.py
jmsmkn: What if you have inputs that are subsets of another e.g. `{A, C}` and `{A, B}`?

amickan: That case is covered. This function would not return jobs with partially overlapping inputs. I have also added a test for that. 

I have opted to not pass the interface to this function as it does not seem to be necessary to me.

PR #3760, opened 2024-12-18
Replace deprecated strtobool function

koopmant:
The `distutils` module has been deprecated in Python 3.10 and removed in Python 3.12, [see docs distutils](https://docs.python.org/3/library/distutils.html). 

The [migration advice](https://peps.python.org/pep-0632/#migration-advice):
> For these functions, and any others not mentioned here, you will need to reimplement the functionality yourself. 
> `distutils.util.strtobool`

Copied the strtobool function from 3.11 and adjusted it to return booleans, like our function did.

### DIFF
--- a/app/grandchallenge/core/utils/__init__.py
+++ b/app/grandchallenge/core/utils/__init__.py
@@ -1,4 +1,3 @@
-from distutils.util import strtobool as strtobool_i
 from functools import wraps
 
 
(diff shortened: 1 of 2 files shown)


jmsmkn (approved):

jmsmkn:
Maybe good to add some tests?

jmsmkn (approved):

PR #3761, opened 2024-12-18
Delete user upload after creating civ for file

### LINKED ISSUE #3687: JSON file uploads not removed after adding to a display set
**Describe the bug**
JSON file uploads are not removed after being added to a reader study. This might also affect algorithm jobs and archive items after the recent refactoring in #3465. A consequence is that a verified user cannot upload more than 1000 files per day.

**To Reproduce**
Try to upload a JSON file:

```python
upload_display_sets.append(
            {
                "baseline-lesion-distances": [Path(bl_distances)], # JSON file
                "follow-up-lesion-distances": [Path(fu_distances)], # JSON file
            },
        )
 
display_set_pks = client.add_cases_to_reader_study(
    reader_study=upload_reader_study_slug,
    display_sets=upload_display_sets
)
```

The files will appear in the display set as expected. However, the uploads will not have been removed.

**Expected behavior**

The uploads are removed after processing

ammar257ammar:
closes #3687

### DIFF
--- a/app/tests/algorithms_tests/test_views.py
+++ b/app/tests/algorithms_tests/test_views.py
@@ -1359,34 +1359,11 @@ class TestJobCreateView:
         assert Job.objects.count() == 1
 
     @override_settings(task_eager_propagates=True, task_always_eager=True)
-    @pytest.mark.django_db
-    @pytest.mark.parametrize(
-        "file_json_content,job_status,error_message,detailed_error_message,ci_count,upload_exists",
-        [
-            (
-                b'{"Foo": "bar"}',
-                Job.CANCELLED,
-                "One or more of the inputs failed validation.",
-                {
-                    "ci_json_file": "JSON does not fulfill schema: instance is not of type 'array'"
-                },
-                0,
-                True,
-            ),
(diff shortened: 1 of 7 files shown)


app/tests/algorithms_tests/test_views.py
```diff
@@ -1357,11 +1359,34 @@ def test_create_job_is_idempotent(
...
+                    "ci_json_file": "JSON does not fulfill schema: instance is not of type 'array'"
+                },
+                0,
+                True,
+            ),
+            (b'[{"Foo": "bar"}]', Job.PROVISIONED, "", {}, 1, False),
+        ],
+    )
+    def test_create_job_with_file_input(
         self,
```
amickan, line 1380: I think this test should only test the faulty file input - that's what the test name suggests. So no parametrization necessary. There should be another test for the correct file input, one that comes to mind is `test_create_job_with_multiple_new_inputs` - this test exists for the API create view and the UI view.

ammar257ammar: Done.
I retracted the parametrisation and added a check for upload file deletion in `test_create_job_with_multiple_new_inputs` for both the view and API.

amickan (approved):

PR #3764, opened 2024-12-20
Fix buggy code blocks in markdown renderer

### LINKED ISSUE #3592: Code blocks buggy in markdown renderer
MarkDown such as:

Screenshot 2024-10-08 at 14 24 24

Should render as:

  shouldn't be a code block

```python
first_line = 1
for foo in bar:
  print(foo)
```

But actual result is:

Screenshot 2024-10-08 at 14 24 49

Note:
- The first line should not be a code block
- The first line of the python block for some reason has a space
- Multiple blank lines at the start of a MarkDown block should not result in a code block

chrisvanrun:
## Old HTML render:

image

## New HTML render:

image

Likely this PR fixes some other incorrect visuals since the base extension only handled snippets. I've only discovered the `<code>` one.


app/tests/core_tests/test_markdown.py
```diff
@@ -202,54 +204,57 @@ def test_markdown_rendering(markdown_with_html, expected_output):
 
 
 @pytest.mark.parametrize(
-    "markdown_with_html, expected_output",
```
chrisvanrun, line 205: There were needlessly complex.

koopmant:
I propose to add some extra css to fix a missing bottom margin after the codeblocks have been converted:
in `app/grandchallenge/core/static/css/base.scss`:
```
div.codehilite {
    margin-bottom: $paragraph-margin-bottom;
}
```

image
⤹
image

app/grandchallenge/core/templatetags/bleach.py
```diff
@@ -87,7 +87,7 @@ def md2html(
     post_processors = [*settings.MARKDOWN_POST_PROCESSORS]
 
     if process_youtube_tags:
-        post_processors.append(YOUTUBE_TAG_SUBSTITUTION)
+        post_processors = [YOUTUBE_TAG_SUBSTITUTION, *post_processors]
```
jmsmkn, line 90: Why does `YOUTUBE_TAG_SUBSTITUTION` need to be first now?

app/grandchallenge/core/utils/markdown.py
```diff
@@ -1,61 +1,59 @@
...
-                    self.set_css_class(element=el, class_name=tag_class)
-                    self.md.htmlStash.rawHtmlBlocks[i] = str(bs4block)
 
-    @staticmethod
-    def set_css_class(*, element, class_name):
-        if isinstance(element, ElementTree.Element):
-            current_class = element.attrib.get("class", "")
+class ExtendHTMLTagClasses:
+    def __init__(self, tag_classes):
+        # Make extensions safe
```
jmsmkn, line 31: Escaping is not the same as safety.

app/grandchallenge/core/utils/markdown.py
```diff
@@ -1,61 +1,59 @@
...
+        self.clean_tag_classes = {
```
jmsmkn, line 32: These are not clean classes, they're escaped classes.

app/grandchallenge/core/utils/markdown.py
jmsmkn: Why do they anyway need to be stripped and escaped if they are provided by the class configuration?

app/grandchallenge/core/utils/markdown.py
```diff
@@ -1,61 +1,59 @@
...
+            t: [escape(c).strip() for c in classes]
+            for t, classes in tag_classes.items()
+        }
+
+    def __call__(self, html):
+        input_is_safe = isinstance(html, SafeString)
```
jmsmkn, line 38: In what situations in the codebase does this extension work with safe html?

app/grandchallenge/core/templatetags/bleach.py
chrisvanrun: The post_processors, such as applying GC-specific classes, should happen way at the end as final clean-up. The `YOUTUBE_TAG_SUBSTITUTION`, if IIRC, doesn't introduce any targed HTML-tags. Still I thought it better to have the system-wide post_processors have the final say in the html.

app/grandchallenge/core/utils/markdown.py
```diff
@@ -1,61 +1,59 @@
...
+                element["class"] = classes
 
-            current_class = element["class"]
+        new_markup = soup.decode()
 
-            for name in class_name.split(" "):
-                if class_name not in current_class:
-                    current_class.append(name)
+        if input_is_safe:
+            new_markup = mark_safe(new_markup)
```
jmsmkn, line 54: I do not think that we can be certain that the modifications done are safe.

app/grandchallenge/core/utils/markdown.py
```diff
@@ -1,61 +1,59 @@
...
-    def run(self, root):
-        el_class_dict = {
-            "img": "img-fluid",
-            "blockquote": "blockquote",
-            "table": "table table-hover table-borderless",
-            "thead": "thead-light",
-            "code": "codehilite",
-        }
+    This messes up things if the HTML is decoded into a string again.
+    """
```
jmsmkn, line 16: Why is preserving them necessary?

app/config/settings.py
```diff
@@ -789,13 +789,30 @@ def get_private_ip():
...
+            "img": ["img-fluid"],
+            "blockquote": ["blockquote"],
+            "table": [
+                "table",
+                "table-hover",
+                "table-borderless",
+            ],
+            "thead": ["thead-light"],
+            "code": ["codehilite"],
+        }
```
jmsmkn, line 805: This class is only used once and the configuration is static, it would be better if this configuration is just done on the class once so that the logic is not split up in multiple places unnecessarily.

app/grandchallenge/core/templatetags/bleach.py
jmsmkn: They have different properties - the `YOUTUBE_TAG_SUBSTITUTION` is adding things which are not allowed by cleaning the html. `ExtendHTMLTagClasses` is only adding things which are allowed. `ExtendHTMLTagClasses` should be done before `clean` on L85, and the `mark_safe` handling should be removed from `ExtendHTMLTagClasses`.

jmsmkn (changes requested):
This needs to be reworked so that `mark_safe` is not used as this opens up XSS vulnerabilities.

app/grandchallenge/core/templatetags/bleach.py
jmsmkn: I'm still wondering why this cannot be done using the standard markdown processors. Shouldn't it be implemented as a https://python-markdown.github.io/reference/markdown/postprocessors/ ? e.g.:

```python
from markdown.extensions import Extension
from markdown.postprocessors import Postprocessor
from bs4 import BeautifulSoup

class ClassMapPostprocessor(Postprocessor):
    def run(self, html):
        soup = BeautifulSoup(html, 'html.parser')

        class_map = {
            "img": ["img-fluid"],
            "blockquote": ["blockquote"],
            "table": ["table", "table-hover", "table-borderless"],
            "thead": ["thead-light"],
            "code": ["codehilite"]
        }
        
        for element in soup.find_all([*class_map.keys()]):
            existing_classes = element.get('class', [])
            element['class'] = existing_classes + self.class_map[element.name]

        return str(soup)

class ClassMapExtension(Extension):
    def extendMarkdown(self, md):
        post_processor = ClassMapPostprocessor(md)
        md.postprocessors.register(post_processor, 'class_map', 100)
```

jmsmkn: I spent too long trying to understand this that I ended up implementing it, as Chris is off today I will push and go with it as I think this is the way to go. We lose the email obfuscation as discussed but that can be replaced by a tag if absolutely necessary. I don't think that in 2025 it is.

jmsmkn (approved):

PR #3768, opened 2024-12-27
Apply ordering of CIVs to API serializations

chrisvanrun:
Ensures the ordering of CIVs on the rendered views is now also applied to the API representations. This way, the order is the same when viewing them within a view and CIRRUS.

### DIFF
--- a/app/grandchallenge/components/serializers.py
+++ b/app/grandchallenge/components/serializers.py
@@ -14,6 +14,7 @@ from grandchallenge.components.models import (
     ComponentInterface,
     ComponentInterfaceValue,
 )
+from grandchallenge.components.templatetags.civ import sort_civs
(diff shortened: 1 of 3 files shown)


app/grandchallenge/components/serializers.py
```diff
@@ -162,6 +163,12 @@ def validate(self, attrs):
         return attrs
 
 
+class SortedCIVSerializer(serializers.ListSerializer):
+    def to_representation(self, data):
+        sorted_data = sort_civs(data.all())
+        return super().to_representation(sorted_data)
```
amickan, line 169: LGTM, but maybe add a test for this new serializer?

pkcakeout:
Holy cabowly:

image

That is a branch-name to behold!

amickan (approved):

PR #3771, opened 2024-12-30
Fix incorrectly defined table data when there are no evaluations

### LINKED ISSUE #3770: Datatable error on uploading new ground-truth
> I keep getting this error after I upload a new ground truth tar file. The upload still works fine though. Just thought I'd let you know

image

chrisvanrun:
Closes #3770 

The `<td>` were missconstrued when the table had no data. The previous version of Datatables might have been reslient to this. The default is adequate:

image

Also applied some non-sortable classification to the relevant columns.

### DIFF
--- a/app/grandchallenge/evaluation/templates/evaluation/partials/evaluations_for_object_table.html
+++ b/app/grandchallenge/evaluation/templates/evaluation/partials/evaluations_for_object_table.html
@@ -4,10 +4,10 @@
         <tr>
             <th>Created</th>
             <th>Updated</th>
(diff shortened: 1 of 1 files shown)


amickan (approved):
Good to know. I was about to use the same construct for the algorithm interface table, but I'll be sure not to.

chrisvanrun:
> Good to know. I was about to use the same construct for the algorithm interface table, but I'll be sure not to.

I think there is a decent option via the datables API: `language.emptyTable` at https://datatables.net/reference/option/language

PR #3774, opened 2025-01-07
Tweak image upload help text

chrisvanrun:
A user mentioned some confusion by clinicians trying to upload dicoms. Imho the user had a valid point. The help text also had the suggestion that .tiff was converted to MHA. That is incorrect. I opted to sacrifice some horizontal space for clarity.

## Before

image

## After

image

Note: text is used in a widget, no need for a migration.

### DIFF
--- a/app/grandchallenge/cases/forms.py
+++ b/app/grandchallenge/cases/forms.py
@@ -8,20 +8,26 @@ from grandchallenge.core.guardian import get_objects_for_user
 from grandchallenge.uploads.models import UserUpload
 from grandchallenge.uploads.widgets import UserUploadMultipleWidget
 
-IMAGE_UPLOAD_HELP_TEXT = (
-    "The total size of all files uploaded in a single session "
-    "cannot exceed 10 GB. "
-    "A maximum of 100 files can be uploaded per session. "
-    "Please only upload one volume per session. "
-    "If your volume consists of more than this then please compress the "
-    "directory and upload a single .zip file.<br>"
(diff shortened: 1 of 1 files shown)


app/grandchallenge/cases/forms.py
```diff
@@ -8,20 +8,26 @@
...
-    "directory and upload a single .zip file.<br>"
-    "The following file formats are supported and will be converted to MHA: "
-    ".mha, .mhd, .raw, .zraw, .dcm, .nii, .nii.gz, .nrrd, "
-    ".fda, .fds, .tiff, .png, .jpeg and .jpg.<br>"
-    "The following file formats can be uploaded and will be converted to "
-    "TIFF: Aperio (.svs), Hamamatsu (.vms, .vmu, .ndpi), Leica (.scn), "
-    "MIRAX (.mrxs), Ventana (.bif) and DICOM-WSI (.dcm)."
-)
+IMAGE_UPLOAD_HELP_TEXT = """
+<p>
```
jmsmkn, line 12: I do not like all of this extra whitespace. Now you have a bunch of floating statements in the form. Before, it was clear all statements belong to the widget, now, do some belong to the widget and some to the button below? Unclear. I would prefer just to keep to line breaks if necessary. I don't think that adding empty lines will mean that users will suddenly start reading the docs.

chrisvanrun: Check! Reverted to the line breaks, screenshots updated.

PR #3776, opened 2025-01-08
Remove verification required mixin from some algorithm views

### LINKED ISSUE #3744: Remove verification required from some algorithm views
Verification should only be required for container uploads, which is why it is in `MethodCreate`. Changing the algorithm settings like this should not require verification. Verification is only there to protect our resources somewhat. Looking through the history:

- AlgorithmImageCreate: Should stay to prevent spam containers (consumes storage)
- JobCreate: Should stay to prevent spam jobs (consumes storage and compute)
- AlgorithmRepositoryUpdate: Should stay to prevent spam containers (consumes storage)
- AlgorithmModelCreate: Should stay to prevent spam models (consumes storage)

However:

- AlgorithmCreate (#1979): This was added when we allowed all challenge users to create algorithms, we now have other limits in place via the phase algorithm create so I do not think that this is necessary anymore.
- AlgorithmUpdate (#2030): Added for GitHub checking, those settings are no longer there and are now in AlgorithmRepositoryUpdate. I don't think there is anything in there that now requires verification.
- AlgorithmDescriptionUpdate (#2171): Copied from #2030 when the settings were split, again I don't think this is necessary.

ammar257ammar:
closes #3744

### DIFF
--- a/app/grandchallenge/algorithms/views.py
+++ b/app/grandchallenge/algorithms/views.py
@@ -105,7 +105,6 @@ logger = logging.getLogger(__name__)
 class AlgorithmCreate(
     LoginRequiredMixin,
     PermissionRequiredMixin,
-    VerificationRequiredMixin,
     UserFormKwargsMixin,
     CreateView,
 ):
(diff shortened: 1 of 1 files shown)


ammar257ammar:
Removing `VerificationRequiredMixin` from three views (AlgorithmCreate, AlgorithmUpdate, AlgorithmDescriptionUpdate) did not trigger any error in the tests. Looking at the PRs mentioned in the issue, the used user was verified in all tests.
Now the mixin is removed, what test is recommended here?

PR #3778, opened 2025-01-09
Preview emails using the full page editor and standard email rendering

### LINKED ISSUE #3682: Extend usage new MD widget
A few places where I think it can also be used nicely:
- [ ] Email writing (is custom rendering possible without ANY site CSS?)
- [x] Blog post creation
- [x] Page documentation

jmsmkn:
Adds the full page markdown editor for emails along with iframed previews using the standard email rendering. As there are differences in email clients and browsers a test email still needs to be sent out, but at least the previews are isolated from the GC CSS/JS environment.

Screenshot 2025-01-09 at 11 37 36

Screenshot 2025-01-09 at 11 37 56

Closes #3682

### DIFF
--- a/app/grandchallenge/emails/forms.py
+++ b/app/grandchallenge/emails/forms.py
@@ -1,8 +1,9 @@
 from django import forms
 
 from grandchallenge.core.forms import SaveFormInitMixin
-from grandchallenge.core.widgets import MarkdownEditorFullPageWidget
 from grandchallenge.emails.models import Email
+from grandchallenge.emails.widgets import MarkdownEditorEmailFullPageWidget
+from grandchallenge.subdomains.utils import reverse
 
 
 class EmailMetadataForm(SaveFormInitMixin, forms.ModelForm):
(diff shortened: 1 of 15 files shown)


koopmant (commented):
Could not find any issues. 

Only downside I see is that the view to edit the title is split from the view to edit the email body. One way to avoid that is to place the `rendered_body()` function outside the model, and then use a regular view to return a httpresponse. (Similar to how it is implemented in markdownx.) Because then you don't need an object pk in the url to get to the view and you can use it directly in the create view as well. I'm not sure if there's anything against that?

jmsmkn:
Indeed, but I followed the existing implementations where only the content is split out from the metadata. Unfortunately here the only metadata for an email is the title. I think having a view that depends on the pk or not is much of a muchness - either would work, this seemed most logical to me due to how it was built up from re-using the rendering on the detail page I had. Maybe in the future we will add more metadata fields that affect the rendering, in which case we definitely would need the object. I don't think that it needs refactoring here, do you agree?

koopmant:
I have to admit it's bugging me a little bit. But if we would need the object in the future, it does not make sense to convert it to a view that's independent of a pk. Let's keep it like this then.

amickan (approved):
Also looks good to me! It is weird to have the title in a separate form, but I don't think it's worth spending effort on given that this is anyway only accessible to a restricted set of users, and the title is not something that people update either.

PR #3782, opened 2025-01-09
Reset start time, end time and error message on job requeue from admin

### LINKED ISSUE #3435: ComponentJob start time, end time and error message are not reset if retried
If you have a failed job which is then retried through the admin action the start time, end time and error message should be reset along with the status.

ammar257ammar:
closes #3435

### DIFF
--- a/app/grandchallenge/components/admin.py
+++ b/app/grandchallenge/components/admin.py
@@ -1,6 +1,5 @@
 from django.contrib import admin
 from django.db.transaction import on_commit
-from django.utils.timezone import now
 from guardian.admin import GuardedModelAdmin
 
 from grandchallenge.components.models import (
@@ -94,7 +93,7 @@ def requeue_jobs(modeladmin, request, queryset):
 
     for job in queryset:
         job.status = ComponentJob.RETRY
-        job.started_at = now()
+        job.started_at = None
         job.completed_at = None
         job.error_message = ""
         job.attempt += 1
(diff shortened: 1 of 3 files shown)


app/grandchallenge/components/admin.py
```diff
@@ -93,12 +94,24 @@ def requeue_jobs(modeladmin, request, queryset):
 
     for job in queryset:
         job.status = ComponentJob.RETRY
+        job.started_at = now()
```
amickan, line 97: Isn't the start time set in `update_status`? Then I think `started_at` should be `None` as well.

ammar257ammar: Done

amickan (commented):
LTGM, just one clarification question: `stdout` and `stderr` get overwritten already when a job is rerun (not appended), right?

ammar257ammar:
True, the `stdout` and `strerr` are updated (overwritten not appended) through the `update_status` function and the the job rerun (i.e. requeue admin function) calls `job.execute` which in turn run `provision_job` -> `execute_job` -> `update_status`

amickan:
> I believe they should also be reset on admin requeue, do you agree?

Yes, I think that makes sense.

amickan:
> However, `stdout`, `strerr` and `detailed_error_message` are updated only in the case the job failed again in the rerun. If the job succeeded, they will keep the values of the failed run.

Actually, are you sure? I think in `handle_event` `stdout` and `stderr` get overridden (via `update_status`) even if the job succeeds. The `detailed_error_message` should still be set to an empty dictionary, but no need to reset `stdout` and `stderr` then, I think.

ammar257ammar:
Indeed, you are right. I also tested it by resetting the fields and retry the failed job and they are overwritten when it succeeded too. okay then, no need to reset.

amickan (approved):

PR #3785, opened 2025-01-14
Forbid activating an algorithm image if the current active one is used by an evaluation in progress

### LINKED ISSUE #3544: Users should not be able to change the active version of an algorithm image whilst an evaluation that is using them is in progress
Otherwise the container can be removed from the registry. Goes for container uploads or for the activation button.

ammar257ammar:
closes #3544

### DIFF
--- a/app/grandchallenge/algorithms/forms.py
+++ b/app/grandchallenge/algorithms/forms.py
@@ -812,7 +812,6 @@ class ImageActivateForm(Form):
             )
             .exists()
         ):
-
             raise ValidationError(
                 "Current active algorithm image cannot be deactivated. An evaluation using it is in progress."
             )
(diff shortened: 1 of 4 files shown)


amickan:
Thanks for starting to tackle this! After discussing with James, we think there is a more user-friendly approach to fixing the underlying problem. The underlying problem is that we remove non-active images from the registry once a day. Rather than keeping users from developing (i.e. changing the active version of an algorithm or an evaluation), we can just make sure that we don't remove images from the registry if they are still being used. 

So that would require adding a check to the periodic task that takes care of removing inactive images: `remove_inactive_container_images`. We would need to check for active evaluations and jobs and ensure that the linked algorithm images and methods are not added to the list of images to remove.

ammar257ammar:
Does that mean when a new algorithm image is uploaded, it will be activated even if an evaluation is running using the current active one?

amickan:
> Does that mean when a new algorithm image is uploaded, it will be activated even if an evaluation is running using the current active one?

Yes. The activating process does not need to be touched, only the logic about removing inactive images from the registry

ammar257ammar:
continues in #3788

PR #3786, opened 2025-01-15
Unmix presentation and content for docpages

koopmant:
This adds css for docpages, to enable writing simple markdown focused on content rather then presentation. 

For example this removes the need to write 
`<h3 class="text-center">heading text</h3>`
to create a centered heading at the top of the page. 
This can now be written simply as 
`### heading text`


app/grandchallenge/core/static/css/base.scss
```diff
@@ -54,3 +54,15 @@ blockquote {
 div.codehilite {
     margin-bottom: $paragraph-margin-bottom;
 }
+
+.docpage h3:first-of-type {
+    text-align: center;
+}
+
+.docpage h3:has(+h4){
+    margin-bottom: 2rem;
```
jmsmkn, line 63: Should a bootstrap variable be used here such as `$spacer` or `$paragraph-margin-bottom`?

jmsmkn (changes requested):
Looks good! Small question and in the new script there are some images that are being used from the production instance. Could you instead create some images on the local instance and use those? They do not have to be the same ones, we have some other jpg/png images in the codebase.

app/grandchallenge/core/utils/markdown.py
```diff
@@ -14,7 +14,7 @@ def run(self, text):
         soup = BeautifulSoup(text, "html.parser")
 
         class_map = {
-            "img": ["img-fluid"],
+            "img": ["border shadow img-fluid"],
```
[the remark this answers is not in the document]
jmsmkn: You could change the processor to have a specific one for documentation like we do for emails, but probably it is overkill. I think I prefer being consistent throughout the site - probably adding a shadow is okay but maybe not a border?

app/grandchallenge/core/static/css/base.scss
koopmant: Ah good point; forgot to replace this!

koopmant:
> Could you instead create some images on the local instance and use those? They do not have to be the same ones, we have some other jpg/png images in the codebase.

Will do!

jmsmkn (approved):

PR #3789, opened 2025-01-17
Algorithm interface delete view

amickan:
This enables users with the `add_algorithm` permission to remove interfaces from their algorithm. 


app/grandchallenge/algorithms/views.py
```diff
@@ -1205,3 +1207,45 @@ def get_context_data(self, *args, **kwargs):
...
+    AlgorithmInterfacePermissionMixin, DeleteView
+):
+    model = AlgorithmAlgorithmInterface
+
+    @property
+    def algorithm_interface(self):
+        return get_object_or_404(
+            klass=AlgorithmAlgorithmInterface,
+            algorithm=self.algorithm,
+            interface__pk=self.kwargs["interface_pk"],
```
jmsmkn, line 1222: You could add the filter `is_default=False` here to cover yourself. I think that returning 404 here is fine rather than 403 as the construction in `dispatch` is a little strange (that sort of check should be done in a Form).

jmsmkn (approved):
Looks good, small suggestion but up to you whether to incorporate it or not.

app/grandchallenge/algorithms/views.py
amickan: It felt a bit awkward to me too and I like your suggestion, so made the change.

PR #3793, opened 2025-01-21
Exclude slugs from historical records 

jmsmkn:
If the slug is included in a historical record the
object can only be saved 100 times before the
AutoSlugField from Django Extensions gives up on trying
to find a unique slug, and the object cannot be saved.


chrisvanrun (approved):
Looks good!

PR #3795, opened 2025-01-22
[Onboarding Tasks] model and admin

chrisvanrun:
Part of the pitch:
- https://github.com/DIAGNijmegen/rse-roadmap/issues/339

This PR adds the model and admin (actions) for the onboarding tasks.

## Showcase

https://github.com/user-attachments/assets/e2d84e4b-8634-49f1-915d-232775a44acf


app/grandchallenge/challenges/models.py
```diff
@@ -1402,3 +1403,81 @@ class ChallengeRequestGroupObjectPermission(GroupObjectPermissionBase):
...
+
+class OnboardingTask(FieldChangeMixin, UUIDModel):
+    ResponsibleChoices = TaskResponsibleChoices
+
+    class Meta:
+        permissions = [
+            ("complete_onboaringtask", "Can mark this task as completed")
+        ]
+
+    created = models.DateTimeField(editable=False)
```
chrisvanrun, line 1421: Add this field without the `auto_now_add` as to have `completed_at` and `created_at` match perfectly when objects are created with `complete=True`.

app/grandchallenge/challenges/admin.py
```diff
@@ -132,6 +138,111 @@ def send_status_update_email(self, request, queryset):
             )
 
 
+class OnTimeFilter(admin.SimpleListFilter):
```
chrisvanrun, line 141: Inverted the overdue. Mainly from the rational that if things are good in the admin, things should be green.

The overdue annotation will likely move to a model manager in a later PR. So keeping that as the overdue.

app/grandchallenge/challenges/admin.py
```diff
@@ -132,6 +138,111 @@ def send_status_update_email(self, request, queryset):
...
+        move_task_deadline_4_weeks,
+    )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._reference_time = now()
+
+    def get_queryset(self, request):
+        qs = super().get_queryset(request)
+        return qs.annotate(
```
chrisvanrun, line 233: Use database-based annotation here for both speed and the option to filter things.

app/grandchallenge/challenges/admin.py
```diff
@@ -141,3 +252,7 @@ def send_status_update_email(self, request, queryset):
     ChallengeRequestGroupObjectPermission, GroupObjectPermissionAdmin
 )
 admin.site.register(ChallengeSeries)
+
+admin.site.register(
```
chrisvanrun, line 256: Not sure about this one, only the Group side of things will be used. But perhaps it is best practice to add a UserObject version as well?

app/grandchallenge/challenges/models.py
```diff
@@ -1402,3 +1403,81 @@ class ChallengeRequestGroupObjectPermission(GroupObjectPermissionBase):
...
+        blank=True, help_text="Description of this task."
+    )
+    complete = models.BooleanField(
+        default=False,
+        help_text="If true, this task is complete.",
+    )
+    completed_at = models.DateTimeField(
+        null=True,
+        blank=True,
+        help_text="Time this task was last marked completed.",
```
amickan, line 1442: Again for my understanding: the help text suggests that we allow tasks to be completed multiple times. Is that the case, and if so, why?

app/grandchallenge/challenges/models.py
```diff
@@ -1402,3 +1403,81 @@ class ChallengeRequestGroupObjectPermission(GroupObjectPermissionBase):
...
+    )
+    responsible = models.CharField(
+        blank=False,
+        max_length=3,
+        choices=ResponsibleChoices.choices,
+        default=ResponsibleChoices.CHALLENGE_ORGANIZERS,
+        help_text="Who is responsible for completion of this task.",
+    )
```
amickan, line 1450: Maybe call this `assigned_to` or `responsible_group`, rather than `responsible`?

app/grandchallenge/challenges/models.py
```diff
@@ -1402,3 +1403,81 @@ class ChallengeRequestGroupObjectPermission(GroupObjectPermissionBase):
...
+    deadline = models.DateTimeField(
+        help_text="Deadline for this task.",
+    )
+
+    def is_overdue(self, reference_time):
+        return not self.complete and self.deadline < reference_time
```
amickan, line 1456: Wouldn't the `reference_time` always just be now if you want to know if a task is overdue at any given point in time?

app/grandchallenge/challenges/models.py
```diff
@@ -1402,3 +1403,81 @@ class ChallengeRequestGroupObjectPermission(GroupObjectPermissionBase):
...
+        super().save(*args, **kwargs)
+
+        if self.responsible == self.ResponsibleChoices.CHALLENGE_ORGANIZERS:
+            assign_perm(
+                "complete_onboaringtask", self.challenge.admins_group, self
+            )
+        elif not adding:
+            remove_perm(
+                "complete_onboaringtask", self.challenge.admins_group, self
+            )
```
amickan, line 1477: Better to separate this out into a `assign_permissions()` function like we do for most models. Makes it easier to follow. 

And for the logic: you are never explicitly assigning the permission to the support group. I understand that we implicitly get that permission through our admin rights, but it took me a moment to understand and I think it would be cleaner to explicitly assign that permission to users with `is_staff=True`. Then we will also be able to do it through the UI since that'll be the permission to get access to the views as well. 

I'm not 100% sure about this, but if you don't explicitly add the permission, I think it'd be nice to at least leave a comment here that explains the logic. 

And what scenario does the elif part cover? If this is for the case that the responsible group changes, why not check `self.has_changed("responsible")`?

app/grandchallenge/challenges/admin.py
amickan: I think the group object one is enough

app/grandchallenge/challenges/admin.py
```diff
@@ -132,6 +138,111 @@ def send_status_update_email(self, request, queryset):
...
+    search_fields = ("title", "description")
+    actions = (
+        mark_task_complete,
+        move_task_deadline_1_week,
+        move_task_deadline_4_weeks,
+    )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._reference_time = now()
```
amickan, line 229: What is `_reference_time` used for?

app/grandchallenge/challenges/models.py
chrisvanrun: This sort-of stemmed from the idea of us looking back at what the resolve time is of the tasks in the future. I added a corner-case in a test and my inner nerd wanted to get that perfect. A rational for reverting it would be that it adds a small amount of complexity.

Adding the `at` would not have it overwrite the parent class, so I'd then need to drop the parent class and add uuid etc myself. That is not something I am fond of doing.

app/grandchallenge/challenges/models.py
chrisvanrun: I think it wise to allow tasks to be have a toggable complete:

- Allow for accidental complete correction.
- Allow for: "no you did **not** complete this" reverting.

The help text is to indicate that it is only the last time someone marked it complete. If it would not mention it: it could also be the first time.

app/grandchallenge/challenges/models.py
chrisvanrun: I agree it's better to change it. I would like to avoid user or group naming since it could suggest an actual implemented user or group. How about `responsible_party`?

app/grandchallenge/challenges/models.py
chrisvanrun: Practically: yes. Technically: no.

The `now()` should be a fixed reference or else you have slightly varying notion of overdue per Task. If two tasks are in the list, having the same deadline, but one is at the start of the list and the other at the end. The one at the start could be considered on time and the task at the end could be considered overdue.

app/grandchallenge/challenges/models.py
chrisvanrun: > Better to separate this out into a assign_permissions() function like we do for most models. Makes it easier to follow.

Right. Makes sense.

> I'm not 100% sure about this, but if you don't explicitly add the permission, I think it'd be nice to at least leave a comment here that explains the logic.

I'll do that. There is no staff group: only users that are marked. As such you'd need to hook a signal whenever that boolean changes to make sure the permissions are setup correctly. I think adding a staff group is out of scope.

> And what scenario does the elif part cover? If this is for the case that the responsible group changes, why not check self.has_changed("responsible")?

My rational, which might be silly, so please let me know, is the following: we should always have a remove_perm call when things are changing, irrespective what the previous state was. So it could be an `else`, however, for performance it's only executed when things are changing (and hence not when adding).

app/grandchallenge/challenges/admin.py
chrisvanrun: Ah! Good find. I had implemented the overdue check on the model level intially, but I found that did now allow for filtering and hence remove it (partially).

It might be usefull for a future PR but for now I've removed things.

chrisvanrun:
@amickan , I wonder: do you for see any other common admin actions we'd like to apply to all tasks of a challenge? We now have 'delete', 'complete' and either delay by 1 week or delay by 4 weeks. Those seemed practical.

app/grandchallenge/challenges/models.py
amickan: Yeah ok, makes sense then. I think this should all be reserved for us staff users though.  A user should only be able to mark it as completed once.

app/grandchallenge/challenges/models.py
amickan: I see why you want to avoid `responsible_group`. What is wrong with `assigned_to` - that feels neutral and clear to me. But `responsible_party` is also fine with me.

app/grandchallenge/challenges/models.py
amickan: A bit of a theoretical cornercase, I think, since the difference in a list of say 25 tasks (our default pagination length in the UI) would be in the millisecond range, so unlikely to make a difference, right?   But I understand the theoretical point, so fine.

app/grandchallenge/challenges/models.py
amickan: > There is no staff group: only users that are marked. As such you'd need to hook a signal whenever that boolean changes to make sure the permissions are setup correctly. I think adding a staff group is out of scope.

Good point, hand't thought of that.

amickan: > we should always have a remove_perm call when things are changing, irrespective what the previous state was.

Why do you think we should always have a remove_perm call when things are changing? Maybe I'm being silly, but you should only remove perms when there is a reason to remove, right? And in this case, I think the only reason to remove the perm would be if the responsible group changes, right? Maybe I'm missing something here.

amickan:
> @amickan , I wonder: do you for see any other common admin actions we'd like to apply to all tasks of a challenge? We now have 'delete', 'complete' and either delay by 1 week or delay by 4 weeks. Those seemed practical.

I think those cover the most plausible actions. We can always add more if the need arises in practice. Maybe we want to able to bulk (re)-assign tasks to either challenge admins or staff, but we can also wait and see if that is actually something we do, or not.

app/grandchallenge/challenges/models.py
chrisvanrun: I imagined some BI in the future, calculating the resolve time for tasks and finding some in `ms` range. Only to find out that happens when a task was created with a `complete=True`. Then having to filter, oddly, on very low resolve time instead of being able to exclude via `created_is==completed_at`.

app/grandchallenge/challenges/models.py
chrisvanrun: Why?

We could have a 'hot' period for the complete where a user is not allowed to undo the complete if it's been marked complete for more than, say, a day?

app/grandchallenge/challenges/models.py
chrisvanrun: AFAIK It's common practice in the code base. Anytime a remove_perm is used in a `save() >> assign_permissions()` it is done so irrespenctive what if the permission is there or not. Mostly it's based on `publi`c states in other models.

I've given it the follow reason: given how important permissions are it makese sense: a minor performance hit is okay as to ensure it is always removed.

app/grandchallenge/challenges/models.py
amickan: Ah, got it! I'm still wondering why we would ever want to add tasks that are already completed, but if there is a use case for that, then I agree that it would be nice for those tasks to have `created==completed_at.`

app/grandchallenge/challenges/models.py
amickan: On second thought, I guess it's ok to allow admins to uncomplete a task in case they accidentally close it. I had a workflow where it's us doing to completion marking in mind, just like with challenge requests, and then you'd want to lock it down, but never mind!

app/grandchallenge/challenges/models.py
amickan: That's true, but then I would prefer an `if ... else ...` statement for clarity.

chrisvanrun: Good point, I'll just drop the `adding` check performance improvement.

app/grandchallenge/challenges/models.py
```diff
@@ -1402,3 +1403,84 @@ class ChallengeRequestGroupObjectPermission(GroupObjectPermissionBase):
...
         ChallengeRequest, on_delete=models.CASCADE
     )
+
+
+class TaskResponsiblePartyChoices(models.TextChoices):
+    SUPPORT = "SUP", "Support"
+    CHALLENGE_ORGANIZERS = "ORG", "Challenge Organizers"
+
+
+class OnboardingTask(FieldChangeMixin, UUIDModel):
```
amickan, line 1413: Are you using the FieldChangeMixin for anything at the moment?

amickan: Ah, yes you are!. Never mind.

amickan (dismissed):

chrisvanrun:
Sorry Anne, I saw I had missed another 'hidden' place for the rename.

amickan (approved):

PR #3797, opened 2025-01-23
Unique input sets per algorithm

amickan:
Adds restriction that interfaces for an algorithm need to have unique sets of inputs, discussed [here](https://github.com/DIAGNijmegen/rse-roadmap/issues/153#issuecomment-2602182907).
Also fixes an oversights from an earlier PR.

### DIFF
--- a/app/grandchallenge/algorithms/admin.py
+++ b/app/grandchallenge/algorithms/admin.py
@@ -263,15 +263,16 @@ class AlgorithmInterfaceAdmin(GuardedModelAdmin):
     def algorithm_outputs(self, obj):
         return oxford_comma(obj.outputs.all())
 
-    # interfaces should only be created through the UI
-    # and cannot be modified or deleted
     def has_change_permission(self, request, obj=None):
+        # interfaces cannot be modified
         return False
 
     def has_delete_permission(self, request, obj=None):
+        # interfaces cannot be deleted
         return False
 
     def has_add_permission(self, request, obj=None):
+        # interfaces should only be created through the UI
         return False
 
 
@@ -285,12 +286,12 @@ class AlgorithmAlgorithmInterfaceAdmin(GuardedModelAdmin):
     )
     list_filter = ("is_default", "algorithm")
 
-    # through table entries should only be created through the UI
-    # and cannot be modified
     def has_add_permission(self, request, obj=None):
+        # through table entries should only be created through the UI
         return False
 
     def has_change_permission(self, request, obj=None):
+        # through table entries should only be updated through the UI
         return False
 
 
--- a/app/grandchallenge/algorithms/admin.py
+++ b/app/grandchallenge/algorithms/admin.py
@@ -263,6 +263,8 @@ class AlgorithmInterfaceAdmin(GuardedModelAdmin):
     def algorithm_outputs(self, obj):
         return oxford_comma(obj.outputs.all())
 
+    # interfaces should only be created through the UI
+    # and cannot be modified or deleted
     def has_change_permission(self, request, obj=None):
         return False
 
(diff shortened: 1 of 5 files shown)


app/grandchallenge/algorithms/admin.py
```diff
@@ -283,6 +283,12 @@ class AlgorithmAlgorithmInterfaceAdmin(GuardedModelAdmin):
     )
     list_filter = ("is_default", "algorithm")
 
+    def has_add_permission(self, request, obj=None):
+        return False
+
+    def has_change_permission(self, request, obj=None):
+        return False
```
amickan, line 290: Locked down the admin actions for the through table since I don't think we need to or should be creating or changing them in the admin. They should always be created through the UI form, which has all the necessary checks in place. 

I left deleting an option for any possible bulk deletion requests.

app/grandchallenge/algorithms/forms.py
```diff
@@ -1401,17 +1401,37 @@ def clean_set_as_default(self):
...
         cleaned_data = super().clean()
 
         inputs = cleaned_data.get("inputs", [])
         outputs = cleaned_data.get("outputs", [])
 
-        if not inputs or not outputs:
-            raise ValidationError(
-                "You must provide at least 1 input and 1 output."
-            )
-
```
amickan, line 1414: As far as I can tell this check is unnecessary. The fields in the form are required and make it impossible to submit without at least one selected choice. Also, the create() method later on will raise an error if there are no inputs or outputs.

app/grandchallenge/algorithms/models.py
```diff
@@ -983,11 +983,15 @@ def get_jobs_with_same_inputs(
...
-                    "inputs", filter=Q(inputs__in=existing_civs)
-                )
+                input_count=Count("inputs", distinct=True),
+                input_match_count=Count(
+                    "inputs", filter=Q(inputs__in=existing_civs), distinct=True
+                ),
+            )
+            .filter(
+                input_count=input_interface_count,
+                input_match_count=input_interface_count,
```
amickan, line 993: Oversight in my last PR. James, I think you hinted at this with your comment about partially overlapping inputs, but I didn't grasp it back then and wrote an incomplete test for it. This fixes it, and I also added to the test to cover the case I had missed. 

We use a very similar annotation and filter pattern in 5 different places in the codebase now. I tried to look for a way to re-use it, but since it's on 5 different models with sometimes different field names I didn't find a way to move the logic to one place and then re-use it everywhere. If you have suggestions, I'm all ears. Otherwise, just being careful in the future and good coverage tests are the way to go for now.

app/tests/algorithms_tests/test_forms.py
```diff
@@ -1557,32 +1574,3 @@ def test_default_interface_for_algorithm_not_updated_when_adding_new_non_default
         assert new_io != io
         assert not new_iot.is_default
         assert old_iot.is_default
-
-    @pytest.mark.django_db
-    def test_is_default_is_updated_when_adding_an_already_existing_interface(
-        self,
-    ):
```
amickan, line 1564: With the new restriction, it is no longer possible to make a non-default interface the default interface of an algorithm through the create form. The user will need to delete the non-default interface first, and then add it again as default to achieve that. I think that is still a totally ok work-around, but this test here has to go.

jmsmkn (approved):
Looks good, it is probably worth putting comments like these into the code itself. I only understood the motivation here (the why) due to reading them, and a future developer wondering why certain things are forbidden will not have that benefit!

app/grandchallenge/algorithms/forms.py
amickan: The create() method actually doesn't raise an error, so I left this check in the form after all. I have also added a check to the create method to cover us from all sides.

amickan: The check for duplicates in the input-output combination is actually redundant now that we enforce uniqueness on inputs, so I have removed that check as well.

jmsmkn (approved):
Looks good, only I think the comments in admin.py should be in the function body otherwise they can get separated.

amickan:
> Looks good, only I think the comments in admin.py should be in the function body otherwise they can get separated.

That makes a lot of sense! Updated.

PR #3798, opened 2025-01-24
Create job with interface through API

amickan:
This adds input-to-interface matching to the JobPostSerializer so that algorithm jobs created through the API also have the interface configured. 
This also adds a serializer for algorithm interfaces.


amickan:
Wrong merge target, close in favour of: https://github.com/comic/grand-challenge.org/pull/3799

PR #3799, opened 2025-01-24
Create job with interface through API

amickan:
This adds input-to-interface matching to the JobPostSerializer so that algorithm jobs created through the API also have the interface configured.
This also adds a serializer for algorithm interfaces.


app/config/settings.py
```diff
@@ -1443,7 +1443,7 @@ def sentry_before_send(event, hint):
 # CONTENT SECURITY POLICY
 ##########################
 
-CSP_REPORT_ONLY = strtobool(os.environ.get("CSP_REPORT_ONLY", "False"))
+CSP_REPORT_ONLY = strtobool(os.environ.get("CSP_REPORT_ONLY", "True"))
```
chrisvanrun, line 1446: Rogue CSP danger!

amickan: Woopsy!! Thanks for catching that

app/grandchallenge/algorithms/serializers.py
```diff
@@ -107,6 +126,7 @@ class JobSerializer(serializers.ModelSerializer):
 
     algorithm_image = StringRelatedField()
 
+    algorithm_interface = AlgorithmInterfaceSerializer(read_only=True)
```
amickan, line 129: Unsure if the job serializers need the interface. It's of no use to the user, just added it for completeness really.

jmsmkn (approved):
LGTM, doesn't need another round of review.

PR #3803, opened 2025-01-29
Add algorithm interfaces to phases and migrate

amickan:
This adds a `algorithm_interfaces` M2M to the `Phase` model and adds a custom through model for it. 
It also removes the input / output configuration option from both the phase admin and the `ConfigureAlgorithmPhases` view.
In anticipation of PR 2, it refactors the interface create form so that it can be easily used for both phases and algorithms alike. 

Views to create and manage interfaces for phases will be added in a separate PR. 


app/grandchallenge/algorithms/migrations/0065_create_algorithm_interfaces.py
```diff
@@ -17,6 +17,10 @@ def create_algorithm_interfaces(apps, _schema_editor):
         inputs = algorithm.inputs.all()
         outputs = algorithm.outputs.all()
 
+        if not inputs or not outputs:
+            # Skip algorithms without inputs or outputs
+            continue
+
```
amickan, line 23: I realized that this checks is necessary since the migration cannot use the custom create method on the manager (where I added this check in the last PR).

app/grandchallenge/evaluation/models.py
```diff
@@ -1198,6 +1204,23 @@ def parent_phase_choices(self):
             )
         )
 
+    @cached_property
+    def default_interface(self):
+        try:
+            return self.algorithm_interfaces.get(
+                phasealgorithminterface__is_default=True
+            )
+        except ObjectDoesNotExist:
```
amickan, line 1213: For evaluations, we already have a property called `default_interfaces` that refers to the evaluation's inputs and outputs, not the algorithm inputs and outputs. I think it might be better to update `default_interface` to `default_algorithm_interface`. I would like to use this in the AlgorithmInterfaceForm, and so need it to be the same as the property on the algorithm. 

So I suggest to change it in both places to `default_algorithm_interface`. That property is used in multiple places (including about 50 tests), so I suggest to make the naming change in a separate PR after this one is merged.

app/grandchallenge/algorithms/migrations/0065_create_algorithm_interfaces.py
jmsmkn: Do we have such algorithms? Would be very strange if so, I think it is/should be forbidden.

app/grandchallenge/evaluation/models.py
```diff
@@ -1207,6 +1230,33 @@ class PhaseGroupObjectPermission(GroupObjectPermissionBase):
     content_object = models.ForeignKey(Phase, on_delete=models.CASCADE)
 
 
+class PhaseAlgorithmInterface(models.Model):
+    phase = models.ForeignKey(Phase, on_delete=models.CASCADE)
+    interface = models.ForeignKey(AlgorithmInterface, on_delete=models.CASCADE)
+    is_default = models.BooleanField(default=False)
```
jmsmkn, line 1236: Why do we need a default algorithm interface for Phases? I thought this was only used to select a default for the "Try out algorithm" page. Do we even use it there?

app/grandchallenge/evaluation/migrations/0073_migrate_interfaces_for_phases.py
```diff
@@ -0,0 +1,52 @@
...
+        Phase.objects.filter(submission_kind=SubmissionKindChoices.ALGORITHM)
+        .prefetch_related("algorithm_inputs", "algorithm_outputs")
+        .all()
+    ):
+        inputs = phase.algorithm_inputs.all()
+        outputs = phase.algorithm_outputs.all()
+
+        if not inputs or not outputs:
+            # Skip phases without inputs or outputs
+            continue
```
jmsmkn, line 25: I am unsure if this is necessary.

jmsmkn (commented):
Looks good, reading this just made me think about why we need the `is_default` field. I don't think that we really use it anywhere, other to enforce that there is at least one interface per algorithm in a few places.

app/grandchallenge/algorithms/migrations/0065_create_algorithm_interfaces.py
amickan: We should not, but better to check. For phases, this can happen since those used to be configured in the admin and there was no check for whether inputs and outputs are both present when converting to algorithm submission type.

app/grandchallenge/evaluation/migrations/0073_migrate_interfaces_for_phases.py
jmsmkn: But that would mean that there is a borked configuration in the database, so should be fixed before migrating rather than just silently skipping over it, no?

amickan: Not sure. It could be phases that are in the middle of set-up - the BigPicture PoC for example. I think it's fine to silently skip over these since we are the ones to configure the phases (not the challenge admins) and if they are not fully set-up yet, we can do it later from the UI rather than postponing the migration until the onboarding is complete, right? 

We could do a quick check in the shell to see how many phases this concerns. In theory, it should not happen and both inputs and outputs should be set-up at the same time, but it is theoretically possible (see DemoBP).

jmsmkn: In the next migration you will presumably remove the old fields though, so that information would anyway be lost.

amickan: Discussed this synchronously, I will raise an error here rather than continue. I have also added a todo item to the pitch to check if there are any algorithms or phases with incomplete set-up.

app/grandchallenge/evaluation/models.py
amickan: Removed

app/grandchallenge/evaluation/models.py
amickan: Removed

amickan:
I have removed the `is_default` field from the through model and now raise a RuntimeError during the migration when a phase or algorithm are not properly configured. I will now merge this into the feature branch.

PR #3806, opened 2025-01-30
Cleanup flexible image adjustments

koopmant:
some refactoring in preparation for implementing flexible file widget. 

related to https://github.com/DIAGNijmegen/rse-roadmap/issues/363


app/grandchallenge/components/form_fields.py
```diff
@@ -40,36 +37,22 @@ def _join_with_br(a, b):
 INTERFACE_FORM_FIELD_PREFIX = "__INTERFACE_FIELD__"
 
 
-class InterfaceFormField:
+class InterfaceFormField(forms.Field):
```
koopmant, line 40: I've added the parent class here because the django docs say we should and I haven´t seen a comment indicating why this was not done when this was first implemented. So I assumed it was just overlooked and I'm relying on the tests to ensure that things keep working...

app/grandchallenge/cases/widgets.py
```diff
@@ -51,7 +50,7 @@ def __init__(
     ):
         widgets = (
             ImageSearchWidget(),
-            UserUploadMultipleWidget(),
+            UserUploadSingleWidget(),
```
amickan, line 53: For images, we might need to keep the `UserUploadMultipleWidget`, but I'm not fully sure. @jmsmkn do you know?

app/grandchallenge/cases/views.py
```diff
@@ -126,11 +126,11 @@ def get(self, request, *args, **kwargs):
...
             )
             return HttpResponse(html_content)
-        elif widget_name == WidgetChoices.IMAGE_UPLOAD.name:
+        elif widget_choice == ImageWidgetChoices.IMAGE_UPLOAD.name:
             html_content = render_to_string(
-                UserUploadMultipleWidget.template_name,
+                UserUploadSingleWidget.template_name,
                 {
-                    "widget": UserUploadMultipleWidget().get_context(
+                    "widget": UserUploadSingleWidget().get_context(
```
amickan, line 133: Same here. We need to keep `UserUploadMultipleWidget` for images.

app/grandchallenge/components/form_fields.py
amickan: Have you also tested in the UI yourself?

amickan: It looks good to me, but some thorough testing would be good ;-)

amickan:
LGTM, except for the change from `UserUploadMultiple` to `UserUploadSingle` - that change should be reverted, and that will most likely also fix the tests that are currently failing.

amickan (approved):
LGTM

app/grandchallenge/components/form_fields.py
koopmant: Also tested manually; everything seems to still work just fine.

jmsmkn (approved):
LGTM!

PR #3807, opened 2025-01-30
Flexible file widget

koopmant:
Add the flexible file widget. The widget is not yet used on any form.


app/grandchallenge/components/views.py
```diff
@@ -472,3 +494,113 @@ def get(self, request, *args, **kwargs):
...
+            )
+            return HttpResponse(html_content)
+        elif (
+            widget_choice == FileWidgetChoices.FILE_SELECTED.name
+            and current_value
+            and (
+                ComponentInterfaceValue.objects.filter(
+                    pk=current_value
+                ).exists()
+                or UserUpload.objects.filter(pk=current_value).exists()
```
koopmant, line 545: The test is are currently failing due to this section. I took this from the `ImageWidgetSelectView` where instead of `ComponentInterfaceValue` we use `Image` objects. That works fine, because both `Image` and `UserUpload` are both derived from `UUIDModel`. `ComponentInterfaceValue` is not and it's pk is `id` which is an integer field; not a uuid. 
I think I can do something similar to what I saw in the code history: checking if the current_value is an integer and then using it for the appropriate filter. Not sure if there is a more elegant solution here...

koopmant: @jmsmkn
@amickan and I discussed this and we agreed that some type checking is fine to fix this issue. We we're wondering if `ComponentInterfaceValue` should become a `UUIDModel` at some point? Currently its pk is the `id` attribute which is a `BigAutoField`, (hence the need for type checking). If it would be a uuid that would be more consistent with other models. Changing it would require a data migration. Are there other implications, upsides and downsides? If it's worth the effort I could create a separate PR for it.

app/grandchallenge/cases/widgets.py
```diff
@@ -85,12 +85,7 @@ def value_from_datadict(self, data, files, name):
             except KeyError:
                 value = None
         if value:
-            if value in ImageWidgetChoices.names:
-                return [None, None]
-            elif Image.objects.filter(pk=value).exists():
-                return [value, None]
-            else:
-                return [None, [value]]
+            return self.decompress(value)
```
koopmant, line 89: Some more refactoring. Not strictly part of this PR, but I already merged the other one.

koopmant:
@amickan do you think these tests are enough? I've adapted existing tests from the `FlexibleImageWidget` but I'm not sure if I'm missing something obvious. I'll take another look at it tomorrow.

app/grandchallenge/cases/widgets.py
amickan: Is the `if value` necessary here? It's also checked in `decompress`. What needs to be returned here if `value=None`?

app/grandchallenge/components/static/components/js/flexible_file_widget.js
```diff
@@ -0,0 +1,8 @@
+$(document).ready(() => {
+    const elements = document.querySelectorAll("[id^=widgetSelect]");
+    for (let i = 0, len = elements.length; i < len; i++) {
+        if (elements[i].value === "FILE_SEARCH") {
+            htmx.trigger(elements[i], "widgetSelected");
+        }
+    }
+});
```
amickan, line 8: Could you turn this and `flexible_image_widget.js` into a module (`.mjs`)?

app/grandchallenge/components/templates/components/file_search_widget.html
```diff
@@ -0,0 +1,38 @@
...
+<div class="form-group mb-1"
+     id="search-file-{{ widget.name }}"
+     hx-get="{% url 'components:file-search' %}"
+     hx-trigger="load once, keyup[target.value.length > 1] delay:500ms from:#query-{{ widget.name }}, click[keyCode==8] delay:500ms from:#query-{{ widget.name }}"
+     hx-target="#selected-file-{{ widget.name }}"
+     hx-swap="outerHTML"
+     hx-include="[id='file-input-group-{{ widget.name }}']"
+>
+    <div class="input-group" id="file-input-group-{{ widget.name }}">
+        <div class="input-group-prepend"><span class="input-group-text">Select an file</span></div>
```
amickan, line 14: Small typo
```suggestion
        <div class="input-group-prepend"><span class="input-group-text">Select a file</span></div>

```

app/grandchallenge/components/urls.py
```diff
@@ -59,4 +61,10 @@
         FileUploadFormFieldView.as_view(),
         name="file-upload",
     ),
+    path(
+        "select-file-widget",
```
amickan, line 65:
```suggestion
        "select-file-widget/",
```

app/grandchallenge/components/views.py
```diff
@@ -472,3 +494,114 @@ def get(self, request, *args, **kwargs):
...
         )
         return HttpResponse(html_content)
+
+
+class FileWidgetSelectView(LoginRequiredMixin, View):
+    def get(self, request, *args, **kwargs):
+        interface = request.GET.get("interface-slug")
+        widget_choice = request.GET.get(f"widget-choice-{interface}")
+        help_text = request.GET.get("help-text")
+        current_value = request.GET.get("current-value")
```
amickan, line 504: Noticing here that you changed the names of the input fields from underscore to dash. Is there a specific reason for doing this? In general, I'm wondering whether the two widgets could make use of the same base templates to avoid duplication of the same logic. It would require adding some widget-specific context variables to the views, but should otherwise be possible, right?

amickan: On second thought, refactoring to use the same templates might break once you add the extra filter option for the FileWidget though

app/grandchallenge/components/views.py
```diff
@@ -472,3 +494,114 @@ def get(self, request, *args, **kwargs):
...
+            # this happens when switching back from one of the
+            # above widgets to the "Choose data source" option
+            return HttpResponse()
+        else:
+            raise RuntimeError("Unknown widget type")
+
+
+class FileSearchResultView(LoginRequiredMixin, ListView):
+    template_name = "components/file_search_result_select.html"
+    search_fields = ["pk", "name"]
```
amickan, line 575: The `ComponentInterfaceValue` model does not have a `name` field. What you probably want is the `file__name`.

app/grandchallenge/components/views.py
```diff
@@ -472,3 +494,114 @@ def get(self, request, *args, **kwargs):
...
+        qs = self.get_queryset()
+        interface = request.GET.get("interface_slug")
+        query = request.GET.get("query-" + interface)
+        if query:
+            q = reduce(
+                or_,
+                [Q(**{f"{f}__icontains": query}) for f in self.search_fields],
+                Q(),
+            )
+            qs = qs.filter(q).order_by("name")
```
amickan, line 599: I think this also needs changing: 
```suggestion
            qs = qs.filter(q).order_by("file__name")

```

app/grandchallenge/components/widgets.py
```diff
@@ -1,4 +1,86 @@
...
+        try:
+            value = data[name]
+        except KeyError:
+            # this happens if the data comes from the DS create / update form
+            try:
+                value = data[f"widget-choice-{name}"]
+            except KeyError:
+                value = None
+        if value:
+            return self.decompress(value)
```
amickan, line 83: Same question as above, could this just be
```suggestion
      return self.decompress(value)
```

amickan:
> @amickan do you think these tests are enough? I've adapted existing tests from the `FlexibleImageWidget` but I'm not sure if I'm missing something obvious. I'll take another look at it tomorrow.

The tests look comprehensive to me. The only thing that might be worth an additional test is the search view (see also the comment I left about the search fields). I don't think we have a test for the image search view, maybe they can be combined?

app/grandchallenge/components/views.py
amickan: Thinking about the view design here, maybe it would be good to add some checks on whether the interface  exists and fail early if not: 

```
@cached_property
 def interface(self):
       return get_object_or_404(
            ComponentInterface, slug=self.request.GET.get("interface-slug") # why not use an underscore here?
       )
```

amickan: You could add a similar check for the widget choice 

```
@property
def widget_choice(self):
     try:
          return FileWidgetChoices(self.request.GET.get("widget-choice-(self.interface.slug)"))
     except ValueError:
         return Http404()
```

But you still need to check the widget_choice in `get()`, so not sure if this adds anything

amickan: And thinking about permission, currently, anyway can visit this view. I think it should be locked down to users who have access to (any) files on GC, similar to how we do it for `FileUploadFormFieldView`.

app/grandchallenge/components/views.py
```diff
@@ -472,3 +494,114 @@ def get(self, request, *args, **kwargs):
...
+            return HttpResponse(html_content)
+        elif widget_choice == FileWidgetChoices.UNDEFINED.name:
+            # this happens when switching back from one of the
+            # above widgets to the "Choose data source" option
+            return HttpResponse()
+        else:
+            raise RuntimeError("Unknown widget type")
+
+
+class FileSearchResultView(LoginRequiredMixin, ListView):
```
amickan, line 573: Thinking about permissions, currently, anyone can visit this view. I think it should be locked down to users who have access to (any) files on GC, similar to how we do it for `FileUploadFormFieldView`.

app/grandchallenge/cases/widgets.py
koopmant: I initially did not check `if value` but the return value then becomes `[None, None]`. To be honest, I'm not sure if that's important? In the test it was explicitly checked that if the value is `None` then the return value should not be decompressed; so I added the check again to comply with the test...

app/grandchallenge/components/views.py
koopmant: > you changed the names of the input fields from underscore to dash. Is there a specific reason for doing this?  

Not really, I initially started changing them to make sense of what was what and I found them dificult to read. I guess the mix of upper camelcase and lower camelcase combined with delimiter separation just triggered me :sweat_smile:  
I then tried to change everything to hyphen separated so I could more easily recognize when things were coming from the request data, and where it is used in the templates. 

> refactoring to use the same templates might break ...

Yes, I've just been updating the templates after some testing and there are some small differences already. After implementing the extra filter option we'll have to see. I'll continue with separate templates first and will take a look if I can reduce duplication when the work is done.

koopmant: > maybe it would be good to add some checks on whether the interface exists and fail early if not

Thanks, I'll look into this!

app/grandchallenge/components/views.py
koopmant: Yes, I found out this morning that this does not work :sweat_smile: 
I tried file__name first, but that did not work either because of the following error:
> django.core.exceptions.FieldError: Unsupported lookup 'name' for FileField or join on the field not permitted.

I therefore went with "file". That works, but it does include the complete path to the file, so I'm not sure of that's a potential problem.

app/grandchallenge/components/views.py
koopmant: Same error as mentioned above. However, ordering by "file" leads to actually ordering by pk, because the pk is in the file path... I have been trying to find a solution for this, but haven't found anything yet.

app/grandchallenge/components/templates/components/file_search_widget.html
koopmant: Fixed, thanks!

app/grandchallenge/cases/widgets.py
koopmant: Removed it.

app/grandchallenge/components/views.py
koopmant: Implemented checking for the interface existence and valid widget choice for failing early. I did not use a cached property because I only check it once in the get request. I also renamed data in the requests to `prefixed-interface-slug` for clarity.

app/grandchallenge/components/views.py
amickan: Ah of course, I think `file` is fine.

app/grandchallenge/components/views.py
amickan: If you really wanted to, you could probably annotate the queryset with the filename and then filter on that, but I honestly think ordering by pk is fine for now.

app/grandchallenge/cases/templates/cases/flexible_image_widget.html
```diff
@@ -25,4 +25,4 @@
 
 <div class="mt-3" id="area-{{ widget.name }}"></div>
 
-<script type="text/javascript" src="{% static "cases/js/flexible_image_widget.js" %}"></script>
+<script type="text/javascript" src="{% static "cases/js/flexible_image_widget.mjs" %}"></script>
```
amickan, line 28:
```suggestion
<script type="module" src="{% static "cases/js/flexible_image_widget.mjs" %}"></script>

```

app/grandchallenge/cases/widgets.py
amickan: Ah ok, well, I'm not sure either, but I don't see the value shouldn't be decompressed if it's `None`. so removing it sounds good to me.

app/grandchallenge/cases/templates/cases/flexible_image_widget.html
koopmant: thank you :smiling_face_with_tear:

app/grandchallenge/components/views.py
koopmant: Moved discussion about `ComponentInterfaceValue` deriving from `UUIDModel` to the pitch https://github.com/DIAGNijmegen/rse-roadmap/issues/363#issuecomment-2631389409

app/grandchallenge/components/views.py
koopmant: I've been wrecking my brain for a little while to try an annotate with the filename but I could not figure it out. I'll simply leave it at ordering by pk.

koopmant:
> The tests look comprehensive to me. The only thing that might be worth an additional test is the search view (see also the comment I left about the search fields). I don't think we have a test for the image search view, maybe they can be combined?

We do have a test for the image search view, just found it after writing a test for the file search results view. I did not combine at the moment... I can pick that up when I will reduce the duplication between the two.

app/grandchallenge/components/views.py
```diff
@@ -429,7 +452,7 @@ def form_valid(self, form):
         return super().form_valid(form)
 
 
-class FileUploadFormFieldView(LoginRequiredMixin, AccessMixin, View):
+class FileAccessRequiredMixin:
```
amickan, line 455: I think this should subclasses `AccessMixin`. Then `FileUploadFormFieldView` and the other new view only need to inherit from `FileAccessRequiredMixin` and no longer `AccessMixin`.

app/grandchallenge/components/widgets.py
```diff
@@ -1,4 +1,91 @@
...
+
+    def decompress(self, value):
+        if value:
+            if value in FileWidgetChoices.names:
+                return [None, None]
+            else:
+                try:
+                    qs = ComponentInterfaceValue.objects.filter(pk=value)
+                except ValueError as e:
+                    if "Field 'id' expected a number but got" not in str(e):
```
amickan, line 68: This feels a little fragile to me. What if the error message changes in the future? Wouldn't it be better to explicitly try to convert `value` into a UUID and check if that fails, and choose the model type accordingly?

amickan: Ah wait, here you need a number, not a uuid. Then convert to a number?

app/grandchallenge/components/views.py
```diff
@@ -472,3 +500,134 @@ def get(self, request, *args, **kwargs):
...
+                return HttpResponse(html_content)
+            case FileWidgetChoices.FILE_SELECTED:
+                try:
+                    qs = UserUpload.objects.filter(pk=current_value)
+                except ValidationError as e:
+                    if "uuid" not in e.message.lower():
+                        raise
+                    qs = ComponentInterfaceValue.objects.filter(
+                        pk=current_value
+                    )
```
amickan, line 567: See my other comment. Why not try to convert `current_value` to a uuid and choose the model according to whether that succeeds?

app/grandchallenge/components/widgets.py
amickan: Wait I'm confused by the logic here. If `value` is a uuid you return `[None, value]` without checking whether it's an existing UserUpload? Or am I reading this wrong? I think it might be clearer if you try to convert to uuid, and then check for a UserUpload if that works, or for a CIV if it doesn't work.

app/grandchallenge/components/views.py
koopmant: Cool, I'll do that.

app/grandchallenge/components/widgets.py
koopmant: Yeah, it's not great. I was trying to avoid having to check the type of the pk, but I guess that is inherently still what I'm doing... I'll update it and add a comment explaing the intention, because I'm afraid it will remain a bit confusing. 
Currently trying to understand what types I should expect/allow here, because in [the tests](https://github.com/comic/grand-challenge.org/actions/runs/13143629458/job/36676464320) this code is sometimes presented with a list (`['abcd-some-uuid-pk']`)

amickan: The list is because the Image widget uses the multiple-uploads subwidget, which returns a list. In this case, the value should always be a string or None as far as I understand it.

koopmant: I've reverted it to type checking the pk. I've been trying a different route yesterday and I'm close to an alternative, but it's taking me a very long time to get it working. I suggest to stick with this for now, I can try out the alternative when refactoring in the end if there is time.

app/grandchallenge/components/views.py
koopmant: Currently checking for integer type first. When it's not, I'm assuming it must be a uuid and check if a user upload object exists. I could do it the other way around with:
```python
try:
    UUID(current_value)
except ValueError:
    # assume it's an integer type pk
    # check if civ object exists
else:
    # check if user upload object exists
```

amickan: Doesn't matter which way around you do it I guess. We know that the value should either be a user upload pk or a civ pk, so maybe it's worth checking for both explicitly, and else log an error to sentry (without raising the error for the user, the decompressed value will just be [None, None] in that case). What do you think? 

To log the error to sentry without raising it, you could do `logger.error(e, exc_info=True)`.

koopmant: That's great! I did not realize we could log errors like that
I tried to incorporate it. Could you check if I did it right?

amickan: I take it back. I was under the wrong impression that this was about the decompress() function on the widget and not the view. In the view, we should raise the error or else the user doesn't get feedback that something went wrong and might keep trying and trying, DOSing the site.

amickan (approved):
LGTM

app/grandchallenge/components/views.py
```diff
@@ -472,3 +499,132 @@ def get(self, request, *args, **kwargs):
...
+        )
+        try:
+            widget_choice = FileWidgetChoices(widget_choice_name)
+        except ValueError:
+            raise Http404(f"Widget choice {widget_choice_name} not found")
+        help_text = request.GET.get("help-text")
+        current_value = request.GET.get("current-value")
+
+        match widget_choice:
+            case FileWidgetChoices.FILE_SEARCH:
```
jmsmkn, line 526: Style wise I think we should only use `match`-`case` if an `if`-`elif` will do, which in this case it is as it is only doing a simple comparison each time. The rebinding problem with `match`-`case` is too easy to trigger and can cause serious bugs.

app/grandchallenge/components/views.py
```diff
@@ -472,3 +499,132 @@ def get(self, request, *args, **kwargs):
...
+                    },
+                )
+                return HttpResponse(html_content)
+            case FileWidgetChoices.FILE_SELECTED:
+                if current_value and (
+                    ComponentInterfaceValue.objects.filter(
+                        pk=current_value
+                    ).exists()
+                    if current_value.isdigit()
+                    else UserUpload.objects.filter(pk=current_value).exists()
```
jmsmkn, line 564: Shouldn't the permissions be checked here? If the user does not have permission to view the object it should still return 404 otherwise information about the existence of the object will leak.

koopmant: Fixing this in #3833

PR #3808, opened 2025-01-30
Use title instead of slug for label interface widgets

chrisvanrun:
There is a minor typo in a slug. Imho it should be the title anyhow here so this saves us some minor time on redoing the interface:

Screenshot 2025-01-30 at 12 16 44


amickan (approved):
LGTM

PR #3812, opened 2025-01-31
Managements views for algorithm interfaces for phases

amickan:
This adds the management views for algorithm interfaces for phases. 
The views are only accessible to staff users, and only for algorithm submission phases that are not external. 
The logic behind the views is the same as for the algorithm interfaces, but it only made sense to refactor the create view logic, and parts of the templates. 

image


app/config/urls/challenge_subdomain.py
```diff
@@ -15,6 +15,10 @@
         ),
         name="subdomain_robots_txt",
     ),
+    path(
+        "components/",
+        include("grandchallenge.components.urls", namespace="components"),
+    ),
```
amickan, line 21: The interface create form links to the component overview pages. In order to make that possible for the phase views, the subdomain url config needs to be adjusted, or so I think. Is there a better way of solving this?

jmsmkn: I think this is the correct thing to do - this is why we have `markdownx/` here.

app/grandchallenge/evaluation/urls.py
```diff
@@ -89,6 +92,21 @@
...
+        AlgorithmInterfacesForPhaseList.as_view(),
+        name="interface-list",
+    ),
+    path(
+        "<slug>/interfaces/create/",
+        AlgorithmInterfaceForPhaseCreate.as_view(),
+        name="interface-create",
+    ),
+    path(
+        "<slug>/interfaces/<uuid:interface_pk>/delete/",
```
jmsmkn, line 106: Nit: shouldn't this be `<slug:slug>` throughout?

app/grandchallenge/evaluation/views/__init__.py
```diff
@@ -1252,3 +1255,94 @@ def get_object(self, queryset=None):
...
+
+
+class AlgorithmInterfaceForPhasePermissionMixin(AccessMixin):
+    @property
+    def phase(self):
+        return get_object_or_404(
+            Phase,
+            challenge=self.request.challenge,
+            challenge__phase__submission_kind=SubmissionKindChoices.ALGORITHM,
+            challenge__phase__external_evaluation=False,
```
jmsmkn, line 1267: Why are external evaluations excluded? I thought the interfaces had to match the parent phase for those?

app/grandchallenge/evaluation/views/__init__.py
```diff
@@ -1252,3 +1255,94 @@ def get_object(self, queryset=None):
...
+            slug=self.kwargs["slug"],
+        )
+
+    def dispatch(self, request, *args, **kwargs):
+        if request.user.is_staff:
```
jmsmkn, line 1272: I think we want a finer grained permission for this as probably not all staff users should be able to do it.

app/grandchallenge/pages/templates/pages/phase_menu_sidebar.html
```diff
@@ -37,6 +37,14 @@
                         <i class="fas fa-database fa-fw"></i> Linked archive
                     </a>
                 </li>
+                {% if request.user.is_staff %}
```
jmsmkn, line 40: Here too.

jmsmkn (changes requested):
Looks good! Just some small comments.

app/grandchallenge/evaluation/views/__init__.py
amickan: Ok, fair point. We could use the existing `configure_algorithm_phase` permission for that. This permission is currently assigned to the RSE group in the Django admin. Or do you think this requires a new permission just for Chris, Miriam, you and me?

amickan: Using `configure_algorithm_phase` makes sense to me given that this is the permission that is required to change a phase into an algorrithm phase as well. This is a related action I would say.

app/grandchallenge/evaluation/views/__init__.py
amickan: Oh you're right! I was under the impression that we shouldn't be configuring interfaces for external phases, but it's crucial that we can given that they need to match those of the parent phase.

app/grandchallenge/evaluation/urls.py
amickan: Sure. We don't do this anywhere else in the evaluation urls. Should I update all of them once I'm here?

app/grandchallenge/evaluation/views/__init__.py
jmsmkn: That sounds like the perfect one!

app/grandchallenge/evaluation/urls.py
amickan: Done!

PR #3817, opened 2025-02-04
Add html user completion for challenge admin users

chrisvanrun:
Was annoying not to be able to check if I had the correct user in the workshop challenge add admin page.

## Before

image

## After

image


jmsmkn (changes requested):
Whilst we're here to future proof this and remove duplication could the `AdminsForm` be updated to use `UserGroupForm`? Rest looks good!

chrisvanrun:
Bit of a rabbit hole with origins, login detection etc. Side stepped it by having an within domain end-point for user-completion. Much like it was the case already but now implemented via the other view.

You don't need permissions to autocomplete users imho.

jmsmkn (approved):
LGTM!

PR #3820, opened 2025-02-06
Enable submission with optional inputs

amickan:


amickan:
I still need to expand the tests to cover the case of multiple interfaces defined for a phase or algorithm, but otherwise this is ready for review, and I would appreciate if you could already take a look, if you find time, @jmsmkn

app/grandchallenge/algorithms/tasks.py
```diff
@@ -163,105 +160,134 @@ def create_algorithm_jobs(
...
+                        for civ in item.values.all()
+                    }
                 ),
-            )
+                distinct=True,
+            ),
+        )
+        .filter(
+            input_count=F("inputs_match_count"),
+            input_count__in=valid_input_counts,
```
jmsmkn, line 273: I have a feeling this isn't quite right, don't you need to limit it to the count of the number of inputs for the one relevant interface, not all interfaces? So you would then need to find:

- What interface does this job match?
- How many inputs are there for this interface?
- Does the number of inputs match the number of inputs for this interface?

You could do this slowly by taking the original query and then iterating over each interface.

app/grandchallenge/evaluation/models.py
```diff
@@ -1695,23 +1702,9 @@ def successful_jobs(self):
...
-                            civ
-                            for civ_set in self.valid_archive_item_values
-                            for civ in civ_set
-                        }
-                    ),
-                ),
-            )
-            .filter(
-                inputs_match_count=self.algorithm_inputs.count(),
+                algorithm_interface__in=self.submission.phase.algorithm_interfaces.all(),
```
jmsmkn, line 1705: This is only checking the interfaces and not their values? What happens when jobs are created for that algorithm through the API?

jmsmkn (commented):
Looks good, this is quite complex! We should review things very carefully each time as we know this scheduling is problematic.

app/grandchallenge/algorithms/tasks.py
```diff
@@ -116,7 +113,7 @@ def create_algorithm_jobs_for_archive(
 def create_algorithm_jobs(
     *,
     algorithm_image,
-    civ_sets,
+    archive_items,
```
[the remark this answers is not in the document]
amickan: I wondered the same. This is only used in 2 places, and in both those places we are going from archive items to civ_sets. It made more sense to me to just pass through the archive item queryset. Then I can annotate on it. 

I think this function might have been used in other places in the past, where no archive items were involved, but I'm not sure.

app/grandchallenge/algorithms/tasks.py
```diff
@@ -163,105 +160,134 @@ def create_algorithm_jobs(
...
-                input_civ_set=civ_set,
-            )
-            on_commit(job.execute)
+    for interface, archive_items in valid_job_inputs.items():
+        for ai in archive_items:
+            if max_jobs is not None and len(jobs) >= max_jobs:
+                break
+
+            if len(jobs) >= settings.ALGORITHMS_JOB_BATCH_LIMIT:
+                raise TooManyJobsScheduled
```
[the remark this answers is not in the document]
amickan: One should raise an exception so we retry the task, the other should just stop the loop and not retry the task. So I don't think they can be unified, right?

app/grandchallenge/evaluation/models.py
amickan: Yes, this is not correct. I just realized when adding more tests. It's actually quite complex. I need to return the successful jobs per interface, making sure to check the actual values, and then check that against `valid_archive_item_count_per_interface`.

app/grandchallenge/algorithms/tasks.py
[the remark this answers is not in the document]
jmsmkn: Okay, clear! Can you add those as comments for future developers?

app/grandchallenge/components/models.py
```diff
@@ -1339,7 +1339,10 @@ def title(self):
 
     @property
     def has_value(self):
-        return self.value is not None or self.image or self.file
+        if self.interface.value_required:
+            return self.value is not None or self.image or self.file
+        else:
+            return True
```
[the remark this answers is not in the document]
amickan: I changed this because the value can be none for interfaces of kind `ANY`, but you're right about the cornercase with booleans, so this needs to change again. 

It's anyway a bit strange that we allow some json type inputs to remain empty. What's the reason for that again?

app/grandchallenge/algorithms/tasks.py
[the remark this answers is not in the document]
jmsmkn: I seem to remember algorithm jobs going this way before the API and HTML Views were unified, unsure, but okay!

app/grandchallenge/algorithms/tasks.py
amickan: Hm, you might be right. It's similar to the `successful_jobs` property further below.

amickan:
> Looks good, this is quite complex! We should review things very carefully each time as we know this scheduling is sensitive.

Thanks for thinking along James! I needed a second brain to look at this. I'll work on updating the logic and the tests and hope that I have an updated PR ready by Monday.

app/grandchallenge/components/models.py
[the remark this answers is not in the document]
jmsmkn: I think that it is only the case if `None` is allowed: https://github.com/comic/grand-challenge.org/blob/18b50ce644e3e1ada866fe6bfe31c069aa65a674/app/grandchallenge/components/models.py#L703

app/grandchallenge/algorithms/tasks.py
amickan: Regarding "What interface does this job match?" - the job has an interface set, but you're suggesting that we should not trust the a job is actually valid for the interface that is set on the job, and hence double check here? 

I think that is indeed necessary given how we decided to migrate the interfaces for existing jobs. I think then I should first subset to jobs that have inputs that conform with their configured interfaces, and only then do the input comparison with the archive items.

app/grandchallenge/components/models.py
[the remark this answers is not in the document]
amickan: Exactly, but it is allowed sometimes, and for these cases the inputs are still complete even if the input is None. Maybe the place to update the logic is in `inputs_complete` and not `has_value`.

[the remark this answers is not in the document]
amickan: I have reverted the change here and added an extra condition to `inputs_complete` on the algorithm model together with a comment.

app/grandchallenge/algorithms/tasks.py
amickan: I have updated the logic here and in the model property. They now use the same underlying code to get valid jobs per interface

app/grandchallenge/evaluation/models.py
```diff
@@ -140,6 +140,70 @@
...
+                    distinct=True,
+                ),
+            )
+            .filter(
+                # first filter out jobs whose inputs don't match its interface,
+                # this can happen because we migrated assuming each job for
+                # an algorithm was done with the latest sets of inputs/outputs
+                input_count=input_count,
+                input_interface_match_count=input_count,
+            )
```
amickan, line 186: This might not be necessary. It's only relevant as long as there are jobs whose inputs don't match their interface. For jobs created from now on, this should never happen, so as time goes by we will run into this less and less. For now, it reduces the number of jobs to iterate over in the for-loop below and hence might help with speed, but the annotation also costs time of course, possibly resulting in net 0 difference. We should benchmark this. I will put together some queries for comparison.

app/grandchallenge/evaluation/models.py
```diff
@@ -140,6 +140,70 @@
...
+            # subset to jobs whose input set exactly matches
+            # one of the valid archive items' value sets
+            list_of_job_inputs = list(job.inputs.values_list("pk", flat=True))
+            list_of_archive_item_value_sets = [
+                list(item.values.values_list("pk", flat=True))
+                for item in valid_archive_items_per_interface[interface]
+            ]
+
+            if list_of_job_inputs in list_of_archive_item_value_sets:
+                jobs_per_interface[interface].append(job)
```
amickan, line 202: I tried doing this with an annotation, but I don't think that's possible. We used to do the following annotation:
```python
.annotate(
    inputs_match_count=Count(
        "inputs",
         filter=Q(
              inputs__in={
                   civ
                   for civ_set in self.valid_archive_item_values
                   for civ in civ_set
               }
        ),
   ),
)
```

But this would match a job that ran with a combination of values from different archive items. To exclude those jobs, we need to check for an exact match. I tried doing what I do in python above with an ArrayAgg annotation, but that requires extra type conversions, and got pretty complex pretty fast, so in the interest of readability and clarity I opted to filter in python. 

This also needs to be benchmarked, and I will put together a query for that shortly.

amickan:
This is ready for another review.

app/grandchallenge/evaluation/models.py
amickan: This is no more expensive than the annotation we currently do, so needs no extra benchmarking I think.

jmsmkn:
> This is ready for another review.

Given the discussions on the pitch I don't know what is still relevant or not. It's probably good to go through everything here and check for things like ensuring that set lookups are used rather than list lookups, that things are being correctly selected/prefetched as necessary, that sets of sets are only being constructed once and not many times in a for loop. I can see occurrences of those here but probably you know about them from the other discussions.

app/grandchallenge/evaluation/models.py
jmsmkn: I think the benchmarking showed that for ~1000 items and jobs that doing this in python is fine. I don't know how it will be with larger challenges.

amickan: I agree. Time will tell, I guess.

amickan:
> > This is ready for another review.
> 
> Given the discussions on the pitch I don't know what is still relevant or not. It's probably good to go through everything here and check for things like ensuring that set lookups are used rather than list lookups, that things are being correctly selected/prefetched as necessary, that sets of sets are only being constructed once and not many times in a for loop. I can see occurrences of those here but probably you know about them from the other discussions.

Indeed. I have started working on updating this PR, and will push changes tomorrow and let you know once it's ready.

app/grandchallenge/evaluation/models.py
jmsmkn: Do we know how many UNICORN will have? It would be better to find out what the performance would be now rather than in August 🙈

amickan: No idea, maybe @miriam-groeneveld does?

amickan:
Ready for review again.

app/grandchallenge/algorithms/tasks.py
```diff
@@ -163,105 +159,137 @@ def create_algorithm_jobs(
...
-        civ_interfaces = {civ.interface for civ in civ_set}
-        if input_interfaces.issubset(civ_interfaces):
-            # If the algorithm works with a subset of the interfaces
-            # present in the set then only feed these through to the algorithm
-            valid_input = {
-                civ for civ in civ_set if civ.interface in input_interfaces
-            }
-        else:
-            continue
+    (archive,) = archives
```
jmsmkn, line 240: For my understanding, what does `(archive,) = archives` do?

app/grandchallenge/algorithms/tasks.py
```diff
@@ -163,105 +159,137 @@ def create_algorithm_jobs(
...
+    if algorithm_model:
+        extra_filter = {"algorithm_model": algorithm_model}
+    else:
+        extra_filter = {"algorithm_model__isnull": True}
+
+    relevant_jobs = Job.objects.filter(
+        algorithm_image=algorithm_image,
+        algorithm_interface__in=valid_job_inputs.keys(),
+        creator=None,
+        inputs__archive_items__archive=archive,
```
jmsmkn, line 264: You have the archive items here, so you might be able to do:

```suggestion
        inputs__archive_items__in=archive_items,
```

Then, you might not need the `archive` from above.

app/grandchallenge/algorithms/tasks.py
```diff
@@ -163,105 +159,137 @@ def create_algorithm_jobs(
...
+    filtered_valid_job_inputs = {}
+    for interface, archive_items in valid_job_inputs.items():
+        items_with_job = [
+            ai.pk
+            for ai in archive_items
+            if frozenset(ai.values.all())
+            in [
+                frozenset(j.inputs.all())
+                for j in existing_jobs_for_interfaces[interface]
+            ]
```
jmsmkn, line 286: It may or may not matter here but doing set lookups is quicker as the items are hashed, so python does not need to iterate over the items and perform a comparison each time.

```suggestion
            in {
                frozenset(j.inputs.all())
                for j in existing_jobs_for_interfaces[interface]
            }
```

Also double check that this isn't being repeatedly calculated. It may be better to extract this to a variable which is declared before `items_with_job`.

app/grandchallenge/algorithms/tasks.py
```diff
@@ -163,105 +159,137 @@ def create_algorithm_jobs(
...
+        return {}
 
-    existing_jobs = {
-        frozenset(j.inputs.all())
-        for j in Job.objects.filter(
-            algorithm_image=algorithm_image, algorithm_model=algorithm_model
+    archives = {ai.archive for ai in archive_items}
+    if len(archives) != 1:
+        raise RuntimeError(
+            "The archive items should all belong to the same archive."
```
jmsmkn, line 238: Is there a reason for this restriction? I don't think we ever send that, but if it is due to the logic of the algorithm below then that should be mentioned in a comment (the why). If we filter the jobs by `archive_item` rather than `archive` then I don't think it is.

app/grandchallenge/algorithms/tasks.py
```diff
@@ -163,105 +159,137 @@ def create_algorithm_jobs(
...
+            ai.pk
+            for ai in archive_items
+            if frozenset(ai.values.all())
+            in [
+                frozenset(j.inputs.all())
+                for j in existing_jobs_for_interfaces[interface]
+            ]
+        ]
+        filtered_valid_job_inputs[interface] = list(
+            valid_job_inputs[interface].exclude(pk__in=items_with_job)
```
jmsmkn, line 289: This will go back to the database one time per interface. As you have all of the archive items in python already it may be more efficient to send over an iterable of the ones you want to keep.

app/grandchallenge/evaluation/models.py
```diff
@@ -1671,59 +1725,64 @@ def algorithm_inputs(self):
...
+        # subset to relevant jobs only to avoid querying
+        # the full table again for each interface
+        relevant_jobs = Job.objects.filter(
+            algorithm_image=self.submission.algorithm_image,
+            status=Job.SUCCESS,
+            algorithm_interface__in=algorithm_interfaces,
+            creator=None,
+            inputs__archive_items__archive=self.submission.phase.archive,
+            **extra_filter,
+        )
```
jmsmkn, line 1749: It seems like you send a similar query to `get_valid_jobs_for_interfaces_and_archive_items` both times and in `get_valid_jobs_for_interfaces_and_archive_items` you have to remember to use `distinct` due to the query outside. The `algorithm_interfaces` argument is also duplicated. It's also called `get` rather than `filter` or `group`. Maybe `get_valid_jobs_for_interfaces_and_archive_items` should be responsible for initialising the query?

jmsmkn (commented):
Looks good! It's definitely easier to follow. I remember Chris spending quite some effort in optimising these queries, have you double checked the number of database calls and that everything is still being prefetched/selected efficiently?

app/grandchallenge/algorithms/tasks.py
amickan: Ah that's a good point. Then indeed I don't need the archive above.

app/grandchallenge/algorithms/tasks.py
amickan: `archives` is a set that contains 1 archive, and this extracts that archive so I can use it in the filter. `*archives` is not supported in queries, so this seemed like the best way to extract the archive. But it is anyway not needed anymore given that I'm now filtering by archive_items directly.

app/grandchallenge/evaluation/models.py
amickan: > It seems like you send a similar query to `get_valid_jobs_for_interfaces_and_archive_items` both times

The key difference between the 2 queries is that we're subsetting by job status `SUCCESS` here whereas in the other case we don't. I could pass this as an extra argument to `get_valid_jobs_for_interfaces_and_archive_items` and do it there, but then I also need to pass the algorithm_image, algorithm_model and archive_items to that function to do the full query there. It seemed clearer to me to narrow down the queryset in the respective places. 

> in `get_valid_jobs_for_interfaces_and_archive_items` you have to remember to use `distinct` due to the query outside

I'm already using `.distinct()` in `get_valid_jobs_for_interfaces_and_archive_items` (on line 178) for that reason, and I think it works fine, or do you think that I'm using it in the wrong place? 

> The `algorithm_interfaces` argument is also duplicated.

Yeah ok, it's unnecessary to subset by interface here already, that's true. I can remove that filter. In `get_valid_jobs_fro_interfaces_and_archive_items` I need to filter by interface since I'm grouping them by interface.

amickan: > Maybe `get_valid_jobs_for_interfaces_and_archive_items` should be responsible for initialising the query?

I have decided to move the query here. It also avoids repeating the logic about the algorithm_model filter. One place for the logic is worth the more complex function signature.

app/grandchallenge/algorithms/tasks.py
amickan: I double checked and can confirm that this is not repeatedly calculated. `existing_jobs_for_interfaces` contains an already evaluated list of jobs per interface.

amickan: Actually, moving this out of the loop helps with readability and I also realized that I can directly filter for items withOUT a job and hence can avoid the later list comprehension all together.

amickan:
> Looks good! It's definitely easier to follow. I remember Chris spending quite some effort in optimising these queries, have you double checked the number of database calls and that everything is still being prefetched/selected efficiently?

I incorporated your latest feedback, and also took another close look at the queries that all of this generates, and made some final minor adjustments. The number of queries scales with the number of interfaces, but I don't think there is a way around that. Other than that I think the queries are now as efficient as they can be.

jmsmkn (approved):
LGTM, great work!

app/grandchallenge/evaluation/models.py
jmsmkn: We should have this as a follow up to the pitch so we can benchmark it next cycle and make any corrections in the April cycle.

PR #3821, opened 2025-02-07
Parent object layer for flexible file

koopmant:
Add a layer for selecting the parent object type (algorithm job, display set or archive item) to the flexible file widget. 

The parent object is used to make the query more specific and smaller. 

related to https://github.com/DIAGNijmegen/rse-roadmap/issues/363

### DIFF
--- a/app/grandchallenge/serving/models.py
+++ b/app/grandchallenge/serving/models.py
@@ -1,6 +1,5 @@
 from django.conf import settings
 from django.db import models
-from django.db.models import Exists, OuterRef
 
 from grandchallenge.algorithms.models import (
     AlgorithmImage,
@@ -12,10 +11,7 @@ from grandchallenge.cases.models import Image
 from grandchallenge.challenges.models import ChallengeRequest
 from grandchallenge.components.models import ComponentInterfaceValue
 from grandchallenge.components.widgets import ParentObjectTypeChoices
-from grandchallenge.core.guardian import (
-    filter_by_permission,
-    get_objects_for_user,
-)
+from grandchallenge.core.guardian import filter_by_permission
 from grandchallenge.evaluation.models import Submission
 from grandchallenge.reader_studies.models import DisplaySet
 from grandchallenge.workstations.models import Feedback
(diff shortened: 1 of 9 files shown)


koopmant:
You can test the widget manually on this branch:
https://github.com/comic/grand-challenge.org/tree/integrate-flexible-file-widget

app/grandchallenge/serving/models.py
```diff
@@ -47,26 +53,78 @@ class Download(models.Model):
...
-        user=user, perms="algorithms.view_job"
-    ).filter(outputs__pk__in=OuterRef("pk"))
+    civs = ComponentInterfaceValue.objects.filter(**extra_filter_kwargs)
 
-    display_set_query = get_objects_for_user(
-        user=user, perms="reader_studies.view_displayset"
-    ).filter(values__pk__in=OuterRef("pk"))
+    pks_for_filter = []
+    if parent_object_type_choice in (None, ParentObjectTypeChoices.JOB):
+        job_query = filter_by_permission(
```
amickan, line 76: Why include `None` here and in the other 3 cases? Shouldn't we only get civs for jobs/ds/ai if that was specifically requested?

amickan: Ah I see why, you still need to filter the civs by permission for the user.

app/tests/serving_tests/test_models.py
```diff
@@ -19,43 +19,68 @@
 @pytest.mark.django_db
 def test_get_component_interface_values_for_user():
```
amickan, line 20: Maybe also add a check for subsettting by parent object choice? I see that you are already testing this in the views for the widget, but I think it should also be here since it is inherent to the `get_component_interface_values_for_user` function.

amickan (approved):
LGTM!

amickan:
> You can test the widget manually on this branch: https://github.com/comic/grand-challenge.org/tree/integrate-flexible-file-widget

I gave it a quick test: looks great! Two things I noticed (maybe you're still working on it, since this is for the next PR, but leaving the comments here anyway):
1. I think you are still loading the full file queryset for the user on initial widget load (i.e before parent object selection) and on form validation. I think the queryset should be empty at the beginning and appropriately subset at validation, so as to force the user to make a parent object choice. I only took a quick look, so it's possible that I missed the code that does this somewhere, but thought I'd mention it.
2. The upload widget didn't load fully for me. I didn't get the draf-drop area and hence could not actually upload a new file. You're probably missing some uppy initialization magic somewhere after an htmx load event.

app/tests/serving_tests/test_models.py
koopmant: Good point, thanks! Added it :)

koopmant:
> I gave it a quick test: looks great! Two things I noticed (maybe you're still working on it, since this is for the next PR, but leaving the comments here anyway):

>- I think you are still loading the full file queryset for the user on initial widget load (i.e before parent object selection) and on form validation. I think the queryset should be empty at the beginning and appropriately subset at validation, so as to force the user to make a parent object choice. I only took a quick look at the integration branch, so it's possible that I missed the code that does this somewhere, but thought I'd mention it.
>- The upload widget didn't load fully for me. I didn't get the draf-drop area and hence could not actually upload a new file. You're probably missing some uppy initialization magic somewhere after an htmx load event.

I´ll address these in the [next PR](https://github.com/comic/grand-challenge.org/pull/3824).

PR #3823, opened 2025-02-07
Add download example link to interface form field

### LINKED ISSUE #3809: Add download example link when trying out algorithm
Some formats are best explained when downloading an example, the following has an example CIV:

Image

I think adding a download link in the help text area would help here.

ammar257ammar:
closes #3809


ammar257ammar:
This is how the download example link looks now

Screenshot 2025-02-10 at 10 57 24

app/grandchallenge/components/templates/components/partials/example_download_link.html
```diff
@@ -0,0 +1,8 @@
+{% load json %}
+
+{% if object.json_kind_example %}
+    Example usage: <a href="data:text/plain;charset=utf-8,{{ object.json_kind_example.value|json_dumps|urlencode }}"
+        download="example.json">
```
chrisvanrun, line 5: If there are a lot of files to download, would it make sense to incorporate the interface slug in the download filename somehow?

ammar257ammar: sure, good idea. It should be possible.

app/tests/components_tests/test_forms.py
```diff
@@ -190,3 +196,24 @@ def test_image_widget_current_value_in_archive_item_and_display_set_update_forms
...
     )
+
+
+@pytest.mark.parametrize(
+    "component_interface_kind",
+    [kind for kind in INTERFACE_TYPE_JSON_EXAMPLES.keys()],
+)
+@pytest.mark.parametrize("store_in_db", [True, False])
+@pytest.mark.django_db
+def test_interface_form_field_help_text_example_download_link(
```
chrisvanrun, line 207: I think you could:
- move this to a dedicated `test_fields.py`, create it if it doesn't already exists
- not use the django_db but rather pass user=None, this speeds things up a lot
- add a negative test: i.e. add an case where there should _not_ be an example

app/tests/components_tests/test_forms.py
```diff
@@ -190,3 +196,24 @@ def test_image_widget_current_value_in_archive_item_and_display_set_update_forms
...
+    component_interface_kind, store_in_db
+):
+    user = UserFactory()
+    ci = ComponentInterfaceFactory(
+        kind=component_interface_kind, store_in_database=store_in_db
+    )
+    field = InterfaceFormField(instance=ci, user=user, form_data={})
+
+    encoded_example = quote(json.dumps(ci.json_kind_example.value, indent=2))
```
chrisvanrun, line 216: I would make this explicit. Add a few concretely encoded examples and provide them via the parametrize. This seems to be too 'smart'.

I think it is OK if does not test every single INTERFACE_TYPE_JSON_EXAMPLES, I would focus the test on three cases:
- Should show an example
- Should not show an example
- Special case of the first, but not a JSONField: so no example

chrisvanrun (changes requested):
Looks pretty good already! A few questions and suggestions for spit and polish.

app/grandchallenge/components/templates/components/partials/example_download_link.html
ammar257ammar: Done

app/grandchallenge/components/form_fields.py
```diff
@@ -65,6 +66,11 @@ def __init__(self, *, instance=None, user=None, form_data=None, **kwargs):
             "label": instance.title.title(),
         }
 
+        self.example_download_link = render_to_string(
```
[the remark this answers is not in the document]
ammar257ammar: I agree. It would be cleaner.
Done

app/grandchallenge/components/templates/components/partials/example_download_link.html
```diff
@@ -0,0 +1,8 @@
+{% load json %}
+
+{% if object.json_kind_example %}
+    Example usage: <a href="data:text/plain;charset=utf-8,{{ object.json_kind_example.value|json_dumps|urlencode }}"
```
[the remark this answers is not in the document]
ammar257ammar: I would go with `<a>Download example</a>`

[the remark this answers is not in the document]
ammar257ammar: Looks like this now:

Screenshot 2025-02-12 at 10 15 06

app/grandchallenge/components/templates/components/partials/example_download_link.html
```diff
@@ -0,0 +1,8 @@
+{% load json %}
```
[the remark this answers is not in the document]
ammar257ammar: Thanks!

app/tests/components_tests/test_forms.py
ammar257ammar: Done

app/grandchallenge/components/form_fields.py
```diff
@@ -138,7 +144,9 @@ def get_json_field(self):
             JSONValidator(schema=default_schema),
             JSONValidator(schema=self.instance.schema),
         ]
-        extra_help = ""
+
+        extra_help = self.example_download_link
```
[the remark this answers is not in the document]
ammar257ammar: I changed it to show only for CIs of JSONField type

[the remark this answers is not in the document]
ammar257ammar: Done

app/tests/components_tests/test_forms.py
ammar257ammar: All points were taken except:
> not use the django_db but rather pass user=None, this speeds things up a lot

There are factories for component interface creation which need access to the database to create them.

Maybe I am mistaken, can you please clarify that point and if it is meant for this type of tests?

app/grandchallenge/components/form_fields.py
```diff
@@ -176,8 +182,13 @@ def get_file_field(self):
...
                 )
 
+        extra_help = _join_with_br(
+            f"{file_upload_text} {self.instance.file_extension}",
+            self.example_download_link,
+        )
+
         self.kwargs["widget"] = FlexibleFileWidget(
-            help_text=self.help_text,
+            help_text=_join_with_br(self.help_text, extra_help),
```
ammar257ammar, line 191: After the new implementation of file widget #8ae9479 
The download link is not showing correctly.
Looking into it.

amickan: Thomas is still in the process of heavily refactoring the logic there, so maybe best to wait until he is done.

ammar257ammar: Got it, I will wait then.

app/tests/components_tests/test_forms.py
chrisvanrun: The factories for the component interface creation are indeed DB heavy. However, you can also call them in create object only mode, not having to hit the database. 

There might be another gotcha keeping us from making thist test database free, but not having to hit the database means a one (or maybe two) magnitudes quicker testing.

app/grandchallenge/components/form_fields.py
```diff
@@ -144,7 +145,12 @@ def get_json_field(self):
             JSONValidator(schema=default_schema),
             JSONValidator(schema=self.instance.schema),
         ]
+
         extra_help = ""
+
+        if self.instance.default_field == forms.JSONField:
+            extra_help = self.example_download_link
```
chrisvanrun, line 152:
```suggestion
        if self.instance.default_field == forms.JSONField:
            extra_help = self.example_download_link
        else:
            extra_help = ""
```

Sorry! Nitpick code styling.

app/tests/components_tests/test_forms.py
ammar257ammar: okay got it, I look into it to avoid calling the database

app/grandchallenge/components/form_fields.py
ammar257ammar: Done

app/tests/components_tests/test_forms.py
ammar257ammar: I have been trying today to achieve this in multiple ways. For model factories, by calling `build()` on the factory class, it is possible to create an instance without hitting the database. However, `InterfaceFormField` still needs access to the database since it calls `filter_by_permission` and that requires accessing the database through the `guardian` package.

The test fundamentally tests database-backed functionality, so mocking or using pure objects would not provide the same level of integration testing that validates the form field works correctly with the actual database models and relationships.

Lets discuss this further on Monday.

ammar257ammar: Done

chrisvanrun:
I hope to cycle onto this tomorrow.

PR #3824, opened 2025-02-10
Integrate flexible file widget

koopmant:
Use the flexible file widget in the forms by integrating it in the `InterfaceFormField` class. 

Usage of parent object type choice in the widget has been removed. See discussion in the pitch. 

Related to https://github.com/DIAGNijmegen/rse-roadmap/issues/363

### DIFF
--- a/app/grandchallenge/components/form_fields.py
+++ b/app/grandchallenge/components/form_fields.py
@@ -183,8 +183,8 @@ class InterfaceFormField(forms.Field):
 
     @cached_property
     def civs_for_user_for_interface(self):
-        return get_component_interface_values_for_user(user=self.user).filter(
-            interface=self.instance
+        return get_component_interface_values_for_user(
+            user=self.user, interface=self.instance
         )
 
     @property
--- a/app/grandchallenge/components/form_fields.py
+++ b/app/grandchallenge/components/form_fields.py
@@ -156,7 +156,7 @@ class InterfaceFormField(forms.Field):
             if isinstance(self.initial, ComponentInterfaceValue):
                 # This can happen on display set or archive item update forms, the value is then taken from the model
                 # instance unless the value is in the form data.
-                current_value = self.initial.file
+                current_value = self.initial
             # Otherwise the value is taken from the form data and will always take the form of a pk for either
             # a ComponentInterfaceValue object (in this case the pk is a digit) or
             # a UserUpload object (then the pk is a UUID).
--- a/app/grandchallenge/components/form_fields.py
+++ b/app/grandchallenge/components/form_fields.py
@@ -10,17 +10,13 @@ from grandchallenge.cases.widgets import (
 )
 from grandchallenge.components.models import ComponentInterfaceValue
 from grandchallenge.components.schemas import INTERFACE_VALUE_SCHEMA
-from grandchallenge.components.widgets import (
-    FlexibleFileWidget,
-    SelectUploadWidget,
-)
+from grandchallenge.components.widgets import FlexibleFileWidget
 from grandchallenge.core.guardian import get_objects_for_user
 from grandchallenge.core.validators import JSONValidator
 from grandchallenge.core.widgets import JSONEditorWidget
 from grandchallenge.serving.models import (
     get_component_interface_values_for_user,
 )
-from grandchallenge.subdomains.utils import reverse
 from grandchallenge.uploads.models import UserUpload
 from grandchallenge.uploads.widgets import (
     UserUploadMultipleWidget,
@@ -50,6 +46,7 @@ class InterfaceFormField(forms.Field):
         UserUploadSingleWidget,
         JSONEditorWidget,
         FlexibleImageWidget,
+        FlexibleFileWidget,
     }
 
     def __init__(self, *, instance=None, user=None, form_data=None, **kwargs):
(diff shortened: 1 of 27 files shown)


koopmant:
From @amickan in [previous PR](https://github.com/comic/grand-challenge.org/pull/3821)
> I gave it a quick test: looks great! Two things I noticed (maybe you're still working on it, since this is for the next PR, but leaving the comments here anyway):

>- I think you are still loading the full file queryset for the user on initial widget load (i.e before parent object selection) and on form validation. I think the queryset should be empty at the beginning and appropriately subset at validation, so as to force the user to make a parent object choice. I only took a quick look at the integration branch, so it's possible that I missed the code that does this somewhere, but thought I'd mention it.
>- The upload widget didn't load fully for me. I didn't get the draf-drop area and hence could not actually upload a new file. You're probably missing some uppy initialization magic somewhere after an htmx load event.

koopmant:
> The upload widget didn't load fully for me. I didn't get the draf-drop area and hence could not actually upload a new file. You're probably missing some uppy initialization magic somewhere after an htmx load event.

I had fixed the upload widget not loading in a different branch earlier (branch of refactoring and cleanup; not pushed yet). Had to search for the issue again. I was accidentally using a string representation of the interface instead of the prefixed interface slug as id for the element. That's an invalid selector which is why the javascript was throwing an error.

koopmant:
I added the `ALL` option in parent object type choices. This allows the user to search everywhere if they want to. I imagine that will still result in a time-out for the users that we started this pitch for, but for most users this will be nice to revert back to? This has also clarified the code in `get_component_interface_values_for_user`.

koopmant:
I have not added any new tests. I take a look at it again tomorrow to see if I can come up with some.

app/grandchallenge/components/form_fields.py
```diff
@@ -144,42 +148,39 @@ def get_json_field(self):
...
+                isinstance(self.initial, int) or self.initial.isdigit()
+            ) and ComponentInterfaceValue.objects.filter(
+                pk=self.initial
+            ).exists():
+                current_value = ComponentInterfaceValue.objects.get(
+                    pk=self.initial
+                )
+            elif UserUpload.objects.filter(pk=self.initial).exists():
+                current_value = UserUpload.objects.get(pk=self.initial)
+            else:
```
amickan, line 166: Could you add comments to the code to explain when each of these situations arises? I think that would help with debugging in the future.

app/grandchallenge/components/form_fields.py
```diff
@@ -144,42 +148,39 @@ def get_json_field(self):
...
+                raise RuntimeError(
+                    f"Unknown type for initial value: {self.initial}"
+                )
+
+        self.kwargs["widget"] = FlexibleFileWidget(
+            help_text=self.help_text,
+            user=self.user,
+            current_value=current_value,
+            # also passing the CIV as current value here so that we can
+            # show the image name to the user rather than its pk
```
amickan, line 176: Maybe this comment can go if you add comments above. 

```suggestion
            # also passing the CIV as current value here so that we can
            # show the filename to the user rather than its pk
```

app/grandchallenge/components/form_fields.py
```diff
@@ -217,6 +230,23 @@ def __init__(
...
+            parent_object_type_choice = ParentObjectTypeChoices(
+                parent_object_type_choice_name,
+            )
+            self.fields[0].queryset = get_component_interface_values_for_user(
+                user=self.user,
+                interface=self.interface,
+                parent_object_type_choice=parent_object_type_choice,
+            )
+        return super().clean(value)
+
```
amickan, line 249: What's the rationale for doing this in the clean method rather than in the init method? I think the clean method should only be checking the value and prepping the field for that to work, right? 

Is the order of the fields guaranteed to stay the same? If you do it in the init method, it'll be clearer (and more future-proof) which field the queryset is set for.

app/grandchallenge/components/form_fields.py
```diff
@@ -199,11 +200,23 @@ class FlexibleFileField(MultiValueField):
...
+        user=None,
+        interface=None,
+        form_data=None,
         disabled=False,
         **kwargs,
     ):
+        self.user = user
+        self.interface = interface
+        if form_data is None:
+            form_data = {}
```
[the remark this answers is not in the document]
koopmant: It's one of those python quirks. Default argument values are only evaluated once, so if the dict is defined there it will persist between function calls. See the warning in https://docs.python.org/3/tutorial/controlflow.html#default-argument-values

app/grandchallenge/components/form_fields.py
koopmant: Ah, this is a lack of understanding on my part. I was under the impression the form field was already instantiated when the form is presented. (I was wondering how that worked...)
Yes, in the init function makes a lot more sense. 

Regarding the order of the fields, good question. I'm not sure and will look into it. But since this will move to the init function, the potential issue will be resolved anyway.

app/grandchallenge/components/form_fields.py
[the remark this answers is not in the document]
amickan: Ah ok, thanks for explaining!

amickan:
LGTM, see questions/comments above. However, I still seem to have the issue that the upload widget doesn't fully load. I tried just now on the archive item update page.

app/grandchallenge/components/form_fields.py
koopmant: Coming back to the ordering question. (Thanks for the question. Ever since James told me relying on list order can create flaky tests I wanted to look into this a bit further.)

The order in a python list is persistent. So with the fields being defined explicitly as they are in the init function, you could rely on the ordering like this. Of course, if the list is replaced or sorted in-place somewhere else in the code this would still break. 

As a side note, many collections in python are ordered. The order of the items depends on how the collection is created (and changed if mutable). 
Relying on the order is a bad idea when the items in the list are not created in a particular order, for example when a serializer isn´t ordering the items in a certain way. 

In general, I think, relying on order when it can be avoided adds unneccessary complexity.

app/grandchallenge/components/form_fields.py
```diff
@@ -144,42 +150,41 @@ def get_json_field(self):
...
-            return ModelChoiceField(
-                queryset=self.civs_for_user_for_interface,
-                widget=SelectUploadWidget(
-                    attrs={"upload_link": file_upload_link}
-                ),
-                **self.kwargs,
-            )
+        if self.initial:
+            if isinstance(self.initial, ComponentInterfaceValue):
+                # This can happen when a form is presented for an existing item.
```
amickan, line 157: More specifically, this is when a form is rendered after a validation error on another field and then re-submitted, right?

koopmant: Ah, no. I believe this only happens on display set and archive item UpdateForms. 
The relevent lines are in the `MultipleCIVForm` here: [components/forms.py#L136-L139](https://github.com/comic/grand-challenge.org/blob/integrate-flexible-file-widget/app/grandchallenge/components/forms.py#L136-L139)
Am I understanding this correctly? Then I'll update the comment to clarify.

amickan: Ah ok, that could be. Yes, please update the comment. And in the case that I described above, we send along the pk, not the CIV?

amickan (dismissed):
LGTM

koopmant:
> I think I found a bug: on the ArchiveItemUpdate and DisplaySetUpdate views, already existing file values are not attached to the widget properly. The widget is empty and when you hit save, the previously attached file is removed.

Thanks for checking! I noticed something odd in the code when I was cleaning up the code in the next branch. So I know exactly where this happens.

amickan (approved):
LGTM! Great work!

PR #3826, opened 2025-02-11
Add evaluation error message to evaluation details view

### LINKED ISSUE #3467: Evaluation error message missing from detail view

ammar257ammar:
closes #3467 

This PR adds a message section to the evaluation details view (admin section) to better communicate with the user the reason for evaluation failure.

I also adds an error message column to the prerequisite jobs table to show the error of failed jobs.

### DIFF
--- a/app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
+++ b/app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
@@ -210,14 +210,14 @@
 
                 {% if not incomplete_jobs %}
 
-                    All prequisite jobs succeeded.
+                    All prerequisite jobs succeeded.
 
                     {% if object.error_message %}
                         Evaluation failed due to the following error: {{ object.error_message }}
                     {% endif %}
 
                 {% else %}
-                    Some jobs are incomplete. Check the status and error messages of the prequisite jobs above.
+                    Some jobs are incomplete. Check the status and error messages of the prerequisite jobs above.
                 {% endif %}
             </div>
 
--- a/app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
+++ b/app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
@@ -56,6 +56,13 @@
             {% include 'evaluation/evaluation_status_detail.html' %}
         </dd>
 
+        {% if object.error_message %}
+            <dt>Error Message</dt>
+            <dd>
+                {{ object.error_message }}
+            </dd>
+        {% endif %}
+
         <dt>User</dt>
         <dd>
             {{ object.submission.creator|user_profile_link }}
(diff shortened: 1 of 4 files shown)


app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
```diff
@@ -205,6 +205,19 @@ <h3 class="card-title">Contact User</h3>
 
             {% include "evaluation/evaluation_incomplete_jobs_detail.html" %}
 
+            <div class="card-body">
+                <h3 class="card-title">Error Message</h3>
```
ammar257ammar, line 209: Since the message in this section is not necessarily an error (e.g. in the case of incomplete jobs), should the title be renamed to "Status Message" rather than "Error Message"?

app/grandchallenge/evaluation/templates/evaluation/evaluation_incomplete_jobs_detail.html
```diff
@@ -31,6 +32,9 @@ <h3 class="card-title">Prerequisite Jobs</h3>
                         <td>
                             {% include "algorithms/job_status_detail.html" with object=job %}
                         </td>
+                        <td>
+                            {% include "algorithms/job_error_message.html" with object=job %}
```
jmsmkn, line 36: This does not need to be a separate partial and the entirety of stderr should not be presented here:

```suggestion
                            {{ object.error_message }}
```

Please also remove the separate file.

ammar257ammar: sure, I will remove it

ammar257ammar: Done

app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
```diff
@@ -205,6 +205,21 @@ <h3 class="card-title">Contact User</h3>
...
+                    <h3 class="card-title">Error Message</h3>
+
+                    {% if not incomplete_jobs %}
+                        All prerequisite jobs succeeded.
+                        {% if object.error_message %}
+                            Evaluation failed due to the following error: {{ object.error_message }}
+                        {% endif %}
+                    {% else %}
+                        Some jobs are incomplete. Check the status and error messages of the prerequisite jobs above.
+                    {% endif %}
```
jmsmkn, line 219: Unnecessary negation of boolean test.

```suggestion
                    {% if incomplete_jobs %}
                        Some jobs are incomplete. Check the status and error messages of the prerequisite jobs above.
                    {% else %}
                        All prerequisite jobs succeeded.
                        {% if object.error_message %}
                            Evaluation failed due to the following error: {{ object.error_message }}
                        {% endif %}
                    {% endif %}
```

app/grandchallenge/evaluation/templates/evaluation/evaluation_detail.html
```diff
@@ -205,6 +205,21 @@ <h3 class="card-title">Contact User</h3>
 
             {% include "evaluation/evaluation_incomplete_jobs_detail.html" %}
 
+            {% if object.status != object.SUCCESS or incomplete_jobs %}
```
jmsmkn, line 208: Unnecessary check.

```suggestion
            {% if object.status != object.SUCCESS %}
```

jmsmkn: Actually this will show up when the jobs are in pending states. The only check and display needs to be on the `object.error_message`.

jmsmkn (approved):

PR #3828, opened 2025-02-11
Add production danger theme

chrisvanrun:
This should clearly show that one is working on production:

image


chrisvanrun:
> Thanks! As you are away today I made a few minor changes. I know that you followed what the Django docs say but one is about organisation and the second removed the inline styles to make it CSP compatible.

Thank you for the heads up on the changes.

PR #3832, opened 2025-02-13
Remove inputs and outputs fields from Phase and Algorithm models

amickan:
Final PR for https://github.com/DIAGNijmegen/rse-roadmap/issues/153


app/grandchallenge/algorithms/forms.py
```diff
@@ -178,7 +178,7 @@ def __init__(self, *args, algorithm, user, interface, **kwargs):
                 instance=inp,
                 initial=initial if initial else inp.default_value,
                 user=self._user,
-                required=True,
+                required=inp.value_required,
```
amickan, line 181: Booleans and json inputs that are allowed to be empty should not be required. 

This was changed to `True` in [this PR](https://github.com/comic/grand-challenge.org/pull/3620) because back then a job's inputs were not considered complete with an empty json input even if that was valid for the interface. That was fixed in my last PR, so jobs will now run just fine with empty json inputs (when empty is a valid value for the interface).

app/grandchallenge/algorithms/templates/algorithms/job_detail.html
```diff
@@ -52,10 +52,9 @@
                    aria-selected="false"><i class="fas fa-terminal fa-fw"></i>&nbsp;Logs
                 </a>
             {% endif %}
-
             {% if object.status == object.SUCCESS and perms.reader_studies.add_readerstudy %}
                 <a class="nav-link"
-                   href="{% url "algorithms:display-set-from-job-create" slug=object.algorithm_image.algorithm.slug interface_pk=object.pk %}">
+                   href="{% url "algorithms:display-set-from-job-create" slug=object.algorithm_image.algorithm.slug pk=object.pk %}">
                     <i class="fas fa-edit fa-fw"></i>&nbsp;Edit in Reader Study
```
amickan, line 58: I had accidentally changed this from `pk` to `interface_pk` in one of my earlier PRs to the feature branch. It should stay at `pk` (refers to job pk).

app/grandchallenge/hanging_protocols/forms.py
```diff
@@ -183,20 +183,23 @@ def __init__(self, *args, **kwargs):
         )
 
     def _get_interface_lists(self):
+        sorted_interfaces = sorted(
+            list(self.instance.linked_component_interfaces), key=lambda x: x.pk
```
jmsmkn, line 187: I remember that Chris recently worked on this. Does it make sense to sort by the pk? I thought this was sorted by slug?

app/grandchallenge/algorithms/serializers.py
```diff
@@ -56,10 +56,6 @@ class Meta:
 
 class AlgorithmSerializer(serializers.ModelSerializer):
     average_duration = SerializerMethodField()
-    # TODO: Still to be addressed for optional inputs pitch
```
jmsmkn, line 59: Have all of these TODOs been checked and removed?

jmsmkn (commented):
Looks good! Couple of small questions.

app/grandchallenge/algorithms/forms.py
jmsmkn: Can't say that I understand that fully but I trust you!

app/grandchallenge/hanging_protocols/forms.py
amickan: Sorting by slug is also fine. This is only used for generating the view_content examples, so doesn't really matter as long as they are sorted somehow. I will update it to sorting by slug.

app/grandchallenge/algorithms/serializers.py
amickan: Yes

app/grandchallenge/algorithms/forms.py
amickan: Ok. This ultimately boils down to the question: should a json interface for which empty values are allowed by its schema be marked as "required" as an algorithm input or not? 

My implementation here does not mark the corresponding field as required. A user can leave the json input field empty and that will be considered valid input for this interface. When a job gets created like that, a CIV for that interface without a value gets created. If we were to make the field required (like it used to be), the user would be forced to enter something even though the schema does not require that. I think it makes sense to make the `required` property schema-dependent, but doing this essentially makes the json input an optional input. 

Booleans are a different story: those should definitely not be required or else only True would be valid.

amickan:
Some session control related tests keep failing even though they pass locally. I cant figure out why that is at the moment.

jmsmkn:
It's strange, test case teardown stuff. Did you destroy your local database and run the test without `--reuse-db`?

amickan:
> It's strange, test case teardown stuff. Did you destroy your local database and run the test without `--reuse-db`?

Now I get the error locally as well. I'm investigating.

amickan:
The error is this: https://github.com/3YOURMIND/django-deprecate-fields/issues/22 

This is only a problem in `TransactionTestCase`, which the `live_server` fixture that we're using for the playrwright test relies on. Since this seems restricted to the test, I think marking that with `xfail` until we've fully removed the field would be the way to go?

jmsmkn:
Yes that's fine so long as we know to enable them again.

amickan:
Done! I will add a to-do to the pitch to remove the fields and then also re-enable the test.

PR #3833, opened 2025-02-14
Cleanup flexible widget

koopmant:
Cleanup after changes for the flexible file widget to make future maintenance easier. 

`InterfaceFormField` has been renamed to `InterfaceFormFieldFactory` and has been stripped of much of the logic, which had become very confusing. 

User object permissions are now checked before rendering, to avoid leaking information about object existance. 

This also contains a functional change to the htmx for the `FileSearchWidget` and the `ImageSearchWidget`. 
`hx-trigger` now includes `input` ensure triggering when text is removed. 
`hx-swap` is now the innerHTML of the select element. This makes it easier to see that this element has the widgetname which is where the value is taken from. 

Related to https://github.com/DIAGNijmegen/rse-roadmap/issues/363


app/grandchallenge/cases/views.py
```diff
@@ -107,73 +107,67 @@ class RawImageUploadSessionViewSet(
 
 class ImageWidgetSelectView(LoginRequiredMixin, View):
     def get(self, request, *args, **kwargs):
-        interface = request.GET.get("interface_slug")
-        widget_choice = request.GET.get(f"WidgetChoice-{interface}")
-        help_text = request.GET.get("help_text")
```
amickan, line 112: Are you sure the help-text is also set appropriately? We have some standard help text (the upload one you are loading below), but we also add the interface description to the help text when it exists, and I think you're not passing that in here, right?

app/grandchallenge/components/form_fields.py
```diff
@@ -55,147 +47,75 @@ class InterfaceFormField(forms.Field):
...
-        ).filter(status=UserUpload.StatusChoices.COMPLETED)
-        image_queryset = get_objects_for_user(self.user, "cases.view_image")
-        return FlexibleImageField(
-            upload_queryset=upload_queryset,
-            image_queryset=image_queryset,
-            **self.kwargs,
-        )
+        kwargs = {
+            "required": required,
+            "help_text": help_text,
```
amickan, line 68: Since you have access to the interface here, maybe best to set the help_text here to `interface.description`. Then you don't need to pass that in from the form, right?

app/grandchallenge/components/views.py
```diff
@@ -519,42 +515,27 @@ def get(self, request, *args, **kwargs):
...
-                            },
-                        )["widget"],
-                    },
+                return HttpResponse(
+                    UserUploadSingleWidget().render(
+                        name=prefixed_interface_slug,
+                        value=None,
+                        attrs={
+                            "id": prefixed_interface_slug,
+                            "help_text": f"{file_upload_text} {interface.file_extension}",
```
amickan, line 535: Should this help text also be added to the InterfaceFormField, so to the multi-widget (before selecting one of the 2 options)?

koopmant: No, I don't think so. This help text is for the upload widget in particular.

app/grandchallenge/cases/views.py
koopmant: Passing it as a keyword argument to the `FlexibleFileField` which passes it on to `MultiValueField`. I haven't quite figured out how django does it, but that help text is automatically rendered underneath the widget. So we don't need to render it in our template anymore. and don't need to pass it along in this view.

amickan: Testing this out, I don't see the interface specific help-text added, so I think you do need to pass it in here.

koopmant: For future reference. I looked into how django renders the help text. Rendering of this and other items around fields (like labels, tags and errors) are created by the renderer of the form.

app/grandchallenge/components/form_fields.py
koopmant: Yes, that is excellent actually. Before, we only did it on the `JobCreateForm`; on the `SingleCIVForm` and `MultipleCIVForm` we did not set a help text. Now setting it here so that the interface description shows up on all the forms.

app/grandchallenge/cases/views.py
koopmant: > Testing this out, I don't see the interface specific help-text added, so I think you do need to pass it in here.

Hmm, maybe I'm misunderstanding then. I'll contact you tomorrow.

app/grandchallenge/cases/widgets.py
```diff
@@ -94,20 +83,52 @@ class FlexibleImageField(MultiValueField):
...
+                # This can happen on display set or archive item update forms, the value is then taken from the model
+                # instance unless the value is in the form data.
+                self.current_value = initial.image
+                initial = initial.image.pk
+            # Otherwise the value is taken from the form data and will always take the form of a pk for either
+            # an Image object or a UserUpload object.
+            elif Image.objects.filter(pk=initial).exists():
+                self.current_value = Image.objects.get(pk=initial)
+            elif UserUpload.objects.filter(pk=initial).exists():
+                self.current_value = UserUpload.objects.get(pk=initial)
```
jmsmkn, line 104: Should permission checks also be done here?

app/grandchallenge/components/form_fields.py
```diff
@@ -207,17 +128,38 @@ def __init__(
...
+            # a ComponentInterfaceValue object (in this case the pk is a digit) or
+            # a UserUpload object (then the pk is a UUID).
+            elif (
+                isinstance(initial, int) or initial.isdigit()
+            ) and ComponentInterfaceValue.objects.filter(pk=initial).exists():
+                self.current_value = ComponentInterfaceValue.objects.get(
+                    pk=initial
+                )
+            elif UserUpload.objects.filter(pk=initial).exists():
+                self.current_value = UserUpload.objects.get(pk=initial)
```
jmsmkn, line 153: Should there be permission checks here too?

app/grandchallenge/cases/views.py
amickan: Your latest commit fixes this!

app/grandchallenge/components/form_fields.py
koopmant: Yes, thanks.

app/grandchallenge/cases/widgets.py
koopmant: Yes. I added them. 
The user object permission filtered query is now used in three places: the ModelChoiceField validation, the FlexibleField initialization (for rendering the widget with an existing value) and the WidgetSelectView (when an exisiting value is selected, this renders a hidden input). I don't think I can somehow cache this query for use in those different places, can I?

jmsmkn: There should be no need - if you do `filter_by_permission(queryset=UserUpload.objects.filter(pk=initial), user=..., codename=...)` it will do an index lookup and should be fast.

jmsmkn: Actually, that is a strange construction. You can do `user.has_perm(permission, obj)` and that is very quick. If you need to check multiple objects you can use `ObjectPermissionChecker`.

koopmant: Although I don't have the object itself, only the pk.

jmsmkn: Then `filter_by_permission(queryset=UserUpload.objects.filter(pk=initial), user=..., codename=...).exists()` may work, not sure that you can use `exists` after a union.

It would be more idiomatic to get the object and then check the permissions, both are fast operations.

koopmant: I changed it to how you suggested, except for where I use `get_component_interface_values_for_user`. In this function we check if the user has permission to the parent object (algorithm job, display set, or archive item). I don't see how I could do that after getting the component interface value. But maybe you have an idea?

jmsmkn: Hmm, the way it's changed doesn't seem very Pythonic. The code is currently checking that the object exists, then fetching it, then checking the permissions. Things can go wrong if the state changes in the meantime. Generally, you want to be doing EAFP (See https://realpython.com/python-lbyl-vs-eafp/). You might need a method like:

```python
def get_object_if_allowed(*, model, pk, user, codename):
    try:
        obj = model.objects.get(pk=pk)
    except ObjectDoesNotExist:
        return

    if user.has_perm(codename, obj):
        return obj
    else:
        return
```

You would need some isolated tests on that! Then it can be used with:

```python
    if obj := get_object_if_allowed(*, model=...):
        # do something with obj
```

jmsmkn: Can you link or make a comment to the exact bit of code you're referring to?

koopmant: > Can you link or make a comment to the exact bit of code you're referring to?

For example here: [components/views.py L539-541](https://github.com/comic/grand-challenge.org/pull/3833/files#diff-95b17dfa75c7cc9285d3505525bdfccf7049d361c0db4f3dd12d05f907b2999aR539-R541)

koopmant: We run into the same issue in `serve_component_interface_value` [serving/views.py L127-149](https://github.com/comic/grand-challenge.org/blob/cleanup-flexible-widget/app/grandchallenge/serving/views.py#L127-L149)

koopmant: I updated the function in `serving.views` as well.

Silly question perhaps:
In the function we first check if the object exists at all, then check if it exists in the files the user has access to. However, since that second step also involves filtering by pk, I don't think we need the first step. The code of the second step is this:

```python
def get_component_interface_values_for_user(
    *,
    user,
    civ_pk=None,
    interface=None,
):
    extra_filter_kwargs = {}
    if interface:
        extra_filter_kwargs["interface"] = interface
    if civ_pk:
        extra_filter_kwargs["pk"] = civ_pk

    civs = ComponentInterfaceValue.objects.filter(**extra_filter_kwargs)

    job_query = filter_by_permission(
        queryset=Job.objects.all(),
        user=user,
        codename="view_job",
        accept_user_perms=False,
    )

    job_inputs = (
        job_query.filter(inputs__in=civs)
        .distinct()
        .values_list("inputs__pk", flat=True)
    )
    job_outputs = (
        job_query.filter(outputs__in=civs)
        .distinct()
        .values_list("outputs__pk", flat=True)
    )

    display_sets = (
        filter_by_permission(
            queryset=DisplaySet.objects.all(),
            user=user,
            codename="view_displayset",
            accept_user_perms=False,
        )
        .filter(values__in=civs)
        .distinct()
        .values_list("values__pk", flat=True)
    )

    archive_items = (
        filter_by_permission(
            queryset=ArchiveItem.objects.all(),
            user=user,
            codename="view_archiveitem",
            accept_user_perms=False,
        )
        .filter(values__in=civs)
        .distinct()
        .values_list("values__pk", flat=True)
    )

    return civs.filter(
        pk__in=[
            *job_inputs,
            *job_outputs,
            *display_sets,
            *archive_items,
        ]
    )
```

So if the object does not exist at all, this query is empty early on. I don't think any of the subsequent "steps" (filtering, annotating, filtering) take any time, correct?

koopmant: Okay. I added an early return to `get_component_interface_value_for_user` for this case.

app/tests/serving_tests/test_models.py
```diff
@@ -1,19 +1,71 @@
...
+            model=UserUpload, pk=upload.pk, user=creator, codename=codename
+        )
+        == upload
+    )
+    assert (
+        get_object_if_allowed(
+            model=UserUpload, pk=upload.pk, user=user, codename=codename
+        )
+        is None
+    )
```
koopmant, line 68: @jmsmkn Do you think these tests are sufficient?

app/grandchallenge/serving/models.py
```diff
@@ -52,6 +52,16 @@ class Download(models.Model):
     )
 
 
+def get_object_if_allowed(*, model, pk, user, codename):
```
jmsmkn, line 55: I think this best lives in `core/guardian.py`. Also good to add some documentation to the method.

app/grandchallenge/serving/views.py
```diff
@@ -137,16 +137,16 @@ def serve_component_interface_value(
...
+        pass
+    else:
+        if get_component_interface_values_for_user(
+            user=user, civ_pk=civ.pk
+        ).exists():
+            return protected_storage_redirect(
+                name=civ.file.name, creator=user, component_interface_value=civ
+            )
+
+    raise Http404("No ComponentInterfaceValue found.")
```
jmsmkn, line 149: It's best not to rely on fall-through so we should be explicit with our denials, a 404 message is not necessary so we can omit it for consistency.

```suggestion
        raise Http404

    if get_component_interface_values_for_user(
        user=user, civ_pk=civ.pk
    ).exists():
        return protected_storage_redirect(
            name=civ.file.name, creator=user, component_interface_value=civ
        )

    raise Http404
```

app/tests/serving_tests/test_models.py
jmsmkn: Yes I think they're good!

jmsmkn (commented):
Looks good! Couple of small things and I would like @amickan to take a final look.

app/grandchallenge/serving/models.py
koopmant: Thanks, moved it!

app/grandchallenge/serving/views.py
koopmant: Good to know!

amickan (approved):
LGTM! Great work!

PR #3834, opened 2025-02-17
Onboarding Tasks Creation

chrisvanrun:
Part of the pitch:
- https://github.com/DIAGNijmegen/rse-roadmap/issues/339

This PR adds the actual task creation when a challenge gets created.

## Showcase
image

### DIFF
--- a/app/grandchallenge/challenges/apps.py
+++ b/app/grandchallenge/challenges/apps.py
@@ -27,3 +27,6 @@ class ChallengesConfig(AppConfig):
 
     def ready(self):
         post_migrate.connect(init_reviewers_group, sender=self)
+
+        # noinspection PyUnresolvedReferences
+        import grandchallenge.challenges.signals  # noqa: F401
--- a/app/grandchallenge/challenges/signals.py
+++ b/app/grandchallenge/challenges/signals.py
@@ -8,55 +8,59 @@ from grandchallenge.challenges.models import Challenge, OnboardingTask
 @receiver(post_save, sender=Challenge)
 def create_onboarding_tasks(sender, instance, created, **__):
     if created:  # Ensures tasks are only created on first save
-        OnboardingTask.objects.bulk_create(
-            [
-                OnboardingTask(
-                    title="Create Phases",
-                    description="Define, name and create the different phases of the challenge.",
-                    responsible_party=OnboardingTask.ResponsiblePartyChoices.CHALLENGE_ORGANIZERS,
-                    deadline=instance.created + timedelta(weeks=1),
-                ),
-                OnboardingTask(
-                    title="Define Inputs and Outputs",
-                    description="To support, communicate the required input and output data formats for participant's algorithms.",
-                    responsible_party=OnboardingTask.ResponsiblePartyChoices.CHALLENGE_ORGANIZERS,
-                    deadline=instance.created + timedelta(weeks=3),
-                ),
-                OnboardingTask(
-                    title="Create Archives",
-                    description="Create an archive per algorithm-type phase for the challenge.",
-                    responsible_party=OnboardingTask.ResponsiblePartyChoices.SUPPORT,
-                    deadline=instance.created + timedelta(weeks=3),
-                ),
-                OnboardingTask(
-                    title="Upload Data",
-                    description="Prepare and upload necessary datasets to archives, contact support if no archives are available.",
-                    responsible_party=OnboardingTask.ResponsiblePartyChoices.CHALLENGE_ORGANIZERS,
-                    deadline=instance.created + timedelta(weeks=5),
-                ),
-                OnboardingTask(
-                    title="Example Algorithm",
-                    description="Implement and document a baseline example algorithm for participants to use as a reference.",
-                    responsible_party=OnboardingTask.ResponsiblePartyChoices.CHALLENGE_ORGANIZERS,
-                    deadline=instance.created + timedelta(weeks=6, seconds=0),
-                ),
-                OnboardingTask(
-                    title="Evaluation Method",
-                    description="Implement and document the evaluation method for assessing participant submissions.",
-                    responsible_party=OnboardingTask.ResponsiblePartyChoices.CHALLENGE_ORGANIZERS,
-                    deadline=instance.created + timedelta(weeks=6, seconds=1),
-                ),
-                OnboardingTask(
-                    title="Scoring",
-                    description="Configure the leaderboard scoring to accurately interpret the evaluation results.",
-                    responsible_party=OnboardingTask.ResponsiblePartyChoices.CHALLENGE_ORGANIZERS,
-                    deadline=instance.created + timedelta(weeks=6, seconds=2),
-                ),
-                OnboardingTask(
-                    title="Test Evaluation",
-                    description="Run test evaluations using sample submissions to ensure the scoring system and evaluation method function correctly before launching the challenge.",
-                    responsible_party=OnboardingTask.ResponsiblePartyChoices.CHALLENGE_ORGANIZERS,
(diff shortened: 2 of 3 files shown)


app/grandchallenge/challenges/signals.py
```diff
@@ -0,0 +1,65 @@
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+from django.utils.timezone import timedelta
+
+from grandchallenge.challenges.models import Challenge, OnboardingTask
+
+
+@receiver(post_save, sender=Challenge)
```
jmsmkn, line 8: Signals should only be used as a last resort and are not necessary here. This is to prevent business logic being located in multiple locations to make it easier for future developers. See https://lincolnloop.com/insights/django-anti-patterns-signals/

There are already post save methods for `Challenges` in `models.py` and they should be extended with a new class method here: https://github.com/comic/grand-challenge.org/blob/bcaa9ed57915b07733eea065e7d17284f2218630/app/grandchallenge/challenges/models.py#L505

app/grandchallenge/challenges/signals.py
```diff
@@ -0,0 +1,65 @@
...
+                title="Test Evaluation",
+                description="Run test evaluations using sample submissions to ensure the scoring system and evaluation method function correctly before launching the challenge.",
+                responsible_party=OnboardingTask.ResponsiblePartyChoices.CHALLENGE_ORGANIZERS,
+                deadline=instance.created + timedelta(weeks=6, seconds=3),
+            ),
+        ]
+
+        for t in tasks:
+            t.challenge = instance
+            t.created = instance.created
```
jmsmkn, line 64: `created` should not be manually set. It is better to use `OnboardingTask.create(challenge=self, ...)` in the list above here.

app/grandchallenge/challenges/apps.py
```diff
@@ -27,3 +27,6 @@ class ChallengesConfig(AppConfig):
 
     def ready(self):
         post_migrate.connect(init_reviewers_group, sender=self)
+
+        # noinspection PyUnresolvedReferences
+        import grandchallenge.challenges.signals  # noqa: F401
```
jmsmkn, line 32: This should not be necessary when using the save method.

jmsmkn (changes requested):
Looks good but would prefer that signals are not used. I have not reviewed the content of the tasks, that is for @miriam-groeneveld

app/grandchallenge/challenges/signals.py
chrisvanrun: Thanks for the link to the blog. It was a good read and I agree with its argument.

app/tests/challenges_tests/test_views.py
```diff
@@ -554,6 +554,8 @@ def get():
 @pytest.mark.django_db
 def test_onboarding_task_list_view_permissions(client):
     ch = ChallengeFactory()
+    ch.onboarding_tasks.all().delete()
```
chrisvanrun, line 557: Long live sanity tests a few lines down.

app/grandchallenge/challenges/signals.py
chrisvanrun: It indeed should not have. This was a left-over after I had neglected to account for permissions in the save() when using the bulk_create.

app/grandchallenge/challenges/models.py
```diff
@@ -664,6 +665,65 @@ def create_default_pages(self):
...
+            challenge=self,
+            title="Create Archives",
+            description="Create an archive per algorithm-type phase for the challenge.",
+            responsible_party=OnboardingTask.ResponsiblePartyChoices.SUPPORT,
+            deadline=self.created + timedelta(weeks=3),
+        )
+        OnboardingTask.objects.create(
+            challenge=self,
+            title="Upload Data",
+            description="Prepare and upload necessary datasets to archives, contact support if no archives are available.",
```
jmsmkn, line 693: Do the challenge organisers see that the archives need to be created by support first?

app/grandchallenge/challenges/models.py
```diff
@@ -664,6 +665,65 @@ def create_default_pages(self):
...
+            responsible_party=OnboardingTask.ResponsiblePartyChoices.CHALLENGE_ORGANIZERS,
+            deadline=self.created + timedelta(weeks=5),
+        )
+        OnboardingTask.objects.create(
+            challenge=self,
+            title="Example Algorithm",
+            description="Implement and document a baseline example algorithm for participants to use as a reference.",
```
jmsmkn, line 700: Do they know that support will provide them with a template for this and the evaluation containers?

jmsmkn (dismissed):
Looks good, couple of questions about the tasks that Miriam should also look at but this is fine for me.

app/grandchallenge/challenges/models.py
chrisvanrun: They do not see the support task ticket for this.

app/grandchallenge/challenges/models.py
chrisvanrun: I hope that by the time this plays they have had a onboarding meeting.

Maybe a new 'task' for the orgs: receive challenge pack with examples and a task for support: create challenge pack?

tbh I think we'll be revisiting the task definitions a few times the next few months to get it right.

amickan: I think it would be enough to add something along the lines of:
```suggestion
            description="Implement and document a baseline example algorithm for participants to use as a reference. Use the provided challenge pack as starting point for this.",
```

chrisvanrun: Same should hold for the evaluation method descr.

miriam-groeneveld: I would suggest adding the onboarding meeting as a task, that way everybody can see in the admin if that has happened already.

miriam-groeneveld (commented):
A few comments, indeed might be revisited a few times.

chrisvanrun:
Thanks everyone. I'll revise the questions tomorrow and will ask Miriam for the final approval.

jmsmkn (dismissed):

app/grandchallenge/challenges/models.py
```diff
@@ -664,6 +665,65 @@ def create_default_pages(self):
             permission_level=Page.ALL,
         )
 
+    def create_default_onboarding_tasks(self):
+        OnboardingTask.objects.create(
+            challenge=self,
+            title="Create Phases",
+            description="Define, name and create the different phases of the challenge.",
```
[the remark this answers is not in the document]
miriam-groeneveld: Sorry, my bad for not being specific, I meant the description. Define, name and create all seem to be the same thing (more or less)

chrisvanrun:
> Tests need to be updated, perhaps they are too sensitive?

Not sure. It's a bit of a hassle now but I would def like to know in the future if any onboarding task went missing. I've just limited to the titles now but forgot to update them pre-lunch =P

app/grandchallenge/challenges/models.py
```diff
@@ -664,6 +665,75 @@ def create_default_pages(self):
...
+            challenge=self,
+            title="Create Archives",
+            description="Create an archive per algorithm-type phase for the challenge.",
+            responsible_party=OnboardingTask.ResponsiblePartyChoices.SUPPORT,
+            deadline=self.created + timedelta(weeks=3),
+        )
+        OnboardingTask.objects.create(
+            challenge=self,
+            title="Upload Data to Archives",
+            description="Archives must be created by support before preparing and uploading the necessary datasets.",
```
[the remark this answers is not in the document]
chrisvanrun: You're somewhat describing the old description. I'll see if I can combine them somewhat.

miriam-groeneveld (approved):

PR #3841, opened 2025-02-19
Adds algorithm create redirect view

### LINKED ISSUE #3671: Create algorithm from challenge form visibility
Currently, it's only a small link in a help text. We could improve this by having a clear button on the form that links to the algorithm-creation-for-challenge form.

I've seen reports of multiple users confused by this.

jmsmkn:
Users are presented with the phases that they have the `"create_phase_submission"` which is currently public phases for the participants group, or all phases for the admins group. 

Screenshot 2025-02-19 at 17 11 18

The check on whether the phase is open is very complex and in a mixin that cannot really be re-used here, so we just clean the phase to see if it is open for non-admins. The full check is then done on the redirect but the errors are not as nice.

Screenshot 2025-02-19 at 17 11 25

If the phase is open then the user will end up on the usual page.

Screenshot 2025-02-19 at 17 11 33

Those with custom algorithm permissions will see the extra link.

Screenshot 2025-02-19 at 17 15 39

Closes #3671

### DIFF
--- a/app/grandchallenge/algorithms/forms.py
+++ b/app/grandchallenge/algorithms/forms.py
@@ -41,7 +41,7 @@ from django.urls import Resolver404, resolve
 from django.utils.functional import cached_property
 from django.utils.html import format_html
 from django.utils.text import format_lazy
-from django_select2.forms import Select2MultipleWidget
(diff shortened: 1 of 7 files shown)


amickan (dismissed):
LGTM!

PR #3853, opened 2025-02-25
Split off the Ground Truth page

chrisvanrun:
Part of the pitch:

A quick pre-lude:
- split the ground truth from being a tab in the the reader study detail to its own view
- add a warning when the reader study is not configured for educational purposes

Other than above, I did not change the content of the page.

Considered to refactor all reader study views to have the configuration side-panel, much like the algorithms. But have put that out of scope (for now).

## Showcase
afbeelding

### DIFF
--- a/app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_ground_truth.html
+++ b/app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_ground_truth.html
@@ -20,7 +20,7 @@
 <h2>Ground Truth</h2>
 {% if not object.is_educational %}
 <div class="alert alert-warning" role="alert">
-        <i class="fa fa-info-triangle pr-1" aria-hidden="true"></i>
+        <i class="fa fa-exclamation-triangle pr-1" aria-hidden="true"></i>
         This reader study is currently not configured as educational: any ground truth is not utilized.
         Mark the reader study as educational via the <a href="{% url 'reader-studies:update' slug=object.slug %}">settings</a>.
     </div>
--- a/app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_ground_truth.html
+++ b/app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_ground_truth.html
@@ -1,0 +1,66 @@
+{% extends "base.html" %}
+{% load url %}
+{% load static %}
+
+{% block title %}
+    Ground Truth - {{ object }} - {{ block.super }}
+{% endblock %}
+
+{% block breadcrumbs %}
+    <ol class="breadcrumb">
+        <li class="breadcrumb-item"><a href="{% url 'reader-studies:list' %}">Reader
+            Studies</a></li>
+        <li class="breadcrumb-item"><a href="{{ object.get_absolute_url }}">{{ object }}</a></li>
+        <li class="breadcrumb-item active" aria-current="page">Ground Truth</li>
+    </ol>
+{% endblock %}
+
+{% block content %}
+
+<h2>Ground Truth</h2>
+
+<p>
+    If you wish to assess the performance of the readers you
+    can upload the CSV file containing the ground truth for
+    this reader study here.
+</p>
+<p>
+    The first row of the CSV file must contain the headings
+    <code>case</code>, followed by the question text for
+    each of the questions in the reader study (excluding those
+    of type <code>heading</code>).
+</p>
+<p>
+    The consecutive lines contain the id per case.
+    Then include the answer for the question
+    corresponding to the question text defined in the header for that
+    column. All answers must be enclosed in single quotes. Strings
+    must be enclosed in both single and double quotes, i.e.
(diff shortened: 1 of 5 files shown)


jmsmkn:
> Considered to refactor all reader study views to have the configuration side-panel, much like the algorithms. But have put that out of scope (for now).

It is not the case for algorithms, we do not have such a feature anywhere. Please do not do large refactoring like that without discussion in a pitch first.

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_ground_truth.html
```diff
@@ -0,0 +1,74 @@
...
+        <li class="breadcrumb-item"><a href="{% url 'reader-studies:list' %}">Reader
+            Studies</a></li>
+        <li class="breadcrumb-item"><a href="{{ object.get_absolute_url }}">{{ object }}</a></li>
+        <li class="breadcrumb-item active" aria-current="page">Ground Truth</li>
+    </ol>
+{% endblock %}
+
+{% block content %}
+
+<h2>Ground Truth</h2>
```
jmsmkn, line 20: The indentation doesn't seem correct, should be 4 spaces over as it's inside a block.

app/grandchallenge/reader_studies/urls.py
```diff
@@ -64,6 +65,11 @@
         AnswersRemoveForUser.as_view(),
         name="answers-remove",
     ),
+    path(
+        "<slug>/ground-truth/",
```
jmsmkn, line 69: Missing a type

```suggestion
        "<slug:slug>/ground-truth/",
```

app/grandchallenge/reader_studies/views.py
```diff
@@ -175,6 +175,28 @@ def form_valid(self, form):
         return response
 
 
+class ReaderStudyGroundTruth(
+    LoginRequiredMixin, ObjectPermissionRequiredMixin, DetailView
+):
+    model = ReaderStudy
+    permission_required = (
+        f"{ReaderStudy._meta.app_label}.change_{ReaderStudy._meta.model_name}"
```
jmsmkn, line 183: Using the meta stuff means that this isn't searchable, I think that we should just bite the bullet and use the explicit permissions.

app/grandchallenge/reader_studies/views.py
```diff
@@ -175,6 +175,28 @@ def form_valid(self, form):
...
+    )
+    raise_exception = True
+    template_name = "reader_studies/readerstudy_ground_truth.html"
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context.update(
+            {
+                "example_ground_truth": self.object.get_example_ground_truth_csv_text(
```
jmsmkn, line 192: Can this be removed from the context of the detail view?

app/tests/reader_studies_tests/test_views.py
```diff
@@ -931,3 +931,47 @@ def test_display_set_upload_corrupt_image(
...
+
+    response = get_view_for_user(
+        client=client,
+        viewname="reader-studies:ground-truth",
+        reverse_kwargs={"slug": rs.slug},
+        user=editor,
+    )
+
+    educational_warning = (
+        "This reader study is currently not configured as educational"
```
jmsmkn, line 961: We could hide the link if it is not educational?

chrisvanrun: Mhhh. That would also hide the feature. Given that most people don't have the best track record of reading documents or feature lists I am generally fond of showing a disabled action unless the UI clutter calls for it.

Then again, this is really only applicable to ppl who want to have an educational reader study. Which is a minority of the reader studies and the feature **is** documented here: https://grand-challenge.org/documentation/creating-questions/#educational-reader-studies .

We could then also hide the leaderboard link.

Locking away the deletion of ground truth when a reader study is not not educational seems an OK side effect. I think?

chrisvanrun:
> It is not the case for algorithms, we do not have such a feature anywhere. Please do not do large refactoring like that without discussion in a pitch first.

Definitively! I thought algorithms implemented it already but it actually are the Challenge admin pages (i.e. phase settings, general config, et cetera) that do that.

app/tests/reader_studies_tests/test_views.py
chrisvanrun: Done! Both are now hidden.

app/grandchallenge/reader_studies/views.py
chrisvanrun: I've bitten all the local bullets and have updated all the permissions in this file. Given this PR is reasonably small I thought it justified.

chrisvanrun:
Given the changes were out of style with the other apps and it was suggested to update them, I've taken the oppertunity to update the style of permissions and route param typing within the reader study app.

jmsmkn (commented):
LGTM, thanks!

jmsmkn (approved):

PR #3857, opened 2025-02-25
Postpaid payment status

koopmant:
Add new invoice payment status choice "postpaid".

Related to https://github.com/DIAGNijmegen/rse-roadmap/issues/373


jmsmkn:
I think this misses the requirements that we need to track the status of the invoice within our own administration, we would that have two places to track the payment status - the choice field and the dates. The dates are hardly used as we are not forced to in the forms. An extension to this pitch would be to force that!

PR #3860, opened 2025-02-27
Change overflow of register form to avoid showing scroll bars

### LINKED ISSUE #3845: Scroll bars displayed on the register page
They should be able to be hidden, this is due to the honeypot field.

Image

ammar257ammar:
closes #3845

### DIFF
--- a/app/grandchallenge/core/static/css/core.css
+++ b/app/grandchallenge/core/static/css/core.css
@@ -242,9 +242,3 @@ a[aria-expanded="false"] .fa-chevron-down {
 .htmx-request.htmx-indicator {
     display: inline-block;
 }
-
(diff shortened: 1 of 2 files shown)


jmsmkn (changes requested):
This matching is going to be too brittle as we are not in control of those actions. The form should be modified to include the bootstrap classes to solve this which should be possible with a crispy form helper.

ammar257ammar:
Looking into it, since the allauth template does not use crispy forms, the crispy form helper did not work. It was even simpler to achieve by overriding the form element template, similar to how other element templates were overridden, and add the Bootstrap css classes there.

ammar257ammar:
Tested on Brave and Safari and the scrollbar is gone.

jmsmkn (approved):

PR #3861, opened 2025-02-27
Invoice email alerts

koopmant:
Add invoice email alerts

If we have not received payment for an invoice after four weeks, email alerts will be sent out to the challenge admins and site managers. 
We also send an email to the challenge admins when an invoice has been sent out (i.e. when it is set to `issued`).

Related to https://github.com/DIAGNijmegen/rse-roadmap/issues/379


koopmant:
to be reviewed after https://github.com/comic/grand-challenge.org/pull/3855 has been merged

app/grandchallenge/invoices/templates/invoices/partials/challenge_invoice_issued_notification_email.md
koopmant, line 1: Should I send a different message to the challenge admins? Right now they get the same message as the invoice contact person, but that might lead to some confusion...

app/grandchallenge/core/context_processors.py
```diff
@@ -41,6 +46,10 @@ def challenge(request):
         )
         .with_overdue_status()
         .status_aggregates,
+        "num_outstanding_invoices": Invoice.objects.filter(challenge=challenge)
+        .exclude(payment_type=PaymentTypeChoices.COMPLIMENTARY)
+        .filter(payment_status=PaymentStatusChoices.ISSUED)
```
amickan, line 51: Shouldn't we include all but paid invoices here?

app/grandchallenge/invoices/views.py
```diff
@@ -0,0 +1,52 @@
...
+        queryset = queryset.filter(challenge=self.request.challenge)
+        return queryset
+
+
+class InvoiceDetail(
+    LoginRequiredMixin, ObjectPermissionRequiredMixin, DetailView
+):
+    model = Invoice
+    permission_required = "change_challenge"
+    template_name_suffix = "_detail"
```
amickan, line 47: I don't think you need this suffix here. Should be inherited from `DetailView`

app/grandchallenge/invoices/views.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class InvoiceList(
+    LoginRequiredMixin,
+    ObjectPermissionRequiredMixin,
+    ListView,
+):
+    model = Invoice
+    permission_required = "change_challenge"
+    template_name_suffix = "_list"
```
amickan, line 16: I think this is also not necessary, right?

app/config/settings.py
```diff
@@ -1133,6 +1133,7 @@ def sentry_before_send(event, hint):
 
 CHALLENGES_DEFAULT_ACTIVE_MONTHS = 12
 CHALLENGE_ONBOARDING_TASKS_OVERDUE_SOON_CUTOFF = timedelta(days=7)
+CHALLENGE_INVOICE_OUTSTANDING_CUTOFF = timedelta(weeks=4)
```
amickan, line 1136: Probably good to add the timeunit to the name of this variable for clarity

app/grandchallenge/core/context_processors.py
amickan: Ah no, I think this is correct. It doesn't make sense to include initialized and requested here since those haven't been sent to the challenge orgs yet of course. Sorry!

app/grandchallenge/invoices/emails.py
```diff
@@ -0,0 +1,88 @@
...
+            get_unsubscribe_link=lambda *_, **__: None
+        )
+        contact_user = namedtuple(
+            "User", ["username", "email", "user_profile"]
+        )(
+            username=username,
+            email=invoice.contact_email,
+            user_profile=user_profile,
+        )
+        recipients.append(contact_user)
```
amickan, line 31: Should we check if there is a user with this email address in our system and if so use that?

app/grandchallenge/invoices/tasks.py
```diff
@@ -0,0 +1,36 @@
...
+        payment_type__in=[
+            Invoice.PaymentTypeChoices.PREPAID,
+            Invoice.PaymentTypeChoices.POSTPAID,
+        ],
+        payment_status=Invoice.PaymentStatusChoices.ISSUED,
+        issued_on__lt=_now - settings.CHALLENGE_INVOICE_OUTSTANDING_CUTOFF,
+    )
+    for invoice in outstanding_invoices:
+        send_challenge_outstanding_invoice_reminder(invoice)
+
```
amickan, line 28: When is this task run? I suppose that this is a periodic task, but I don't see it added to the beat schedule.

app/grandchallenge/invoices/tasks.py
```diff
@@ -0,0 +1,36 @@
...
+
+@acks_late_micro_short_task
+@transaction.atomic
+def send_challenge_invoice_issued_notification_emails(*, pk):
+    from grandchallenge.invoices.models import Invoice
+
+    invoice = Invoice.objects.get(pk=pk)
+    send_challenge_invoice_issued_notification(invoice)
```
amickan, line 36: Does this need to be done asynchronously? It's just sending a handful of emails at most, right?

app/grandchallenge/challenges/emails.py
```diff
@@ -222,3 +224,57 @@ def send_onboarding_task_due_reminder(challenge, task_info):
...
+        user_profile = namedtuple("UserProfile", ["get_unsubscribe_link"])(
+            get_unsubscribe_link=lambda *_, **__: None
+        )
+        contact_user = namedtuple(
+            "User", ["username", "email", "user_profile"]
+        )(
+            username=username,
+            email=invoice.contact_email,
+            user_profile=user_profile,
+        )
```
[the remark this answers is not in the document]
amickan: I suggested below to maybe check if there is a User in our systems with that email, but you still need this here for cases where there is no user instance.

app/grandchallenge/invoices/templates/invoices/partials/challenge_invoice_issued_notification_email.md
amickan: What confusion do you foresee? I think it's fine to send this to challenge admins.

app/config/settings.py
jmsmkn: I don't think this is necessary as the value is a `timedelta`, follows the pattern of `CHALLENGE_ONBOARDING_TASKS_OVERDUE_SOON_CUTOFF` above.

app/grandchallenge/invoices/tasks.py
jmsmkn: This is to decouple the saving and email sending as sometime these take a while and you can get lock contention (for instance with the notification email creation on forum posts).

app/grandchallenge/invoices/views.py
```diff
@@ -0,0 +1,52 @@
...
+    LoginRequiredMixin, ObjectPermissionRequiredMixin, DetailView
+):
+    model = Invoice
+    permission_required = "change_challenge"
+    template_name_suffix = "_detail"
+    raise_exception = True
+    login_url = reverse_lazy("account_login")
+
+    def get_permission_object(self):
+        return self.request.challenge
```
jmsmkn, line 52: The permission object should be the `self.object.challenge`, otherwise you could view the invoice on another challenge domain if you knew the pk.

It would be better if we used the `view_invoice` permission which would be automatically assigned to the invoice challenge admins group. Could you make that change in another PR and then use the object based permissions for these two views?

app/grandchallenge/invoices/templates/invoices/partials/challenge_invoice_issued_notification_email.md
koopmant: The message is very much directed at the person who should handle the invoice and will receive the message from radboud. The other admins won't receive that message (right?) so might be a little confused about what they should do. It's just an fyi for them, and I think most of them will understand. I thought I could perhaps explain it in a separate message to them. But if you think it's fine like this I won't bother.

app/grandchallenge/invoices/views.py
koopmant: Right, I wanted to remove this but forgot. Thanks!

app/grandchallenge/invoices/views.py
koopmant: correct.

app/grandchallenge/invoices/emails.py
koopmant: Yes, I'll add that.

app/grandchallenge/invoices/tasks.py
koopmant: I did not add it to the schedule indeed. Misunderstanding on my part. 
I'll schedule it!

app/grandchallenge/invoices/views.py
koopmant: new PR is here: https://github.com/comic/grand-challenge.org/pull/3885

I copied this from other parts in the codebase. We don't appear to use `self.object.challenge` anywhere. Shall I create another PR to fix usages of `self.request.challenge`?

jmsmkn: I think that's correct as the other views should be implemented with the correct object permissions (I looked through them again yesterday). For some things you need to use `self.request.challenge` e.g. with a create view you have no `self.object` but need to create a new page for the challenge, say.

koopmant: Yes, I figured as much after looking through them. Thank for clarifying!

app/grandchallenge/invoices/emails.py
koopmant: this should be good now

app/grandchallenge/invoices/tasks.py
koopmant: added it to the schedule

koopmant:
@jmsmkn in the next PR I'm changing the views again. Since this one has not been merged yet, would you like me to remove the views from this one?

jmsmkn:
Sure

koopmant:
> Sure

Done!

app/grandchallenge/invoices/tasks.py
```diff
@@ -0,0 +1,36 @@
...
+from django.utils.timezone import now
+
+from grandchallenge.core.celery import acks_late_micro_short_task
+from grandchallenge.invoices.emails import (
+    send_challenge_invoice_issued_notification,
+    send_challenge_invoice_overdue_reminder,
+)
+
+
+@acks_late_micro_short_task
```
jmsmkn, line 12: Maybe put it on the 2xlarge queue, just in case a lot of emails need to be sent.

app/grandchallenge/invoices/models.py
```diff
@@ -108,6 +125,16 @@ def save(self, *args, **kwargs):
         super().save(*args, **kwargs)
         if adding:
             self.assign_permissions()
+        if (
+            self.payment_type != PaymentTypeChoices.COMPLIMENTARY
+            and self.has_changed("payment_status")
```
jmsmkn, line 130: Does this work for invoice that are being created with payment status issued?

app/tests/invoices_tests/test_tasks.py
```diff
@@ -0,0 +1,313 @@
...
+        )
+
+    mocker.patch(
+        "grandchallenge.invoices.tasks.now",
+        return_value=_fixed_now,
+    )
+
+    send_challenge_invoice_overdue_reminder_emails()
+
+    if send_email:
```
jmsmkn, line 89: Ideally tests do not have any logic in them as this can cause issues. Try to follow the arrange, act, assert pattern, see https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/. This does mean that things can be duplicated in tests, but that is okay as it should be very straightforward code. If you need to duplicate a huge arrange section then you can use a pytest fixture.

koopmant: Thanks for the reminder. Should be much better now! 
And the duplication wasn't too bad.

app/grandchallenge/invoices/models.py
koopmant: Sharp, it did not. Fixed it and added an accompanying test.

PR #3864, opened 2025-03-05
Set time limit, gpu type and memory for external evaluations

amickan:
Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/463

### DIFF
--- a/app/grandchallenge/evaluation/tasks.py
+++ b/app/grandchallenge/evaluation/tasks.py
@@ -54,9 +54,11 @@ def create_evaluation(*, submission_pk, max_initial_jobs=1):  # noqa: C901
     if submission.phase.external_evaluation:
         external_evaluation, created = Evaluation.objects.get_or_create(
             submission=submission,
-            time_limit=submission.phase.evaluation_time_limit,
-            requires_gpu_type=submission.phase.evaluation_requires_gpu_type,
-            requires_memory_gb=submission.phase.evaluation_requires_memory_gb,
+            defaults={
+                "time_limit": submission.phase.evaluation_time_limit,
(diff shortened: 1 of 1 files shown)


app/grandchallenge/evaluation/tasks.py
```diff
@@ -53,7 +53,10 @@ def create_evaluation(*, submission_pk, max_initial_jobs=1):  # noqa: C901
 
     if submission.phase.external_evaluation:
         external_evaluation, created = Evaluation.objects.get_or_create(
-            submission=submission
+            submission=submission,
+            time_limit=submission.phase.evaluation_time_limit,
+            requires_gpu_type=submission.phase.evaluation_requires_gpu_type,
+            requires_memory_gb=submission.phase.evaluation_requires_memory_gb,
```
jmsmkn, line 59: Do we want to create a new Evaluation object each time here given the values may be ignored? Should these values be in `defaults` instead?

amickan: We'd only be creating new evaluation objects here if the values on the phase change, and those are unlikely to change for external phases, right? But you're right that it might be better to be defensive here, just in case. 

What do you mean by putting these values in `defaults` exactly? Adding defaults to these model field definitions on the `ComponentJob` model, or is there a different way to define default that I'm not aware of?

jmsmkn: You can set defaults in `get_or_create` that are only used for the create and not the get: https://docs.djangoproject.com/en/5.1/ref/models/querysets/#get-or-create

amickan: Ah that's an elegant solution, thanks!

jmsmkn (approved):

PR #3868, opened 2025-03-06
Remove results template

jmsmkn:
See https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/467


amickan (approved):
LGTM!

There are still quite some algorithms that use the generic json result CI. Of the highlighted algorithms at least 3 have a rendering template for their json result in place. Should we contact the algorithm developers to update their interfaces and algorithm? If at all, I would only do this for the highlighted algorithms, of course.

PR #3871, opened 2025-03-07
Allow additional evaluation inputs

amickan:
This adds input fields for all configured inputs of a phase to the SubmissionForm, and the necessary code to process those inputs and attach them to the linked evaluation object.


app/grandchallenge/cases/models.py
```diff
@@ -174,12 +174,12 @@ def update_status(
...
                 action_object=self,
             )
-        from grandchallenge.algorithms.models import Job
+        from grandchallenge.components.models import ComponentJob
 
         if (
             self.error_message
             and linked_object
-            and isinstance(linked_object, Job)
+            and isinstance(linked_object, ComponentJob)
```
jmsmkn, line 182: Are we certain that these messages will not leak any information to the challenge participants?

app/grandchallenge/evaluation/admin.py
```diff
@@ -112,7 +112,16 @@ def challenge_forge_json(obj):
 def reevaluate_submissions(modeladmin, request, queryset):
     """Creates a new evaluation for an existing submission"""
     for submission in queryset:
-        submission.create_evaluation()
+        if submission.phase.inputs.exists():
```
jmsmkn, line 115: Is there an API difference here - the phase `inputs` are ComponentInterfaces, whereas for ComponentJobs `inputs` are ComponentInterfaceValues?

app/grandchallenge/evaluation/models.py
```diff
@@ -1519,7 +1520,7 @@ def save(self, *args, **kwargs):
                     send_action=False,
                 )
 
-    def create_evaluation(self):
+    def create_evaluation(self, *, additional_inputs=None):
```
jmsmkn, line 1523: Should we explicitly require the caller setting `additional_inputs` to `None`? Do we check that the `additional_inputs` are set when they are required?

app/grandchallenge/evaluation/models.py
```diff
@@ -1850,6 +1860,52 @@ def inputs_complete(self):
...
             return False
 
+    @cached_property
+    def additional_inputs_complete(self):
+        phase_input_slugs = self.submission.phase.inputs.values_list(
+            "slug", flat=True
+        )
+        return self.inputs.filter(
+            interface__slug__in=phase_input_slugs
+        ).count() == len(phase_input_slugs)
```
jmsmkn, line 1870: Does this work if the evaluation is rescheduled for algorithm submission evaluations where there are many inputs of the same type?

app/grandchallenge/evaluation/tasks.py
```diff
@@ -49,8 +49,15 @@ def prepare_and_execute_evaluation(
...
+    if not evaluation.additional_inputs_complete:
+        logger.info("Nothing to do, inputs are still being validated.")
+        return
+
     if evaluation.status != evaluation.VALIDATING_INPUTS:
-        raise RuntimeError("Evaluation is not ready to be executed.")
+        # this task can be called multiple times with complete inputs,
+        # and might have been queued for execution already, so ignore
+        logger.info("Evaluation has already been scheduled for execution.")
+        return
```
jmsmkn, line 60: Are you certain that this is the correct behaviour and not only works for the happy path? Previously we raised the exception here as it indicated that something went wrong with the celery task ordering.

app/grandchallenge/evaluation/templates/evaluation/submission_form.html
```diff
@@ -84,7 +84,11 @@ <h3>Create a new submission</h3>
             {% include 'evaluation/partials/algorithm_interfaces_information.html' with phase=phase form=form %}
         {% endif %}
 
-        {% crispy form %}
+        <form method="POST">
+            {% csrf_token %}
+            {{ form|crispy }}
+            <input type="submit" name="save" value="Submit" class="btn btn-primary">
+        </form>
```
jmsmkn, line 91: Shouldn't all of the form setup, csrf token and button be handled by crispy forms?

app/tests/conftest.py
```diff
@@ -887,3 +890,145 @@ def submission_with_model_for_optional_inputs(jobs_for_optional_inputs):
         civs_for_interface2=jobs_for_optional_inputs.civs_for_interface2,
         output_civs=jobs_for_optional_inputs.output_civs,
     )
+
+
+class PhaseWithInputs(NamedTuple):
```
jmsmkn, line 895: Large fixtures like this should be avoided if possible as you can create weird dependencies between tests. The fixture anyway does not need to live in conftest where it is available to everything and can instead be confined to the module where it is used.

jmsmkn (commented):
Looks good, some questions mostly for my understanding.

app/grandchallenge/evaluation/admin.py
amickan: Yes. This is about whether or not to allow creating a new Evaluation (ComponentJob). In the admin we can only allow this if the phase has no input sockets  (ie. ComponentInterfaces) configured. If it does, re-evaluation needs to happen through the challenge UI because we'll need to also provide the values for these input sockets.

app/grandchallenge/cases/models.py
jmsmkn: I would probably err on the side of caution, just in case they are private or we change things later so that all inputs somehow go through this route.

app/grandchallenge/evaluation/templates/evaluation/submission_form.html
amickan: I thought the same, but couldn't get it to work. The additional inputs (which are added to form.fields on form initialization) did not render when using `{% crispy form %}`. I must admit that I was a bit lazy and just copied the implementation from the algorithm form (where we also have dynamic form fields). If you have a hunch why the crispy form filter isn't working for those fields, I'm all ears.

app/grandchallenge/cases/models.py
amickan: Ok, are you suggesting to replace the specific error message with a generic "input validation failed" instead? But the challenge admins should probably be able to see the full error message. Would it be an idea to hide the error messages of evaluations from participants altogether?

app/tests/conftest.py
amickan: I will move it to the evaluation tests. Repeating all the set-up for every test that uses this also seems suboptimal. What would you suggest doing instead?

app/grandchallenge/cases/models.py
jmsmkn: Generic would be fine, likely we're not going to hit this as likely we're just using JSON configuration to these things which are going to be validated on the client side too.

amickan: That's true. Thinking about whether to log the error to sentry so that we have it on record - is that worth it in your eyes ?We'd need to take care to only do this for evaluations or else we would spam sentry.

app/grandchallenge/evaluation/templates/evaluation/submission_form.html
jmsmkn: Without looking at it could it be that the helper is added before the fields are present?

app/tests/conftest.py
jmsmkn: Yes sometimes it's necessary, but usually it is better to limit the number of things that are created for a test. Otherwise things get slow as you're creating a lot of (potentially unnecessary) objects, and if you need to change the fixture you do not know if you are breaking an implicit assumption of a test that is using it.

For this reason we moved away from things like the two challenge sets, and wrote tests that were more isolated instead, or just duplicated a lot of factories.

app/config/urls/challenge_subdomain.py
```diff
@@ -14,6 +14,10 @@
     path(
         "", include("grandchallenge.well_known.urls", namespace="well-known")
     ),
+    path(
+        "cases/",
```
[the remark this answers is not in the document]
amickan: This is needed for rendering the additional input fields for images (the image widget has htmx calls to views in the cases app that need to be resolved when rendering the template). Turning those into absolute urls does not work I think, or at least I can't get it to work.

app/grandchallenge/cases/models.py
jmsmkn: Probably fine to log it to Sentry given that we do not expect for it to occur!

app/grandchallenge/evaluation/templates/evaluation/submission_form.html
amickan: That was it! Thanks! I will update the algorithm form as well.

app/grandchallenge/evaluation/models.py
amickan: Good idea. I made it a required input to the function now and added a check in `create_evaluation`.

app/grandchallenge/evaluation/templates/evaluation/submission_form.html
jmsmkn: Wow, I will buy a lottery ticket on the way home.

app/grandchallenge/cases/models.py
amickan: I have updated the error handlers accordingly and also the tests.

app/config/urls/challenge_subdomain.py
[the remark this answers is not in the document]
amickan: These views fail with a CORS error, so this has definitely to do with the subdomain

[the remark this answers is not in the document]
amickan: The specific error message is `Request header field hx-trigger is not allowed by Access-Control-Allow-Headers in preflight response`. Should we allow this?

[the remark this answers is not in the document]
jmsmkn: In the CORS section of settings you should be able to do:

```python
from corsheaders.defaults import default_headers

...

CORS_ALLOW_HEADERS = (
    *default_headers,
    "hx-trigger",
)
```

amickan:
Just fyi: I'm having issues with my dev environment, specifically Docker. Working on fixing this, but it's blocking me from working on this.

app/config/urls/challenge_subdomain.py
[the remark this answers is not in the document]
amickan: I feel like passing the subdomain url configuration when rendering the template (only if the request comes from a subdomain) might fix it, but it seems hacky.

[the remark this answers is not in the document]
jmsmkn: Are you sending the credentials with the request? https://htmx.org/attributes/hx-request/

You can also try stripping out the headers if they are not necessary, then you can avoid adding them to CORS_ALLOW_HEADERS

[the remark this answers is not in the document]
amickan: Including the credentials seems to do the trick, thanks! Stripping out the headers did not work, so leaving those in there.

app/grandchallenge/evaluation/models.py
```diff
@@ -1854,6 +1869,58 @@ def inputs_complete(self):
         else:
             return False
 
+    @property
+    def additional_inputs(self):
+        # additional inputs as currently defined on the phase
+        phase_input_slugs = self.submission.phase.inputs.values_list(
+            "slug", flat=True
+        )
+        return self.inputs.filter(interface__slug__in=phase_input_slugs)
```
amickan, line 1878: I'm using this to display the inputs on the detail page. The way I implemented it now, if half way through the challenge, the inputs are removed from the Phase, they would not show for any of the evaluations, even if those evaluations do have these inputs. 

I could alternatively, collect a list of other input slugs (predictions-json, predictions-csv, algorithm-outputs defined on the phase) and exclude those from the inputs rather than subsetting by inputs currently set on the phase. Then the inputs would still show for old evaluations even if the phase configuration changes.

jmsmkn:
Please could you remove my assignment if you are still working on it?

app/grandchallenge/evaluation/models.py
jmsmkn: I think given the current state of this part of the codebase this is fine. Maybe the field `phase.inputs` should not be used at all and we should drop it and start with a new thing, `phase.evaluation_inputs`, that handles things in the standard way, I don't know.

amickan:
I'm done now

app/grandchallenge/evaluation/models.py
amickan: It is indeed a bit confusing, but I think it will always be because the evaluation inputs will always be a combination of standard handled inputs (validated and added on submission creation from user input) and a set of algorithm outputs, which are added at a later point in time.

amickan: For now, I will leave it at this.

app/config/urls/challenge_subdomain.py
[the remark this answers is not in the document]
jmsmkn: The URLs could get removed from the subdomain then, right? Same for the component urls?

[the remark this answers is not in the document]
amickan: No, they are still necessary to render the widgets (and for the components urls also for the link to the socket options on the interface views).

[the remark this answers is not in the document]
jmsmkn: Why though? Going cross-domain should be fine.

app/grandchallenge/evaluation/models.py
```diff
@@ -1864,14 +1869,20 @@ def inputs_complete(self):
...
+    @property
+    def additional_inputs(self):
+        # additional inputs as currently defined on the phase
         phase_input_slugs = self.submission.phase.inputs.values_list(
             "slug", flat=True
         )
-        return self.inputs.filter(
-            interface__slug__in=phase_input_slugs
-        ).count() == len(phase_input_slugs)
+        return self.inputs.filter(interface__slug__in=phase_input_slugs)
```
jmsmkn, line 1878: Would this work if an additional input is also an algorithm output? Maybe we need to say that the inputs and algorithm outputs must be disjoint. I can't think of a situation where that would be desired but you never know what users will come up with.

amickan: Ah that's a cornercase where this would break indeed. I think it's a very sensible restriction to add. I'll work on adding validation to the admin form.

amickan: I'll make a seperate PR for that restriction. I'm wondering if just checking this in the admin form is enough, or if I also need to add a check to the AlgorithmInterfaceForm (in case someone first configures the inputs and then the algorithm interfaces). Adding a check to the AlgorithmInterfaceForm will be a bit trickier I think.

jmsmkn: I guess it needs to be checked in every place that these two fields could change.

app/config/urls/challenge_subdomain.py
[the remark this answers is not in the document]
jmsmkn: Is `{% load url %}` being used in that template? Might be easier to link to the code on GitHub so I can see exactly what you are referring to.

app/grandchallenge/evaluation/models.py
amickan: I agree. I'll work on that probably tomorrow, and make a seperate PR. The places that will need to have a check are the PhaseAdmin and the AlgorithmInterfaceForm.

app/config/urls/challenge_subdomain.py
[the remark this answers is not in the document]
jmsmkn: So then if you add `{% load url %}` it should work.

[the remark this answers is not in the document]
amickan: Huh, now I also excluded the components urls, and now the image widget loads fine. WTH?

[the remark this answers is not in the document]
amickan: Ok, removing the cases and components urls from the subdomain urlconfig and then adding `{% load url %}` to all templates that require urls from those apps works.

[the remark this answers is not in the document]
jmsmkn: Nice!

app/grandchallenge/evaluation/admin.py
amickan: Does my comment address you question / concern?

jmsmkn (approved):
LGTM!

app/config/urls/challenge_subdomain.py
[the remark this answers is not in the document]
amickan: Just started up my dev environment again, and for some reason the credentials issue popped up again when loading the widgets. I have tried adding `hx-request='{"credentials":"include"}'` to both the flexible_image_widget and flexible_file_widget templates, but to no avail. I don't understand why this was working yesterday at the end, and is not working now. I will park this issue for now and continue with the more crucial parts of the pitch, but still needs to be revisited.

PR #3872, opened 2025-03-10
Add phase interface configuration links to challenge admin

amickan:
Accessing the pages to set up the algorithm interfaces for phases is difficult for non-superusers since those are hidden within the challenge admin panel. This adds the urls to the relevant pages to the challenge admin, just underneath the configure-algorithm-phase link.


jmsmkn (commented):
Shouldn't these links be on the Phase model and not the Challenge model?

amickan:
Since the other link to configure algorithm phases (which is unique per challenge, not phase), is on the Challenge model, I thought it would be easiest to put them there - one place for all configuration links.

PR #3879, opened 2025-03-10
Ground Truth csv form seperation

chrisvanrun:
Part of the pitch:

This PR further pull apart the ground truth main view and the csv upload form.

This is to create a nice niche where the answer copy form can be placed in both UI and workflow.

It also:
- Fixes having non ground-truth applicable questions in the example csv (e.g. annotations)
- Adds a warning for 'incomplete' ground truth that should give some warnings to reader study admins when they added a new question or have forgotten to set the ground truth when it is set to educational.
- Add a confirmation modal when deleting GT

## Showcase

https://github.com/user-attachments/assets/8ff85dc5-e6f6-4d02-9d7b-1c099394afbc


app/grandchallenge/reader_studies/urls.py
```diff
@@ -76,14 +76,14 @@
...
-        name="add-ground-truth",
+        "<slug:slug>/ground-truth/add/csv/",
+        AddGroundTruthViaCSVToReaderStudy.as_view(),
+        name="add-ground-truth-csv",
     ),
     path(
-        "<slug:slug>/ground-truth/example/",
-        ReaderStudyExampleGroundTruth.as_view(),
-        name="example-ground-truth",
+        "<slug:slug>/ground-truth/add/csv/example/",
```
jmsmkn, line 84: add/csv/example is misleading. Maybe these should be `<slug:slug>/ground-truth/csv/create/` and `<slug:slug>/ground-truth/csv/example/`?

app/grandchallenge/reader_studies/views.py
```diff
@@ -486,18 +475,34 @@ def get_context_data(self, **kwargs):
...
 
-class AddGroundTruthToReaderStudy(BaseAddObjectToReaderStudyMixin, FormView):
-    form_class = GroundTruthForm
-    template_name = "reader_studies/readerstudy_add_object.html"
-    type_to_add = "ground truth"
+class AddGroundTruthViaCSVToReaderStudy(
+    BaseAddObjectToReaderStudyMixin, FormView
+):
+    form_class = GroundTruthCSVForm
+    template_name = "reader_studies/ground_truth_via_csv.html"
```
jmsmkn, line 482: This template name is not constant with the rest of the application. Maybe something like `ground_truth_csv_form.html`?

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_ground_truth.html
```diff
@@ -16,49 +16,85 @@
...
 
 {% block content %}
-    <h2>Ground Truth</h2>
+    <h2>Ground Truth of {{ object }}</h2>
 
     <p>
-        If you wish to assess the performance of the readers you
-        can upload the CSV file containing the ground truth for
-        this reader study here.
+        A ground truth consists of the correct answers to the questions for each case, it serves as a
```
jmsmkn, line 22: Ground Truth would not be singular.

jmsmkn (commented):
Looks good, small things for consistency

app/grandchallenge/reader_studies/views.py
chrisvanrun: I agree, that makes it better suite the rest of the naming scheme.

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_ground_truth.html
chrisvanrun: T.b.h. I am a bit confused on the correct usage of singular and plural forms with respect to 'ground truth' or akin collections. Likely too subtle for a brute-force Dutchie!

When referencing it as a singular collection I'd be inclined to use 'ground truth'. I am not sure when or where to use the plural 'ground truths'. Or should perhaps the problem lies with the 'A ...' and it should be 'The ground truth...'?

jmsmkn: I would say The Ground Truth, A Ground Truth sounds wrong to me but unfortunately I cannot explain why - the style of English education at the time was to do away with rules.

chrisvanrun: Roger! Let us then fully lean on your English roots! Thank you for noticing.

On the uptick, I did find a misspelling of 'wish' as 'widh' that was part of the CSV paragraphs for the longest time!

PR #3881, opened 2025-03-11
Add border to user profile image

### LINKED ISSUE #3863: Add a small border or shadow to user profile images
A user uploaded their profile photo as the same background as the site, it now looks broken:

Image

ammar257ammar:
closes #3863 

Adding the border helps in cases where the user profile image is as the same background of the site which makes it looks like broken.

Not it looks like this:

Screenshot 2025-03-11 at 10 08 35

### DIFF
--- a/app/grandchallenge/profiles/templatetags/profiles.py
+++ b/app/grandchallenge/profiles/templatetags/profiles.py
@@ -22,7 +22,7 @@ def user_profile_link(user: AbstractUser | None) -> str:
         )
         mugshot = format_html(
             (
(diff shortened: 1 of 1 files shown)


chrisvanrun:
What does it look like with `shadow-sm`?

ammar257ammar:
It looks like this
Screenshot 2025-03-12 at 09 16 47

Which one do you prefer?
To me, the border looks better but I don't mind either.

jmsmkn (approved):

PR #3882, opened 2025-03-11
Add `ALGORITHMS_MAX_ACTIVE_JOBS_PER_ALGORITHM`

jmsmkn:
Makes this separately adjustable and default to 1 so that the queue is cleared for the particular algorithm image before adding more jobs.

See https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/470


amickan:
So the default of 1 means that there can only be 1 active job per algorithm. Aren't we losing the parallelization benefit with that? Or are you planning on changing that setting in production to a higher number?

amickan:
Oh right, I understand it now.

amickan (approved):

Issue #64, opened 2012-09-18
Function to copy an entire site


sjoerdk:
This is what many want to do when they create a project: Take that one, which I like, and alter it to suit my needs.  It should be easy to start a new project based on an existing project.

Pages need to be easily copyable as well. Either by show/copy source button or via an interface when creating a new page (e.g. some meny copy existing > project > page )

PR #3883, opened 2025-03-11
Fix numberical slugs breaking collapsables

chrisvanrun:
Seems like bootstrap does not like HTML element IDs that start with digits (e.g. `#64-collapse`). Simple fix.


amickan (approved):
Thanks!!

PR #3884, opened 2025-03-11
Ground truth copy from answers

chrisvanrun:
Part of the pitch:

This PR adds a form that, when submitted, copies the answers of a reader/editor as ground truth. 

Only users that have provided answers are selectable. However, a 100% completion of the reader study is required to consider the select user valid.

Since readers are limited in number I've explicitly opted to not use an autocomplete view + widget and have instead used a simple Select.

# Showcase

https://github.com/user-attachments/assets/2c1d21d0-8177-4aee-9ffe-308938fcf0ac


app/grandchallenge/reader_studies/urls.py
```diff
@@ -85,6 +86,11 @@
         ReaderStudyExampleGroundTruthCSV.as_view(),
         name="example-ground-truth-csv",
     ),
+    path(
+        "<slug:slug>/ground-truth/copy-answers/",
```
jmsmkn, line 90: Maybe?

```suggestion
        "<slug:slug>/ground-truth/answers/create/",
```

app/grandchallenge/reader_studies/views.py
```diff
@@ -511,6 +512,28 @@ def get_success_url(self):
...
+    def get_form_kwargs(self):
+        kwargs = super().get_form_kwargs()
+        kwargs.update({"reader_study": self.reader_study})
+        return kwargs
+
+    def form_valid(self, form):
+        is_valid = super().form_valid(form)
+        if is_valid:
+            form.create_ground_truth()
+        return is_valid
```
jmsmkn, line 531: `form_valid` is only called when the form is valid. `super().form_valid(form)` does not tell you if the form is valid or not, it returns a redirect response. There is no need to check it as it is always a non-empty string, so this can be simplified:

```suggestion
        response = super().form_valid(form)
        
        form.create_ground_truth()
        
        return is_valid
```

app/grandchallenge/reader_studies/forms.py
```diff
@@ -688,6 +689,54 @@ class Meta(PermissionRequestUpdateForm.Meta):
...
+
+    def create_ground_truth(self):
+        answers = Answer.objects.filter(
+            question__reader_study=self._reader_study,
+            creator=self.cleaned_data["user"],
+        )
+        for answer in answers:
+            answer.pk = None
+            answer.is_ground_truth = True
+            answer.save()
```
jmsmkn, line 737: This will take too long for a request-response cycle and would need to be done in an asynchronous task. Alternatively, it could be done with `answers.update(is_ground_truth=True)` if we are sure there are no side effects or dependencies on the models save method. Maybe there are for image permissions for linked images (masks).

app/grandchallenge/reader_studies/forms.py
```diff
@@ -688,6 +689,54 @@ class Meta(PermissionRequestUpdateForm.Meta):
         model = ReaderStudyPermissionRequest
 
 
+class GroundTruthCopyAnswersForm(SaveFormInitMixin, Form):
+    user = ModelChoiceField(
+        queryset=get_user_model().objects.none(),
+        required=True,
+        help_text="Select a user who's answers will be copied.",
```
jmsmkn, line 696: who's is a contraction of who is

```suggestion
        help_text="Select a user whose answers will be copied.",
```

jmsmkn (changes requested):
Looks good but it currently won't fit into a request-response cycle for realistic reader studies.

app/grandchallenge/reader_studies/urls.py
chrisvanrun: Also doubting. It's not really creating answers... or maybe it is but based off other answers? Creating answers is something that the CSV route also does.

Having had a think about this I think your suggestion is best. The /csv/ or /answers/ sort points to the source. answers/copy/ might be applicable but doesn't quite fit the rest of the URL routes in the app.

app/grandchallenge/reader_studies/forms.py
chrisvanrun: The update route is hazardous.  Even if it might fit now: it might unintenionally break in the future. IIRC the permissions are set in the save method so it's not really an option.

I think a simple push to a single async task is enough for particial use, no need to batch it.

app/grandchallenge/reader_studies/views.py
chrisvanrun: Ah, I had missinterpreted its relation to the `is_valid`. If the form_valid execution is already dependent on the form being valid this is fine.

app/grandchallenge/reader_studies/forms.py
chrisvanrun: Also, if this is the case, should the CSV extraction then not also be pushed async?

jmsmkn: > it might unintenionally break in the future.

That is what tests are for.

If the CSV answer processing is in the request response cycle then maybe people don't use this for many cases/questions and inline is fine.

If so, you will still need to update how the questions are created. Note the section on inheritance: https://docs.djangoproject.com/en/5.2/topics/db/queries/#copying-model-instances

A quicker way would be not to create new answers but update the state on existing ones, either with update or saving each one.

If I were you I would use `queryset.update`, write a test, and call it a day.

chrisvanrun: Having looked at how ground_truth are actually being serialized I just realized they are a free-for-all: 0 permission checks. As such, I think we can get away with skipping save() all together.

Part of the requirement is leaving the old ones in place, but I hadn't realized we needed to update `_state`. I think we can use `Answer.objects.create_batch` as alternative.

Deleting the ground-truth currently leans on generic `Answer` deletion and hence checks `"reader_studies.delete_answer"`. I think it acceptable, given the creation of the ground truth is checked against `"reader_studies.change_readerstudy"` to rewrite the deletion to do similar.

This does mean we dip our toes into the 'get_permission_object` smell (https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/335) but ground-truth is a bit of an odd duck anyway given they are a special class of answers.

I'll make the changes and ping for a re-review once done.

jmsmkn: I do not see a need to leave the old ones in place.

> I just realized they are a free-for-all: 0 permission checks. 

I do not see what you mean here. The object based permissions are in place and are used. The reader cannot get the ground truth answers unless they created them.

jmsmkn: That uses object permissions, set as an attribute on the view set, so will require the view reader study permission.

chrisvanrun: Yes, the reader study object. But not `Answer`, so that should work. I expected some permissions to be involved there for answers.

jmsmkn: But that's the point, the reader does not have permission to view the answer directly, but needs the ground truth answers are they are scored client side. This provides that. Whether that is the correct setup now is a question for another pitch (https://github.com/DIAGNijmegen/rse-roadmap/issues/395)

chrisvanrun: Indeed. As part of checking if we can get away with a `bulk_create` / `build_update` I am quickly checking the different workflows wrt ground truth:

**Creating** ground truth is bound by the `"change_readerstudy"` on the reader study itself.

**Viewing** the ground truth, only the `"view_readerstudy"` is used from the reader study itself, ignoring permissions on the underlying `Answer`.

**Updating** the ground truth is explicitly made impossible. (Quick double check on the api route. Indeed: the `AnswerSerializer` has the `is_ground_truth` set to read only).

**Deleting** the ground truth is dependent on the `"delete_answer"` permission on the `Answer` object.

So I am currently refactoring the deleting atm to rely on the `"change_readerstudy"` instead.

jmsmkn: > Deleting is the problem in this case so I am currently refactoring the deleting atm to rely on the "change_readerstudy" instead.

Please do not do that! Using the delete permissions to delete things is correct. Please discuss things before you make changes like this.

app/grandchallenge/reader_studies/tasks.py
```diff
@@ -128,3 +128,20 @@ def copy_reader_study_display_sets(*, orig_pk, new_pk):
...
+    reader_study_pk,
+    user_pk,
+):
+    answers = Answer.objects.filter(
+        question__reader_study=reader_study_pk,
+        creator=user_pk,
+    )
+    for answer in answers:
+        answer.pk = None
+        answer.is_ground_truth = True
```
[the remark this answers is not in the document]
chrisvanrun: Ah! I missed the `id` for inheritence. What an odd requirement. Mea cupla. I dismissed the adding state since it's a grey area when bulk_creating and I don't think it's needed in this case? I'll add both since we'll be switching back to deletion.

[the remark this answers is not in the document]
jmsmkn: When doing bulk create you are anyway skipping save, which was your argument for not doing this with update. I would really prefer that this was kept simple rather than adding all of these complications.

chrisvanrun:
# Updated showcase:


https://github.com/user-attachments/assets/2503c0d7-c66f-448b-b491-3bf298872d76

app/grandchallenge/reader_studies/tasks.py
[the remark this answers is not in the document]
chrisvanrun: I've reverted back to the `save()`. Forgoing any optimalizations because of permissions considerations.

app/grandchallenge/reader_studies/forms.py
chrisvanrun: I see why you'd only expect a single form. 

However, the first objection in the review was that doing this in a response-request was too slow. Implementing that gave some trouble in considering subsequent request. Then the response-request being slow  was suddenly ok, with a question if we could use a bulk operation of sorts. The later was maybe OK if we'd push the split permission-checking scheme one way? Which involved tweaking the deletion of ground truth which, I admit, wasn't thought out very well on my side.

So we've went around and are now back at the somewhat starting point, with some tweaks and discoveries along the way.

jmsmkn: The request-response cycle isn't okay, this will break when you get a serious reader study. That will then be on you to fix, if you want that, then fine. Given my experience of running this platform for 10 years I have repeatedly advised you, both [a week ago](https://github.com/comic/grand-challenge.org/pull/3884#discussion_r1991582393) and [again yesterday](https://github.com/comic/grand-challenge.org/pull/3884#discussion_r1998050953) to pick for an easy life as the time that this feature will break will not come at your choosing and to use `update`.

app/grandchallenge/reader_studies/forms.py
```diff
@@ -688,6 +689,67 @@ class Meta(PermissionRequestUpdateForm.Meta):
...
+        return user
+
+    def create_ground_truth(self):
+        answers = self.answers
+
+        for answer in answers:
+            answer.is_ground_truth = True
+
+            answer.pk = None  # Ensures new ones are created
+            answers.id = None  # Required to ensure inheritence works correctly
```
jmsmkn, line 747: `id` is being set on `answers` here and not `answer`. This sets the Python identifier and not the Django `id`.

jmsmkn (changes requested):

app/grandchallenge/reader_studies/forms.py
chrisvanrun: I feel like a dufus for misstyping this. Sorry!

app/grandchallenge/reader_studies/forms.py
chrisvanrun: In that case I missread your reply when I mentioned this is already the case for the CSV-source route.

I'll double check where the initial requirement of copying came from in the pitch.

chrisvanrun:
Answers are now consumed instead of copied. The form and views have been renamed to pre-pare for the reverse action in a follow-up PR.

chrisvanrun:
The early flaky test was from a permission check that was a single set in a parameterized test. I've left it for now.

app/grandchallenge/reader_studies/forms.py
```diff
@@ -688,6 +689,61 @@ class Meta(PermissionRequestUpdateForm.Meta):
...
+        if (
+            self._reader_study.answerable_questions.exists()
+            and progress["diff"] != 0
+        ):
+            raise ValidationError("User has not completed the reader study!")
+
+        return user
+
+    def create_ground_truth(self):
+        self.answers.update(is_ground_truth=True)
```
koopmant, line 744: Not sure if it is within scope, but could they also be copied instead? If the user wants to update the ground truth they have to delete it, and then all the answers are gone...

app/grandchallenge/reader_studies/templates/reader_studies/ground_truth_from_answers_form.html
```diff
@@ -0,0 +1,14 @@
+{% extends "reader_studies/ground_truth_csv_form.html" %}
+
+{% block form_heading %}
+    <h2>Add {{ type_to_add }} to {{ object }}</h2>
+
+    <p>
+        All the answers of the selected user will be consumed and remade into ground truth.
```
koopmant, line 7: I'd make this perhaps a bit less technical and (even) more explicit. Perhaps 'deleted' instead of 'consumed'.

app/grandchallenge/reader_studies/forms.py
```diff
@@ -688,6 +689,61 @@ class Meta(PermissionRequestUpdateForm.Meta):
...
+        )
+
+    @cached_property
+    def answers(self):
+        return Answer.objects.filter(
+            question__in=self._reader_study.ground_truth_applicable_questions,
+            creator=self.cleaned_data["user"],
+        )
+
+    def clean(self):
```
koopmant, line 721: I think this should return cleaned_data

app/grandchallenge/reader_studies/forms.py
```diff
@@ -688,6 +689,61 @@ class Meta(PermissionRequestUpdateForm.Meta):
...
+    def clean_user(self):
+        user = self.cleaned_data["user"]
+
+        progress = self._reader_study.get_progress_for_user(user)
+
+        if (
+            self._reader_study.answerable_questions.exists()
+            and progress["diff"] != 0
+        ):
+            raise ValidationError("User has not completed the reader study!")
```
koopmant, line 739: This is somehow not raised when I test it manually... Not sure why though

koopmant (commented):
Code looks good to me, but I found an issue in testing it.

app/grandchallenge/reader_studies/forms.py
koopmant: I tested it with one question, two cases. Only one of the cases was answered.

app/grandchallenge/reader_studies/templates/reader_studies/ground_truth_from_answers_form.html
chrisvanrun: I doubt. The problem with less technical is that things become inaccurate quite quickly. Then again, consumed doesn't quite have the ring like 'deleted'.

What about:

"The selected user's answers will be used to generate ground truth, meaning their individual answers will be removed."

app/grandchallenge/reader_studies/forms.py
chrisvanrun: Good catch. This was a copy and paste of a bad code example in this module. I've also fixed the bad example.

app/grandchallenge/reader_studies/forms.py
chrisvanrun: Interesting! There might be a bug with the progress calculation. Investigating!

chrisvanrun: I had incorrectly assumed the `"diff"` was used as an inverse progression. Fixed!

chrisvanrun:
I've also pushed changed that will update the score when deleting and creating GT from answers. I thought it wise to add it here as to not have a 'broken' main. This already works for the CSV route via signals.

I'll admit that the CSV route is very inefficient and the use of a signal is not ideal. For instance it spawns a task for **each answer** when saving a G). I started refactoring it a bit but quickly decided that it is out of the scope of this PR / pitch and dropped it.

app/grandchallenge/reader_studies/forms.py
```diff
@@ -153,11 +157,11 @@ class Meta:
         }
 
     def clean(self):
-        super().clean()
+        cleaned_data = super().clean()
```
chrisvanrun, line 160: Sanitation duty

app/grandchallenge/reader_studies/views.py
```diff
@@ -772,15 +795,33 @@ def get_success_url(self):
...
+        return context
 
     def get_queryset(self):
         return Answer.objects.filter(
             question__reader_study=self.reader_study,
             is_ground_truth=True,
         )
 
+    def form_valid(self, *args, **kwargs):
+        Answer.objects.filter(question__reader_study=self.reader_study).update(
```
chrisvanrun, line 820: No need to re-calculate scores here: this view will explicilty remove all ground truth so all scores are null.

app/tests/reader_studies_tests/test_views.py
```diff
@@ -958,3 +957,105 @@ def test_ground_truth_view(client):
     )
 
     assert response.status_code == 200
+
+
+@pytest.mark.django_db
+@override_settings(task_eager_propagates=True, task_always_eager=True)
+def test_ground_truth_from_answers_workflow(client):
```
chrisvanrun, line 964: Sort of an integration test. If for instance, one form doesn't set the permissions correctly another form would fail: this checks if they integrate correctly.

chrisvanrun:
Ready for re-review.

koopmant:
I'll take. But James will need to look at it also.

app/grandchallenge/reader_studies/tasks.py
```diff
@@ -54,6 +54,35 @@ def add_scores_for_display_set(*, instance_pk, ds_pk):
...
+    def key(answer):
+        return (str(answer.question_id), str(answer.display_set_id))
+
+    ground_truth_lookup = {}
+    for answer in ground_truth:
+        ground_truth_lookup[key(answer)] = answer
+
+    answers = Answer.objects.filter(
+        question__reader_study__pk=reader_study_pk
+    ).prefetch_related("question")
```
jmsmkn, line 73: I think only non-ground-truth answers should be set:

```suggestion
    answers = Answer.objects.filter(
        question__reader_study__pk=reader_study_pk,
        is_ground_truth=False,
    ).prefetch_related("question")
```

app/grandchallenge/reader_studies/forms.py
```diff
@@ -688,6 +691,56 @@ class Meta(PermissionRequestUpdateForm.Meta):
...
+            raise ValidationError("User has not completed the reader study!")
+
+        return user
+
+    def create_ground_truth(self):
+        answers = Answer.objects.filter(
+            question__in=self._reader_study.ground_truth_applicable_questions,
+            creator=self.cleaned_data["user"],
+        )
+        answers.update(is_ground_truth=True, score=None)
```
jmsmkn, line 738: Should all answers for the reader study have their scores reset?

jmsmkn (commented):
Looks good, couple of questions.

app/grandchallenge/reader_studies/forms.py
chrisvanrun: It is intentional: it is a clear signal that all scores are invalid until we can get to re-calculating them.

chrisvanrun: In practice that is the case: when deleting the ground-truth all scores are reset and a user (should) not get here without deleting the ground-truth.

I could play it super safe and have it reset all scores here? It is only one additional query to the DB after all?

app/grandchallenge/reader_studies/tasks.py
chrisvanrun: Agreed! Albeit there are filters that make this redundant I still think we should have a clear split here.

chrisvanrun:
Ugh. precommit CI is failing on building. Not sure why, I'll check it out tomorrow.

jmsmkn (approved):
LGTM!

PR #3885, opened 2025-03-12
Give challenge admins view invoice permission

koopmant:
This allows us to create views for invoices that challenge admins have access to.


jmsmkn (commented):
Looks good but we need a data migration and the test could be simplified with https://django-guardian.readthedocs.io/en/stable/api/guardian.shortcuts.html#get-users-with-perms and https://django-guardian.readthedocs.io/en/stable/api/guardian.shortcuts.html#get-groups-with-perms

app/grandchallenge/invoices/migrations/0005_assign_perms_challenge_admins.py
```diff
@@ -0,0 +1,18 @@
+from django.db import migrations
+from guardian.shortcuts import assign_perm
+
+
+def assign_perms_challenge_admins(apps, schema_editor):
+    Invoice = apps.get_model("invoices", "Invoice")  # noqa: N806
+    for invoice in Invoice.objects.all():
+        assign_perm("view_invoice", invoice.challenge.admins_group, invoice)
```
koopmant, line 8: @jmsmkn It's probably not this simple, is it?

koopmant: I've updated the code and I think it's gone in the right direction. But I have to admit I don't fully understand how it works...

app/grandchallenge/invoices/models.py
```diff
@@ -100,3 +102,29 @@ class Meta:
                 name="payment_status_in_choices",
             ),
         ]
+
+    def save(self, *args, **kwargs):
+        adding = self._state.adding
+        super().save(*args, **kwargs)
+        if adding:
+            self.assign_permissions()
```
koopmant, line 110: In other models I see we assign permissions on every save. Is there a reason we are not just doing when `adding`?

app/grandchallenge/invoices/migrations/0005_assign_perms_challenge_admins.py
jmsmkn: Indeed, it can be simplified. Why did you switch to iterating over challenges? It may be more straightforward to go back to iterating over invoices:

```python
Invoice = apps.get_model("invoices", "Invoice")  # noqa: N806
InvoiceGroupObjectPermission = apps.get_model("invoices", "InvoiceGroupObjectPermission")  # noqa: N806
Permission = apps.get_model("auth", "Permission") # noqa: N806

view_permission = Permission.objects.get(
    codename=f"view_invoice",
    content_type__app_label="invoices",
)

for invoice in Invoice.objects.all():
  InvoiceGroupObjectPermission.objects.create(
      content_object=invoice,
      group=invoice.challenge.admins_group,
      permission=view_permission,
  )
```

app/grandchallenge/invoices/models.py
koopmant: okay, then I think it is fine like this.

app/grandchallenge/invoices/migrations/0005_assign_perms_challenge_admins.py
koopmant: I copied what we did in `grandchallenge/archives/migrations/0017_archive_item_delete_permission_granting.py`. I presumed using `invoice.challenge.admins_group` might not work, and I thought that was why we are using `prefetch_related("editors_group")` for the archive items.

jmsmkn: It should work, prefetch_related is an optimisation technique. It is important to know about that, N+1 problems and select_related, and where to use each one. Maybe this is a good post on the topic? https://johnnymetz.com/posts/combine-select-related-prefetch-related/

The tests fail as there is something wrong with this query (Permission matching query does not exist.):

```
view_permission = Permission.objects.get(
    codename=f"view_invoice",
    content_type__app_label="invoices",
)
```

I can't work it out. Does that work in the shell?

koopmant: Yes, that works fine in the shell

koopmant: > It is important to know about that, N+1 problems and select_related, and where to use each one. Maybe this is a good post on the topic? https://johnnymetz.com/posts/combine-select-related-prefetch-related/

Yes, that one is good. I found this one helpful as well:
https://hakibenita.com/all-you-need-to-know-about-prefetching-in-django

jmsmkn: Okay, maybe move it into the for loop as it must not be created until after the first migration. Then please test this by checking out main, creating some invoices for different challenges, then checking out the branch and applying the migration. Then check the correct permissions are applied (e.g. the correct code and group are applied). It might be worth switching the Invoice admin model to GuardedModelAdmin (?)

koopmant: If adjusted the migration to check if there are invoices first. That fixes the tests issue, but mainly because this part is no longer reached now:
```python
view_permission = Permission.objects.get(
    codename="view_invoice",
    content_type__app_label="invoices",
)
```

I tried to write a test for it, but kept running into issues. I tested it manually by going back to main, running migrations, creating some invoices, testing there are no groups with permissions on the invoices (pass), going to this branch, running last migrations, testing if there are groups with permissions on the invoices (pass).

koopmant: Ah I see I missed your comment. At least we were on the same track. 
I'll look into the `GuardedModelAdmin`

koopmant: I understand what you mean now. Yes the GuardedModelAdmin is helpful!

jmsmkn (approved):
LGTM!

PR #3889, opened 2025-03-14
Reevaluate submission with extra inputs

amickan:
This enables reevaluating a submission with different inputs. 


app/grandchallenge/evaluation/models.py
```diff
@@ -1710,6 +1720,74 @@ class EvaluationGroundTruthGroupObjectPermission(GroupObjectPermissionBase):
...
+                        "inputs",
+                        filter=(
+                            Q(
+                                inputs__interface__slug__in=configured_input_interface_slugs
+                            )
+                        ),
+                        distinct=True,
+                    ),
+                    0,
+                ),
```
amickan, line 1772: I can't say that I fully understand this, but it appears that annotations with zero counts need special treatment for the filter below to work properly. I tried various things, but the `Coalesce` was ultimately the cleanest solution I could find. 

This also needs to be fixed in `annotate_input_output_counts` which we use for Jobs. It's not been an issue since we currently never pass empty inputs to this function and when we pass empty outputs, we don't filter on those subsequenty. But it should of course be fixed nonetheless.

jmsmkn: I think we've seen this before as empty matches will return `None`. Can't you use `Count(..., default=0)`? https://docs.djangoproject.com/en/5.1/ref/models/querysets/#default

amickan: For count the `default` argument is not supported, see the note on the page you linked.

jmsmkn: Ah okay. Then the behaviour is strange, as they set `empty_result_set_value = 0`.

amickan: Indeed, I don't quite understand what's going on either. I can see that the count annotation is 0, but then when I filter 0 != 0 for some reason. I thought it was a type difference, but the Count outputs an Integer, so that can't be it.

jmsmkn: Weird, surprised that we've not hit this before as we've used this pattern before, but maybe it's only been when the filter count is non-zero?

amickan: Indeed, I think in all other places we either don't end up with zero-counts, or if we do (it can happen for `annotate_input_output_counts` when we don't pass in outputs), we at least don't filter on those.

jmsmkn: Good to know! Please add this to our knowledge base somewhere as this is a strange gotcha.

jmsmkn (commented):
Looks good, only concerns are the change in the unique together as discussed, but if it's covered by tests then hopefully it's fine - after all, I guess the same problem exists for algorithm jobs and those are okay? Small question about the tests.

amickan:
> only concerns are the change in the unique together as discussed, but if it's covered by tests then hopefully it's fine

Yes, I will double check again that there are tests in place for all entry-points.

> I guess the same problem exists for algorithm jobs and those are okay?

Correct. Same principle that we check for uniqueness in the form and the API views and block creation there.

app/tests/evaluation_tests/test_views.py
```diff
@@ -2607,3 +2607,113 @@ def test_parent_phase_algorithm_interfaces_locked(client):
...
+    )
+    eval1 = EvaluationFactory(
+        submission=submission,
+        method=method,
+        status=Evaluation.SUCCESS,
+        time_limit=10,
+    )
+    eval1.inputs.set([civ_str, civ_bool])
+
+    with django_capture_on_commit_callbacks(execute=True):
```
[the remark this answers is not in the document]
amickan: Hm I thought it didn't since I'm just creating a dummy method, but I'll double check. It is indeed not necessary for the test. I think I could just set `execute=False` actually. I'll try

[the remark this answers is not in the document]
amickan: `execute=False` works in this case since the additional inputs are of json type stored in DB and hence are validated synchronously. We have a more comprehensive test that covers all possible input types elsewhere and there I mock the call to execution. In any case, fine to not execute the callbacks.

amickan:
There are now the following tests for uniqueness:
- `test_create_evaluation_is_idempotent`: checks that the `create_evaluation` method does the uniqueness check
- `test_reevaluate_submission_idempotent`: just added a separate check for the admin action, in case we ever change the implementation and move away from using `create_evaluation` under the hood for this
- `test_reschedule_evaluation_with_additional_inputs` checks idempotency for the re-evaluate view

jmsmkn (approved):
LGTM, thanks!

PR #3890, opened 2025-03-14
Invoice validation

koopmant:
The following constraints were added:
- Complimentary invoices require internal comments
- Non-complimentary invoices require the following details:
  - contact name
  - contact email
  - billing address
  - vat number
- Paid non-complimentary invoices must have a paid_on date
- Issued non-complimentary invoices must have:
  - issued_on date
  - internal invoice number
  - internal client number
- The total amount cannot change after the invoice is created (but costs can be redistributed between compute, support and storage)


jmsmkn:
Please could you fix the tests?

koopmant:
> Please could you fix the tests?

Sorry about that. They are fixed now.

app/grandchallenge/invoices/models.py
```diff
@@ -94,15 +98,90 @@ class Invoice(models.Model):
...
             models.CheckConstraint(
-                check=models.Q(payment_status__in=PaymentStatusChoices.values),
+                check=Q(payment_status__in=PaymentStatusChoices.values),
                 name="payment_status_in_choices",
             ),
+            models.CheckConstraint(
+                name="issued_on_date_required_for_issued_payment_status",
+                check=Q(issued_on__isnull=False)
+                | ~Q(payment_status=PaymentStatusChoices.ISSUED)
+                | Q(payment_type=PaymentTypeChoices.COMPLIMENTARY),
```
jmsmkn, line 112: Maybe this is more straightforward: `check=~Q(payment_status=PaymentStatusChoices.ISSUED) | Q(issued_on__isnull=False)`? Similar for the paid constraint below.

app/grandchallenge/invoices/models.py
```diff
@@ -94,15 +98,90 @@ class Invoice(models.Model):
...
+                check=Q(paid_on__isnull=False)
+                | ~Q(payment_status=PaymentStatusChoices.PAID)
+                | Q(payment_type=PaymentTypeChoices.COMPLIMENTARY),
+                violation_error_message="When setting the payment status to 'Paid',"
+                " you must set the 'Paid on' date.",
+            ),
+            models.CheckConstraint(
+                name="comments_required_for_complimentary_payment_type",
+                check=Q(internal_comments__gt="")
+                | ~Q(payment_type=PaymentTypeChoices.COMPLIMENTARY),
```
jmsmkn, line 127: Maybe `check=~(Q(payment_type=PaymentTypeChoices.COMPLIMENTARY) & Q(internal_comments=""))` is more readable?

app/tests/invoices_tests/test_models.py
```diff
@@ -254,3 +268,100 @@ def test_most_recent_submission_datetime_multiple_submissions():
...
     )
     assert challenge.most_recent_submission_datetime == last_submission.created
+
+
+@pytest.mark.django_db
+def test_payment_status_issued_requires_issued_on():
+    invoice = InvoiceFactory()
+
+    invoice.payment_status = invoice.PaymentStatusChoices.ISSUED
+    with pytest.raises(IntegrityError) as e, transaction.atomic():
```
jmsmkn, line 278: Why are atomic blocks required here and not in the 2nd call?

jmsmkn (commented):
Looks good! Couple of suggestions to make the constraints easier to follow.

app/tests/invoices_tests/test_models.py
koopmant: The atomic block makes the exception catching work. Without it, the exception is somehow not caught by `pytest.raises` and the test fails because of it. In the later calls, there is no exception to be caught. 
(I have tested that the exception still occurs and causes the test to fail if the constraint is violated in the second call.)

app/grandchallenge/invoices/models.py
koopmant: Yes, that is a bit easier to read indeed.

app/tests/invoices_tests/test_models.py
jmsmkn: Hmm, the construction is strange. What about doing `full_clean` in both places rather than `save`?

jmsmkn (approved):
LGTM! I will hold off on merging this as the data will need to be manually migrated first so that the constraints are satisfied.

jmsmkn (changes requested):
The check on the fields contact name etc use `__gt`, I don't think that is correct?

jmsmkn:
We should also check that if an invoice is issue then the internal invoice number and internal client number are set.

koopmant:
> The check on the fields contact name etc use `__gt`, I don't think that is correct?

Thanks for checking! You're right, it's not. It was bugging me when I wrote it, but it did not seem to work with the inverse (`~Q(field_name="")`). I must have made a mistake when I was testing it back then, because it works fine now.

koopmant:
> We should also check that if an invoice is issue then the internal invoice number and internal client number are set.

I've set it like that now.

koopmant:
I've added an additional constraint to set the payment status to 'Paid' when the paid_on date is provided.

koopmant:
> I've added an additional constraint to set the payment status to 'Paid' when the paid_on date is provided.

Reverted this again, because it makes working with the InvoiceFactory too complicated. (I'd either have to set the default status to 'Paid' because there is a paid_on date generated by the factory, or I'd have to not set that date in the factory, but then it would need to be provided manually in tests when a paid invoice is required.)

jmsmkn:
The tests should not dictate the functionality we need, please find a way to implement it.

jmsmkn:
Actually, I see what you mean. This is probably fine but will need to wait until next week to deploy now as the existing data needs a lot of cleaning, and I have higher priority work to do to finish the cycle.

jmsmkn:
Could you merge main and resolve the conflicts? Still need to keep this up to date but still not found time to do a data migration.

It honestly might be better to implement this as part of a clean method and not using database constraints and accept that the old data does not fit. Right now it means looking up contact names and emails from invoices from 3 years ago.

koopmant:
Updated and conflicts solved.

Using clean might be more pragmatic indeed... Although, if we're doing that and thus not really fix old invoices, why not put placeholders in the old ones?

amickan:
I can help with updating the invoices.

amickan:
Questions:
- Non-EU invoices don't require a VAT number, so unless you want to add a very fine grained check, I would drop that requirement for non-complimentary invoices
- For internal charges, Anne-Marie does not list a client-nr, probably because there is no client number for internal customers. So I think that requirements needs to be dropped as well, unless we want to fill that with a placeholder, like "internal"

amickan:
RE VAT: ok, but then it is still the case that all previous invoices from outside-EU customers don't have a VAT. But I think James anyway suggested moving this to the clean method on the form, so then it would only apply to future invoices (or old ones we update). 

RE client-nr: a placeholder would be fine with me as well. Easy enough to replace.

amickan:
Ok, I finished updating all invoices.

VAT is still empty for the vast majority of invoices.
Client-nr for internal invoices is also still empty.

koopmant:
@jmsmkn @amickan Sorry, is this one waiting for action from my side? 
Do we go with placeholders for all the old ones, or should I remove the constraints? (If so, should I remove them for all: contact name, email, vat and address?)

amickan:
I think the checks need to be moved to the clean method still, right? At least the check on the VAT, because the old invoices will not fullfill this requirement.

koopmant:
Sure, but a placeholder like `"-"` does, that's why I'm asking.

amickan:
> Sure, but a placeholder like `"-"` does, that's why I'm asking.

True, that would be fine with me as well. Let's see what James thinks.

jmsmkn:
Yes, fine. Only risk is emails being sent out with these placeholders.

jmsmkn:
There are still blockers:

- 6x When setting the payment status to 'Issued', you must specify the internal invoice number.
- 6x When setting the payment status to 'Issued', you must specify the internal client number.
- 2x Please explain why the invoice is complimentary in the internal comments.
- 2x Contact name is required for non-complimentary invoices.
- 1x Contact email is required for non-complimentary invoices.
- 1x Billing address is required for non-complimentary invoices.
- 47x VAT number is required for non-complimentary invoices.

jmsmkn:
Please can someone address those?

jmsmkn:
> Blocked because we're waiting for Anne-Marie to update the excel sheet:
> 
> * [ ]      6x When setting the payment status to 'Issued', you must specify the internal invoice number.
> * [ ]      6x When setting the payment status to 'Issued', you must specify the internal client number.

Those 6 invoices have now become 10.

jmsmkn:
I have filled out the information for the 6, and for the 4 new ones set them to Invoice Requested.

PR #3896, opened 2025-03-18
Remove usages of transaction.atomic in tests

koopmant:


jmsmkn:
The use of atomic is not necessarily a bad thing. My question was more about the difference in the tests in the other PR where the same action was being carried out, but one in an atomic block and one not. It was about consistency of action rather than the use of atomic itself.

koopmant:
> The use of atomic is not necessarily a bad thing. My question was more about the difference in the tests in the other PR where the same action was being carried out, but one in an atomic block and one not. It was about consistency of action rather than the use of atomic itself.

I see, I did not get that yesterday. In that case, I'll go ahead and close this one because I don't think this is an improvement. 

Btw, one thing I found interesting to learn here is that `with pytest.raises` works fine if the transaction is at the end of the test. It makes sense, but it wasn't obvious to me.

PR #3897, opened 2025-03-18
Fix flaky test due to existing browser sessions

koopmant:
Fix flaky test when orphan browser sessions exist. Flaky test failure:
```
>       assert BrowserSession.objects.count() == 1
E       assert 11 == 1
E        +  where 11 = count()
```

### DIFF
--- a/app/tests/sessions_tests/test_models.py
+++ b/app/tests/sessions_tests/test_models.py
@@ -4,13 +4,9 @@ from grandchallenge.browser_sessions.models import BrowserSession
 from tests.factories import UserFactory
 
 
-@pytest.fixture(scope="module", autouse=True)
-def clear_sessions():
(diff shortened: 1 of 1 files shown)


jmsmkn:
This is a strange one. Do we know why this sometimes occurs?

koopmant:
> This is a strange one. Do we know why this sometimes occurs?

I don't have a clue...

koopmant:
It appears the browsersessions are persistant between tests for some reason. 
I think we could also use `@pytest.mark.django_db(transaction=True)` here to avoid the issue.

jmsmkn:
Okay - then it might be worth rewriting the test. It assumes that the table is initially empty, so instead fetch the existing pks, then do the action, get the new session by excluding pks in that list of existing ones.

jmsmkn (approved):

PR #3902, opened 2025-03-19
EvaluationInputSet model

amickan:


app/grandchallenge/evaluation/migrations/0077_create_evaluation_input_set.py
```diff
@@ -0,0 +1,117 @@
...
+            model_name="evaluation",
+            name="input_set",  # Temporary field name
+            field=models.ForeignKey(
+                blank=True,
+                null=True,
+                on_delete=django.db.models.deletion.PROTECT,
+                to="evaluation.evaluationinputset",
+            ),
+        ),
+        migrations.RunPython(update_evaluation_inputs, elidable=True),
```
jmsmkn, line 91: This needs to be split up into 3 migrations as data migrations need to be separate from table migrations.

app/grandchallenge/evaluation/models.py
```diff
@@ -1720,6 +1770,9 @@ class Evaluation(CIVForObjectMixin, ComponentJob):
     ground_truth = models.ForeignKey(
         EvaluationGroundTruth, null=True, blank=True, on_delete=models.PROTECT
     )
+    inputs = models.ForeignKey(
```
jmsmkn, line 1773: I think that overriding the field name is dangerous. When interacting with the base class `inputs` will return two separate things depending on the concrete model, one returns an object, the other a queryset, right?

The quickest fix here might be to give it a different name, but then we would need to make sure that `evaluation.inputs` is never used.

I think that a complete implementation would be to migrate other concrete classes of `ComponentJob` too.

app/grandchallenge/evaluation/migrations/0077_create_evaluation_input_set.py
```diff
@@ -0,0 +1,117 @@
...
+
+        evaluation.input_set = input_set
+        updated_evaluations.append(evaluation)
+
+        if len(updated_evaluations) >= batch_size:
+            Evaluation.objects.bulk_update(updated_evaluations, ["input_set"])
+            updated_evaluations = []
+
+    if updated_evaluations:
+        Evaluation.objects.bulk_update(evaluations, ["input_set"])
```
jmsmkn, line 44: This then updates all evaluations? Wasn't the point of the iterator not to have them in memory?

app/grandchallenge/evaluation/models.py
```diff
@@ -1710,6 +1712,54 @@ class EvaluationGroundTruthGroupObjectPermission(GroupObjectPermissionBase):
...
+        if not obj:
+            obj = super().create(**kwargs)
+            obj.values.set(inputs)
+
+        return obj
+
+
+class EvaluationInputSet(UUIDModel):
+    values = models.ManyToManyField(
+        ComponentInterfaceValue, related_name="eval_inputs"
```
jmsmkn, line 1734: We don't really need to use an abbreviation, the full name is clearer.

```suggestion
        ComponentInterfaceValue, related_name="evaluation_inputs"
```

app/grandchallenge/challenges/costs.py
```diff
@@ -67,6 +71,9 @@ def annotate_compute_costs_and_storage_size(*, challenge):
 def update_size_in_storage_and_registry(
     *, challenge, algorithm_jobs, evaluation_jobs
 ):
+    evaluation_inputs = EvaluationInputSet.objects.filter(
```
jmsmkn, line 74: Why do the separate lookup? This should be directly accessible as before.

jmsmkn (changes requested):
I think this looks okay, but I'm concerned about the API changes and the inconsistencies they introduce. Algorithm Jobs and Evaluations are coupled through ComponentJob so they should roughly be the same, but things like this, the normal vs special inputs and the permission differences are causing problems.

I think we need a bit of a redesign as working with them now is clearly difficult, I don't know if making this change so last minute in the cycle makes sense as we need to be sure on the direction.

amickan:
> I think this looks okay, but I'm concerned about the API changes and the inconsistencies they introduce. Algorithm Jobs and Evaluations are coupled through ComponentJob so they should roughly be the same, but things like this, the normal vs special inputs and the permission differences are causing problems.
> 
> I think we need a bit of a redesign as working with them now is clearly difficult, I don't know if making this change so last minute in the cycle makes sense as we need to be sure on the direction.

This exactly summarizes my concerns with this change as well. It feels messy, and I do agree that it would be better to migrate the `ComponentJob` base class instead, but also don't think that this is the kind of change you make lightly at the end of the cycle.... 

Not doing this now, does mean that we will need to sacrifice the unique constraint on the Evaluation though. As I explained on the pitch, I don't think it's a huge issue at this point since both entry points to creating evaluations (if we ignore that we can create evaluations through the shell technically) have a uniqueness check in place (see the other open PR). 

Nonetheless, I think it was worth trying this out, to see how much effort it would be to migrate to a FK field. Doing this on the base class will actually make things easier since the handling would the same everywhere again. I think if we do migrate the inputs on the ComponentJob model, we should do the same for the outputs. It would be very strange to have them be implemented differently.

app/grandchallenge/evaluation/models.py
amickan: I think this clashes with the related name for `inputs` on the `Phase` model...

app/grandchallenge/evaluation/models.py
amickan: I share your concerns, but that is why the migrations is a multi step process. I first add a new field with a different name (input_set), then migrate the inputs, then remove the old `inputs` field and then rename the new field from `input_set` to `inputs`. I tried this out, and from what I can tell it works fine. 

But I think we both agree that only migrating the inputs for the `Evaluation` anyway is suboptimal

jmsmkn:
Agreed, we're adding to the technical debt which does need to be paid back now but we can address that with a pitch.

app/grandchallenge/evaluation/migrations/0077_create_evaluation_input_set.py
amickan: Right, sorry, I had this as separate migrations initially, and then forgot why I did that and merged them into one. Sigh.

app/grandchallenge/evaluation/models.py
jmsmkn: Agreed, thank for trying!

PR #3905, opened 2025-03-19
Add `inputs.json`

jmsmkn:
This PR adds metadata about the jobs inputs through `inputs.json`. The structure is the same as the inputs section of `predictions.json`.

Closes https://github.com/DIAGNijmegen/rse-roadmap/issues/397

### DIFF
--- a/app/grandchallenge/components/backends/base.py
+++ b/app/grandchallenge/components/backends/base.py
@@ -25,7 +25,15 @@ from grandchallenge.components.backends.exceptions import (
     UncleanExit,
 )
 from grandchallenge.components.backends.utils import user_error
+from grandchallenge.components.models import (
+    ComponentInterface,
+    ComponentInterfaceValue,
+    InterfaceKindChoices,
+)
 from grandchallenge.components.schemas import GPUTypeChoices
+from grandchallenge.components.serializers import (
+    ComponentInterfaceValueSerializer,
+)
 from grandchallenge.core.utils.error_messages import (
     format_validation_error_message,
 )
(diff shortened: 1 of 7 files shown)


app/grandchallenge/components/backends/base.py
```diff
@@ -325,28 +308,60 @@ def _provision_inputs(self, *, input_civs, input_prefixes):
...
+        """
+        yield from input_civs
+
+        serializer = ComponentInterfaceValueSerializer(input_civs, many=True)
+        yield ComponentInterfaceValue(
+            value=serializer.data,
+            interface=ComponentInterface(
+                relative_path="inputs.json", kind=InterfaceKindChoices.ANY
+            ),
+        )
```
amickan, line 336: Do I understand this correctly, that both the `ComponentInterface` and the corresponding CIV are generated on the fly and never saved to the database? I guess the reason is that we don't want anyone to be able to define this as input/output, since it's an implicit input for all algorithms? And we don't want to create thousands of unnecessary CIVs  either. 

I guess you could technically create a dedicated interface for this and then exclude that interface from the input/output options on the interface form. Then you would also not need the extra validator you added below, and the CIV would only be saved to the DB once for a given input set and then reused for subsequent algorithm jobs, so it wouldn't be too many extra CIVs. 

Your implementation makes, and is more elegant, but it took a moment to understand, and I'm just curious what the thought process behind choosing this design was.

amickan (approved):

app/grandchallenge/components/backends/base.py
jmsmkn: That was exactly it, it's implicit based on the inputs. I didn't want exceptions anywhere and the CIV interface is how SageMaker Shim gets stuff into the container at the correct time and place. I'm wondering if the same thing should be done with predictions.json. But that all relates to the comments on the other PR as I think what we have has become too complex.

jmsmkn: The reason for not creating the inputs.json interface is so that values for that definitely will not get created as they would otherwise be overwritten by this implementation.

amickan: Ok, that makes perfect sense. Thanks for sharing your thoughts. I'll take a look at your comments on my PR now.

PR #3907, opened 2025-03-20
Make sure algorithm sockets don't overlap with evaluation sockets

amickan:
This adds validation to the forms to make sure that evaluation inputs and outputs (as defined on the phase) do not overlap with the algorithm inputs and outputs of the phase. 

### DIFF
--- a/app/grandchallenge/algorithms/forms.py
+++ b/app/grandchallenge/algorithms/forms.py
@@ -194,8 +194,8 @@ class JobCreateForm(AdditionalInputsMixin, Form):
         return cleaned_data
 
 
-# Exclude interfaces that are not aimed at algorithms from user selection
-NON_ALGORITHM_INTERFACES = [
+# Exclude sockets that are not aimed at algorithms or evaluations from user selection
+RESERVED_SOCKET_SLUGS = [
     "predictions-csv-file",
     "predictions-json-file",
     "predictions-zip-file",
@@ -1359,13 +1359,13 @@ class AlgorithmModelVersionControlForm(Form):
 class AlgorithmInterfaceForm(SaveFormInitMixin, ModelForm):
     inputs = ModelMultipleChoiceField(
         queryset=ComponentInterface.objects.exclude(
-            slug__in=NON_ALGORITHM_INTERFACES
+            slug__in=RESERVED_SOCKET_SLUGS
         ),
         widget=Select2MultipleWidget,
     )
     outputs = ModelMultipleChoiceField(
         queryset=ComponentInterface.objects.exclude(
-            slug__in=NON_ALGORITHM_INTERFACES
+            slug__in=RESERVED_SOCKET_SLUGS
         ),
         widget=Select2MultipleWidget,
     )
@@ -1450,7 +1450,7 @@ class AlgorithmInterfaceForm(SaveFormInitMixin, ModelForm):
                 str(obj) for obj in overlapping_sockets
             )
             raise ValidationError(
-                "The following sockets are already configured as sockets on"
+                "The following sockets are already configured as additional inputs or outputs on"
                 f" {self._base_obj}: {overlapping_names}"
             )
 
--- a/app/grandchallenge/algorithms/forms.py
+++ b/app/grandchallenge/algorithms/forms.py
@@ -1387,6 +1387,11 @@ class AlgorithmInterfaceForm(SaveFormInitMixin, ModelForm):
         if not inputs:
             raise ValidationError("You must provide at least 1 input.")
 
+        if self._base_obj.additional_inputs_field:
+            self.check_for_overlapping_sockets(
+                sockets=inputs,
+            )
+
         if (
             self._base_obj.algorithm_interface_through_model_manager.annotate(
                 input_count=Count("interface__inputs", distinct=True),
@@ -1412,7 +1417,36 @@ class AlgorithmInterfaceForm(SaveFormInitMixin, ModelForm):
         if not outputs:
             raise ValidationError("You must provide at least 1 output.")
 
+        if self._base_obj.additional_outputs_field:
+            self.check_for_overlapping_sockets(
+                sockets=outputs,
+            )
+
         return outputs
+
+    def check_for_overlapping_sockets(self, *, sockets):
+        overlapping_input_sockets = (
+            self._base_obj.additional_inputs_field.filter(
+                slug__in=sockets.values_list("slug", flat=True)
+            )
+        )
+        overlapping_output_sockets = (
+            self._base_obj.additional_outputs_field.filter(
+                slug__in=sockets.values_list("slug", flat=True)
+            )
+        )
+        overlapping_sockets = list(
+            chain(overlapping_input_sockets, overlapping_output_sockets)
+        )
+
+        if overlapping_sockets:
+            overlapping_names = ", ".join(
+                str(obj) for obj in overlapping_sockets
+            )
+            raise ValidationError(
+                "The following sockets are already configured as sockets on"
+                f" {self._base_obj}: {overlapping_names}"
+            )
 
     def clean(self):
         cleaned_data = super().clean()
--- a/app/grandchallenge/algorithms/forms.py
+++ b/app/grandchallenge/algorithms/forms.py
@@ -1387,7 +1387,10 @@ class AlgorithmInterfaceForm(SaveFormInitMixin, ModelForm):
         if not inputs:
             raise ValidationError("You must provide at least 1 input.")
 
-        if self._base_obj.additional_inputs_field:
+        if (
+            self._base_obj.additional_inputs_field
+            or self._base_obj.additional_outputs_field
+        ):
             self.check_for_overlapping_sockets(
                 sockets=inputs,
             )
@@ -1417,7 +1420,10 @@ class AlgorithmInterfaceForm(SaveFormInitMixin, ModelForm):
         if not outputs:
             raise ValidationError("You must provide at least 1 output.")
 
-        if self._base_obj.additional_outputs_field:
+        if (
+            self._base_obj.additional_inputs_field
+            or self._base_obj.additional_outputs_field
+        ):
             self.check_for_overlapping_sockets(
                 sockets=outputs,
             )
--- a/app/grandchallenge/evaluation/admin.py
+++ b/app/grandchallenge/evaluation/admin.py
@@ -1,6 +1,7 @@
 import json
 
 from django.contrib import admin, messages
+from django.core.exceptions import ValidationError
 from django.forms import ModelForm
 from django.utils.html import format_html
 from guardian.admin import GuardedModelAdmin
@@ -11,6 +12,7 @@ from grandchallenge.components.admin import (
     deprovision_jobs,
     requeue_jobs,
 )
+from grandchallenge.components.models import ComponentInterface
 from grandchallenge.core.admin import (
     GroupObjectPermissionAdmin,
     UserObjectPermissionAdmin,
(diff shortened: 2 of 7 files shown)


app/grandchallenge/evaluation/admin.py
```diff
@@ -44,13 +55,65 @@ class Meta:
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        self.fields["inputs"].queryset = ComponentInterface.objects.exclude(
+            slug__in=NON_EVALUATION_INTERFACES
+        )
```
amickan, line 60: Somehow this does not work. I still get the full list of CIs in the admin. Not sure what I'm doing wrong. 
Also, `NON_EVALUATION_INTERFACES` == `NON_ALGORITHM_INTERFACES`, but it seemed semantically wrong to use the latter here. We might also want to differentiate the two in the future.

jmsmkn: I think that the data from the autocomplete comes from how it is defined on the related model https://docs.djangoproject.com/en/5.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.autocomplete_fields

I think the only thing you can do here is to make sure that it is cleaned and rejected if selected.

amickan: Ah that makes sense. Thanks! I'll add check in the clean method.

app/grandchallenge/algorithms/forms.py
```diff
@@ -1387,6 +1387,11 @@ def clean_inputs(self):
         if not inputs:
             raise ValidationError("You must provide at least 1 input.")
 
+        if self._base_obj.additional_inputs_field:
```
koopmant, line 1390: Do you also need to check `if self._base_obj.additional_outputs_field`? (Or, I guess, no need to check at all in that case, just run the method.) Because I see you're using both in `check_for_overlapping_sockets`.

app/grandchallenge/evaluation/admin.py
```diff
@@ -51,6 +61,68 @@ def __init__(self, *args, **kwargs):
...
+
+        return inputs
+
+    def clean(self):
+        cleaned_data = super().clean()
+
+        if self.instance.submission_kind == SubmissionKindChoices.ALGORITHM:
+            input_slugs = (
+                set(
+                    cleaned_data.get("inputs", []).values_list(
```
koopmant, line 83: minor thing, but I don't think you need `.get("inputs", [])` here? Since you're also checking `if "inputs" in cleaned_data`

app/grandchallenge/evaluation/admin.py
```diff
@@ -51,6 +61,68 @@ def __init__(self, *args, **kwargs):
...
+                        "slug", flat=True
+                    )
+                )
+                if "inputs" in cleaned_data
+                else set()
+            )
+            output_slugs = (
+                set(
+                    cleaned_data.get("outputs", []).values_list(
```
koopmant, line 92: same

app/tests/evaluation_tests/test_admin.py
```diff
@@ -119,3 +122,38 @@ def test_reevaluate_submission_idempotent(rf):
...
+    )
+    assert ci1.slug in str(form.errors)
+    assert ci2.slug in str(form.errors)
+    assert ci3.slug not in str(form.errors)
+    assert ci4.slug not in str(form.errors)
+
+
+@pytest.mark.django_db
+def test_non_evaluation_interfaces():
+    ci = ComponentInterface.objects.get(slug="predictions-json-file")
```
koopmant, line 152:
```suggestion
@pytest.mark.parametrize("slug", NON_EVALUATION_INTERFACES)
def test_non_evaluation_interfaces(slug):
    ci = ComponentInterface.objects.get(slug=slug)
```

app/grandchallenge/evaluation/admin.py
```diff
@@ -35,6 +36,15 @@
     SubmissionGroupObjectPermission,
     SubmissionUserObjectPermission,
 )
+from grandchallenge.evaluation.utils import SubmissionKindChoices
+
+NON_EVALUATION_INTERFACES = [
```
koopmant, line 41:
```suggestion
NON_EVALUATION_SOCKET_SLUGS = [
```

app/grandchallenge/evaluation/models.py
```diff
@@ -1298,6 +1298,14 @@ def algorithm_interface_manager(self):
     def algorithm_interface_through_model_manager(self):
         return PhaseAlgorithmInterface.objects.filter(phase=self)
 
+    @property
+    def additional_inputs_field(self):
```
koopmant, line 1302: Why is it called `additional_inputs_field` if it returns `inputs`?

koopmant (commented):
I left some small remarks!

PS there is a typo in the PR title
<sub>evaluatiooooon, let's gooooooo</sub>

app/grandchallenge/evaluation/admin.py
amickan: True!

app/grandchallenge/evaluation/models.py
amickan: Because the inputs defined are by definition additional inputs. The standard inputs (algorithm job outputs, predictions.json or the predictions file) are not defined as socket inputs and also not processed like socket values. You could argue that those should be the special, additional inputs then, but since the socket-inputs that we're adding in this pitch are the new thing, and not the standard inputs, I've opted to call them additional inputs.

app/grandchallenge/algorithms/forms.py
amickan: I am doing the same check for outputs in `clean_outputs`, or what do you mean?

amickan: You're right though that I'm basically calling the same function twice, for no reason. I'll move the check to the clean() method so that it runs only once

amickan: I need to run this check twice, because once I check whether the chosen inputs overlap with either inputs or outputs on the phase (additional inputs/outputs), and then I do the same check for the outputs chosen on the form.

amickan: Ah right, that's true (currently there is no case where we have only additional outputs and not inputs, but of course that could exist in the future). I'll update this

app/grandchallenge/evaluation/models.py
koopmant: Ah I get it. Would it make sense to rename the field then? (I'm not saying we should, but it would be nice if it's no effort)

app/grandchallenge/evaluation/admin.py
koopmant: You're still using `get()` though 😅

app/grandchallenge/evaluation/models.py
amickan: The `inputs` field on the `Phase` model you mean? I think it's never easy to rename a model field, really. I'm also not sure if `additional_inputs` would be any clearer, or what would you rename it to? It would be nicest to rework the evaluation pipeline and have all inputs defined as sockets. That would work for prediction submission phases, but that doesn't really work for algorithm-submission phases, those will always need special treatment.

app/grandchallenge/evaluation/admin.py
amickan: Oops, sorry! Fixed ;-)

app/grandchallenge/evaluation/admin.py
```diff
@@ -35,6 +36,15 @@
     SubmissionGroupObjectPermission,
     SubmissionUserObjectPermission,
 )
+from grandchallenge.evaluation.utils import SubmissionKindChoices
+
+NON_EVALUATION_SOCKET_SLUGS = [
```
jmsmkn, line 41: This seems like duplicate data, shouldn't `NON_ALGORITHM_INTERFACES` be reused?

app/grandchallenge/evaluation/admin.py
```diff
@@ -51,6 +61,60 @@ def __init__(self, *args, **kwargs):
...
+
+            algorithm_socket_slugs = set(
+                self.instance.algorithm_interfaces.values_list(
+                    "inputs__slug", flat=True
+                )
+            ) | set(
+                self.instance.algorithm_interfaces.values_list(
+                    "outputs__slug", flat=True
+                )
+            )
```
jmsmkn, line 102: Can the algorithm interfaces be changed by this form? If so, I don't think that this lookup will be in sync.

jmsmkn (commented):
Looks good. Couple of questions, additionally: it looks like `clean_inputs` and `clean_outputs` depend on other fields. Should they be moved to the models `clean` method instead?

app/grandchallenge/evaluation/models.py
jmsmkn: I think we need to be really careful and should distinguish between algorithm inputs and additional inputs. It would be very easy for a site admin to accidentally put the sockets into `inputs` rather than `algorithm_inputs`, and potentially expose the values in the leaderboard.

In this case I don't think that renaming the field is that difficult as we can drop the existing field as there currently is no data, and add a new one with the proper name and through table.

app/grandchallenge/evaluation/admin.py
amickan: Not sure, it seemed semantically wrong to use `NON_ALGORITHM_INTERFACES` here and I also thought that we might want to distinguish between the two in the future, they just happen to overlap now.

app/grandchallenge/evaluation/admin.py
amickan: No, they cannot be changed by this form.

app/grandchallenge/evaluation/admin.py
jmsmkn: Okay then it's probably best to unify them and rename to save them from unintentionally getting out of sync.

amickan:
> it looks like `clean_inputs` and `clean_outputs` depend on other fields. Should they be moved to the models `clean` method instead?

Those are field specific clean methods, and they each depend on their field and then on 2 fields on the parent model. They do not depend on other fields from the same model. To me this seems a very typical use of field specific clean methods. What would be the benefit of moving the checks to the model's `clean` method?

app/grandchallenge/evaluation/models.py
amickan: You are right that in this case, renaming the inputs field would be ok since it does not currently hold data. I forgot that we wiped that field earlier on.

I don't really share the concern with mixing things up to be honest. The algorithm interfaces are not configured in the admin any longer, and trying to do so would be prevented anyway by the overlapping check that this PR implements. Even if someone were to add algorithm inputs as evaluation inputs, those would not be part of existing evaluations' inputs, and hence would not be exposed on the leaderboard. Adding algorithm outputs as evaluation outputs would be just as problematic - again that should not be possible with the check I'm adding here, and in any case would not be solved by only renaming the inputs field.  

If we rename the inputs field to `additional_evaluation_inputs`, for fear of people confusing it with algorithm inputs, I think we should also rename the outputs field to `evaluation_outputs`. The outputs field does hold data, which is actually being used in the pipeline. Not sure how much work it would be to update the references since I believe the code that parses the outputs is the same for jobs and evaluations, and assumes the outputs to be defined in an outputs fields on the parent model. So that would need to be pulled apart.

jmsmkn:
Okay, so long as those values are from a different model and they are not being changed in the same transaction then it is fine.

app/grandchallenge/evaluation/admin.py
amickan: Ok, makes sense. How about calling the unified list `RESERVED_SOCKET_SLUGS`?

app/grandchallenge/evaluation/models.py
amickan: Maybe to clarify/summarize, I'm ok with renaming the inputs field if you think that would be clearer, but I think it would be strange to not also rename the outputs field, and that is a bit trickier to do. And regarding the name, renaming it to `additional_inputs` doesn't really clarify / unobscure things, IMO, so I'd go for `additional_evaluation_inputs` and `èvaluation_outputs` respectively.

jmsmkn: Agreed on the naming. I think it would be clearer but I will leave it up to you. Does not necessarily need to be in this PR.

jmsmkn (approved):
LGTM!

PR #3908, opened 2025-03-21
Speculative fix for pycurl>7.45.2

### LINKED ISSUE #3252: Pycurl 7.45.3 breaks our celery workers
Pycurl version 7.45.3 is broken. Import fails with `ImportError: pycurl: libcurl link-time ssl backend (gnutls) is different from compile-time ssl backend (openssl)`

pkcakeout:
Most likely issue was this: https://github.com/pycurl/pycurl/blob/d92395d9fe1e18365c4e4812f4351bd6985872e1/RELEASE-NOTES.rst#pycurl-7455---2025-03-06

Due to lack of details, I was not able to confirm that this fixes #3252. But this should fix a C dependency issue stemming from pycurl now being built with a non-system libcurl library (likely statically linked) and libvips indirectly requiring the system libcurl version to be loaded which is a different version.

:warning: I pinned version `pycurl==7.45.6` - Not sure if this is intended, but for now to force that version.

### DIFF
--- a/dockerfiles/web-base/Dockerfile
+++ b/dockerfiles/web-base/Dockerfile
@@ -133,5 +133,5 @@ RUN apt-get update \
     && rm -rf /var/lib/apt/lists/*
 USER django:django
 
-RUN uv sync --no-progress
+RUN uv sync --no-progress --no-binary-package pycurl
 RUN playwright install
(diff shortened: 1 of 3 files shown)


jmsmkn:
Still no dice I'm afraid. There is a CI test for this which fails, but only on the PR branch: `ImportError: pycurl: libcurl link-time version (7.88.1) is older than compile-time version (8.12.1-DEV)`.

jmsmkn:
Thanks @pkcakeout ! I merged this in #3911.

PR #3910, opened 2025-03-24
Show additional inputs on leaderboard

amickan:
This adds the additional inputs to the leaderboard and the evaluation list:

sshot-87

### DIFF
--- a/app/grandchallenge/evaluation/templates/evaluation/evaluation_list.html
+++ b/app/grandchallenge/evaluation/templates/evaluation/evaluation_list.html
@@ -4,6 +4,7 @@
 {% load guardian_tags %}
 {% load url %}
 {% load static %}
+{% load civ %}
 
 {% block title %}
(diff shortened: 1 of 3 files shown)


jmsmkn:
Are the additional inputs being prefetched/selected in both places?

koopmant (commented):
Looks fine. See question of James. Also, would a test be overkill here? (If you do decide to write one you can use [assertInHTML(needle, haystack)](https://docs.djangoproject.com/en/4.2/topics/testing/tools/#django.test.SimpleTestCase.assertInHTML))

jmsmkn:
Those are using the UnitTest constructs, which we do not use as we run our tests with pytest. Instead, the pytest API for that would be `assert needle in haystack`, much nicer.

koopmant:
Right, I did not look closely enough at what I was linking to. I actually meant `assertInHTML` from `django_pytest`:
`from pytest_django.asserts import assertInHTML`. Django_pytest imports the function from SimpleTestCase, so the earlier link still explains what the function does.

jmsmkn (approved):
LGTM!

PR #3919, opened 2025-03-25
Avoid celery and kombu 5.5.x

jmsmkn:
See https://github.com/celery/kombu/issues/2258

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,8 +5,9 @@ authors = [
 license = {text = "Apache-2.0"}
 requires-python = "<3.13,>=3.11"
 dependencies = [
-    "celery[redis,sqs]!=5.0.6,!=5.3.0",
-    "kombu!=5.3.0,!=5.5.0",
+    # Pin kombu and celery due to performance issues, see https://github.com/celery/kombu/issues/2258
+    "celery[redis,sqs]!=5.0.6,!=5.3.0,!=5.5",
+    "kombu!=5.3.0,!=5.5",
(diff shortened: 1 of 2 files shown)


spawn-guy:
don't forget to add `pycurl` as a requirement if you plan to use `pycurl` instead of `urllib3` ;) as kombu will work the same new way on the next minor version as in the current one

jmsmkn:
Thanks! Hopefully the performance issues can be addressed in 5.7 as we would like to drop the dependency too. Don't know if it helps you but we saw a slowdown within the AWS environment (ECS Fargate->SQS, same region, no proxy) with `"--concurrency", "1", "--pool", "prefork"` workers, ACKS Late, prefetch multiplier 1.

spawn-guy:
thanks for the slowdown report. and for the full settings.

the thing is.. i can't reliably state that pycurl-ullib3 is the problem. 'cause my change was not the only one since 5,4,2.

on the other hand i have also enabled ssl connections for sqs which were disabled instead of "fixed" previously

it looks like i still need to test pycurl on aws anyway :(

jmsmkn:
No worries. We also `CELERY_BROKER_USE_SSL`, or at least I hope so as our IAM SQS roles have always had the condition:

```
    condition {
      test = "Bool"
      values = [
        true
      ]
      variable = "aws:SecureTransport"
    }
```

Celery is very complex and I get overwhelmed every time I try to take a look at an issue there, thanks for your efforts!

PR #3922, opened 2025-03-25
Ground Truth from Answers

chrisvanrun:
This PR is the final part on:

It adds a form that allows the ground truth to be transformed into answers for a selected user. The selected user should not have any existing answers.

To simplify the permissions I've opted to fully create new answers in an asyncronous task. However, because of wish to have the whole conversion in a single transaction I ran into problems with signals.

As a consequence I've also opted to boot the signal on the post_save of an `Answer` and pull the logic there into the `Answer.save()`. It used to be the case that an async task got spawned for each new `Answer` added. That is now done synchronously: the ground truth was being retrieved anyway. When added ground truth, which is done in bulk anyway, the re-calculation is now batched in a single task. Should be a bit more performant than a typical 1000 tasks spawned instantly.

Not sure about the current flag-based calculation of scores approach to `Answer.save()`: perhaps there is a better ay? It fixes an N+1 run off on creating answers from ground truth.

## Showcase

Showing the loop: answers to ground truth to answers to ground truth

https://github.com/user-attachments/assets/bd4f2915-6eb5-4a79-9a0b-ac888b3d519a

### DIFF
--- a/app/grandchallenge/reader_studies/forms.py
+++ b/app/grandchallenge/reader_studies/forms.py
@@ -16,6 +16,7 @@ from crispy_forms.layout import (
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
+from django.db import transaction
 from django.db.models import BLANK_CHOICE_DASH
 from django.forms import (
     BooleanField,
@@ -67,6 +68,7 @@ from grandchallenge.reader_studies.models import (
     CASE_TEXT_SCHEMA,
     Answer,
     AnswerType,
+    AnswerUserObjectPermission,
     CategoricalOption,
     Question,
     ReaderStudy,
@@ -783,15 +785,41 @@ class AnswersFromGroundTruthForm(SaveFormInitMixin, Form):
         return user
 
     def convert_answers(self):
+        user = self.cleaned_data["user"]
         all_answers = Answer.objects.filter(
             question__reader_study=self._reader_study,
         )
-        ground_truth = all_answers.filter(is_ground_truth=True)
-        ground_truth.update(
-            is_ground_truth=False,
-            creator=self.cleaned_data["user"],
-        )
+
+        # Reset all scores: ground truth has been removed
         all_answers.update(score=None)
+
+        # Convert and reassign answers
+        all_answers.filter(is_ground_truth=True).update(
+            is_ground_truth=False,
+            creator=user,
+        )
+
+        # Fix permissions
+        with transaction.atomic():
+            user_answers = all_answers.filter(creator=user)
+
+            # First, remove all other user permissions
+            AnswerUserObjectPermission.objects.filter(
+                content_object__in=user_answers
+            ).exclude(user=user).delete()
+
+            # Second, assign new permissions.
+            # Note: this already handles existing permissions
+            AnswerUserObjectPermission.objects.bulk_assign_perm(
+                perm="view_answer",
+                user_or_group=user,
+                queryset=user_answers,
+            )
+            AnswerUserObjectPermission.objects.bulk_assign_perm(
+                perm="change_answer",
+                user_or_group=user,
+                queryset=user_answers,
+            )
 
 
 class GroundTruthCSVForm(SaveFormInitMixin, Form):
--- a/app/grandchallenge/reader_studies/forms.py
+++ b/app/grandchallenge/reader_studies/forms.py
@@ -16,7 +16,6 @@ from crispy_forms.layout import (
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
-from django.db import transaction
 from django.db.models import BLANK_CHOICE_DASH
 from django.forms import (
     BooleanField,
@@ -68,13 +67,13 @@ from grandchallenge.reader_studies.models import (
     CASE_TEXT_SCHEMA,
     Answer,
     AnswerType,
-    AnswerUserObjectPermission,
     CategoricalOption,
     Question,
     ReaderStudy,
     ReaderStudyPermissionRequest,
 )
 from grandchallenge.reader_studies.tasks import (
+    answers_from_ground_truth,
     bulk_assign_scores_for_reader_study,
 )
 from grandchallenge.subdomains.utils import reverse, reverse_lazy
@@ -764,11 +763,6 @@ class AnswersFromGroundTruthForm(SaveFormInitMixin, Form):
 
         self._reader_study = reader_study
 
-        # self.fields["user"].queryset = (
-        #     get_user_model()
-        #     .objects.filter(answer__question__reader_study=reader_study)
-        #     .distinct()
-        # )
         self.fields["user"].queryset = (
             self._reader_study.editors_group.user_set.all()
             | self._reader_study.readers_group.user_set.all()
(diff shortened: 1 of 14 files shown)


app/grandchallenge/reader_studies/tasks.py
```diff
@@ -34,24 +26,29 @@ def add_image(obj, image):
...
+    reader_study = ReaderStudy.objects.get(pk=reader_study_pk)
+    target_user = get_user_model().objects.get(pk=target_user_pk)
+
+    all_answers = Answer.objects.filter(question__reader_study=reader_study)
+    if all_answers.filter(is_ground_truth=False, creator=target_user).exists():
+        raise ValueError("User already has answers")
+
+    ground_truth = all_answers.filter(is_ground_truth=True)
+
+    with transaction.atomic():
```
jmsmkn, line 40: Code style - wrap the entire task in `@transaction.atomic` if possible.

app/grandchallenge/reader_studies/forms.py
```diff
@@ -741,8 +743,55 @@ def create_ground_truth(self):
...
+            ).apply_async
+        )
+
+
+class AnswersFromGroundTruthForm(SaveFormInitMixin, Form):
+    user = ModelChoiceField(
+        queryset=get_user_model().objects.none(),
+        required=True,
+        help_text=format_html(
+            "Select a user to whom the answers will assigned. "
```
jmsmkn, line 759: This is having one user assign answers for another. I think we want to avoid that for traceability reasons, especially thinking about the requirements that just came in and the problems that it caused with submitting on behalf of someone else. Maybe we should limit it to only being able do this for the current user?

jmsmkn (commented):
LGTM, small things and a question about who the answers can be transferred to.

app/grandchallenge/reader_studies/forms.py
chrisvanrun: I think that is a very reasonable limitation. I've opted for an open-user selection initially as I could vaguely imagine someone asking to double check the answer, generally though: I suspect that it's always the editor re-assigning it to themselves.

chrisvanrun:
Had a slight change of heart: instead of both copying the answers AND deleting the ground truth, I've made it only copy the ground truth as answers.

This action can only be done for the current user. Who needs permission to change the reader study.

# Latest showcase

https://github.com/user-attachments/assets/c1048f87-0fe3-4e5b-a7b2-d2e66556df50

app/grandchallenge/reader_studies/forms.py
```diff
@@ -741,8 +744,59 @@ def create_ground_truth(self):
...
+            is_ground_truth=False,
+        ).exists():
+            raise ValidationError(
+                "User already has answers. Delete these first"
+            )
+
+        return super().clean()
+
+    def schedule_answers_from_ground_truth_task(self):
+        answers_from_ground_truth.apply_async(
```
jmsmkn, line 795: `apply_async` should be always be wrapped in an `on_commit`, otherwise the celery task could run before the changes are committed to the database. Preferred style is how it is below with `on_commit(task.signature(...).apply_async)`.

app/grandchallenge/reader_studies/tasks.py
```diff
@@ -20,38 +18,33 @@
...
+        raise ValueError("User already has answers")
+
+    for answer in all_answers.filter(is_ground_truth=True).all():
+        # Simplify permissions and create new answers
+        answer._state.adding = True
+        answer.id = None
+        answer.pk = None
+        answer.is_ground_truth = False
+        answer.creator = target_user
+        answer.save(calculate_score=False)
```
jmsmkn, line 45: There is a race condition here - there is a check to see if the user has answers, then the answers are created. In the meantime, the user could create some and there would be duplicates. To solve that `answer.full_clean()` should be called before `answer.save(...)`. That would then make the pre-flight check redundant. Hopefully duplicates are checked in the models clean method, if not, that should be added there.

See also https://realpython.com/python-lbyl-vs-eafp/

app/grandchallenge/reader_studies/tasks.py
```diff
@@ -20,38 +18,33 @@
...
+    if all_answers.filter(is_ground_truth=False, creator=target_user).exists():
+        raise ValueError("User already has answers")
+
+    for answer in all_answers.filter(is_ground_truth=True).all():
+        # Simplify permissions and create new answers
+        answer._state.adding = True
+        answer.id = None
+        answer.pk = None
+        answer.is_ground_truth = False
+        answer.creator = target_user
```
jmsmkn, line 44: This is potentially changing the creator. That should be kept the same and maybe just check that the target user is the same? That could be added to the filter above. Also note that the `.all()` is unnecessary.

app/grandchallenge/reader_studies/forms.py
chrisvanrun: My thought process on not adding it here: there are no changes that need to be commited. Hence it is not wrapped in an on_commit. I could add it, but it would it not be purely decorative?

The only thing I can think of if is that if the transaction fails somehow this would still run. Not sure how it could considering the transaction is empty?

jmsmkn: That's exactly what this form prevents.

The transaction can fail for a wide array of reasons that are out of your control. We have this guideline as we have faced these sorts of bugs in the past and they are a massive pain. The construction isn't "purely decorative", I would never ask for that and it's a little bit offensive that you consider my suggestions that. I ask for changes as I am the one who has to deal with outages and this construction prevents an entire class of them.

If there is no transaction then the callback will run immediately. If there is a transaction then the callback will be called once everything is done. There are no downsides to adding it.

app/grandchallenge/reader_studies/tasks.py
jmsmkn: Like what was discussed in the last review, the owner shouldn't change. This task should not be concerned with permissions, those should be checked in the view and instead the task should only be responsible for carrying out the work it was asked to do.

app/grandchallenge/reader_studies/forms.py
chrisvanrun: > The transaction can fail for a wide array of reasons that are out of your control. We have this guideline as we have faced these sorts of bugs in the past and they are a massive pain. The construction isn't "purely decorative", I would never ask for that and it's a little bit offensive that you consider my suggestions that. I ask for changes as I am the one who has to deal with outages and this construction prevents an entire class of them.

No offense was ever meant! We sometimes style things because it is done that way, in that particular app, such as adding `View` to some class-based view names and others.

I was honestly wondering if the `on_commit` was useful here, and what rational behind adding it would be. I've added it to the coding guideline.

app/grandchallenge/reader_studies/tasks.py
chrisvanrun: As discussed offline: overwriting the creator on the copy is reasonable.

app/grandchallenge/reader_studies/tasks.py
chrisvanrun: Saving throws an `IntegrityError` here if an answer already exists: as expected. So for clearness I've removed the earlier global check.

I thought it good practice to also add a `full_clean` here. However, **that** opened up a whole can of worms with respect to the `Answer.full_clean` being broken and that it is currently being crutched by `Answer.validate()`. A small dip into that rabbit hole convinced me very quickly that this PR is not the place and I've opted to just use `Answer.validate()`. That should mark this clearly as a place where a `full_clean` should occur.

jmsmkn: Urgh, yes, I remember that now. That should be refactored if possible, someday.

app/grandchallenge/reader_studies/tasks.py
```diff
@@ -20,38 +18,38 @@
...
+        Answer.validate(
+            creator=answer.creator,
+            question=answer.question,
+            answer=answer.answer,
+            display_set=answer.display_set,
+            is_ground_truth=answer.is_ground_truth,
+        )
+        answer.save(calculate_score=False)
+
+    all_answers.update(score=None)
```
jmsmkn, line 52: Why are the scores reset here? The ground truth still exists so the answers should still have scores, right? I would have thought that they would only need to be reset when the ground truth is removed.

jmsmkn (commented):
Looks good, one small question

app/grandchallenge/reader_studies/tasks.py
chrisvanrun: Ah, that is a left over from the initial logic where we were actively destroying ground truth instead of copying it. Removed it now and also removed the associated tests snippets.

jmsmkn (approved):
LGTM!

chrisvanrun:
> LGTM!

Thanks for many reviews James!

PR #3924, opened 2025-03-26
Add filter for overdue status of invoices to admin

### LINKED ISSUE #3917: Add filter for overdue invoices in django admin
A filter would be useful for us here

Image

koopmant:
Closes #3917


jmsmkn (dismissed):
LGTM but it may be more robust to do this with `TextChoice` and matching rather than hardcoding the keys.

koopmant:
Better!

jmsmkn (approved):
LGTM, thanks!

PR #3926, opened 2025-03-26
Cache the RDS token generation

jmsmkn:
The auth tokens expire after 15 minutes and should be reused, see https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.html#UsingWithRDS.IAMDBAuth.Limitations

We don't use redis for this as the tokens should be local, so can do everything locally.

See https://github.com/DIAGNijmegen/rse-cloud-infrastructure/issues/214


app/grandchallenge/core/db/postgres_iam/base.py
```diff
@@ -1,17 +1,72 @@
...
+from datetime import timedelta
+
 import boto3
 from django.db.backends.postgresql import base
+from django.utils import timezone
+
+
+def thread_local_cache(ttl, max_size=128):
+    def decorator(func):
+        function_cache = {}
```
amickan, line 11: Doesn't this create a cache that all threads can access, rather than a thread-local cache? I'm probably misunderstanding what you mean by thread-local...

jmsmkn: Okay, I meant just local to the worker running the code, not shared across workers as done with the django cache. I have replaced it with a third party package for simplicity.

PR #3927, opened 2025-03-27
Rename phase inputs and outputs

amickan:
This renames the `inputs` and `outputs` fields on the `Phase` model to `additional_evaluation_inputs` and `evaluation_outputs` respectively. 


amickan:
After testing the migration, it looks like it doesn't create through model entries automatically. So that means that migration will need to happen in multiple steps, probably:
- create new M2M fields (rather than renaming the old ones) + custom through tables
- data migration to create through table entries
- deprecate the original fields

jmsmkn (approved):
LGTM!

PR #3928, opened 2025-03-27
Refactor usage of UserPassesTestMixin

koopmant:
Replace usage of UserPassesTestMixin with AccessMixin and place the test directly in dispatch to allow for subclassing with additional tests. 

Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/347

### DIFF
--- a/app/grandchallenge/admins/views.py
+++ b/app/grandchallenge/admins/views.py
@@ -1,6 +1,6 @@
-from django.contrib.auth.mixins import LoginRequiredMixin
 from django.contrib.messages.views import SuccessMessageMixin
 from django.views.generic import FormView, ListView
+from guardian.mixins import LoginRequiredMixin
 
 from grandchallenge.admins.forms import AdminsForm
 from grandchallenge.core.guardian import ObjectPermissionRequiredMixin
--- a/app/grandchallenge/admins/views.py
+++ b/app/grandchallenge/admins/views.py
@@ -1,6 +1,6 @@
+from django.contrib.auth.mixins import LoginRequiredMixin
 from django.contrib.messages.views import SuccessMessageMixin
 from django.views.generic import FormView, ListView
-from guardian.mixins import LoginRequiredMixin
 
 from grandchallenge.admins.forms import AdminsForm
 from grandchallenge.core.guardian import ObjectPermissionRequiredMixin
--- a/app/grandchallenge/algorithms/views.py
+++ b/app/grandchallenge/algorithms/views.py
@@ -6,8 +6,8 @@ from django.conf import settings
 from django.contrib import messages
 from django.contrib.auth.mixins import (
     AccessMixin,
-    LoginRequiredMixin,
     PermissionRequiredMixin,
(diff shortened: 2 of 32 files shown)


koopmant:
2f15d4f is not strictly necessary, but I think it helps by showing that you can and should call super().test_func() if you inherit these classes.

amickan:
Thomas and I synchronously discussed this and realized that there is another `LoginRequiredMixin` (form `django.contrib.auth.mixins`) that subclasses `AccessMixin`, but otherwise does the exact same thing as the guardian mixin. So we think that's the one we should be using instead.

amickan:
If we all agree that this is the correct solution, then it might be worth adding the guardian mixin to the banned modules in setup.cfg to make sure we use the correct one in the future.

koopmant:
I just removed the usages of UserPassesTestMixin again, but realized we have the same problem as before. Also replacing `guardian.mixins.LoginRequiredMixin` with `django.contrib.auth.mixins.LoginRequiredMixin` shows the same issue with a lot of other tests. The mixin from django.contrib.auth simply returns the redirect from `handle_no_permission()`. But that means you have to inspect the returned object if you want to stop there and just return the redirect before doing anything else. 

At this point I think creating the new class is the best way to go.

koopmant:
Anne and I discussed it synchronously again. 

Indeed the problem is that you cannot update the dispatch method with a user test in a class where you use guardian LoginRequiredMixin and still have the benefit of checking if the user is logged in first (and if not logged-in automatically skipping the user test). 

An alternative would be to create dedicated mixins whereever you want to create a user test together with LoginRequiredMixin while you want (or would expect) that the login required check is performed first and stops the chain if it fails. Such a mixin would then inherit AccessMixin and the dispatch method would be updated in the mixin. Then it would still be wrapped by LoginRequiredMixin. But you would need to create such a mixin even if you use it on only one class, so that is not intuitive. And nothing prevents you from forgetting it and updating dispatch in the class itself. 

We also discussed using this new class everywhere we are now using django's AccessMixin combined with LoginRequiredMixin, since this has the same behaviour (if you update the dispatch method in the final class, the login check is not wrapped.) However, in these cases we add tests not related to the user. So the login check does not necessarily need to go first. If we do want to go this route, I would rename to new class to AccessMixin and ban the use of both UserPassesTestMixin and django's AccessMixin.

jmsmkn:
I'm not sure about this one, especially with leaving the default implementation to return `True`, I worry that could somehow end up as the only implementation and allow permission pass through. Do we need our own implementation of `LoginRequiredMixin` instead?

koopmant:
@jmsmkn Did you have a pattern in mind there? I've not yet found an alternative to the guardian and django implementations, but that may be a case of WYSIATI...

jmsmkn:
I have no idea I'm afraid! All options have their drawbacks, I wonder if we should then just stick with what the frameworks provide.

jmsmkn:
Not sure where to go here so closing this, can be re-opened.

PR #3930, opened 2025-03-28
Image processing and assignment: account for object not existing

### LINKED ISSUE #3929: Building images for objects - strange states when target object no longer exists
**Describe the bug**

**To Reproduce**
Create an `ArchiveItem`/`DisplaySet` with an image.

Before the `build_image` and `assign_image_to_object` can run, remove the `ArchiveItem`/`DisplaySet`.

Suddenly the `RawImageUploadSession` is stuck in REQUEUE state and the `assign_image_to_object` fails.

**Expected behavior**

Tasks succeed gracefully: it is OK if the target no longer exists, nothing to do.

chrisvanrun:
Closes #3929

There are three steps involved in image processing:

1. `RawImageUploadSession.process_images` (sync)
    A. Interval
2.  `tasks.build_images` (async)
    B. Interval
3. `task.assign_image_to_object` (async)

During both interval A and B the target `DisplaySet` / `ArchiveItem` could be removed by the user.

This PR adds some handling for that: it should gracefully ignore it when it refers to `DisplaySets` / `ArchiveItems`, never a `Job`.

It also applies the same logic to the adding of files.

### DIFF
--- a/app/grandchallenge/cases/tasks.py
+++ b/app/grandchallenge/cases/tasks.py
@@ -138,10 +138,6 @@ def build_images(  # noqa:C901
         app_label=RawImageUploadSession._meta.app_label,
         model_name=RawImageUploadSession._meta.model_name,
     )
-    if linked_interface_slug:
-        ci = ComponentInterface.objects.get(slug=linked_interface_slug)
-    else:
-        ci = None
 
     if linked_object_pk:
         try:
@@ -156,11 +152,22 @@ def build_images(  # noqa:C901
                 DisplaySet._meta.model_name,
             ]:
                 # users can delete archive items and display sets before this task runs
+                logger.info(
+                    f"Nothing to do here: {linked_model_name} no longer exists."
+                )
+                upload_session.update_status(
+                    status=RawImageUploadSession.CANCELLED
+                )
                 return
             else:
                 raise e
     else:
         linked_object = None
+
+    if linked_interface_slug:
+        ci = ComponentInterface.objects.get(slug=linked_interface_slug)
+    else:
+        ci = None
 
     error_handler = upload_session.get_error_handler(
         linked_object=linked_object
--- a/app/grandchallenge/cases/tasks.py
+++ b/app/grandchallenge/cases/tasks.py
@@ -156,7 +156,9 @@ def build_images(  # noqa:C901
                     f"Nothing to do here: {linked_model_name} no longer exists."
                 )
                 upload_session.update_status(
-                    status=RawImageUploadSession.CANCELLED
+                    status=RawImageUploadSession.CANCELLED,
+                    error_message="Image processing canceled. "
+                    f"The associated {linked_model_name} not longer exists.",
                 )
                 return
             else:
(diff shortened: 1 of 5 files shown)


app/grandchallenge/cases/tasks.py
```diff
@@ -156,12 +152,23 @@ def build_images(  # noqa:C901
                 DisplaySet._meta.model_name,
             ]:
                 # users can delete archive items and display sets before this task runs
+                logger.info(
+                    f"Nothing to do here: {linked_model_name} no longer exists."
+                )
+                upload_session.update_status(
+                    status=RawImageUploadSession.CANCELLED
```
amickan, line 159: maybe add an error message here for the user? Something along the lines of "Linked object no longer exists."

app/tests/components_tests/test_tasks.py
```diff
@@ -487,6 +489,83 @@ def test_add_image_to_object_marks_job_as_failed_on_validation_fail(
...
+        )
+    if task is add_file_to_object:
+        ci = ComponentInterfaceFactory(kind="PDF")
+        task_kwargs.update(
+            {
+                "user_upload_pk": None,
+                "interface_pk": ci.pk,
+            }
+        )
+
```
amickan, line 560: Can this be part of the parametrization so you don't need the logic here?

amickan (commented):
LGTM, just two minor comments / questions.

app/grandchallenge/cases/tasks.py
chrisvanrun: We could, but then the user get's informed about it. Do we really want that?

amickan: Well, it's not secret, right? They deleted the object, so they know already. 
I agree that it's not strictly necessary, but it might help them understand why certain uploads were cancelled if they later look at their overview.

chrisvanrun: I am in doubt since generally they aren't even aware that an upload is being processed in the backend.  I doubt, but you seem certain so let's roll with a notification!

chrisvanrun: Done!

amickan (dismissed):

app/grandchallenge/components/tasks.py
```diff
@@ -1191,6 +1191,25 @@ def validate_voxel_values(*, civ_pk):
...
+            app_label=app_label, model_name=model_name, pk=object_pk
+        )
+    except ObjectDoesNotExist as e:
+        if model_name in [
+            ArchiveItem._meta.model_name,
+            DisplaySet._meta.model_name,
+        ]:
+            return None
+        else:
+            raise e
```
koopmant, line 1209: suggestion, but not important:
```suggestion
    except (ArchiveItem.ObjectDoesNotExist, DisplaySet.ObjectDoesNotExist):
        return None
```
Also fyi, you can use raise without having to define "e". It will reraise the exception.

koopmant (commented):
I saw this task failure today, nice to see it fixed!

Added a tiny suggestion.

jmsmkn (changes requested):
The clean-up suggested by @koopmant would be good to add here.

app/grandchallenge/components/tasks.py
chrisvanrun: Far cleaner, nice suggestion!

jmsmkn (dismissed):
LGTM! I removed a specialist function as given its name and signature I would have expected it to behave differently - it is more pythonic to do exception handling rather than returning `None` and checking for that, so used the original method which behaves the same for all objects.

jmsmkn (approved):

PR #3939, opened 2025-04-02
Allow algorithms with additional interfaces to be submitted to phase

### LINKED ISSUE #3931: Algorithm that implements more than the required phase algorithm interfaces should still be selectable on the submission form
For UNICORN, users should be able to initially create an algorithm for the "all interfaces" phase and then submit that algorithm to the task-specific phases. So in other words, algorithms that implement more than the required interfaces should still be allowed for submission. Currently, we only allow algorithms that implement the exact same interfaces (and hence also the same number) as defined on the phase. 

This is a requirement that was not described in the pitch, and hence the difference in implementation, but I do not see why we would not allow algorithms that support more than the required interfaces to be submitted. 

The requirement for the algorithm should be that it implements at least the defined interfaces on the phase. 

Would you agree, @jmsmkn ?

amickan:
Closes #3931

### DIFF
--- a/app/grandchallenge/algorithms/forms.py
+++ b/app/grandchallenge/algorithms/forms.py
@@ -493,7 +493,6 @@ class UserAlgorithmsForPhaseMixin:
         return (
             get_objects_for_user(self._user, "algorithms.change_algorithm")
             .annotate(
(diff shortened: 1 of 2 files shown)


jmsmkn (approved):
LGTM!

PR #3942, opened 2025-04-04
Leaderboard accessible to readers

amickan:
This introduces the option to make the leaderboard accessible to readers. When enabled, readers can compare their own performance to those of other readers, while keeping their names and profiles hidden. 

Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/386


app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
```diff
@@ -86,6 +86,11 @@
                     </a>
                 {% endif %}
             {% endif %}
+            {% if 'change_readerstudy' not in readerstudy_perms and 'read_readerstudy' in readerstudy_perms and object.is_educational and object.leaderboard_accessible_to_readers %}
+                <a class="nav-link" href="{% url 'reader-studies:leaderboard' slug=object.slug %}">
+                       <i class="fas fa-trophy fa-fw"></i>&nbsp;Leaderboard
+                    </a>
+            {% endif %}
```
amickan, line 93: Editors also get the `read` permission, hence the double check on permissions here. I also considered moving the leaderboard tab down here for editors as well, but then it's put at the very bottom, and I think it's better to keep it where it is now for editors.

app/grandchallenge/reader_studies/views.py
```diff
@@ -354,13 +354,25 @@ def get_context_data(self, **kwargs):
 
 
 class ReaderStudyLeaderBoard(
-    LoginRequiredMixin, ObjectPermissionRequiredMixin, DetailView
+    LoginRequiredMixin, UserPassesTestMixin, DetailView
```
amickan, line 357: I'm not 100% sure anymore what the correct way is. As Thomas found out, using `AccessMixin` together with `LoginRequiredMixin` doesn't work unless you create a seperate class for the permission check. `UserPassesTestMixin` can be used like this directly, but can't be combined with other UserPassesTestMixins - not a problem now, per se, but could be in the future. The choice depends a bit on what we decide to do in https://github.com/comic/grand-challenge.org/pull/3928

jmsmkn (changes requested):
This looks good but I think that this would be cleaner with a custom `view_leaderboard` permission, which is assigned to the editors group and then added/removed to the readers group when the reader study is saved. That should save a lot of the repeated logic in the view and templates.

app/grandchallenge/reader_studies/views.py
amickan: No longer relevant thanks to James' suggestion of using a custom permission.

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
amickan: I have opted to pull the leaderboard tab up to avoid checking for it in two places.

jmsmkn (approved):
LGTM!

PR #3948, opened 2025-04-09
Update leaderboard and evaluation detail page

amickan:
This makes sure that the inputs column only shows on the leaderboard, if there are additional inputs. 
This also fixes the evaluation detail page so that it shows additional outputs (again).


jmsmkn (approved):
Looks good! Curious as to where `{% if json %}` in the evaluation detail was from, do you know?

PR #3950, opened 2025-04-09
Add created and modified fields to verification admin

HarmvZ:
Adds fields so that admins can easily see how old the verification requests are.

### DIFF
--- a/app/grandchallenge/verifications/admin.py
+++ b/app/grandchallenge/verifications/admin.py
@@ -55,6 +55,8 @@ class VerificationAdmin(admin.ModelAdmin):
         "email_is_verified",
         "is_verified",
         "verified_at",
+        "created",
+        "modified",
     )
     list_select_related = ("user__user_profile",)
     list_filter = ("email_is_verified", "is_verified")
--- a/app/grandchallenge/verifications/admin.py
+++ b/app/grandchallenge/verifications/admin.py
@@ -56,7 +56,6 @@ class VerificationAdmin(admin.ModelAdmin):
         "is_verified",
         "verified_at",
         "created",
-        "modified",
     )
     list_select_related = ("user__user_profile",)
     list_filter = ("email_is_verified", "is_verified")


jmsmkn:
The table is already very wide. Do you really need both?

HarmvZ:
Depends on what we decide for the SOP (see [discussion](https://teams.microsoft.com/l/message/19:5e954fc2c34246a7b846532578ec87c3@thread.tacv2/1744199058785?tenantId=b208fe69-471e-48c4-8d87-025e9b9a157f&groupId=4289f730-75fb-4e70-b879-98c686114ada&parentMessageId=1744199058785&teamName=Grand%20Challenge&channelName=Support%20Duty&createdTime=1744199058785)). If we want to know when the verification request was made(=created) and when/if the confirmation email was resent (=modified) we need both. 

If we automate the process by creating a task that resends the confirmation email when the token first expires and marks as not verified when the second token expires. We won't need this at all.

jmsmkn:
If the token has expired it can be marked as rejected or the request can be deleted, no need to send the email again.

jmsmkn:
I also don't really see the need to clean up. What is wrong with also just leaving them as-is? This prevents a user from creating multiple to spam multiple addresses. If they want to try again they can ask support to delete it.

HarmvZ:
Sure we can also leave them. Then the _created_ field can be used to see when you've arrived at old verifications when going through the list.

jmsmkn:
That can also be achieved with the existing filters, no?

Screenshot 2025-04-09 at 18 26 03

That will show you only the ones that need action in the daily support duty.

HarmvZ:
I may be misunderstanding the SOP. But I think you still need to check the users with a non-verified email to actively mark them as not verified when they pass any of those conditions (user info incomplete, team accounts, etc).

jmsmkn:
You do not need to do that.

HarmvZ:
Ok, I've updated the SOP.

PR #3956, opened 2025-04-14
Reduce number of challenge leaderboard queries

amickan:


app/grandchallenge/evaluation/views/__init__.py
```diff
@@ -835,8 +840,12 @@ def get_queryset(self, *args, **kwargs):
                 "submission__phase__challenge",
                 "submission__algorithm_image__algorithm",
             )
-            .prefetch_related("outputs__interface", "inputs__interface")
+            .prefetch_related("outputs__interface")
         )
+
+        if self.additional_inputs_defined_on_phase:
+            queryset = queryset.prefetch_related("inputs")
```
jmsmkn, line 847: At the moment this will prefetch all inputs, including the 1000's of values for each evaluation. Should a `Prefetch` object be used to only get the inputs that need to be displayed? See `https://docs.djangoproject.com/en/5.2/ref/models/querysets/#django.db.models.query.QuerySet.prefetch_related`.

amickan: Very good point. I have made the changes.

jmsmkn (approved):

PR #3965, opened 2025-04-16
Create model for session costs

koopmant:
Adds new model `SessionCost` which contains relevant information on the costs of a session. 

Related to https://github.com/DIAGNijmegen/rse-roadmap/issues/365


app/grandchallenge/workstations/models.py
```diff
@@ -728,3 +739,79 @@ class FeedbackUserObjectPermission(UserObjectPermissionBase):
 
 class FeedbackGroupObjectPermission(GroupObjectPermissionBase):
     content_object = models.ForeignKey(Feedback, on_delete=models.CASCADE)
+
+
+class SessionCost(UUIDModel):
+    session = models.OneToOneField(
```
jmsmkn, line 745: I think `models.OneToOneField` is only a soft restriction in Django. Can we add a database constraint to ensure that there is only one `SessionCost` per `Session`?

app/grandchallenge/workstations/models.py
```diff
@@ -728,3 +739,79 @@ class FeedbackUserObjectPermission(UserObjectPermissionBase):
...
+        Session,
+        related_name="session_cost",
+        null=True,
+        on_delete=models.SET_NULL,
+    )
+    duration = models.DurationField()
+    creator = models.ForeignKey(
+        settings.AUTH_USER_MODEL, null=True, on_delete=models.SET_NULL
+    )
+    reader_studies = models.ManyToManyField(
```
jmsmkn, line 755: `models.ManyToManyField` should have a through model, and that model should have constraints on the relationship.

app/grandchallenge/workstations/models.py
```diff
@@ -728,3 +739,79 @@ class FeedbackUserObjectPermission(UserObjectPermissionBase):
...
+                    },
+                    "uniqueItems": True,
+                }
+            )
+        ],
+    )
+    credits_consumed = models.PositiveIntegerField(
+        null=True,
+        blank=True,
+        editable=False,
```
jmsmkn, line 784: The combination of blank and non-editable may be meaningless, doesn't blank only count when editing?

app/grandchallenge/workstations/models.py
```diff
@@ -728,3 +739,79 @@ class FeedbackUserObjectPermission(UserObjectPermissionBase):
...
+    )
+
+    def save(self, *args, **kwargs) -> None:
+        adding = self._state.adding
+
+        super().save(*args, **kwargs)
+
+        if adding and self.session:
```
jmsmkn, line 792: `self.session` should always be set here right? Probably we should error out if not, which is easy enough to do by just setting `if adding:`.

app/grandchallenge/workstations/models.py
```diff
@@ -728,3 +739,79 @@ class FeedbackUserObjectPermission(UserObjectPermissionBase):
...
+            self.creator = self.session.creator
+            self.reader_studies.set(self.session.reader_studies.all())
+            self.interactive_algorithms = list(
+                Question.objects.filter(
+                    reader_study__in=self.reader_studies.all()
```
jmsmkn, line 797: Can you reliably query self at this point? You can anyway locally keep the reader studies with `reader_studies = self.session.reader_studies.all()` and then do the two usages of it.

app/grandchallenge/workstations/models.py
```diff
@@ -728,3 +739,79 @@ class FeedbackUserObjectPermission(UserObjectPermissionBase):
...
+                )
+                .exclude(interactive_algorithm="")
+                .values_list("interactive_algorithm", flat=True)
+                .order_by()
+                .distinct()
+            )
+            self.credits_consumed = self.calculate_credits_consumed()
```
jmsmkn, line 804: Do we really need to store this in the database? Looking at the calculation below it could be a straightforward property.

jmsmkn (changes requested):
Looks good! Just some changes needed with the model, important to get the right from the start to avoid difficult migrations.

koopmant:
Thanks for the quick review!

app/grandchallenge/workstations/models.py
koopmant: This made the most sense to me, because the cost objects are like receipts or historical records. If the calculation changes over time, this should not affect objects that were created earlier. Do you think I should change it to a property?

app/grandchallenge/workstations/models.py
koopmant: Good point, thanks. In my tests it was no issue, but storing it locally indeed makes more sense.

app/grandchallenge/workstations/models.py
```diff
@@ -728,3 +739,79 @@ class FeedbackUserObjectPermission(UserObjectPermissionBase):
...
+            super().save()
```
jmsmkn, line 805: This creates a loop where the object is saved twice, why not do the property manipulation pre-saving?

app/grandchallenge/workstations/models.py
jmsmkn: I think that a property is better until it is needed.

app/grandchallenge/workstations/models.py
koopmant: Thanks, changed it.

app/grandchallenge/workstations/models.py
jmsmkn: Okay! I thought that we hit that problem earlier but I must be mistaken.

app/grandchallenge/workstations/models.py
koopmant: Err, yes. This is silly. changed it.

app/grandchallenge/workstations/models.py
koopmant: Ahh, thanks for this! It was a leftover from when I still did this in save():
```python
Question.objects.filter(
    reader_study__in=self.reader_studies.all()
```
At which point I couldn't query self until after saving. 
Having fixed that, we no longer need the weird double save().

app/grandchallenge/workstations/models.py
```diff
@@ -728,3 +739,79 @@ class FeedbackUserObjectPermission(UserObjectPermissionBase):
...
+                        "enum": InteractiveAlgorithmChoices.values,
+                        "type": "string",
+                    },
+                    "uniqueItems": True,
+                }
+            )
+        ],
+    )
+    credits_consumed = models.PositiveIntegerField(
+        null=True,
```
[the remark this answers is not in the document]
koopmant: I suppose there could be a meaning to it, but it's not something we need. Removed it.

app/grandchallenge/workstations/models.py
```diff
@@ -583,6 +589,11 @@ def stop(self) -> None:
         if self.auth_token:
             self.auth_token.delete()
 
+        SessionCost.objects.create(
```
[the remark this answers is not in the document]
koopmant: Moved it.

app/grandchallenge/workstations/models.py
koopmant: Done.

koopmant: For documentation.

Setting the intermediary model also removes the widget from the admin, [as explained here in the docs]

When the intermediary model is not set, django automatically creates this multi select widget. 
(https://docs.djangoproject.com/en/4.2/ref/contrib/admin/#working-with-many-to-many-intermediary-models). 
image

I don't think that can be achieved when the intermediary model is set. Instead you'd have to use a tabular inline widget. 

Currently displaying the reader studies using `readonly_fields` because I don't think we'll be editing these relationships:
image

app/grandchallenge/workstations/models.py
koopmant: Changed it.

jmsmkn (approved):
LGTM!

PR #3968, opened 2025-04-17
Disable reader studies without credits

koopmant:
Disable launching a reader study when all credits have been consumed. 

When max_credits is none, there is no limit. 

Banner for reader
image

Banner for editor
image

Related to https://github.com/DIAGNijmegen/rse-roadmap/issues/365


app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
```diff
@@ -124,7 +124,11 @@ <h2>{{ object.title }}</h2>
             {% if "read_readerstudy" in readerstudy_perms %}
                 <p>
                     <button class="btn btn-primary"
-                        {% workstation_session_control_data workstation=object.workstation context_object=object reader_study=object %}
+                        {% if object.is_launchable %}
+                            {% workstation_session_control_data workstation=object.workstation context_object=object reader_study=object %}
+                        {% else %}
+                            disabled
```
chrisvanrun, line 130: Not sure if we should disable the button or just remove it completely.

If we disable it though: I suggest adding a title attribute to provide a tooltip explaining things.

pkcakeout: I think that tooltips makes a lot more sense than just removing it which leaves people confused when documentation does not match their interface.

jmsmkn: Easiest things to do would be to hoist the warning up and adjust the message for the admin/reader.

app/grandchallenge/reader_studies/models.py
```diff
@@ -339,6 +341,10 @@ class ReaderStudy(
             "Usernames and avatars will be hidden to protect other readers' privacy."
         ),
     )
+    max_credits = models.PositiveSmallIntegerField(
+        default=0,
```
jmsmkn, line 345: Here, zero means infinite credits right? Maybe it is better to make this nullable and default to null. This way, zero credits means zero, none means unbound. The check then becomes:

```python
@property
def is_launchable(self):
    return self.max_credits is None or self.credits_consumed < self.max_credits
```

koopmant: Thanks! done

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
koopmant: Thanks for the input all. I replaced it with an informative banner.

app/grandchallenge/components/templates/components/civ_set_detail.html
```diff
@@ -26,15 +26,19 @@ <h2>{{ object|verbose_name|title }} </h2>
     <h4>{{ object.pk }}</h4>
 
     <p>
-        <a class="btn btn-primary"
+        <button class="btn btn-primary"
```
jmsmkn, line 29: Why the change from `a` to `button`? Same in `civ_set_row`.

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_detail.html
```diff
@@ -170,6 +178,11 @@ <h3>Admin Info</h3>
                         Each reader will read the cases in the same order.
                     {% endif %}
                 </p>
+                {% if object.questions_with_interactive_algorithm.exists and object.max_credits is not None %}
+                    <p>
+                        This reader study has {{ object.questions_with_interactive_algorithm.count|pluralize:"a question,questions" }} with an interactive algorithm. This doubles the credit consumption rate for this study.
```
jmsmkn, line 183: Better not to commit to how:

```suggestion
                        This reader study has {{ object.questions_with_interactive_algorithm.count|pluralize:"a question,questions" }} with an interactive algorithm. This increases the credit consumption rate for this study.
```

app/grandchallenge/reader_studies/forms.py
```diff
@@ -656,7 +656,7 @@ class Meta:
         initial=initial_interactive_algorithm,
         choices=interactive_algorithm_choices,
         required=False,
-        help_text="Select an interactive algorithm for this question.",
+        help_text="Select an interactive algorithm for this question. Please note that setting an interactive algorithm will double the credit consumption rate.",
```
jmsmkn, line 659: Flexibility:

```suggestion
        help_text="Select an interactive algorithm for this question. Please note that setting an interactive algorithm will increase the credit consumption rate.",
```

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_statistics.html
```diff
@@ -47,8 +47,13 @@ <h2>Statistics per case</h2>
...
                         <td data-order="{{ entry.id }}">
-                            <button title="Open in viewer" class="btn badge badge-primary"
-                                    {% workstation_session_control_data workstation=object.workstation context_object=object display_set=entry reader_study=object %}
+                            <button class="btn btn-primary badge badge-primary"
+                                    {% if object.is_launchable %}
+                                        {% workstation_session_control_data workstation=object.workstation context_object=object display_set=entry reader_study=object %}
+                                    {% else %}
+                                        disabled title="Reader study cannot be launched"
+                                    {% endif %}
+                                    title="Open in viewer"
```
jmsmkn, line 56: The title is set twice here if the study is not launchable.

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_progress.html
```diff
@@ -54,7 +54,11 @@ <h2>Users Progress for {{ reader_study.title }}</h2>
                     <div class="col-6 col-md-3 p-1">
                         <div class="d-flex justify-content-md-end justify-content-start align-items-center">
                             <button class="btn btn-primary"
-                                    {% workstation_session_control_data workstation=reader_study.workstation context_object=reader_study reader_study=reader_study user=user.obj %}
+                                    {% if reader_study.is_launchable %}
+                                        {% workstation_session_control_data workstation=reader_study.workstation context_object=reader_study reader_study=reader_study user=user.obj %}
+                                    {% else %}
+                                        disabled title="Reader study cannot be launched"
+                                    {% endif %}
```
jmsmkn, line 61: This block seems to be repeated a lot. Could it make more sense to move the logic into `workstation_session_control_data`?

app/grandchallenge/workstations/models.py
```diff
@@ -680,18 +680,16 @@ def handle_reader_study_switching(self, *, workstation_path):
...
-            Question.objects.filter(reader_study=reader_study)
-            .exclude(interactive_algorithm="")
-            .exists()
-        ):
-            on_commit(
-                preload_interactive_algorithms.signature(
-                    queue=f"workstations-{self.region}"
-                ).apply_async
-            )
+        if reader_study.is_launchable:
```
jmsmkn, line 684: Else shouldn't they get a permission denied or something?

jmsmkn (commented):
Looks good, some small things.

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_statistics.html
koopmant: I started this in civ_set_row, where we have a nested if-statement. In practice, the second definition is ignored. I think it's best explained here: https://stackoverflow.com/a/43859478.
I see that in this case it's not needed though.

app/grandchallenge/components/templates/components/civ_set_detail.html
koopmant: We have styling for :disabled but this attribute is not supported by an `a` element. Alternatively we could use the disabled class, but using the attribute is easier because we can then add it together with the title. Otherwise the closing quote for the classes is in between. 

Given that this is a navigation link rather than an in-page action, one could argue that using an anchor is more appropriate here. However, we use the button element a lot for navigation links.

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_progress.html
koopmant: I thought of doing that. But it makes the function more complicated. I'd rather not call it at all if we don't want to have the session control data. It's also a bit difficult to do in `civ_set_row` and `civ_set_detail` because we don't pass the reader_study to the function there. 

I could maybe write a new function or template partial?

app/grandchallenge/workstations/models.py
koopmant: Yes they should. I wasn't sure about where to do that yesterday. I've done it in `SessionCreate.form_valid()` now. The session is still created though, as the new test demonstrates. Do you think we should prevent that as well?

jmsmkn: The method should not return True or False, an exception should be raised, see https://realpython.com/python-lbyl-vs-eafp/

If that error is not appropriate for the view then a `try ... except` should be used to handle this appropriately.

A more difficult problem is that `form_valid` should just work as at this point all of the data should be valid, and all that needs to happen is hooking in any side effects, e.g. scheduling of async tasks.

The problem here is that the reader study is invalid for a session, so that should be caught earlier on form validation, not after the object is created. That might need a bit more refactoring, it might be tricky.

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_progress.html
jmsmkn: Okay that's fine, then yes, a partial or function might be best.

app/grandchallenge/components/templates/components/civ_set_detail.html
jmsmkn: Clear!

app/grandchallenge/workstations/models.py
koopmant: > The method should not return True or False, an exception should be raised, see https://realpython.com/python-lbyl-vs-eafp/

Always the same mistake 😅 

> That might need a bit more refactoring, it might be tricky.

I'll look into it.

koopmant: I think I got it. Added validation to `SessionForm`.

app/grandchallenge/reader_studies/templates/reader_studies/readerstudy_progress.html
koopmant: Now that I look at it, the block is slightly different each time because of different arguments to `workstation_session_control_data` and once because of an additional title attribute. I think it's better to leave it like this.

jmsmkn (approved):
LGTM!

PR #3970, opened 2025-04-18
Update forge

chrisvanrun:
Part of pitch:
- https://github.com/DIAGNijmegen/rse-roadmap/issues/393

This updates forge version to v0.7.1, introducing:
- adaption to `AlgorithmInterface`
- Parameterized Evaluations
- Host of other minor fixes (permissions, comments et cetera)

Side bonus is that locally the generation of an algorithm templates is now 70-90 ms ( was 100-120ms).


jmsmkn (changes requested):
This looks good but please rebase on main, the changes to uv.lock will need to be regenerated.

app/grandchallenge/core/utils/grand_challenge_forge.py
```diff
@@ -22,10 +46,41 @@ def get_forge_challenge_pack_context(challenge, phase_pks=None):
...
+        )
+        .order_by("slug")
+        .prefetch_related(
+            Prefetch(
+                "archive",
+                queryset=Archive.objects.order_by("created"),
+            ),
+            Prefetch(
+                "additional_evaluation_inputs",
+                queryset=ComponentInterface.objects.order_by("slug"),
```
jmsmkn, line 64: Why force ordering for all of these things, including the interfaces below? This changes the ordering from what users see in the UI.

chrisvanrun: Consistent generation of templates: there is no innate ordering on the contained sockets. IMHO regenerating the template should be a repeatable action.

jmsmkn: They should be ordered by pk already https://github.com/comic/grand-challenge.org/blob/f2a784cb8ae1cb4575a767ceb219402267496e13/app/grandchallenge/components/models.py#L703

chrisvanrun: Ah, not sure how I missed that! It's been there for over 4 years!

I do remember a different point about `Meta.ordering` long ago: explicit usage that requires ordering (or assumes it) should explicit require it. And having a 'central' ordering was generally lead to a lot of useless DB actions.

Ofc we now already have that ordering in the `CI.Meta`. So I think it is fine to not be explicit about it downstream since in practice we never remove central ordering? Would you agree?

jmsmkn: We should do ordering in one place which is in the models Meta method. That way we can ensure that the appropriate indices are set so that ordering is efficient as it is never free (gratis).

Of course there are cases where explicit ordering is needed, but there needs to be a good reason for it. In this case, I think we can go with the default ordering everywhere.

chrisvanrun: Ah I see now why I originally did this: The `AlgorithmInterface` are not ordered by default. Should we, in this case, add a central `AlgorithmInterface.Meta.ordering` or here explicitly?

Moreover, the `"created"` would be my first to go for ordering but it seems those are not indiced by default. Considering all related DB queries will not really have a large number of final objects would it be fine to pick a non-indiced column?

jmsmkn: Yes that would be fine.

uv.lock
```diff
@@ -1891,6 +1891,15 @@ wheels = [
     { url = "https://files.pythonhosted.org/packages/21/2c/5e05f58658cf49b6667762cca03d6e7d85cededde2caf2ab37b81f80e574/pillow-11.2.1-pp311-pypy311_pp73-win_amd64.whl", hash = "sha256:208653868d5c9ecc2b327f9b9ef34e0e42a4cdd172c2988fd81d62d2bc9bc044", size = 2674751, upload_time = "2025-04-12T17:49:59.628Z" },
 ]
 
+[[package]]
+name = "pkgconfig"
```
chrisvanrun, line 1895: This is a bit of an odd one. It is marked as a build dependency for `pyvips` packaged. I updated my local `uv` and the Python version to see if it got excluded. However, I suspect the inclusion might be based off MacOS vs Linux (Dependency update GitHub action). A few additional local-package upgrades later I've thrown in the towel for such a small abbaration.

chrisvanrun:
Thanks for the review James!

PR #3973, opened 2025-04-24
Prevent reevaluating under certain conditions

### LINKED ISSUE #3800: Mark evaluations with (outdated) invalid algorithm as failed
When re-evaluating submissions for a challenge phase, it is possible that a submission is no longer valid for the challenge. This can happen if the interface for the phase changed in the meantime (i.e. the algorithm inputs are different now). Re-evaluating such a submission will currently leave the linked evaluation stuck in `Executing algorithm` state, because:
- in `create_algorithm_jobs` we `filter_civs_for_algorithm` based on the inputs that are set on the algorithm. So even if new archive items are passed to `create_algorithm_jobs` those get filtered out --> no new jobs to create --> `set_evaluation_inputs` gets called
- `set_evaluation_inputs` does nothing if not `self.inputs_complete`, where we check that there is exactly 1 successful job per valid archive item

The newly created evaluation for such submission should be marked as Failed (or Cancelled?) with an appropriate error message (i.e. "Algorithm inputs don't match those defined for the phase"). We should check and catch that early, probably in `create_algorithm_jobs`. 

First discovered here: https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/428#issuecomment-2615173942

### LINKED ISSUE #3847: Do not allow challenge admins to reevaluate submissions in certain scenarios
- There is an existing evaluation for the algorithm image
- There are failed/cancelled jobs for the archive + algorithm image + model

amickan:
Closes #3800 
I think it also closes #3847


app/grandchallenge/evaluation/models.py
```diff
@@ -1670,6 +1670,20 @@ def create_evaluation(self, *, additional_inputs):
...
+            )
+
+        if self.has_blocking_algorithm_jobs:
+            evaluation.update_status(
+                status=Evaluation.CANCELLED,
+                error_message="There are non-successful jobs for this submission. "
+                "These need to be handled first before you can "
+                "re-evaluate. Please contact support.",
+            )
+
```
amickan, line 1686: I have opted to create the evaluation and mark it as failed for more direct feedback to the admins. The alternative would be to send a notification like we do for missing methods, but I feel like those just get lost, and we (siet admins) wouldn't get those.

app/grandchallenge/evaluation/models.py
```diff
@@ -1715,6 +1729,46 @@ def get_absolute_url(self):
...
+            items = get_archive_items_for_interfaces(
+                algorithm_interfaces=algorithm_interfaces,
+                archive_items=self.phase.archive.items.all(),
+            )
+            non_success_statuses = [
+                status[0]
+                for status in Job.STATUS_CHOICES
+                if status[0] != Job.SUCCESS
+            ]
```
amickan, line 1757: The issue only mentions failed and cancelled jobs, but I think we want to check for any non-successful jobs as re-evaluating should not be possible if there are queued jobs etc either.

app/grandchallenge/evaluation/models.py
```diff
@@ -1715,6 +1730,42 @@ def get_absolute_url(self):
             },
         )
 
+    @cached_property
+    def has_matching_algorithm_interfaces(self):
+        return sorted(
+            [int.pk for int in self.phase.algorithm_interfaces.all()]
+        ) == sorted(
+            [int.pk for int in self.algorithm_image.algorithm.interfaces.all()]
+        )
```
jmsmkn, line 1739: Easier as a set comparison?

```suggestion
        return (
            {*self.phase.algorithm_interfaces.all()} == {*self.algorithm_image.algorithm.interfaces.all}
        )
```

app/grandchallenge/evaluation/models.py
```diff
@@ -1715,6 +1730,42 @@ def get_absolute_url(self):
...
+            algorithm_interfaces=algorithm_interfaces,
+            archive_items=self.phase.archive.items.all(),
+        )
+        non_success_statuses = [
+            status[0]
+            for status in Job.STATUS_CHOICES
+            if status[0] != Job.SUCCESS
+        ]
+
+        jobs = get_valid_jobs_for_interfaces_and_archive_items(
```
jmsmkn, line 1756: I'm a little bit worried that this will take too long for a synchronous view. Could we short-cut it? Something like:

```python
    @cached_property
    def has_blocking_algorithm_jobs(self):
        return Job.objects.exclude(status=Job.SUCCESS).filter(inputs__archive_items__archive__phase=self.phase, algorithm_image=self.algorithm_image, algorithm_model=self.algorithm_model).exists()
```

This way you wouldn't need to modify `get_valid_jobs_for_interfaces_and_archive_items`.

amickan: This would match jobs with partial inputs (or additional inputs). We only want to know about non-successful jobs with the exact same inputs as the archive items, so I'm afraid the complex check is necessary.

amickan: I get your point about this possibly taking too long though. Is this something we could compute periodically and save on the evaluation object? I think it shouldn't change, technically, after an evaluation finishes, but there might be corner cases.

jmsmkn: Hmm. We shouldn't have a periodic task as the data would then not be in sync. Why not do this asynchronously and show the evaluation as failed? It is anyway not possible to evaluate in this scenario.

amickan: That could work.

jmsmkn (approved):
LGTM!

PR #3974, opened 2025-04-25
Phase Starter kit Download

chrisvanrun:
Sort of an extension on pitch:
-    git push --set-upstream origin starter-kit-generation

This PR adds the option to download a starter kit (aka challenge pack) for a phase:

image


app/grandchallenge/evaluation/templates/evaluation/phase_starter_kit.html
```diff
@@ -0,0 +1,58 @@
...
+       The starter kit is a Git-ready, custom-tailored set of examples designed specifically for this phase,
+       including input handling and output generation. It provides a strong foundation for building the necessary components
+       for this challenge phase.
+    </p>
+
+    <p>
+        The starter kit includes:
+        <ul>
+            <li>A script to <strong>automate uploading data</strong> to an archive</li>
+            <li>An <strong>submission algorithm</strong> that can be uploaded to run as a submission in a challenge phase</li>
```
amickan, line 33:
```suggestion
            <li>A <strong>submission algorithm</strong> that can be uploaded to run as a submission in a challenge phase</li>
```

app/grandchallenge/evaluation/views/__init__.py
```diff
@@ -1412,3 +1419,56 @@ def get_context_data(self, *args, **kwargs):
             }
         )
         return context
+
+
+class PhaseStarterKitInfo(
```
amickan, line 1424: We usually call information views details views, right? So maybe rename to `PhaseStarterKitDetail`? Then again, it's not a typical detail view, as in we don't have a PhaseStarterKit model.

app/grandchallenge/evaluation/views/__init__.py
```diff
@@ -1412,3 +1419,56 @@ def get_context_data(self, *args, **kwargs):
...
+    LoginRequiredMixin,
+    ObjectPermissionRequiredMixin,
+    DetailView,
+):
+    model = Phase
+    permission_required = "evaluation.change_phase"
+    raise_exception = True
+
+    def get(self, *_, **__):
+        phase = self.get_object()
```
amickan, line 1452: Don't you need a custom `get_object` here as well to subset by challenge?

amickan: You could probably subclass the `PhaseStartKitDetail` view here, since the permissions are the same as well.

app/grandchallenge/evaluation/templates/evaluation/phase_starter_kit.html
```diff
@@ -0,0 +1,58 @@
...
+        <p>
+            This phase does not have an archive linked to it or is not set up as an algorithm-accepting phase.
+        </p>
+        <p>
+            These are pre-requisites for generating a starter kit.
+        </p>
+        <p>
+            Contact <a href="mailto:support@grand-challenge.org">Grand Challenge Support</a> for help with getting these set up.
+        </p>
+
```
amickan, line 55: My first reaction here was: why show this tab at all for prediction submission challenges? But on second thought, I think it's good advertisement for the starter packs and for algorithm submission phases. Makes the admins aware of the fact that their set-up is not complete yet, and the prospect of getting a starter kit might just give them the extra incentive to get things set-up.

app/grandchallenge/evaluation/views/__init__.py
jmsmkn: Agreed, I think that this should be a `TemplateView`, that provides the `get_permission_object` returning the phase. The `PhaseStarterKitDownload` is probably just a `View`. No need to go crazy with the mixins for permissions, usually it is a single statement and covered by tests.

app/grandchallenge/evaluation/templates/evaluation/phase_starter_kit.html
chrisvanrun: I've mirrored the choices made for the linked algorithms.

app/grandchallenge/evaluation/views/__init__.py
chrisvanrun: I agree, 'info' is probably an incorrect name here. Since we don't have a direct model we are viewing or manipulating I think the `DetailView` can indeed far better be replaced by `TemplateView` / `View`.

However, afaik these don't have any permission checks in them. Instead of patching `dispatch()` I found it cleaner to keep using the permission mixin. Not sure: shadowing the `dispatch()` feels wrong for these typical permission checks.

Since these comments also apply to `PhaseArchiveInfo` I've taken the liberty of also applying these changes there.

amickan (approved):

PR #3975, opened 2025-04-25
Save session cost object before setting reader studies relation

koopmant:
Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/506

### DIFF
--- a/app/grandchallenge/workstations/models.py
+++ b/app/grandchallenge/workstations/models.py
@@ -759,10 +759,11 @@ class SessionCost(UUIDModel):
     )
 
     def save(self, *args, **kwargs) -> None:
-        if self._state.adding:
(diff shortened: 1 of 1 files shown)


koopmant:
Should be as simple as this I think. I'm not sure why this doesn't happen locally for me though...

amickan (approved):
Makes sense! As for prod vs dev: maybe this is related to auto-commit mode in de vs. everything running in a transaction in production? Not sure, but I concur that your fix should work.

koopmant:
I've looked into this a bit further and it could have been caught by [testing transactions](https://pytest-django.readthedocs.io/en/latest/database.html#testing-transactions), although this slows down the test:
```python
@pytest.mark.django_db(transaction=True)
```

PR #3977, opened 2025-04-29
Add task to remove old container images

jmsmkn:
This PR adds a task that can remove a container image from the registry and storage. Another PR will follow that adds cleanup of old items once this has been tested in production.

See https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/297


amickan (approved):

amickan:
LGTM. For my understanding: `remove_tag_from_registry` only removes the image from ECR. We're not removing the image from S3, so technically we're not fully removing the image from storage, or am I wrong?

jmsmkn:
> LGTM. For my understanding: `remove_tag_from_registry` only removes the image from ECR. We're not removing the image from S3, so technically we're not fully removing the image from storage, or am I wrong?

That is correct.

jmsmkn:
It would be nice to store everything in ECR, but our realised costs there are ~10x more than S3 for the same amount of data.

amickan:
Right, and S3 costs are so low (with intelligent-tiering) that it's not worth deleting from there, I guess.

jmsmkn:
Oh no, what this task does is remove it from s3 (storage). There are a lot of junk containers that do not work and just take up space, it is a significant amount.

amickan:
> Oh no, what this task does is remove it from s3 (storage). There are a lot of junk containers that do not work and just take up space, it is a significant amount.

Oh so when you delete the `image` from the model, that also deletes the referenced file on s3.

jmsmkn:
It does.

PR #3982, opened 2025-05-01
Seperate storage and compute cost calculation

amickan:
Speculative fix for https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/507


app/grandchallenge/challenges/costs.py
```diff
@@ -40,7 +41,12 @@ def annotate_job_duration_and_compute_costs(*, phase):
     )
 
 
-def annotate_compute_costs_and_storage_size(*, challenge):
+class JobsAndEvaluations(NamedTuple):
+    algorithm_jobs: list[Job]
+    evaluation_jobs: list[Evaluation]
+
+
+def filter_object_for_compute_and_storage_update(*, challenge):
```
koopmant, line 49:
```suggestion
def filter_objects_for_compute_and_storage_update(*, challenge):
```

koopmant:
Apart from the suggestions, LGTM.

jmsmkn:
I'm going to close this for the time being as I think that we need to profile all queries to get an understanding of what is going on, we may need to do extra work to ensure the costs and storage tasks do not run at the same time.

PR #3983, opened 2025-05-01
Do not allow a default value when sockets require a file

### LINKED ISSUE #3945: Level: Error 'dict' object has no attribute 'isdigit' in FlexibleFileField
In https://github.com/comic/grand-challenge.org/blob/fee46ae9bd9b143048d5c893a87a82d0270a24b1/app/grandchallenge/components/form_fields.py#L164

`initial` can be a dictionary in some circumstances:

```
{
disabled: False,
help_text: "JSON Task description ",
label: "foo",
required: True
}
```

This raises `AttributeError 'dict' object has no attribute 'isdigit'`

koopmant:
closes #3945


jmsmkn (approved):
LGTM

PR #3984, opened 2025-05-02
Allow cancelling jobs in validating inputs state

amickan:
There are currently 2 jobs stuck in validating inputs state. We should be able to cancel those in the admin. 

We should also look into why these got stuck, of course, but cancelling them should be possible anyway.

### DIFF
--- a/app/grandchallenge/components/admin.py
+++ b/app/grandchallenge/components/admin.py
@@ -129,6 +129,7 @@ def cancel_jobs(modeladmin, request, queryset):
             ComponentJob.PARSING,
             ComponentJob.RETRY,
             ComponentJob.EXECUTING_PREREQUISITES,
+            ComponentJob.VALIDATING_INPUTS,
         ]
     ).select_for_update(of=("self",), skip_locked=True).update(
         status=ComponentJob.CANCELLED


koopmant:
Not entirely sure why you assigned me, but the change looks good. 

If I look at the rest of the code I do have some questions, but none of them are important for this PR. 
Let me ask them anyway...
- Why can we not cancel `STARTED` and `PROVISIONING` jobs? 
- What's the difference between `STARTED` and `EXECUTING`?
- I don't think we ever use `PROVISIONING` nor `PARSING`, I cannot find it in the code at least. Should we remove those from the admin filter? 
- Can we remove this comment in ComponentJob? (I don't this is up-to-date since we have more than the original five)
  ```
  # The job statuses come directly from celery.result.AsyncResult.status:
  # http://docs.celeryproject.org/en/latest/reference/celery.result.html
  ```

amickan:
>     * Why can we not cancel `STARTED` and `PROVISIONING` jobs?
There might not be a specific reason. I think we kept extending the filter here as we needed, and neither of these 2 statuses is used, so there just hasn't been a need for adding them, would be my guess. 
 
>     * What's the difference between `STARTED` and `EXECUTING`?
Good question and I have no idea. I don't think we use `STARTED` at all. 

>     * I don't think we ever use `PROVISIONING` nor `PARSING`, I cannot find it in the code at least. Should we remove those from the admin filter?
True, I don't think we use these, at least not anymore. Not sure if it applies here, but sometimes these intermediate statuses can't be set, because it all happens in a transaction that ends with another status update. The intermediate status update then would never be committed. I guess if they're not used, they could indeed be deleted. 

>     * Can we remove this comment in ComponentJob? (I don't this is up-to-date since we have more than the original five)
>       ```
>       # The job statuses come directly from celery.result.AsyncResult.status:
>       # http://docs.celeryproject.org/en/latest/reference/celery.result.html
>       ```
I tend to agree that this is no longer true, and would hence be ok to remove.

amickan:
> Not entirely sure why you assigned me, but the change looks good.

I assigned you because neither James nor Chris are in, and you seemed like a logical choice. ;-)

koopmant:
> I assigned you because neither James nor Chris are in, and you seemed like a logical choice. ;-)

I meant because there is hardly anything to review 😅 
But you have my approval!

koopmant (approved):

amickan:
> I meant because there is hardly anything to review 😅 But you have my approval!

Oh yeah, well I need someone to approve it for me ;-)  Thanks!

PR #3985, opened 2025-05-02
Enable downloading of evaluation outputs

amickan:
Necessary to enable downloading of additional evaluation outputs. Currently, only MONKEY uses this feature (and possibly one more challenge, but can't remember the name now), but they would like to download the detailed metrics file that they are writing in addition to metrics.json. 

Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/508

### DIFF
--- a/app/grandchallenge/serving/models.py
+++ b/app/grandchallenge/serving/models.py
@@ -131,7 +131,10 @@ def get_component_interface_values_for_user(
     # they have access to as output.
     # Serving of evaluation inputs might become necessary in the future if
     # challenges start using additional evaluation inputs of type file. Serving of
-    # those should be restricted to admins as well.
+    # those MUST be restricted to admins.
+    #
+    # Challenge Participants MUST NOT have access to evaluation inputs as those
+    # contain the results on the hidden test set.
     evaluation_outputs = (
         evaluation_query.filter(outputs__in=civs)
         .distinct()
--- a/app/grandchallenge/serving/models.py
+++ b/app/grandchallenge/serving/models.py
@@ -121,7 +121,7 @@ def get_component_interface_values_for_user(
     evaluation_query = filter_by_permission(
         queryset=Evaluation.objects.all(),
         user=user,
-        codename="view_evaluation",
+        codename="change_evaluation",
         accept_user_perms=False,
     )
     evaluation_outputs = (
--- a/app/grandchallenge/serving/models.py
+++ b/app/grandchallenge/serving/models.py
@@ -124,6 +124,14 @@ def get_component_interface_values_for_user(
         codename="change_evaluation",
         accept_user_perms=False,
     )
+    # We restrict downloading of evaluation outputs to challenge admins since those
+    # might not be intended for participants to see (unlike the metrics).
+    # Note: we currently do not allow direct serving of evaluation inputs.
+    # Challenge admins can already download those because they are attached to jobs
+    # they have access to as output.
+    # Serving of evaluation inputs might become necessary in the future if
+    # challenges start using additional evaluation inputs of type file. Serving of
+    # those should be restricted to admins as well.
     evaluation_outputs = (
         evaluation_query.filter(outputs__in=civs)
         .distinct()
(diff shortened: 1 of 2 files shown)


app/grandchallenge/serving/models.py
```diff
@@ -118,11 +118,24 @@ def get_component_interface_values_for_user(
         .values_list("values__pk", flat=True)
     )
 
+    evaluation_query = filter_by_permission(
+        queryset=Evaluation.objects.all(),
+        user=user,
+        codename="change_evaluation",
```
amickan, line 124: Unsure, but I think serving should only be allowed for challenge admins, not for participants. Participants should only be allowed to see the metrics that are exposed through the leaderboard (which might only be a subset in metrics.json), for example.

amickan: Then again, the evaluation detail page is accessible to participants, and we print the entire metrics.json there, and also metrics.json is not a file type socket, so not actually an output to download anyway. Then by extension the same rights should apply to additional evaluation outputs, so maybe I should change the permission back to `view`.

amickan: Thinking about whether to allow serving of evaluation inputs (might become necessary down the line if a challenge uses additional input of type file): that should definitely be restricted to admins then since the other inputs are the hidden test data.

jmsmkn: Could this be added as a comment in the code here for future developers?

app/tests/serving_tests/test_models.py
```diff
@@ -75,64 +76,83 @@ def test_get_component_interface_values_for_user():
...
     )
 
     job_with_perm = AlgorithmJobFactory(creator=user, time_limit=60)
     job_without_perm = AlgorithmJobFactory(time_limit=60)
 
-    job_with_perm.inputs.set([civ1, civ9])
-    job_without_perm.inputs.set([civ2, civ10])
-    job_with_perm.outputs.set([civ3, civ11])
-    job_without_perm.outputs.set([civ4, civ12])
+    job_with_perm.inputs.set([civa1, civb1])
```
jmsmkn, line 89: This giant test I find very hard to follow. It would nice to have explicit tests for making sure that no-one can download the inputs, and only challenge admins can download the outputs.

app/grandchallenge/serving/models.py
amickan: Sure, but what is your opinion on which permission to use for downloading the outputs?

app/tests/serving_tests/test_models.py
amickan: Why would noone be able to download the inputs though? It should be ok for challenge admins to download the inputs, right?

amickan: True, but we have additional evaluation inputs now which they cannot yet download since they do not get attached to jobs. At the moment, no one is using additional inputs of type file, but if that changes, we might need to enable downloading of inputs for challenge admins. Not relevant now, but I will add a comment about that in the code.

app/grandchallenge/serving/models.py
```diff
@@ -118,11 +118,32 @@ def get_component_interface_values_for_user(
...
+        accept_user_perms=False,
+    )
+    # We restrict downloading of evaluation outputs to challenge admins since those
+    # might not be intended for participants to see (unlike the metrics).
+    # Note: we currently do not allow direct serving of evaluation inputs.
+    # Challenge admins can already download those because they are attached to jobs
+    # they have access to as output.
+    # Serving of evaluation inputs might become necessary in the future if
+    # challenges start using additional evaluation inputs of type file. Serving of
+    # those should be restricted to admins as well.
```
jmsmkn, line 134:
```suggestion
    # Serving of evaluation inputs might become necessary in the future if
    # challenges start using additional evaluation inputs of type file. Serving of
    # those MUST be restricted to admins.
    #
    # Challenge Participants MUST NOT have access to evaluation inputs as those
    # contain the results on the hidden test set.
```

jmsmkn (approved):

PR #3990, opened 2025-05-07
Add container cleanup tasks

jmsmkn:
See https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/297

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,7 @@ jobs:
             - name: Run the django tests
               run: |
                   make minio
+                  docker network create grand-challengeorg_workstations
                   docker compose run --rm celery_worker pytest --durations 10
 
     docs:
--- a/app/grandchallenge/components/tasks.py
+++ b/app/grandchallenge/components/tasks.py
@@ -222,6 +222,29 @@ def remove_inactive_container_images():
 
 @acks_late_2xlarge_task
 @transaction.atomic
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -45,6 +45,7 @@ jobs:
             - name: Run the django tests
               run: |
                   make minio
+                  docker network create grand-challengeorg_workstations
```
jmsmkn, line 48: The latest release of docker compose only creates the network if they are attached to one of the running containers, so we need to manually create this here.

app/grandchallenge/components/tasks.py
```diff
@@ -279,13 +347,37 @@ def remove_container_image_from_registry(
...
+        ).exists()
+    elif isinstance(instance, AlgorithmImage):
+        should_be_protected = Job.objects.filter(
+            algorithm_image=instance,
+            status=Job.SUCCESS,
+        ).exists()
+    elif isinstance(instance, WorkstationImage):
+        should_be_protected = instance.created > (
+            now() - relativedelta(years=1)
+        )
```
pkcakeout, line 374: Suggestion: Would it make sense to also spare the "active image"? Maybe has been though about, I do not know...

jmsmkn: If it is over a year old then the license has expired and it will not run.

PR #3991, opened 2025-05-07
Use standard group names for challenges

jmsmkn:


amickan (approved):
Why the change now? This only applies to new challenges. Are you planning on migrating the existing ones too?

jmsmkn:
This would be for standardisation, currently there is a mismatch if the `short_name` is changed and we have requests for that.

It is anyway not possible to do this as the pk is not known before save, we would need to add a uuid, no appetite for that right now.

amickan:
Thanks for the context!

PR #3994, opened 2025-05-08
Fix WorkstationConfig causing server errors on blank input

chrisvanrun:
Sentry noted me to that the  form allows blanking of these JSON fields. 

This PR ensures (as is done in other places) that there are no server errors when this is submitted.

### DIFF
--- a/app/grandchallenge/workstation_configs/models.py
+++ b/app/grandchallenge/workstation_configs/models.py
@@ -437,15 +437,6 @@ class WorkstationConfig(TitleSlugDescriptionModel, UUIDModel):
             "workstation-configs:detail", kwargs={"slug": self.slug}
         )
 
-    def clean(self):
-        super().clean()
-        if self.overlay_segments is None:
-            self.overlay_segments = self._meta.get_field(
-                "overlay_segments"
-            ).default()
-        if self.key_bindings is None:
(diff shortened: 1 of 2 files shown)


app/grandchallenge/workstation_configs/models.py
```diff
@@ -437,6 +437,15 @@ def get_absolute_url(self):
             "workstation-configs:detail", kwargs={"slug": self.slug}
         )
 
+    def clean(self):
```
jmsmkn, line 440: I might be mistaken don't think that clean should modify values as otherwise you end up with different data in the instance and in cleaned data. These validations do not depend on any other fields so I would expect this fix to be on the form, and using the clean methods for the individual fields.

chrisvanrun: I agree — cleaning at the form on the per-field level makes more sense here, since this isn’t strictly validation but rather interpretation of submitted data. Updated!

My initial thought was inspired by ReaderStudy.view_content, which approaches this a bit differently. However, that implementation relies on a mixin that’s pretty entagled with other models. Had a quick inspiration to update that as well, but threw in the towel when I saw the mixin.

jmsmkn (approved):

PR #3997, opened 2025-05-08
Discussion forum models

amickan:


app/grandchallenge/challenges/models.py
```diff
@@ -809,27 +761,33 @@ def add_participant(self, user):
         if user != get_anonymous_user():
             user.groups.add(self.participants_group)
             follow(
```
amickan, line 763: I had to update the follows here already because otherwise all tests including the ChallengeFactory would fail.... Migration of the existing follows will be done in a separate PR.

app/grandchallenge/discussion_forums/models.py
```diff
@@ -0,0 +1,120 @@
...
+class TopicTypeChoices(models.TextChoices):
+    DEFAULT = "DEFAULT", "Default topic"
+    STICKY = "STICKY", "Sticky topic"
+    ANNOUNCE = "ANNOUNCE", "Announcement topic"
+
+
+class Forum(UUIDModel):
+    # name gets populated from challenge title, so use same max_length
+    name = models.CharField(max_length=60)
+    slug = AutoSlugField(populate_from="name", max_length=60)
```
jmsmkn, line 17: Quick thought - does the forum need a name and slug? We could limit it to one forum per challenge/parent object (one to one related) and have it accessible at `/forum/`. I don't think that we have ever seen the need for multiple forums and it is cleaner.

amickan: Good point. We used the slug previously for the urls, but since we're changing those anyway, I think I agree that the slug and name are redundant. The relationship between forum and challenge is already OneToOne and I indeed don't foresee the need for multiple forums in the future either. 

Worst case, if a name and slug became necessary at some point, it also wouldn't be too hard to add in retrospect since we're anyway auto-populating those fields from the challenge short_name.

app/grandchallenge/discussion_forums/migrations/0002_migrate_old_forums.py
```diff
@@ -0,0 +1,48 @@
...
+            new_topic = Topic.objects.create(
+                forum=new_forum,
+                creator=topic.poster,
+                subject=topic.subject,
+                type=topic_type_matching_dict[topic.type],
+                last_post_on=topic.last_post_on,
+            )
+
+            for post in topic.posts.all():
+                Post.objects.create(
```
jmsmkn, line 33: I think that the post creation time needs to be maintained.

PR #3998, opened 2025-05-08
Remove retry if dropped

jmsmkn:
With the changes with database proxying this should no longer happen


amickan (approved):
Changes look good, but could you briefly explain why/how database proxying fixes this? It's not immediately clear to me.

jmsmkn:
As the timeout times have been more carefully set, I hope!

PR #4000, opened 2025-05-09
Check for partial algorithm interface match

amickan:
`has_matching_algorithm_interfaces` should return `True` for algorithms that implement at least the interfaces defined for the phase, not only algorithms with an exact match.


chrisvanrun (approved):
LGTM!

chrisvanrun (approved):

PR #4001, opened 2025-05-09
Add job utilization models

### LINKED ISSUE #3898: `test_job_updated_start_and_complete_times_after_admin_requeue` is slow
25-30s for a single test, it's unnecessarily running another container job. Introduced in #3782

koopmant:
Add models for component job utilizations and migrate data from jobs and evaluations to utilizations.

Related to https://github.com/DIAGNijmegen/rse-roadmap/issues/408
Closes #3898


app/grandchallenge/algorithms/models.py
```diff
@@ -1334,6 +1335,23 @@ class JobGroupObjectPermission(GroupObjectPermissionBase):
...
+    job = models.OneToOneField(
+        Job,
+        related_name="job_utilization",
+        null=True,
+        on_delete=models.SET_NULL,
+    )
+
+    def save(self, *args, **kwargs) -> None:
+        if self._state.adding:
+            self.creator = self.job.creator
```
koopmant, line 1348: I haven't yet figured out how to link to `Phase` and `Challenge`. Jobs don't have a direct link to that. Would it make sense to add a field to the `Job` model linking to `Submission`?

app/grandchallenge/evaluation/models.py
```diff
@@ -2150,6 +2151,30 @@ def status_url(self) -> str:
...
+        null=True,
+        on_delete=models.SET_NULL,
+    )
+
+    def save(self, *args, **kwargs) -> None:
+        if self._state.adding:
+            self.creator = self.evaluation.submission.creator
+            self.phase = self.evaluation.submission.phase
+            if self.evaluation.submission.phase is not None:
+                # Todo: remove if-statement after data migration. Self.evaluation.submission.phase is nullable, but we don't need to expect that when we create new utilization objects
```
koopmant, line 2167: Is my assumption correct here?

app/grandchallenge/evaluation/admin.py
```diff
@@ -211,10 +212,35 @@ class EvaluationAdmin(admin.ModelAdmin):
...
+        "algorithm_image",
+        "algorithm",
+    )
+    search_fields = (
+        "creator__username",
+        "pk",
+        "evaluation__pk",
+        "phase__slug",
+        "challenge__short_name",
+    )
```
koopmant, line 241: Should I add `algorithm` and `algorithm_image` here?

koopmant:
It's not ready yet, but I could do with a review so I can continue.

app/grandchallenge/algorithms/models.py
jmsmkn: Right now we do it based on who has access to the `Job`, see https://github.com/comic/grand-challenge.org/blob/be6d2bd4e67a5c06b21f0330d3df178904d4757c/app/grandchallenge/challenges/costs.py#L49

However, the downside to this is that it will not be available pre-save, so you will need to find a different way to do it.

jmsmkn: I think that is just a case of timing and structure though, I do not think that there is a need for another relationship, that may add too much complexity.

app/grandchallenge/evaluation/admin.py
jmsmkn: The `search_fields` should be as short as possible as each one increases the duration of how long it takes to look something up. What is there now is probably more than sufficient.

app/grandchallenge/evaluation/models.py
jmsmkn: Which of those are nullable? I think all of those should exist.

The submission creator might not though!

koopmant: `phase` is nullable. I forgot my own line of though here. I see that the foreign key `phase` is set to on_protect; so I won't need to expect any submissions that would have lost their relation with submission. I believe that is what I was worried about for the data-migration. Why does it have `null=True` though?
```python
class Submission(FieldChangeMixin, UUIDModel):
    phase = models.ForeignKey(Phase, on_delete=models.PROTECT, null=True)
```

Good point on the creator. That is for when the user no longer exists, correct? 
It should not be an issue though, it would just set the value as None here, right?

jmsmkn: Indeed the phase shouldn't be null. Maybe make a separate PR removing that and if there is a reason then CI should enlighten us.

Ah indeed the creator would just be none.

jmsmkn (changes requested):
The new models need to be in a separate app before I can review this fully. This needs to be done from the start as it is really very difficult to move models between apps (forcing the use of the scary `SeparateDatabaseAndState`), and we would otherwise end up with them in the migration history.

app/grandchallenge/challenges/costs.py
```diff
@@ -181,19 +186,3 @@ def update_compute_cost_euro_millicents(
     obj.compute_cost_euro_millicents = sum(
         item["compute_cost_euro_millicents__sum"] or 0 for item in items
     )
-
-
-def update_average_algorithm_job_duration(*, phase, algorithm_jobs):
-    aggregates = algorithm_jobs.filter(
-        status=Job.SUCCESS, duration__gt=timedelta(seconds=0)
```
koopmant, line 188: Replaced this with the implementation we already had in ComponentJobUtilizationManager. The only difference is that here we used `filter(duration__gt=timedelta(seconds=0))` whereas in the manager we use `exclude(duration=None)`. Do we have instances where the duration is set, but it is zero? If so, should the implementation on the manager be updated?

app/grandchallenge/utilization/models.py
```diff
@@ -99,3 +100,83 @@ class Meta:
...
             )
         ]
+
+
+class ComponentJobUtilizationManager(models.QuerySet):
+    def average_duration(self):
+        """Calculate the average duration that completed jobs ran for"""
+        return self.exclude(duration=None).aggregate(
+            duration__avg=Avg("duration")
+        )["duration__avg"]
```
koopmant, line 110: This has been moved from the ComponentJobManager. I have deleted `total_duration`, because it was not used anywhere.

app/grandchallenge/algorithms/models.py
jmsmkn: That is correct.

app/grandchallenge/challenges/costs.py
jmsmkn: Unsure, it depends what the duration is set to with failed or cancelled jobs.

app/grandchallenge/utilization/models.py
```diff
@@ -99,3 +100,83 @@ class Meta:
...
+        null=True,
+        on_delete=models.SET_NULL,
+    )
+    external_evaluation = models.BooleanField()
+
+    def save(self, *args, **kwargs) -> None:
+        if self._state.adding:
+            self.creator = self.evaluation.submission.creator
+            self.phase = self.evaluation.submission.phase
+            self.external_evaluation = (
```
jmsmkn, line 171: Does `self.external_evaluation` exist?

app/tests/algorithms_tests/test_admin.py
```diff
@@ -1,103 +0,0 @@
...
-from tests.components_tests.factories import (
-    ComponentInterfaceFactory,
-    ComponentInterfaceValueFactory,
-)
-from tests.factories import ImageFileFactory
-from tests.utils import recurse_callbacks
-
-
-@pytest.mark.django_db
-def test_job_updated_start_and_complete_times_after_admin_requeue(
```
jmsmkn, line 25: This needs to be replaced with a test that the duration is correctly set after re-queuing.

app/grandchallenge/utilization/migrations/0003_auto_20250520_1012.py
```diff
@@ -0,0 +1,121 @@
...
+def create_job_utilizations(apps, schema_editor):
+    Job = apps.get_model("algorithms", "Job")  # noqa: N806
+    JobUtilization = apps.get_model(  # noqa: N806
+        "utilization", "JobUtilization"
+    )
+    for job in (
+        Job.objects.annotate(duration=F("completed_at") - F("started_at"))
+        .filter(job_utilization__isnull=True)
+        .iterator()
+    ):
```
jmsmkn, line 45: This will need a `.iterator(batch_size=...)` as all of the jobs will not fit into memory, and an `order_by` to ensure that it is not affected by inserts.

The algorithm needs to be selected/prefetched.

app/grandchallenge/utilization/migrations/0003_auto_20250520_1012.py
```diff
@@ -0,0 +1,121 @@
...
+            content_type__app_label="algorithms",
+            content_type__model="job",
+        )
+        for challenge in challenges:
+            jobs = Job.objects.filter(
+                jobgroupobjectpermission__group=challenge.admins_group,
+                jobgroupobjectpermission__permission=permission,
+            ).distinct()
+            for job in jobs:
+                job.update_utilization(challenge=challenge)
```
jmsmkn, line 74: Has this been tested? I do not think that you have access to custom methods here. Also check the iteration and memory requirements. It may be possible to replace this with a batch update https://docs.djangoproject.com/en/5.2/ref/models/querysets/#update

app/grandchallenge/utilization/migrations/0003_auto_20250520_1012.py
```diff
@@ -0,0 +1,121 @@
...
+    Job = apps.get_model("algorithms", "Job")  # noqa: N806
+    Phase = apps.get_model("evaluation", "Phase")  # noqa: N806
+    for phase in Phase.objects.all():
+        jobs = Job.objects.filter(
+            job_utilization__phase__isnull=False,
+            inputs__archive_items__archive__phase=phase,
+            algorithm_image__submission__phase=phase,
+        ).distinct()
+        for job in jobs:
+            job.update_utilization(
```
jmsmkn, line 87: Both comments apply here, the iteration and the custom update. Instead, it may be possible to do a batch update with https://docs.djangoproject.com/en/5.2/ref/models/querysets/#update

app/grandchallenge/utilization/migrations/0003_auto_20250520_1012.py
```diff
@@ -0,0 +1,121 @@
...
+    Evaluation = apps.get_model("evaluation", "Evaluation")  # noqa: N806
+    EvaluationUtilization = apps.get_model(  # noqa: N806
+        "utilization", "EvaluationUtilization"
+    )
+    for evaluation in (
+        Evaluation.objects.annotate(
+            duration=F("completed_at") - F("started_at")
+        )
+        .filter(evaluation_utilization__isnull=True)
+        .iterator()
```
jmsmkn, line 15: The iterator needs a batch size and an order_by is required. Several properties are being accessed so a select_related (I think, if that's not possible then prefetch_related) should be used for the submission, phase and challenge.

app/grandchallenge/components/models.py
```diff
@@ -1676,15 +1655,17 @@ class ComponentJob(FieldChangeMixin, UUIDModel):
...
-    compute_cost_euro_millicents = models.PositiveIntegerField(
-        # We store euro here as the costs were incurred at a time when
-        # the exchange rate may have been different
-        editable=False,
-        null=True,
-        default=None,
-        help_text="The total compute cost for this job in Euro Cents, including Tax",
+    started_at = deprecate_field(models.DateTimeField(null=True))
+    completed_at = deprecate_field(models.DateTimeField(null=True))
+    compute_cost_euro_millicents = deprecate_field(
```
jmsmkn, line 1660: The fields cannot be deprecated before the data migration is run otherwise the values will always return None, see https://pypi.org/project/django-deprecate-fields/

app/grandchallenge/components/models.py
```diff
@@ -1954,6 +1921,18 @@ def runtime_metrics_chart(self):
...
+        raise NotImplementedError
+
+    @property
+    def utilization(self):
+        raise NotImplementedError
+
+    def update_utilization(self, **kwargs):
+        for key, value in kwargs.items():
+            setattr(self.utilization, key, value)
+        self.utilization.save(update_fields=kwargs.keys())
```
jmsmkn, line 1934: This is too magic, style wise it is better to be explicit about what fields are going to be set here.

jmsmkn (changes requested):
Looks good but the data migration needs some work and thorough testing. Be sure to run it against a local database with lots of existing objects and make sure the values for the old and new implementations match.

app/grandchallenge/algorithms/serializers.py
```diff
@@ -152,8 +152,6 @@ class Meta:
             "inputs",
             "outputs",
             "status",
-            "started_at",
-            "completed_at",
```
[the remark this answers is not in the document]
jmsmkn: I think that would be fine, ensure that the utilisation is selected/prefetched so that we do not end up with an N+1 query.

app/grandchallenge/utilization/models.py
koopmant: It's defined as a field on this model only. I did not define it on ComponentJobUtilization because I think it does not apply to algorithm jobs. Did I make the wrong assumption there?

jmsmkn: Ah I missed it, sorry. Would be good to have these properties covered by tests.

app/grandchallenge/utilization/models.py
```diff
@@ -99,3 +100,101 @@ class Meta:
...
+    objects = ComponentJobUtilizationManager.as_manager()
+
+    class Meta:
+        abstract = True
+
+
+class JobUtilization(ComponentJobUtilization):
+    job = models.OneToOneField(
+        "algorithms.Job",
+        related_name="jobutilization",
```
jmsmkn, line 162: `jobutilization` and `evaluationutilization` is really ugly, where did the underscore go? Adding them back would be better, or even using `utilization` for both.

jmsmkn: I think that you could set:

```python3
class Meta(ComponentJobUtilization.Meta):
  default_related_name = "job_utilizations"
```

etc. in each of the concrete classes and leave them unset in the fields. See https://docs.djangoproject.com/en/5.2/ref/models/options/

app/grandchallenge/components/models.py
koopmant: Updated it. Using `object` as the default so the arguments are optional but None can still be valid value.

jmsmkn: Does this even need to be a method? It is not doing anything other than calling save, so rather than add the complexity the caller can update the object and call save itself.

app/grandchallenge/utilization/models.py
koopmant: Good point, thanks. Added some tests.

app/grandchallenge/components/models.py
```diff
@@ -1794,15 +1776,13 @@ def update_status(  # noqa: C901
...
-            and self.completed_at is None
+            and self.utilization.duration is None
+            and "duration" not in utilization_kwargs.keys()
+            and self.started_at is not None
         ):
-            self.completed_at = now()
-            if duration and self.started_at:
-                # TODO: maybe add separate timings for provisioning, executing, parsing and total
-                self.started_at = self.completed_at - duration
+            utilization_kwargs.update(duration=now() - self.started_at)
```
koopmant, line 1783: Ultimately, I want to get the duration from the executor and not calculate it here. But I have not completely figured out how to do that yet, so I'll keep it out of this PR.

app/grandchallenge/utilization/migrations/0003_auto_20250520_1012.py
koopmant: I've set the chunk_size to 2000, which seems to be a sensible default value. 
I'm still in the process of generating the data for testing the migration.

koopmant: > and an order_by to ensure that it is not affected by inserts.

I cannot say I fully understand what you mean here. Can you maybe point me to a good source to read up on this?

jmsmkn: If you're iterating over objects you want to ensure that any inserts do not affect the pagination that you're currently doing, so they should be at the end. Otherwise, new objects could be inserted at the start or somewhere in the middle, and if you are selecting by pages (give me results 1000 to 1999) then you could miss some.

app/grandchallenge/components/models.py
koopmant: I made it while trying to minimize changes made to `ComponentJob.update_status()`. That method is often called without the duration and compute_cost arguments. So I ended up with this logic: I'm calling save with the update_fields argument, so if that argument is empty, save is not actually called and we save a database hit. Am I correct there?

In the end, I'm not sure if utilization updates should go through the update_status method at all. I think I would prefer it if they don't. But since the PR is already a bit much, I did not want to tackle that yet.

app/tests/algorithms_tests/test_admin.py
koopmant: I've replaced it with a simple test. Unlike the previous test I'm not capturing callbacks, since I only wanted to check if duration is set to None.

app/grandchallenge/components/models.py
jmsmkn: So why not just do it in `update_status`, and only save if the values are set? It would then follow the same logic as already implemented on main, fundamentally nothing should be changing logic wise, I don't see why you would need to set the utlisation back to `None`.

koopmant: > I don't see why you would need to set the utlisation back to None.

I'm assuming you mean `utilization.duration` here. But I think you're right, you wouldn't. Not unless you're explicitly retrying the job, in which case we set the duration to None explicitly anyway.

> So why not just do it in update_status, and only save if the values are set? 

It just felt cleaner like this, no other reason. I've just removed the method and put it in update_status.

app/grandchallenge/utilization/migrations/0003_auto_20250520_1012.py
jmsmkn: You should be selecting/prefectching things and leaving it unset in combination with prefetch is deprecated from Django 5, so just good to get in the habit https://docs.djangoproject.com/en/dev/internals/deprecation/#deprecation-removed-in-5-0

koopmant:
This is ready for review again. I've tested the data migration and it seems to work. I've tested it with ~16000 jobs in 500 challenges, which was no problem. I'm not sure how realistic the state of my database was though. I did not have any jobs that were created for an evaluation, so no jobs were related to a challenge, phase, or archive. I tested that part after creating a few manually.

app/grandchallenge/utilization/migrations/0003_auto_20250520_1012.py
koopmant: Replaced with a batch update

app/grandchallenge/utilization/migrations/0003_auto_20250520_1012.py
koopmant: This should be fixed now.

app/tests/algorithms_tests/test_admin.py
jmsmkn: Indeed, I linked this PR to an issue about that test as it was poorly implemented.

koopmant:
I've pushed another change to include `submission__phase__challenge` and `submission__creator` in select_related for the Evaluation query in the data migration, and `algorithm_image__algorithm` and `creator` in select_related for the Job query. 
I added these because when I inspect the database queries locally, these fields required additional separate queries. However, I don't really understand why they are needed. Can I assume that it will be the same in production?

app/grandchallenge/evaluation/tasks.py
```diff
@@ -271,6 +272,13 @@ def create_algorithm_jobs_for_evaluation(*, evaluation_pk, max_jobs=1):
         requires_memory_gb=evaluation.submission.algorithm_requires_memory_gb,
     )
 
+    job_utilizations = JobUtilization.objects.filter(job__in=jobs)
+    job_utilizations.update(
+        phase=evaluation.submission.phase,
+        archive=evaluation.submission.phase.archive,
+        challenge=evaluation.submission.phase.challenge,
+    )
```
jmsmkn: `create_algorithm_jobs` can raise `TooManyJobsScheduled`, in which case this metadata will not get added.

PR #4002, opened 2025-05-09
Add (phase) algorithm interface copy form

chrisvanrun:
This PR adds a form that allows copying the algorithm interfaces configured for one phase to another.

Generally there are at least 2 Phases that require te same setup (sometimes more). Generating these interfaces is tidious and also error prone. Moreover, the complexer the interfaces the more likely the initial setup was missing something and it has to be re-done, sometimes multiple times.

Adding an interface edit form would be an option, but Anne warned that might be complex and adding this (simple) copy form already cuts the work in half.

## Showcase

https://github.com/user-attachments/assets/cf27a1a6-391e-406a-a9d4-30d0e3283727

### DIFF
--- a/app/grandchallenge/evaluation/forms.py
+++ b/app/grandchallenge/evaluation/forms.py
@@ -985,3 +985,29 @@ class EvaluationGroundTruthVersionManagementForm(Form):
             raise ValidationError("Ground truth updating already in progress.")
 
         return ground_truth
+
+
+class AlgorithmInterfaceForPhaseCopyForm(Form):
+    phases = ModelMultipleChoiceField(
+        queryset=Phase.objects.none(),
+        label="Select the phases to copy the interfaces to",
+        widget=CheckboxSelectMultiple,
+    )
+
+    def __init__(self, *args, user, phase, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._phase = phase
+
+        self.fields["phases"].queryset = filter_by_permission(
+            queryset=Phase.objects.filter(
+                challenge=phase.challenge,
+                submission_kind=phase.submission_kind,
+            ).exclude(pk=phase.pk),
+            codename="evaluation.configure_algorithm_phase",
+            user=user,
+        )
+
+    def copy_algorithm_interfaces(self):
+        for phase in self.cleaned_data["phases"]:
+            for interface in self._phase.algorithm_interfaces.all():
+                phase.algorithm_interface_manager.add(interface)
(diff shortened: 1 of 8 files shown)


app/grandchallenge/evaluation/forms.py
```diff
@@ -985,3 +985,25 @@ def clean_ground_truth(self):
...
+        widget=CheckboxSelectMultiple,
+    )
+
+    def __init__(self, *args, phase, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._phase = phase
+
+        self.fields["phases"].queryset = Phase.objects.filter(
+            challenge=phase.challenge,
+            submission_kind=phase.submission_kind,
```
amickan, line 1003: Since we only allow interfaces for algorithm submission phases, you can subset by that directly, right?

app/grandchallenge/evaluation/views/__init__.py
```diff
@@ -1384,6 +1385,46 @@ def get_context_data(self, *args, **kwargs):
         return context
 
 
+class AlgorithmInterfacesForPhaseCopy(
+    ConfigureAlgorithmPhasesPermissionMixin,
+    AlgorithmInterfaceForPhaseMixin,
+    SuccessMessageMixin,
+    FormView,
+):
```
amickan, line 1393: I think we might need to look this down further to make sure that we're not accidentally updating locked phases (i.e. those that are in a parent-child relationship). We also lock the other interface views for those phases. There is a property on the phase model you can use for that `algorithm_interfaces_locked`, see also `LockParentAndChildPhasesMixin`

amickan: The subsetting would best be done in the form, not here, of course.

chrisvanrun: Good idea. I've made the changes.

app/grandchallenge/evaluation/forms.py
```diff
@@ -985,3 +985,37 @@ def clean_ground_truth(self):
...
+
+    def clean_phases(self):
+        phases = self.cleaned_data["phases"]
+
+        for phase in phases:
+            if phase.algorithm_interfaces_locked:
+                raise ValidationError(
+                    "You cannot copy algorithm interfaces to a phase "
+                    "where the algorithm interfaces are locked."
+                )
```
amickan, line 1014: It would probably help here to say which phase(s) is locked so that you can avoid selecting that phase when trying again.

amickan (dismissed):
LGTM! Just one small suggestion for the form error reporting.

app/grandchallenge/evaluation/forms.py
chrisvanrun: Done!

amickan (approved):

chrisvanrun:
Thanks for the review Anne!

PR #4003, opened 2025-05-09
Topic create, detail, list and delete views

amickan:
First set of forum views: `TopicCreate`, `TopicDelete`, `TopicList` and `TopicDetail`. 

These views also include links to subscribe and unsubscribe since those views already exist. They don't work fully yet, but I will tackle subscriptions in a later PR. 


amickan:
Wrong base branch

PR #4004, opened 2025-05-09
Topic create, detail, delete and list views

amickan:
First set of forum views: `TopicCreate`, `TopicDelete`, `TopicList` and `TopicDetail`. 

These views also include links to subscribe and unsubscribe since those views already exist. They don't work fully yet, but I will tackle subscriptions in a later PR. 

Forum home: 
image

Topic detail:
image

Topic delete:
image


app/grandchallenge/profiles/models.py
```diff
@@ -104,7 +104,7 @@ def get_absolute_url(self):
 
     def get_mugshot_url(self):
         if self.mugshot:
-            return self.mugshot.x02.url
+            return self.mugshot.x20.url
```
amickan, line 107: Is this intentional or a typo? Everywhere else we seem to be referencing `x20`. `x02` returned a blurry image.

app/grandchallenge/notifications/templates/actstream/partials/follow_unfollow_links.html
```diff
@@ -0,0 +1,28 @@
...
+{% else %}
+    {% get_content_type object as object_ct %}
+    {% follow_form user=request.user content_type=object_ct object_id=object.id as follow_form %}
+    <form class="d-inline" method="post" action="{% url 'notifications:follow-create' %}">
+        {% csrf_token %}
+        {{ follow_form|crispy }}
+        <button type="submit" class="btn btn-primary btn-sm btn-subscription"><i class="fas fa-check mr-1"></i>Subscribe
+            to {{ object|verbose_name }}
+        </button>
+    </form>
```
amickan, line 27: It would be nice if we could just do `{% crispy follow_form %}` here. That would require setting all attributes in the template tag. I almost got it to work except that the form then renders on a separate row with a huge lower margin and I couldn't change that (I tried with all kinds of form helper attributes), so I gave up and think this is what it has to be.

app/grandchallenge/discussion_forums/urls.py
```diff
@@ -0,0 +1,30 @@
...
+
+urlpatterns = [
+    path("", TopicListView.as_view(), name="topic-list"),
+    path(
+        "topics/create/",
+        TopicCreate.as_view(),
+        name="topic-create",
+    ),
+    path(
+        "topics/<uuid:pk>/",
```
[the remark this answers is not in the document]
amickan: Ok. The same should also go for the `post` views that I still need to create. Would a maximum of 100 for the slug field be ok?

app/grandchallenge/profiles/models.py
amickan: Clear. I'll change the forum implementation then.

app/grandchallenge/challenges/models.py
```diff
@@ -582,7 +582,7 @@ def assign_permissions(self):
         self.assign_forum_permissions()
 
     def assign_forum_permissions(self):
-        if self.display_forum_link and not self.hidden:
+        if self.display_forum_link:
```
amickan, line 585: This was approved in the previous PR, but I'm actually wondering if it makes sense now. Shouldn't hidden challenges be able to have a forum for their users as well? 

I misunderstood the current permission assignment for forums. Currently, hidden challenges are allowed to have forums, and both participants and admins can access those forums. 

Assuming that we want to keep it that way, this removes the self.hidden check.

app/grandchallenge/discussion_forums/views.py
```diff
@@ -0,0 +1,92 @@
...
+        context = super().get_context_data()
+        context.update({"forum": self.object.forum})
+        return context
+
+
+class TopicDelete(ObjectPermissionRequiredMixin, DeleteView):
+    model = Topic
+    permission_required = "discussion_forums.delete_topic"
+    raise_exception = True
+    success_message = "Successfully deleted topic."
```
amickan, line 81: I again took the current implementation as guidance. For the `TopicDelete` view this means that we don't block deletion of topics with posts and that deleting a topic also deletes all of its posts. This is also how the `ForeignKey` `on_delete` is currently set-up on the DB level. The permission to do this is reserved to challenge admins. 

My initial thought here as the the creator should also get the right to delete their own topics, but that would essentially give them the permission to delete other users posts to their topic, so I removed the permission from the creator.

app/grandchallenge/discussion_forums/urls.py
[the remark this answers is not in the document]
jmsmkn: 64 is probably plenty.

app/grandchallenge/challenges/models.py
jmsmkn: Yes, hidden challenges should have forums but they should not be visible to non-participants. I guess the same goes for public challenges?

app/grandchallenge/discussion_forums/views.py
jmsmkn: I think that users should be able to delete their own posts. If there are then no posts for a topic then it should also be deleted.

app/grandchallenge/challenges/models.py
amickan: Yes ,at the moment only admins and participants of a challenge can view the challenge's forum, Non-participants and anonymous users cannot. I think that makes sense. 

At the moment, I believe that public challenges also have a public forum, but I actually think that is a bit weird. Locking it down to participants makes more sense to me.

app/grandchallenge/discussion_forums/views.py
amickan: Yes, posts can be deleted by the creator and if the last post of a topic is deleted, so is the topic. But what I meant here is whether the creator of a topic should be able to delete the topic. Currently, that is not possbile, and I've kept it that way since that would allow them to delete other user's posts.

app/grandchallenge/discussion_forums/templates/discussion_forums/forum_base.html
```diff
@@ -0,0 +1,38 @@
...
+{% block content %}
+    <div class="row mb-3 justify-content-between">
+        <div class="col-8"><h1>Forum</h1></div>
+        {% get_obj_perms request.user for forum as "user_perms" %}
+        {% if 'view_forum' in user_perms %}
+            <div class="col-4">
+                <div class="float-right controls-link-wrapper">
+                    {# TODO: my posts view button #}
+                    {% include 'actstream/partials/follow_unfollow_links.html' with object=forum %}
+                    <a href="{% url 'notifications:follow-list' %}" class="btn btn-dark btn-sm"><i class="fas fa-bookmark">&nbsp;</i>My Subscriptions</a>
```
jmsmkn, line 29: I think there are a few inconsistent uses of non-breaking spaces to separate icons and text. Here there is one within the `<i>` tag, others have one or two outside the `<i>` tag, others do not have any. We should try and be consistent about how these are used.

app/grandchallenge/discussion_forums/templates/discussion_forums/topic_confirm_delete.html
```diff
@@ -0,0 +1,38 @@
...
+        </div>
+        <div class="card-body">
+            <div class="row px-3">
+                <form action="" method="post">
+                    {% csrf_token %}
+                    <p>Are you sure that you want to delete the topic "{{ object }}"?</p>
+                    <p>
+                        <b class="text-danger">WARNING: You are not able to undo this action. Once the topic is deleted,
+                            it is
+                            deleted forever.</b>
```
jmsmkn, line 25: Quite the poetic layout 😅

app/grandchallenge/discussion_forums/models.py
```diff
@@ -78,6 +80,69 @@ class Meta:
     def __str__(self):
         return self.subject
 
+    def save(self, *args, **kwargs):
```
jmsmkn, line 83: You may want the `slug` and `parent_object` to be unique together, that way the slugs do not need to be globally unique, but unique per challenge. If you do that, you will need an extra filter on the parent object anywhere an object is looked-up with a slug in the url, so far this would be the detail and delete views.

app/grandchallenge/discussion_forums/models.py
```diff
@@ -45,7 +47,7 @@ class Topic(UUIDModel):
     )
 
     subject = models.CharField(max_length=255)
-    slug = AutoSlugField(populate_from="subject", max_length=255)
+    slug = AutoSlugField(populate_from="subject", max_length=64)
 
     TopicTypeChoices = TopicTypeChoices
     type = models.CharField(
```
jmsmkn, line 53: We shouldn't use `type` as a property as it clashes with the built-in. I think we use `kind` elsewhere to avoid this. Could you also add the database constraint on the choices?

With the boolean `locked` I think that using `is_locked` makes the code easier to read.

jmsmkn (commented):
Looks good!

app/grandchallenge/discussion_forums/models.py
amickan: Good point. Not forcing global uniqueness makes sense here. I'll make the necessary changes.

app/grandchallenge/discussion_forums/templates/discussion_forums/forum_base.html
amickan: I completely agree. I had started removing the non-breaking spaces, but clearly missed a few.

app/grandchallenge/discussion_forums/models.py
amickan: Thinking about this some more: extending this to reader studies or other parent objects can be done with the current design. The uniqueness constraint would then just need a condition added to it, and I think we could also add a check constraint to ensure that there is only either a linked_challenge or linked_reader_study, for example. It's less elegant, but possible, and in my opinion better than the GFK for performance and ease of working with the parent object relation.

amickan: That all has me thinking that it would be good to add a constraint now that each forum MUST have a linked challenge. We don't want dangling forums, right?

amickan: @jmsmkn could you comment on the latter (making a parent_object required for forums)?

amickan: Turns out that I also can't add a uniqueness constraint for the related field `linked_challenge`, only fields on the model can be included in unique constraints. But then again, I think we don't need to enforce uniqueness with the parent object, but rather between slug and forum. That makes a lot of the above irrelevant.

amickan: I have not added the constraint that a forum needs to have a linked challenge for now. If we want to add it, it needs to happen in a separate migration after both the forum model and the new field on the challenge model have been created. I can add it later, if we think it's important.

amickan: I think there is no need to add an extra filter on any of the views, since the views already include the challenge short_name filter, and a challenge only has one forum, so that filters it down sufficiently already, right?

jmsmkn: You're right that it should be unique `slug` and `forum`, sorry!

> since the views already include the challenge short_name filter, and a challenge only has one forum

`TopicDetail` and `TopicDelete` does not appear to me to be filtered by forum. The only URL lookup there is the slug. The challenge is not considered by the default object resolver.

jmsmkn: Should the post have a title/slug? It is always viewed in context of an existing topic.

amickan: A post can be updated though, so we need a way to refer to a single post. If there is no slug, the url for the update view would then contain the pk. If that's ok, then we could indeed do without a title and slug.

jmsmkn: Yes, I think that is fine.

amickan: Looking at current post slugs, they often start with one (or multiple) "re:" (which gets prepended automatically), because it indeed doesn't really make sense to add a new title to a reply post. So yeah, we're not losing meaningful information by getting rid of the slug.

app/grandchallenge/discussion_forums/models.py
```diff
@@ -120,3 +191,47 @@ def delete(self, *args, **kwargs):
...
+    group = models.ForeignKey(
+        "auth.Group",
+        on_delete=models.CASCADE,
+        related_name="forums_postgroupobjectpermissions",
+    )
+    permission = models.ForeignKey(
+        "auth.Permission",
+        on_delete=models.CASCADE,
+        related_name="forums_postgroupobjectpermissions",
+    )
```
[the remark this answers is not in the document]
amickan: Because there is name clash with the Blog's Post model group and user object permissions otherwise. I had to change the related_name of the `group` and `permission` FKs to get it to work.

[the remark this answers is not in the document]
amickan: I think I also need to use your new implementation of the `GroupObjectPermissionBase`.

app/grandchallenge/discussion_forums/forms.py
```diff
@@ -0,0 +1,59 @@
...
+        model = Topic
+        fields = ("forum", "creator", "subject", "kind", "content")
+
+    def __init__(self, *args, forum, user, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.fields["forum"].queryset = Forum.objects.filter(id=forum.id)
+        self.fields["forum"].initial = forum
+
+        self.fields["creator"].initial = user
```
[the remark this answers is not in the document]
amickan: You're absolutely right. Thanks for catching that. A quick search suggests that everywhere else we do this, we also subset the queryset to only the passed in user, so we're good. I will fix this here.

app/grandchallenge/discussion_forums/models.py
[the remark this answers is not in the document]
amickan: Sure. I will also rename the `Topic` model to `ForumTopic` then. No name clash there, but for consistency with the `ForumPost` model.

amickan:
Not sure if you want to take another look, but it is now ready for re-review.

jmsmkn (approved):
LGTM

PR #4005, opened 2025-05-12
Fix container protection and image deletion

jmsmkn:
If container images fail imports then they should not be protected. Sometimes the image will not be set at all so it can immediately be marked as removed.

Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/514


amickan (approved):
Makes sense!

PR #4008, opened 2025-05-13
Make phase fk on submission not nullable

koopmant:
The phase shouldn't be null for submissions. 

See also https://github.com/comic/grand-challenge.org/pull/4001#discussion_r2084850395


koopmant:
@jmsmkn We need to manually check if there are existing rows with null values I think?

jmsmkn (approved):
LGTM

PR #4013, opened 2025-05-13
Remove use of `get_objects_for_user` and `get_objects_for_group`

jmsmkn:
We have an optimised version implemented so should always use it.


amickan (approved):
LGTM

PR #4022, opened 2025-05-14
Add cancelled payment status option for invoices

### LINKED ISSUE #4021: Add cancelled payment status option for invoices
https://github.com/DIAGNijmegen/rse-challenge-request-admin/issues/27#issuecomment-2879921060
> The idea of the invoices model is that we will eventually be able to replace to excel sheet and have one source of truth. So if invoices are marked as cancelled in excel, they should be marked as cancelled in our database.

koopmant:
Closes #4021


jmsmkn (approved):
LGTM

PR #4024, opened 2025-05-14
Allow combination of `ObjectPermissionRequiredMixin` and `ViewObjectPermissionListMixin`

jmsmkn:
Allows for checking a parent object for object permissions (e.g. change reader study) and filtering the list by view permissions (e.g. view display sets). The list permissions are now always view, there was one exception to this in the codebase which is now handled by the additional parent object permissions.


amickan (approved):
LGTM

PR #4026, opened 2025-05-15
Avoid use of union in permission filter

jmsmkn:
This PR allows using the most optimal filter in all cases. By defining which permissions are allowed on the user and group models we can avoid having to search over both tables. Now we should be able to see the cases where this is necessary - currently these are "change_job", "view_image", "view_challengerequest", "view_submission", "view_answer", "change_session", and "view_session".

The allowed_permissions have been populated by inspecting the production database. If they are any extra ones then this only affects performance. Errors will be raised if permissions are attempted to be assigned that are not allowed.


amickan (approved):
LGTM! I didn't actually go through the codebase to double check that the allowed permissions are complete, but I assume you did that carefully. I just checked for a few examples, also to double check I understood this correctly.

jmsmkn:
Yep, generated them automatically and a couple that were missing as they were not assigned to anyone in prod were found by the tests, will anyway keep an eye on sentry.

PR #4029, opened 2025-05-15
Forum post views

amickan:


app/grandchallenge/discussion_forums/models.py
```diff
@@ -176,44 +176,115 @@ class Meta:
...
-        else:
-            super().delete(*args, **kwargs)
+        elif self.is_last_post:
+            update_last_post_on = True
+            topic = self.topic
+
+        super().delete(*args, **kwargs)
 
+        if update_last_post_on:
+            topic.last_post_on = topic.last_post.created
```
jmsmkn, line 211: Maybe it makes sense that a query is done here, or a separate task launched to update this? I imagine that in most cases `last_post` can be cached (or it would seem to a future developer an easy thing to cache).

app/grandchallenge/discussion_forums/models.py
```diff
@@ -176,44 +176,115 @@ class Meta:
...
+            topic.save()
 
-class ForumUserObjectPermission(NoUserPermissionsAllowed):
+    def assign_permissions(self):
+        # challenge admins and participants can see this post
+        assign_perm(
+            "discussion_forums.view_forumpost",
```
jmsmkn, line 217: Maybe we should prefer using just the codename (e.g. "view_forumpost", without the app name) everywhere? That would make our codebase more searchable and I do not think that anything is lost by removing the app name. 

Same applies for the permissions in the views and the change to be added to our style guide.

app/grandchallenge/discussion_forums/templates/discussion_forums/forumpost_confirm_delete.html
```diff
@@ -0,0 +1,51 @@
...
+                    Confirm Post Deletion
+                </h5>
+                <button type="button" class="close" data-dismiss="modal"
+                        aria-label="Close">
+                    <span aria-hidden="true">&times;</span>
+                </button>
+            </div>
+            <dd class="text-left modal-body">
+                <div class="row px-3">
+                    <form action="{% url 'discussion-forums:post-delete' slug=object.topic.slug pk=object.pk %}"
```
jmsmkn, line 26: Did we decide to serve these on the subdomain? If so, I think that you're missing `{% load url %}`. Same for the other views.

app/grandchallenge/discussion_forums/views.py
```diff
@@ -1,20 +1,30 @@
...
     ForumTopic,
     ForumTopicKindChoices,
 )
 from grandchallenge.subdomains.utils import reverse
 
 
-class ForumTopicListView(PermissionListMixin, ListView):
+class ForumTopicListView(ViewObjectPermissionListMixin, ListView):
     model = ForumTopic
     permission_required = "discussion_forums.view_forumtopic"
```
jmsmkn, line 29: `permission_required` is not used by `ViewObjectPermissionListMixin`

app/grandchallenge/discussion_forums/views.py
```diff
@@ -108,3 +118,116 @@ def get_success_url(self):
...
+    raise_exception = True
+
+    @cached_property
+    def forum(self):
+        return self.request.challenge.discussion_forum
+
+    def get_object(self, queryset=None):
+        return get_object_or_404(
+            ForumPost,
+            pk=self.kwargs["pk"],
```
jmsmkn, line 165: It may not be an issue but for these three views only filtering by pk and not forum will allow the views in the context of another challenge or topic. You may want to add the `topic__forum=self.forum` filter too.

app/grandchallenge/discussion_forums/views.py
```diff
@@ -108,3 +118,116 @@ def get_success_url(self):
...
+    permission_required = "discussion_forums.delete_forumpost"
+    raise_exception = True
+    success_message = "Successfully deleted post."
+
+    @cached_property
+    def forum(self):
+        return self.request.challenge.discussion_forum
+
+    @cached_property
+    def topic(self):
```
jmsmkn, line 180: The topic should be obtained from the object, not the url, this is to prevent incorrect references. Same goes for `ForumPostUpdate`.

app/grandchallenge/discussion_forums/forms.py
```diff
@@ -63,3 +65,73 @@ def save(self, commit=True):
...
+
+        self.fields["topic"].queryset = ForumTopic.objects.filter(id=topic.id)
+        self.fields["topic"].initial = topic
+
+        self.fields["creator"].queryset = get_user_model().objects.filter(
+            pk=user.pk
+        )
+        self.fields["creator"].initial = user
+
+        if not self.instance._state.adding:
```
jmsmkn, line 101: Accessing the `_state` is a little risky. Can this be solved with a form kwarg?

app/grandchallenge/discussion_forums/forms.py
```diff
@@ -63,3 +65,73 @@ def save(self, commit=True):
...
+                "hx-post": hx_post_url,
+                "hx-target": hx_target,
+                "hx-swap": "outerHTML",
+            }
+        )
+
+    def clean(self):
+        if (
+            self._topic.is_locked
+            and not self._topic.forum.parent_object.is_admin(self._user)
```
jmsmkn, line 132: Both topic and user (creator) should be obtained from `cleaned_data`.

jmsmkn (commented):
Looks good! Just a few style things.

app/grandchallenge/discussion_forums/templates/discussion_forums/forumpost_confirm_delete.html
jmsmkn: I honestly have no idea how that works then, didn't this cause problems for other htmx views? Is it because these are only partials and are not full views? If so, shouldn't they be in a partials subdirectory?

amickan: I also don't really understand to be honest. For the forumpost_form template it doesn't matter if I add {% load urk %} or not - works either way, but for forumpost_confirm_delete and forumpost_detail it actually breaks if I load the url template tag. 

I can move them to the partials folder, but I hadn't done that since the views expect the templates in their current location. I can of course override the template on the view, if you think it makes things clearer if they are in the partials folder.

amickan: What the delete and the detail templates have in common is that both are included into their parent template (not loaded via HTMX, as I claimed before, that is only true for the create and update views, for which it doesn't matter if I load the custom template tag or not). So this is unrelated to HTMX, and instead must have something to do with how include works in Django templates. Looks like the request context is not shared between the parent and the partial template since the challenge_short_name is no longer in the context for the partial (hence causing the resolve error when I load the template tag in the partial template).

app/grandchallenge/discussion_forums/models.py
amickan: Good point, I will turn this into a query.

app/grandchallenge/discussion_forums/templates/discussion_forums/forumpost_confirm_delete.html
amickan: Ok, so included templates share the context and template tags loaded in the parent template, which is why there is no need to load the url tag again in the partial. I still do not understand why loading it again is an issue (the docs don't help and ChatGPT also not), but I have moved all templates to the partial folder now to make it clear that they are partials, and that hopefully also explains why loading the tag is not necessary. @jmsmkn does this make sense to you?

jmsmkn: It now seems that they're used as both partials (through include tags) and full views (via the CBVs in views.py). Does that make sense? I would have thought that if they are HTMX views they would only be used as that, so they would be full views and then rather than using an include using a div that is replaced by htmx in a separate call.

amickan: The only HTMX views are the `PostCreate` and `PostUpdate` views. These views load partials into the DOM via HTMX, they are not included anywhere via the {% include %} tag. Since the templates are only partial templates (they don't inherit from the base template, it's just a div basically), I think it makes sense for these templates to live in the partials subfolder. I used the CBV for these 2 views since functionally they are still classic update and create views. Are you suggesting to change those to `FormViews`?

amickan: The `PostDelete` and `PostDetail` views are included into the `TopicDetail` templates via `{% include %}`. Again I used the CBVs for those since the functionality still is that of classic detail and delete views. We just don't render them like normal views. 

I guess for the detail view you could argue that maybe that should be a `TemplateView`, but for the DeleteView I wouldn't know how else to implement it.

jmsmkn: Partials are things that only get included in other templates via include tags. If a template is used as the base of a view, then it is a full view, regardless of whether it uses the base template. A full view should not be included in other templates. We have to be clear about how templates are used and be strict about them.

Following that:

- discussion_forums/partials/forumpost_form.html is not a partial as it is used in ForumPostCreate and ForumPostUpdate, it does not use any urls so doesn't need the tag. That template needs to move.
- discussion_forums/partials/forumpost_detail.html is not a partial as it is used by ForumPostDetail. However, is also used as an include. It should be one or the other. If it is going to be a full view then that needs to be called from htmx from the other view as a full view. As it is a full view and has urls then it needs the url tags. If it is going to be a partial then the full post detail view needs to be removed.
- discussion_forums/partials/forumpost_confirm_delete.html see discussion_forums/partials/forumpost_detail.html, but this does not have the issues with urls.

jmsmkn: This has nothing to do with the type of views (detail, create, update, delete), but the organisation and the design of the implementation with htmx.

amickan: Thanks, James. Makes sense. I will move the templates back to the main folder and call them through HTMX. However, adding the url tags still does not work. I get a resolver error then since the request object does not contain the challenge. I really don't know why that is. Might be a peculiarity of HTMX.

jmsmkn: That needs to be traced, HTMX shouldn't have anything to do with it as this is being set purely in the Django world depending on what domain the request is on. If it is not on a challenge domain then the view shouldn't be resolved at all as the urls are not included there. Check that the request from the browser is going to the correct domain and that it is being added by the challenge middleware.

amickan: Ok, so the issue was a missing kwarg for `challenge_short_name` on the two templates that get called through HTMX. If I add that to the urls, then loading the url tag works. 

I see that often we don't add that kwarg. I was trying to see if it gets added automatically in some cases, but didn't find anything. How does this work normally?

jmsmkn: The kwarg is only needed for views on subdomains, so the resolver knows to use the subdomain urlconf. If the view is on the main domain then the kwarg is not needed.

amickan: I don't understand. For example, in the TopicDetail view I can link to `{% url 'discussion-forums:topic-list' %}` without needing to provide the `challenge_short_name`. This view is on the subdomain, so according to what you just said this should not work and the kwarg should be needed here.

jmsmkn: Please link to that location.

amickan: See for example here: https://github.com/comic/grand-challenge.org/blob/9f43cca15b9bf52d82b8e485f01081c004bb9eb8/app/grandchallenge/discussion_forums/templates/discussion_forums/forumtopic_detail.html#L16

app/grandchallenge/discussion_forums/templates/discussion_forums/forumtopic_detail.html
```diff
@@ -26,41 +23,27 @@
...
-                            </div>
-                            <div class="col-md-2 d-none d-md-block">
-                                {% include 'profiles/partials/mugshot.html' with profile=post.creator.user_profile with_userlink=True %}
-                            </div>
-                        </div>
-                    </div>
-                </div>
+                <div class="my-3 card col-12"
+                     id="post-{{ post.pk }}"
+                     hx-get="{% url 'discussion-forums:post-detail' slug=object.slug pk=post.pk %}"
```
jmsmkn, line 45: This is going to create one request per post, similar to an N+1 problem, it would be better if the user does not have to go back to the server each time. Why is each post its own individual view? Do we really need that or can everything be done as part of the topic detail view? If you then need to go to a particular post you could use anchors. Shouldn't the posts also be paginated?

amickan: The only reason I added the detail view was to be able to easily return the newly created post and swap that into the DOM after post creation, but I can also just return the entire topic detail view again and make sure the entire body is replaced. Then, we can just include each post via a partial and remove the detail view.

amickan: Maybe it is easiest to turn the `TopicDetail` into a `TopicPostList` view to enable easy pagination.

jmsmkn: If you use the PaginatedListView the table of topics will update asynchronously without a page refesh as the user goes through the pages. The announcements can be a separate block at the top which is static, would need a separate queryset to get those objects.

amickan: Ah yeah ok, that might work. it does require a rework of the templates though since currently it's not a table at all. I'll do that in a seperate PR. Easier to get this merged first.

amickan: I have changed the `TopicDetail` to a list view and added pagination. I will make the changes to the `TopicList` in a seperate PR.

jmsmkn: True, maybe a normal list view with pagination and an anchor link to the top of the normal topic from page 2 onwards would fit better, otherwise you have a table header and everything.

jmsmkn (approved):
Looks much better! Some suggestions for permissions and url style, up to you.

app/grandchallenge/discussion_forums/views.py
```diff
@@ -1,22 +1,25 @@
...
 from grandchallenge.discussion_forums.models import (
+    ForumPost,
     ForumTopic,
     ForumTopicKindChoices,
 )
 from grandchallenge.subdomains.utils import reverse
 
 
-class ForumTopicListView(PermissionListMixin, ListView):
+class ForumTopicListView(ViewObjectPermissionListMixin, ListView):
```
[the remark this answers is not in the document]
amickan: Ok, but the permission would then need to be `view_forum` since the object we'd check would be the `forum`, or did I misunderstand the idea here?

app/grandchallenge/discussion_forums/urls.py
```diff
@@ -21,26 +20,21 @@
...
-        ForumTopicDetail.as_view(),
-        name="topic-detail",
-    ),
     path(
         "topics/<slug:slug>/delete/",
         ForumTopicDelete.as_view(),
         name="topic-delete",
     ),
+    path(
+        "topics/<slug:slug>/posts/",
```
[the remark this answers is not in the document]
amickan: Would you still call it `topic-post-list`, or is `topic-detail` the better url name?

app/grandchallenge/discussion_forums/views.py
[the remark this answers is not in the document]
jmsmkn: Why? The `view_forumtopic` is being assigned to the correct groups, and you're looking at a `ForumTopic` here.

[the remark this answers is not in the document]
amickan: I think this is actually nice, since it would raise a 403 for non-participants (which so far just got an empty page).

app/grandchallenge/discussion_forums/views.py
```diff
@@ -108,3 +120,116 @@ def get_success_url(self):
...
+
+    def get_form_kwargs(self):
+        kwargs = super().get_form_kwargs()
+        kwargs.update(
+            {"user": self.request.user, "topic": self.topic, "instance": None}
+        )
+        return kwargs
+
+    def get_success_url(self):
+        return f"{self.topic.get_absolute_url()}#post-{self.object.pk}"
```
jmsmkn, line 157: You may need to go to the last page here to ensure that the post is on the page. Same in `ForumPostUpdate`: this could all go in `get_absolute_url` for the post, no need to duplicate it.

app/grandchallenge/discussion_forums/urls.py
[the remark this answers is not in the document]
amickan: `topic-post-list` doesn't really make sense if the path is `topics/<slug:slug>`, so I will change it back to `topic-detail`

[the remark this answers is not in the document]
jmsmkn: Your call, in a way it's both. Maybe stick with post list as that is how it is implemented.

app/grandchallenge/discussion_forums/views.py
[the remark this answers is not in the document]
amickan: I'm looking at a list of forum topic (for which the view permission is checked by `ViewObjectPermissionListMixin`. The one object to check object permissions for besides all the topics listed would be the forum.

app/grandchallenge/discussion_forums/views.py
amickan: Ah right, with the pagination added, the page also neeeds to be taken into account when getting and displaying the form. At the moment I just append to the bottom of the post list, but that may not be the end of the list. I think I might need to not use HTMX for the create and update views after all. Looking into it now.

amickan: If I switch to just a regular view that only shows the post form, the downside is that the user no longer sees the messages they're responding to. This was the reason why I went with HTMX in the first place. I guess the best is then to mimic the post list view and have the form view show the last page of posts for the topic with the form at the bottom.

amickan: Mimicking the post list on the post create page is messy and requires quite a bit of logic to work properly (with the paginator especially), so I don't think that's the way to go. Is it an ok assumption to say that people will already be on the last page when replying, and hence just do what I do now and append the form to the bottom of the page. The redirect on save would still load the appropriate page and the new post would still appear as the last post. Having had a look through GC I have yet to find a topic with more than 10 posts, so we might never hit the pagination for the post list anyway.

amickan: Especially if we also implement the last read functionality discussed on the pitch, a topic view would always open on the last page for the user. So I think it is fine to just append the form to the end of the current page.

jmsmkn: MIDOG has an announcement with 18 replies in case you wanted to see how pagination is currently implemented: https://grand-challenge.org/forums/forum/midog2021-598/

amickan: Thanks for finding that. At the moment, the post create page shows the form to create a post and then dumps the most recent 10 posts underneath that form in a scrollable card. That's not very convenient, but users do have a way to see the latest posts they are replying to if they scroll down and find the that section. 

I also see in one of the comments that the organizer missed some posts which appeared on the second page because she wasn't aware of the second page. I guess the "open topic at last unread post" would fix that issue.

amickan: Alright, I have bitten in the sour apple and turned both the PostCreate and PostUpdate posts into normal Django views. The create view now displays the latest 5 posts (seems sufficient to me) before the form, the update view does not (no need IMHO). 

I'm not 100% convinced that this a lot better than what I had before in terms of user friendliness, but it is slightly simpler from the developer perspective and hence easier to maintain.

jmsmkn: I thought that you were going to have the post create form at the bottom of the post list view? You can have a post a reply button and reveal it https://getbootstrap.com/docs/4.0/components/collapse/

amickan: That is what I had (implemented with HTMX), but I couldn't figure out how that would work with pagination if the user is on the first page.

jmsmkn: I don't get it? Why not:

```
{% for post in page %}
  # render post
{% endfor %}

{% if last_page %}
  # Render new post form
{% else %}
  # Reply Button that links to last page with the anchor for the new post form
{% endif %}
```

Or always render it?

amickan: I don't know. It seemed strange to me to pass in a form to a list view context. I haven't seen that done anywhere in the codebase, so then standard Django views seemed cleaner. What you suggests would make for the most user-friendly design though. I will continue working on this tomorrow.

jmsmkn: Can't quickly find a list view but we include forms in other views in several places, e.g. see `get_conversation_form` of `EvaluationDetail` or the user groups forms of `ReaderStudyDetail`, `AlgorithmDetail`, `JobDetail` etc.

amickan: I tried this out and it works well. The only weird thing maybe is that if you edit the last post in a topic, you end up with an edit form (rendered into place via HTMX) just above a create form (which is always there after the last post). They have clear titles, and the user only sees the other form if they scroll down, so probably fine:

image

PR #4030, opened 2025-05-15
Allow model-based POSTs to create SocketValues

chrisvanrun:
In rse-gcapi we have this definition of a request:
```Python
@dataclass
class ComponentInterfaceValuePostRequest(BaseModel):
    interface: str
    value: Optional[Any]
    file: Optional[bytes]
    image: Optional[str]
    upload_session: Optional[str]
    user_upload: Optional[str]
 ```
 
However, if one would try to submit using this model as a base we run into trouble with fields which are not allowed to be null.
 
This PR cleans up the fields and validation of the PostRequest as to account for some of the fields to be there but have a null value.
 
 This opens up the option to use local-validated models when POSTing from gcapi:
 
 ```Python
 
  c.archive_items.partial_update(
        pk=ai.pk,
        values=[
            ComponentInterfaceValuePostRequest(
                interface=socket.slug,
                value="foo",
                file=None,
                image=None,
                upload_session=None,
                user_upload=None,
            )
        ],
    )
```

### DIFF
--- a/app/grandchallenge/components/serializers.py
+++ b/app/grandchallenge/components/serializers.py
@@ -130,16 +130,6 @@ class ComponentInterfaceValuePostSerializer(serializers.ModelSerializer):
 
         interface = attrs["interface"]
 
-        # Check if exactly one of the specified keys is set
-        keys_to_check = ["image", "value", "user_upload", "upload_session"]
-        set_keys = [key for key in keys_to_check if attrs.get(key) is not None]
-
-        if len(set_keys) != 1:
-            raise serializers.ValidationError(
-                "Exactly one of 'image', 'value', 'user_upload', or "
-                "'upload_session' must be set."
-            )
-
         if interface.is_image_kind:
             if not any(
                 [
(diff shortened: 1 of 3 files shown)


app/grandchallenge/components/serializers.py
```diff
@@ -148,23 +141,37 @@ def validate(self, attrs):
...
+                    f"value is required for interface "
+                    f"kind {interface.kind}"
+                )
+        elif interface.requires_file:
+            if not any(
+                [
+                    attrs.get("file"),
+                    attrs.get("user_upload"),
+                ]
+            ):
```
amickan, line 158: Does this mean that we could now technically send over a `file` value, i.e. basically reuse file CIVs?

amickan (approved):
Just a small question, but otherwise LGTM

app/grandchallenge/components/serializers.py
chrisvanrun: I think using `file` is already currently possible: but it's the `file` field directly from the model. I **think** it entails that you could skip the creation of a UserUpload first.... but not sure. Haven't tested it: that path should already work now.

PR #4033, opened 2025-05-19
Topic lock and my posts view

amickan:
Remaining views for https://github.com/DIAGNijmegen/rse-roadmap/issues/407


amickan:
Failing tests are the Docker tests.

jmsmkn:
That should be fixed on main, could you rebase?

app/grandchallenge/discussion_forums/templates/discussion_forums/forumpost_list.html
```diff
@@ -0,0 +1,30 @@
...
+        <a href="{% url 'discussion-forums:my-posts' %}" class="btn btn-dark btn-sm"><i class="fas fa-comments mr-1"></i>My posts</a>
+    </div>
+{% endblock %}
+
+{% block extra_content %}
+    <h3>My forum posts</h3>
+    <div class="row">
+        <div class="col-12" id="post-list">
+            {% for post in object_list %}
+                {% include 'discussion_forums/forumpost_detail.html' with object=post include_topic_link=True %}
```
jmsmkn, line 26: Same issue as the other PR, this is a full view being treated as a partial.

app/grandchallenge/discussion_forums/models.py
```diff
@@ -122,7 +122,12 @@ def assign_permissions(self):
             self.forum.parent_object.participants_group,
             self,
         )
-        # only challenge admins can delete this topic
+        # only challenge admins can change and delete this topic
```
jmsmkn, line 125: What about changing the topic subject separately from the lock? We may want a specific lock topic permission.

app/grandchallenge/discussion_forums/urls.py
```diff
@@ -9,13 +9,16 @@
...
+    ForumTopicLockUpdate,
+    MyForumPosts,
 )
 
 app_name = "discussion_forums"
 
 
 urlpatterns = [
     path("", ForumTopicListView.as_view(), name="topic-list"),
+    path("my-posts/", MyForumPosts.as_view(), name="my-posts"),
```
jmsmkn, line 21: Maybe `/posts/mine/` ? I think we have that pattern elsewhere.

app/grandchallenge/discussion_forums/views.py
```diff
@@ -245,3 +268,20 @@ def get_form_kwargs(self):
...
+        qs = super().get_queryset()
+        return qs.filter(creator=self.request.user)
+
+    @cached_property
+    def forum(self):
+        return self.request.challenge.discussion_forum
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data()
+        context.update({"forum": self.forum})
```
jmsmkn, line 286: Does it make sense to show posts from other forums in the context of this forum?

jmsmkn:
This PR has the same issues as #4029

jmsmkn (approved):
LGTM!

app/grandchallenge/discussion_forums/models.py
```diff
@@ -122,7 +123,12 @@ def assign_permissions(self):
             self.forum.parent_object.participants_group,
             self,
         )
-        # only challenge admins can delete this topic
+        # only challenge admins can lock and delete this topic
+        assign_perm(
+            "discussion_forums.lock_forumtopic",
```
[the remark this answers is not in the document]
amickan: Oh right, I missed that when refactoring this.

app/grandchallenge/discussion_forums/templates/discussion_forums/forumpost_list.html
```diff
@@ -1,46 +1,92 @@
...
+                          hx-post="{% url 'discussion-forums:topic-lock-update' slug=topic.slug %}"
+                          hx-vals='{"is_locked": "False"}'
+                          hx-swap="outerHTML"
+                          hx-target="body"
+                    >
+                        {% csrf_token %}
+                        <button type="submit" class="btn btn-warning btn-sm btn-subscription"><i class="fas fa-lock-open mr-1"></i>
+                            Unlock topic
+                        </button>
+                    </form>
```
jmsmkn, line 75: FYI - htmx is not needed here if you are doing target="body". This can be a normal Django form view and post which will then redirect. Then you shouldn't need the `topic_type_matching_dict` or any of the special handlers for htmx.

amickan: True. The `topic_type_matching_dictionary` is still necessary though. That's for the mirgation.

app/grandchallenge/discussion_forums/templates/discussion_forums/forumpost_list.html
```diff
@@ -1,46 +1,90 @@
...
+        <h3>My posts</h3>
+    {% endif %}
+
+    {% if topic %}
+        {% get_obj_perms request.user for topic as "user_perms" %}
+        <div class="row">
+            <div class="col-8">
+                {% if 'create_topic_post' in user_perms %}
+                    <div class="d-inline" title="{% if topic.is_locked %}This topic is locked. You cannot reply to it.{% endif %}">
+                        <a hx-get="{% url 'discussion-forums:post-create' slug=topic.slug %}" hx-swap="beforeend show:bottom" hx-target="#post-list" class="btn btn-primary btn-sm {% if topic.is_locked %}disabled{% endif %}">
```
jmsmkn, line 42: This is just disabling the button, the user can still post data and create a new post. Should the `create_topic_post` permission be removed and added when locking and unlocking the post, or it should be checked on PostCreate perhaps by adding `is_locked=False` in `get_object_or_404`. Both would need a test.

app/grandchallenge/discussion_forums/templates/discussion_forums/forumpost_list.html
jmsmkn: Ah I see!

app/grandchallenge/discussion_forums/templates/discussion_forums/forumpost_list.html
amickan: There is validation in the form, so the user can post but it wouldn't create a post when the topic is locked. There is a test for it. I thought that would be enough?

jmsmkn: Ah missed that, sounds good!

amickan: I guess I can also check whether the user is an admin in the view, of course, so I could use the `AccessMixin` to add the check for whether it's unlocked or the user is an admin.

jmsmkn: But so long as the user cannot make a post for a locked topic then that is fine.

amickan: I'm wondering if it makes sense to allow challenge admins to post to locked topics. We could simplify things and just not allow posting to locked topics for anyone. The challenge admin can always unlock the topic again, post and lock it again if they really need to post to the topic. Then the check in the form could go and we could revoke `lock_forumtopic` for everyone when a topic gets locked. This also simplifies the templates, because else we need to check in the template if the user is a challenge admin in multiple place to decide whether to render the form and disable the buttons etc.

amickan: I have implemented the permission revoking / regranting now. I like that solution best, I think. It also simplifies the templates: the reply button just won't appear for locked topics.

Issue #4014, opened 2025-05-13
Session Costs should be moved to Utilization application


jmsmkn:
To avoid a circular dependency

PR #4037, opened 2025-05-20
Remove circular dependency in migrations

koopmant:
Related to #4014


koopmant:
Note to self, migration workstations 0028 can be deleted after rebase

app/tests/test_dependencies.py
```diff
@@ -0,0 +1,36 @@
+import importlib
+from collections import defaultdict
+from pathlib import Path
+
+
+def test_no_circular_dependencies(settings):
```
jmsmkn, line 6: I would like to keep the tests folder clean if possible, so this could go in the existing core_tests/test_migrations.py

The migrations have been deployed to production so this can go too.

koopmant: Ah excellent, that's a better place for it

jmsmkn (approved):
LGTM

PR #4054, opened 2025-05-26
Fix API schema for `Answer`

chrisvanrun:
Related:
- https://github.com/DIAGNijmegen/rse-gcapi/issues/201#issuecomment-2909524299

This allows the deduced schema to set `nullable: True`, and subsequent derived models to allow for `None` values.


jmsmkn (approved):
LGTM, this is the correct solution for these sorts of bugs as the models themselves allow null.

PR #4057, opened 2025-05-28
Forum follows and notifications

amickan:


app/grandchallenge/notifications/forms.py
```diff
@@ -39,16 +36,17 @@ def clean(self):
...
         elif cleaned_data["content_type"] == ContentType.objects.get(
-            app_label="forum_conversation", model="topic"
+            app_label="discussion_forums", model="forumtopic"
         ):
-            perm = self.perm_handler.can_subscribe_to_topic(
-                Topic.objects.filter(id=obj_id).get(), self.user
+            perm = self.user.has_perm(
+                "view_forumtopic",
+                ForumTopic.objects.filter(pk=obj_id).get(),
             )
```
amickan, line 50: It seems strange to me to do a permission check in the form clean method. If I had to implement this now from scratch, I would move the check to the view. That would require constructing the form in the views where it is used and pass it into the context of the template through the view. The template tag that we're currently using could then be removed. That feels like a cleaner solution, but I figured I first get things working, and then optimize. What do you think about this, James?

jmsmkn: > figured I first get things working, and then optimize

Yes, this! Additionally always separate refactoring from new functionality. Make a separate PR with the refactoring onto main, then rebase your feature.

app/grandchallenge/notifications/filters.py
```diff
@@ -79,42 +83,50 @@ class FollowFilter(FilterSet):
...
+            app_label = None
+            model = None
+            kwargs = {}
+
+        if model:
+            name_qs = [x.pk for x in model.objects.filter(**kwargs).all()]
+            extra_filter = {
+                **{"object_id__in": name_qs},
+                **{
+                    "content_type__exact": ContentType.objects.filter(
```
jmsmkn: Style wise there are a couple of places in this file where `.filter(...).get()` is used, `.get(...)` would suffice.

Same with this dictionary construction, why the double unpacking? Wouldn't:

```python3

extra_filter = {
  "object_id__in": name_qs,
  "content_type": ContentType.objects.get(model=model_name, app_label=app_label)
}
```

Be sufficient?

app/grandchallenge/notifications/migrations/0008_migrate_forum_follows_and_notifications.py
```diff
@@ -0,0 +1,111 @@
...
+    Forum = apps.get_model("discussion_forums", "Forum")  # noqa: N806
+    ForumTopic = apps.get_model(  # noqa: N806
+        "discussion_forums", "ForumTopic"
+    )
+    MachinaForum = apps.get_model("forum", "Forum")  # noqa: N806
+    MachinaTopic = apps.get_model("forum_conversation", "Topic")  # noqa: N806
+    ContentType = apps.get_model("contenttypes", "ContentType")  # noqa: N806
+
+    new_topic_ct = ContentType.objects.get_for_model(ForumTopic)
+    new_forum_ct = ContentType.objects.get_for_model(Forum)
```
jmsmkn, line 23: Didn't know about `ContentType.objects.get_for_model`, useful!

app/grandchallenge/notifications/migrations/0008_migrate_forum_follows_and_notifications.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def get_matching_forum(*, old_forum_id):
+        old_forum = MachinaForum.objects.get(pk=old_forum_id)
```
jmsmkn, line 26: What about when the old forum does not have a challenge, such as the general forum?

app/grandchallenge/notifications/migrations/0008_migrate_forum_follows_and_notifications.py
```diff
@@ -0,0 +1,111 @@
...
+        old_topic = MachinaTopic.objects.get(pk=old_topic_id)
+        new_forum = get_matching_forum(old_forum_id=old_topic.forum.pk)
+        return ForumTopic.objects.get(
+            forum=new_forum,
+            creator=old_topic.poster,
+            subject=old_topic.subject,
+        )
+
+    notifications_to_update = []
+    for notification in Notification.objects.filter(
```
jmsmkn: This is going to go OOM, we have 600k post notifications and 7k reply notifications, modifying that many objects will take several hours even with batching. I don't know what the time limit is on the ECS task. See the utilization app for some strategy. 1000 objects fetched even when using `only` still took up 25% of the memory.

Maybe we just have an amnesty and delete the existing notifications?

app/grandchallenge/notifications/migrations/0008_migrate_forum_follows_and_notifications.py
```diff
@@ -0,0 +1,111 @@
...
+            NotificationTypeChoices.FORUM_POST_REPLY,
+        ]
+    ):
+        if notification.type == NotificationTypeChoices.FORUM_POST:
+            new_target = get_matching_forum(
+                old_forum_id=notification.target_object_id
+            )
+            new_action_object = get_matching_topic(
+                old_topic_id=notification.action_object_object_id
+            )
```
jmsmkn: How do we know if a notification is migrated? Does this run atomically?

app/grandchallenge/notifications/migrations/0008_migrate_forum_follows_and_notifications.py
```diff
@@ -0,0 +1,111 @@
...
+            "action_object_object_id",
+        ],
+        batch_size=1000,
+    )
+
+    ct_forum = ContentType.objects.get_for_model(MachinaForum)
+    ct_forumtopic = ContentType.objects.get_for_model(MachinaTopic)
+
+    follows_to_update = []
+    for follow in Follow.objects.filter(
```
jmsmkn: Follow is missing from the Django admin so I cannot get insights into how much data we have there, but for the object permissions it looks like there are about 1M.

amickan: There is a `Follows` admin in the Django admin. There are 1148776 objects, but I can't filter by content type easily, so not sure how many of those would need to be updated. In any case, a lot though.

app/grandchallenge/notifications/migrations/0008_migrate_forum_follows_and_notifications.py
amickan: Oh, I thought we had deleted that forum. Good to add a check here though and skip forums that don't have a challenge attached.

app/grandchallenge/notifications/migrations/0008_migrate_forum_follows_and_notifications.py
amickan: Right, I thought of batching when doing the bulk update later, but forgot that this is of course reads them all into memory as well.... 

Deleting seems ok to me in this case. I doubt people would miss old notifications. Arguably, the value of a notification disappears once it's been read...

app/grandchallenge/notifications/migrations/0008_migrate_forum_follows_and_notifications.py
amickan: No longer relevant if we don't migrate them anymore, but just for my understanding: my thinking was that we run this migration once and so each object gets updated once, and so we don't need to know if it's been migrated or not. 

If a notification has been migrated though, `get_matching_forum` would fail, so we could catch it with a try ... except block and skip notifications for which we can't retrieve a challenge forum (that would also cover the General Forum notifications).

app/grandchallenge/notifications/migrations/0008_migrate_forum_follows_and_notifications.py
amickan: So this also needs batching here, and not just while updating below. Follows do need to be updated, no option to just delete those.

app/grandchallenge/notifications/migrations/0008_migrate_forum_follows_and_notifications.py
amickan: However, even if we delete objects, we still need to batch process them to avoid OOM.

jmsmkn: You can do `queryset.delete()`, does everything db side.

jmsmkn:
Had some more questions:

1. Are we sure new notifications aren’t generated when migrating the topics and posts?
2. Should we check follow and notification permission for user and object when migrating?
3. How do we ensure that the notification deletion problem from the missing signals doesn’t happen again? Should all old notifications just be deleted as the data might have drifted?

amickan:
>     1. Are we sure new notifications aren’t generated when migrating the topics and posts?

Good question. My tests suggest that no new notifications and follows are created when migrating posts and topics. My best guess is that this is because migrations reference historical models and that those don't trigger signals. It's actually a bit of a bummer because if those were created automatically, all we would need to do in the migration is delete old follows and notifications with the machina content types, but alas that is not how it works. 

>     2. Should we check follow and notification permission for user and object when migrating?

For follows there is no need, in my opinion, since the follows are deleted when an admin or participant is removed from a challenge. For notifications, it might be a good idea. We would then need to check that the necessary follow object is still in place. But since we've decided not to migrate notifications, but rather to delete them, we don't need to worry about it. 

>     3. How do we ensure that the notification deletion problem from the missing signals doesn’t happen again? Should all old notifications just be deleted as the data might have drifted?

I was thinking about that as well, but couldn't really find a solution. Deleting old notifications would be fine, but it doesn't ensure that a future developer remembers to add new follow or notification targets to the respective signals. For follows, there is a `registry` that keeps track of all possible follow object types. We could use that to construct a test to make sure that all models registered in that registry are covered by a signal, I suppose.

amickan:
> It's actually a bit of a bummer because if those were created automatically, all we would need to do in the migration is delete old follows and notifications with the machina content types, but alas that is not how it works.

On second thought this is not true because people can unsubscribe from forums and topics, or subscribe to topics they are not directly part of. Those wouldn't be captured by the standard follows.

app/grandchallenge/notifications/filters.py
amickan: Absolutely. No idea why I did that.

jmsmkn:
> future developer remembers to add new follow or notification targets to the respective signals

Maybe have a test that iterates through all of the follow kinds, creates an object and tests the deletion? Then it would fail and the developer would be forced to update the test and add the signal. May be messy.

app/grandchallenge/notifications/migrations/0008_migrate_forum_follows_and_notifications.py
amickan: I have added the batching here. I also checked and ECS tasks do not have a time limit, so this should work.

jmsmkn:
> My best guess is that this is because migrations reference historical models and that those don't trigger signals.

Please can you research this and work out exactly why. This is important as our system relies on signals.

app/grandchallenge/notifications/migrations/0008_migrate_forum_follows_and_notifications.py
```diff
@@ -0,0 +1,95 @@
...
+            follow.content_type = new_forum_ct
+        else:
+            topic = get_matching_topic(old_topic_id=follow.object_id)
+            follow.object_id = topic.pk
+            follow.content_type = new_topic_ct
+
+        follows_to_update.append(follow)
+
+        if len(follows_to_update) >= batch_size:
+            Follow.objects.bulk_update(
```
jmsmkn, line 76: Would be nice to have some feedback on the progress, both when an individual batch is updated and when it has completed.

app/grandchallenge/notifications/migrations/0008_migrate_forum_follows_and_notifications.py
```diff
@@ -0,0 +1,95 @@
...
+    ):
+        if follow.content_type == ct_forum:
+            try:
+                forum = get_matching_forum(old_forum_id=follow.object_id)
+            except ObjectDoesNotExist:
+                continue
+            follow.object_id = forum.pk
+            follow.content_type = new_forum_ct
+        else:
+            topic = get_matching_topic(old_topic_id=follow.object_id)
```
jmsmkn: What if the topic belongs to a forum without a challenge?

jmsmkn:
Okay, clear, then we need to be extremely careful about performing data migrations and always need to remember this.

jmsmkn (approved):
LGTM

PR #4059, opened 2025-05-30
Shorten paths found in starter-kit / algorithm templates

chrisvanrun:
Related:
-https://github.com/DIAGNijmegen/rse-grand-challenge-forge/issues/92

This PR updates to latest forge version that has some shorted path names.

It also hides the options to download the kit if there are no interfaces configured.

Finally, it removes the 'starter' from 'starter-kit' in the generated .ZIP file. It had already removed adding the containing directory: however several unzipping apps tend to automatically add the name of the filename as holding directory if there is not a singular folder/item in the ZIP. Which makes sense but this does add to the path length.

I think that was all the 'fat' that was on the path lengths in the forge and it can't be made much leaner. When the forge gets pulled into GC we should likely rewrite the generation to work on the Phase level. It currently operates on the Challenge level with a singular phase in the context. That would allow us to also remove the intermediate phase folder.

I'll mention unpacking is most wise under Linux env in the doc page on the starter kit.

### DIFF
--- a/app/grandchallenge/evaluation/views/__init__.py
+++ b/app/grandchallenge/evaluation/views/__init__.py
@@ -1507,6 +1507,6 @@ class PhaseStarterKitDownload(
         return FileResponse(
             streaming_content=buffer,
             as_attachment=True,
-            filename=f"{phase.challenge.short_name}-{phase.slug}-starter-kit.zip",
+            filename=f"{phase.challenge.short_name}-{phase.slug}-kit.zip",
             content_type="application/zip",
         )


app/grandchallenge/evaluation/templates/evaluation/phase_starter_kit.html
```diff
@@ -35,16 +35,16 @@ <h2 class="mb-3">Starter kit for {{ phase.title }}</h2>
         </ul>
     </p>
 
-    {% if phase.archive and phase.submission_kind == phase.SubmissionKindChoices.ALGORITHM %}
+    {% if phase.archive and phase.submission_kind == phase.SubmissionKindChoices.ALGORITHM and phase.algorithm_interfaces.exists %}
```
amickan, line 38: Does calling 'exists' work? I would have thought that it does not since you can't typically call methods in Django templates with the exception of a few (like `all()` on the queryset), but I don't think `exists()` is part of the list of auto-called methods.

chrisvanrun: It does. I tested it manually which seems to work. Maybe you are confusing with a set of filters we have?

Both methods and properties can readily be used in Django templates. As in: an no-arg method (`self` does not count) accessed as a property will get autocalled and this is not limited to a numbered set: https://docs.djangoproject.com/en/5.2/ref/templates/language/#accessing-method-calls

amickan: Ah sorry, I had a vague memory that this didn't work, but if you've tested it, great!

PR #4070, opened 2025-06-04
Enable tracking of read status of topics

amickan:


app/grandchallenge/discussion_forums/migrations/0002_migrate_old_forums.py
```diff
@@ -21,6 +22,36 @@ def migrate_challenge_forums(apps, schema_editor):
 
     for challenge in Challenge.objects.all():
         new_forum = Forum.objects.create()
+
+        # assign permissions
```
amickan: I had missed this because most of the time when testing the migrations, I was logged in a superuser so as to be able to see the django admin and the objects there. Doing a full check with normal users post migration, I realized that we need to explicitly grant these permissions here since they are otherwise only created when a challenge is created.

amickan: We could also move these to the forum model itself and assign permissions when adding a new forum. Then we wouldn't need those permissions here in the migration. Only downside is that we would then assume that each forum has a linked challenge, but this is currently true, and could be updated in the future if this changes.

app/grandchallenge/discussion_forums/models.py
```diff
@@ -230,6 +240,11 @@ def save(self, *args, **kwargs):
 
         if adding:
             self.assign_permissions()
+            # mark topic (and hence this post) as read by the user
+            TopicReadRecord.objects.update_or_create(
+                user=self.creator,
+                topic=self.topic,
+            )
```
amickan: I'm wondering if we should do a permission check that `self.creator` has view permission for the topic. Where would be the best place for such a check? In the `save` method of the `TopicReadRecord` model? 
We're currently creating `TopicReadRecord`s here and in the view - in the view, we are already checking the permission.

app/grandchallenge/discussion_forums/migrations/0002_migrate_old_forums.py
```diff
@@ -44,11 +76,47 @@ def migrate_challenge_forums(apps, schema_editor):
...
+        return old_forum.challenge.discussion_forum
+
+    def get_matching_topic(*, old_topic_id):
+        old_topic = MachinaTopic.objects.get(pk=old_topic_id)
+        new_forum = get_matching_forum(old_forum_id=old_topic.forum.pk)
+        return ForumTopic.objects.get(
+            forum=new_forum,
+            creator=old_topic.poster,
+            subject=old_topic.subject,
+        )
```
amickan: These are identical to the helper methods in the migration in the notifications app, but I couldn't find a way to reuse them. I can't put them in `models.py` because they need to reference historical models, which are only available in migrations, not in models.py.

app/grandchallenge/discussion_forums/migrations/0002_migrate_old_forums.py
amickan: I just realized that this does not work as is actually. I need to retrieve the historical model records for the groups to get this to work. No big deal, but I'm thinking that it actually is easier and clearer to move this to the Forum model. Would you agree, James?

amickan: Hm, on second thought that does not save us from also adding the permissions in the migration since the assign_permissions function would not be available to the migration code and would also not be triggered since the custom save() method is not avaialble either. Ok, never mind then, I'll update the migration here to use the appropriate historical Group model and leave the rest as is.

jmsmkn: 🦆 happy to help.

amickan: Learning a lot about migrations today: `assign_perm` does not work in migrations (https://github.com/django-guardian/django-guardian/issues/281#issuecomment-156264129).

amickan: I think I got now, thanks 🦆 James ;-)

amickan: Too fast: looks like the discussion forum permissions are not available yet. I vaguely remember that we ran into this issue before, and that the solution was to first migrate the models and in a seperate step (as in a separate deployment) to migrate the permissions. Could you confirm that this is correct, James?

jmsmkn: Yes, the permissions are created on post migration, so there needs to be at least one `manage.py migrate` in between. If you're happy with the models then can go through stand-alone in a separate PR, that is what I did earlier in the week for the utilization app.

amickan: Ok. Yes, I think the models are good to go. This PR is the last PR, from what I can tell, that adds functionality and a new model. I will remove the permission stuff from the migration, and then this PR is ready for review. Once this PR is merged, I'll create a PR to separate out all models.

app/grandchallenge/discussion_forums/migrations/0002_migrate_old_forums.py
jmsmkn: Maybe create a function in a common accessible location that takes in the models as arguments?

app/grandchallenge/discussion_forums/migrations/0002_migrate_old_forums.py
```diff
@@ -44,11 +45,47 @@ def migrate_challenge_forums(apps, schema_editor):
...
+
+    for track in TopicReadTrack.objects.iterator(chunk_size=1000):
+        try:
+            new_topic = get_matching_topic(old_topic_id=track.topic.pk)
+        except ObjectDoesNotExist:
+            continue
+        TopicReadRecord.objects.create(topic=new_topic, user=track.user)
```
jmsmkn: Again, this is going to be a huge number of objects, so the writes also need to be chunked and progress added.

app/grandchallenge/discussion_forums/migrations/0002_migrate_old_forums.py
```diff
@@ -44,11 +45,47 @@ def migrate_challenge_forums(apps, schema_editor):
                 )
 
 
+def migrate_topic_tracks(apps, schema_editor):
```
jmsmkn, line 63: Not modified here, but on L42: as created is set to auto_now_add any setting of it will be ignored. You will need to check other places where that is done. To work around it I think you can unset auto_now_add on the initial model, then add it back in a migration after the data migration - but please test that.

app/grandchallenge/discussion_forums/models.py
jmsmkn: Save should just save, so definitely not here. In this case the permissions should be checked in the view (create permission on the topic) and/or the form (filter by permission for the foreign key to the topic).

app/grandchallenge/discussion_forums/models.py
```diff
@@ -193,6 +194,13 @@ def last_page_num(self):
         posts_per_page = ForumTopicPostList.paginate_by
         return math.ceil(post_count / posts_per_page)
 
+    def get_unread_topic_posts_for_user(self, *, user):
+        try:
+            read_record = TopicReadRecord.objects.get(user=user, topic=self)
+            return self.posts.exclude(created__lt=read_record.modified)
```
jmsmkn, line 232: The modified date will be set to the time that the data migration ran, essentially marking all posts as read for a user-topic relationship. See the comments about created time, but in this case you have `auto_now` to deal with.

app/grandchallenge/discussion_forums/templatetags/discussion_forum_extras.py
```diff
@@ -0,0 +1,8 @@
+from django import template
+
+register = template.Library()
+
+
+@register.simple_tag()
+def get_first_unread_topic_post_for_user(*, topic, user):
+    return topic.get_unread_topic_posts_for_user(user=user).first()
```
jmsmkn: N+1 problem - this is going to perform one or two database queries per topic, one to get the `TopicReadRecord`, the other to get the post for that topic.

app/grandchallenge/discussion_forums/views.py
```diff
@@ -124,10 +125,24 @@ def get_context_data(self, **kwargs):
...
             }
         )
         return context
 
+    def get(self, request, *args, **kwargs):
+        response = super().get(request, *args, **kwargs)
+
+        # mark topic as read by the user
+        TopicReadRecord.objects.update_or_create(
+            user=self.request.user,
```
jmsmkn: Shouldn't the anonymous user be excluded?

jmsmkn (commented):
Sorry I was so late with reviewing this. Please can you add some tests for the number of database queries for the list view to ensure that there is not an N+1 problem and that we are not tracking what the anonymous user has or has not read.

app/grandchallenge/discussion_forums/models.py
amickan: That makes sense. I've added the check there.

app/grandchallenge/discussion_forums/views.py
amickan: The anonymous user does not have permission to access this view, and so `self.request.user` would never be the anonymous user here. I can add an explicit check, of course, but I think it's not strictly necessary.

app/grandchallenge/discussion_forums/migrations/0002_migrate_old_forums.py
amickan: Right, sorry for the oversight, again!

app/grandchallenge/discussion_forums/migrations/0002_migrate_old_forums.py
amickan: That could work. Another option would be to use `bulk_create` in the migration which should not trigger `auto_now_add`. I still have to test if that is how it works, but if so, that would have my preference.

app/grandchallenge/discussion_forums/models.py
amickan: Good catch. I think again bulk creating them (which I do now anyway) and transferring the modified time from the original object is the solution here.

app/grandchallenge/discussion_forums/migrations/0002_migrate_old_forums.py
amickan: Not sure why I thought bulk_create would ignore auto_now_add, but it does not, so this is not a solution to the problem. I will need to do what you suggested initially.

app/grandchallenge/discussion_forums/models.py
amickan: Same as above, I will need to unset auto_now

app/grandchallenge/discussion_forums/migrations/0002_migrate_old_forums.py
amickan: First creating the models without the auto_now_add, then migrating and then adding it back appears to work.

app/grandchallenge/discussion_forums/views.py
amickan: I have added the extra check just in case

app/grandchallenge/discussion_forums/migrations/0002_migrate_old_forums.py
jmsmkn: Good to know, thanks for confirming that.

app/grandchallenge/discussion_forums/templatetags/discussion_forum_extras.py
amickan: True. I had not considered that yet. I'm only using this to scroll to the first unread post on the topic detail page. I could move that logic to the topic detail view, check there for unread posts for the user for just the topic in question (thus avoiding the N+1) and then redirect to the correct page accordingly. 

While looking at the queries, I also realized that we are wasting a lot of queries on getting the `user_profile_link` for the creator of the last post. The last post is currently a property on the Topic model, so we can't prefetch the user profile and verification. In the original machina implementation, last_post if a FK field that is updated whenever a new post is created. Having it as a model field would enable us to prefetch it accordingly, which I think might be necessary given that we call this twice per topic on the topic list view.

amickan: I do also need to check if there are unread posts for each topic to mark them accordingly, so I do still need some version of this template tag on the topic list view. I think I could turn it into a simple check like this, which together with making `last_post` a FK field on the topic model, which I prefetch on the view, would solve the N+1 problem, I think: 


```python
@register.simple_tag()
def check_unread_topic_posts_for_user(*, topic, user):
    try:
        record = TopicReadRecord.objects.get(user=user, topic=topic)
        if topic.last_post.created >= record.modified:
            return True
        else:
            return False
    except ObjectDoesNotExist:
        return True
```

amickan: Confirmed that this approach drastically reduces the number of queries, and gets rid of all duplicate queries.

amickan:
> Sorry I was so late with reviewing this. Please can you add some tests for the number of database queries for the list view to ensure that there is not an N+1 problem and that we are not tracking what the anonymous user has or has not read.

I have addressed all comments, and added / updated tests, and this is ready for re-review. The only test I'm struggling with is the one that checks the number of queries. When using the debug toolbar, I see that with each new topic there is 1 additional query (to get the TopicReadRecord), but in the test it doesn't matter how many extra topics I add ,the number of queries stays the same. I'm probably doing something wrong, but I can't see at the moment. I'll check again tomorrow.

app/grandchallenge/discussion_forums/models.py
```diff
@@ -342,3 +385,31 @@ class ForumPostUserObjectPermission(UserObjectPermissionBase):
...
+    class Meta:
+        unique_together = [
+            "user",
+            "topic",
+        ]
+
+    def save(self, *args, **kwargs):
+        if self.user.username == settings.ANONYMOUS_USER_NAME or isinstance(
+            self.user, AnonymousUser
+        ):
```
amickan: The second condition might be redundant since I think `AnonymousUser` is not an instance of the `User` model and hence could not be saved to the DB anyway, but I'm not 100% sure.

amickan: Overall it does feel a little overkill to check this on the model given that we also have guardrails on the view and in the method that is used to create these instances.

amickan:
I think I have found a way to reduce the number of queries even further by doing some custom prefetching and then using those prefetched records in the template tag.

app/grandchallenge/discussion_forums/models.py
jmsmkn: How do you know in the future that things will be the same? That developers will always call things through these routes. We have these sorts of checks everywhere, like when adding users to groups. It is much better practice to code defensively.

amickan: True. The check here is a good addition then.

jmsmkn (approved):
LGTM, nice use of prefetch.

PR #4073, opened 2025-06-05
Replace`add_images_to_archive_item` with `add_image_to_object`

chrisvanrun:
There was some lock contention trouble with `add_images_to_archive_item`. Instead of adding a retry wrapper I favored replacing it with the already lock-contention safe `add_image_to_object`.

Less code!

### DIFF
--- a/app/grandchallenge/archives/tasks.py
+++ b/app/grandchallenge/archives/tasks.py
@@ -1,14 +1,10 @@
 from django.db import transaction
-from django.db.transaction import on_commit
 
 from grandchallenge.archives.models import Archive, ArchiveItem
-from grandchallenge.cases.models import Image, RawImageUploadSession
+from grandchallenge.cases.models import Image
 from grandchallenge.components.models import (
     ComponentInterface,
     ComponentInterfaceValue,
-)
-from grandchallenge.components.tasks import (
-    add_image_to_component_interface_value,
 )
 from grandchallenge.core.celery import acks_late_micro_short_task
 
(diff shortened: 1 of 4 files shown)


app/grandchallenge/cases/serializers.py
```diff
@@ -268,9 +265,11 @@ def _get_linked_task(*, targets, interface):
             kwargs["interface_pk"] = interface.pk
         return add_images_to_archive.signature(kwargs=kwargs, immutable=True)
     elif "archive_item" in targets:
-        return add_images_to_archive_item.signature(
+        return add_image_to_object.signature(
```
amickan, line 268: Is there a test for the create method of the serializer somewhere? It all looks good to me but would be good to know for sure that this runs as expected.

app/grandchallenge/archives/tasks.py
```diff
@@ -39,34 +35,3 @@ def add_images_to_archive(*, upload_session_pk, archive_pk, interface_pk=None):
                 continue
             item = ArchiveItem.objects.create(archive=archive)
             item.values.set([civ])
-
-
-@acks_late_micro_short_task
-def add_images_to_archive_item(
```
jmsmkn, line 45: You cannot just remove celery tasks as there may be existing ones on the queue. You need to leave them in and handle the migration (in this case, create a new `add_image_to_object` task), deploy, then make a new PR to remove the task, then deploy again.

jmsmkn: If you don't do that and there is an old task then celery panics about not having the task, makes an exception, puts the message back on the queue, flooding sentry.

app/grandchallenge/components/tasks.py
```diff
@@ -1186,30 +1186,6 @@ def preload_interactive_algorithms():
     return consolidation_results
 
 
-@acks_late_micro_short_task
-@transaction.atomic
-def add_image_to_component_interface_value(
```
jmsmkn, line 1191: See message above.

app/grandchallenge/archives/tasks.py
chrisvanrun: Ah, I hadn't considered two-stepping this to clear the queues.  Good tip.

app/grandchallenge/cases/serializers.py
chrisvanrun: Yes. Albeit its in an older section and hence a bit harder to find:

https://github.com/comic/grand-challenge.org/blob/8c23ecbe2cb8c93f009bbaf5e0d8ae43561ccf47/app/tests/cases_tests/test_api.py#L395

jmsmkn (approved):
LGTM

PR #4083, opened 2025-06-10
Add headline and breadcrumbs to docpage search results

koopmant:
Make the search results more user friendly by showing the breadcrumbs and headlines. 

This adds a new field for storing the plain text of a documentation page for searching the headlines without markup. 

image


app/grandchallenge/documentation/models.py
```diff
@@ -72,8 +75,21 @@ def save(self, *args, **kwargs):
...
+
         super().save(*args, **kwargs)
 
+    def update_content_plain(self):
+        self.content_plain = BeautifulSoup(
+            md2html(
+                self.content,
+                link_blank_target=False,
+                create_permalink_for_headers=False,
+                process_youtube_tags=False,
```
jmsmkn: Why set these kwargs? Isn't the page normally rendered with the defaults? Same with the use in the migration.

app/grandchallenge/documentation/templates/documentation/docpage_detail.html
```diff
@@ -143,7 +143,25 @@
...
+                                        </a></li>
+                                    {% endif %}
+                                    <li class="breadcrumb-item"><a href="{% url 'documentation:detail' slug=result.parent.parent.slug %}">
+                                            {{ result.parent.parent.title }}
+                                    </a></li>
+                                {% endif %}
+                                <li class="breadcrumb-item"><a href="{% url 'documentation:detail' slug=result.parent.slug %}">
+                                        {{ result.parent.title }}
+                                </a></li>
+                            {% endif %}
```
jmsmkn: Maybe this can be generalised in a template tag?

app/grandchallenge/documentation/templates/documentation/docpage_detail.html
```diff
@@ -143,7 +143,25 @@
...
+                        </ol>
+                        <a href="{% url 'documentation:detail' slug=result.slug %}"><h5 class="pl-3 pt-1">{{ result.title }}</h5></a>
+                        <div class="mb-3 pl-3">{{ result.headline|clean }}</div>
```
jmsmkn, line 146: Why does the headline need cleaning?

app/grandchallenge/documentation/views.py
```diff
@@ -43,9 +44,11 @@ def get_context_data(self, **kwargs):
 
         if keywords:
             query = SearchQuery(keywords)
-            vector = SearchVector("title", "content")
+            vector = SearchVector("title", "content_plain")
```
jmsmkn: How is the performance of all of this? Do we need to add any indexes etc to the database? https://docs.djangoproject.com/en/5.2/ref/contrib/postgres/search/#performance

Shouldn't the vector be added to the model itself? https://docs.djangoproject.com/en/5.2/ref/contrib/postgres/search/#django.contrib.postgres.search.SearchVectorField

jmsmkn (commented):
Looks good, small questions.

app/grandchallenge/documentation/templates/documentation/docpage_detail.html
koopmant: Hmm, good question. I merely reverted this to what we had before. But the cleaning doesn't do anything here.

app/grandchallenge/documentation/templates/documentation/docpage_detail.html
koopmant: Good idea.

app/grandchallenge/documentation/models.py
koopmant: I wanted create permalink for headers to be false, because this adds the permalink with a paragraph break character to the text, which I did not want to show in the headlines. 
The other two are irrelevant, but I did not realize that. I'll remove those.

app/grandchallenge/documentation/templates/documentation/docpage_detail.html
koopmant: We do need to use `|safe` to avoid HTML escaping.

app/grandchallenge/documentation/views.py
koopmant: I'm not sure if it's necessary. But adding the vector to the model makes sense, then we don't have to create it for each query. I added it there and update it on `save()`. I've added an index for it as well.

app/grandchallenge/documentation/templates/documentation/docpage_detail.html
jmsmkn: As a guideline safe should never be used. If the string does contain html that needs to be rendered and it comes from a user source then clean should be used. If the string contains html that needs to be rendered and it comes from a safe source, then that means that we need to refactor something to use more appropriate constructs (e.g. format_html).

app/grandchallenge/documentation/templates/documentation/docpage_detail.html
koopmant: Added

app/grandchallenge/documentation/views.py
koopmant: > How is the performance of all of this?

As far as I can tell, adding the searchvector field with the index reduces the search time from ~30ms to 3-9ms locally.

app/grandchallenge/documentation/templates/documentation/docpage_detail.html
koopmant: SearchHeadline removes the XML tags before highlighting the search terms in the headline. But from the [PostgreSQL documentation](https://www.postgresql.org/docs/current/textsearch-controls.html#TEXTSEARCH-HEADLINE):
> The output from ts_headline is not guaranteed to be safe for direct inclusion in web pages. 

In our case, the search is performed on (what should be) plain text. So I'd say it is pretty safe already. 

I don't yet see how I can use format_html here, though.

jmsmkn: >(what should be) plain text

Exactly, but that is not guaranteed. In fact, if it were plain text there would be no need for either filter, from what you said this contains html markup that you want to preserve. Given this, it is not safe to use `safe`, so `clean` should be used to defend against potential XSS vulnerabilities.

koopmant: It really should be plain text. The markup comes from the highlighting done by `SearchHeadline`. 
But we're using beautifulsoup `get_text()` to get it, and I don't see any guarantees there either. So, `clean` it is.

app/grandchallenge/documentation/models.py
```diff
@@ -38,10 +44,15 @@ class DocPage(models.Model):
         related_name="children",
     )
 
-    history = HistoricalRecords(excluded_fields=["order", "parent", "slug"])
+    history = HistoricalRecords(
```
jmsmkn: We can remove the historical records whilst we're here, see https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/494

app/grandchallenge/documentation/models.py
```diff
@@ -23,6 +27,8 @@ class DocPage(models.Model):
     slug = AutoSlugField(populate_from="title", max_length=1024)
 
     content = models.TextField()
+    content_plain = models.TextField(default="", editable=False)
+    search_vector = SearchVectorField(null=True)
```
jmsmkn: Do we want to allow nulls after the data have been migrated?

app/grandchallenge/documentation/models.py
```diff
@@ -72,8 +83,20 @@ def save(self, *args, **kwargs):
                     + 1
                 )
 
+        self.update_content_plain()
+
         super().save(*args, **kwargs)
 
+        DocPage.objects.filter(pk=self.pk).update(
+            search_vector=SearchVector("title", "content_plain")
+        )
```
jmsmkn, line 87: Why is this done in a separate query and not set pre-save with `self.search_vector = SearchVector("title", "content_plain")`?

app/grandchallenge/documentation/models.py
koopmant: No

app/grandchallenge/documentation/models.py
koopmant: SearchVector only supports updating existing objects. Let me trace back the documentation on that.

jmsmkn: Then can it be wrapped in `if adding:`? Would save a double hit on each save.

app/grandchallenge/documentation/models.py
jmsmkn: Is it possible to remove then? Maybe not if SearchVector only supports updating existing objects.

app/grandchallenge/documentation/models.py
koopmant: So, the documentation does not explain it explicitly. But it does say about the `SearchVectorField`: "you'll need to keep it populated with triggers" [link](https://docs.djangoproject.com/en/5.2/ref/contrib/postgres/search/#:~:text=You%E2%80%99ll%20need%20to%20keep%20it%20populated%20with%20triggers)

The issue came up when creating a new record, at which point you get an error: 
```
Failed to insert expression "SearchVector ... " on documentation.DocPage.search_vector. 
F() expressions can only be used to update, not to insert.
```

I've tested the following:
```python
        if not self._state.adding:
            self.search_vector = SearchVector("title", "content_plain")

        super().save(*args, **kwargs)

        if self._state.adding:
            DocPage.objects.filter(pk=self.pk).update(
                search_vector=SearchVector("title", "content_plain")
            )
```

This does not trigger the error. But after updating an existing page, the search vector is then populated with the old values. 

An alternative solution to the post save update would be the following:
```python
self.search_vector = SearchVector(Value(self.title), Value(self.content_plain))
```

I'll change it to that last solution, that saves a query.

koopmant: That makes the query literally contain the content of the page, so it becomes quite big. 
In terms of timing, I don't see a difference.

app/grandchallenge/documentation/models.py
koopmant: Sorry, forgot to push. I've replaced it with `default=""`.

app/grandchallenge/documentation/models.py
koopmant: Having thought about it some more, I think I prefer what we had previously. (With the separate update call.)
Using `Value(self.content_plain)` should be safe, but I'm not entirely sure. It could also affect performance. And could we hit query size limits? 
What is your preference?

jmsmkn: > But after updating an existing page, the search vector is then populated with the old values.

That is because `self._state.adding` is no longer true after calling `super().save(...)`. You need to set `adding = self._state.adding` at the start of the save method, then use `if adding` elsewhere. You will be able to find existing values in the codebase.

That would then solve the follow up issue.

app/grandchallenge/documentation/models.py
koopmant: I've created a separate pull request to remove all of them. https://github.com/comic/grand-challenge.org/pull/4104

app/grandchallenge/documentation/models.py
koopmant: You're right, I did not use self._state.adding correctly. Thanks for the tip. 

That does not solve the issue though. I tested it like this:
```python
        if not adding:
            self.search_vector = SearchVector("title", "content_plain")

        super().save(*args, **kwargs)

        if adding:
            DocPage.objects.filter(pk=self.pk).update(
                search_vector=SearchVector("title", "content_plain")
            )
```
manual test:
```console
> from grandchallenge.documentation.models import DocPage
> doc_page = DocPage.objects.create()
> doc_page.content = "1"
> doc_page.save()
> DocPage.objects.filter(search_vector="1")
<QuerySet []>
> doc_page.content = "2"
> doc_page.save()
> DocPage.objects.filter(search_vector="2")
<QuerySet []>
> DocPage.objects.filter(search_vector="1")
<QuerySet [<DocPage: >]>
```

jmsmkn: Super strange, I am learning about these at the same time. Then indeed what you originally had might be best.

koopmant: Ok, reverted to a post save update.

jmsmkn: Thanks! Sorry for the goose chase.

jmsmkn (approved):
LGTM!

PR #4085, opened 2025-06-11
Temporarily prevent new forum notifications and follows

amickan:

This ensures that no new posts and topics can be created while we migrate to the new forum implementation and that no new notifications or follows can be created in the meantime either.

This will make the forums temporarily unavailable to users, so it should only be merged and deployed directly before we merge and deploy the migration to the new forums.


jmsmkn (commented):
I'm not too sure that this is the best way of going about it. The tables still exist, so new objects can be created. Maybe that only happens when challenges are created?

Alternatively - you could remove the urls from the urlconf, that would be quicker to deploy but you might need to make some changes to the templates to stop the resolution.

amickan:
These permissions are created when a challenge is created and when a challenge is changed from hidden to public or vice versa. I had forgotten about the latter, and foolishly thought it was fine (no open requests, so no new challenge while we deploy), but you're right - just removing the permissions is not save enough. Removing the urls is easy enough, so I went with that approach. Thanks for the suggestion.

jmsmkn:
> It's no ideal to break challenge participant management during challenge high season, but I think we have no choice. We will need to plan the migration well to minimize downtime.

We can't do that at any point - this migration is going to take on the order of a day. Please can you think about another way to migrate these data? You will need to re-work things so that we can interrupt the process and resume from where we left off, that should handle anything created in the meantime.

amickan:
I think then we might need to switch to doing this in (a) management command(s) and to adding a boolean field `migrated` to all models for which data needs to be migrated (all forum models and the Follow model). These fields can be removed again when the migration is complete. 

We can have separate management commands for the separate migration steps to keep things more manageable and configure them to only process unmigrated objects. That should enable us to interrupt them and to rerun them without duplicating data and would deal with records created in the meantime. 

The way I understand migrations, I don't think that would work with migrations since these run in transactions and so if we interrupt them, we'd start from square one.

jmsmkn:
It is better not to modify the existing objects. If the new objects don't naturally reflect the source objects, then it is better to add a source foreign key to the original.

You could use multiple data migrations as the transaction is only for each individual migration to break things up. I think that a management command is worth it too as you get direct feedback and can run that prior to migrating, then the migration step cleans up everything created in-between.

PR #4088, opened 2025-06-11
Remove machina settings and deprecate forum field

amickan:


amickan:
I can't work out why the 2 tests fail. From what I see, none of my changes should affect these tests. For the healthcheck test, the number of queries has reduced but that is not due to the change in the `select_related()` on the subdomain middleware. Rather it seems that the session data is not fetched, which might also be the cause of the 2nd test failure. I don't see why that is though.

amickan:
Turns out that retrieving the session was done by the `ForumPermissionMiddleware`, which I removed here. I've updated the test accordingly. The other test succeeds locally, so apparently a flaky test.

amickan:
Actually `test_no_user_is_assigned_to_session` does fail locally as well and the reason is the removal of the `ForumPermissionMiddleware`. That middleware modified the session, which triggered it to be saved to the DB. If not modified, [session data is not automatically saved to the DB](https://docs.djangoproject.com/en/5.2/topics/http/sessions/#when-sessions-are-saved), which is why the test now fails.

I'm not sure what the solution to this should be. Do we want to always save session data to the DB? In that case, there is a setting in Django we could enable `SESSION_SAVE_EVERY_REQUEST`. I don't see any (custom) code that uses the session data though, so maybe we are fine with just not saving the session data unless modified?

PR #4091, opened 2025-06-13
Add GPU availability issue warning

amickan:
image

See https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/537

### DIFF
--- a/app/grandchallenge/algorithms/templates/algorithms/algorithm_detail.html
+++ b/app/grandchallenge/algorithms/templates/algorithms/algorithm_detail.html
@@ -112,7 +112,7 @@
 
 {% block content %}
     {% if 'change_algorithm' in algorithm_perms %}
-        {% if object.job_requires_gpu_type == gpu_type_choices.A10G and object.job_requires_memory_gb > 16 %}
+        {% if object.job_requires_gpu_type == object.GPUTypeChoices.A10G and object.job_requires_memory_gb > 16 %}
             <div class="alert alert-danger" role="alert">
                 <p><i class="fas fa-exclamation-triangle mr-1"></i><u>Instance Availability Issue</u></p>
                 <p>
--- a/app/grandchallenge/algorithms/views.py
+++ b/app/grandchallenge/algorithms/views.py
@@ -83,7 +83,6 @@ from grandchallenge.components.backends.exceptions import (
     CIVNotEditableException,
 )
 from grandchallenge.components.models import ImportStatusChoices
-from grandchallenge.components.schemas import GPUTypeChoices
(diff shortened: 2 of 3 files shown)


miriam-groeneveld (approved):
LGTM!

app/grandchallenge/algorithms/templates/algorithms/algorithm_detail.html
```diff
@@ -112,6 +112,21 @@
 
 {% block content %}
     {% if 'change_algorithm' in algorithm_perms %}
+        {% if object.job_requires_gpu_type == gpu_type_choices.A10G and object.job_requires_memory_gb == 32 %}
```
jmsmkn:
```suggestion
        {% if object.job_requires_gpu_type == gpu_type_choices.A10G and object.job_requires_memory_gb > 16 %}
```

app/grandchallenge/algorithms/templates/algorithms/algorithm_detail.html
```diff
@@ -112,6 +112,21 @@
...
+            <div class="alert alert-danger" role="alert">
+                <p><i class="fas fa-exclamation-triangle mr-1"></i><u>A10G GPU Availability Issue</u></p>
```
jmsmkn: Don't know if there is better wording? It is only an issue for high memory instances.

```suggestion
                <p><i class="fas fa-exclamation-triangle mr-1"></i><u>Instance Availability Issue</u></p>
```

app/grandchallenge/algorithms/views.py
```diff
@@ -267,6 +268,7 @@ def get_context_data(self, **kwargs):
                 "pending_permission_requests": pending_permission_requests,
                 "algorithm_perms": get_perms(self.request.user, self.object),
                 "best_evaluation_per_phase": self.best_evaluation_per_phase,
+                "gpu_type_choices": GPUTypeChoices,
```
jmsmkn: It should already be available on `object.GPUTypeChoices`

amickan: It's not. `GPUTypeChoices` is a separate class outside of the Algorithm model.

jmsmkn: Ah, it's only set on the ComponentImage, it is a choice for the Algorithm so it is worth setting there too to keep the same pattern

https://github.com/comic/grand-challenge.org/blob/8a231b55cdd06ee43d7e18a37838eb3bb0ee2314/app/grandchallenge/components/models.py#L1986

amickan: Ok. Makes sense. I'll make the necessary changes.

jmsmkn (approved):
LGTM!

Issue #4082, opened 2025-06-10
Job status button no longer refreshes on job list or detail page


jmsmkn:
The automatic refreshing of the job status badge was removed in #3767. @chrisvanrun  do you know why? If this was a mistake could you bring it back? I would expect the `span` in:

```html
    <span class="badge badge-{{ object.status_context }}">
        {% if object.animate %}
            <span class="spinner-border spinner-border-sm" role="status" aria-hidden="true"></span>
        {% endif %}
        {{ object.get_status_display }}
    </span>
```

to have these extra attributes:

```html
 <span ... {% if not object.finished %}hx-get="{{ object.status_url }}" hx-trigger="load delay:30s" hx-swap="outerHTML"{% endif %}>
```

chrisvanrun:
I think this might have been a merge mess up. I'll figure it out.

PR #4092, opened 2025-06-13
Re-add auto-updating of job status

chrisvanrun:
Fixes: #4082 

Was indeed a merge problem.

Now updates as one would expect:

https://github.com/user-attachments/assets/7e0f77f6-488d-472b-a7e2-faafbaabd172

### DIFF
--- a/app/grandchallenge/algorithms/templates/algorithms/job_status_detail.html
+++ b/app/grandchallenge/algorithms/templates/algorithms/job_status_detail.html
@@ -87,7 +87,6 @@
         {% if object.animate %}
             <span class="spinner-border spinner-border-sm" role="status" aria-hidden="true"></span>
         {% endif %}
-        {{ object.get_status_display }}{% if object.status == object.SUCCESS and object.stderr %},
-            with warnings{% endif %}
+        {{ object.get_status_display }}
     </span>
 {% endif %}


jmsmkn:
Does everything still work when you have multiple pages of stuff and you go back and forth between them? There's no issue with the button disappearing from datatables?

chrisvanrun:
> Does everything still work when you have multiple pages of stuff and you go back and forth between them? There's no issue with the button disappearing from datatables?

I've focused testing if the extra error reporting modals are placed where one would expect them. I'll quickly thow together some tests that push the datatables refresh, resize etc.

chrisvanrun:
Result table is server side rendered. So page switching shouldn't be problem. Small chance a weird table element gets replaced: not that I can tell. So htmx doesn't accidentally target the wrong element. Double checked if the htmx queries would continue if the dt switched pages: seems like it is correctly hooked into destroyed elements and stops when a new page gets send.

I did realize a potential problem. How bad is it that if all loaded Jobs in the list do a burst refresh of the job status? I.e. 50 GETs at the same time? We could add a random delay at to the render (30~40 sec).

chrisvanrun:
It seems HTMX blocks request in batches of 6, blocking batches with roughly 50 ms delays. Jeej for htmx!

PR #4098, opened 2025-06-16
Use window annotation for standard index

jmsmkn:
Follow on from #4097, this saves having to lookup the index of a display set by having them all in memory. This only works when **not** shuffling the hanging list, that is much more difficult to change.

### DIFF
--- a/app/grandchallenge/reader_studies/serializers.py
+++ b/app/grandchallenge/reader_studies/serializers.py
@@ -121,7 +121,16 @@ class DisplaySetSerializer(HyperlinkedModelSerializer):
                 # The list is empty if no reader study is specified.
                 return None
         else:
-            return obj.standard_index
+            try:
+                return obj.standard_index - 1
+            except AttributeError:
+                # The annotation wasn't made when getting this object
+                return (
+                    DisplaySet.objects.with_standard_index()
+                    .get(pk=obj.pk)
+                    .standard_index
+                    - 1
+                )
 
     class Meta:
         model = DisplaySet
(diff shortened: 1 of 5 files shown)


app/grandchallenge/reader_studies/serializers.py
```diff
@@ -115,13 +115,20 @@ class DisplaySetSerializer(HyperlinkedModelSerializer):
...
+            except AttributeError:
+                # The annotation wasn't made when getting this object
+                row_number = (
+                    DisplaySet.objects.with_row_number()
+                    .get(pk=obj.pk)
+                    .row_number
+                    - 1
+                )
+
+            return row_number - 1
```
amickan: Here you subtract 1 twice for cases where the annotation wasn't made when getting the object, right?

amickan: Ah I guess that's because the RowNumber() window function returns 1-based numbering, not 0-based.

jmsmkn: Good catch!

app/grandchallenge/reader_studies/serializers.py
```diff
@@ -115,13 +115,20 @@ class DisplaySetSerializer(HyperlinkedModelSerializer):
...
-            return obj.standard_index
+            try:
+                row_number = obj.row_number
+            except AttributeError:
+                # The annotation wasn't made when getting this object
+                row_number = (
+                    DisplaySet.objects.with_row_number()
+                    .get(pk=obj.pk)
+                    .row_number
+                    - 1
```
jmsmkn:
```suggestion
```

amickan (approved):

PR #4099, opened 2025-06-17
Fix prediction submissions in algorithm kind phase

chrisvanrun:
Currently one is able to change the Phase submission kind if there are existing submissions.

This is problematic because nigh all checks in templates etc are just looking at the **current** `submission_kind` of the phase and not when it was submitted.

This PR adds a check on the model level to ensure that one does not change the model when submissions exists and addresses the reported errors on templates that result from this mix.


jmsmkn (approved):
LGTM!

PR #4104, opened 2025-06-17
Remove usage historical records

koopmant:
Related to https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/494

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,6 +24,7 @@ dependencies = [
     "django-storages[s3]>=1.14.2",
     "boto3>1.18.8",
     "django-extensions",
+    "django-simple-history",
     "sentry-sdk",
     "django-cors-headers",
     "django-celery-beat!=1.6.0",
--- a/uv.lock
+++ b/uv.lock
@@ -983,6 +983,18 @@ wheels = [
 ]
 
 [[package]]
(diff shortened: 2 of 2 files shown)


app/grandchallenge/blogs/migrations/0001_initial.py
```diff
@@ -119,6 +118,6 @@ class Migration(migrations.Migration):
                 "ordering": ("-history_date", "-history_id"),
                 "get_latest_by": "history_date",
             },
-            bases=(simple_history.models.HistoricalChanges, models.Model),
+            bases=(models.Model,),
```
jmsmkn: Is it safe to remove these from old migrations? I'm not sure on the consequences of modifying bases.

koopmant: I'm not sure about it either, but I don't have an alternative yet.

jmsmkn: Maybe it is better to keep simple history installed for the time being, until all the data are removed?

koopmant: Sounds good to me.

jmsmkn (approved):
LGTM

PR #4108, opened 2025-06-18
Highlight headline paragraphs after following a documentation search result

koopmant:
Make searching through the documentation more user friendly by highlighting the headline on the page after clicking on a search result. 

Example search result:
image

Highlighted elements containing the headline:
image

### DIFF
--- a/app/grandchallenge/documentation/static/documentation/js/highlight.mjs
+++ b/app/grandchallenge/documentation/static/documentation/js/highlight.mjs
@@ -43,121 +43,62 @@ function highlightRangeAndScroll(startIndex, endIndex, rootElement) {
 
     const context = {
         characterCount: 0,
-        highlighting: false,
         first_node: null,
     };
 
     traverseAndHighlight(rootElement, startIndex, endIndex, context);
 
     if (context.first_node) {
-        // scrollIntoView works on elements, so if first_node is a text node,
-        // we use its parentElement.
-        const scrollTarget =
-            context.first_node.nodeType === Node.TEXT_NODE
-                ? context.first_node.parentElement
-                : context.first_node;
-
-        scrollTarget.scrollIntoView({
+        context.first_node.scrollIntoView({
             behavior: "smooth",
             block: "center",
         });
     }
 }
 
-// List of tags that should be wrapped entirely, not split.
-const FULL_HIGHLIGHT_NODES = ["STRONG", "B", "EM", "I", "MARK", "SPAN", "A"];
+// List of element tags that should be treated as a single, indivisible unit for highlighting.
+const ATOMIC_NODES = ["STRONG", "B", "EM", "I", "MARK", "SPAN", "A"];
 
 /**
- * Recursively traverses the DOM and highlights text based on character indices.
+ * Recursively traverses the DOM, highlighting full nodes that fall within a
+ * character range.
  */
 function traverseAndHighlight(node, startIndex, endIndex, context) {
-    if (context.characterCount >= endIndex || context.done) {
+    if (context.characterCount > endIndex) {
         return;
     }
 
-    if (node.nodeType === Node.ELEMENT_NODE) {
-        // Check if the current element is one we should wrap whole.
-        if (FULL_HIGHLIGHT_NODES.includes(node.tagName)) {
-            const nodeText = node.textContent.replace(/\s+/g, " ");
-            const nodeLength = nodeText.length;
-            const startOfNode = context.characterCount;
-            const endOfNode = startOfNode + nodeLength;
+    const isAtomic =
+        node.nodeType === Node.TEXT_NODE ||
+        (node.nodeType === Node.ELEMENT_NODE &&
+            ATOMIC_NODES.includes(node.tagName));
 
-            // Check if this node's text content falls within the highlight range
-            const shouldHighlight =
-                (context.highlighting && endOfNode <= endIndex) || // Already highlighting and this node is fully contained
-                (startOfNode < endIndex && endOfNode > startIndex); // Node overlaps with the highlight range
-
-            if (shouldHighlight) {
-                // If this is the very first thing we highlight, store it for scrolling
-                if (!context.first_node) {
-                    context.first_node = node;
-                }
-                // wrapElementNode(node); // Wrap the entire element
-                node.classList.add("mark"); // Add a class for styling (or wrap the element as desired)
-                context.highlighting = true; // Ensure we stay in highlighting mode
-            }
-
-            // Update character count and, crucially, DO NOT traverse children.
-            // We've treated this element as an atomic unit.
-            context.characterCount = endOfNode;
-            return;
-        }
-
-        for (const child of Array.from(node.childNodes)) {
-            traverseAndHighlight(child, startIndex, endIndex, context);
-        }
-        return;
-    }
-
-    if (node.nodeType === Node.TEXT_NODE) {
-        // Your whitespace handling is complex. For simplicity in this example,
-        // we'll use a more direct textContent length, but your logic
-        // for collapsing whitespace would go here. To correctly align with `normalizedInnerText`,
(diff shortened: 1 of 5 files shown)


koopmant:
I've based this branch on https://github.com/comic/grand-challenge.org/pull/4083 because it relies on the headline in the search results. Will rebase after that one has been merged.

app/grandchallenge/documentation/static/documentation/js/highlight.mjs
```diff
@@ -0,0 +1,47 @@
+document.addEventListener("DOMContentLoaded", () => {
```
jmsmkn, line 1: Do we need custom js for this? Can't it be done with built-in browser functionality: https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Fragment/Text_fragments

koopmant: Oh, that's silly. I thought that was Chrome specific... Ok, that makes this a lot simpler.

jmsmkn: This site is a good guide: https://caniuse.com/mdn-html_elements_a_text_fragments

koopmant: Thanks! 🙌

koopmant: I've been trying to make this work for a couple of hours. But there are some issues with text fragments. 
Most importantly, we cannot limit the target area to the docpage content `#pageContainer`. 
This is especially an issue with the page title, which is shown in the breadcrumbs and at the top of the docpage. If the headline starts with a part of the title, the text fragment usually finds this part in the breadcrumbs and highlight everything from there. For example, https://grand-challenge.org/documentation/create-your-own-algorithm/#:~:text=your%20own%20algorithm,to%20get%20your%20algorithm
image

This is usually resolved by providing more context with prefix and suffix strings. But since we only have the provided headline from the search results, we cannot analyze the target page to create this extra context.

I propose we go with the custom js.

app/grandchallenge/documentation/static/documentation/js/highlight.mjs
```diff
@@ -0,0 +1,47 @@
...
+            startNode = el;
+            capturing = true;
+        }
+
+        if (capturing) {
+            el.classList.add("highlight");
+            num_highlighted_elements += 1;
+        }
+
+        if (num_highlighted_elements === 3) break;
```
HarmvZ: Why stop at 3?

app/grandchallenge/documentation/static/documentation/js/highlight.mjs
```diff
@@ -0,0 +1,47 @@
...
+
+        if (!endText && startNode) break;
+
+        if (endText && capturing && text.includes(endText)) break;
+
+        last_two_words = text.split(" ").slice(-2).join(" ");
```
HarmvZ: Why is this needed?

app/grandchallenge/documentation/static/documentation/js/highlight.mjs
```diff
@@ -0,0 +1,47 @@
...
+    if (!highlightText) return;
+
+    const content = document.getElementById("pageContainer");
+    if (!content) return;
+
+    const [startText, endText] = highlightText.split(":::");
+
+    if (!startText) return;
+
+    const elements = content.querySelectorAll("p, h3, h4, h5, pre, td, li");
```
HarmvZ: This seems fragile. Why not broaden this to any element inside the doc page?

HarmvZ (changes requested):
Nice feature! I have a few questions.

This method matches only the first occurrence of the string, right? If there is duplicate text on the page, it could end up highlighting the wrong section. Can we instead use numerical indexing to highlight text from the start_n character to the end_n character?

app/grandchallenge/documentation/static/documentation/js/highlight.mjs
koopmant: I get what you mean. Maybe there is a more generic selector for what we need? We need to go though the subsequent blocks which are at the right level. They should also not be nested. I selected these to for example go through the cells of the table, but not the whole table, go though the list items, but not the whole list.

app/grandchallenge/documentation/static/documentation/js/highlight.mjs
koopmant: It's a failsafe, in case the endText is not found. (Then the rest of the page would get highlighted as well.) The search result headline is always a short piece of text, so I capped it at three elements. 

I added this before I used the `last_two_words` to search across elements. Since then, the search performs much better; so I'm not sure if this is still needed. If the search fails and the rest of the page is highlighted, it looks really bad though. Better safe than sorry?

koopmant:
That sounds a lot better. I'll try that!

koopmant:
Updated it with highlighting parts of the elements, instead of the complete element. And now using the start and end chars of the headline in the full text. 

A small remaining issue is that mark tags are added within `strong` and `a` elements. The mark tags should go around these, otherwise these elements get extra space around them and the formatting for the anchors breaks. 

Another one I just found is that the start index can be incorrect. The script already ignores repeated whitespace, whitespace introduced by the DOM and generated `¶` characters. But there is apparently another cause of incorrect index. 

image

If the above is fixed and when the headline starts or ends halfway in the anchor text, the whole anchor should be highlighted, otherwise you get the same issue of broken formatting:

image

app/grandchallenge/documentation/static/documentation/js/highlight.mjs
```diff
@@ -0,0 +1,182 @@
+document.addEventListener("DOMContentLoaded", () => {
+    const params = new URLSearchParams(window.location.search);
+    const highlightText = params.get("highlight").replace(/\s+/g, " ");
```
HarmvZ: Is it possible to determine and send the start and end index in the search params instead of the text? That way it will not match incorrectly when the highlight text appears multiple times on the page.

koopmant: Right, forgot to mention why we don't do that yet. It's a bit difficult, because we use a postgresql function to search for the headline. https://www.postgresql.org/docs/current/textsearch-controls.html#TEXTSEARCH-HEADLINE
This function returns the highlighted headline excerpt, but no information about where in the text it was found. So on the search results page we have the same problem. I therefore forwarded the entire headline as highlight text and solve the problem of finding the position of headline during highlighting. 

It can still happen that we highlight the "incorrect" match, although it seems highly unlikely that the entire headline is repeated. From the user perspective you wouldn't even be able to tell, because in the search results you are also presented with the excerpt only.

koopmant:
Highlighting the entire nodes reduced the complexity a lot, thanks. That seems sensible. 
I still need to fix an issue with the text length normalization, which is still off sometimes. (But because we highlight entire nodes now, it's only noticable if the index is incorrectly before the start of a node)

koopmant:
> Highlighting the entire nodes reduced the complexity a lot, thanks. That seems sensible. I still need to fix an issue with the text length normalization, which is still off sometimes. (But because we highlight entire nodes now, it's only noticable if the index is incorrectly before the start of a node)

It's the same issue I had yesterday. When we normalize the full text we (have to) get rid of repeated whitespace. When we go through the nodes, we have to keep track of whitespace across nodes. For example a node ending with whitespace and the next node starting with it.

app/grandchallenge/documentation/static/documentation/js/highlight.mjs
HarmvZ: Allrigt, makes sense to keep it like this.

koopmant:
Had to revert the "search and highlight directly" approach, because that wasn't working. But in the process I managed to correct the character count during traversal. It appears to work well now!

PR #4111, opened 2025-06-19
Add VRAM to GPU Specs

jmsmkn:
Now that VRAM is getting more attention users are confusing this with DRAM.


chrisvanrun (approved):
Sensible cleanup of the tests.

PR #4114, opened 2025-06-20
New forum models

amickan:


amickan:
I have added and tested management commands for migrating the forum objects (forums, topics, posts and read records). 

I can create a management command for deleting notifications, but I think we would only want to do that right before we migrate. Since the deletion is done in the DB `.delete()` on the queryset, I think it would be fine to do this just as part of the migration, right? 

For the follows, I'm a bit unsure: the migration I wrote previously updates existing follows, rather than creating new ones and deleting old ones. We should definitely not do that until we actually migrate, so it makes no sense to turn that into a management command. However, this is probably the migration that takes the longest, so is the most problematic one. I'm now considering changing the strategy away from updating existing follows to creating new follows for the new objects and then once that is done delete all old follows in bulk. I could then write a management command that already creates new follows for the new objects. The deletion of the old follows could be its own command (maybe together with notifications) that we run right before deploying, or we just do this in the migration. The downside to this is that we would temporarily end up with a whole lot (hundreds of thousands) of "duplicate" follows and make the table even bigger than it already is, but that is probably the lesser evil, right? 

If I go for a management command, I would need to figure out how to keep track of which follow has been created and which still needs creating. If we don't want to add a new field to the Follow model to keep track of migration state, I think the only option would be to run 2 queries for each follow to check if a matching follow exists with the matching new forum/topic... not efficient either. Are there are other options that I'm not seeing?

app/grandchallenge/discussion_forums/models.py
```diff
@@ -0,0 +1,412 @@
...
+
+
+class Forum(UUIDModelNoAutoNow):
+
+    source_object = models.OneToOneField(
+        MachinaForum,
+        related_name="migrated_forum",
+        null=True,
+        on_delete=models.SET_NULL,
+    )
```
amickan: I made all these `OneToOne` fields rather than `ForeignKey`s because with a foreign key any given machina forum/topic could have multiple migrated forums/topics etc. We do not want that, so a one to one relation made more sense to me. Or did I miss something?

app/grandchallenge/discussion_forums/management/commands/migrate_topic_tracks.py
```diff
@@ -0,0 +1,80 @@
...
+
+            batch.append(
+                TopicReadRecord(
+                    source_object=track,
+                    topic=new_topic,
+                    user=track.user,
+                    created=track.mark_time,  # the original model does not have a creation time stamp
+                    modified=track.mark_time,
+                )
+            )
```
amickan: Currently any user on the platform can view any post in any forum. In the new implementation only challenge participants and admins can read their forum. I think that makes sense. However, when it comes to migrating the `TopicReadTracks`, it means we might end up adding inconsistent data if we just blindly copy over (non-participant read record for a forum topic). 

So I will need to add a check here that `track.user` has view permission for the `new_topic` before creating the `TopicReadRecord`. We already have this check in place on the new views/forms (see recent PR), but I had forgotten to add this in the migration.

amickan: I have added the permission check

amickan:
> 3 will not work as part of a migration as you cannot send argument. I thought the idea was to reverse relationship and exclude those where the relationship exists?

Yes, that's how the migration is written, but updating breaks notifications and the links to the follow objects on subscription overview. Not so much of an issue if we do this right before migrating, but the point of the management command would be to be able to do it in stages before, right? This is likely going to take quite a while anyway, so would break things for a while.

amickan:
Ok, I've added the missing management commands. For the follows I ultimately went with the update strategy that we also use in the migration. To mitigate the issue on the follow-list, I just disable the links. We should still hold off running the follow migration until maybe a day before we deploy the feature branch. That would then mean a day of no notifications for new forum posts, but I think that would be acceptable. Would you agree, @jmsmkn ?

app/grandchallenge/discussion_forums/management/commands/migrate_topic_tracks.py
```diff
@@ -0,0 +1,119 @@
...
+            TopicReadRecord.objects.bulk_create(batch)
+            total_created += len(batch)
+            self.stdout.write(
+                self.style.SUCCESS(f"Created {total_created} new records")
+            )
+            batch = []
+
+        for track in ForumReadTrack.objects.exclude(
+            user__username=settings.ANONYMOUS_USER_NAME
+        ).iterator(chunk_size=1000):
```
amickan, line 74: This entire section is new (hasn't been reviewed yet and is not yet part of the migration on the feature branch). I realized after some very thorough testing that machina only creates `TopicReadTracks` if a user either creates or replies to a topic, not if they "just" read it. When a topic is read, the `ForumReadTrack` gets either created or updated. 

So only migrating `TopicReadTracks` will leave a bunch of topics marked as unread when they have actually been read. For the full picture, we need to also look at `ForumReadTracks` in the migration to create the missing topic records. That again adds quite some complexity. I tried to keep the DB load at a minimum by creating mappings in python that I use in the checks, as well as batching the DB calls. I don't see a way to make this more efficient. 

I did consider adding a `ForumReadRecord` and mimicking the logic in machina, but I think our current solution with only having `TopicReadRecords` makes more sense and is more efficient down the line (in templates when checking whether a topic has been read for example).

app/grandchallenge/core/models.py
```diff
@@ -34,6 +34,20 @@ class Meta:
         abstract = True
 
 
+class UUIDModelNoAutoNow(models.Model):
```
jmsmkn, line 37: I guess this is just temporary? It's fine if it will be deleted. Just FYI - another option would be to explicitly set the overridden fields in the models that need it, saves duplicating an entire model. But for a small, temporary, model like this this is fine!

amickan: Yes, this is just temporary and will be removed in the PR for the feature branch

amickan:
Ok, let's merge and deploy this then. I'll take a look later today at how to run the commands. I thought I needed shell access, hadn't thought of doing this as an ECS task like the migration, but that makes sense.

jmsmkn:
Yes, it is anyway better as the shell has an interactivity timeout, you can follow the logs for progress in cloudwatch and stop the task via the console if necessary.

amickan:
Ok, I might need some help setting up the task properly, but I'll take a look first and create a PR for you to look at. Won't happen until later today though.

PR #4116, opened 2025-06-20
Add page title to doc page navigation

koopmant:
The next and previous buttons shown below a docpage are replace with navigation links showing the target page title. 
The links are also shown above the docpage. 

Example:
image

### DIFF
--- a/app/grandchallenge/documentation/models.py
+++ b/app/grandchallenge/documentation/models.py
@@ -6,6 +6,7 @@ from django.db import models
 from django.db.models import Max
 from django.db.models.signals import post_save
 from django.dispatch import receiver
+from django.utils.functional import cached_property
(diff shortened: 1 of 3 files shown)


jmsmkn (approved):

jmsmkn (changes requested):
Actually, did you check the number of queries with this? I'm not sure if using `queryset.first()` and `queryset.last()` forces going to the database again. If so, quickest thing to do would be to make the properties `cached_property` instead.

koopmant:
> Actually, did you check the number of queries with this? 

I did not, thanks. Will check.

koopmant:
Good catch! The query was repeated 6 times!

jmsmkn (approved):

PR #4117, opened 2025-06-20
Revert "Use window annotation for standard index"

pkcakeout:
Reverts comic/grand-challenge.org#4098 for breaking reader studies.


pkcakeout:
Closing this after a debugging session with @amickan - the underlying issue seems to only affect my account for an unknown reason. I was the first person to cause sentry 500-errors, but in case other accounts are affected as well, I will leave the status-message on grand-challenge status. But as it stands, not all accounts seem to be affected.

PR #4124, opened 2025-06-25
Fix N+1 query in documentation pages

koopmant:
Prefetch children four levels deep for nav menu. 

Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/566


koopmant:
I looked at `documentation_detail.html` to see if we could do with less than four levels, but we need them all.

koopmant:
> Why 4? Is that something to do with our layout? 

Yes, we build up the navigation sidebar everytime we load a docpage and we nest each page under it's parent. We have four levels of nesting. 

> Good idea to add a test here using `assert_max_num_queries`.

Cool, I'll look into it!

koopmant:
> But do we only have 4 because that is what is in production right now?
 
Yes

> Can someone add a 5th without changing the codebase?

No. It's defined in the `docpage_detail.html` template. You can create a fifth level child page, but the nav pane won't handle that: it will not show a navigation element for that page, nor will it expand the parent navigation elements.

koopmant:
I've use `assert_num_queries` instead of `assert_max_num_queries`, because I felt being more specific would be better.

PR #4130, opened 2025-06-30
Avoid creating challenge page with invalid slug

koopmant:
Fixes a bug where some titles pass form validation while still leading to invalid slug `evaluation`.

Reported in https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/577


jmsmkn:
I'm wondering if there is a better way to solve this? I think the new forum urls may also cause a problem. We should allow slugs like this. Is the problem just the update page? Maybe we could shift those urls to:

```python3
    path(
        "pages/<slug>/content-update/",
        PageContentUpdate.as_view(),
        name="content-update",
    ),
    path(
        "pages/<slug>/metadata-update/",
        PageMetadataUpdate.as_view(),
        name="metadata-update",
    ),
    path("pages/<slug>/delete/", PageDelete.as_view(), name="delete"),
```

(note the `pages/` prefix).

koopmant:
Right, of course. 
In that case, the following are safe:

- `evaluation`
  (There is this pattern `path("<slug:slug>/", EvaluationList.as_view(), name="list")`, 
but that is no issue since the subpatterns for pages now all have `pages/`)

- `markdownx`

The following are not:

- "admins/", include("grandchallenge.admins.urls", namespace="admins")
  `path("", AdminsList.as_view(), name="list")`
  propose: `admins/all/`
  There is no subpattern starting with a variable. 

- "invoices/", include("grandchallenge.invoices.urls", namespace="invoices")
  `path("", InvoiceList.as_view(), name="list")`
  propose: `invoices/all/`
  There is no subpattern starting with a variable. 

- "onboarding-tasks/", OnboardingTaskList.as_view(), name="challenge-onboarding-task-list")
  propose: `onboarding-tasks/all`
  There is only `onboarding-tasks/<uuid:pk>`, so no risk. 

- "participants/", include("grandchallenge.participants.urls", namespace="participants")
  `path("", ParticipantsList.as_view(), name="list")`
  propose: `participants/all`
  There is no subpattern starting with a variable. 

- "teams/", include("grandchallenge.teams.urls", namespace="teams")
  `path("", TeamList.as_view(), name="list")`
  propose: `teams/all/`
  There is no subpattern starting with a variable. 

- "update/", ChallengeUpdate.as_view(), name="challenge-update")
  propose: `challenge/update/`

koopmant:
Also had to change the url for `pages/` (`pages/all/`) and `statistics/` (`pages/statistics/` or `challenge/statistics/`?) otherwise we could still clash there.

jmsmkn:
`challenge/statistics/` would make the most sense I think

app/grandchallenge/teams/urls.py
```diff
@@ -13,7 +13,7 @@
 app_name = "teams"
 
 urlpatterns = [
-    path("", TeamList.as_view(), name="list"),
+    path("all/", TeamList.as_view(), name="list"),
```
jmsmkn, line 16: Maybe, I think otherwise there would be a clash with `all` as a page slug.

```suggestion
    path("t/all/", TeamList.as_view(), name="list"),
```

jmsmkn:
Thanks for looking into this, much better solution to the one we had from years ago. Did you work out if it is possible to cover it with a test?

app/grandchallenge/teams/urls.py
koopmant: I don't think `all/` would be an issue, because we add the teams urls as `teams/`. 
I think we could also get rid of all `t/` usages here. Since `teams/1/delete` is much nicer than `teams/t/1/delete/`. For members we could go with `teams/member/<int:pk>/delete`.

koopmant:
> Did you work out if it is possible to cover it with a test?

Good point, I think I threw it out at some point. Let me get it back and extend it with the other page names.

app/grandchallenge/teams/urls.py
jmsmkn: Agreed!

app/tests/pages_tests/test_pages.py
```diff
@@ -524,3 +524,91 @@ def test_should_show_verification_warning():
...
     assert challenge.should_show_verification_warning is False
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize(
+    "slug",
+    [
+        "0",
+        "1",
+        "admins",
```
jmsmkn: Is there a way to automatically generate this parameterisation? I'm thinking of a test that alerts developers when they add a new prefix (e.g. `forums/`). I don't know if this is possible but maybe we can avoid parametrize altogether:

- inspect all of the full paths in the resolver
- if any of them only contain one `/`, and it is not the page detail view, error out?

koopmant: That's smart. 

Added it and tested on main. Works like a charm:
```console
AssertionError: 
These patterns will clash with page urls if challenge admins create a page with the same slug:   
['invoices/', 
'teams/', 
'participants/', 
'admins/', 'update/', 
'onboarding-tasks/', 
'pages/', 
'statistics/']
```

jmsmkn (approved):
Nice work!

PR #4135, opened 2025-07-01
Create specialized docpages for FAQs

koopmant:
Enable the use of docpages for FAQs by adding a toggle to the `DocPage` model.

Docpages marked as FAQ are presented like normal docpages, but the navigation panes for FAQs and normal docpages are independent. When viewing an FAQ page, the navigation pane only shows other FAQ pages, and vice versa. The navigation pane for FAQs is also made wider to make the questions easier to read. 

image

A button is added below the navigation pane to navigate from the documentation to the FAQ and back. 

image

When searching through the documentation, the search results contain both normal docpages and FAQs. 

image


app/grandchallenge/documentation/models.py
```diff
@@ -134,15 +139,19 @@ def get_absolute_url(self):
     @cached_property
     def next(self):
         try:
-            next_page = DocPage.objects.filter(order__gt=self.order).first()
+            next_page = DocPage.objects.filter(
+                order__gt=self.order, is_faq=False
+            ).first()
```
amickan, line 148: For FAQ pages, next and previous will return bogus pages, right? I guess you don't want there to be next and previous pages for FAQ pages, right? Wouldn't it make more sense than to raise a NotImplementedError for this property for those pages?

koopmant: Ah good point. Yes that would be better.

amickan (approved):
LGTM!

PR #4137, opened 2025-07-02
Implement autosaving flag

pkcakeout:
Add `enable_autosaving` flag for the new autosaving feature for reader-studies.


app/grandchallenge/reader_studies/models.py
```diff
@@ -291,6 +291,17 @@ class ReaderStudy(
             "be checked as well to enable this setting."
         ),
     )
+    enable_autosaving = models.BooleanField(
+        default=False,
+        help_text=(
+            "If true, answers to questions are saved in the background while a "
+            "user reads a case. 'Allow Answer Modification' must be "
```
jmsmkn, line 298: Please check that this is the case in form validation and add a test for it.

pkcakeout: Done (similar to other flags)

PR #4139, opened 2025-07-02
Update forum migrations

amickan:
This updates the forum migrations to match the management commands introduced in https://github.com/comic/grand-challenge.org/pull/4114

No need to review this until https://github.com/comic/grand-challenge.org/pull/4114 has been reviewed.

### DIFF
--- a/app/grandchallenge/discussion_forums/migrations/0002_migrate_old_forums.py
+++ b/app/grandchallenge/discussion_forums/migrations/0002_migrate_old_forums.py
@@ -249,9 +249,12 @@ def migrate_challenge_forums(apps, schema_editor):  # noqa C901
         )
 
 
-def migrate_topic_tracks(apps, schema_editor):
+def migrate_topic_tracks(apps, schema_editor):  # noqa: C901
     TopicReadTrack = apps.get_model(  # noqa: N806
         "forum_tracking", "TopicReadTrack"
+    )
+    ForumReadTrack = apps.get_model(  # noqa: N806
+        "forum_tracking", "ForumReadTrack"
     )
     TopicReadRecord = apps.get_model(  # noqa: N806
         "discussion_forums", "TopicReadRecord"
@@ -265,15 +268,18 @@ def migrate_topic_tracks(apps, schema_editor):
     batch_size = 1000
     batch = []
     total_created = 0
-    num_tracks_to_create = (
-        TopicReadTrack.objects.filter(migrated_track__isnull=True)
-        .exclude(user__username=settings.ANONYMOUS_USER_NAME)
-        .count()
-    )
 
-    if num_tracks_to_create == 0:
-        print("No tracks to create")
-        return
+    topic_mapping = {}
+    for old_id in MachinaTopic.objects.values_list("pk", flat=True):
+        try:
+            topic_mapping[old_id] = get_matching_topic(
+                old_topic_id=old_id,
+                old_topic_model=MachinaTopic,
+                new_topic_model=ForumTopic,
+                old_forum_model=MachinaForum,
+            )
+        except ObjectDoesNotExist:
+            continue
 
     for track in (
         TopicReadTrack.objects.filter(migrated_track__isnull=True)
@@ -281,13 +287,8 @@ def migrate_topic_tracks(apps, schema_editor):
         .iterator(chunk_size=1000)
     ):
         try:
-            new_topic = get_matching_topic(
-                old_topic_id=track.topic.pk,
-                old_topic_model=MachinaTopic,
-                new_topic_model=ForumTopic,
-                old_forum_model=MachinaForum,
-            )
-        except ObjectDoesNotExist:
+            new_topic = topic_mapping[track.topic.pk]
+        except KeyError:
             continue
 
         admins_group = new_topic.forum.linked_challenge.admins_group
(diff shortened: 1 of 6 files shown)


app/grandchallenge/core/models.py
```diff
@@ -34,20 +34,6 @@ class Meta:
         abstract = True
 
 
-class UUIDModelNoAutoNow(models.Model):
```
amickan, line 37: I already changed the base classes of the forum models back to `UUIDModel` when rebasing the feature branch. Didn't realize at that point that that change wouldn't be visible, but I hope that's ok.

app/grandchallenge/challenges/models.py
```diff
@@ -411,13 +411,11 @@ class Challenge(ChallengeBase, FieldChangeMixin):
...
         related_name="external_evaluators_of_challenge",
     )
-    forum = deprecate_field(
-        models.OneToOneField(
-            machina_forum_models.Forum,
-            null=True,
-            editable=False,
-            on_delete=models.PROTECT,
-        )
+    forum = models.OneToOneField(
```
amickan, line 414: Can't deprecate the field just yet since the migration and management command reference the forum field.

amickan:
Given the newly added constraint on subdomain url patterns, should we change `/forum/` to `/discussion-forum`?

app/grandchallenge/discussion_forums/urls.py
```diff
@@ -16,7 +16,7 @@
 
 
 urlpatterns = [
-    path("", ForumTopicListView.as_view(), name="topic-list"),
+    path("home/", ForumTopicListView.as_view(), name="topic-list"),
```
jmsmkn: I think this might fit a little better?

```suggestion
    path("topics/", ForumTopicListView.as_view(), name="topic-list"),
```

app/grandchallenge/discussion_forums/migrations/0002_migrate_old_forums.py
```diff
@@ -8,7 +8,168 @@
...
-def migrate_challenge_forums(apps, schema_editor):
+def init_forum_permissions(*, apps, challenge):
+    Permission = apps.get_model("auth", "Permission")  # noqa: N806
+    ForumGroupObjectPermission = apps.get_model(  # noqa: N806
+        "discussion_forums", "ForumGroupObjectPermission"
+    )
+
+    admins_group = challenge.admins_group
+    participants_group = challenge.participants_group
+    view_permission = Permission.objects.get(
```
jmsmkn, line 19: The permissions would only have been created on post migrate. Has that step already been run?

app/grandchallenge/discussion_forums/migrations/0002_migrate_old_forums.py
```diff
@@ -8,7 +8,168 @@
...
+        permission=view_permission,
+    )
+    ForumPostGroupObjectPermission.objects.create(
+        content_object=post,
+        group=admins_group,
+        permission=delete_permission,
+    )
+    ForumPostUserObjectPermission.objects.create(
+        content_object=post,
+        user=post.creator,
```
jmsmkn: What if the creator is none? Same below.

app/grandchallenge/discussion_forums/migrations/0002_migrate_old_forums.py
```diff
@@ -25,45 +186,76 @@ def migrate_challenge_forums(apps, schema_editor):
...
 
-            new_topic.last_post = latest_new_post
-            new_topic.save()
+        print(
+            f"Finished migrating forum, topics and posts for {challenge.short_name}. Migrated {topic_count} topics and {post_count} posts"
+        )
 
 
-def migrate_topic_tracks(apps, schema_editor):
+def migrate_topic_tracks(apps, schema_editor):  # noqa: C901
```
jmsmkn, line 253: Can this handle partial applications of the migration?

app/grandchallenge/discussion_forums/migrations/0002_migrate_old_forums.py
```diff
@@ -25,45 +186,76 @@ def migrate_challenge_forums(apps, schema_editor):
...
+                modified=challenge.forum.updated,
             )
+            challenge.discussion_forum = new_forum
+            challenge.save()
+            init_forum_permissions(apps=apps, challenge=challenge)
+
+        topic_count = 0
+        post_count = 0
+
+        for topic in challenge.forum.topics.all():
```
jmsmkn, line 207: Shouldn't ones that have already been migrated be excluded?

app/grandchallenge/discussion_forums/migrations/0002_migrate_old_forums.py
amickan: Yes, the models were deployed, so the post migrate signal of that migration added the permissions. I also see them in the admin already (created when running the management commands).

app/grandchallenge/discussion_forums/migrations/0002_migrate_old_forums.py
amickan: Good catch, it would fail to create the permission then. I will add a check here and only create those if the creator is not None.

app/grandchallenge/discussion_forums/migrations/0002_migrate_old_forums.py
amickan: No, we need to loop over all topics because a topic might have been migrated but there might be new posts for that topic.

app/grandchallenge/discussion_forums/migrations/0002_migrate_old_forums.py
amickan: What would be a partial migration in this case?

jmsmkn: The management command would have run right, so the data migration would already be in a partial state?

amickan: `TopicReadTrack` which have been migrated are excluded on subsequent runs of this migration. For `ForumReadTrack` we can't match those 1-to-1, so we need to always loop over all, no matter how many times the migration has run, but it is set up with that in mind. So unless you mean something else, I would say, yes, this can handle partial migrations.

jmsmkn: Great!

PR #4147, opened 2025-07-08
Add node environment with jest and some preliminary tests

HarmvZ:
This does not contain the real tests we need to validate our deidentification procedure, just some tests of the file matcher and dummy creation functions. The deidentification tests will follow in a later PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,6 +47,20 @@ jobs:
                   make minio
                   docker compose run --rm celery_worker pytest --durations 10
 
+    javascript-tests:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v4
+            - name: Setup Node.js
+              uses: actions/setup-node@v4
+              with:
+                  node-version: '20'
+                  cache: 'npm'
+            - name: Install dependencies
+              run: npm ci
+            - name: Run tests
+              run: npm test
+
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -0,0 +1,19 @@
...
+    },
+    "keywords": [],
+    "author": "DIAG",
+    "license": "Apache-2.0",
+    "devDependencies": {
+        "jest": "^30.0.4",
+        "jest-environment-jsdom": "^30.0.4"
+    },
+    "dependencies": {
+        "dcmjs": "^0.42.0"
```
jmsmkn, line 17: Is there a way we can keep this in sync with amass, or potentially replace what amass is doing there? Is node_modules already ignored in git?

HarmvZ: node_modules is already ignored indeed. We might be able to replace amass fully with npm but need to research that a bit. I'll look into it but I think it is better to leave that out of this PR.

jmsmkn: Indeed, would be good to consolidate but I think it is tricky. All of the unpkg stuff could be replaced by npm, but not things on cdnjs. The reason I made amass was to prevent having to maintain a JS toolchain as it changes so much (I think yeoman or grunt was _the_ thing when I last looked at it). Nice that this only uses npm.

jmsmkn (approved):
LGTM!

PR #4149, opened 2025-07-09
Fix datatables sorting

### LINKED ISSUE #4131: Columns do not sort correctly
On the challenge stats page:

Image

use string sorting, not value sorting. Same with 

Image

on the cost overview page.

Also seen on the Pages table. Others need to be checked and fixed too.

koopmant:
Revert setting all column data types to string and clean up data-order attributes. 

Closes #4131


jmsmkn (approved):
LGTM!

PR #4152, opened 2025-07-10
Remove the docker executor

jmsmkn:
Replaced with an executor that copies the i/o

### DIFF
--- a/app/tests/evaluation_tests/test_tasks.py
+++ b/app/tests/evaluation_tests/test_tasks.py
@@ -10,7 +10,6 @@ from django.utils.html import format_html
 from redis.exceptions import LockError
 
 from grandchallenge.algorithms.models import Job
-from grandchallenge.components.backends import docker_client
 from grandchallenge.components.models import InterfaceKind
 from grandchallenge.components.tasks import (
     push_container_image,
@@ -172,11 +171,7 @@ def test_method_validation(algorithm_io_image):
     """The validator should set the correct sha256 and set the ready bit."""
     method = MethodFactory(image__from_path=algorithm_io_image)
 
-    image = docker_client.inspect_image(repo_tag="test-algorithm-io:latest")
-    sha256 = image["Id"]
-
-    # The method factory fakes the sha256 on creation
-    assert method.image_sha256 != sha256
+    original_sha256 = method.image_sha256
     assert method.is_manifest_valid is None
     assert method.is_in_registry is False
     assert method.can_execute is False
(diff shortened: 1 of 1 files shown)


app/tests/algorithms_tests/test_tasks.py
```diff
@@ -315,7 +302,7 @@ def test_algorithm(
     )
     heatmap_interface = ComponentInterface.objects.get(slug="generic-overlay")
     interface = AlgorithmInterfaceFactory(
-        inputs=[input_interface],
+        inputs=[input_interface, heatmap_interface],
         outputs=[json_result_interface, heatmap_interface],
     )
```
amickan, line 307: Why this change?

app/tests/algorithms_tests/test_tasks.py
```diff
@@ -273,34 +271,23 @@ def test_jobs_workflow(django_capture_on_commit_callbacks):
     assert len(callbacks) == 2
 
 
-@pytest.mark.flaky(reruns=3)
 @pytest.mark.django_db
-def test_algorithm(
-    algorithm_image, client, settings, django_capture_on_commit_callbacks
-):
+def test_algorithm(client, settings, django_capture_on_commit_callbacks):
```
amickan, line 275: I'm sorry but I don't really understand how this tests works now that we're no longer attaching an algorithm image to the AI here. What is being executed then?

app/tests/algorithms_tests/test_tasks.py
jmsmkn: The new executor just copies the io. The test used a different container that was creating a bunch of stuff, including the deprecated overlay interface. The test does too much. This allows minimal changes to the test by letting the new executor just copy the io.

app/tests/algorithms_tests/test_tasks.py
jmsmkn: Nothing is being executed, the test does not require a container as the executor is not using docker. This is about separation of concerns - the responsiblity of the thing run by the executor is to put files in the output prefix on S3, that is what the new executor does. We decide to do it without using a docker container. We do not need to test docker as that is not used.

amickan: I see it now, when this just calls the execute command on the `IOCopyExecutor` that you added here.

app/tests/algorithms_tests/test_tasks.py
amickan: ahh right, ok, makes sense now!

amickan (approved):

amickan:
I think the development fixtures need to be updated as well, specifically `_create_algorithm_demo`.

PR #4156, opened 2025-07-11
Improve discussion forum admin

amickan:
This makes most fields readonly fields and also adds the `TopicReadRecords` to the admin.


pkcakeout:
(Hope that premumed typo fix was correct?)

chrisvanrun (approved):
LGTM! Just admin read fields, so not much that can break here.

PR #4159, opened 2025-07-11
Clean up algorithm detail page

koopmant:
The following things were done to clean up the algorithm detail page and make it easier to read:

- The "Model Facts" heading is removed and subheadings are upgraded to h3.
- Interfaces are shown as part of the "Mechanism" section
- Challenge performance is shown as part of the "Validation and Performance" section
- Editors are now displayed inline instead of as a list. 
- The last updated datetime has been moved to inline after the version uuid and shows only the date.
- Admin buttons moved to admin section, making title full-width of the container
- Admin section moved to the top, just below title. 
- The algorithm logo image is displayed with fixed height instead of 50% width. 
- Model facts explanation text at bottom is made small with muted text.

Also,
- "Creator(s)" has been replaced with "Editor(s)"
- Editors are now shown in reverse order. This makes the initial creator the first in the list.

### DIFF
--- a/app/grandchallenge/algorithms/views.py
+++ b/app/grandchallenge/algorithms/views.py
@@ -260,6 +260,8 @@ class AlgorithmDetail(ObjectPermissionRequiredMixin, DetailView):
             ).count()
         )
 
+        editors = reversed(self.object.editors_group.user_set.all())
+
         context.update(
             {
                 "form": form,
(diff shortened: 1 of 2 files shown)


app/grandchallenge/algorithms/views.py
```diff
@@ -1220,6 +1223,15 @@ def get_success_url(self):
...
 
+    def get_context_data(self, *args, **kwargs):
+        context = super().get_context_data(*args, **kwargs)
+        context.update(
+            {
+                "algorithm": self.algorithm,
+            }
+        )
+        return context
+
```
koopmant, line 1234: This change has nothing to do with this PR. I just noticed the link in the breadcrumb wasn't working for this view.

koopmant:
Some examples:

image
image

image
image

amickan (approved):
LGTM!

PR #4160, opened 2025-07-14
New challenge forums

amickan:
This implements the new forum functionality as described in https://github.com/DIAGNijmegen/rse-roadmap/issues/407

Should not be merged and deployed until the migrations have been run.


app/grandchallenge/discussion_forums/templates/discussion_forums/forumpost_list.html
```diff
@@ -0,0 +1,122 @@
...
+                {% endif %}
+            </div>
+            <!-- Version for medium screens and up (≥768px) -->
+            <div class="d-none d-md-block col-4 text-right">
+                {% include 'actstream/partials/follow_unfollow_links.html' with object=topic %}
+            </div>
+            <!-- Version for small screens (phones) (<768px) -->
+            <div class="d-block d-md-none col-12 text-left">
+                {% include 'actstream/partials/follow_unfollow_links.html' with object=topic %}
+            </div>
```
jmsmkn: These constructions are a little strange. I think that I would be better to avoid the duplication they bring in. Doesn't `<div class="col-md-4 text-right">` have the same effect?

jmsmkn: This is the first time I recall seeing inline comments like `<!-- Version for medium screens and up (≥768px) -->`. They're in other places too. Is this from an LLM?

amickan: Yes, I did consult an LLM for these. In this case, `col-md-4 text-right` would do the same thing, you're right, but in the other places where I added those, I don't think there is a similarly elegant alternative, but I'll double check.

jmsmkn: I don't think that we need to worry about people using phones, we're 90% desktop. Would rather have a solution without any duplication as that is annoying to keep in sync.

app/grandchallenge/notifications/migrations/0008_migrate_forum_follows_and_notifications.py
```diff
@@ -0,0 +1,115 @@
...
+            follow.object_id = topic.pk
+            follow.content_type = new_topic_ct
+
+        follows_to_update.append(follow)
+
+        if len(follows_to_update) >= batch_size:
+            Follow.objects.bulk_update(
+                follows_to_update,
+                fields=["content_type", "object_id"],
+            )
```
amickan, line 92: In test run 1001 just now, I realized that after we migrate the follows (i.e. updating the content type and object reference of existing follows), new follows of the old content type can be added for the same object id and user (i.e. when a user replies to a topic they were subscribed to). Migrating those in a subsequent run of the migration will fail because the corresponding new follow already exists. The management command and the migration need to be updated to check for whether a follow already exists before adding it to the list to be updated. That makes this migration a bit more complex still but I don't think we have another option.

app/grandchallenge/discussion_forums/templates/discussion_forums/forumpost_list.html
amickan: Ok, I will go through all places where I added this and refactor accordingly.

app/grandchallenge/notifications/migrations/0008_migrate_forum_follows_and_notifications.py
amickan: See https://github.com/comic/grand-challenge.org/pull/4160/commits/bfebfe124d5b3f9511c7e8ceada139e9021945ac for my proposed solution

app/grandchallenge/discussion_forums/templates/discussion_forums/forumpost_list.html
amickan: I've removed or reworked all of them.

amickan:
Sigh. I have noticed one more issue with the current migrations: for the `TopicReadRecords`, since we only loop over them once (we exclude migrated records on subsequent runs), we ignore potential update in their `mark_time` in between migrations. That means that the migrated records might have an outdated `modified` timestamp, which could mean that some topics will be marked as unread for some users while they actually have been read. Changing the migration to fix this would mean always looping over all `TopicReadRecords` and comparing the timestamp of every single one of them to its migrated counter part (if that exists). It will make the migration more intense from a DB-load perspective and more complex in general. I'm happy to work on updating the command (it's now or never), but I am wondering if it is worth the effort. The worst that can happen is that some users will have some topics marked as unread even though they have read them. This can only happen if there was a new post for an existing (migrated) topic and if the user has opened the topic since. I see that there have been in total 14 new posts to existing topics, so the problem is not huge....

jmsmkn (approved):
LGTM!

PR #4166, opened 2025-07-15
Avoid expanding phases navigation if there are too many phases

### LINKED ISSUE #3878: Challenge nav bar ugly with many phases
Look at this list:

Image

Perhaps having a dropdown menu of sorts would help here? Make it responsive as to now show for challengs with less than 5 (?) phases?

koopmant:
closes #3878

### DIFF
--- a/app/grandchallenge/evaluation/templates/evaluation/partials/phase_navbar.html
+++ b/app/grandchallenge/evaluation/templates/evaluation/partials/phase_navbar.html
@@ -13,7 +13,7 @@
         </button>
     </div>
     <div class="collapse navbar-collapse" id="collapsiblePhasesNav">
-        <ul class="nav nav-pills flex-wrap">
+        <ul class="navbar-nav nav-pills flex-wrap">
             {% if request.resolver_match.view_name in 'evaluation:combined-leaderboard-detail,evaluation:leaderboard,evaluation:detail' %}
                 {% for combined_leaderboard in challenge.combinedleaderboard_set.all %}
                     <li class="nav-item">
--- a/app/grandchallenge/evaluation/templates/evaluation/partials/phase_navbar.html
+++ b/app/grandchallenge/evaluation/templates/evaluation/partials/phase_navbar.html
@@ -13,7 +13,7 @@
         </button>
     </div>
     <div class="collapse navbar-collapse" id="collapsiblePhasesNav">
-        <ul class="nav nav-pills">
+        <ul class="nav nav-pills flex-wrap">
             {% if request.resolver_match.view_name in 'evaluation:combined-leaderboard-detail,evaluation:leaderboard,evaluation:detail' %}
                 {% for combined_leaderboard in challenge.combinedleaderboard_set.all %}
                     <li class="nav-item">
--- a/app/grandchallenge/evaluation/templates/evaluation/partials/phase_navbar.html
+++ b/app/grandchallenge/evaluation/templates/evaluation/partials/phase_navbar.html
@@ -2,44 +2,49 @@
 {% load profiles %}
 {% load evaluation_extras %}
 
-<div class="nav-pill-dropdown-container">
-    <a class="d-lg-none btn btn-outline-dark dropdown-toggle phaseButton mr-1 mb-3 px-4"
-       data-toggle="dropdown"
-       href="#"
-       role="button"
-       aria-expanded="false">
-        Choose a Phase</a>
-    <ul class="nav nav-pills col-12 mb-3">
-        {% if request.resolver_match.view_name in 'evaluation:combined-leaderboard-detail,evaluation:leaderboard,evaluation:detail' %}
-            {% for combined_leaderboard in challenge.combinedleaderboard_set.all %}
-                <li class="nav-item">
-                     <a class="nav-link mr-1 px-4 py-1 {% if request.resolver_match.view_name in 'evaluation:combined-leaderboard-detail,evaluation:combined-leaderboard-update,evaluation:combined-leaderboard-delete' and request.resolver_match.kwargs.slug == combined_leaderboard.slug %}active{% endif %}"
-                        href="{% url 'evaluation:combined-leaderboard-detail' challenge_short_name=challenge.short_name slug=combined_leaderboard.slug %}">{{ combined_leaderboard.title|title }}</a>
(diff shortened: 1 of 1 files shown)


app/grandchallenge/challenges/static/js/challenges/dropdown.js
```diff
@@ -1,86 +1,8 @@
...
-        const ulPages = pageContainer.querySelector("ul");
-
-        if (horizPagePills === true) {
-            ulPages.classList.remove(...ulVertPagePillClasses);
-            ulPages.classList.add(...ulHorizPagePillClasses);
-        } else {
-            ulPages.classList.add(...ulVertPagePillClasses);
-            ulPages.classList.remove(...ulHorizPagePillClasses);
-        }
-    }
```
koopmant, line 40: This was used on the page detail page to make nav pills show without flex-column on small screens. I initially replaced it with flex-sm-column, which does exactly the same thing without this js. However, I removed it again because I think the pills should always be shown in column. We currently show it like this, which I think is rather ugly:
Screenshot 2025-07-25 at 14 41 34

app/grandchallenge/pages/templates/pages/page_detail.html
```diff
@@ -22,12 +22,12 @@
 {% endblock %}
 
 {% block sidebar %}
-    <div class="nav-pill-pages-container col-12 col-sm-5 col-md-4 col-lg-3 pl-3">
-        <ul class="nav nav-pills flex-column">
+    <div class="col-xl-5 col-lg-4 col-md-3">
```
koopmant, line 25: Not really part of this work, but I noticed we made the width for the page content extra small on smaller screens. Turned it around now, so the page content width is more or less constant

amickan:
I thought you went width full container widht for the dropdown? Right now, it looks like it is much smaller and for leaderboards, the dropdown is not underneath the Leaderboard tab at all. 
image

koopmant:
Just realized I messed up the normal nav-tabs now:

Screenshot 2025-07-25 at 16 46 05

Fixing it now. (though these original tabs are refreshing in a way)

amickan:
> Just realized I messed up the normal nav-tabs now:
> Screenshot 2025-07-25 at 16 46 05
> 
> Fixing it now. (though these original tabs are refreshing in a way)

Huh? For me it still looks normal.

koopmant:
What happens when you hard reload the page? (Shft + reload)

amickan:
> What happens when you hard reload the page? (Shft + reload)

No longer relevant. I cleared my browser cache and that fixed it

koopmant:
> I thought you went width full container widht for the dropdown? Right now, it looks like it is much smaller and for leaderboards, the dropdown is not underneath the Leaderboard tab at all.

That is really weird. Can you see that the html looks like?

amickan:
> > I thought you went width full container widht for the dropdown? Right now, it looks like it is much smaller and for leaderboards, the dropdown is not underneath the Leaderboard tab at all.
> 
> That is really weird. Can you see that the html looks like?

```html
<div class="dropdown-menu mt-n3 border-top-0 rounded-0 phase-nav-dropdown show">
<nav class="navbar navbar-expand p-0">
    <ul class="navbar-nav nav-pills p-1 flex-wrap">
                <li class="nav-item col-lg-6 px-0">
                        <a class="nav-link pl-4 pr-2 py-1 " href="http://demo.gc.localhost:8000/evaluation/phase-1/submissions/create/">Phase 1</a>
                </li>        
                <li class="nav-item col-lg-6 px-0">
                        <a class="nav-link pl-4 pr-2 py-1 " href="http://demo.gc.localhost:8000/evaluation/phase-2/submissions/create/">Phase 2</a>
                </li>
            <li class="nav-item col-lg-6 px-0">
                <a class="nav-link pl-4 pr-2 py-1 " href="http://demo.gc.localhost:8000/evaluation/submissions/">All  Submissions
                </a>
            </li>

    </ul>
</nav>
</div>
```

koopmant:
Is that inside a `<div class="dropdown">` underneath the `<nav>` that is the challenge-topbar?

amickan:
> Is that inside a `<div class="dropdown">` underneath the `<nav>` that is the challenge-topbar?

Yes

koopmant:
> I messed up the normal nav-tabs

This should be fixed now.

koopmant:
Should be fixed now. 🙈

amickan:
I can confirm that the dropdown is now full width. I think this works, but still noticed a few things that are not optimal IMHO.

- On smaller screens the dropdown is visually separated from the Submit / Leaderboard tab. I find this a bit confusing, and would honestly prefer the current design where you go to the page and then select from a button dropdown... 

image

- On all screen sizes, I noticed that you don't see which tab you clicked on (Submit or Leaderboard) until you've also chosen a phase (then it gets underlined because it's the active tab). Not a huge deal, but confusing is you get distracted for a second, look away and then return to the screen.

Have you played around with visually separating the dropdown from the rest (through a light grey background for example)? If that gets the same color as the nav you've clicked on (Submit / Leaderboard), it might help with the above 2 points. Maybe it's ugly though...  

Finally, you have the dropdown default to 2 columns. For most titles on GC that is a bit "wasteful" space-wise. I quickly double checked and the average phase title has 24 characters, and even with a bit more than that, it's not the most economic use of space:

image

pkcakeout:
Quick clarification, because you showed this in the meeting: Does the dropdown on handheld also work all of a sudden? On the example we revisited, the first item was cut off due to a "bug in bootstrap" as you explained if I remember correctly. That is not a problem anymore?

koopmant:
> Quick clarification, because you showed this in the meeting: Does the dropdown on handheld also work all of a sudden? On the example we revisited, the first item was cut off due to a "bug in bootstrap" as you explained if I remember correctly. That is not a problem anymore?

Yes, that is fixed now. The dropdowns are no longer positioned dynamically, which caused them to start at the top of the page to show as much as possible. Normally you'd have to add `data-display="static"`, except for in navbars. So it got fixed automatically by using the .navbar class. If you're interested: https://getbootstrap.com/docs/4.6/components/dropdowns/#responsive-alignment

amickan (approved):
LGTM!

Nice work. I think the normal dropdown option is a good solution considering all options.

PR #4170, opened 2025-07-17
Add redirects for old forums

amickan:
This adds redirects for the old forum pages to the new ones. The machina urls no longer exist, so I had to construct the original paths manually.

### DIFF
--- a/app/grandchallenge/challenges/templates/challenges/challenge_topbar.html
+++ b/app/grandchallenge/challenges/templates/challenges/challenge_topbar.html
@@ -1,5 +1,6 @@
 {% load url %}
 {% load static %}
+{% load guardian_tags %}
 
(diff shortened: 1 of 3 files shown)


jmsmkn (approved):
LGTM!

jmsmkn (changes requested):
Actually, the tests don't seem to think that there is a Redirects app, not sure why.

jmsmkn:
Hmm, same error. Maybe switch it to a management command?

amickan:
It's a different error `site.get_current()` doesn't work in the migration because it relies on middleware. I quickly checked the pk of the production site, and hardcoded that. Now it should work.

jmsmkn (approved):
LGTM!

PR #4172, opened 2025-07-17
Correctly send notification at job failure

chrisvanrun:
Source: https://github.com/comic/grand-challenge.org/commit/d8af5b8cca7c133c8a652d92b49e437c9ed0fbb0

I think this is a re-name gone wrong. Way old. `task_on_success` should be `task_on_failure`

A sentry issue had this pop-up and I could not figure out why it would fire when the job had success. Now I know!

### DIFF
--- a/app/grandchallenge/algorithms/tasks.py
+++ b/app/grandchallenge/algorithms/tasks.py
@@ -36,11 +36,6 @@ def execute_algorithm_job_for_inputs(*, job_pk):
         app_label="algorithms", model_name="job", pk=job_pk
     )
 
-    # Notify the job creator on failure
-    linked_task = send_failed_job_notification.signature(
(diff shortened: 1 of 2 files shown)


jmsmkn:
See the comments in the review of #2036 - is this still the case? Likely there are others like this.

chrisvanrun:
> Did you check `task_on_failure` too?

Only usage is via:
`task_on_failure = handle_failed_jobs.signature(` Seems correct!

chrisvanrun:
> `send_failed_session_jobs_notifications` was removed in #2780

Should that be reintroduced?

I agree, a test would be appropriate here. I'll whip something together next week.

jmsmkn (approved):
LGTM!

PR #4173, opened 2025-07-17
Ignore conflicts during redirect creation

amickan:
Some redirect paths for the old forums get too long (longer than the max char count of 200) - those should be skipped. Rather than adding a check whether a path would be too long (or whether a redirect already exists, another error source when rerunning the command), I've opted to just run the `bulk_create` with `ignore_conflicts=True`. I think for this one off migration this is fine.

### DIFF
--- a/app/grandchallenge/discussion_forums/management/commands/add_redirects.py
+++ b/app/grandchallenge/discussion_forums/management/commands/add_redirects.py
@@ -2,7 +2,7 @@ from django.contrib.redirects.models import Redirect
 from django.contrib.sites.models import Site
 from django.core.management import BaseCommand
 
-from grandchallenge.discussion_forums.models import Forum, ForumTopic
+from grandchallenge.discussion_forums.models import ForumTopic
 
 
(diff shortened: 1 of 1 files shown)


app/grandchallenge/discussion_forums/management/commands/add_redirects.py
```diff
@@ -15,26 +15,6 @@ def handle(self, *args, **options):
         redirects_to_add = []
         n_created = 0
 
-        for forum in Forum.objects.filter(source_object__isnull=False):
```
amickan, line 18: this can go because the forum redirects were all created successfully.

jmsmkn (approved):

amickan:
`ignore_conflicts` actually does not catch the `DataError`. I had to add a length check instead. `ignore_conflicts` is still necessary to ignore the `IntegrityError` we'll get when rerunning this for the already created topic redirects. I double checked that integrity errors are handled by `ignore_conflicts`.

jmsmkn (approved):

PR #4175, opened 2025-07-18
Replace amass with npm

HarmvZ:


HarmvZ:
> Does grand-challenge get its own webpack/npm build environment now for JavaScript?

That is the idea. We already included npm so we can run javascript unit tests. Now I'll see If I can use npm to build our custom and vendored js.

HarmvZ:
Closing in favor of https://github.com/comic/grand-challenge.org/pull/4178

PR #4176, opened 2025-07-21
Upgrade mc

chrisvanrun:
`mc` is always throwing warnings in my face while pulling up my local dev instance.

The latest update no longer supports the `mc config` but suggests `mc alias`.

This PR updates the `minio.py` script and updates the `mc` used for local testing.

### DIFF
--- a/dockerfiles/web-base/Dockerfile
+++ b/dockerfiles/web-base/Dockerfile
@@ -117,7 +117,7 @@ FROM base AS test-base
 USER root
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
-    # Add java and graphviz for plantuml documentation
+        # Add java and graphviz for plantuml documentation
         default-jre \
         graphviz \
         # make for sphinx docs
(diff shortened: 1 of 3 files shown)


dockerfiles/web-base/Dockerfile
```diff
@@ -61,8 +61,8 @@ RUN mkdir -p /opt/git-lfs \
 
 # Get the minio client for development
 RUN mkdir -p /opt/mc \
-    && wget https://dl.min.io/client/mc/release/linux-amd64/archive/mc.RELEASE.2024-01-28T16-23-14Z -O /opt/mc/mc \
-    && echo "fc195c3ad5c19b91d96bbb42d79af94173aa491ebcda9ca372c4f382546135ec  /opt/mc/mc" | shasum -c - || exit 1 \
+    && wget https://dl.min.io/client/mc/release/darwin-arm64/archive/mc.RELEASE.2025-07-16T15-35-03Z -O /opt/mc/mc \
```
jmsmkn: This is a linux-amd64 based container so cannot use binaries built for darwin-arm64 (macOS M series)

jmsmkn: The reason this isn't caught in CI is that you need to force a rebuild of the containers if the Dockerfile is changed. That can be done by adding a blank line somewhere in `uv.lock`. It should/would get caught in the automatic dependency update each Tuesday.

jmsmkn (approved):
LGTM!

Issue #4163, opened 2025-07-14
Replace amass with node to build js dependencies


HarmvZ:
Since we added a node environment to run javascript tests, we can also use this to replace amass and build our js dependencies. See short discussion here: https://github.com/comic/grand-challenge.org/pull/4147#discussion_r2194221688. It will consolidate installation of js libraries into package.json. We can also use it to bundle and minify our own javascript files.

jmsmkn:
No appetite for this right now.

PR #4178, opened 2025-07-21
PoC for building custom JS with webpack

HarmvZ:
This is a proof of concept for building our JS dependencies and our custom mjs modules as javascript using webpack.
This works by adding any dependency or custom js file here:

https://github.com/comic/grand-challenge.org/blob/eb5a2a4c3dfe6428b5fe4a18c1e5618a89b08bcf/app/webpack.config.js#L10-L12

As you can see this PoC is implemented for jquery and jsoneditor. The rest of our libraries and custom js will follow later.

We use webpack to build and postprocess those files. That means it gets built into a single bundle per entry. They get built and put in the `npm_vendored` directory (as opposed to the `vendored` directory to keep the distinction for now). The custom js that is built is moved from the `static` directory to a `javascript` directory outside of static so that those source files are not served as static files anymore.

In the custom javascript files, any nested dependencies can be imported like this:
https://github.com/comic/grand-challenge.org/blob/eb5a2a4c3dfe6428b5fe4a18c1e5618a89b08bcf/app/grandchallenge/core/javascript/jsoneditor_widget.mjs#L1-L2
Those dependencies will be bundled and treeshaken in the entry file. `webpack-bundle-tracker` is used to generate a json file that describes all the created bundles. `django-webpack-loader` is used to read that json file and allow easy importing of js files in templates like this:

https://github.com/comic/grand-challenge.org/blob/eb5a2a4c3dfe6428b5fe4a18c1e5618a89b08bcf/app/grandchallenge/core/templates/grandchallenge/partials/head_script.html#L13

To include js in widgets, I've added a mixin that allows this:

https://github.com/comic/grand-challenge.org/blob/eb5a2a4c3dfe6428b5fe4a18c1e5618a89b08bcf/app/grandchallenge/core/widgets.py#L8-L10

There is a development and a production build. For development it uses source maps allowing easy debugging. To allow easy debugging in sentry we need to upload our sourcemaps to sentry. That is not integrated yet.

Because we want to be able to build the dev and production environment in the containers, we've added node there. This also required that we move the node environement 1 level deeper into `/app` so that it was not exposed in the root of the container.

Part of https://github.com/comic/grand-challenge.org/issues/4163

Improvement of https://github.com/comic/grand-challenge.org/pull/4175 so that will be closed when this is merged.


jmsmkn:
I think there's a way:

- Run `python manage.py collectstatic` which will put the static files in a single folder
- Serve the files from the folder locally, you could use a separate python process with `python -m http.server` but you might have problems with mixed http and https in the browser, unsure, exercise left to the reader
- Start the server with `DEBUG=False DJANGO_STATIC_HOST=<url from above> python manage.py runserver_plus`

jmsmkn:
I am going to close this as stale, feel free to reopen if you are working on it again.

PR #4183, opened 2025-07-24
Add admin action to cancel image imports

amickan:
Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/601

### DIFF
--- a/app/grandchallenge/components/admin.py
+++ b/app/grandchallenge/components/admin.py
@@ -2,6 +2,7 @@ from django.contrib import admin
 from django.db.transaction import on_commit
 
 from grandchallenge.components.models import (
+    ComponentImage,
     ComponentInterface,
     ComponentInterfaceExampleValue,
     ComponentInterfaceValue,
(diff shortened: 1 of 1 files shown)


app/grandchallenge/components/admin.py
```diff
@@ -11,6 +12,23 @@
...
+)
+def cancel_image_imports(modeladmin, request, queryset):
+    queryset.filter(
+        import_status__in=[
+            ComponentImage.ImportStatusChoices.STARTED,
+            ComponentImage.ImportStatusChoices.QUEUED,
+            ComponentImage.ImportStatusChoices.INITIALIZED,
+            ComponentImage.ImportStatusChoices.RETRY,
+        ]
+    ).select_for_update(of=("self",), skip_locked=True).update(
```
chrisvanrun, line 27: Does select_for_update work when not in a transaction?

amickan: No, it does not, but admin actions are wrapped in transactions (in fact every view is due to `ATOMIC_REQUESTS=True`.

chrisvanrun: Ah right, admin actions are requests too ofc. They have such a different structure I hadn't realized the request still goes through all the middleware.

chrisvanrun (approved):
LGTM!

PR #4185, opened 2025-07-25
Block forms on submit

chrisvanrun:
This PR adds some vanilla JS that effectively disables several forms after submitting. Both the input elements and the actual subsequent submitting is disabled. This is a client-side fix that should be followed-up by a server-side fix to prevent duplicate slugs.

The element `<form>` does not have a way to disable all elements within. Going through a list of possible input elements however, doesn't really sound fool proof either. Luckily, an container within forms: `<fieldset>` does disable and disables all elements withing when it disables. So I've opted to use that.

Since the vector is a much used mixin, I'v tried to hammer down as many corner cases I could come up with and covered it with JS tests.

## Showcase
Here the response is slow, on purpose:

https://github.com/user-attachments/assets/694f6594-c77e-481a-876f-e8f00d6f49e1

Related (private issues):
- https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/107
- https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/558

### DIFF
--- a/app/grandchallenge/core/forms.py
+++ b/app/grandchallenge/core/forms.py
@@ -27,13 +27,13 @@ class SaveFormInitMixin:
             Fieldset(
                 None,  # Legend
                 self.helper.layout,
-                StrictButton(
-                    "Save",
-                    css_class="btn-primary",
-                    type="submit",
-                    css_id="submit-id-save",
-                ),
-            )
+            ),
+            StrictButton(
+                "Save",
+                css_class="btn-primary",
+                type="submit",
+                css_id="submit-id-save",
+            ),
         )
 
     class Media:
--- a/app/grandchallenge/core/forms.py
+++ b/app/grandchallenge/core/forms.py
@@ -16,7 +16,7 @@ class SaveFormInitMixin:
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.helper = FormHelper(self)
-        self.helper.attrs["gc-disable-fieldsets-after-submit"] = True
+        self.helper.attrs["gc-disable-after-submit"] = True
         self.helper.layout = Layout(
             Fieldset(
                 None,  # Legend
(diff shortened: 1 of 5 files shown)


app/grandchallenge/core/static/js/disable_after_submit.mjs
```diff
@@ -0,0 +1,149 @@
...
+const spinnerAttributeName = `${baseAttributeName}-spinner`;
+
+// Holds timers for later cancelation
+const formTimers = new Set();
+
+function initDisableAfterSubmit() {
+    const forms = document.querySelectorAll(`form[${baseAttributeName}]`);
+
+    for (const form of forms) {
+        if (!form.hasAttribute(initAttributeName)) {
```
chrisvanrun, line 30: I thought about using the `.dataset` for elements. I purposefully did not as I thoroughly dislike having to switch between `'data-attr-name'` and `'attrName'` pointers and mentally translate those

app/grandchallenge/core/forms.py
```diff
@@ -1,19 +1,48 @@
...
+                css_id="submit-id-save",
+            ),
+        )
+
+    class Media:
+        js = [
+            format_html(
+                '<script type="module" src="{}"></script>',
+                static("js/disable_after_submit.mjs"),
+            )
```
HarmvZ: Why did you use this? You should be able to do the following, right?

```suggestion
            "js/disable_after_submit.mjs",
```

app/grandchallenge/core/forms.py
```diff
@@ -10,10 +11,31 @@
...
+            StrictButton(
+                "Save",
+                css_class="btn-primary",
+                type="submit",
+                css_id="submit-id-save",
+            ),
+        )
+
+    class Media:
+        js = ["js/disable_after_submit.mjs"]
```
jmsmkn, line 38: We might need to check that inheritance of js media works correctly here.

app/grandchallenge/core/forms.py
chrisvanrun: AFAIK usting the regular notation would result in a regular `<script>` and not a module import. T.b.h. that was an assumption and I didn't check it. I guess it worked out?

PR #4193, opened 2025-07-29
Replace datatable initialization and configuration javascript with data attributes

HarmvZ:
While working on the javascript conversion I notices lot's of simple javascript files that contained only a few lines for initialization of DataTable. In the new setup these would all require separate endpoints to be built by npm, but that is quite a big overhaul for such a trivial feature. Therefore, I decided to create a helper function that initializes and configures DataTable when certain data attributes are present on the table element. This allows for easy configuration without needing to write JavaScript. To enable this add the `data-data-table` attribute on the table element. Then other attributes prefixed with `data-dt-` can be used to configure DataTable.

 Example:
 ```html
  <table data-data-table data-dt-page-length="50" data-dt-order='[[0, "asc"]]'>
 ```


app/grandchallenge/documentation/templates/documentation/docpage_list.html
```diff
@@ -23,7 +23,7 @@ <h2>Documentation Pages</h2>
         {% endif %}
     </div>
     <div class="table-responsive">
-        <table class="table table-hover table-borderless table-sm w-100" id="pagesTable" data-reorderable-rows="true">
```
amickan, line 26: `data-reorderable-rows` did nothing?

amickan (approved):
LGTM!

app/grandchallenge/documentation/templates/documentation/docpage_list.html
HarmvZ: I don't think so. It's a bootstrap extensions that requires some dependencies that we do not include anywhere. I suspect it's an old leftover. I think it should allow reordering of rows by dragging them, but I tested it in production and it does not seem possible in that table: https://grand-challenge.org/documentation/overview/.

amickan: Ok, save to remove then

PR #4194, opened 2025-07-29
Remove evaluations list from method detail page

### LINKED ISSUE #3934: Method detail page should not list evaluations
Takes too long to load for many entries.

koopmant:
Closes #3934

### DIFF
--- a/app/grandchallenge/evaluation/templates/evaluation/method_detail.html
+++ b/app/grandchallenge/evaluation/templates/evaluation/method_detail.html
@@ -117,8 +117,6 @@
         <dd>{{ object.comment }}</dd>
     </dl>
 
-    <h2>Evaluations for this Evaluation Method</h2>
-    {% include 'evaluation/partials/evaluations_for_object_table.html' %}
 {% endblock %}
 
 {% block script %}


amickan:
LGTM! Another option might be to load this asynchronously and only on request from the user (i.e. have a button that loads the table via HTMX or a button that redirects to a separate view). I have found this table handy in the past, although only to debug mysterious submission failures, so possibly I'm the only one who has used it.

amickan (dismissed):

app/grandchallenge/evaluation/views/__init__.py
```diff
@@ -303,6 +303,18 @@ class MethodImportStatusDetail(
     login_url = reverse_lazy("account_login")
 
 
+class MethodEvaluationList(
+    LoginRequiredMixin,
+    ObjectPermissionRequiredMixin,
+    DetailView,
+):
+    permission_required = "view_method"
```
amickan: I think to be 100% safe this needs to also check the permission on the evaluations that are shown by the view. It would then be a `ListView` for the `Evaluation` model with a combination of `ViewObjectPermissionListMixin` and `ObjectPermissionRequiredMixin`. We use this pattern, for example, on the `InvoiceList`.

koopmant: Cool, thanks! I was wondering if I had to do something like that here.

koopmant: Should it be a `PaginatedTableListView`?

amickan: That would probably make sense, yes.

koopmant:
Just pushed the last commit as I finally had it working last week. 

> We can always add a button if we find that we need it again.

When we do, see last commit. 

Will revert this to just removing it now, as discussed in the RSE meeting today.

jmsmkn (approved):
LGTM!

PR #4199, opened 2025-07-31
Improve whitespace placement for pages

### LINKED ISSUE #4198: Sidebar on challenge pages has doubled in width
See the panther challenge for example:

Image

This is the case on all challenge info pages on the largest responsive size. This is likely not intended?

koopmant:
Navigation pills for pages were very wide, they are now a relative size. 
Page content width is controlled separately and whitespace is placed around the content. This moves the page content from the right of the screen more to the center, on large screens. 

Closes #4198


amickan (approved):
LGTM

PR #4201, opened 2025-08-01
Restrict sockets for archive uploads

### LINKED ISSUE #4151: Values for RESERVED_SOCKET_SLUGS should not be allowed to be added to archives
We define several reserved sockets:

https://github.com/comic/grand-challenge.org/blob/19cbf3abfccb087b2d796d670d32ebeaa85213ba/app/grandchallenge/algorithms/forms.py#L206

These are not allowed to be added to the interfaces of algorithms, so should not be allowed to be added to archives too. This is to prevent incorrect setup of challenge archives. This should not be possible to add via both the templates and the REST API.

amickan:
Closes #4151

### DIFF
--- a/app/grandchallenge/components/models.py
+++ b/app/grandchallenge/components/models.py
@@ -2440,9 +2440,11 @@ class CIVForObjectMixin:
                 f"{self} is not editable. CIVs cannot be added or removed from it.",
             )
 
-        base_object = self.base_object
         try:
-            if civ_data.interface_slug not in base_object.allowed_socket_slugs:
+            if (
+                civ_data.interface_slug
+                not in self.base_object.allowed_socket_slugs
+            ):
                 raise CINotAllowedException(
                     f"Socket {civ_data.interface_slug!r} is not allowed "
                     f"for this {self.base_object._meta.model_name}."


koopmant:
This code is making my head spin. Not the new code added here, but the way we implemented autocomplete. For example:

We forward some information in the htmx,

https://github.com/comic/grand-challenge.org/blob/538900e7c35fcf48f43aa47f4fce44cfdf9507a0/app/grandchallenge/components/forms.py#L340

Which is later put in the `dal-forward-conf script` But we then override that with some js:

https://github.com/comic/grand-challenge.org/blob/538900e7c35fcf48f43aa47f4fce44cfdf9507a0/app/grandchallenge/components/static/components/js/autocomplete_htmx.mjs#L36

Something else:

https://github.com/comic/grand-challenge.org/blob/538900e7c35fcf48f43aa47f4fce44cfdf9507a0/app/grandchallenge/components/templates/components/civ_set_form.html#L37

This causes all hidden inputs to be included in the htmx when we click "Add value for new socket". Initially, there is nothing there, but when we add more sockets, any hidden input on the previous socket widgets is now included for our next socket.

It's taking me quite some time to understand what the code is doing.

PR #4202, opened 2025-08-01
Build more JS dependencies with npm

HarmvZ:
This is a next step after https://github.com/comic/grand-challenge.org/pull/4178 to build more dependencies with webpack:

- bootstrap
- datatables
- floating-scroll
- sentry
- vega

These were all slightly intertwined so I added them all together in this PR. Besides those there are some extra fixes and additions in here:

- Clears the `npm_vendored` directory. This directory contains the vendored npm files and is now created on the fly during the docker build. This will also be done when running `make runserver` for development. **This means that `node` and `npm` are now required to run the dev environment.**
- Fixes django-tests CI workflow
- Adds more tests for datatables setup
- Adds a chapter to the developer documentation on how to develop and build the JS dependencies


jmsmkn:
Did something go wrong with merging main here? I see some strange commits? Maybe they also need to be applied to the target branch.

PR #4204, opened 2025-08-04
Revert pages width change

### LINKED ISSUE #4198: Sidebar on challenge pages has doubled in width

koopmant:
Revert pages width to before #4166. Changes compared to then are: 

- the page pills now shrink with the screen size, no longer exacerbating the problem of reduced space available for the page content
- page content breaks to the next row on screens smaller than `md` rather then `sm`
- the right margin for the page content is removed and is only shown on the left on screens of `lg` or wider.

Screenshot 2025-08-04 at 22 41 38

Closes #4198

### DIFF
--- a/app/grandchallenge/core/templates/base.html
+++ b/app/grandchallenge/core/templates/base.html
@@ -90,7 +90,7 @@
                 {% endblock %}
                 <div class="row">
                     {% block sidebar %}{% endblock %}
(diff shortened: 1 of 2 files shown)


jmsmkn:
See also #4199

PR #4211, opened 2025-08-06
Update utilisation chart

jmsmkn:
Now that we offer machines with 192 CPUs the memory utilisation was difficult to see. Instead we now scale the chart from 0-100%, noting the single thread limit.

Before:
Screenshot 2025-08-05 at 17 46 33

After:
Screenshot 2025-08-05 at 17 46 18

### DIFF
--- a/app/grandchallenge/charts/specs.py
+++ b/app/grandchallenge/charts/specs.py
@@ -190,7 +190,7 @@ def world_map(*, values):
     }
 
 
-def components_line(*, values, title, cpu_limit, tooltip):
+def components_line(*, values, title, single_thread_limit, tooltip):
     return {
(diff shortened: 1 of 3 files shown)


koopmant:
Great improvement! Even with 4 CPU's the impact on the chart is huge. 

Not sure how useful the indication of the single CPU thread is? With few CPU's it's easy to calculate and with many you won't really see it. I'd remove it entirely.

jmsmkn:
I'm thinking for less experienced users here. It's easy to calculate but might not be easy to interpret for them who may not be able to understand why their CPU utilisation is pinned at some, seemingly arbitrary for them, percentage.

PR #4227, opened 2025-08-08
Equalize component height in ground truth list table

### LINKED ISSUE #3936: Chunk missing in active cell of method list table when hovering
On safari:

Image

That white tab at the bottom of the active cell should not be there.

koopmant:
Closes #3936


jmsmkn (approved):
LGTM!

PR #4228, opened 2025-08-11
Restrict sockets for archive uploads

koopmant:
Add an `allowed_socket_slugs` property to limit the sockets if an archive is linked to a phase. Then only the sockets defined on the phase's algorithm input interfaces can be added.

Related to #4151

### DIFF
--- a/app/grandchallenge/components/views.py
+++ b/app/grandchallenge/components/views.py
@@ -115,6 +115,8 @@ class ComponentInterfaceAutocomplete(
         if self.forwarded:
             object_slug = self.forwarded.pop("object")
             model_name = self.forwarded.pop("model")
+            image_only = self.forwarded.pop("image_only")
+
             if model_name == ReaderStudy._meta.model_name:
                 object = ReaderStudy.objects.get(slug=object_slug)
             elif model_name == Archive._meta.model_name:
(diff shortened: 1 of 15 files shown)


app/grandchallenge/components/views.py
```diff
@@ -112,24 +112,36 @@ class ComponentInterfaceAutocomplete(
...
-            else:
-                raise RuntimeError(
-                    f"Autocomplete for objects of type {model_name} not defined."
-                )
-            qs = ComponentInterface.objects.exclude(
-                slug__in=object.values_for_interfaces.keys()
-            ).exclude(pk__in=self.forwarded.values())
+        object_slug = self.forwarded.pop("object_slug")
+        model_name = self.forwarded.pop("model_name")
+        image_only = self.forwarded.pop("image_only", False)
```
amickan, line 117: Are these keys always present now?

koopmant: Yes!

koopmant:
Yes, I was so focused on the second point I forgot about the first point, I think I thought it solved the first point automatically.

koopmant:
Updated the description. @amickan would you mind checking it for me?

PR #4229, opened 2025-08-11
Show initial sort order in MethodList table

### LINKED ISSUE #3935: Method list page has no ordering
Should be sorted by creation date.

koopmant:
Closes #3935

### DIFF
--- a/app/tests/evaluation_tests/test_views.py
+++ b/app/tests/evaluation_tests/test_views.py
@@ -534,11 +534,6 @@ class TestObjectPermissionRequiredViews:
         group.user_set.add(u)
 
         for view_name, kwargs, permission, obj in [
-            (
-                {"slug": e.submission.phase.slug},
-                "view_method",
-                m,
-            ),
             ("submission-list", {}, "view_submission", s),
         ]:
             assign_perm(permission, group, obj)


app/grandchallenge/datatables/templates/datatables/list_challenge_settings_base.html
```diff
@@ -0,0 +1,16 @@
+{% extends "pages/challenge_settings_base.html" %}
+{% load static %}
+
+{% block content %}
+    {% include "datatables/partials/datatable.html" with columns=columns %}
```
jmsmkn, line 5: Does this work with pagination? I think that is forgotten sometimes.

app/grandchallenge/evaluation/views/__init__.py
```diff
@@ -268,11 +268,24 @@ class MethodList(
     LoginRequiredMixin,
     ViewObjectPermissionListMixin,
     CachedPhaseMixin,
-    ListView,
+    PaginatedTableListView,
```
jmsmkn, line 272: Optional but maybe better for security - now that `ViewObjectPermissionListMixin` works with object based permissions too we could check that the user also has change permissions on the challenge.

app/grandchallenge/datatables/templates/datatables/list_challenge_settings_base.html
koopmant: Yes, this `partials/datatable.html` template initially only loads the table header row. `js/datatables/list.mjs` then creates the request to load the data afterwards with `pageLength: 25`.

app/grandchallenge/evaluation/views/__init__.py
koopmant: Added it.

jmsmkn (approved):
LGTM! I made a slight change unifying the permissions on the phase, didn't realise that we had those.

jmsmkn (approved):

PR #4230, opened 2025-08-11
Replace EvaluationList with EvaluationAdminList

### LINKED ISSUE #4225: Error when compressing files
```
RUN python manage.py collectstatic --noinput && python manage.py compress --force

Compressing... Error parsing template evaluation/evaluation_list.html: Invalid template name in 'extends' tag: ''. Got this from the 'base_template' variable.
```

koopmant:
Closes #4225

### DIFF
--- a/app/tests/evaluation_tests/test_views.py
+++ b/app/tests/evaluation_tests/test_views.py
@@ -248,7 +248,7 @@ class TestLoginViews:
                 "submission-detail",
                 {"pk": e.submission.pk, "slug": e.submission.phase.slug},
             ),
-            ("list", {"slug": e.submission.phase.slug}),
+            ("evaluation-admin-list", {"slug": e.submission.phase.slug}),
             ("update", {"pk": e.pk}),
         ]:
             response = get_view_for_user(


app/grandchallenge/evaluation/views/__init__.py
```diff
@@ -912,7 +870,7 @@ class EvaluationUpdate(
     def get_success_url(self):
         next = self.request.GET.get("next")
         admin_list_url = reverse(
-            "evaluation:evaluation-admin-list",
+            "evaluation:list",
```
jmsmkn: I think I would prefer to keep the name Evaluation Admin and Evaluation Admin List. This is intended for admins only so that makes it clear to future developers what the permission should be. We have an evaluation list view for other users but call that leaderboard detail.

koopmant: All right. I was unsure what to go for and went for removing `Admin` as it reminded me of the django admin. But keeping it like this also makes it easier to backtrack in the future.

app/grandchallenge/evaluation/urls.py
```diff
@@ -84,9 +83,8 @@
         CombinedLeaderboardDelete.as_view(),
         name="combined-leaderboard-delete",
     ),
-    path("<slug:slug>/", EvaluationList.as_view(), name="list"),
     path(
-        "<slug:slug>/admin/",
```
koopmant, line 89: I've still removed admin from the pattern here, since we don't use it anywhere else and it is no longer needed now. I guess this is fine?

app/grandchallenge/evaluation/urls.py
```diff
@@ -84,9 +83,8 @@
...
+        "<slug:slug>/",
```
jmsmkn: Let's keep it at admin.

```suggestion
        "<slug:slug>/admin/",
```

app/grandchallenge/evaluation/views/__init__.py
jmsmkn: This should fix the tests:

```suggestion
            "evaluation:evaluation-admin-list",
```

koopmant: Oops, thanks.

jmsmkn (approved):

PR #4236, opened 2025-08-12
Complete list of views for test_login_redirect

koopmant:
Added missing views from test_login_redirect and fixed views with incorrect redirect.


jmsmkn (approved):
LGTM!

PR #4238, opened 2025-08-13
Use server side pagination for EvaluationAdminList

### LINKED ISSUE #3933: `/evaluation/{slug}/admin/` list view should use server side pagination
Very slow for a large number of entries right now. Tricky bit could be the interactions for hiding/unhiding evaluations.

koopmant:
Closes #3933

### DIFF
--- a/app/grandchallenge/datatables/views.py
+++ b/app/grandchallenge/datatables/views.py
@@ -28,15 +28,19 @@ class PaginatedTableListView(ListView):
         )
         return context
 
-    def render_row(self, *, object_, page_context):
+    def render_row(self, *, object_, page_context, request):
         return render_to_string(
-            self.row_template, context={**page_context, "object": object_}
+            self.row_template,
+            context={**page_context, "object": object_},
+            request=request,
         ).split("<split></split>")
 
-    def render_rows(self, *, object_list):
+    def render_rows(self, *, object_list, request):
         page_context = self.get_context_data(object_list=object_list)
         return [
-            self.render_row(object_=o, page_context=page_context)
+            self.render_row(
+                object_=o, page_context=page_context, request=request
+            )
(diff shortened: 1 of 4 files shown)


app/grandchallenge/datatables/views.py
```diff
@@ -28,15 +28,19 @@ def get_context_data(self, *, object_list=None, **kwargs):
         )
         return context
 
-    def render_row(self, *, object_, page_context):
+    def render_row(self, *, object_, page_context, request):
         return render_to_string(
-            self.row_template, context={**page_context, "object": object_}
+            self.row_template,
+            context={**page_context, "object": object_},
+            request=request,
```
jmsmkn, line 35: Why is the request now necessary here?

app/grandchallenge/evaluation/templates/evaluation/evaluation_list_row.html
jmsmkn, line 1: Might be worth changing the template names to include `admin` now that all of the permission checks are gone (which seems fine)

app/grandchallenge/datatables/views.py
koopmant: That is to get the csrf token to work. I got the following in the console:

```
A {% csrf_token %} was used in a template, but the context did not provide the value. 
This is usually caused by not using RequestContext
```

By supplying the request, `RequestContext` is automatically created. (Specifically in `django.template.backends.django.Template.render` which calls `django.template.context.make_context`)

app/grandchallenge/evaluation/templates/evaluation/evaluation_list_row.html
koopmant: Good point! I think that is also the last name the said evaluation list instead of evaluation admin list.

PR #4241, opened 2025-08-13
Add explicit accepting of each policy

### LINKED ISSUE #4234: Registration form should have a tick box to confirm that this is the users only account
"I have read and agree to the Terms of Service and the Privacy.*" is also awkwardly worded.

jmsmkn:
Get user to confirm only account

Closes #4234

### DIFF
--- a/app/grandchallenge/profiles/forms.py
+++ b/app/grandchallenge/profiles/forms.py
@@ -3,11 +3,10 @@ from crispy_forms.layout import Submit
 from django import forms
 from django.core.exceptions import ValidationError
 from django.forms import CheckboxInput, Select
-from django.utils.safestring import mark_safe
+from django.utils.html import format_html
 from django.utils.translation import gettext_lazy as _
 
 from grandchallenge.core.forms import SaveFormInitMixin
-from grandchallenge.core.templatetags.remove_whitespace import oxford_comma
 from grandchallenge.policies.models import Policy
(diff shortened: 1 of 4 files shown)


app/grandchallenge/profiles/forms.py
```diff
@@ -82,21 +81,26 @@ def clean(self):
...
+
+        for idx, policy in enumerate(Policy.objects.all()):
+            self.fields[f"accept_policy_{idx}"] = forms.BooleanField(
+                required=True,
+                label=format_html(
+                    "I have read and agree to the <a href={absolute_url}>{title}</a> policy",
+                    absolute_url=policy.get_absolute_url(),
+                    title=policy.title,
+                ),
             )
```
koopmant, line 103: Suggestion to avoid multiple checkboxes. This will read "I have read and agree to the the Terms of Service and the Privacy policy". 

```suggestion
self.fields["accept_policies"] = forms.BooleanField(
            required=True,
            label=mark_safe(
                f'I have read and agree to the {oxford_comma([format_html("the <a href={url!r}>{title}</a>", url=policy.get_absolute_url(), title=policy.title) for policy in Policy.objects.all()])} policy'
            ),
        )
```

jmsmkn: I wanted to remove the use of `mark_safe` as it is dangerous.

PR #4246, opened 2025-08-13
Complete list of copy fields for Question

### LINKED ISSUE #4244: Reader study setting for "Empty answer confirmation" not copied
User reported that the following setting was not copied when they copied a reader study.

Image

koopmant:
Closes #4244

### DIFF
--- a/app/grandchallenge/reader_studies/models.py
+++ b/app/grandchallenge/reader_studies/models.py
@@ -1443,6 +1443,31 @@ class Question(UUIDModel, OverlaySegmentsMixin):
             )
         ]
 
+    copy_fields = [
+        "question_text",
+        "help_text",
+        "answer_type",
+        "image_port",
+        "default_annotation_color",
+        "required",
+        "direction",
+        "scoring_function",
+        "order",
+        "interface",
+        "look_up_table",
+        "overlay_segments",
+        "widget",
+        "interactive_algorithm",
+        "answer_max_value",
+        "answer_min_value",
+        "answer_step_size",
+        "answer_min_length",
+        "answer_max_length",
+        "answer_match_pattern",
+        "empty_answer_confirmation",
+        "empty_answer_confirmation_label",
+    ]
+
     def __str__(self):
         return (
             f"{self.question_text} "
--- a/app/grandchallenge/reader_studies/views.py
+++ b/app/grandchallenge/reader_studies/views.py
@@ -632,28 +632,10 @@ class ReaderStudyCopy(
             for question in reader_study.questions.all():
                 q = Question.objects.create(
                     reader_study=rs,
-                    question_text=question.question_text,
-                    help_text=question.help_text,
-                    answer_type=question.answer_type,
-                    image_port=question.image_port,
-                    default_annotation_color=question.default_annotation_color,
-                    required=question.required,
-                    direction=question.direction,
-                    scoring_function=question.scoring_function,
-                    order=question.order,
-                    interface=question.interface,
-                    look_up_table=question.look_up_table,
-                    overlay_segments=question.overlay_segments,
-                    widget=question.widget,
-                    interactive_algorithm=question.interactive_algorithm,
-                    answer_max_value=question.answer_max_value,
-                    answer_min_value=question.answer_min_value,
-                    answer_step_size=question.answer_step_size,
-                    answer_min_length=question.answer_min_length,
-                    answer_max_length=question.answer_max_length,
-                    answer_match_pattern=question.answer_match_pattern,
-                    empty_answer_confirmation=question.empty_answer_confirmation,
-                    empty_answer_confirmation_label=question.empty_answer_confirmation_label,
+                    **{
+                        field: getattr(question, field)
+                        for field in Question.copy_fields
+                    },
                 )
                 for option in question.options.all():
                     CategoricalOption.objects.create(
--- a/app/grandchallenge/reader_studies/views.py
+++ b/app/grandchallenge/reader_studies/views.py
@@ -636,6 +636,7 @@ class ReaderStudyCopy(
                     help_text=question.help_text,
                     answer_type=question.answer_type,
                     image_port=question.image_port,
+                    default_annotation_color=question.default_annotation_color,
                     required=question.required,
                     direction=question.direction,
                     scoring_function=question.scoring_function,
@@ -644,12 +645,15 @@ class ReaderStudyCopy(
                     look_up_table=question.look_up_table,
                     overlay_segments=question.overlay_segments,
                     widget=question.widget,
+                    interactive_algorithm=question.interactive_algorithm,
                     answer_max_value=question.answer_max_value,
                     answer_min_value=question.answer_min_value,
                     answer_step_size=question.answer_step_size,
                     answer_min_length=question.answer_min_length,
                     answer_max_length=question.answer_max_length,
                     answer_match_pattern=question.answer_match_pattern,
+                    empty_answer_confirmation=question.empty_answer_confirmation,
+                    empty_answer_confirmation_label=question.empty_answer_confirmation_label,
                 )
                 for option in question.options.all():
                     CategoricalOption.objects.create(


app/grandchallenge/reader_studies/models.py
```diff
@@ -1443,6 +1443,31 @@ class Meta:
             )
         ]
 
+    copy_fields = [
+        "question_text",
```
jmsmkn, line 1464: How do we keep this list up to date? If we don't have a way of checking this then we will get the same problem as developers add new fields in the future.

koopmant: Initially I thought of doing something like:

```python

@property
def copy_fields(self):
    return [f.name for f in self._meta.get_fields() if f.name not in self.non_copy_fields]
```

But that just moves the issue to keeping a list up to date that defines the fields that should not be copied. 

Maybe we can define both lists and add a test that all fields are covered in both lists? Then when adding a new field you have to explicitly add it to either the copy or list or the non-copy list.

app/grandchallenge/reader_studies/models.py
```diff
@@ -365,7 +365,35 @@ class Meta(UUIDModel.Meta, TitleSlugDescriptionModel.Meta):
...
+        "id",
+        "created",
+        "modified",
+        "view_content",
+        "hanging_protocol",
+        "editors_group",
+        "readers_group",
+        "max_credits",
+        "workstation_sessions",
+        "case_text",
```
koopmant: I'd expected`case_text` should be copied, but in `test_reader_study_copy` we assert it is not.

koopmant: It's conditional, never mind.

app/grandchallenge/reader_studies/models.py
```diff
@@ -365,7 +365,35 @@ class Meta(UUIDModel.Meta, TitleSlugDescriptionModel.Meta):
             ("view_leaderboard", "Can view leaderboard"),
         ]
 
-    copy_fields = (
+    _non_copy_fields = {
```
koopmant: Some fields in this list are copied if the user indicates it on the form. So "non_copy_fields" might be a bit of a misnomer.

jmsmkn: Maybe 3 sets - always, optional and never? Another test could then be added to ensure the optional ones are on the form.

app/grandchallenge/reader_studies/models.py
```diff
@@ -365,7 +365,38 @@ class Meta(UUIDModel.Meta, TitleSlugDescriptionModel.Meta):
...
+        "created",
+        "modified",
+        "max_credits",
+        "workstation_sessions",
+        "actor_actions",
+        "target_actions",
+        "action_object_actions",
+    }
+
+    _optional_copy_fields = {
```
jmsmkn: Is this used anywhere other than the tests? I would expect to see it handled in code.

koopmant: No, it's just there for the test. Same with `_non_copy_fields`. 
I thought about handling it, but for example for the readers and editors we use `reader_study.add_reader()`, instead of `reader_study.readers_group.add()`. I didn't bother looking into it further, the lists are just there so we don't forget to define what should be copied and what shouldn't when we add new fields.

jmsmkn: But then what is the point of having the separate lists? Doesn't the problem just get pushed to the other locations in the codebase and the developer is unaware?

koopmant: I wouldn't think so. If you now add a new field, the test will tell you it is not defined in any of the lists. If you add it to the copy list, it will be automatically copied. If you add it to any of the other lists, you indicate it is handled separately (either it should not be copied, or we first check the form whether to copy or not). I only separated non_copy and optional_copy for the sake of clarity.

koopmant: Though I admit, optional_copy_fields sounds like it will be handled automatically...

koopmant: I think we can add another test to check if all copy and optional fields are handled.

jmsmkn: Couldn't you generate the form fields from this list?

jmsmkn: It would also be nicer to move the never copyable fields to the test if they are not being used in the application code.

koopmant: > Couldn't you generate the form fields from this list?
> It would also be nicer to move the never copyable fields to the test if they are not being used in the application code.

Good points, thanks.

app/tests/reader_studies_tests/test_forms.py
```diff
@@ -548,312 +557,450 @@ def test_image_port_only_with_bounding_box(
...
-    rs.hanging_protocol = HangingProtocolFactory()
-    rs.case_text = {im1.name: "test", im2.name: "test2"}
-    rs.workstation_config = WorkstationConfigFactory()
-    rs.save()
+    assert {f.name for f in model._meta.get_fields()} == union
 
-    assert ReaderStudy.objects.count() == 1
 
-    response = get_view_for_user(
+question_copy_fields = Question.copy_fields.union({"options"})
```
koopmant, line 608: `options` values are copied, but are not in `Question.copy_fields` because they are handled differently. This creates a difference in what we consider "copy fields" in the model (fields we can copy in the same way) and "copy fields" in the test (fields that are copied, no matter how).

app/tests/reader_studies_tests/test_forms.py
```diff
@@ -548,312 +557,450 @@ def test_image_port_only_with_bounding_box(
...
+        ):
+            # cannot test if there can be no value
+            continue
 
-    assert copied_question.options.get().title == question.options.get().title
-    assert copied_question.options.get().pk != question.options.get().pk
+        if field_name in [
+            "actor_actions",
+            "target_actions",
+            "action_object_actions",
```
koopmant: I could not figure out how to generate these. Without a value, we cannot assert they are not copied.

app/tests/reader_studies_tests/test_forms.py
```diff
@@ -548,312 +557,450 @@ def test_image_port_only_with_bounding_box(
...
-        "id",
-        "created",
-        "modified",
-        "options",
-    ]
-    for field in copied_question._meta.get_fields():
-        if field.name not in non_copied_fields:
-            assert getattr(copied_question, field.name) == getattr(
-                question, field.name
-            )
```
koopmant, line 721: This assert was not working for new fields, because we would not give them a value in the test.

app/tests/reader_studies_tests/test_forms.py
jmsmkn: Where do those fields even come from? I have no idea what they are.

koopmant: They were created after adding `ReaderStudy` to the actstream registry:
https://github.com/comic/grand-challenge.org/blob/e3d15e1a9f9e8d8b6fe08af6a228043558777420/app/grandchallenge/notifications/apps.py#L45

I can create some [actions](https://django-activity-stream.readthedocs.io/en/latest/actions.html) manually. But I'm still unsure how these are usually created.

app/grandchallenge/reader_studies/models.py
pkcakeout: Keeping the lists up to date is indeed a hassle. I ran into this myself when I was adding an option for "background saving" and found that you had to add that option to 3-4 separate lists, manually.

My 2 cents: I think a custom decorator would be great but this does not work:

```
class TheModel(...):
    @copy_on_clone
    copy_this_bool_field = BoolField(...)

    do_not_copy_this_field = BoolField(...)
```

Decorators can only decorate functions, not assignments. Therefore, a working substitute would be:

```.py
class TheModel(...):
    fields_to_copy = []

    fields_to_copy.append("copy_this_bool_field")
    copy_this_bool_field = BoolField(...)

    do_not_copy_this_field = BoolField(...)

    fields_to_copy = tuple(fields_to_copy) # Make the list immutable at the end
```

You then get a TheModel.fields_to_copy with fieldnames that should be copied. This is minimal fanciness (no metaclasses or `__set_name__` magic) giving us a bit of syntactic sugar which allows us to annotate what fields to copy, right there in the model.

Maybe?

koopmant: > found that you had to add that option to 3-4 separate lists

These lists all mean something different. The model could define all of that, I suppose. But it would go against the pattern now. 

The copy list is already in the model, so you're not wrong here. It's just a different way of writing it though. With the new test we'll fail if we forget to specify a field in one of the lists, so we don't have to clutter the field definitions with an extra line for each field defining what to do on copy.

pkcakeout: That is totally fine as well, I think! And yes, the lists mean something different each. It is handy to have a blueprint to copy off, which is where my suggestion would come in - one would see the pattern and simply copy it. On a very abstract level, this is a dependency/cohesion issue of features that depend on other things in modules where they are not defined (like copying and rendering). Again, testing that things are explicitly names is also a fine solution to the issue.

app/tests/reader_studies_tests/test_forms.py
jmsmkn: I think this is best added as a comment in the codebase!

app/tests/reader_studies_tests/test_forms.py
```diff
@@ -548,312 +558,446 @@ def test_image_port_only_with_bounding_box(
...
-            client=client,
-            method=client.post,
-            reverse_kwargs={"slug": rs.slug},
-            data={"title": "3", "copy_display_sets": True},
-            user=editor,
-            follow=True,
-        )
+        if field_name == "readerstudygroupobjectpermission":
+            assert value_in_copy.count() == 10
+            assert set(value_in_copy.all()) != set(original_value.all())
```
jmsmkn: This tests are testing that the sets are not the same, but shouldn't they be disjoint? This happens in a couple of places.

Generally, it's best to avoid for loops and if statements in tests. If possible, it might be worth trying to rework the tests to use parametrise, and then write different tests for the special cases that require an if statement, putting any common code into a fixture (although we do not worry so much about duplication in tests).

app/grandchallenge/reader_studies/models.py
jmsmkn: The pattern follows Django's patterns - that you need to declare on the form the behaviour of that field in that situation - so I would prefer to follow that. Django does have an `__all__` escape hatch, but that tends to be dangerous, so I prefer the pattern of being explicit about what is editable, and in this case, copyable.

app/tests/reader_studies_tests/test_forms.py
jmsmkn: Me either, a question for @amickan

app/tests/reader_studies_tests/test_forms.py
koopmant: > but shouldn't they be disjoint?

Yes, thank you. I couldn't remember it was called disjoint.

app/tests/reader_studies_tests/test_forms.py
amickan: We don't generate actions. We only use the `Follow` model from the actstream library, as far as I know. For the test, the only way would be to manually generate actions as you already suggested.

amickan: We ended up writing our own `Notification` model rather than using the action streams from this library. I don't remember the reasons and can't find them documented anywhere either, but I see that we used to use action streams for forum notifications, but then moved away from those in favor of our own notification model.

PR #4263, opened 2025-08-20
Add capacity reservation breakdown for invoices and enable budget calculation for more than two phases

koopmant:
The challenge request budget table has been updated to break down the calculation for the capacity reservation. 

New fields have been added to the `ChallengeRequest` model to enable more than two phases in the calculation and breakdown. 

The ChallengeRequestBudgetUpdateForm has been updated to allow changing these new fields. The new fields are initially filled with values from the still existing fields that are used by the ChallengeRequestCreateForm, which has not changed. 

Example:
Screenshot 2025-08-20 at 14 21 45

Related to https://github.com/DIAGNijmegen/rse-roadmap/issues/421


jmsmkn:
Could you split this PR into two:

- One to enable budget calculation for more than two phases
  - with a quick skim it doesn't look like any data are being migrated, or old fields deprecated, that will need to be done in the same PR. 
- One to add capacity reservation breakdown for invoices

app/grandchallenge/challenges/admin.py
```diff
@@ -115,18 +115,10 @@ class ChallengeRequestAdmin(ModelAdmin):
         "creator",
         "created",
         "status",
-        "total_cost",
     )
     actions = ["create_challenge", "send_status_update_email"]
     list_filter = ["status"]
 
-    @admin.display(description="Total cost")
-    def total_cost(self, obj):
```
jmsmkn, line 124: It would still be useful to have this number in the admin overview.

app/grandchallenge/challenges/models.py
```diff
@@ -1290,42 +1295,60 @@ def compute_euro_cents_per_hour(self):
         usd_cents_per_hour = max(
             executor.usd_cents_per_hour for executor in executors
         )
-        return usd_cents_per_hour * settings.COMPONENTS_USD_TO_EUR
+        return (
+            usd_cents_per_hour
+            * settings.COMPONENTS_USD_TO_EUR
+            * (1 + settings.COMPONENTS_TAX_RATE_PERCENT)
```
jmsmkn: I thought that this was already accounted for in `executor.usd_cents_per_hour`. It looks like it is being applied correctly when used to calculate the utilisation, but do we ever need to know the number without tax? If not, it could be incorporated into the value set on the executor and removed from the other places where it is added.

app/grandchallenge/challenges/models.py
```diff
@@ -1243,39 +1243,44 @@ def total_compute_time(self):
...
             self.phase_2_number_of_test_images
             * (self.average_size_of_test_image_in_mb * settings.MEGABYTE)
             * self.number_of_tasks
         )
 
     @property
-    def docker_s3_storage_size_bytes(self):
+    def phase_1_docker_storage_size_bytes(self):
         return (
-            # 1 unique container image per submission
```
jmsmkn, line 1264: Why have the docker costs been doubled? If you have two phases then users (should) be using the same container between the phases, so there are no extra docker storage costs.

app/grandchallenge/challenges/models.py
```diff
@@ -1411,69 +1414,85 @@ def total_compute_costs_euros(self):
...
+            )
+            / settings.CHALLENGE_ADDITIONAL_COMPUTE_AND_STORAGE_PACK_SIZE_IN_EURO
         )
 
     @property
-    def total_storage_to_be_invoiced(self):
-        return self.calculate_invoiced_amount(
-            cost=self.total_storage_costs_euros,
-            ratio=1 - self.compute_cost_ratio,
+    def capacity_reservation_euros_per_unit(self):
```
jmsmkn: Does this need to be a separate property? I would have thought that the setting could be used directly, and if the setting name is not clear then that should be updated.

app/grandchallenge/challenges/models.py
```diff
@@ -1411,69 +1414,85 @@ def total_compute_costs_euros(self):
...
-        return (
-            self.minimal_challenge_cost
-            + self.additional_compute_and_storage_costs
-        )
+        return self.base_cost_euros + self.capacity_reservation_euros
+
+    @cached_property
+    def costs_for_phases(self):
+        return [
+            {
```
jmsmkn, line 1460: This is quite a lot to keep track of, might be worth setting up a dataclass for this.

app/tests/challenges_tests/test_models.py
```diff
@@ -145,42 +144,19 @@ def test_total_challenge_cost(settings):
...
-    assert request2.storage_and_compute_cost_surplus == -270
+    assert request2.total_compute_and_storage_costs_euros == 709.07
     assert request2.total_challenge_cost == 6000
 
-    assert request3.storage_and_compute_cost_surplus == 1380
+    assert request3.total_compute_and_storage_costs_euros == 2361.98
     assert request3.total_challenge_cost == 7500
 
-    assert request4.storage_and_compute_cost_surplus == 2580
+    assert request4.total_compute_and_storage_costs_euros == 3555.5
```
jmsmkn, line 156: How have these numbers changed from before?

app/tests/challenges_tests/test_models.py
```diff
@@ -145,42 +144,19 @@ def test_total_challenge_cost(settings):
...
     assert request4.total_challenge_cost == 9000
 
 
-@pytest.mark.django_db
-def test_storage_and_compute_cost_add_up_to_total():
```
jmsmkn, line 162: Why was this test deleted?

app/grandchallenge/challenges/admin.py
koopmant: I'll add it. (This was broken by some past update, so I assumed we weren't using it.)

app/grandchallenge/challenges/models.py
koopmant: > but do we ever need to know the number without tax?

I don't think so. That simplifies things :)

app/grandchallenge/challenges/models.py
koopmant: I only added it so it was available in the template. I'll add it to the context instead and remove this property.

app/grandchallenge/challenges/models.py
koopmant: This will be removed in the next PR, so I stopped optimizing.

app/tests/challenges_tests/test_models.py
koopmant: Note the surplus at the end of the old variable name

We used to have the logic like this:
`total` is `minimal_cost` + `surplus`
`minimal_cost` is `base_cost` plus `minimal_compute_and_storage` (1000 euros)
`surplus` is compute and storage costs above the minimum. 

Now the `surplus` is no longer used:
`total` is `base_cost` + `total_compute_and_storage` 
(`total_compute_and_storage` has an inherent minimum of 1000 euros)

That explains the approximate 1000 euro difference
The remaing difference is because we no longer round the amounts to 10 euros.

app/tests/challenges_tests/test_models.py
koopmant: We used to have a roundabout way of adding up, so this test made sense. Now, we just add them and the test would become the definition of the property:
`storage + compute == storage + compute`

So I deleted it.

app/grandchallenge/challenges/models.py
koopmant: We did not include tax in `storage_bytes_to_euro_per_year` and `registry_bytes_to_euro_per_year`. So we showed it without tax in the costs overview, but I think it should have been with tax included there as well, right?

app/grandchallenge/challenges/models.py
koopmant: They have not been doubled, they have been split up for the phases. (As for why, I think this gives insight in how the storage costs can be high for debug phases when it allows many submissions. I don't think we communicate that clearly.)

I see what you mean, but that was not implemented in the previous calculation either?

```
    @property
    def docker_s3_storage_size_bytes(self):
        return (
            # 1 unique container image per submission
            (self.average_algorithm_container_size_in_gb * settings.GIGABYTE)
            * self.total_num_submissions
        )

    @property
    def total_num_submissions(self):
        return self.phase_1_num_submissions + self.phase_2_num_submissions
```

I think if there are multiple phases, assuming they are for the same task, their will be at least one container that is counted twice. (say, first phase with submission limit 5, second phase with submission limit 2, we anticipate 7 docker images, but we will in reality have 6. Using the last container image of phase 1 for the first submission of phase 2. I can incorporate that in the calculation.

jmsmkn: I see, that was a mistake previously. With multiple phases the containers carry through, so in your example:

> first phase with submission limit 5, second phase with submission limit 2, we anticipate 7 docker images, but we will in reality have 6

We would have 5 containers. You wouldn't (shouldn't) create a container for a later phase that has not been submitted to an earlier one.

app/grandchallenge/challenges/models.py
jmsmkn: Yes both of those should have tax added.

app/grandchallenge/challenges/models.py
koopmant: Ah I see, so before submitting a new container to the second phase the participant would go back to the first phase and then run a sanity check before submitting in the second phase. Then indeed it does not make sense to split up the docker costs like this. In that case:

```python
    @property
    def docker_storage_size_bytes(self):
        return (
            (self.average_algorithm_container_size_in_gb * settings.GIGABYTE)
            * self.num_unique_docker_containers
        )

    @property
    def num_unique_docker_containers(self):
        # A container for a later phase should also have been submitted to an earlier one.
        return max(self.phase_1_num_submissions, self.phase_2_num_submissions)
```

app/config/settings.py
```diff
@@ -144,18 +144,18 @@
...
 COMPONENTS_USD_TO_EUR = float(
     os.environ.get("COMPONENTS_USD_TO_EUR", "0.92472705")
 )
 COMPONENTS_S3_USD_MILLICENTS_PER_YEAR_PER_TB = (
-    12_300_000  # Last calculated 23/08/2023
+    14_883_000  # Last calculated 23/08/2023
 )
 COMPONENTS_ECR_USD_MILLICENTS_PER_YEAR_PER_TB = (
-    39_600_000  # Last calculated 23/08/2023
+    47_916_000  # Last calculated 23/08/2023
```
jmsmkn: These are calculated from the cost explorer which exclude tax, so future maintainers would need to remember to apply them. Maybe update the variable name to make it clear that it is exclusive and apply the tax rate where they are used?

app/grandchallenge/core/templatetags/costs.py
```diff
@@ -27,7 +42,8 @@ def storage_bytes_to_euro_per_year(storage_size):
 @register.filter
 def registry_bytes_to_euro_per_year(storage_size):
     return millicents_to_euro(
-        settings.COMPONENTS_ECR_USD_MILLICENTS_PER_YEAR_PER_TB
+        settings.COMPONENTS_S3_USD_MILLICENTS_PER_YEAR_PER_TB_EXCLUDING_TAX
```
jmsmkn: Incorrect replacement, this should be COMPONENTS_ECR_USD_MILLICENTS_PER_YEAR_PER_TB_EXCLUDING_TAX

koopmant: That was sloppy, thanks for catching it.

jmsmkn: It's why we review!

Issue #3787, opened 2025-01-16
Add a admin page to calculate challenge costs


miriam-groeneveld:
For creating quotes, and see the effects of various options (batching vs non-batching, GPU choices, etc.), it would be good to have a page that can be used to calculate the costs of a challenge. Preferably one that has the fields on the left, and the overview on the right, and have that update instantly.

jmsmkn:
Updated admin to staff to make it clearer who we want this to apply to.

miriam-groeneveld:
Could the maximum runtime be removed? As we currently support a runtime of up to 28 days.

jmsmkn:
That restriction needs to be kept for the public form but the staff form can have a different limit. We do not currently support 28 days but could go there if necessary.

miriam-groeneveld:
> That restriction needs to be kept for the public form but the staff form can have a different limit. We do not currently support 28 days but could go there if necessary.

Should we not make it slighty more for the public form, maybe 24 hours? Or maybe we should add in the form that longer is possible after checking with support.

miriam-groeneveld:
For some algorithms (thinking of pathology), the 1 hour limit is on the short side. There is no indication on the challenge request form that this is a soft limit.

jmsmkn:
Okay that can be increased to the current limit of 6, the help text can mention that if they think they need to go beyond that they should mention it in the form and we can discuss it in onboarding.

chrisvanrun:
Someone created a random challenge request, possibly expecting to see the results of the budget fields.

jmsmkn:
That is correct. There is a github gist with the current calculator I provided to them instead.

PR #4264, opened 2025-08-20
Enable budget calculation for more than two phases

koopmant:
New fields have been added to the ChallengeRequest model to define any number of phases for any number of tasks for the budget calculation. 

The ChallengeRequestBudgetUpdateForm has been updated to allow changing these new fields. The new fields are initially filled with values from the old fields that are used by the ChallengeRequestCreateForm, which has not changed from the users perspective.

Because the challenge request can now define multiple tasks, runtime limits are set for the phases by supplying the corresponding task when they are converted to algorithm phases; unless there is only one task defined on the challenge request, then the input is not needed.

Screenshot 2025-09-16 at 20 54 19

Related to https://github.com/DIAGNijmegen/rse-roadmap/issues/421
Related to #3787

### DIFF
--- a/app/grandchallenge/challenges/forms.py
+++ b/app/grandchallenge/challenges/forms.py
@@ -47,8 +47,6 @@ registration_items = (
     "registration_page_markdown",
 )
 
-HTMX_BLANK_CHOICE_KEY = "__HTMX_BLANK_CHOICE_KEY__"
-
 
 class ChallengeUpdateForm(forms.ModelForm):
     def __init__(self, *args, **kwargs):
@@ -598,56 +596,22 @@ class ChallengeRequestStatusUpdateForm(forms.ModelForm):
 
 
 class ChallengeRequestBudgetUpdateForm(forms.ModelForm):
-    algorithm_selectable_gpu_type_choices = forms.MultipleChoiceField(
-        choices=[
-            (HTMX_BLANK_CHOICE_KEY, GPUTypeChoices.NO_GPU.label),
-            (GPUTypeChoices.T4, GPUTypeChoices.T4.label),
-            (GPUTypeChoices.A10G, GPUTypeChoices.A10G.label),
-        ],
-        widget=forms.CheckboxSelectMultiple,
-        label="Selectable GPU types for algorithm jobs",
-        help_text="The GPU type choices that participants will be able to select for "
-        "their algorithm inference jobs.",
-    )
-
     class Meta:
         model = ChallengeRequest
         fields = (
-            "expected_number_of_teams",
-            "number_of_tasks",
-            "inference_time_limit_in_minutes",
-            "algorithm_selectable_gpu_type_choices",
-            "algorithm_maximum_settable_memory_gb",
-            "average_size_of_test_image_in_mb",
-            "phase_1_number_of_submissions_per_team",
-            "phase_1_number_of_test_images",
-            "phase_2_number_of_submissions_per_team",
-            "phase_2_number_of_test_images",
+            "task_ids",
+            "algorithm_selectable_gpu_type_choices_for_tasks",
+            "algorithm_maximum_settable_memory_gb_for_tasks",
+            "average_size_test_case_mb_for_tasks",
+            "inference_time_average_minutes_for_tasks",
+            "task_id_for_phases",
+            "number_of_teams_for_phases",
+            "number_of_submissions_per_team_for_phases",
+            "number_of_test_cases_for_phases",
         )
-        labels = {
-            "phase_1_number_of_submissions_per_team": "Expected number of submissions per team to Phase 1",
-            "phase_2_number_of_submissions_per_team": "Expected number of submissions per team to Phase 2",
-            "inference_time_limit_in_minutes": "Average algorithm job run time in minutes",
-            "algorithm_maximum_settable_memory_gb": "Maximum memory for algorithm jobs in GB",
-        }
-        help_texts = {
-            "inference_time_limit_in_minutes": (
-                "The average time that you expect an algorithm job to take in minutes. "
-                "This time estimate should account for everything that needs to happen "
-                "for an algorithm container to process <u>one single image, including "
-                "model loading, i/o, preprocessing and inference.</u>"
-            ),
-        }
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        if "" in (
-            initial := self.instance.algorithm_selectable_gpu_type_choices
-        ):
-            initial[initial.index("")] = HTMX_BLANK_CHOICE_KEY
-            self.fields["algorithm_selectable_gpu_type_choices"].initial = (
-                initial
-            )
         self.helper = FormHelper(self)
         self.helper.form_id = "budget"
         self.helper.attrs.update(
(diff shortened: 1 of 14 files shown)


app/grandchallenge/evaluation/forms.py
```diff
@@ -811,37 +798,6 @@ def __init__(self, *args, challenge, **kwargs):
             .all()
         )
 
-        try:
-            challenge_request = ChallengeRequest.objects.get(
```
koopmant, line 808: I'm not sure if there was a good reason to do this on the form. I've moved it to the view.

app/grandchallenge/challenges/migrations/0060_auto_20250820_1420.py
```diff
@@ -0,0 +1,61 @@
+# Generated by Django 4.2.23 on 2025-08-20 14:20
+
+from django.db import migrations
+
+
+def copy_data_to_new_budget_fields(apps, schema_editor):  # noqa: C901
```
jmsmkn, line 6: I think that the data migration will need to change here. Have you tried it on a database with data? I think that any of the fields that have been deprecated will return `None` here, rather than the values actually stored.

I think that 3 migrations are then necessary - the first to create the new fields, a second to migrate the data, and finally a third to deprecate/remove the old fields.

koopmant: I tried it yesterday and it worked. I can make separate migrations to be sure?

koopmant: I checked again to be sure. The deprecated field will return None, but that uses the python `__get__` method, which is not used in migrations. So during the data migration it works fine.

koopmant: Should I already add the migration to remove the fields btw?

koopmant: I've added them.

jmsmkn: Ah sorry - that cannot be added yet.

This is a problem as when deploying you:

1. apply the migrations
2. launch new versions of the platform and wait for them to come online
3. switch traffic from the old processes to the new processes
4. terminate the old processes once the connections to the old processes are drained

So, during 3 you will have both the old and the new versions of the codebase running. If you remove the fields when the old version is still running all of those processes will error out, and will cause problems for the users and hundreds of errors in Sentry.

That is what deprecate fields is for. It allows you to remove the use of the field from the codebase and can handle the field not existing with ease.

In order to use it you need to deprecate the fields, deploy, then remove the fields, deploy. So the fields cannot be deprecated and removed in the same PR.

jmsmkn: So you will need 4 migrations to do this:

1. Field creation
2. Field migration
3. Field deprecation
4. Field removal

1-3 can be in the same PR, 3 and 4 must be in different PRs and deployed separately.

app/grandchallenge/evaluation/views/__init__.py
```diff
@@ -1168,13 +1154,28 @@ def turn_phase_into_algorithm_phase(
...
-            algorithm_maximum_settable_memory_gb
-        )
+        try:
+            challenge_request = ChallengeRequest.objects.get(
+                short_name=phase.challenge.short_name
+            )
+        except ObjectDoesNotExist:
+            pass
+        else:
+            # Fill in runtime limits from challenge request if there is only one task.
```
jmsmkn: The other cases need to be handled as these three properties must be set. If it cannot be done automatically then these values need to be provided by the form.

app/grandchallenge/core/templatetags/index.py
```diff
@@ -0,0 +1,8 @@
+from django import template
+
+register = template.Library()
+
+
+@register.filter()
+def index(indexable, i):
```
jmsmkn: Why is this necessary? There is a `slice` built in and can't you also do `{{ indexable.3 }}` etc directly in the template?

https://docs.djangoproject.com/en/5.2/ref/templates/builtins/#slice

app/grandchallenge/challenges/templates/challenges/partials/budget_table.html
```diff
@@ -94,14 +94,132 @@ <h4>To be invoiced</h4>
...
+                        {% with n_submissions=object.number_of_submissions_per_team_for_phases|index:phase_idx %}
+                        {% with n_teams=object.number_of_teams_for_phases|index:phase_idx %}
+                        {% with n_images=object.number_of_test_images_for_phases|index:phase_idx %}
+                        {% with avg_job_time=object.inference_time_average_minutes_for_tasks|index:task_idx %}
+                        {% with compute_costs_per_hour=object.compute_costs_euros_per_hour_for_tasks|index:task_idx %}
+                        {% with compute_time=object.compute_time_for_phases|index:phase_idx %}
+                        {% with compute_costs=object.compute_costs_euros_for_phases|index:phase_idx %}
+                        {% with data_storage_size=object.data_storage_size_gb_for_phases|index:phase_idx %}
+                        {% with data_storage_costs=object.data_storage_costs_euros_for_phases|index:phase_idx %}
+                        {% with avg_image_size=object.average_size_test_image_mb_for_tasks|index:task_idx %}
```
jmsmkn: I think we can do better here. Maybe construct what you need as a context variable in python?

app/grandchallenge/challenges/templatetags/budget.py
```diff
@@ -0,0 +1,10 @@
+from django import template
+
+register = template.Library()
+
+
+@register.simple_tag
+def get_phase_indices_for_task(task_id_for_phases, task_id):
```
jmsmkn: This seems too specific and suggests that a context variable should be used instead.

app/grandchallenge/challenges/forms.py
```diff
@@ -586,10 +580,104 @@ def __init__(self, *args, **kwargs):
...
+                "All fields defining phases must be of equal length."
+            )
+        self._clean_later_phases_not_more_teams_or_submissions(
+            task_id_for_phases, cleaned_data
+        )
+
+        return cleaned_data
+
+    def _clean_all_fields_nonempty_cleaned_data(self, cleaned_data):
+        print(
```
jmsmkn: There shouldn't be print statements in the codebase, use logging if necessary.

app/grandchallenge/challenges/models.py
```diff
@@ -1164,159 +1244,194 @@ def create_challenge(self):
...
 
     @property
-    def phase_1_data_storage_size_bytes(self):
-        return (
-            self.phase_1_number_of_test_images
-            * (self.average_size_of_test_image_in_mb * settings.MEGABYTE)
-            * self.number_of_tasks
-        )
+    def number_of_docker_images_per_team_for_tasks(self):
+        # A docker for a later phase should also be submitted to an earlier one.
```
jmsmkn: There is no such thing as "A docker". Docker is a company. They provide tooling around container images.

app/grandchallenge/challenges/migrations/0060_auto_20250820_1420.py
koopmant: Thanks for explaining this again! I forgot about the issues during deployment. 

For the migrations, 1 and 3 can technically be combined. Do you still want to have them separate?

app/grandchallenge/evaluation/views/__init__.py
koopmant: Hmm, all right. Not sure yet how to implement that on the form. I suppose I can just request input about which task each phase belongs to.

app/grandchallenge/core/templatetags/index.py
koopmant: Had not found the slice builtin, thanks. But I want just one item, not a slice. `{{ indexable.3 }}` does not work with variables, so something like `{{ indexable.forloop.counter0 }}` does not work. 
I could let the view or the model provide a more appropriate structure that is easier to handle in the template.

app/grandchallenge/challenges/forms.py
koopmant: 🙈

app/grandchallenge/challenges/templates/challenges/partials/budget_table.html
koopmant: yes, indeed.

app/grandchallenge/challenges/migrations/0060_auto_20250820_1420.py
jmsmkn: 1-3 cannot be combined - the database does not allow schema migrations in the same transaction as data migrations.

app/grandchallenge/core/templatetags/index.py
jmsmkn: It supports the same syntax, so would be `{{ indexable|slice:"3" }}` no?

app/grandchallenge/challenges/migrations/0060_auto_20250820_1420.py
koopmant: Not 1-3; 1 and 3.

app/grandchallenge/core/templatetags/index.py
koopmant: That still returns a slice (of the first three items in this case).

app/grandchallenge/challenges/migrations/0060_auto_20250820_1420.py
jmsmkn: I don't understand what you mean. None of the migrations can be combined.

koopmant: I'm referring to this earlier comment https://github.com/comic/grand-challenge.org/pull/4264#discussion_r2355824696

I believe we can create new fields and deprecate fields at the same time, because deprecated fields can still be used in the data migration. Do I understand this correctly or am I mistaken?

jmsmkn: It is safer to have them separate, then the data migration will not need to handle Null data unless it is already possible on the field prior to 1.

koopmant: Ah right!

app/grandchallenge/evaluation/views/__init__.py
koopmant: I've added a dynamic field for the task id for each phase. I've made it required if the phase is selected.

koopmant:
@jmsmkn I don't think it is worth splitting up the PR or creating stacked commits, but let me know if this is annoying to review.

chrisvanrun:
Sorry! Heavy conflict with https://github.com/DIAGNijmegen/rse-grand-challenge/pull/4495. We'll need to discuss how to combine this and what makes more sense to merge first.

koopmant:
Ah, this get's automatically closed when the branch is up-to-date with main. I wanted to use this as a base for a feature branch with stacked PR's. Will open it again after the first PR.

chrisvanrun:
It is alive again!

PR #4270, opened 2025-08-25
DicomImageSetUpload model

amickan:


app/grandchallenge/healthimaging/models.py
```diff
@@ -0,0 +1,101 @@
...
+    creator = models.ForeignKey(
+        to=settings.AUTH_USER_MODEL,
+        null=True,
+        default=None,
+        on_delete=models.SET_NULL,
+    )
+
+    user_uploads = models.ManyToManyField(
+        UserUpload, blank=True, related_name="dicom_import_jobs"
+    )
```
amickan: These fields are not needed yet, maybe better to remove for now?

app/config/settings.py
```diff
@@ -216,6 +216,21 @@
...
 AWS_SES_REGION_NAME = os.environ.get("AWS_SES_REGION_NAME")
+AWS_HEALTH_IMAGING_DATASTORE_ID = os.environ.get(
+    "AWS_HEALTH_IMAGING_DATASTORE_ID"
+)
+AWS_HEALTH_IMAGING_INPUT_BUCKET_NAME = os.environ.get(
+    "AWS_HEALTH_IMAGING_INPUT_BUCKET_NAME"
+)
+AWS_HEALTH_IMAGING_OUTPUT_BUCKET_NAME = os.environ.get(
+    "AWS_HEALTH_IMAGING_OUTPUT_BUCKET_NAME"
+)
```
jmsmkn: Why have two buckets rather than one with permissions based upon the prefix?

app/config/settings.py
```diff
@@ -216,6 +216,21 @@
...
+AWS_HEALTH_IMAGING_IMPORT_ROLE_ARN = os.environ.get(
+    "AWS_HEALTH_IMAGING_IMPORT_ROLE_ARN"
+)
+AWS_HEALTH_IMAGING_REGION_NAME = os.environ.get(
+    "AWS_HEALTH_IMAGING_REGION_NAME", "eu-west-1"
+)
```
jmsmkn: Is this necessary? I don't think we'll ever plan on running it in a different region, and this will anyway need to match the region of the S3 bucket(s).

app/config/settings.py
```diff
@@ -563,6 +578,7 @@ def get_private_ip():
     "grandchallenge.well_known",
     "grandchallenge.utilization",
     "grandchallenge.discussion_forums",
+    "grandchallenge.healthimaging",
```
jmsmkn: Why a separate app and not part of `cases`?

app/grandchallenge/healthimaging/models.py
```diff
@@ -0,0 +1,101 @@
...
+    STATUS_CHOICES = (
+        (PENDING, "Queued"),
+        (DEIDENTIFYING, "De-Identifying"),
+        (DEIDENTIFIED, "De-Identified"),
+        (STARTED, "Importing"),
+        (REQUEUED, "Re-Queued"),
+        (FAILURE, "Failed"),
+        (SUCCESS, "Succeeded"),
+        (CANCELLED, "Cancelled"),
+    )
```
jmsmkn: This is the old-school way of defining choices, please use `TextChoices` with a matching database constraint (this should be in the style guide).

app/grandchallenge/healthimaging/models.py
jmsmkn: I think this is okay for now, I think that we want to match the pattern used by `RawImageUploadSession`.

app/grandchallenge/healthimaging/models.py
```diff
@@ -0,0 +1,101 @@
...
+            jobName=self._import_job_name,
+            datastoreId=settings.AWS_HEALTH_IMAGING_DATASTORE_ID,
+            dataAccessRoleArn=settings.AWS_HEALTH_IMAGING_IMPORT_ROLE_ARN,
+            inputS3Uri=self._import_input_s3_uri,
+            outputS3Uri=self._import_output_s3_uri,
+        )
+        return job["jobId"]
+
+
+class DicomImportJob(UUIDModel):
```
jmsmkn: Is this not a `DicomImageSetUpload`?

app/grandchallenge/healthimaging/models.py
```diff
@@ -0,0 +1,101 @@
...
+    )
+
+    status = models.PositiveSmallIntegerField(
+        choices=STATUS_CHOICES, default=PENDING, db_index=True
+    )
+
+    error_message = models.TextField(blank=False, null=True, default=None)
+
+    def get_importer(self):
+        return HealthImagingWrapper(job_id=self.pk)
```
jmsmkn: This seems like unnecessary complexity, why not have the methods on this class?

app/config/settings.py
amickan: This is what the HI developer guide recommends doing. You are suggesting 1 bucket with 2 prefixes "source" and "logging" and the inputs and logs nested under those accordingly? What would be the benefit of that?

app/config/settings.py
amickan: Good point. We can just use `AWS_S3_REGION_NAME` here then.

app/config/settings.py
amickan: It seemed cleaner to separate the two to me. They hold fundamentally different models and workflow logic. But if you prefer to keep it all in the cases app that's fine by me as well. The main benefit I see with keeping it in the cases app is for url paths, which would then also be `cases/...`, which does make sense.

app/config/settings.py
jmsmkn: You shouldn't need to set a region name in the boto arguments at all. I think a lot of existing settings come from migrating between regions and trying to balance things with minio.

app/grandchallenge/healthimaging/models.py
amickan: I chose this name to stay with the Healthimaging nomenclature. `DicomImageSetUpload` could work, but it doesn't quite describe what this is I think. DicomImageSets are the result of the upload and it's HI that decides how to group the dicom files that we feed it into DicomImageSets. What we are doing, i.e. what this model is doing, is just starting a dicom import job, so I personally find `DicomImportJob` more fitting.

app/config/settings.py
jmsmkn: It reduces the number of resources we create and all the configuration can be seen in one place.

app/config/settings.py
jmsmkn: The models are not fundamentally different - we are doing this to create `Image` objects which already exist in cases.

app/grandchallenge/healthimaging/models.py
amickan: What do you mean? We are starting the import from a `DicomImportJob` instance, so it's the `import_images` method on this class that needs to somehow instantiate the `HealthImagingWrapper` with the proper kwargs (at this point just the `job_id`). We don't necessarily need the `get_importer` method here, I can also just do `HealthImagingWrapper(job_id=self.pk)` in `import_images`, but I don't see how moving this to the `HealthImagingWrapper` would work.

app/grandchallenge/healthimaging/models.py
jmsmkn: It is doing more than that - its responsibility is to create `Image` objects that are singularly related to a `DICOMImageSet` (note that DICOM is an acronym), from a set of `Upload` objects that it also needs to de-identify. That is much more than just starting a job on AWS, and the name I suggest fits more closely with the names that already exist in our codebase.

app/grandchallenge/healthimaging/models.py
jmsmkn: The wrapper isn't wrapping anything nor is it creating an abstraction that we will want to swap out. All of its methods can move directly to the model, then the `job_id` can be replaced with `self.pk`.

amickan: Ah this way around, ok, I understand your point. I'll move everything to the model.

app/config/settings.py
amickan: Ok, I will update the infrastructure then.

app/grandchallenge/healthimaging/models.py
amickan: Ok, convinced.

app/config/settings.py
amickan: It is true that I don't need to specify a region, it will just use the `AWS_DEFAULT_REGION` in the environment variables. Somehow the tests where I use a Stubber fail with a "NoRegion" error though, even though `AWS_DEFAULT_REGION` should be in the test environment variables (it's set to default to `eu-central-1` in our settings file. Have you encountered this before?

amickan: This only happens on Github by the way, locally the tests pass. I also see that in all other places where we use a Stubber, we do pass a region_name explicitly to the underlying clients (it's all cases where the service is not located in the default region, so that is necessary in those cases).

jmsmkn: Ah okay, whatever then, pass through `settings.AWS_DEFAULT_REGION`.

amickan: I can do that, but actually I'm now questioning the usefulness of the test with the stubber. All it tests is that we make the api call with the right parameters. All other tests have been updated to not use a Stubber since those don't really benefit from that in my eyes. Maybe let me know what you think about the tests and then I can still add the region_name to the client if we think the stubber test is useful, and otherwise just drop it.

amickan: Ok, I've passed through the default region, just because I guess that we will be using a Stubber anyway in later tests, even the current one gets scrapped.

amickan: The test still fails but because of the values I pass in for the role arn (and possibly others) - empty strings. I guess the way to go then is to pass fake, but close to real life values to the test (overriding the env vars for the test).

jmsmkn: Yes, that is what is done in other tests

https://github.com/comic/grand-challenge.org/blob/7f59a9098f2e77e03e46b7cc493b5dc83e4b258a/app/tests/components_tests/test_amazon_sagemaker_training_backend.py#L155

Good to always return a string for environment variables rather than `None`.

amickan: That fixed it indeed.

app/grandchallenge/cases/models.py
```diff
@@ -837,3 +839,94 @@ def save(self, *args, **kwargs):
...
+class DicomImageSetUploadStatusChoices(models.TextChoices):
+    INITIALIZED = "INITIALIZED", "Initialized"
+    PENDING = "PENDING", "Queued"
+    DEIDENTIFYING = "DEIDENTIFYING", "De-Identifying"
+    DEIDENTIFIED = "DEIDENTIFIED", "De-Identified"
+    STARTED = "STARTED", "Importing"
+    REQUEUED = "REQUEUED", "Re-Queued"
+    FAILURE = "FAILURE", "Failed"
+    SUCCESS = "SUCCESS", "Succeeded"
+    CANCELLED = "CANCELLED", "Cancelled"
```
jmsmkn: Given how we run Celery tasks we cannot display to the user the `DEIDENTIFYING` state, there is a mismatch between `"STARTED", "Importing"` and many of the states are not in use (yet). Maybe have a look at `PostProcessImageTaskStatusChoices` and `PostProcessImageTask` as that implements our current best practices regarding Celery tasks.

app/grandchallenge/cases/models.py
```diff
@@ -837,3 +839,94 @@ def save(self, *args, **kwargs):
...
+    user_uploads = models.ManyToManyField(
+        UserUpload, blank=True, related_name="dicom_import_jobs"
+    )
+
+    DicomImageSetUploadStatusChoices = DicomImageSetUploadStatusChoices
+    status = models.CharField(
+        max_length=13,
+        choices=DicomImageSetUploadStatusChoices.choices,
+        default=DicomImageSetUploadStatusChoices.PENDING,
+        db_index=True,
```
jmsmkn: Please use `Meta.indexes` instead as `db_index` is on the deprecation path https://docs.djangoproject.com/en/5.2/ref/models/fields/#django.db.models.Field.db_index

app/grandchallenge/cases/models.py
```diff
@@ -837,3 +839,94 @@ def save(self, *args, **kwargs):
...
+    )
+
+    error_message = models.TextField(blank=False, null=True, default=None)
```
jmsmkn: Should `None` be used to represent nothing? `blank` is only enforced when `clean` is run, which we hardly do when working with the objects outside of forms (really we should always run `full_clean` before `save`, maybe now is a good time to start?). Otherwise we may get two representations of nothing - `None` and `""`. Maybe `models.TextField(blank=False, default="")` is better?

app/grandchallenge/cases/models.py
```diff
@@ -837,3 +839,94 @@ def save(self, *args, **kwargs):
                     kwargs={"post_process_image_task_pk": self.pk}
                 ).apply_async
             )
+
+
+class DicomImageSetUploadStatusChoices(models.TextChoices):
```
jmsmkn: `Dicom` should be `DICOM` everywhere, same for `DICOMImageSetUpload` which should also have the `verbose_name` set for correct conversion from camel case.

app/grandchallenge/cases/models.py
```diff
@@ -837,3 +839,94 @@ def save(self, *args, **kwargs):
...
+            self.__health_imaging_client = boto3.client(
+                "medical-imaging",
+                region_name=settings.AWS_DEFAULT_REGION,
+            )
+        return self.__health_imaging_client
+
+    @property
+    def _import_job_name(self):
+        # HealthImaging requires job names to be max 64 chars
+        return f"{settings.COMPONENTS_REGISTRY_PREFIX}-HI-{self.pk}"
```
jmsmkn: Is `-HI-` necessary and is the job name really limited in the same way or has this been copy pasted from components? The pk maps directly to a `DICOMImageSetUpload` object so the identifier could be safely removed.

app/grandchallenge/cases/models.py
```diff
@@ -837,3 +839,94 @@ def save(self, *args, **kwargs):
...
+                datastoreId=settings.AWS_HEALTH_IMAGING_DATASTORE_ID,
+                dataAccessRoleArn=settings.AWS_HEALTH_IMAGING_IMPORT_ROLE_ARN,
+                inputS3Uri=self._import_input_s3_uri,
+                outputS3Uri=self._import_output_s3_uri,
+            )
+        except ClientError:
+            self.status = DicomImageSetUploadStatusChoices.FAILURE
+            self.error_message = "An unexpected error occurred"
+            self.save()
+            raise
```
jmsmkn: If the exception is raised the transaction will get rolled back and the object won't actually be saved. You would need to use `logger.error(..., exc_info=True)` to log the error and return normally to save the object.

The things that can be raised here will be more than `ClientError`, so if this is for the catch all then this should be replaced with `Exception` as some client errors we actually want to handle, such as rate limits.

app/grandchallenge/cases/models.py
amickan: Yes, the job name is limited to 64 characters (see https://docs.aws.amazon.com/healthimaging/latest/APIReference/API_StartDICOMImportJob.html#healthimaging-StartDICOMImportJob-request-jobName). `-HI-` is not strictly necessary and just using `{settings.COMPONENTS_REGISTRY_PREFIX}-{self.pk}` is fine by me.

app/grandchallenge/cases/models.py
amickan: Oh right, thanks for catching this! 

RE which errors to catch and handle in the code: the throttling exception should raise a RetryStep and the celery task should be retried on that, right? I can add that. 

As for the other error messages (https://docs.aws.amazon.com/healthimaging/latest/APIReference/API_StartDICOMImportJob.html#API_StartDICOMImportJob_Errors), I don't think we would want to automatically retry for any of those and so just raising is fine for now at least. Would you agree? 

The ValidationException might be something we want to tell the user about, in a redacted way, like "The data did not meet AWS Healthimaging requirements."?

app/grandchallenge/cases/models.py
amickan: I always get confused with this detail, but what you're saying makes sense. That way we will only have one representation for empty, i.e. empty string. I will also add the call to full_clean() for good measure.

app/grandchallenge/cases/models.py
amickan: Ok, I have removed all unused statuses for now and matched the naming to `PostProcessImageTaskStatusChoices`

app/grandchallenge/cases/models.py
jmsmkn: You would need to check the real world reasons but I don't think that `ValidationException` has anything to do with the users data, but is a problem with the arguments that we send to `start_dicom_import_job` as the call is synchronous. Those errors would be unexpected I think. Any data problems you then would see later in the stopped job event.

I think the only ones that needs to be handled would be `ThrottlingException` and `ServiceQuotaExceededException`, see the handling in `_create_sagemaker_job`. I don't know if they are both `ClientErrors`, errors in `boto` are pretty annoying to handle.

amickan: > is a problem with the arguments that we send to `start_dicom_import_job` as the call is synchronous

That makes a lot of sense. I hadn't thought about that properly. Issues with the data would be described in the logs of a failed import job of course. 



> I think the only ones that needs to be handled would be `ThrottlingException` and `ServiceQuotaExceededException`

I made a suggestion for handling the throttling error. For the service quota exception - that would require us to increase the limits before we retry, no?

amickan: As for what the error looks like, I'm not 100% sure, but reading https://boto3.amazonaws.com/v1/documentation/api/latest/guide/error-handling.html I think catching a ClientError and then parsing the error code is the way to go.

jmsmkn: > For the service quota exception - that would require us to increase the limits before we retry, no?

From what I understand:

> The request caused a service quota to be exceeded.

So your service quota could be 100 concurrent import jobs, and you tried to launch your 101st. Here is a list of the quotas, I think the only one relevant for `start_import_jobs` is "Maximum concurrent import jobs per data store", which would reduce as jobs complete so would be retry-able:

Screenshot 2025-08-28 at 11 21 19

app/grandchallenge/cases/tasks.py
```diff
@@ -520,3 +522,27 @@ def _check_post_processor_result(*, post_processor_result, image):
...
+            "Upload is not ready for importing into HealthImaging."
+        )
+
+    upload.status = DICOMImageSetUploadStatusChoices.STARTED
+    upload.save()
+
+    try:
+        on_commit(upload.start_dicom_import_job)
+    except RetryStep:
+        raise
```
jmsmkn: The try ... except block doesn't have any effect as nothing is happening in the except, so everything will just get raised. The `on_commit` is also unnecessary as you are not calling a celery task nor does the call need to be deferred (we are anyway at the end of the transaction here):

```suggestion
    upload.start_dicom_import_job()
```

amickan: I somehow thought the RetryStep had to be raised again for the celery task to pick it up properly, but on second thought that does not make sense.

app/grandchallenge/cases/models.py
amickan: > would reduce as jobs complete so would be retry-able

Ah of course, yes, in that case I agree that retrying would make sense.

app/grandchallenge/cases/models.py
```diff
@@ -837,3 +840,111 @@ def save(self, *args, **kwargs):
...
+            )
+        ]
+        indexes = (models.Index(fields=["status"]),)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.__health_imaging_client = None
+
+    def save(self, *args, **kwargs):
+        self.full_clean()
```
jmsmkn, line 890: Not sure about doing this here, `save` should just save as any `ValidationError` needs to be handled somehow, that is not possible here and that is different for everywhere the model is used (e.g. Forms directly or through a foreign key relationship, Celery Tasks, the admin etc). This is why it's not in `UUIDModel` or in Djangos core. For now, fine, but I think we just need to get in the habit of calling it or we might introduce new problems.

PR #4271, opened 2025-08-25
Implement end of study message

pkcakeout:
Pitch https://github.com/DIAGNijmegen/rse-roadmap/issues/426

Implements the end-of-reader-study message option in the reader study settings. Please tell me if tests are needed somehow/somewhere for this. There does not seem to be much logic beyond the storing of the values, so it is relatively basic...?

Also I deviated from the suggested naming `completion_message_markdown` and named the field `end_of_study_text[_markdown]`. I did not like "completion message" because it seemed a bit too generic, e.g. what is completed?


pkcakeout:
Tests added, please tell me if this is overboard. I added the big-naughty-list-of-strings because those actually try to avoid cleaning of data in some more advanced ways, e.g. binary data. However, I bet that the `bleach` library used is with 99% certainty tested against this list as well, and having an opaque base64 data file for strings is not super-ideal. So I could also replace the data file with a few representative strings if you prefer this. So what is the preferred?

Also, I added `_save` suffixes creating `end_of_study_text_safe` and also `help_text_safe` with the idea that we can phase out `help_text` to have a consistent naming scheme. Phasing out `help_text` would need the two-cycle delay again so that cirrus and cirrus-previous-release both use `help_text_safe`. Thought that makes sense!?

pkcakeout:
Ok... changed it to a single string again.

app/grandchallenge/reader_studies/models.py
```diff
@@ -555,7 +560,28 @@ def remove_reader(self, user):
     @property
     def help_text(self) -> str:
         """The cleaned help text from the markdown sources"""
-        return md2html(self.help_text_markdown, link_blank_target=True)
+        return md2html(
+            self.help_text_markdown,
+            link_blank_target=True,
+            create_permalink_for_headers=False,
+        )
```
jmsmkn:
```suggestion
        # Deprecated method
        return self.help_text_safe
```

jmsmkn (approved):
LGTM!

PR #4274, opened 2025-08-26
Improve responsiveness of navbar in header and on challenge page

HarmvZ:
The responsiveness of the main navbar and the secondary navbar on the challenge page is a bit too aggressive in my opinion. Both were set to `navbar-expand-lg` which meant that they collapse to a vertical form (+ hidden behind hamburger menu for main navbar) when the page width is less than 991px. I think this is too soon. The main navbar easily fits until the next breakpoint at 768px. The secondary navbar on the challenge page fits until the last breakpoint (576px). I've adjusted them accordingly, check the videos for a comparison.

## Before

 https://github.com/user-attachments/assets/896416af-a7e9-4c9f-a3ab-41eece27130c

## After

https://github.com/user-attachments/assets/3ead6e2b-7a6e-499d-9ad9-584ea50f1fb6

### DIFF
--- a/app/grandchallenge/challenges/templates/challenges/challenge_topbar.html
+++ b/app/grandchallenge/challenges/templates/challenges/challenge_topbar.html
@@ -4,7 +4,7 @@
 
 {% if challenge %}
     <nav class="navbar navbar-expand-sm navbar-light p-0 mb-3 mx-0 border-bottom align-items-start challenge-topbar">
-        <div class="col-auto mr-auto px-0">
+        <div class="col mr-auto px-0">
             <ul class="nav navbar-nav nav-tabs">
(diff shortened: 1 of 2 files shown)


jmsmkn:
Isn't this only the case when you're not a participant of the challenge and not all of the features are enabled. This is what you get as an admin:

Screenshot 2025-08-26 at 16 23 36

Would this still work with all of that?

jmsmkn (approved):
Unsure - users of these devices won't see that the items are falling in to place and may just see a jumble. I do prefer a flat list in those cases, but I guess you never know at what device size the breaks will appear. We shouldn't spend much time on this so if you're happy with this fine by me.

PR #4277, opened 2025-08-29
DICOM import job event handling

koopmant:


app/grandchallenge/cases/models.py
```diff
@@ -948,3 +980,87 @@ def start_dicom_import_job(self):
             self._mark_failed(
                 error_message="An unexpected error occurred", exc=e
             )
+
+    def get_job_output_manifest(self, *, import_job):
+        output_uri = import_job["outputS3Uri"]
+        parsed = urlparse(output_uri)
+        bucket = parsed.netloc
+        key = parsed.path.lstrip("/") + "job-output-manifest.json"
```
amickan, line 1025: You could also just get the bucket name from the settings `settings.AWS_HEALTH_IMAGING_BUCKET_NAME` and then maybe add 2 new environment variables for the 2 prefixes we have in that bucket `logs` and `inputs` and then just construct the key from the those environment variables. For the uploading of the deidentified data to the HI bucket, I was going to do it like that.

app/grandchallenge/cases/models.py
```diff
@@ -948,3 +980,87 @@ def start_dicom_import_job(self):
...
+            elif job_status == "FAILED":
+                self.handle_failed_job(event=event)
+                raise ValueError("Import job failed")
+            else:
+                raise ValueError("Invalid job status")
+        except Exception as e:
+            self._mark_failed(
+                error_message="An unexpected error occurred", exc=e
+            )
+            raise
```
amickan: Why raise here? `_mark_failed` already pushes the error to sentry.

app/grandchallenge/cases/models.py
```diff
@@ -948,3 +980,87 @@ def start_dicom_import_job(self):
...
+    def handle_event(self, *, event):
+        job_status = event["jobStatus"]
+        try:
+            if job_status == "COMPLETED":
+                self.handle_completed_job(event=event)
+            elif job_status == "FAILED":
+                self.handle_failed_job(event=event)
+                raise ValueError("Import job failed")
+            else:
+                raise ValueError("Invalid job status")
```
amickan: In these case you would also want to mark the upload as failed, right?

app/grandchallenge/cases/models.py
amickan: If I understand correctly, you could get here from a completed job with invalid image sets. In that case, would we want the error message to be different?

app/grandchallenge/cases/models.py
```diff
@@ -948,3 +980,87 @@ def start_dicom_import_job(self):
...
+            )
+            raise
+        else:
+            self.status = self.DICOMImageSetUploadStatusChoices.COMPLETED
+            self.save()
+
+    def handle_completed_job(self, *, event):
+        image_sets = self.get_image_sets_for_dicom_import_job(import_job=event)
+        if not image_sets:
+            raise ValueError("No image sets created.")
```
amickan: Is that a possible outcome for a completed import job? If no image sets were created, wouldn't the job be marked as a failed from HealthImaging?

app/grandchallenge/cases/models.py
koopmant: I was going to do it like that. But the output_s3_uri we specify in the `start_dicom_import_job` call is not the final location of the job manifest. A new directory is created for the dicom import job. 
In call:
```
start-dicom-import-job 
  --output-s3-uri 
  s3://rumc-gc-s-thomas-2608-healthimaging/logs/cf94e169[...]/
```
In event response:
```
'outputS3Uri': 
  's3://rumc-gc-s-thomas-2608-healthimaging/logs/cf94e169[...]/5bb1dcedf7c14ece969d7fe73c5b87a1-DicomImport-9b64353134ee28d8242d46d618300aac/'
```

app/grandchallenge/cases/models.py
```diff
@@ -948,3 +980,87 @@ def start_dicom_import_job(self):
...
+        valid = self.validate_image_sets(image_sets)
+        if not valid:
+            self.cleanup_image_sets(image_sets)
+            raise ValueError("Image sets invalid.")
+        self.convert_image_sets_to_internal(image_sets)
+
+    def handle_failed_job(self, *, event):
+        image_sets = self.get_image_sets_for_dicom_import_job(import_job=event)
+        if image_sets:
+            self.cleanup_image_sets(image_sets)
```
amickan: For failed jobs, you also want to know the exact error message and hence would need to inspect the failure.ndjson, right? Even if we don't want to expose it to participants, we might want to log it to sentry given that the S3 bucket contents get deleted eventually, right?

app/grandchallenge/cases/models.py
```diff
@@ -948,3 +980,87 @@ def start_dicom_import_job(self):
...
+
+        try:
+            data = json.load(manifest)
+        except json.decoder.JSONDecodeError:
+            return []
+
+        expression = jmespath.compile(
+            "jobSummary.imageSetsSummary[].imageSetId"
+        )
+        image_sets = expression.search(data)
```
amickan: Why do this over just a standard dictionary lookup?

app/grandchallenge/cases/models.py
amickan: Also, why raise again in these cases? That would make the celery task fail, but that is not what we want, right?

app/grandchallenge/cases/models.py
amickan: Ah ok, thanks for explaining!

app/grandchallenge/cases/models.py
```diff
@@ -948,3 +980,87 @@ def start_dicom_import_job(self):
...
+        parsed = urlparse(output_uri)
+        bucket = parsed.netloc
+        key = parsed.path.lstrip("/") + "job-output-manifest.json"
+
+        # Try to get the manifest.
+        try:
+            obj = self._s3_client.get_object(bucket=bucket, key=key)
+            return obj["Body"]
+        except self._s3_client.exceptions.NoSuchKey as e:
+            raise RetryStep("Manifest not (yet) found for job output") from e
```
amickan: And this is necessary because Healthimaging fires the event off before it has saved the manifest file to S3?

koopmant: That was my assumption. I copied the retry logic from the example, and then narrowed it down to this error. I'm not sure why else we would retry. (The other expected error is [InvalidObjectState](https://boto3.amazonaws.com/v1/documentation/api/1.35.66/reference/services/s3/client/exceptions/InvalidObjectState.html))
I cannot find if we can rely on the event being send after the manifest is saved.

app/grandchallenge/cases/models.py
koopmant: I guess it's a bit overkill. I'll replace it.

app/grandchallenge/cases/models.py
koopmant: Yes, but it will be by `except Exception` right? I wanted to use `self._mark_failed` in places like this, but it has an exception argument, so I started doing it like this everywhere.

app/grandchallenge/cases/models.py
koopmant: Ah right, I didn't realize.

app/grandchallenge/cases/models.py
koopmant: Good point! I'll add it.

app/grandchallenge/cases/models.py
```diff
@@ -948,3 +980,87 @@ def start_dicom_import_job(self):
...
+        self.convert_image_sets_to_internal(image_sets)
+
+    def handle_failed_job(self, *, event):
+        image_sets = self.get_image_sets_for_dicom_import_job(import_job=event)
+        if image_sets:
+            self.cleanup_image_sets(image_sets)
+
+    def cleanup_image_sets(self, image_sets):
+        for image_set in image_sets:
+            self.delete_image_set(image_set["imageSetId"])
```
koopmant: I think I should handle this in a separate task...

app/grandchallenge/cases/models.py
amickan: Ah yeah ok, I misread the code. `_mark_failed` doesn't require an exception though, you should be able to call it without one.

amickan: Hm well you do need an exception if you want to push it to sentry.... and I guess we want to report all failed imports to sentry at least at the beginning until we start seeing patterns and can handle recurring error messages. But then I think it would be clearer if the exception was raised in `handle_failed_job` and if it contained the actual error message from healthimaging.

amickan: and maybe create a new exception class `FailedDICOMImportException` or something like that?

app/grandchallenge/cases/models.py
amickan: That might be a good idea if the call can take a while and waits for the deletion to be complete to return.

app/grandchallenge/cases/models.py
```diff
@@ -948,3 +980,87 @@ def start_dicom_import_job(self):
...
+    @staticmethod
+    def validate_image_sets(image_sets):
+        for image_set in image_sets:
+            if not image_set["imageSetId"]:
+                return False
+            if not image_set["isPrimary"]:
+                return False
+            if not image_set["imageSetVersion"] == 1:
+                return False
+        return True
```
amickan: I'm not at all sure about this, but I thought that we wanted a single import job to result in a single image set? If so, that needs to be checked here or above as well, but I might be wrong.

amickan: I'm taking this from the following comment from James: https://github.com/DIAGNijmegen/rse-roadmap/issues/302#issuecomment-3155536239

But it is possible that I misunderstood that comment.

app/grandchallenge/cases/migrations/0019_remove_dicomimagesetupload_dicomuimagesetupload_status_valid_and_more.py
jmsmkn, line 1: Please merge any new migrations that will land on main.

app/grandchallenge/cases/models.py
```diff
@@ -921,6 +938,24 @@ def _mark_failed(self, *, error_message, exc):
...
+            delete_results = self._health_imaging_client.delete_image_set(
+                imageSetId=image_set_id,
+                datastoreId=settings.AWS_HEALTH_IMAGING_DATASTORE_ID,
+            )
+        except self._health_imaging_client.exceptions.ThrottlingException as e:
+            raise RetryStep("Request throttled") from e
+        except (
+            self._health_imaging_client.exceptions.InternalServerException
+        ) as e:
+            raise RetryStep("Server side error") from e
```
jmsmkn: Probably you do not want to retry this, you could be hammering it when it is 504.

app/grandchallenge/cases/models.py
```diff
@@ -948,3 +983,85 @@ def start_dicom_import_job(self):
             self._mark_failed(
                 error_message="An unexpected error occurred", exc=e
             )
+
+    def get_job_output_manifest(self, *, event):
```
jmsmkn: This is returning the job summary from the output manifest, so maybe rename?

```suggestion
    def get_job_summary(self, *, event):
```

app/grandchallenge/cases/models.py
```diff
@@ -948,3 +983,85 @@ def start_dicom_import_job(self):
...
+        raise DICOMImportJobFailedError(
+            message=f"Import job {job_id} failed", message_details=failure_log
+        )
+
+    def cleanup_image_sets(self, *, manifest):
+        image_sets = manifest.get("imageSetsSummary", [])
+        for image_set in image_sets:
+            self.delete_image_set(image_set["imageSetId"])
+
+    def revert_image_set_to_initial_version(self, image_set):
```
jmsmkn: Things may be more robust if `revert_image_set_to_initial_version` and `cleanup_image_sets` are performed in a separate celery task.

app/grandchallenge/cases/models.py
```diff
@@ -948,3 +983,85 @@ def start_dicom_import_job(self):
...
+            )
+        self.convert_image_set_to_internal(image_set)
+
+    def handle_failed_job(self, *, event):
+        manifest = self.get_job_output_manifest(event=event)
+        self.cleanup_image_sets(manifest=manifest)
+        failure_log = self.get_job_output_failure_log(manifest=manifest)
+        job_id = manifest["jobId"]
+        raise DICOMImportJobFailedError(
+            message=f"Import job {job_id} failed", message_details=failure_log
```
jmsmkn: What is the intention of message_details? If it is for the user they should not just be sent logs. If it is for us there might be some confusion with the other existing usage for this variable of the same name on other error objects, and may not be necessary at all if it is being sent to sentry.

jmsmkn (commented):
Looks good, couple of small things.

app/grandchallenge/cases/models.py
koopmant: It's for us. Thanks, I'll look at the other usages.

app/grandchallenge/cases/models.py
koopmant: I agree, thanks for confirming.

app/grandchallenge/cases/models.py
jmsmkn: Okay - then it might be worth storing that in a separate field on the model, `internal_failure_log`, etc.

app/grandchallenge/cases/models.py
jmsmkn: I meant to say two celery tasks btw, one for each of those methods.

app/grandchallenge/cases/models.py
koopmant: done

app/grandchallenge/cases/migrations/0019_remove_dicomimagesetupload_dicomuimagesetupload_status_valid_and_more.py
koopmant: done

app/grandchallenge/cases/models.py
koopmant: Is it worth retrying with backoff?

app/grandchallenge/cases/models.py
```diff
@@ -846,6 +852,53 @@ class DICOMImageSetUploadStatusChoices(models.TextChoices):
...
+            raise RetryStep("Request throttled") from e
+        except (
+            self._health_imaging_client.exceptions.InternalServerException
+        ) as e:
+            raise RetryStep("Server side error") from e
+        except self._health_imaging_client.exceptions.ConflictException:
+            # todo: check status and maybe retry
+            raise
+        else:
+            return delete_results
```
koopmant: I've separated the health imaging client from the model again so I can use it without a model instance in the task `cleanup_healthimaging_image_set`.

app/grandchallenge/cases/tasks.py
```diff
@@ -543,3 +545,42 @@ def import_dicom_to_healthimaging(*, dicom_imageset_upload_pk):
...
+
+@acks_late_micro_short_task
+@transaction.atomic
+def revert_image_set_to_initial_version(*, image_set_id, version_id):
+    health_imaging_wrapper = HealthImagingWrapper()
+    health_imaging_wrapper.update_image_set_metadata(
+        image_set_id=image_set_id,
+        version_id=version_id,
+        metadata={"revertToVersionId": "1"},
+    )
```
koopmant: I tested this just now, but the image set just disappeared and is no longer active now. Not sure what went wrong yet.

app/grandchallenge/cases/models.py
koopmant: Having seen some failures, the answer is yes. The job status can be `completed`, yet there are no generated image sets...

app/grandchallenge/cases/models.py
```diff
@@ -948,3 +1019,97 @@ def start_dicom_import_job(self):
...
+            raise DICOMImportJobValidationError(
+                "Instance already exists. This should never happen!"
+            )
+        self.convert_image_set_to_internal(image_set)
+
+    def handle_failed_job(self, *, event):
+        job_summary = self.get_job_summary(event=event)
+        self.internal_failure_log = self.get_job_output_failure_log(
+            job_summary=job_summary
+        )
```
koopmant, line 1098: Currently only doing this after some checks that the job has failed, but so far the failure log always exists (if the job succeeded it's just empty). Maybe we should always get the failure log?

app/grandchallenge/cases/tasks.py
koopmant: update on this here: https://github.com/DIAGNijmegen/rse-roadmap/issues/431#issuecomment-3252131195

app/grandchallenge/cases/models.py
koopmant: You were right, thanks

app/grandchallenge/cases/models.py
jmsmkn: I don't think so, I think you're more likely to get into an infinite loop that way.

app/grandchallenge/cases/exceptions.py
```diff
@@ -0,0 +1,14 @@
+class DICOMImportJobError(Exception):
+
+    def __init__(self, message, message_details=None):
+        super().__init__(message)
+        self.message = message
+        self.message_details = message_details
+
```
amickan: You're not actually using these, are you?

app/grandchallenge/cases/models.py
amickan: It makes sense to me to only get the failure log for failed imports, so I think what you're doing now is good. Or are you saying that there can be a failure log for successful imports?

app/grandchallenge/cases/models.py
```diff
@@ -996,3 +1056,106 @@ def deidentify_user_uploads(self):
...
+        self.save()
+        self.delete_image_sets(job_summary=job_summary)
+        job_id = job_summary["jobId"]
+        raise DICOMImportJobFailedError(message=f"Import job {job_id} failed")
```
amickan: Maybe add the pk of the DICOMImageSetUpload instance to the error message as well?

app/grandchallenge/cases/exceptions.py
koopmant: right, not anymore. I'll clean this up, thanks!

app/grandchallenge/cases/models.py
koopmant: I have blurred the definition of failed a little bit. I'm now also calling it when I get a completed job without generated image sets. I suppose it would have been possible to get a failure and success log when we upload multiple image sets, but we only accept single image sets. You're right, I don't think we can get a failure log and still have a generated image set. So we should be good.

app/grandchallenge/cases/models.py
```diff
@@ -996,3 +1056,108 @@ def deidentify_user_uploads(self):
...
+    def revert_image_set_to_initial_version(*, image_set):
+        from grandchallenge.cases.tasks import (
+            revert_image_set_to_initial_version,
+        )
+
+        revert_image_set_to_initial_version(
+            image_set_id=image_set["imageSetId"],
+            version_id=image_set["imageSetVersion"],
+        )
+
```
amickan: You're calling the tasks here synchronoulsy rather than asynchronously.

amickan: and why handle deletion of each image set in a seperate task and not pass all image sets in one go?

koopmant: Thanks, silly mistake.

koopmant: > why handle deletion of each image set in a seperate task and not pass all image sets in one go?

I figured this would be slightly easier for if the task does not succeed and it needs to retry. There should be only one image set anyway, so it doesn't matter too much I guess.

amickan: Oh right. Makes sense.

koopmant: Should be fixed now.

app/grandchallenge/cases/models.py
```diff
@@ -869,6 +933,7 @@ class DICOMImageSetUpload(UUIDModel):
     )
 
     error_message = models.TextField(blank=True, default="")
+    internal_failure_log = models.TextField(blank=True, default="")
```
jmsmkn: Is the log that we get from HI only text or is it structured JSON?

app/grandchallenge/cases/models.py
```diff
@@ -846,6 +852,64 @@ class DICOMImageSetUploadStatusChoices(models.TextChoices):
...
+        try:
+            return self._health_imaging_client.delete_image_set(
+                imageSetId=image_set_id,
+                datastoreId=settings.AWS_HEALTH_IMAGING_DATASTORE_ID,
+            )
+        except self._health_imaging_client.exceptions.ThrottlingException as e:
+            raise RetryStep("Request throttled") from e
+        except ClientError as err:
+            logger.error(
+                "Couldn't delete image set. Here's why: %s: %s",
```
jmsmkn: Adding the code and message to the overall report will create unique errors in sentry. It is better to do:

```python3
except ClientError:
  logger.error("Couldn't delete image set", exc_info=True)
```

Then all of the information, including the exception, will be in Sentry and grouped appropriately.

app/grandchallenge/cases/models.py
```diff
@@ -846,6 +852,64 @@ class DICOMImageSetUploadStatusChoices(models.TextChoices):
...
+            return self._health_imaging_client.update_image_set_metadata(
+                imageSetId=image_set_id,
+                datastoreId=settings.AWS_HEALTH_IMAGING_DATASTORE_ID,
+                latestVersionId=str(version_id),
+                updateImageSetMetadataUpdates=metadata,
+                force=force,
+            )
+        except ClientError as err:
+            logger.error(
+                "Couldn't update image set metadata. Here's why: %s: %s",
```
jmsmkn: Same here.

app/grandchallenge/cases/models.py
```diff
@@ -944,12 +1006,10 @@ def start_dicom_import_job(self):
...
-            return self._health_imaging_client.start_dicom_import_job(
-                jobName=self._import_job_name,
-                datastoreId=settings.AWS_HEALTH_IMAGING_DATASTORE_ID,
-                dataAccessRoleArn=settings.AWS_HEALTH_IMAGING_IMPORT_ROLE_ARN,
-                inputS3Uri=self._import_input_s3_uri,
-                outputS3Uri=self._import_output_s3_uri,
+            return self._health_imaging_wrapper.start_dicom_import_job(
+                job_name=self._import_job_name,
+                input_s3_uri=self._import_input_s3_uri,
+                output_s3_uri=self._import_output_s3_uri,
```
jmsmkn: Isn't this a bit of a leaky abstraction now? A wrapper has now been implemented so that the Job does not need to be concerned about the implementation, but only some of the methods, settings and error handling have been moved across. I don't think that is clear to me what the responsibility of each class is, and where functionality is located or new functionality should be implemented.

With the other class responsible for these things I would expect to be able to do:

```python3
    @property
    def _health_imaging_wrapper(self):
        if self.__health_imaging_wrapper is None:
            self.__health_imaging_wrapper = HealthImagingWrapper(pk=pk). # add any other necessary setup here
        return self.__health_imaging_wrapper

   def start_dicom_import_job(self):
       return self._health_imaging_wrapper.start_dicom_import_job()
```

Maybe `start_dicom_import_job` isn't even necessary on the model then. Maybe there should be another PR later with a fixup?

app/grandchallenge/cases/models.py
```diff
@@ -996,3 +1056,114 @@ def deidentify_user_uploads(self):
...
+            raise DICOMImportJobValidationError(
+                "New instance is not primary: "
+                "metadata conflicts with already existing instance."
+            )
+        if not image_set["imageSetVersion"] == 1:
+            self.revert_image_set_to_initial_version(image_set=image_set)
+            raise DICOMImportJobValidationError(
+                "Instance already exists. This should never happen!"
+            )
+        return image_set
```
jmsmkn: nit: allow your code to breathe, group logical units together, makes it easier to break things down and follow.

```suggestion
        job_summary = self.get_job_summary(event=event)
        
        if job_summary["numberOfGeneratedImageSets"] == 0:
            self.handle_failed_job(event=event)
        elif job_summary["numberOfGeneratedImageSets"] > 1:
            self.delete_image_sets(job_summary=job_summary)
            raise DICOMImportJobValidationError(
                "Multiple image sets created. Expected only one."
            )
            
        image_set = job_summary["imageSetsSummary"][0]
        
        if not image_set["isPrimary"]:
            self.delete_image_sets(job_summary=job_summary)
            raise DICOMImportJobValidationError(
                "New instance is not primary: "
                "metadata conflicts with already existing instance."
            )
            
        if not image_set["imageSetVersion"] == 1:
            self.revert_image_set_to_initial_version(image_set=image_set)
            raise DICOMImportJobValidationError(
                "Instance already exists. This should never happen!"
            )
            
        return image_set
```

jmsmkn (commented):
Looks good, couple of minor things.

app/grandchallenge/cases/models.py
```diff
@@ -996,3 +1056,114 @@ def deidentify_user_uploads(self):
...
+        if job_summary["numberOfGeneratedImageSets"] == 0:
+            self.handle_failed_job(event=event)
+        if job_summary["numberOfGeneratedImageSets"] > 1:
+            self.delete_image_sets(job_summary=job_summary)
+            raise DICOMImportJobValidationError(
+                "Multiple image sets created. Expected only one."
+            )
+        image_set = job_summary["imageSetsSummary"][0]
+        if not image_set["isPrimary"]:
+            self.delete_image_sets(job_summary=job_summary)
```
jmsmkn, line 1080: Does the task here actually get run? If an error is raised then the transaction is not committed and the celery task not scheduled as it is wrapped in `on_commit`.

Same for `revert_image_set_to_initial_version` below.

app/grandchallenge/cases/models.py
koopmant: it's ndjson

jmsmkn: Okay, then if you want to keep it all convert it to a JSON array and store it in a JSON field. That will be more efficient than text and it's then properly searchable.

app/grandchallenge/cases/models.py
koopmant: Haha, I actually thought about this and still got to the wrong conclusion. 😅 
I'll fix it.

app/grandchallenge/cases/models.py
koopmant: That's much better, thanks!

app/grandchallenge/cases/models.py
koopmant: Good to know, thanks!

app/grandchallenge/cases/models.py
koopmant: I think you are referring to this method in particular:
```python
    def start_dicom_import_job(self):
        """
        Start a HealthImaging DICOM import job.
        """
        try:
            return self._health_imaging_wrapper.start_dicom_import_job(
                job_name=self._import_job_name,
                input_s3_uri=self._import_input_s3_uri,
                output_s3_uri=self._import_output_s3_uri,
            )
        except ClientError as e:
            if e.response["Error"]["Code"] in {
                "ThrottlingException",
                "ServiceQuotaExceededException",
            }:
                raise RetryStep from e
            else:
                raise
```

I agree having the error handling here is not right. I think it should be moved to the task instead. 

Apart from that, there are no other methods where the model calls the health imaging client, so everything else is moved I think. 

As for the set up of the wrapper with a pk, I don't see why that would be beneficial. I figured that class should just be a wrapper around the client. It should be generic and it shouldn't store pk's like that.

app/grandchallenge/cases/models.py
koopmant: Wait no, I think the tasks do get called. We catch the errors in `handle_event`, either there are no errors and the model status is updated to `COMPLETED` or there are errors and we handle them and update the model status to `FAILED`.

jmsmkn: Okay, good!

app/grandchallenge/cases/models.py
koopmant: I have moved the error handling to the task, which I think is slightly cleaner, but does not really address your point. 
Do you think it's okay as it is, or should we have a separate PR to clean up the wrapper?

jmsmkn: I think that this looks better now, but all that the wrapper is really doing is saving a call to:

```
        client = boto3.client(
            "medical-imaging",
             region_name=settings.AWS_DEFAULT_REGION,
        )
```

on each interaction. Is it really worth a separate class for that given that there is no shared state between the methods?

koopmant: It also simplifies the function calls by filling in the variables from settings and by converting the keywords for the arguments to camel case, required by the client. The client is dynamically defined, so it doesn't give you auto-complete. This makes it a bit annoying that the keywords need to be camel case.

jmsmkn: That's true but I don't think that it is worth the extra complexity. The abstraction still leaks as the caller needs to be able to handle all of the weirdness of boto errors anyway, which is the most tedious part.

koopmant: All right, removed it. Although a bit less convenient, the code somehow feels cleaner now. 😄

app/grandchallenge/cases/models.py
```diff
@@ -996,3 +1018,122 @@ def deidentify_user_uploads(self):
...
+        output_uri = event["outputS3Uri"]
+        parsed = urlparse(output_uri)
+        bucket = parsed.netloc
+        key = parsed.path.lstrip("/") + "job-output-manifest.json"
+
+        # Try to get the manifest.
+        try:
+            obj = self._s3_client.get_object(Bucket=bucket, Key=key)
+        except self._s3_client.exceptions.NoSuchKey as e:
+            raise RetryStep("Manifest not (yet) found for job output") from e
```
jmsmkn: Couldn't this put you in an infinite loop if HI never produces the manifest? Did you face the issue of the manifest not being present when expected in development? If not, it's probably not worth explicitly handling.

koopmant: No, I haven't seen this happen. I'll remove it!

jmsmkn (approved):
LGTM!

PR #4283, opened 2025-09-03
Add DICOM upload widget

amickan:

### DIFF
--- a/app/grandchallenge/uploads/widgets.py
+++ b/app/grandchallenge/uploads/widgets.py
@@ -27,14 +27,6 @@ class UserUploadWidgetMixin:
         )
 
 
-class DICOMUserUploadMixin(UserUploadWidgetMixin):
-    class Media(UserUploadWidgetMixin.Media):
-        js = UserUploadWidgetMixin.Media.js + (
-            "vendored/dcmjs/build/dcmjs.min.js",
-            "js/file_preprocessors.js",
-        )
-
-
 class UserUploadSingleWidget(UserUploadWidgetMixin, HiddenInput):
     pass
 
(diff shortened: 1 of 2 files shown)


app/grandchallenge/uploads/widgets.py
```diff
@@ -27,6 +27,14 @@ class Media:
         )
 
 
+class DICOMUserUploadMixin(UserUploadWidgetMixin):
+    class Media(UserUploadWidgetMixin.Media):
+        js = UserUploadWidgetMixin.Media.js + (
```
jmsmkn: This doesn't look correct to me, would it work with multiple mixins (e.g. like in `DICOMUserUploadMultipleWidget` if `MultipleHiddenInput` also defined `Media`)? Is this the recommended way to do it in Django? 

Why is this a mixin anyway? It is only used once, so maybe move it to the concrete implementation and if this is the recommended way ensure that `Media` inherits both subclasses?

amickan: You're right. The default behavior is for the subclass to inherit all media assets from its parent. What I did here breaks that automatic inheritance. And yes, I can move everything to the concrete implementation.

jmsmkn (approved):
LGTM!

app/grandchallenge/uploads/widgets.py
```diff
@@ -36,3 +36,11 @@ def get_context(self, name, value, attrs):
         context = super().get_context(name, value, attrs)
         context["widget"]["attrs"]["multiple"] = True
         return context
+
+
+class DICOMUserUploadMultipleWidget(UserUploadMultipleWidget):
```
HarmvZ, line 41: Shouldn't this restrict `allowed_file_types`?

amickan: It already does through the socket.

PR #4284, opened 2025-09-03
Add deidentification step

amickan:
This adds the de-identification step to the import task. For now, just a placeholder method that copies rather than download and uploads files. 

### DIFF
--- a/app/grandchallenge/cases/tasks.py
+++ b/app/grandchallenge/cases/tasks.py
@@ -6,6 +6,7 @@ from shutil import rmtree
 from tempfile import TemporaryDirectory
 
 from billiard.exceptions import SoftTimeLimitExceeded, TimeLimitExceeded
+from botocore.exceptions import ClientError
 from celery import signature
 from celery.utils.log import get_task_logger
 from django.conf import settings
@@ -524,7 +525,7 @@ def _check_post_processor_result(*, post_processor_result, image):
         raise RuntimeError("Created image IDs do not match")
 
 
-@acks_late_micro_short_task(retry_on=(LockNotAcquiredException, RetryStep))
+@acks_late_2xlarge_task(retry_on=(LockNotAcquiredException, RetryStep))
 @transaction.atomic
 def import_dicom_to_healthimaging(*, dicom_imageset_upload_pk):
     upload = lock_model_instance(
@@ -539,12 +540,22 @@ def import_dicom_to_healthimaging(*, dicom_imageset_upload_pk):
         )
 
     try:
-        upload.deidentify()
+        upload.deidentify_user_uploads()
+        upload.status = DICOMImageSetUploadStatusChoices.STARTED
+        upload.save()
+        upload.start_dicom_import_job()
+    except ClientError as e:
+        if e.response["Error"]["Code"] == "ThrottlingException":
+            raise RetryStep("Request throttled") from e
+        elif e.response["Error"]["Code"] == "ServiceQuotaExceededException":
+            raise RetryStep("Service quota exceeded") from e
+        else:
+            upload._mark_failed(
+                error_message="An unexpected error occurred", exc=e
+            )
     except Exception as e:
         upload._mark_failed(
             error_message="An unexpected error occurred", exc=e
         )
     else:
-        upload.status = DICOMImageSetUploadStatusChoices.STARTED
-        upload.save()
-        upload.start_dicom_import_job()
+        upload.user_uploads.all().delete()
--- a/app/grandchallenge/cases/tasks.py
+++ b/app/grandchallenge/cases/tasks.py
@@ -533,13 +533,18 @@ def import_dicom_to_healthimaging(*, dicom_imageset_upload_pk):
         pk=dicom_imageset_upload_pk,
     )
 
-    # the status to check here will ultimately have to be something like DICOMImageSetUploadStatusChoices.DEIDENTIFIED
     if not upload.status == DICOMImageSetUploadStatusChoices.INITIALIZED:
         raise RuntimeError(
-            "Upload is not ready for importing into HealthImaging."
+            "Upload is not ready for de-identification and importing into HealthImaging."
         )
 
-    upload.status = DICOMImageSetUploadStatusChoices.STARTED
-    upload.save()
-
-    upload.start_dicom_import_job()
+    try:
+        upload.deidentify()
+    except Exception as e:
+        upload._mark_failed(
+            error_message="An unexpected error occurred", exc=e
+        )
+    else:
+        upload.status = DICOMImageSetUploadStatusChoices.STARTED
+        upload.save()
+        upload.start_dicom_import_job()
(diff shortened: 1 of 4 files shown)


app/grandchallenge/cases/tasks.py
```diff
@@ -533,13 +534,28 @@ def import_dicom_to_healthimaging(*, dicom_imageset_upload_pk):
...
+        else:
+            upload._mark_failed(
+                error_message="An unexpected error occurred", exc=e
+            )
+    except Exception as e:
+        upload._mark_failed(
+            error_message="An unexpected error occurred", exc=e
+        )
+    else:
+        upload.user_uploads.all().delete()
```
amickan: Unsure about this, but in order to only run the user upload deletion step if both the deidentification and the importing succeeded without errors, it seemed easiest to me to move the error handling that I had implemented in `start_dicom_import_job` up to the celery task here. This does assume that we want to handle the client errors from both `deidentify_user_uploads` and `start_dicom_import_job` in the same way though.

amickan: The alternative would be to have each of these functions implement their own error handling (like I did before with `start_dicom_import_job`) and have both of these return `True` when successful and then check these return codes here before calling `upload.user_uploads.all().delete()`. In that case, we would not need a try ... except statement in the celery task at all.

app/grandchallenge/cases/tasks.py
```diff
@@ -539,12 +540,22 @@ def import_dicom_to_healthimaging(*, dicom_imageset_upload_pk):
...
+            )
     except Exception as e:
         upload._mark_failed(
             error_message="An unexpected error occurred", exc=e
         )
     else:
-        upload.status = DICOMImageSetUploadStatusChoices.STARTED
-        upload.save()
-        upload.start_dicom_import_job()
+        upload.user_uploads.all().delete()
```
jmsmkn: We definitely want this in an `else` block and not `finally`?

amickan: We only want to clean up if both tasks completed successfully. Putting it in the finally block would mean it would execute regardless, so I would say `else` is the block this should go in, or am I missing something?

amickan: Well technically, we could already clean up after the deidentification step.

amickan: I could also move the cleaning to the deidentify method. I could then leave the error handling for import jobs in that task, nice and contained, but the celery task would still need to catch and handle all errors from the deidentification step since we want the task to return early and not move on to starting the import job if de-id fails.

app/grandchallenge/cases/tasks.py
jmsmkn: I don't really get it. The error handling here seems complicated, this should be minimal. Two separate things are happening here - deidentification and starting a job - so I would imagine you would have two separate function calls, each with their own error handling. The user uploads can be deleted after the de-identification is complete.

app/grandchallenge/cases/tasks.py
amickan: I made a suggestion for that, see latest changeset

app/grandchallenge/cases/tasks.py
amickan: I have updated this based on the other conversation

app/tests/cases_tests/test_tasks.py
```diff
@@ -225,12 +225,39 @@ def test_import_dicom_to_healthimaging_updates_status_when_successful(
...
+    di_upload = DICOMImageSetUploadFactory()
+    with patch.object(
+        DICOMImageSetUpload, "start_dicom_import_job"
+    ) as mocked_method:
+        with patch.object(
+            DICOMImageSetUpload,
+            "deidentify_user_uploads",
+            side_effect=Exception(),
+        ):
+            with django_capture_on_commit_callbacks(execute=True):
```
jmsmkn: Don't know if this would look any better but you can avoid deep nesting with:

```python3
with A() as a, B() as b:
    SUITE
```

or

```python3
with (
    A() as a,
    B() as b,
):
    SUITE
```

https://docs.python.org/3.13/reference/compound_stmts.html#with

app/grandchallenge/cases/tasks.py
```diff
@@ -533,13 +533,18 @@ def import_dicom_to_healthimaging(*, dicom_imageset_upload_pk):
...
+    try:
+        upload.deidentify_user_uploads()
+    except Exception as e:
+        upload._mark_failed(
+            error_message="An unexpected error occurred", exc=e
+        )
+    else:
+        upload.status = DICOMImageSetUploadStatusChoices.STARTED
+        upload.save()
+        upload.start_dicom_import_job()
```
jmsmkn: The PR looks good but maybe we can make the exception handling more consistent. `deidentify_user_uploads` looks good, it does what it says and is not concerned about also managing the state of the upload. But `start_dicom_import_job` has two cases where it marks as failed, one of which is duplicated so it will be called twice (once from models.py, once from tasks.py).

In short, I don't think `self._mark_failed` should be called from `upload.start_dicom_import_job`, and that should only be handled in `tasks.py`. Putting it altogether we would get:

```python3
    def start_dicom_import_job(self):
        """
        Start a HealthImaging DICOM import job.
        """
        try:
            return self._health_imaging_client.start_dicom_import_job(
                jobName=self._import_job_name,
                datastoreId=settings.AWS_HEALTH_IMAGING_DATASTORE_ID,
                dataAccessRoleArn=settings.AWS_HEALTH_IMAGING_IMPORT_ROLE_ARN,
                inputS3Uri=self._import_input_s3_uri,
                outputS3Uri=self._import_output_s3_uri,
            )
        except ClientError as e:
            if e.response["Error"]["Code"] in {"ThrottlingException","ServiceQuotaExceededException"}:
                raise RetryStep from e
            else:
                raise
```

and

```python3
    try:
        upload.deidentify_user_uploads()
    except RetryStep:
        raise
    except Exception as e:
        upload._mark_failed(
            error_message="An unexpected error occurred", exc=e
        )
    else:
        upload.status = DICOMImageSetUploadStatusChoices.STARTED
        upload.save()
```

What do you think?

jmsmkn: Note that `"ServiceQuotaExceededException"` also needed to be corrected.

amickan: Ok, that makes sense. Thanks! 

My only concern would be that eventually as we add more error handling to either function, we would need to separate the calls into 2 separate try-except blocks to handle exceptions differently, but we can do that when we get there. And maybe it's not necessary if we raise specific enough (custom) exceptions.

jmsmkn: You don't want too many custom errors, really they're only there for how the exception should be handled. For components along side `RetryStep` we have `ComponentException` which contains information that will get pushed to the user. There shouldn't be a need for many more than that, we either retry, fail expectedly, or fail unexpectedly.

amickan: Ah ok, then indeed it won't be an issue. Thanks for explaining.

jmsmkn (approved):

Issue #4280, opened 2025-09-01
Remove deprecated API fields


pkcakeout:
The following fields are deprecated:

```
    ("HyperlinkedImage", "patient_id"),
    ("HyperlinkedImage", "patient_name"),
    ("HyperlinkedImage", "patient_birth_date"),
    ("HyperlinkedImage", "patient_age"),
    ("HyperlinkedImage", "patient_sex"),
    ("HyperlinkedImage", "study_date"),
    ("HyperlinkedImage", "study_instance_uid"),
    ("HyperlinkedImage", "series_instance_uid"),
    ("HyperlinkedImage", "study_description"),
    ("HyperlinkedImage", "series_description"),
    ("Question", "help_text"),
    ("Question", "question_text"),
    ("Question", "empty_answer_confirmation_label"),
    ("ReaderStudy", "help_text"),
    ("ReaderStudy", "title"),
    ("DisplaySet", "title"),
```

All can be removed after release 2026.03.

jmsmkn:
There are more fields than just help text, see the linked PR.

Also requires an update to GCAPI models.

PR #4285, opened 2025-09-03
Add safe endpoints for reader studies

pkcakeout:
Related to #4280

Add a series of escaped endpoints to the REST API for more user-defined text.

## TODOs

- [ ] add a few tests


app/grandchallenge/reader_studies/models.py
```diff
@@ -1,6 +1,7 @@
 from math import ceil
 
 from actstream.models import Follow
+from bleach import clean
```
jmsmkn: We shouldn't use this directly, especially as bleach is deprecated, we have a template tag that handles the configuration.

```suggestion
from grandchallenge.core.templatetags.bleach import clean, md2html
```

app/grandchallenge/reader_studies/models.py
```diff
@@ -1872,6 +1881,20 @@ def is_image_type(self):
     def get_absolute_url(self):
         return self.reader_study.get_absolute_url() + "#questions"
 
+    @property
+    def help_text_safe(self) -> str:
+        """The cleaned help text allowing some HTML tags"""
+        return clean(self.help_text)
```
jmsmkn:
```suggestion
        return md2html(self.help_text)
```

pkcakeout: Wanted to do this, but there is a problem with this: We are processing and embedding these strings in CIRRUS in a tabular/enum format. When using md2html, the paragraphs become proper `<p>` blocks and some other formatting is done as well. If we then render this out in our template:

~~~~~
<div class="pt-3">
    <h2>{{reader_study.title}}</h2>
    <p><a href={{url}} target="_blank">View on Grand Challenge</a></p>
    <p>{{unread_case_count}} case{{unread_case_count|pluralize}} left to read.</p>
    <p>{{reader_study.description}}</p>
    <p>{{reader_study.help_text}}</p>

    <hr>
    {% if questions_with_helptext %}
        <h4>Questions</h4>
        <dl>
            {% for question in questions %}
                {% if question.help_text %}
                    <dt>{{question.question_text}}</dt>
                    <dd>{{question.help_text}}</dd>
                {% endif %}
            {% endfor %}
        </dl>
    {% endif %}
</div>
~~~~~

it messes things up considerably by having the compact list expand vertically.

I was hoping to park these issues for now and simply do a clean. Then we can do the full markdown conversion later...

jmsmkn: Okay, that's fine, but use GCs version of clean.

app/grandchallenge/core/templatetags/bleach.py
```diff
@@ -14,11 +14,16 @@
 
 
 @register.filter
-def clean(html: str):
+def clean(html: str, no_tags=False):
```
jmsmkn:
```suggestion
def clean(html: str, *, no_tags=False):
```

jmsmkn:
> Ok, did it! @jmsmkn - I added a "no_tags" mode to suppress and escape all tags for some of the fields. Tell me if that is alright. Tags in (at least) the `question_text` are problematic.

Yes looks good, what about the existing serialised attributes? Shouldn't they point at the new cleaned versions until they can be removed?

pkcakeout:
> Yes looks good, what about the existing serialised attributes?

Had to think about this for a sec, but yes indeed. That will plug the XSS holes for the previous version at the cost of messing up the formatting in CIRRUS. Probably a good deal.

pkcakeout:
I dug my teeth into this one now, but I found out that _maybe_ it makes more sense to leave the `title_safe` endpoints there right next to the `title`-endpoints.

When trying to unify them, I got errors because, for example, `DisplaySet.title` was not settable anymore through the API. That makes sense, because if we interpolate the value with a `clean(...)`, it is not settable anymore. However, we likely want to have many of these values still be settable via the API? In that case it is, imo, very valuable to leave the getters for exact representations intact. Then you can do (pseudocode):

```
reader_study.title.get() -> "<script>evil_code</script>"
reader_study.title.patch("<script>evil_code</script>")
reader_study.title_safe.get() -> "evil_code"
reader_study.title_safe.patch("<script>evil_code</script>") -> ERROR, not settable
```

The non-safe endpoint is for writing the actual value, and "safe" is meant for endpoints that render the value. That would leave automated uploads and patching defining reader studies fully through scripts as features intact...?

jmsmkn:
Hmm, good point. Not sure how to handle that without having a separate read only seralizer.

app/grandchallenge/reader_studies/forms.py
```diff
@@ -120,6 +121,33 @@
 }
 
 
+def verify_cleaned_html(
```
jmsmkn: This is a little bit unique when it doesn't need to be. It would be better implemented as a validator, then it will fit in with the rest of the ecosystem: https://docs.djangoproject.com/en/5.2/ref/validators/

app/grandchallenge/reader_studies/serializers.py
```diff
@@ -138,6 +158,13 @@ class Meta:
             "description",
             "index",
         )
+        validators = [
```
jmsmkn: These are validators for the whole model, this only needs to be implemented on individual fields, then the  field and  readable_field_name arguments can be dropped.

pkcakeout:
I combined everything in a generic model-field validator. It is great, makes way more sense, imo, thanks for pointing me into the correct direction. Learning about django is kind of fun... however... always a "however". Maybe you can help out with a thing I cannot fix for some reason:

https://github.com/comic/grand-challenge.org/blob/6353b1c3019209f10f947d29c4defa15cf9d60c3/app/grandchallenge/reader_studies/models.py#L987-L992

This title-validator does not work... I have a _vague_ idea why this is. Looking at test-uses of updating display sets, it looks to me like the display-set update is forked out into a celery task(?):

https://github.com/comic/grand-challenge.org/blob/3b5c00b34bde8640efa1b6f313aa1aeb94916a74/app/tests/reader_studies_tests/test_views.py#L546-L550

So likely, some celery task is spawned and then that one sets the data on `DisplaySet.title` directly... that is what I _guess_ is happening here. So my validator gets bypassed, because the form never uses it as a validator as it is not a ModelForm. What is actually the best way to fix this?

app/grandchallenge/reader_studies/migrations/0072_alter_readerstudy_title.py
jmsmkn, line 1: Please consolidate new migration files

app/grandchallenge/reader_studies/forms.py
```diff
@@ -957,6 +957,11 @@ class Meta:
     def model(self):
         return self.base_obj.civ_set_model
 
+    def clean(self):
+        cleaned_data = super().clean()
+
+        return cleaned_data
+
```
jmsmkn: This can now be removed

app/grandchallenge/reader_studies/models.py
```diff
@@ -193,6 +219,12 @@ class ReaderStudy(
     a set of questions on a set of images (cases).
     """
 
+    title = models.CharField(
+        _("title"),
+        max_length=255,
+        validators=[CleanedHtmlValidator(no_tags=True)],
+    )
+
```
pkcakeout: Also a _slight_ question about this one: Is this ok? This _redefines_ the title-field from `TitleSlugDescriptionModel` to add the validator. I find it a bit messy. Is there a better way?

pkcakeout: Actually, maybe there is... what about this construct:

```.py
def init_from_serialized(cls_args_kwargs, **overload_kwargs):
    cls, args, kwargs = cls_args_kwargs
    return cls(*args, **{**kwargs, overload_kwargs})

title = init_from_serialized(
    TitleSlugDescriptionModel.title.deconstruct(),
    validators=[validators=[CleanedHtmlValidator(no_tags=True)]]
)
```

Or is that getting too opaque with too much meta programming?

jmsmkn: I agree but think this is it. This is the only way if you consistently want validation in all places - admin, forms, views, drf, etc. Otherwise you would need to add it in the specific places.

jmsmkn: Yes too meta, we pretty much never change the title base class anyway.

pkcakeout: So you would want the `init_from_serialized`?

jmsmkn:
Ah yeah, for some reason DisplaySetUpdate is using a FormView and not an update view, and likely is missing the call to full clean. @amickan any idea why? Looks like this is the same for the create view too.

amickan:
Ah actually I think I know why. The form uses `UniqueTitleCreateFormMixin` which overwrites the `title` field on the form and the new field does not have the validator.

pkcakeout:
Ah, back to square one where we clean when rendering out to the API endpoints? Will work, too, yes!

jmsmkn:
> Ah, back to square one where we clean when rendering out to the API endpoints? That version will work, too, yes!

Indeed, as you saw it was basically impossible to ensure that the validation is always applied so the cleaning needs to be done on read. That is then consistent with how we do it in the rest of the codebase. CIRRUS now needs to move to using the `_safe` endpoints exclusively, the old fields will then be removed after the 2025.10 release.

jmsmkn (approved):

PR #4287, opened 2025-09-04
Allow multiple user uploads through FlexibleImageWidget

amickan:
I accidentally discovered a bug in the `FlexibleImageWidget` that would prevent multiple user uploads from being passed along to the forms where it's used. This bug must have been there for quite a while. I'm surprised no one noticed this.

### DIFF
--- a/app/grandchallenge/cases/widgets.py
+++ b/app/grandchallenge/cases/widgets.py
@@ -120,8 +120,9 @@ class FlexibleImageField(MultiValueField):
             ModelMultipleChoiceField(queryset=upload_queryset, required=False),
         ]
 
-        # The `current_value` is added to the widget attrs to display in the initial dropdown.
-        # We get the object so we can present the user with the image name rather than the pk.
+        # The `current_value` is added to the widget attrs to display an appropriate
+        # title in the initial dropdown and to add the pk(s) as hidden input
+        # in the widget template.
         self.current_value = None
         if initial:
             if isinstance(initial, ComponentInterfaceValue):


amickan:
Tests need a fix still and I also discovered that the code fails when an image is preselected. Need to investigate that further.

jmsmkn:
As a general principle the code shouldn't change just to satisfy a test when the test can instead be re-written to match the real world. If the code is never going to encounter a dictionary in real life, then there is no need to handle it just to satisfy the test and the test is not representative of the real world.

koopmant:
I'm not sure if I fully understand it yet. But looking at [`FileInput.value_from_datadict()`](https://github.com/django/django/blob/stable/4.2.x/django/forms/widgets.py#L442) and [`MultipleHiddenInput.value_from_datadict()`](https://github.com/django/django/blob/stable/4.2.x/django/forms/widgets.py#L404) it seems to make sense to handle both types. 

But if we only ever process standard form data here, than the function shouldn't need to handle both.

app/grandchallenge/cases/widgets.py
```diff
@@ -57,23 +59,37 @@ def __init__(
...
+                item = value[0]
+                if item in ImageWidgetChoices.names:
+                    return [None, None]
+                if Image.objects.filter(pk=item).exists():
+                    return [str(item), None]
+                # can also just be a single UserUpload
+                return [None, value]
+            else:
+                # can be a list of UserUploads or
+                # a single image UUID (when an image is preselected)
```
koopmant: This is a remainder from before you added isinstance(value, UUID) I think. It cannot be a single UUID here, I believe.

koopmant:
I wonder if we also have to handle both types then in `AdditionalInputsMixin` and `MultipleCIVForm`, where we use `getlist()` as well.

amickan:
> I'm not sure if I fully understand it yet. But looking at [`FileInput.value_from_datadict()`](https://github.com/django/django/blob/stable/4.2.x/django/forms/widgets.py#L442) and [`MultipleHiddenInput.value_from_datadict()`](https://github.com/django/django/blob/stable/4.2.x/django/forms/widgets.py#L404) it seems to make sense to handle both types.
> 
> ~But if we only ever process standard form data here, than the function shouldn't need to handle both.~ I think the widget should handle normal dict as well if we pass initial data or redisplay the form after validation errors.

This is why I was asking. I wasn't sure whether we could in real life also encounter dictionaries and under which circumstances. So you figured out that when passing initial data or when redisplaying the form after validation errors that that data is in fact a normal dictionary? I had not tried the latter, but though that when passing in initial data it was also a QueryDict. I'll do some tests.

app/grandchallenge/cases/widgets.py
amickan: The comment makes no sense and needs to go but `value` can still be a list of UserUploads, so this is still needed.

amickan:
> I think the widget should handle normal dict as well if we pass initial data or redisplay the form after validation errors.

I can't reproduce a situation where the data is not a QueryDict. I have tried all possible interactions with the widget in development I could think of. I think it makes sense since in real life form data will always be submitted through a POST request and so the data will be a QueryDict. 

Then again, all Django base widgets gracefully handle both dict and QueryDict, and since that is the case, and maybe that is reason enough to folow the same pattern in our custom widget?

koopmant:
I think that might be there for backwards compatibility only. Not sure

app/grandchallenge/cases/views.py
```diff
@@ -168,6 +172,28 @@ def get(self, request, *args, **kwargs):
...
+                        [
+                            HiddenInput().render(
+                                name=prefixed_interface_slug,
+                                value=upload,
+                            )
+                            for upload in uploads
+                        ]
+                    )
+
             raise Http404(f"Selected image {current_value_pk} not found")
```
koopmant:
```suggestion
            raise Http404(f"Selected image {current_value_list} not found")
```

app/grandchallenge/cases/widgets.py
```diff
@@ -57,32 +59,44 @@ def __init__(
...
+        if isinstance(value, UUID):
+            # when an image or user upload is preselected as current_value
+            if Image.objects.filter(pk=value).exists():
+                return [value, None]
+            elif UserUpload.objects.filter(pk=value).exists():
                 return [None, [value]]
-        else:
-            return [None, None]
+
+        raise RuntimeError("Unrecognized value type")
```
koopmant, line 87: There is a possibility that it's a UUID, but the object does not exist. I think we can just return `[None, None]` in that case. That's what we do in the `FlexibleFileWidget` anyway.

app/grandchallenge/cases/widgets.py
```diff
@@ -113,23 +127,40 @@ def __init__(
...
-                self.current_value = image
+                self.current_value = [image]
             elif upload := get_object_if_allowed(
                 model=UserUpload,
                 pk=initial,
                 user=user,
                 codename="change_userupload",
             ):
-                self.current_value = upload
+                self.current_value = [upload]
```
koopmant, line 166: I thought these lines should be deleted now, but that depends on the answer to [this comment](https://github.com/comic/grand-challenge.org/pull/4287#discussion_r2325936030).

app/grandchallenge/cases/widgets.py
```diff
@@ -57,32 +59,44 @@ def __init__(
...
+                return [None, value]
             else:
+                # can be a list of UserUploads
+                return [None, value]
+
+        if isinstance(value, UUID):
+            # when an image or user upload is preselected as current_value
+            if Image.objects.filter(pk=value).exists():
+                return [value, None]
+            elif UserUpload.objects.filter(pk=value).exists():
```
koopmant, line 82: I'm almost afraid to ask, but this is not an option right?

app/grandchallenge/cases/widgets.py
```diff
@@ -113,23 +127,40 @@ def __init__(
...
+                        if upload := get_object_if_allowed(
+                            model=UserUpload,
+                            pk=i,
+                            user=user,
+                            codename="change_userupload",
+                        ):
+                            uploads.append(upload)
+                if len(uploads) != 0:
+                    self.current_value = uploads
+
```
koopmant, line 153: There is still something weird with `initial` and `current_value`. If we filter the current value list on permissions, I thought we should also update the list in `initial`, but `initial` is not relevant anymore in this particular case (I think?)
Looking at this comment:
```
# The `current_value` is added to the widget attrs to display in the initial dropdown.
# We get the object so we can present the user with the image name rather than the pk.
```
It seems we only get the objects so we can get the title; but the comment is misleading because we also use `current_value.pk` as hidden input in the widget template. 

It makes my head spin, but I think we should do something about the comment, at least.

app/grandchallenge/components/forms.py
```diff
@@ -118,15 +118,12 @@ def init_additional_inputs(self, *, inputs):
...
-                    and input.kind == ComponentInterface.Kind.ANY
-                ):
-                    # interfaces for which the data can be a list need
-                    # to be retrieved with getlist() from the QueryDict
+                try:
                     initial = self.data.getlist(prefixed_interface_slug)
-                else:
-                    initial = self.data[prefixed_interface_slug]
+                    if len(initial) == 1:
+                        initial = initial[0]
```
koopmant, line 124: We have to do this because we later use the type to figure out what case we are in, right? I.e. if it's still a list then, we definitely know we have the multiple uploads case... 
I'm asking because with this a single upload is also converted to "not a list"; so they come in two flavours still, correct?

koopmant (changes requested):
I have thoroughly checked it, so we can hopefully fix "everything" now that we are touching it. But yes, the time this has taken shows that the whole thing is just too complex. 

I wonder if we should completely separate uploading data from input selection. Maybe we should write a pitch for that.

app/grandchallenge/cases/widgets.py
amickan: That makes sense.

app/grandchallenge/components/forms.py
amickan: If we were only dealing with uploads here, we wouldn't need to do this and could leave out the lines that you highlighted [earlier](https://github.com/comic/grand-challenge.org/pull/4287#discussion_r2325869454). But since we can get either uploads, images or any value here in the form (also strings, numbers etc from the json sockets), we need to do this to pass the single value items down as single values. That indeed means that a single UserUpload is passed downstream not as a list, and that in turn makes the section you referenced earlier in the widget necessary. 

I hope that makes sense.

app/grandchallenge/cases/widgets.py
amickan: The `UserUploadMultipleWidget` (and the underlying `MultipleHiddenInput`) expect the value to be a list, so UserUploads always have to be passed to the widget as a list, hence the brackets here (and above). 

In the forms, as I explain below, we now try to retrieve all data with `getlist` and hence get lists for everything. For all data types except uploads, the data should not be passed on as a list or else downstream processing in the other widgets fails. As a proxy for "non-upload-data", we just return single value lists as single values. For a single user upload this means that we need to turn it back into a list before moving on, hence this snippet here. The alternative would be to check the type in the form and pass single user uploads on as a list, but that complicates the form logic, so I think I prefer to keep the logic here. What do you think?

amickan: I see now that we don't actually turn the single user upload back into a list at this point. That is not necessary at this point because that's what our `decompress` method on the `FlexibleImageWidget` does.... 
The explanation above about why we need to unlist single list items in the form still holds though.

app/grandchallenge/cases/widgets.py
amickan: It is possible when we instantiate the form manually with data, i.e. in the tests.

app/grandchallenge/cases/widgets.py
amickan: Hm, good point. `initial` is still passed on as initial value to the field at the end here, but the field has its own permission filter on since it subsets the values it gets based on the queryset we pass it:

```python
list_fields = [
            ModelChoiceField(queryset=image_search_queryset, required=False),
            ModelMultipleChoiceField(queryset=upload_queryset, required=False),
        ]
```

So actually then I wonder if the permission check we do here is necessary? The field itself takes care of the necessary permission checking automatically.

amickan:
Thanks for the thorough review, Thomas! I agree that this widget has become way too complex to work with. It would be best to separate the widgets, and hence drop the `MultiWidget` and `MultiField` approach. Unsure how we would go about presenting the choice between the widgets to the user, but that can be worked out in a pitch.

app/grandchallenge/cases/widgets.py
koopmant: Right, so if `UserUpload`s are always a list at this point, even if it's just one, we don't have to check if the pk is from a UserUpload.

app/grandchallenge/components/forms.py
koopmant: That is what I thought. Thanks for the clarification!

app/grandchallenge/cases/widgets.py
koopmant: All right.

app/grandchallenge/cases/widgets.py
amickan: Well, but they are not always a list. If it's a single UserUpload, it won't be a list at this point.

koopmant: All right, then they should stay.

app/grandchallenge/cases/widgets.py
koopmant: I think it's still necessary because we use the `current_value` as hidden input to pass it on.

amickan: True, but that will end up being passed through the field again in the end - or should at least.

amickan: I would say we leave it as is at the moment. Would you agree?

koopmant: yes

koopmant: Should we remove the comment?

amickan: I have updated the comment. Does that look good to you?

PR #4290, opened 2025-09-05
Remove creator must be verified option

jmsmkn:
Our challenges now require an algorithm phase so being verified is not optional.


amickan (approved):
LGTM

PR #4296, opened 2025-09-10
Fill in deidentify data step

amickan:

### DIFF
--- a/app/grandchallenge/cases/models.py
+++ b/app/grandchallenge/cases/models.py
@@ -1012,7 +1012,15 @@ class DICOMImageSetUpload(UUIDModel):
         deid = DicomDeidentifier(
             study_instance_uid_suffix=self.study_instance_uid,
             series_instance_uid_suffix=self.series_instance_uid,
-            assert_unique_value_for=["StudyInstanceUID", "SeriesInstanceUID"],
+            assert_unique_value_for=[
+                "StudyInstanceUID",
+                "SeriesInstanceUID",
+                "PatientID",
+                "StudyID",
+                "StudyDate",
+                "AccessionNumber",
+                "SeriesNumber",
+            ],
         )
         for upload in self.user_uploads.all():
             with (
--- a/app/tests/cases_tests/test_models.py
+++ b/app/tests/cases_tests/test_models.py
@@ -580,7 +580,15 @@ def test_deidentify_files_processes_all_user_uploads(mocker):
     mock_deid.assert_called_once_with(
         study_instance_uid_suffix=di_upload.study_instance_uid,
         series_instance_uid_suffix=di_upload.series_instance_uid,
-        assert_unique_value_for=["StudyInstanceUID", "SeriesInstanceUID"],
+        assert_unique_value_for=[
+            "StudyInstanceUID",
+            "SeriesInstanceUID",
+            "PatientID",
+            "StudyID",
+            "StudyDate",
+            "AccessionNumber",
+            "SeriesNumber",
+        ],
     )
     assert mock_download.call_count == len(uploads)
     assert mock_upload.call_count == len(uploads)


app/grandchallenge/cases/models.py
```diff
@@ -842,6 +844,26 @@ def save(self, *args, **kwargs):
...
+def generate_dicom_id_suffix(pk: str, suffix_type: str) -> int:
+    """
+    Generates a unique numerical suffix for a DICOM UID based on the primary key
+    and a string to differentiate the type (e.g., 'study' or 'series').
+    """
+    seed = f"{pk}-{suffix_type}"
+
+    digest = hashlib.sha512(seed.encode("utf8")).digest()
+    return int.from_bytes(digest[:14])
+
```
amickan: Having tried this out, I actually think that relying on the pk will cause issues down the line. For that to work, a new model instance needs to be saved first with a default empty string. The DB uniqueness constraint will prevent saving other instances at the same time then. So with multiple concurrent DICOM uploads at the same time, we'd get lots of `IntegrityErrors`. 

I think we will need to create a UUID before the first save and pass that to `generate_dicom_id_suffix` to populate the two fields before calling `super().save()`.

app/grandchallenge/cases/tasks.py
```diff
@@ -543,6 +546,8 @@ def import_dicom_to_healthimaging(*, dicom_imageset_upload_pk):
         upload.start_dicom_import_job()
     except RetryStep:
         raise
+    except RejectedDICOMFileError as e:
+        upload._mark_failed(error_message=e.justification)
```
amickan: Chris confirmed that the justifications for these are all save to return to the user.

uv.lock
```diff
@@ -24,42 +24,42 @@ dependencies = [
     { name = "propcache" },
     { name = "yarl" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/9b/e7/d92a237d8802ca88483906c388f7c201bbe96cd80a165ffd0ac2f6a8d59f/aiohttp-3.12.15.tar.gz", hash = "sha256:4fc61385e9c98d72fcdf47e6dd81833f47b2f77c114c29cd64a361be57a763a2", size = 7823716, upload-time = "2025-07-29T05:52:32.215Z" }
-wheels = [
```
amickan: I used `uv add grand-challenge-dicom-de-identifier` to add the new dependency. Not sure sure why all these other lines here have been changed and whether we need the `upload-time` information?

app/grandchallenge/cases/models.py
```diff
@@ -961,16 +1008,38 @@ def start_dicom_import_job(self):
...
+                infile.seek(0)
+
+                deid.deidentify_file(infile, output=outfile)
+
+                outfile.seek(0)
+                self._s3_client.upload_fileobj(
+                    Fileobj=outfile,
+                    Bucket=settings.AWS_HEALTH_IMAGING_BUCKET_NAME,
+                    Key=f"{self._input_key}/{upload.filename}",
+                )
```
amickan: This set-up means that we would potentially upload one or multiple files in vain to the HI bucket if deidentification errors out on a subsequent file. The alternative would be to write all files to disk and wait for all to be deidentified before uploading.

amickan: If we use `SpooledTemporaryFile` I also still need to set a `max_size`. I see in other places that we use 1GB as the cut-off (on the large worker). Since we have two files in memory here, I guess 500MB would be an ok cut-off?

app/grandchallenge/cases/tasks.py
amickan: See https://github.com/DIAGNijmegen/rse-roadmap/issues/419#issuecomment-3274109165

app/grandchallenge/cases/models.py
jmsmkn: Why would it default to an empty string? The pk UUID is automatically generated on initialisation, so the suffix can be set pre-save.

With 32 digits using 10 characters per position the number of unique strings is going to be 10^32. I really don't see how collisions will be a problem with such a space.

uv.lock
jmsmkn: It looks like your version of uv is not up to date.

app/grandchallenge/cases/models.py
jmsmkn: Don't all of the files need to be sent to the method at the same time so that they are de-identified consistently? How does it check for consistent metadata in this case?

jmsmkn: Oh I see. This is fine, the prefix can be wiped on error, same with removing the users uploads.

app/grandchallenge/cases/models.py
```diff
@@ -870,6 +892,19 @@ class DICOMImageSetUpload(UUIDModel):
 
     error_message = models.TextField(blank=True, default="")
 
+    study_instance_uuid = models.CharField(
```
jmsmkn: Why are different identifiers required? They should also be set without a default.

app/grandchallenge/cases/models.py
amickan: Ok, somehow IK was convinced that `self.pk` was not available until after saving, but I see now that it is. My bad. In that case, never mind.

app/grandchallenge/cases/models.py
amickan: The default can go, that was because of a misconception (see above). But what do you mean by different identifiers? Are you saying we should use the same for study and series uuid?

app/grandchallenge/cases/models.py
jmsmkn: That is true when using auto field pks, in that case the pk comes from the database. For UUIDs they are generated in Python.

app/grandchallenge/cases/models.py
jmsmkn: Yes, what would be the problem with that if we are having 1 Image Set = 1 Study = 1 Series?

amickan: I thought that these 2 needed to be different for a valid Dicom file, but I have not actually checked whether HI cares. I will do that and report back.

amickan: Why I thought that: https://dicom.nema.org/dicom/2013/output/chtml/part05/chapter_9.html See note that says `For example the same UID value cannot be used to identify both a study instance (Study Instance UID) and a series instance (Series Instance UID) within that study or a different study. `

amickan: But HealthImaging might not care, so we might be able to use the same for both.

uv.lock
amickan: That is the problem indeed.

app/grandchallenge/cases/models.py
amickan: Healthimaging does not care. A file with the same study and series UID imports just fine. So do we want to go with just one id then? Not sticking to the DICOM standard might lead to issues downstream if HI ever changes the way they do things in that regard.

jmsmkn: Ah, then yes, if DICOM says they need to be different then they should be. The idea is that the resulting DICOM should still pass the validators after deidentification, so the replacements that you send to the deidentification methods should also conform to the DICOM standard. I don't know if chris' method checks that.

amickan: Ok, I will then stick to what I had here, 2 separate UUIDs.

uv.lock
jmsmkn: Something still seems wrong with the resulting lock file as other dependencies have been touched. We anyway need to generate it fresh from main so both this and the target branch need to be up to date first.

amickan: Yeah I've noticed but all ways I've tried, it updates other dependencies as well. I'll revisit after I've added tests and after the target branch have been rebased on main.

app/grandchallenge/cases/models.py
```diff
@@ -1065,3 +1056,9 @@ def deidentify_user_uploads(self):
         )
 
         self.user_uploads.all().delete()
+
+    def delete_input_files(self):
+        self._s3_client.delete_objects(
+            Bucket=settings.AWS_HEALTH_IMAGING_BUCKET_NAME,
+            Delete={"Objects": [{"Key": self._input_key}]},
```
jmsmkn: This won't actually delete anything, right? I think the `_input_key` property is incorrectly named, this is actually a prefix as it doesn't represent a single object.

All the files each have their own key under this prefix. You would then need to delete all the objects that have a certain prefix. You need to be extremely careful here that you don't touch anything else.

uv.lock
amickan: I think it's good now - right?

jmsmkn: Strange, it does it for me:

```
(grand-challenge-org) ➜  app git:(431-store-dicom-in-hi) uv add grand-challenge-dicom-de-identifier         
Resolved 199 packages in 719ms
Prepared 2 packages in 186ms
Installed 2 packages in 1ms
 + grand-challenge-dicom-de-id-procedure==2025.9.0
 + grand-challenge-dicom-de-identifier==0.0.1
(grand-challenge-org) ➜  app git:(431-store-dicom-in-hi) ✗ uv --version
uv 0.7.3 (Homebrew 2025-05-07)
```

app/grandchallenge/cases/models.py
amickan: Hm I tried this and it did delete all files for the specific upload and left the other data in tact. I will double check once again to be sure.

jmsmkn: See the implementation here, even this is incomplete:

https://github.com/comic/grand-challenge.org/blob/9f88a1d1b3612493e6e769aa288a57126aeb226e/app/grandchallenge/components/backends/base.py#L578-L615

jmsmkn: You could of course keep a record of [all the keys created](https://github.com/comic/grand-challenge.org/blob/8dbb239183b72071a45f607c6c3626b44fb63aa3/app/grandchallenge/cases/models.py#L1032) and pass those to `Delete={"Objects": [{"Key": key} for key in created_keys]},`

app/grandchallenge/cases/models.py
```diff
@@ -961,16 +999,38 @@ def start_dicom_import_job(self):
...
+                )
+                infile.seek(0)
+
+                deid.deidentify_file(infile, output=outfile)
+
+                outfile.seek(0)
+                self._s3_client.upload_fileobj(
+                    Fileobj=outfile,
+                    Bucket=settings.AWS_HEALTH_IMAGING_BUCKET_NAME,
+                    Key=f"{self._input_key}/{upload.filename}",
```
jmsmkn: `upload.filename` is user generated so should not be used internally. For the uploads bucket we give the objects the key `upload.pk`, maybe health imaging needs to know that it is dcm:

```suggestion
                    Key=f"{self._input_key}/{upload.pk}.dcm",
```

app/grandchallenge/cases/models.py
jmsmkn: Actually, I don't think that tracking the created keys will handle the case where the task is interrupted, and then fails on the second time around. So listing the prefix and then deleting is more robust.

amickan: You are right that what I implemented does not work. I mistook a never completed upload for a successful deletion yesterday... I will list all objects and delete them instead.

amickan: What do you mean when you say that the snippet you shared is not complete? It looks pretty complete to me.

jmsmkn: The object list response is paginated and we only get the first page (mea culpa with `objects_list["IsTruncated"]`), so you need something like this: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/paginator/ListObjectsV2.html

I don't know how many objects can then be sent to delete objects so the requests to that would also need to be batched.

amickan: Aha! I had not thought of that at all. I'll work on a batching solution.

amickan: The docs suggest that the max number items returned from `list_objects` and the max number of keys to provide to the `delete_objects` is 1000 (see: https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html and https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html). The current upload widget allows only 100 files max to be uploaded in one go, so that would mean that we would never reach that limit, making pagination redundant for the DICOM upload case at least. 

Do we expect users to upload more than 100 files for a single DICOM series?

amickan: For the Sagemaker usecase, batching is still necessary, so I will still update the common function I created to use the paginator. But we need to think about the upload limit for dicom files as well

jmsmkn: Yes, for DICOM we're going to need to increase the number of files, probably up to around 2000, maybe more for WSI or full body scans.

amickan: I made a suggestion for a batching solution. I tested it with a small batch size (the paginator accepts a `MaxKeys` argument) and confirmed that it works, but I think for real life we can just go with the default of 1000.

jmsmkn: Yes, 1000 is fine. You probably want to check the bucket and prefix to ensure you're not accidentally deleting everything in the protected bucket...

app/grandchallenge/cases/models.py
```diff
@@ -867,6 +869,26 @@ def save(self, *args, **kwargs):
...
+
+def generate_dicom_id_suffix(pk: str, suffix_type: str) -> int:
+    """
+    Generates a unique numerical suffix for a DICOM UID based on the primary key
+    and a string to differentiate the type (e.g., 'study' or 'series').
+    """
+    seed = f"{pk}-{suffix_type}"
+
+    digest = hashlib.sha512(seed.encode("utf8")).digest()
+    return int.from_bytes(digest[:14])
```
koopmant: Need to remove the type hinting here. And I think it would make sense to return a string instead of an integer here.

app/tests/cases_tests/test_models.py
```diff
@@ -488,3 +490,94 @@ def test_handle_completed_job_generated_image_set(
...
+    assert di_upload.study_instance_uid == generate_dicom_id_suffix(
+        pk=di_upload.pk, suffix_type="study"
+    )
+    assert di_upload.series_instance_uid == generate_dicom_id_suffix(
+        pk=di_upload.pk, suffix_type="series"
+    )
+
+
+@pytest.mark.django_db
+def test_deidentify_user_uploads_idempotency(mocker, settings):
```
koopmant, line 508: Do we need another marker to indicate `_deidentify_files` has been called? What happens if that fails half-way and the task runs again?

amickan: If deidentification fails half way, we delete all files on S3. If we then reschedule the task, deidentification would run again because the marker file would not have been created. 

At the moment, the only error I catch from the deidentifier is `RejectedDICOMFileError`. It could of course also fail with an unexpected error and I think it would make sense to clean up the files on S3 in that case as well, so I have added the `delete_input_files` call to the broad exception block as well.

amickan: By the way, `delete_input_files` should probably also be called when an import jobs completes.

amickan:
This is ready for re-review. The deidentifier has been updated and the code now also passes along the study and series UIDs.

app/grandchallenge/cases/models.py
```diff
@@ -982,16 +1009,31 @@ def start_dicom_import_job(self):
...
 
     def _deidentify_files(self):
-        # this will need to iterate over all linked UserUploads
-        # for each, download the file from the uploads bucket, run the
-        # deidentifier and upload the deidentified file to the HI bucket
-        # for the time being just copy each upload to the HI bucket
+        deid = DicomDeidentifier(
+            study_instance_uid_suffix=self.study_instance_uid,
+            series_instance_uid_suffix=self.series_instance_uid,
+            assert_unique_value_for=["StudyInstanceUID", "SeriesInstanceUID"],
```
jmsmkn: Don't all of the HI fields need to be checked here?

amickan: I think that makes sense, yes. I thought that all data within a certain series would be the same for these other fields, but I might be wrong about that and in any case, it's better to be safe and include them all here.

app/tests/cases_tests/test_models.py
amickan: @koopmant if this PR gets merged first, don't forget to call `delete_input_files` on import success. If yours gets merged first, I will add it.

jmsmkn (approved):
LGTM!

PR #4297, opened 2025-09-10
Dicom image set model

koopmant:


app/grandchallenge/cases/models.py
```diff
@@ -1129,6 +1169,52 @@ def revert_image_set_to_initial_version(*, image_set):
...
+                height=0,
+            )
+
+        Image.objects.create(
+            dicom_image_set=dicom_image_set,
+            **image_metadata,
+        )
+
+    def get_image_metadata(self, *, image_set_id):
+        response = self._health_imaging_client.get_image_set_metadata(
```
koopmant: I imagine we already have this data from the deidentification step at this point, but otherwise we could get it from health imaging like this.

amickan: We know the following: StudyDate, AccessionNumber, PatientID, StudyID, SeriesNumber, StudyInstanceUID and SeriesInstanceUID. But the metadata from HI will of course contain more. I think it makes sense to fetch it from HI. That way we know for sure that our ImageSet metadata corresponds to the HI metadata. We could even check that the above defined values are what they should be and reject otherwise?

app/grandchallenge/cases/models.py
```diff
@@ -325,6 +327,32 @@ def image_file_path(instance, filename):
...
+        unique=True,
+        help_text="The ID of the image set in AWS Health Imaging.",
+    )
+    dicom_image_set_upload = models.OneToOneField(
+        to="DICOMImageSetUpload",
+        editable=False,
+        null=True,
+        on_delete=models.SET_NULL,
+        related_name="dicom_image_set",
+    )
```
amickan: Do we need to know which upload this was linked to?

app/grandchallenge/cases/models.py
```diff
@@ -1106,29 +1149,78 @@ def handle_failed_job(self, *, event):
...
+            logger.error(e, exc_info=True)
+            image_kwargs = dict(
+                width=0,
+                height=0,
+            )
+
+        Image.objects.create(
+            dicom_image_set=dicom_image_set,
+            **image_kwargs,
+        )
```
amickan: I think you need to set the image's name as well. That will be used in challenge evaluations, for example.

app/grandchallenge/cases/models.py
```diff
@@ -1106,29 +1149,78 @@ def handle_failed_job(self, *, event):
...
+        return dict(
+            height=instance_metadata["Rows"],
+            width=instance_metadata["Columns"],
+            modality=ImagingModality.objects.get(
+                modality=series_metadata["Modality"]
+            ),
+            study_instance_uid=study_metadata["StudyInstanceUID"],
+            study_date=datetime.strptime(
+                study_metadata["StudyDate"], "%Y%m%d"
+            ),
```
amickan: If you're setting StudyDate, I would also set the other values that we force override, right?

And looking at the Image model, isn't `color_space` also required?

app/grandchallenge/cases/models.py
koopmant: Not really I think. For now it was convenient, because you get a link to the upload instance in the admin. That's the only reason I added it. Would that also be useful later?
I believe there is a similar link to raw upload sessions, the `origin` field on `Image`, but I'm not sure if that serves a specific purpose.

app/grandchallenge/cases/models.py
koopmant: > the other values that we force override

Do you mean these?
```
STUDY_DATE = "20000101"
ACCESSION_NUMBER = ""
PATIENT_ID = "ANONYMOUS"
STUDY_ID = ""
SERIES_NUMBER = "0"
```

koopmant: > And looking at the Image model, isn't color_space also required?

It is, I'm not sure yet how I want to get the correct one. I think I want to rely on `panimg` maybe, so we always do the same thing.

jmsmkn: The study date isn't worth storing as it is set to the same value for everything by the de-identifier.

app/grandchallenge/cases/models.py
```diff
@@ -1106,29 +1149,78 @@ def handle_failed_job(self, *, event):
...
+            image_set_id=image_set_id,
+            dicom_image_set_upload=self,
+        )
+
+        try:
+            metadata = self.get_image_set_metadata(image_set_id=image_set_id)
+            image_kwargs = self.convert_image_set_metadata_to_image_kwargs(
+                metadata=metadata
+            )
+        except Exception as e:
```
jmsmkn: This is super broad. Why create an image in all of these scenarios?

app/grandchallenge/cases/models.py
jmsmkn: We use the origin for tracking who uploaded what, so it is useful to have here.

app/grandchallenge/cases/models.py
koopmant: True. I just wasn't sure yet what to do in this case. Does it make sense to go to health imaging for the metadata at all? We have already seen the images in the deidentification step, so maybe we can get the information at that point?

app/grandchallenge/cases/models.py
koopmant: Ah, so we'll just leave it blank on the `Image`?

jmsmkn: I think so, also width and height may change per instance so those values may not make sense.

app/grandchallenge/cases/models.py
koopmant: Yes, in panimg we use `f"{ds.StudyInstanceUID}-{index}"` where index is a counter:
```
# Since we might need to combine multiple images with different
# series instance UID (in 4D images), we cannot use the series
# as the unique file name - instead, we use the study instance
# uid and a counter (index)
```

I'm not sure if the deidentifier will solve this problem (probably not?). But either way, we only have one set each time. So I suppose we could use `{StudyInstanceUID}-0`, or use the SeriesInstanceUID.

jmsmkn: That will be a de-identified value. Probably we need to get the user to provide the image name on upload.

app/grandchallenge/cases/models.py
jmsmkn: We need to check what metadata actually needs to be set. I think a lot of these fields are actually deprecated or not actually used anywhere. Perhaps it's better to create something minimal and then see what is actually needed.

app/grandchallenge/cases/models.py
koopmant: I've added the get_image_set_metadata method, but will come back to comparing values in a later PR.

app/grandchallenge/cases/models.py
```diff
@@ -325,6 +326,32 @@ def image_file_path(instance, filename):
...
+    image_set_id = models.CharField(
+        max_length=32,
+        unique=True,
+        help_text="The ID of the image set in AWS Health Imaging.",
+    )
+    dicom_image_set_upload = models.OneToOneField(
+        to="DICOMImageSetUpload",
+        editable=False,
+        null=True,
+        on_delete=models.SET_NULL,
```
jmsmkn: Do we want this to be nullable? I think that we always would want information about the origin of the image set, so maybe keep it and protect it on deletion?

app/grandchallenge/cases/models.py
```diff
@@ -325,6 +326,32 @@ def image_file_path(instance, filename):
...
+    )
+
+
+@receiver(post_delete, sender=DICOMImageSet)
+def delete_image_set(*_, instance: DICOMImageSet, **__):
+    from grandchallenge.cases.tasks import delete_healthimaging_image_set
+
+    on_commit(
+        delete_healthimaging_image_set.signature(
+            kwargs=dict(image_set_id=instance.image_set_id)
```
jmsmkn: The style in the codebase is to use these constructions for dictionaries, is there a reason to change?

```suggestion
            kwargs={"image_set_id": instance.image_set_id}
```

app/grandchallenge/cases/models.py
```diff
@@ -1106,29 +1148,47 @@ def handle_failed_job(self, *, event):
     def delete_image_sets(*, job_summary):
         from grandchallenge.cases.tasks import delete_healthimaging_image_set
 
-        image_sets = job_summary.get("imageSetsSummary", [])
-        for image_set in image_sets:
+        for image_set_summary in job_summary.get("imageSetsSummary", []):
             on_commit(
                 delete_healthimaging_image_set.signature(
-                    kwargs=dict(image_set_id=image_set["imageSetId"])
+                    kwargs=dict(image_set_id=image_set_summary["imageSetId"])
```
jmsmkn: Same here

app/grandchallenge/cases/models.py
```diff
@@ -1106,29 +1148,47 @@ def handle_failed_job(self, *, event):
...
             revert_image_set_to_initial_version,
         )
 
         on_commit(
             revert_image_set_to_initial_version.signature(
                 kwargs=dict(
-                    image_set_id=image_set["imageSetId"],
-                    version_id=image_set["imageSetVersion"],
+                    image_set_id=image_set_summary["imageSetId"],
+                    version_id=image_set_summary["imageSetVersion"],
```
jmsmkn: And here.

app/grandchallenge/cases/models.py
```diff
@@ -1106,29 +1148,47 @@ def handle_failed_job(self, *, event):
...
             on_commit(
                 delete_healthimaging_image_set.signature(
-                    kwargs=dict(image_set_id=image_set["imageSetId"])
+                    kwargs=dict(image_set_id=image_set_summary["imageSetId"])
                 ).apply_async
             )
 
     @staticmethod
-    def revert_image_set_to_initial_version(*, image_set):
+    def revert_image_set_to_initial_version(*, image_set_summary):
```
jmsmkn, line 1261: The code passes around image_set_summary a lot, which is an external data structure. Would it be cleaner to extract the information we need into a dataclass which is then clearly used. This would remove the ambiguity between keys not existing and the data structure changing.

app/grandchallenge/cases/models.py
```diff
@@ -1056,8 +1096,8 @@ def handle_event(self, *, event):
             self.save()
 
     def handle_completed_job(self, *, event):
-        image_set = self.validate_image_set(event=event)
-        self.convert_image_set_to_internal(image_set=image_set)
+        image_set_id = self.validate_image_set(event=event)
```
jmsmkn: The validate_image_set function does quite a lot. Maybe the responsibilities should be split out a bit. I wouldn't have thought that a validate function would return anything - if we follow Djangos design then cleaning stuff just raises errors when there are problems.

app/grandchallenge/cases/models.py
```diff
@@ -1106,29 +1148,47 @@ def handle_failed_job(self, *, event):
...
+            dicom_image_set_upload=self,
+        )
+
+        Image.objects.create(
+            dicom_image_set=dicom_image_set,
+            # name="",  # todo: get name for set during upload
+            color_space=ColorSpace.GRAY.value,
+            width=-1,
+            height=-1,
+        )
```
jmsmkn: The new objects should be validated (make sure the caller of `convert_image_set_to_internal` can handle validation errors okay). This would need to be performed atomically, but I think we already are here? We shouldn't be forced to set any other values, `None` should be the value for things not filled out.

```suggestion
        dicom_image_set = DICOMImageSet(
            image_set_id=image_set_id,
            dicom_image_set_upload=self,
        )
        dicom_image_set.full_clean()
        dicom_image_set.save()

        image = Image(
            dicom_image_set=dicom_image_set,
            # name="",  # todo: get name for set during upload
        )
        image.full_clean()
        image.save()
```

app/grandchallenge/cases/models.py
```diff
@@ -1106,29 +1148,47 @@ def handle_failed_job(self, *, event):
...
+
+    def get_image_set_metadata(self, *, image_set_id):
```
jmsmkn: Is this method used?

koopmant: Not yet. I had removed it, but figured I'd keep it in there since we will use it soon.

jmsmkn: I think it's best to remove it for now, it will always be in the git history if we need it.

app/grandchallenge/cases/models.py
koopmant: Not really. I thought it would help with automatically renaming keywords, but I see I am mistaken there. I'll clean it up.

app/grandchallenge/cases/models.py
koopmant: Thanks! 

> make sure the caller of convert_image_set_to_internal can handle validation errors okay

Currently handling this by catching everything in `handle_event` and marking the upload as failed if an error occurs. Should I add explicit handling of validation errors?

app/grandchallenge/cases/models.py
koopmant: All right. I also removed the test fixture for `image_set_metadata`, that wasn't used either.

app/grandchallenge/cases/models.py
jmsmkn: Validation errors are safe to send to the user, so if we have a way of getting that message to them then that would be good.

koopmant: Those fields cannot be `None` at the moment. I've opted to update the `Image` model.

jmsmkn: Yes allowing `None` for height and width is the correct thing to do. For `color_space` allowing blanks would be the best thing as otherwise you could get two representations of nothing - `None` and `""`.

app/grandchallenge/cases/models.py
koopmant: Was struggling a bit with this, because I wanted to do it for the job summary as well, which makes the dataclasses nested and because there needs be conversion between camelCase and snake_case. Not sure if I made it too complicated...

app/grandchallenge/cases/tasks.py
```diff
@@ -569,7 +571,7 @@ def import_dicom_to_healthimaging(*, dicom_imageset_upload_pk):
         upload.save()
 
 
-@acks_late_micro_short_task(retry_on=(LockNotAcquiredException, RetryStep))
+@acks_late_micro_short_task(retry_on=(LockNotAcquiredException,))
```
koopmant, line 574: This task has some interactions with the s3 client. Should I raise RetryStep for things like `ThrottlingException`?

app/grandchallenge/cases/models.py
koopmant: > so if we have a way of getting that message to them then that would be good.

Saving them in `DICOMImageSetUpload.error_message` for the user.

app/grandchallenge/cases/tasks.py
jmsmkn: Throttling yes, but be careful of any others in case they lead to unsolvable loops (they're not infinite as we place a limit on the maximum number of retries but it is a couple of thousand).

koopmant: Actually, I can't find anything about throttling exceptions in the S3 API documentation. There is the `SlowDown` error response, but I think that is only related to AWS lambda.

jmsmkn (approved):
LGTM!

app/grandchallenge/cases/models.py
```diff
@@ -30,6 +34,9 @@
     ImageType,
     PatientSex,
 )
+from pydantic import ConfigDict, Field, field_validator
```
jmsmkn, line 37: Actually, I think that we'll need to add pydantic to our dependencies in `pyproject.toml` now. It's already installed but as a sub-dependency.

jmsmkn: I corrected the last commit. You will need to update your branch. Also, your version of `uv` is out of date so please update that.

jmsmkn (approved):
LGTM!

PR #4298, opened 2025-09-11
Update job locking

jmsmkn:
Attempts to only lock the jobs that require updating, and to request `no_key` locks on objects that are merely referenced.

See https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/612

### DIFF
--- a/app/grandchallenge/components/tasks.py
+++ b/app/grandchallenge/components/tasks.py
@@ -764,13 +764,13 @@ def check_operational_error(error):
         raise error
 
 
-def lock_for_utilization_update(*, algorithm_image_id):
+def lock_for_utilization_update(*, algorithm_image_pk):
     from grandchallenge.algorithms.models import AlgorithmImage
 
     # Lock the algorithm and algorithm image to avoid conflicts
     # when modifying JobUtilization objects
     try:
-        AlgorithmImage.objects.filter(pk=algorithm_image_id).select_related(
+        AlgorithmImage.objects.filter(pk=algorithm_image_pk).select_related(
             "algorithm"
         ).select_for_update(
             of=("self", "algorithm"),
@@ -929,7 +929,7 @@ def handle_event(*, event, backend):  # noqa: C901
         return
 
     if hasattr(job, "algorithm_image"):
-        lock_for_utilization_update(algorithm_image_id=job.algorithm_image_id)
+        lock_for_utilization_update(algorithm_image_pk=job.algorithm_image_id)
 
     try:
         executor.handle_event(event=event)
--- a/app/grandchallenge/evaluation/tasks.py
+++ b/app/grandchallenge/evaluation/tasks.py
@@ -217,7 +217,7 @@ def create_algorithm_jobs_for_evaluation(*, evaluation_pk, first_run):
         return
 
     lock_for_utilization_update(
-        algorithm_image_id=evaluation.submission.algorithm_image_id
+        algorithm_image_pk=evaluation.submission.algorithm_image_id
     )
 
     slots_available = min(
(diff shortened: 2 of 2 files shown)


app/grandchallenge/components/tasks.py
```diff
@@ -905,19 +911,16 @@ def handle_event(*, event, backend):  # noqa: C901
         return
 
     if hasattr(job, "algorithm_image"):
-        # Lock the algorithm image and algorithm to
-        # avoid conflicts when updating later
+        # Lock the algorithm to avoid conflicts when updating later
```
amickan: You are still locking both the algorithm image and the algorithm here, right?

app/grandchallenge/evaluation/tasks.py
```diff
@@ -211,20 +219,6 @@ def create_algorithm_jobs_for_evaluation(*, evaluation_pk, first_run):
         )
         return
 
-    # Lock the algorithm image and algorithm to avoid conflicts when updating later
-    lock_model_instance(
-        pk=evaluation.submission.algorithm_image.pk,
-        app_label="algorithms",
-        model_name="algorithmimage",
-        of=("self",),
-    )
```
amickan, line 220: Why is locking of the algorithm image no longer necessary?

jmsmkn: I think that it is still done in:

```python3
        of=(
            "self",
            "submission__algorithm_image__algorithm",
        ),
```

The Django docs say "specify the related objects you want to lock in select_for_update(of=(...)) using the same fields syntax as select_related()". Given that select related selects all of the intermediates, I assume that means that all intermediates are locked too.

https://docs.djangoproject.com/en/5.2/ref/models/querysets/#django.db.models.query.QuerySet.select_for_update

There isn't a good way to test this, and it is not harmful to deploy and find out. We have had this problem all summer and I only added this fix on Tuesday.

In fact, the fix is probably overkill as we should be able to use `no_key=True` as we're not actually updating the rows as the utilisation is created, only referencing them.

app/grandchallenge/components/tasks.py
jmsmkn: Yes, but given the exceptions we get it is algorithm that is the one causing problems, and we don't know the algorithm id without getting the algorithm image. May as well lock it as both are referenced in the job utilization.

app/grandchallenge/evaluation/tasks.py
amickan: Thanks for explaining. Makes sense.

app/grandchallenge/components/tasks.py
amickan: Yes, indeed, I was just asking because the comment was updated to suggest you are no longer locking the algorithm image. Implementation wise I agree that it makes sense to lock the image as well.

app/grandchallenge/components/tasks.py
```diff
@@ -758,6 +764,24 @@ def check_operational_error(error):
...
+        AlgorithmImage.objects.filter(pk=algorithm_image_pk).select_related(
+            "algorithm"
+        ).select_for_update(
+            of=("self", "algorithm"),
+            nowait=True,
+            no_key=True,
+        ).get()
+    except OperationalError as error:
+        check_operational_error(error)
+        raise
```
amickan, line 782: The only reason why you can't use `lock_model_instance` to do this, is because you also specify `no_key=True` in select_related here, right? Would it make sense to add that as an option to `lock_model_instance`?

jmsmkn: I didn't think that it was worth it as it gets complex if some of the rows need updating, and some do not. We can revisit it if this works and we have more demand for it.

PR #4301, opened 2025-09-16
Improve client deid procedure

HarmvZ:
This copies over some improvements from Chris' python based deidentifier:

- Set PatientIdentityRemoved (00120062) tag to YES.
- Set DeidentificationMethod (00120063) to use value multiplicity and append method with version and timestamp.
- Add grand challenge root UID as prefix for consistently replaced UIDs.

### DIFF
--- a/app/grandchallenge/uploads/static/js/file_preprocessors.js
+++ b/app/grandchallenge/uploads/static/js/file_preprocessors.js
@@ -1,5 +1,3 @@
-const GRAND_CHALLENGE_ROOT_UID = "1.2.826.0.1.3680043.10.1666.";
-
 /**
  * Returns a dummy value based on the DICOM Value Representation (VR).
  * This is used to replace sensitive data during de-identification.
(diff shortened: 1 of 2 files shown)


app/grandchallenge/uploads/static/js/file_preprocessors.js
```diff
@@ -1,3 +1,5 @@
+const GRAND_CHALLENGE_ROOT_UID = "1.2.826.0.1.3680043.10.1666.";
```
jmsmkn: Where does this value come from?

HarmvZ: It's from here: https://github.com/DIAGNijmegen/rse-grand-challenge-dicom-de-identifier/blob/b9fada902a1cfdaef54373602a61905291526adf/grand_challenge_dicom_de_identifier/deidentifier.py#L21-L22

I was in doubt about making it a constant in settings.py.

jmsmkn: Doesn't this one need to be different as it is a different process?

jmsmkn: Or, maybe the root uid doesn't matter at all as we're going to replace it anyway?

HarmvZ: > Doesn't this one need to be different as it is a different process?

I could generate a different one and use that. But indeed, if the serverside deidentifier works correctly, it should be replaced again so it does not really matter.

jmsmkn: Okay, then we can make it a completely random value.

HarmvZ: Good point, I reverted it so it uses a random value again.

PR #4304, opened 2025-09-17
Allow parallel submissions to the same challenge

jmsmkn:
Using the queued status is not ideal here, but it allows us to continue to monitor that the `create_algorithm_jobs_for_evaluation` task is working correctly by filtering on executing pre-requisites.

Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/642

### DIFF
--- a/app/grandchallenge/evaluation/tasks.py
+++ b/app/grandchallenge/evaluation/tasks.py
@@ -146,7 +146,7 @@ def prepare_and_execute_evaluation(*, evaluation_pk):  # noqa: C901
         evaluation.submission.user_upload.delete()
 
     if evaluation.submission.algorithm_image:
-        evaluation.status = Evaluation.EXECUTING_PREREQUISITES
+        evaluation.status = Evaluation.PENDING
         evaluation.save()
         on_commit(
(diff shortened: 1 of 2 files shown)


app/grandchallenge/evaluation/tasks.py
```diff
@@ -263,6 +267,24 @@ def create_algorithm_jobs_for_evaluation(*, evaluation_pk, first_run):
         )
 
     if first_run:
+        # Only go ahead if this is the users only active evaluation to
+        # prevent flooding of the system
+        user_has_other_active_evaluations = (
+            Evaluation.objects.filter(
+                submission__creator=evaluation.submission.creator,
+                status=Evaluation.EXECUTING_PREREQUISITES,
```
amickan, line 275: Why only filter on that specific status here and not use `.active()`?

jmsmkn: Because the other steps we can do in parallel without issue.

amickan: Ah yeah ok, makes sense.

amickan (approved):
LGTM

PR #4309, opened 2025-09-19
Dicom upload widget with name

amickan:
Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/430

This adds a `DICOMUploadWidget` that asks the user for both a file name and the files to upload. 

image

When an archive item or display set already has an image attached, we do not show the widget but instead a disabled field like this:

image

Both fields on the widget are required, i.e. the form cannot be submitted if the user has not supplied a file name next to the files.

### DIFF
--- a/app/grandchallenge/components/models.py
+++ b/app/grandchallenge/components/models.py
@@ -156,7 +156,7 @@ class InterfaceSuperKindChoices(models.TextChoices):
     IMAGE = "I", "Image"
     FILE = "F", "File"
     VALUE = "V", "Value"
-    
+
 
 class InterfaceKind:
     """Interface kind choices."""
@@ -718,10 +718,7 @@ class ComponentInterface(OverlaySegmentsMixin):
         value_required = True
         if self.kind == InterfaceKindChoices.BOOL:
             value_required = False
-        elif (
-            not self.is_image_kind
-            and not self.requires_file
-        ):
+        elif not self.is_image_kind and not self.requires_file:
             try:
                 self.validate_against_schema(value=None)
                 value_required = False


app/grandchallenge/cases/widgets.py
```diff
@@ -200,3 +205,88 @@ def compress(self, values):
...
+
+    def __init__(self, *args, user, initial, **kwargs):
+        upload_qs = filter_by_permission(
+            queryset=UserUpload.objects.all(),
+            user=user,
+            codename="change_userupload",
+        ).filter(status=UserUpload.StatusChoices.COMPLETED)
+
+        fields = [
+            CharField(),
```
amickan, line 254: Not sure if we want the name to end in `.dcm` and if so, if we should force the user to provide a file name with extension or just add it posthoc if not there.

app/grandchallenge/cases/widgets.py
```diff
@@ -200,3 +205,88 @@ def compress(self, values):
...
+            if isinstance(initial, ComponentInterfaceValue):
+                self.current_value = initial.image
+                # turn initial to the internal data type that this widget expects
+                initial = DICOMUploadWithName(
+                    name=initial.image.name, user_uploads=[initial.image.pk]
+                )
+            else:
+                raise RuntimeError(
+                    f"Unexpected initial value of type {type(initial)}"
+                )
```
amickan, line 284: This logic could also move to `InterfaceFormfield`. I kept it here so that it aligns with the other custom fields we have, but since those will likely change in the near future, we might want to move all `initial` handling logic out of the respective fields.

app/grandchallenge/components/forms.py
```diff
@@ -171,7 +172,12 @@ def __init__(self, *args, instance, base_obj, user, **kwargs):  # noqa C901
             interface = ComponentInterface.objects.filter(slug=slug).get()
             prefixed_interface_slug = f"{INTERFACE_FORM_FIELD_PREFIX}{slug}"
 
-            if prefixed_interface_slug in self.data:
+            # For interfaces that use the FlexibleImageWidget or FlexibleFileWidget
+            # we need to pass in the initial value explicitly, for all other
+            # fields the value in self.data is picked up automatically
+            if prefixed_interface_slug in self.data and (
+                interface.is_image_kind or interface.requires_file
+            ):
```
amickan, line 180: This is the revelation I had yesterday, Thomas. All of this extra logic here to retrieve form data for the fields is not necessary if you just use standard form fields. Even for the new `MultiValueField` for the `DICOMUploadField` we don't need any special logic here. It all just works out of the box. I think the goal for refactoring the `FlexibleFileWidget` and `FlexibleImageWidget` would be to have no more custom logic here in the form for those either.

app/grandchallenge/components/models.py
```diff
@@ -468,6 +468,11 @@ def __str__(self):
     def is_image_kind(self):
         return self.kind in InterfaceKind.interface_type_image()
 
+    @property
+    def is_dicom_image_kind(self):
+        # implementation up to Thomas
+        return False
```
amickan, line 474: If you want to check the branch out and test it, you need to actually add the new kind and implement this method. I did that for testing but did not push the code because that is part of https://github.com/DIAGNijmegen/rse-roadmap/issues/423

amickan: You can also return `True` here and have `is_image_kind` return False to test it, but if you do that, the `Add new socket` button / view does not work. So for a full test, you need this method here to be implemented.

amickan: Well actually you need a few more things in place to test fully. Here is the git diff for what needs to be added / changed (quick and dirty, bare bones implementation obviously): https://gist.github.com/amickan/f01c4636d157911688589d0f2e7d7167

app/grandchallenge/cases/widgets.py
koopmant: I don't think it needs to be `.dcm`, it's usually not a single file anyway, so it wouldn't make a lot of sense to me.

amickan: That's a good point. Then I think it's fine to leave at just a CharField without special validation.

app/grandchallenge/components/forms.py
amickan: Hm, not so sure anymore if it will be possible to get rid of all logic here for the FlexibleImage and FlexibleFile fields. As soon as you allow switching back and forth between different widgets for the same field within the form, you need to somehow figure out which field/widget to use for a given field at a given point in time. Anyway, not relevant for this PR. I will look into the options after I'm done with the core pitch.

app/tests/conftest.py
```diff
@@ -596,7 +596,17 @@ def get_interface_form_data(
     existing_data=False,
 ):
     ci = ComponentInterface.objects.get(slug=interface_slug)
-    form_data = {f"{INTERFACE_FORM_FIELD_PREFIX}{interface_slug}": data}
+    if ci.is_dicom_image_kind:
+        form_data = {
+            f"{INTERFACE_FORM_FIELD_PREFIX}{interface_slug}_dicom-image-name": data[
+                0
+            ],
+            f"{INTERFACE_FORM_FIELD_PREFIX}{interface_slug}_dicom-user-uploads": data[
```
koopmant: Take the definitions from `DICOMUploadWidgetSuffixes` here.

app/grandchallenge/components/forms.py
```diff
@@ -205,20 +209,42 @@ def __init__(self, *args, instance, base_obj, user, **kwargs):  # noqa C901
...
+    @staticmethod
+    def parse_slug(*, slug):
+        # remove prefix
+        interface_slug = slug[len(INTERFACE_FORM_FIELD_PREFIX) :]
+
+        # separate known suffix
+        for known_suffix in DICOMUploadWidgetSuffixes:
+            if interface_slug.endswith(f"_{known_suffix}"):
+                base_slug = interface_slug[: -len(f"_{known_suffix}")]
+                return base_slug
```
koopmant, line 241: Finally understand what you were talking about last week!

app/grandchallenge/cases/widgets.py
```diff
@@ -200,3 +205,105 @@ def compress(self, values):
...
+                ):
+                    self.current_value = image
+                    # turn initial to the internal data type that this widget expects
+                    initial = DICOMUploadWithName(
+                        name=image.name,
+                        user_uploads=[
+                            str(upload.pk)
+                            for upload in image.dicom_image_set.dicom_image_set_upload.user_uploads.all()
+                        ],
+                    )
```
koopmant: You can use the compress method here.

app/grandchallenge/cases/widgets.py
```diff
@@ -200,3 +205,105 @@ def compress(self, values):
...
+
+        super().__init__(
+            *args,
+            fields=fields,
+            initial=initial,
+            **kwargs,
+        )
+
+    def compress(self, values: list[str, QuerySet[UserUpload]]):
+        # receives a list of the outputs of the subfields’ clean() methods
```
koopmant: This comment can be removed.

app/grandchallenge/components/forms.py
```diff
@@ -205,20 +209,42 @@ def __init__(self, *args, instance, base_obj, user, **kwargs):  # noqa C901
...
+                ] = InterfaceFormFieldFactory(
                     interface=interface,
                     user=self.user,
                     required=False,
                     initial=current_value,
                 )
 
+    @staticmethod
+    def parse_slug(*, slug):
+        # remove prefix
```
koopmant: remove comment

app/grandchallenge/components/forms.py
```diff
@@ -205,20 +209,42 @@ def __init__(self, *args, instance, base_obj, user, **kwargs):  # noqa C901
...
+        interface_slug = slug[len(INTERFACE_FORM_FIELD_PREFIX) :]
+
+        # separate known suffix
```
koopmant: remove comment

app/grandchallenge/components/forms.py
koopmant: Not entirely sure yet, but I think we're supposed to handle it in `bound_data()` and `prepare_value()`.

app/grandchallenge/cases/widgets.py
```diff
@@ -200,3 +205,105 @@ def compress(self, values):
...
+        # in forms that include this field
+        widgets = {
+            DICOMUploadWidgetSuffixes[0]: DICOMFileNameInput(),
+            DICOMUploadWidgetSuffixes[1]: DICOMUserUploadMultipleWidget(),
+        }
+        super().__init__(widgets, attrs)
+
+    def decompress(self, value: DICOMUploadWithName):
+        # takes a DICOMUploadField value and breaks it into subvalues for
+        # the subwidgets
```
koopmant: This comment can be removed.

app/grandchallenge/cases/widgets.py
```diff
@@ -200,3 +205,105 @@ def compress(self, values):
...
+    template_name = "cases/file_name.html"
+
+
+class DICOMUploadWidget(MultiWidget):
+    template_name = "cases/dicom_upload_widget.html"
+
+    def __init__(self, attrs=None):
+        # pass in the widgets as dictionary with the proper suffixes as keys,
+        # so we can reliably tell these fields apart from regular socket fields
+        # in forms that include this field
```
koopmant: This comment can be removed.

app/grandchallenge/cases/templates/cases/dicom_upload_widget.html
```diff
@@ -0,0 +1,16 @@
...
+            name="{{ widget.name }}"
+            value="{{ widget.attrs.current_value.name }}"
+            disabled
+    >
+    <small id="{{ widget.name }}_helptext" class="form-text text-muted">To upload a different DICOM image for this socket, please delete the entire case/item and start from scratch.</small>
+{% else %}
+    <div class="dicom-upload-widget">
+        {% for subwidget in widget.subwidgets %}
+            {% include subwidget.template_name with widget=subwidget %}
+        {% endfor %}
```
koopmant: I think I would do `{% include 'django/forms/widgets/multiwidget.html' %}` here. 

Alternatively, we could define the contents of `file_name.html` here directly and avoid needing to create a separate widget type `DICOMFileNameInput`.

app/grandchallenge/cases/templates/cases/dicom_image_set_name_input.html
koopmant, line 1: If we keep this, this file needs a better name.

app/grandchallenge/cases/templates/cases/file_name.html
```diff
@@ -0,0 +1,6 @@
+<div class="input-group mb-3">
+    <div class="input-group-prepend">
+        <span class="input-group-text" id="label">Name for resulting DICOM file</span>
```
koopmant: Maybe "resulting DICOM image instance" or "resulting DICOM image set"?

koopmant (commented):
I think using the MultiWidget and MultiValueField like this works really well for what we want. I looked into keeping it as separate fields, by returning multiple fields in the `InterfaceFormFieldFactory` for example, but that just makes looping through the interfaces on the form more complex. 
That is now resolved with the `parse_slug` method. I think we might not need that one either if we fix the logic of retrieving the form data for the fields. But I think it's best to fix that all at once in https://github.com/DIAGNijmegen/rse-roadmap/issues/435 and keep everything uniform for now. 

So all in all: looks good. I just have some cleanup suggestions.

app/grandchallenge/cases/widgets.py
amickan: Good one!

app/grandchallenge/cases/templates/cases/dicom_upload_widget.html
amickan: I think I prefer to change it to `{% include 'django/forms/widgets/multiwidget.html' %}` and keep the seperate widget type. Otherwise, this template here becomes too custom again.

app/grandchallenge/cases/templates/cases/dicom_image_set_name_input.html
amickan: Good point, I agree ;-P How about `file-name-input.html`?

app/grandchallenge/cases/templates/cases/file_name.html
amickan: "resulting DICOM image set" nails it, I think.

app/grandchallenge/cases/templates/cases/dicom_image_set_name_input.html
koopmant: I'd go with `dicom_image_set_name_input.html` and rename the `DICOMFileNameInput` to the same.

amickan: Much better, I'll use that.

app/grandchallenge/components/forms.py
amickan: Yeah, we need to look into how to best do that. With what I proposed in https://github.com/DIAGNijmegen/rse-roadmap/issues/435, we could simply do something like `source_choice = self.data.get(f"{prefixed_interface_slug}_choice", None)`and then pass that to the `InterfaceFormFieldFactory`, but maybe there are better ways.

koopmant (approved):
LGTM!

PR #4317, opened 2025-09-30
Add token generation for fetching data from Health Imaging

jmsmkn:
See https://github.com/DIAGNijmegen/rse-roadmap/issues/432

### DIFF
--- a/app/grandchallenge/cases/views.py
+++ b/app/grandchallenge/cases/views.py
@@ -43,6 +43,7 @@ from grandchallenge.core.guardian import (
 )
 from grandchallenge.core.renderers import PaginatedCSVRenderer
 from grandchallenge.datatables.views import Column, PaginatedTableListView
+from grandchallenge.serving.models import Download
 from grandchallenge.subdomains.utils import reverse_lazy
 from grandchallenge.uploads.models import UserUpload
 from grandchallenge.uploads.widgets import UserUploadMultipleWidget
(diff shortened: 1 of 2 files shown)


app/grandchallenge/cases/views.py
```diff
@@ -94,6 +100,63 @@ class ImageViewSet(ReadOnlyModelViewSet):
...
+            image_frame_requests[frame_id] = self.serialize_aws_request(
+                frame_request
+            )
+
+        metadata_request = AWSRequest(
+            method="POST",
+            url=f"{image_set_url}/getImageSetMetadata",
+            data=json.dumps({"versionId": "1"}),
+        )
+        medical_imaging_auth.add_auth(metadata_request)
```
pkcakeout, line 152: Ha cool! Was that documented somewhere or was this your reverse engineering effort? Looks clean! 

Also: the authentication details get tagged on as query arguments here through the serializer, right?

jmsmkn: It's a mix of docs and reverse engineering.

There is a new field in the serialiser:

```python3
image = get_image(pk=...)

if image["dicom_image_set"]:
  # This can be downloaded from the DICOM store, so fetch the download urls from
  # /api/v1/cases/images/{pk}/dicom/
  # Then use those urls to get the metadata and image frames
elif image["files"]:
  # panimg generated images as before
else:
  # Shouldn't happen!
```

app/grandchallenge/cases/serializers.py
```diff
@@ -32,16 +33,24 @@ class Meta:
         fields = ("pk", "image", "file", "image_type")
 
 
+class DICOMImageSetSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = DICOMImageSet
+        fields = ("image_set_id",)
```
amickan, line 39: The frame ids don't need to be extracted?

amickan: Ah I see that the API endpoint takes care of that and that the client doesn't need to know the single frame ids. Never mind.

koopmant (approved):
LGTM

PR #4319, opened 2025-09-30
Rename image kind

koopmant:
Rename InterfaceKindChoices.IMAGE to InterfaceKindChoices.MHA_OR_TIFF_IMAGE

To avoid confusion in dealing with mha or tiff images and dicom images. 

### DIFF
--- a/app/grandchallenge/components/models.py
+++ b/app/grandchallenge/components/models.py
@@ -131,7 +131,7 @@ class InterfaceKindChoices(models.TextChoices):
     MULTIPLE_CHOICE = "MCHO", _("Multiple choice")
 
     # Image types
-    MHA_OR_TIFF_IMAGE = "IMG", _("Image")
+    MHA_OR_TIFF_IMAGE = "IMG", _("MHA or TIFF Image")
     SEGMENTATION = "SEG", _("Segmentation")
     HEAT_MAP = "HMAP", _("Heat Map")
     DISPLACEMENT_FIELD = "DSPF", _("Displacement field")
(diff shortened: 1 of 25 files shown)


app/grandchallenge/components/models.py
```diff
@@ -131,7 +131,7 @@ class InterfaceKindChoices(models.TextChoices):
     MULTIPLE_CHOICE = "MCHO", _("Multiple choice")
 
     # Image types
-    IMAGE = "IMG", _("Image")
+    MHA_OR_TIFF_IMAGE = "IMG", _("MHA or TIFF Image")
```
koopmant: Is changing the label here safe? How do I figure out if I can do that?

jmsmkn: The value is serialised in the "Kind" field, so the change would likely break users code.

```
    SEGMENTATION = "SEG", _("Segmentation")
    HEAT_MAP = "HMAP", _("Heat Map")
    DISPLACEMENT_FIELD = "DSPF", _("Displacement field")
```

These are also currently MHA/TIFF right so the keys could be updated too.

koopmant: > The value is serialised in the "Kind" field, so the change would likely break users code.

All right, thanks. 

> These are also currently MHA/TIFF right so the keys could be updated too.

Thanks. 

I just realized we also have DZI image type, should the name reflect that?

jmsmkn: The DZI is created from the TIFF, so should never exist without it. Adding `MHA_TIFF_DZI_` is a bit awkward, `MHA_TIFF_` should be sufficient.

For historical reasons we also have MHD, RAW and ZRAW, those can similarly be ignored.

I don't know if there is a good other prefix. `PANIMG_` is a bit too cryptic. Or is that clear enough to distinguish it from `DICOM_`?

koopmant: I like`PANIMG_` actually.

jmsmkn: Okay!

PR #4320, opened 2025-09-30
Clean up components module

koopmant:
- Remove attribute InterfaceKindChoices from InterfaceKind class
- Rename type to kind for references to interface kinds
- Remove `saved_in_object_store` property (it is the inverse of `store_in_database`)
- Clean up code

### DIFF
--- a/app/grandchallenge/components/models.py
+++ b/app/grandchallenge/components/models.py
@@ -1391,7 +1391,9 @@ class ComponentInterfaceValue(models.Model):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._value_orig = self.value
-        self._image_orig = self.image.pk if self.image else None
+        self._image_orig = self._meta.get_field("image").value_from_object(
+            self
+        )
         self._file_orig = self.file
 
     def save(self, *args, **kwargs):


app/grandchallenge/components/models.py
```diff
@@ -713,7 +700,7 @@ def value_required(self):
         value_required = True
         if self.kind == InterfaceKindChoices.BOOL:
             value_required = False
-        elif not self.is_image_kind and not self.requires_file:
+        elif self.requires_value:
```
koopmant, line 703: These two properties are confusing. The difference between the two is
- `value_required` tells you if any value is required, i.e. `self.value`, `self.image` or `self.file`;
  it always returns true unless the type is boolean or the json schema allows the value to be null. (Would such a schema make sense?) If the schema allows null, we set required to False, meaning we will later generate a new schema that also allows null by adding null to the options of allowed values; but it was already allowed. As I said, confusing. @amickan, any insights? 
- `requires_value` means "json type stored in database", in which case we return `SuperKind.VALUE` as the `super_kind`. 
  I think I can replace usages of this with a comparison against the super_kind.

app/grandchallenge/components/models.py
```diff
@@ -573,7 +559,7 @@ def create_instance(self, *, image=None, value=None, fileobj=None):
         elif fileobj:
             container = File(fileobj)
             civ.file.save(Path(self.relative_path).name, container)
-        elif self.saved_in_object_store:
+        elif value and not self.store_in_database:
```
jmsmkn, line 562: I don't think that you can check the value here as it might not be "truthy".

app/grandchallenge/components/models.py
jmsmkn: There are several interfaces where null is allowed - e.g. spacing, a number or null, indicates whether the spacing is nothing, or unknown.

app/grandchallenge/components/models.py
koopmant: good catch, thanks.

app/grandchallenge/components/models.py
amickan: I think null is only ever allowed for sockets of type ANY. Whether that makes sense is debatable now that we have optional inputs, but some algorithms rely on the value being nullable (spacing is a good example), so changing this would break those algorithms. 

As for the schema generation, that was added by Chris [here](https://github.com/comic/grand-challenge.org/pull/3813), he would probably know more about this.

PR #4321, opened 2025-09-30
Refactor InterfaceKind to InterfaceKinds

koopmant:
Refactor `InterfaceKind` to `InterfaceKindSets`, a (set, Enum) class

The sets that were defined as static methods are now defined as immutable attributes.


koopmant:
1. `InterfaceKinds` works, nice and short. 
2. It's already lost? Does that mean I somehow broke the docstrings in a previous commit?
3. I'll put those back as comments.

PR #4318, opened 2025-09-30
New dicom image kind

koopmant:
Add `InterfaceKindChoices.DICOM_IMAGE_SET` kind

Related to https://github.com/DIAGNijmegen/rse-roadmap/issues/423

PR #4322, opened 2025-10-01
Enable DICOM upload and import into HealthImaging

amickan:
This adds:
- a `DICOMUploadWidget` that deidentifies DICOM data client side and allows up to 2000 files to be uploaded
- a `DICOMImageSetUpload` model that handles server-side deidentification of uploaded DICOMs and that starts and tracks imports of that data into AWS HealthImaging
- a `DICOMImageSet` model to reference the resulting image sets in HealthImaging

Closes https://github.com/DIAGNijmegen/rse-roadmap/issues/431


app/grandchallenge/cases/models.py
```diff
@@ -837,3 +913,380 @@ def save(self, *args, **kwargs):
...
+        dicom_image_set = DICOMImageSet(
+            image_set_id=image_set_id,
+            dicom_image_set_upload=self,
+        )
+        dicom_image_set.full_clean()
+        dicom_image_set.save()
+
+        image = Image(
+            dicom_image_set=dicom_image_set,
+            name="placeholder",  # todo: get name for set during upload
```
jmsmkn: That can be done now, right?

app/tests/cases_tests/test_widget.py
```diff
@@ -161,3 +173,102 @@ def test_flexible_image_widget_prepopulated_value():
...
+    im = ImageFactory(
+        name="test_image",
+        dicom_image_set=DICOMImageSetFactory(dicom_image_set_upload=upload),
+    )
+    assign_perm("cases.view_image", user_with_perm, im)
+    ci = ComponentInterfaceFactory(kind=ComponentInterface.Kind.PANIMG_IMAGE)
+    civ = ComponentInterfaceValueFactory(interface=ci, image=im)
+
+    # fake DICOM ci for now
+    # TODO: create a proper DICOM super kind ci above
```
jmsmkn: This needs to be resolved.

koopmant:
> There are still a couple of TODOs here along with open PRs based on this branch.

Sorry, I'm a bit confused. This feature branch was supposed to be merged at the end of last cycle, some of the todo's left open are related to the new cycle. Shouldn't we merge this as is so we can continue in the current cycle without this feature branch still being open? 
Or should we wrap up this feature branch with the new dicom kind and then merge afterwards?

app/grandchallenge/cases/models.py
amickan: True, this can be `self.name` now.

app/tests/cases_tests/test_widget.py
amickan: This can only be resolved after this cycle since it depends on the new kind and all surrounding logic to be in place.

jmsmkn: As far as I can tell it only depends on the key being present and set. Why would it depend on the surrounding logic?

amickan: It depends on `is_dicom_image_kind` returning True in the test - that is all. But for that method to return True the logic around the new kind needs to be in place and Thomas is still working on that.

jmsmkn: Isn't the only thing missing for that in #4318:

```python3
    @property
    def is_dicom_image_kind(self):
        return self.kind == InterfaceKindChoices.DICOM_IMAGE_SET
```

amickan: For this test, yes, that would be enough.

PR #4324, opened 2025-10-01
Remove InterfaceKindChoices from InterfaceKind

koopmant:


app/grandchallenge/components/models.py
```diff
@@ -158,9 +158,6 @@ class InterfaceSuperKindChoices(models.TextChoices):
 
 
 class InterfaceKind:
-    """Interface kind choices."""
-
-    InterfaceKindChoices = InterfaceKindChoices
```
jmsmkn, line 163: Are you sure that this is safe to remove? These things are often transported through for use in templates.

koopmant: Yes, but in this case InterfaceKind is not a model. InterfaceKindChoices are still transported through on `ComponentInterface.Kind`. I've searched for `.InterfaceKindChoices` and it is used nowhere anymore.

jmsmkn (approved):
LGTM!

PR #4325, opened 2025-10-01
Clean up components module

koopmant:
Clean up code and logic for usages of interface kinds and super kinds.


app/tests/components_tests/test_models.py
```diff
@@ -56,129 +55,25 @@
...
+        InterfaceKindChoices.MULTIPLE_TWO_D_BOUNDING_BOXES,
+        InterfaceKindChoices.MULTIPLE_DISTANCE_MEASUREMENTS,
+        InterfaceKindChoices.MULTIPLE_POINTS,
+        InterfaceKindChoices.MULTIPLE_POLYGONS,
+        InterfaceKindChoices.MULTIPLE_LINES,
+        InterfaceKindChoices.MULTIPLE_ANGLES,
+        InterfaceKindChoices.MULTIPLE_ELLIPSES,
+        InterfaceKindChoices.MULTIPLE_THREE_POINT_ANGLES,
+        *InterfaceKind.interface_type_image(),
+        *InterfaceKind.interface_type_file(),
```
jmsmkn: This refactoring removes the point of the test - that all choices are covered by a test that is independent from the implementation. It also removes testing the choices where storing the data in the database is optional, that should still be tested.

app/grandchallenge/components/models.py
```diff
@@ -1404,14 +1386,10 @@ def __str__(self):
...
         super().__init__(*args, **kwargs)
         self._value_orig = self.value
-        self._image_orig = self._dict["image"]
-        self._file_orig = self.file
-
-    @property
-    def _dict(self):
-        return model_to_dict(
-            self, fields=[field.name for field in self._meta.fields]
+        self._image_orig = self._meta.get_field("image").value_from_object(
```
jmsmkn: Would `FieldChangeMixin` work here to replace all of the `_<field>_orig` work?

app/grandchallenge/components/models.py
```diff
@@ -2308,11 +2286,11 @@ def __init__(self, *, interface_slug, value):
...
         ci = ComponentInterface.objects.get(slug=interface_slug)
 
-        if ci.requires_value:
+        if ci.super_kind == ci.SuperKind.VALUE:
             self._init_json_civ_data()
-        elif ci.is_image_kind:
+        elif ci.super_kind == ci.SuperKind.IMAGE:
             self._init_image_civ_data()
-        elif ci.requires_file:
+        elif ci.super_kind == ci.SuperKind.FILE:
```
jmsmkn, line 2295: Shouldn't the fall-through be caught?

app/grandchallenge/components/models.py
koopmant: That works! I wonder why we don't use `model_to_dict` or `field.value_from_object()` in that mixin though?

app/tests/components_tests/test_models.py
koopmant: The test had become useless indeed. I've added a test for `store_in_database=False` with all kind choices and changed the other test with `store_in_database=True` to determine if the validation error should be raised or not. 
I've added another test to check if we are covering all the choices. 

These choices are defined in a variable so I can use them to parametrize the test and test if all choices are covered in the variable. Not sure if there is a better way to do that. I read about pytest-lazy-fixture, but it seemed overkill for what I needed.

app/grandchallenge/components/models.py
koopmant: raising `NotImplementedError` now.

app/grandchallenge/components/models.py
jmsmkn: Likely didn't know about `model_to_dict`.

app/grandchallenge/components/form_fields.py
```diff
@@ -75,33 +75,33 @@ def __new__(
...
-            )
-        elif interface.is_dicom_image_kind:
+        if interface.is_dicom_image_kind:
             return DICOMUploadField(
                 user=user,
                 initial=initial,
                 **kwargs,
             )
-        elif interface.is_image_kind:
+        elif interface.super_kind == interface.SuperKind.IMAGE:
```
jmsmkn: Style wise I found the three if statements on the super kind easier to follow. It may be easier for future developers to follow if the switch on the DICOM image kind was moved in. That way there is no mixing types on the top level check. For example

```python3
if interface.super_kind == interface.SuperKind.IMAGE:
    if interface.is_dicom_image_kind:
         return DICOMUploadField(...)
    else:
         return FlexibleImageField(...)
elif interface.super_kind == interface.SuperKind.FILE:
    ...
elif interface.super_kind == interface.SuperKind.VALUE:
    ...
else:
    raise NotImplementedError(f"Unknown super kind: {interface.super_kind}")
```

app/grandchallenge/components/models.py
```diff
@@ -2317,14 +2285,18 @@ def __init__(self, *, interface_slug, value):
 
         ci = ComponentInterface.objects.get(slug=interface_slug)
 
-        if ci.requires_value:
-            self._init_json_civ_data()
-        elif ci.is_dicom_image_kind:
+        if ci.is_dicom_image_kind:
             self._init_dicom_civ_data()
-        elif ci.is_image_kind:
+        elif ci.super_kind == ci.SuperKind.VALUE:
```
jmsmkn: Same here.

app/grandchallenge/components/form_fields.py
koopmant: Yes, I thought about doing it like that as well. It looks good, I'll change it.

PR #4326, opened 2025-10-01
Improve AdditionalInputsMixin

### LINKED ISSUE #4257: `AdditionalInputsMixin` doesn't play nicely with other form mixins
This mixin:

- Requires explicit instantiation with `init_additional_inputs` rather than using `__init__`
- Requires explicit cleaning calls to `clean_additional_inputs` rather than using `clean`

This causes issues with the use of other mixins such as `SaveFormInitMixin`, see #4256

amickan:
Closes #4257

### DIFF
--- a/app/grandchallenge/algorithms/forms.py
+++ b/app/grandchallenge/algorithms/forms.py
@@ -555,7 +555,11 @@ class AlgorithmForPhaseForm(
         phase,
         **kwargs,
     ):
-        super().__init__(*args, user=user, phase=phase, **kwargs)
+        self._user = user
+        self._phase = phase
+
+        super().__init__(*args, **kwargs)
+
         self.fields["workstation_config"].initial = workstation_config
         self.fields["workstation_config"].disabled = True
         self.fields["hanging_protocol"].initial = hanging_protocol
--- a/app/grandchallenge/algorithms/forms.py
+++ b/app/grandchallenge/algorithms/forms.py
@@ -442,9 +442,9 @@ class AlgorithmForm(
 class UserAlgorithmsForPhaseMixin:
 
     def __init__(self, *args, user, phase, **kwargs):
-        super().__init__(*args, **kwargs)
         self._user = user
         self._phase = phase
+        super().__init__(*args, **kwargs)
 
     @cached_property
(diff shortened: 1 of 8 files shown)


app/grandchallenge/algorithms/forms.py
```diff
@@ -566,7 +555,11 @@ def __init__(
         phase,
         **kwargs,
     ):
-        super().__init__(*args, user=user, phase=phase, **kwargs)
+        self._user = user
+        self._phase = phase
+
+        super().__init__(*args, **kwargs)
```
jmsmkn: This construction is a bit strange now. The developer needs to know to set these attributes in order to use `UserAlgorithmsForPhaseMixin`. Isn't it better to keep that as a responsibility of the mixin?

I think this pattern is repeated with having to set `self.additional_inputs`, so I think there is still a design problem.

amickan: True. I see your point. I agree that it makes more sense to leave this to the mixins.

amickan: Hm, I'm unsure how to properly pass along the user argument to the mixins. Both `UserAlgorithmsForPhaseMixin` and `AdditionalInputsMixin` need it, but they are not always used together, so if I modified `UserAlgorithmsForPhaseMixin`  to pass on the user argument in the call to `super()`, I run into issues in places where that mixin is used without `AdditionalInputsMixin`.

amickan: The only way I found to make this work is by passing the extra kwargs via `kwargs` and not declare them explicitly in the mixin's init method. I haven't seen that pattern elsewhere in the codebase though, and I'm wondering if I should then raise an error if a required argument is not present in kwargs. I also only did this for `AdditionalInputsMixin`, but should probably do the same for `UserAlgorithmsForPhaseMixin` for consistency.

jmsmkn: I have made a PR with a suggestion. In short: we needed to decide on who owned each kwarg, and only consume it in one place.

amickan: Thanks. That is an elegant solution.

PR #4327, opened 2025-10-01
DICOM Storage Fixup

jmsmkn:
From reviewing #4322:

- Tidy up of the settings to separate Django Stroages settings from our internal AWS settings
  - For the future: perhaps we should use another namespace as `AWS_` is overloaded
- Ensures that the upload object exists, we may need to get the series and study ids later
  - For the future: we do not validate the stored metadata to ensure that the series and study ids are as expected, that there really only is one series in the study
- Tidy up of error handling, the specific exceptions were uncaught so generic ones can be used, and the exceptions should not be suppressed if unhandled and there is nothing to commit

### DIFF
--- a/app/grandchallenge/cases/tasks.py
+++ b/app/grandchallenge/cases/tasks.py
@@ -628,10 +628,12 @@ def revert_image_set_to_initial_version(*, image_set_id, version_id):
             updateImageSetMetadataUpdates={"revertToVersionId": "1"},
             force=False,
         )
-    except ClientError as e:
+    except ClientError as error:
         if (
-            e.response["Error"]["Code"] == "ResourceNotFoundException"
-            and e.response["Error"]["Message"]
+            error.response["Error"]["Code"] == "ResourceNotFoundException"
+            and error.response["Error"]["Message"]
             == "Requested version(s) of ImageSet(s) is not the latest."
         ):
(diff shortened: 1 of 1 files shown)


app/grandchallenge/cases/tasks.py
```diff
@@ -609,8 +610,6 @@ def delete_healthimaging_image_set(*, image_set_id):
         pass  # image set already deleted
     except health_imaging_client.exceptions.ThrottlingException as e:
         raise RetryStep("Request throttled") from e
-    except ClientError:
-        logger.error("Couldn't delete image set", exc_info=True)
```
koopmant, line 613: I don't know when to use the logger and when to just raise the exception. What is the difference here?

jmsmkn: In both cases these are for errors that need fixing through future code changes.

The logger should be used if you want to suppress the error from being raised. Suppressing the error means that the transaction will be committed and the celery task will be marked as successful. You want to do that in situations where there is something to commit - e.g. you are saving the state on a Django object for reporting to the user.

In this case, these celery tasks are only making AWS API calls, so there is nothing to commit. As the error is anyway unhandled there is no need to treat `ClientErrors` any differently from any other error that might be raised.

PR #4328, opened 2025-10-02
Refactor Mixins

jmsmkn:
Creates User and Phase mixins to separate the responsibility of what objects own which kward during initialisation.


amickan (approved):
Thanks, James! LGTM!

PR #4329, opened 2025-10-02
Order algorithm images and tarballs by creation date

### LINKED ISSUE #4260: Indicate when algorithm model is missing
We want to encourage algorithms editors to use algorithm models, so indicate when one is not active in the algorithm sidebar.

### LINKED ISSUE #4279: Algorithm models are not ordered by date uploaded
This is different to the algorithm containers and causes confusion. Also check the ground truth ordering.

Image

amickan:
Closes #4279 
Closes #4260


amickan:
I did that to keep it in line with how we order AlgorithmImages. I agree that it's not necessary, but figured that consistency would be good.

PR #4330, opened 2025-10-02
Use `model_to_dict` in `FieldChangeMixin`

jmsmkn:


jmsmkn:
Won't do this - `model_to_dict` excludes non-editable fields, and `value_from_object` will list all objects in many to many relationships.

app/grandchallenge/core/models.py
```diff
@@ -105,34 +106,42 @@ class Meta:
 
 
 class FieldChangeMixin:
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, tracked_properties=(), **kwargs):
```
koopmant, line 109: By also tracking properties, the name of the class is no longer accurate. Perhaps we should change it to `StateChangeMixin` or something.

app/grandchallenge/core/models.py
```diff
@@ -105,34 +106,42 @@ class Meta:
...
+        }
+
+        many_to_many_field_names = {f.name for f in self._meta.many_to_many}
 
-        if "_state" in state:
-            del state["_state"]
+        for attname in tracked_properties:
+            is_already_tracked = attname in self._tracked_attrs
+            is_many_to_many_field = attname in many_to_many_field_names
+            is_property = isinstance(getattr(type(self), attname), property)
```
koopmant, line 124: Are these two not mutually exclusive?

app/grandchallenge/core/models.py
```diff
@@ -105,34 +106,42 @@ class Meta:
...
+                self._meta.concrete_fields, self._meta.private_fields
+            )
+        }
+
+        many_to_many_field_names = {f.name for f in self._meta.many_to_many}
 
-        if "_state" in state:
-            del state["_state"]
+        for attname in tracked_properties:
+            is_already_tracked = attname in self._tracked_attrs
```
koopmant, line 122: Is this one really necessary? When would this happen? And if it does happen, does that really change anything?

koopmant (commented):
Still had these pending, so I'll just leave it here for documentation. 

I don't understand why it doesn't work, but let's not get into that.

app/grandchallenge/core/models.py
jmsmkn: Indeed, the state dict is currently being misused in the codebase to track this by overriding `_current_state` without any of these validation checks, so made it a first class feature.

app/grandchallenge/core/models.py
jmsmkn: I don't think so - we want to allow properties but deny many to many fields.

app/grandchallenge/core/models.py
jmsmkn: If you somehow have a clash between a property and field name, for fields the attr can be different from the name (in one to one fields, foreign keys and many to many fields it is by default, but you can also set this as an option on the Django field), so you need to check that they're independent.

Issue #3211, opened 2024-01-26
FlexibleImageWidget should have a null option 


amickan:
**Is your feature request related to a problem? Please describe.**
For archive items and display sets, users need to be able to remove existing images from archive items and display sets. 

**Describe the solution you'd like**
Similar to the `SelectUploadWidget`, the `FlexibleImageWidget` should have a `----` option.
When a user selects that option, the image CIV should be removed from the archive item / display set. 

As discussed [here](https://github.com/comic/grand-challenge.org/pull/3156#issuecomment-1884507499), before removing the CIV, however, we might need to attach the CIV to a new archive item / display set so that the user keeps permissions to the image. 

I wonder if the latter is really necessary. If a user deletes a case, they might also lose permission to the image, so why not accept that for removing an image from an item as well? Maybe I'm overseeing some of the rammifications here. If we do need to add a new case, that needs to be made very clear so users don't accidentally end up with additional cases without knowing.

PR #4333, opened 2025-10-03
Handle new dicom kind on html forms

koopmant:
Handle the new dicom kind on html forms to attach dicom to objects. 

### DIFF
--- a/app/tests/components_tests/test_tasks.py
+++ b/app/tests/components_tests/test_tasks.py
@@ -597,7 +597,7 @@ def test_add_dicom_image_set_to_object_updates_upload_on_validation_fail(
     assert upload.status == DICOMImageSetUploadStatusChoices.FAILED
     assert (
         upload.error_message
-        == f"Image validation for socket {ci.title} failed with error: Image does not exist"
+        == f"Image validation for socket {ci.title} failed with error: Image imports should result in a single image"
     )
 
 
@@ -634,7 +634,9 @@ def test_add_dicom_image_set_to_object_marks_job_as_failed_on_validation_fail(
     obj.refresh_from_db()
     assert obj.status == obj.CANCELLED
     assert obj.error_message == "One or more of the inputs failed validation."
-    assert "Image does not exist" in str(obj.detailed_error_message)
+    assert "Image imports should result in a single image" in str(
+        obj.detailed_error_message
+    )
     assert "some_async_task" not in str(callbacks)
 
 
@@ -682,7 +684,7 @@ def test_add_dicom_image_set_to_object_sends_notification_on_validation_fail(
     assert ComponentInterfaceValue.objects.filter(interface=ci).count() == 0
     assert Notification.objects.count() == 1
     assert (
-        f"Image validation for socket {ci.title} failed with error: Image does not exist"
+        f"Image validation for socket {ci.title} failed with error: Image imports should result in a single image"
         in Notification.objects.first().description
     )
     assert "some_async_task" not in str(callbacks)


app/grandchallenge/components/templates/components/partials/socket_read_write_paths.html
```diff
@@ -1,6 +1,8 @@
-{% if socket.is_image_kind %}
+{% if socket.is_panimg_kind %}
     <code>/{{ input|yesno:"input,output" }}{% if socket.relative_path %}/{{ socket.relative_path }}{% endif %}/&lt;uuid&gt;.mha</code> or
     <code>/{{ input|yesno:"input,output" }}{% if socket.relative_path %}/{{ socket.relative_path }}{% endif %}/&lt;uuid&gt;.tif</code>
+{% elif socket.is_dicom_image_kind %}
+    <code>/{{ input|yesno:"input,output" }}{% if socket.relative_path %}/{{ socket.relative_path }}{% endif %}/&lt;uuid&gt;/</code>
```
koopmant: I think this is right, right? The name should not be used here and is only so the owners can identify their own image sets?

jmsmkn: The name is only for identification, so used in the templates and API.

The algorithms will receive a folder full of `.dcm` files, the file names will not be the uuid of the socket value.

koopmant: oh right, each set is already under a socket, no need for a `uuid` directory.

app/grandchallenge/cases/models.py
```diff
@@ -1374,3 +1385,12 @@ def convert_image_set_to_internal(self, *, image_set_id):
         image = Image(dicom_image_set=dicom_image_set, name=self.name)
         image.full_clean()
         image.save()
+
+        if self.task_on_success:
+            self.task_on_success.kwargs.update(
+                {"dicom_image_set_pk": dicom_image_set.pk}
+            )
```
koopmant: We do something similar with the linked task in `RawImageUploadSession.process_images`, where we update the task kwargs with the `upload_session_pk`. I was hoping I could do it differently, but I don't know how. The dicom image set does not yet exist when I first create the task signature, so I have to add it later. 
I suppose I could rename the field to show that it is only meant for that specific task?

app/grandchallenge/cases/models.py
```diff
@@ -190,20 +191,17 @@ def update_status(
         self.status = status
 
         if detailed_error_message:
-            notification_description = oxford_comma(
+            self.error_message = oxford_comma(
```
jmsmkn: Rather than setting a property that needs to be saved in a switch statement it is usually more future proof to assign it to a variable, then set it below. That way you ensure that the value is set if future developers edit the switch. It also makes it easier to refactor into a method without side-effects if you'd prefer that.

app/grandchallenge/cases/models.py
```diff
@@ -1374,3 +1397,12 @@ def convert_image_set_to_internal(self, *, image_set_id):
...
         image.save()
+
+        if self.task_on_success:
+            self.task_on_success.kwargs.update(
+                {"dicom_image_set_pk": dicom_image_set.pk}
+            )
+
+    def execute_task_on_success(self):
+        if self.task_on_success:
+            on_commit(signature(self.task_on_success).apply_async())
```
jmsmkn: Be super careful here - `apply_async` should never actually be called otherwise it will be applied at this point, and not after the transaction has been committed.

```suggestion
            on_commit(signature(self.task_on_success).apply_async)
```

app/grandchallenge/components/tasks.py
```diff
@@ -1224,25 +1229,8 @@ def validate_voxel_values(*, civ_pk):
     civ.interface._validate_voxel_values(civ.image)
 
 
-@acks_late_micro_short_task(
-    retry_on=(LockNotAcquiredException,), delayed_retry=False
-)
-@transaction.atomic
-def add_image_to_object(  # noqa: C901
```
jmsmkn, line 1231: Unfortunately celery tasks do not have built in migrations. You cannot rename them or change the arguments without leaving the old task in place as a translation. You then need to deploy and wait for all the existing tasks on the queue that use the old task definition to run, then you can remove the old task definition.

app/grandchallenge/cases/models.py
jmsmkn: I think if you add a note saying that this kwarg will be added then that is okay.

However - you are modifying the task on the object but are not saving it here? Maybe this should be moved to where it is executed if it does not need to be saved.

app/grandchallenge/components/models.py
```diff
@@ -2441,15 +2446,23 @@ def create_civ(self, *, civ_data, user=None, linked_task=None):
...
-                upload_session=civ_data.upload_session,
-                user_upload_queryset=civ_data.user_upload_queryset,
-                linked_task=linked_task,
-            )
+            if ci.is_dicom_image_kind:
+                return self.create_civ_for_dicom_image(
+                    ci=ci,
+                    civ_data=civ_data,
+                    user=user,
+                    linked_task=linked_task,
```
jmsmkn: Don't you need to pass along the `image` and/or `dicom_image_set_upload_pk` here?

app/grandchallenge/components/models.py
```diff
@@ -2567,6 +2582,38 @@ def create_civ_for_image(  # noqa: C901
...
+        from grandchallenge.components.tasks import (
+            add_dicom_image_set_to_object,
+        )
+
+        if not user:
+            raise RuntimeError(
+                f"You need to provide a user along with the user upload "
+                f"queryset for interface {ci}"
+            )
+        upload = DICOMImageSetUpload.objects.create(
```
jmsmkn: The upload isn't cleaned?

app/grandchallenge/components/models.py
```diff
@@ -2567,6 +2582,38 @@ def create_civ_for_image(  # noqa: C901
...
+            creator=user,
+            name=civ_data.image_name,
+            linked_task=add_dicom_image_set_to_object.signature(
```
jmsmkn: If you create the object:

`upload = DICOMImageSetUpload(creator=user, ...)`

It will then have a pk that you can use when creating the task.

```
upload.linked_task = add_dicom_image_set_to_object.signature(kwargs={"dicom_image_set_upload_pk": upload.pk, ...})
upload.full_clean()
upload.save()
```

app/grandchallenge/components/tasks.py
jmsmkn: This is also really hard to review maybe due to code moves? Or is it GitHub just presenting this weirdly.

app/grandchallenge/components/tasks.py
```diff
@@ -1224,25 +1229,8 @@ def validate_voxel_values(*, civ_pk):
...
-    *,
-    app_label,
-    model_name,
-    object_pk,
-    interface_pk,
-    upload_session_pk,
-    linked_task=None,
-):
-    from grandchallenge.algorithms.models import Job
+def get_object(*, app_label, model_name, object_pk):
```
jmsmkn: This is named far too generically, and I think the refactoring has messed up the diff. For now, please go back to the code duplication to see if it gets the PR back to something that is easier to follow. The refactoring can be moved to a separate PR if needs be.

To avoid separate PRs you could try using stacked commits, either way requires active clean up.

app/grandchallenge/components/models.py
jmsmkn: I think not as these are options from the form? Unsure.

app/grandchallenge/cases/models.py
koopmant: I was doing it here because this is where I had a reference to the `dicom_image_set`; but I just realized I can follow the relation to use it where it is executed. However, this is no longer relevant after your other comment about providing the pk of the `DICOMImageSetUpload` object.

app/grandchallenge/cases/models.py
koopmant: Thanks!

app/grandchallenge/cases/models.py
koopmant: All right, makes sense.

app/grandchallenge/components/tasks.py
koopmant: It's because of the code moves. I refactored it so I could make sense of the code. Sorry for the mess.

koopmant: > Unfortunately celery tasks do not have built in migrations. You cannot rename them or change the arguments without leaving the old task in place as a translation. You then need to deploy and wait for all the existing tasks on the queue that use the old task definition to run, then you can remove the old task definition.

Good point, I didn't realize! I'll keep it in mind for the optional refactoring PR.

app/grandchallenge/components/models.py
koopmant: I was passing on the `image_name` and `user_upload_queryset` (but I was doing it wrong by not unpacking civ_data here.) 
Eventually, an existing image will be passed along here as well, but for now the form only produces new uploads. 

I suppose I can already create the code for that, since the widget will be updated later anyway.

app/grandchallenge/components/models.py
koopmant: Thank you, this is exactly what I needed!

app/grandchallenge/components/tasks.py
koopmant: Cleaned up. Refactoring will follow later if there is time.

app/grandchallenge/components/models.py
koopmant: It is cleaned now.

app/grandchallenge/components/models.py
```diff
@@ -2567,6 +2581,77 @@ def create_civ_for_image(  # noqa: C901
...
+            )
+            upload.full_clean()
+            upload.save()
+            upload.user_uploads.set(dicom_upload_with_name.user_uploads)
+
+            on_commit(
+                import_dicom_to_health_imaging.signature(
+                    kwargs={"dicom_imageset_upload_pk": upload.pk}
+                ).apply_async
+            )
```
koopmant, line 2599: In `create_civ_for_image` we call `RawImageUploadSession.process_images` which checks the upload session has state `PENDING` and then updates it to `REQUEUED` to make it idempotent; but I'm not sure why, because the task it calls (`build_images`) is already idempotent, so it seems unneccessary. Am I missing something here?

In `create_civ_for_file` we don't do such an extra step.

jmsmkn: It might be due to the initial state of the Upload session. These are created by users and then files are added to them. Then isn't the state changed so that they cannot be further modified once the user tells us to start importing it? Would need to trace it more.

app/grandchallenge/components/models.py
```diff
@@ -2567,6 +2582,77 @@ def create_civ_for_image(  # noqa: C901
...
+
+            if not user:
+                raise RuntimeError(
+                    f"You need to provide a user along with the user upload "
+                    f"queryset for interface {ci}"
+                )
+            upload = DICOMImageSetUpload(
+                creator=user, name=dicom_upload_with_name.name
+            )
+            upload.task_on_success = add_dicom_image_set_to_object.signature(
```
koopmant: I sometimes see signatures with `immutable=True`. When do I need to use this?

jmsmkn: It is only necessary for things that get added to Celery canvasses, which we shouldn't use as they are too buggy, however, you can end up with them accidentally. We used to use them and it is probably a hangover. See https://docs.celeryq.dev/en/stable/userguide/canvas.html#immutability

jmsmkn: It's probably a good idea to add so that your kwargs do not change if you accidentally  end up in a canvas, but not strictly necessary.

app/grandchallenge/components/models.py
koopmant: Added `image` as argument here. Will not add `upload` as we'll try to avoid creating uploads before this point.

app/grandchallenge/components/models.py
koopmant: Will keep this in mind for when we end up with DICOM uploads that never get processed. For know, we'll try to avoid those.

app/grandchallenge/components/models.py
```diff
@@ -2567,6 +2582,78 @@ def create_civ_for_image(  # noqa: C901
...
+                        error_message=format_validation_error_message(error=e),
+                        user=user,
+                    )
+                    return
+
+            self.remove_civ(civ=current_civ)
+            self.add_civ(civ=civ)
+
+            if linked_task is not None:
+                on_commit(signature(linked_task).apply_async)
```
jmsmkn: I think this is duplicated, can they be unified?

app/grandchallenge/components/models.py
```diff
@@ -2567,6 +2582,78 @@ def create_civ_for_image(  # noqa: C901
...
+
+        elif dicom_upload_with_name:
```
jmsmkn, line 2566: Is the fall through missing for this switch statement?

app/grandchallenge/components/tasks.py
```diff
@@ -1330,6 +1336,114 @@ def add_image_to_object(  # noqa: C901
...
+        logger.info("Nothing to do: upload session was not successful.")
+        return
+
+    error_handler = obj.get_error_handler()
+
+    try:
+        image = Image.objects.get(
+            dicom_image_set__dicom_image_set_upload_id=dicom_image_set_upload_pk
+        )
+    except Image.DoesNotExist:
```
jmsmkn: Shouldn't MultipleObjectsReturned also be handled?

app/grandchallenge/components/tasks.py
```diff
@@ -1330,6 +1336,114 @@ def add_image_to_object(  # noqa: C901
         logger.info("No linked task, task complete")
 
 
+@acks_late_micro_short_task(
+    retry_on=(LockNotAcquiredException,), delayed_retry=False
+)
+@transaction.atomic
+def add_dicom_image_set_to_object(  # noqa: C901
```
jmsmkn: This seems like it is mostly duplicate of `add_image_to_object`. Could that task be updated to take `dicom_image_set_upload_pk`, then either `dicom_image_set_upload_pk` or `upload_session_pk` should be set.

app/grandchallenge/components/tasks.py
```diff
@@ -1330,6 +1336,114 @@ def add_image_to_object(  # noqa: C901
...
+    try:
+        with check_lock_acquired():
+            obj = model.objects.select_for_update(nowait=True).get(
+                pk=object_pk
+            )
+    except (ArchiveItem.DoesNotExist, DisplaySet.DoesNotExist):
+        logger.info(f"Nothing to do: {model_name} no longer exists.")
+        return
+
+    interface = ComponentInterface.objects.get(pk=interface_pk)
```
jmsmkn, line 1278: Do we check anywhere that the Image is of the correct type for the interface - e.g. that only DICOM images can be added for DICOM interfaces. Is that done in the civ's clean method?

app/grandchallenge/components/models.py
koopmant: There is validation in CIVData that ensures it exists. We can still add it of course.

app/grandchallenge/components/tasks.py
koopmant: dicom_image_set is a OneToOne field, so multiple objects should not be possible right?

app/grandchallenge/components/tasks.py
koopmant: Good point! We don't do that anywhere.

app/grandchallenge/components/models.py
koopmant: They can. Added `dicom_image_with_name` as input argument and removed the separate function.

app/grandchallenge/components/tasks.py
koopmant: Yes, we can do that. I think this is what you meant.

app/grandchallenge/components/tasks.py
koopmant: I added `ComponentInterfaceValue._validate_image_kind`

app/grandchallenge/components/models.py
koopmant: With the latest changes, this is no longer True. I see we can also have valid CIVdata with `image=None`. Not sure yet what to do in that case.

app/grandchallenge/components/models.py
```diff
@@ -1420,6 +1421,13 @@ def _validate_image_only(self):
                 f"File ({self.file}) or value should not be set for images"
             )
 
+    def _validate_image_kind(self):
```
jmsmkn, line 1411: Please add a test for this, this really always needs to work.

app/grandchallenge/components/models.py
jmsmkn: Unsure, what would it mean in that case?

app/grandchallenge/components/models.py
```diff
@@ -2508,9 +2513,11 @@ def create_civ_for_image(  # noqa: C901
         image=None,
         upload_session=None,
         user_upload_queryset=None,
+        dicom_upload_with_name=None,
```
jmsmkn, line 2503: Maybe add some validation that if `dicom_upload_with_name` is set, then `upload_session` and `user_upload_queryset` should not be set? Maybe there is more?

app/grandchallenge/components/tasks.py
```diff
@@ -1234,9 +1240,15 @@ def add_image_to_object(  # noqa: C901
...
     object_pk,
     interface_pk,
-    upload_session_pk,
+    upload_session_pk=None,
+    dicom_image_set_upload_pk=None,
     linked_task=None,
 ):
+    if upload_session_pk is None and dicom_image_set_upload_pk is None:
+        raise ValueError(
+            "Neither upload_session_pk nor dicom_image_set_upload_pk"
```
jmsmkn: The message is incomplete. Also shouldn't at least one of those be set?

app/grandchallenge/components/tasks.py
```diff
@@ -1260,27 +1272,50 @@ def add_image_to_object(  # noqa: C901
...
+            logger.info("Upload session should only have one image")
+            return
+    else:
+        upload = DICOMImageSetUpload.objects.get(pk=dicom_image_set_upload_pk)
+
+        if upload.status != DICOMImageSetUploadStatusChoices.COMPLETED:
+            logger.info("Nothing to do: DICOM upload was not successful.")
+            return
+
+        error_handler = obj.get_error_handler()
```
jmsmkn: Why isn't the linked object being passed here?

app/grandchallenge/components/tasks.py
```diff
@@ -1260,27 +1272,50 @@ def add_image_to_object(  # noqa: C901
...
+            image = Image.objects.get(origin_id=upload_session_pk)
+        except (Image.DoesNotExist, Image.MultipleObjectsReturned):
+            error_handler.handle_error(
+                interface=interface,
+                error_message="Image imports should result in a single image",
+                user=upload.creator,
+            )
+            logger.info("Upload session should only have one image")
+            return
+    else:
```
jmsmkn: `elif dicom_image_set_upload_pk is not None` and then raising the fall through later?

app/grandchallenge/components/tasks.py
koopmant: This has to do with the special error handling for `RawImageUploadSession`. The `RawImageUploadSessionErrorHandler` accepts a linked object which is checked in `handle_error` to change the error message and passed on to `RawImageUploadSession.update_status` to update the status of the `linked_object` (in the upload session the linked object is the other end of the link) as `CANCELLED` if the linked object is a `ComponentJob` instance. 

I added a test for this with DICOM, and the jobs get marked as cancelled with the normal error handlers, so maybe I'm missing something here or maybe this special error handler is no longer needed.

app/grandchallenge/components/tasks.py
koopmant: Yes, one of them should be set. Doesn't the message mean that? I'll change it to something more clear.

app/grandchallenge/components/models.py
koopmant: We have that in `CIVData.validate`. This function is only called within the class, so I can make it protected to clarify that? We can also pass civdata directly, to make it more clear. Either way I'd like to do that in a separate PR.

app/grandchallenge/components/tasks.py
jmsmkn: I'm afraid that I don't know enough about it to know for sure.

app/grandchallenge/components/tasks.py
koopmant: I'm already checking at the start of the task that either of them is set. Should I remove that then?

jmsmkn: It's probably the most future proof to have both - fail fast so the users get a nice message, and the second means that future developers do not need to rely on earlier behaviour when modifying the switch statement.

app/grandchallenge/components/tasks.py
koopmant: Following the same structure now.

app/grandchallenge/components/models.py
koopmant: Added `test_validate_image_kind_with_dicom_image` and `test_validate_image_kind_with_panimg_image`

app/grandchallenge/components/models.py
koopmant: I have been searching here https://github.com/comic/grand-challenge.org/pull/3465, but I have no clue. 

Maybe you remember what this line is about:
https://github.com/comic/grand-challenge.org/blob/2e92eebcbd3c0e3472fc3025740be014ff3278f2/app/grandchallenge/components/models.py#L2128

jmsmkn: I have no idea. 

That line was added in this PR: https://github.com/comic/grand-challenge.org/pull/3465/

In this commit: https://github.com/comic/grand-challenge.org/commit/6d2444888141d1ebfb93059a1726d81e4bafcf7f#diff-ba0266ac81d1dc8f7d76478fd0ecacb6fa6f8127589aab6cff4fd6ec912e6662R2129

Which ultimately stemmed from here: https://github.com/comic/grand-challenge.org/commit/d2bc25fd39252fd392b2b8498d688d8a1e29e09e#diff-ba0266ac81d1dc8f7d76478fd0ecacb6fa6f8127589aab6cff4fd6ec912e6662R2109

My guess would be this is used somewhere in the Archive Item or Display Set edit view. Could it be when an image is removed from an archive item? Or several image interfaces are presented and only one is filled?

koopmant: Removing an image or file through the form is not possible for archive items or display sets. 

I think these lines in `create_civ_for_file` makes a lot of sense, but I don't see how or where this is supported:

https://github.com/comic/grand-challenge.org/blob/61e58320fa224e29dbb641ed92cddc644405e37b/app/grandchallenge/components/models.py#L2594-L2597

koopmant: Related: https://github.com/comic/grand-challenge.org/pull/3156#issuecomment-1884507499

jmsmkn: There may be an open issue for that: https://github.com/comic/grand-challenge.org/issues/3211

It should be possible for files though no?

app/grandchallenge/components/models.py
jmsmkn: Yes, as this is a public method the inputs need to be validated.

app/grandchallenge/components/models.py
jmsmkn: I still think that it would be good to add the fall through here.

koopmant: Indeed, it will only need to raise an error, right?

app/grandchallenge/components/tasks.py
```diff
@@ -1260,27 +1276,54 @@ def add_image_to_object(  # noqa: C901
...
+            error_handler.handle_error(
+                interface=interface,
+                error_message="Image does not exist",
+                user=upload.creator,
+            )
+            logger.info("Image for dicom image set does not exist")
+            return
+    else:
+        raise ValueError(
+            "Either upload_session_pk or dicom_image_set_upload_pk must be set."
```
jmsmkn: Reducing the duplication a bit more, I don't think the separate error messages are necessary:

```python3
    if upload_session_pk is not None:
        upload = RawImageUploadSession.objects.get(pk=upload_session_pk)
        expected_status = upload.SUCCESS
        image_lookup_kwargs = {"origin_id": upload_session_pk}
    elif dicom_image_set_upload_pk is not None:
        upload = DICOMImageSetUpload.objects.get(pk=dicom_image_set_upload_pk)
        expected_status = DICOMImageSetUploadStatusChoices.COMPLETED
        image_lookup_kwargs = {"dicom_image_set__dicom_image_set_upload_id": dicom_image_set_upload_pk}
    else:
        raise ValueError(
             "Either upload_session_pk or dicom_image_set_upload_pk must be set"
        )

    if upload.status != expected_status:
        logger.info("Nothing to do: upload session was not in the expected state")
        return

    error_handler = obj.get_error_handler(linked_object=upload)

    try:
        image = Image.objects.get(**image_lookup_kwargs)
    except (Image.DoesNotExist, Image.MultipleObjectsReturned):
        error_handler.handle_error(
            interface=interface,
            error_message="Image imports should result in a single image",
            user=upload.creator,
        )
        logger.info("Upload should result in a single image")
        return
```

What do you think?

app/grandchallenge/core/error_handlers.py
```diff
@@ -125,6 +125,54 @@ def handle_error(self, *, error_message, interface, user=None):
...
+        if not dicom_image_set_upload or not isinstance(
+            dicom_image_set_upload, DICOMImageSetUpload
+        ):
+            raise RuntimeError(
+                "You need to provide a DICOMImageSetUpload instance to this error handler."
+            )
+
+        self._dicom_image_set_upload = dicom_image_set_upload
+
+        if isinstance(linked_object, ComponentJob):
```
jmsmkn: Why are only ComponentJobs handled?

koopmant: We discussed this here: https://github.com/comic/grand-challenge.org/pull/4353#discussion_r2434854825

Moved the type check and added a comment.

app/grandchallenge/components/tasks.py
koopmant: Indeed, now we have the new error handler we can reduce it further.

PR #4334, opened 2025-10-06
Alter dicom endpoint response

jmsmkn:
Return an array of objects for the image frames which allows for more flexibility in the future. Adds the image set and image frame ids for cross-referencing.

### DIFF
--- a/app/grandchallenge/cases/views.py
+++ b/app/grandchallenge/cases/views.py
@@ -130,18 +130,23 @@ class ImageViewSet(ReadOnlyModelViewSet):
 
         image_set_url = f"https://runtime-medical-imaging.{settings.AWS_DEFAULT_REGION}.amazonaws.com/datastore/{settings.AWS_HEALTH_IMAGING_DATASTORE_ID}/imageSet/{image.dicom_image_set.image_set_id}"
 
-        image_frame_requests = {}
+        image_frames = []
 
-        for frame_id in image.dicom_image_set.image_frame_ids:
-            frame_request = AWSRequest(
+        for image_frame_id in image.dicom_image_set.image_frame_ids:
+            image_frame_request = AWSRequest(
                 method="POST",
                 url=f"{image_set_url}/getImageFrame",
-                data=json.dumps({"imageFrameId": frame_id}),
+                data=json.dumps({"imageFrameId": image_frame_id}),
             )
-            medical_imaging_auth.add_auth(frame_request)
+            medical_imaging_auth.add_auth(image_frame_request)
 
-            image_frame_requests[frame_id] = self.serialize_aws_request(
-                frame_request
+            image_frames.append(
+                {
+                    "image_frame_id": image_frame_id,
+                    "get_image_frame": self.serialize_aws_request(
+                        image_frame_request
+                    ),
+                }
             )
 
         metadata_request = AWSRequest(
@@ -153,10 +158,11 @@ class ImageViewSet(ReadOnlyModelViewSet):
 
         return JsonResponse(
             {
+                "image_set_id": image.dicom_image_set.image_set_id,
                 "get_image_set_metadata": self.serialize_aws_request(
                     metadata_request
                 ),
-                "get_image_frames": image_frame_requests,
+                "image_frames": image_frames,
             }
         )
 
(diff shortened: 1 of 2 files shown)


app/grandchallenge/cases/models.py
```diff
@@ -334,19 +334,61 @@ class DICOMImageSet(UUIDModel):
         help_text="The ID of the image set in AWS Health Imaging",
         editable=False,
     )
-    image_frame_ids = models.JSONField(
+    image_frame_metadata = models.JSONField(
```
koopmant, line 340: Maybe rename this and related names in the code to `image_frames_metadata`.

app/grandchallenge/cases/views.py
```diff
@@ -130,33 +158,50 @@ def dicom(self, request, pk=None):
 
         image_set_url = f"https://runtime-medical-imaging.{settings.AWS_DEFAULT_REGION}.amazonaws.com/datastore/{settings.AWS_HEALTH_IMAGING_DATASTORE_ID}/imageSet/{image.dicom_image_set.image_set_id}"
 
-        image_frame_requests = {}
+        image_frames = []
+
+        for image_frame in image.dicom_image_set.image_frame_metadata:
```
koopmant, line 163: Suggest to rename `image_frame` to `image_frame_metadata`

app/grandchallenge/cases/views.py
```diff
@@ -130,33 +158,50 @@ def dicom(self, request, pk=None):
 
         image_set_url = f"https://runtime-medical-imaging.{settings.AWS_DEFAULT_REGION}.amazonaws.com/datastore/{settings.AWS_HEALTH_IMAGING_DATASTORE_ID}/imageSet/{image.dicom_image_set.image_set_id}"
 
-        image_frame_requests = {}
+        image_frames = []
```
koopmant: Why the change from dict to list?

app/grandchallenge/cases/models.py
```diff
@@ -1113,10 +1155,22 @@ def _get_image_set_metadata(self, *, image_set_id):
...
+                "study_instance_uid": metadata["Study"]["DICOM"][
+                    "StudyInstanceUID"
+                ],
+                "series_instance_uid": series["DICOM"]["SeriesInstanceUID"],
+                "sop_instance_uid": instance["DICOM"]["SOPInstanceUID"],
+                "stored_transfer_syntax_uid": instance[
+                    "StoredTransferSyntaxUID"
+                ],
+                "image_frame_id": frame["ID"],
+                "frame_size_in_bytes": frame["FrameSizeInBytes"],
```
koopmant, line 1175: What will `study_instance_uid`, `series_instance_uid`, `sop_instance_uid` and `frame_size_in_bytes` be used for?

koopmant (commented):
looks good, just some comments and questions

app/grandchallenge/cases/views.py
jmsmkn: So that we can add more information in there if necessary in the future. AWS do this by providing the checksums, image min and max, etc. Generally, having id's as keys is a bad design due to these inflexibilities. Now, we tell the client explicitly what sort of id this is.

app/grandchallenge/cases/models.py
jmsmkn: Fetching data from Health Imaging via the DICOM Web endpoints. See https://github.com/DIAGNijmegen/rse-roadmap/issues/436

jmsmkn: frame_size_in_bytes -> going to be useful for accounting.

app/grandchallenge/cases/models.py
jmsmkn: Seems awkward to me, metadata is already plural.

app/grandchallenge/cases/views.py
jmsmkn: I've updated the name of the list to `serialized_image_frames` to differentiate from the inner structure.

app/grandchallenge/cases/models.py
koopmant: The main issue for me was in 
`for image_frame in image.dicom_image_set.image_frame_metadata:` 
where I would have probably written:
`for image_frame_metadata in ...`
but you have already resolved that differently. So this one can be dropped.

PR #4337, opened 2025-10-07
Add option for dicom image kind socket read write path

koopmant:


app/grandchallenge/components/templates/components/partials/socket_read_write_paths.html
```diff
@@ -1,6 +1,8 @@
-{% if socket.is_image_kind %}
+{% if socket.is_panimg_kind %}
     <code>/{{ input|yesno:"input,output" }}{% if socket.relative_path %}/{{ socket.relative_path }}{% endif %}/&lt;uuid&gt;.mha</code> or
     <code>/{{ input|yesno:"input,output" }}{% if socket.relative_path %}/{{ socket.relative_path }}{% endif %}/&lt;uuid&gt;.tif</code>
+{% elif socket.is_dicom_image_kind %}
+    <code>/{{ input|yesno:"input,output" }}/{{ socket.relative_path }}/*.dcm/</code>
```
jmsmkn:
```suggestion
    <code>/{{ input|yesno:"input,output" }}/{{ socket.relative_path }}/*.dcm</code>
```

jmsmkn: Removes the trailing slash. I don't know if this will be clear enough. It will be many files, each named as `<sop_instance_uid>.dcm`.

Initially we will not have support for DICOM output.

koopmant: > Initially we will not have support for DICOM output.

Good point, thanks. 

> I don't know if this will be clear enough. It will be many files

Yes, I hoped the asterisk would indicate to the reader: "multiple files". But I guess it does not. 

We could go with something like:

`/input/images/some_dicom/{<sop_instance_uid_1>.dcm, ..., <sop_instance_uid_n.dcm>}

PR #4338, opened 2025-10-07
List kinds compatible with batch add cases form

koopmant:


app/grandchallenge/components/models.py
```diff
@@ -265,6 +265,14 @@ class InterfaceKinds(set, Enum):
         InterfaceKindChoices.BIOM,
     }
 
+    # Interfaces that can be used in batch uploading
+    batch_upload = {
```
jmsmkn: Isn't this the same as the panimg kinds in the other PR?

koopmant: Yes, initially I had batch_upload = panimg. Though I guess I'm getting ahead of myself here. (At some point we could enable batch uploading for dicom as well) I'll just use `panimg` in the code after rebasing.

app/grandchallenge/hanging_protocols/forms.py
```diff
@@ -209,6 +209,7 @@ def _get_interface_lists(self):
                 *InterfaceKinds.undisplayable,
                 *InterfaceKinds.mandatory_isolation,
                 InterfaceKindChoices.PANIMG_IMAGE,
+                InterfaceKindChoices.DICOM_IMAGE_SET,
```
koopmant, line 212: Should have been added #4339.

PR #4339, opened 2025-10-07
Cleanup dicom filtering

koopmant:
- Add the dicom image kind to `image_interfaces` for hanging protocols
- Filter on super kind for `image_inputs` in handle job status javascript
- Clean up the docstring for `InterfaceKinds`
- Clean up related name for `DICOMImageSetUploads` in `UserUploads`.


app/grandchallenge/components/models.py
```diff
@@ -176,16 +176,24 @@ class InterfaceKinds(set, Enum):
         from grandchallenge.components.models import INTERFACE_KIND_JSON_EXAMPLES
 
         print("Interface kinds that are images:")
+        print("")
```
jmsmkn: Maybe tidy this up a bit by using `\n` in the previous calls? It is also missing from `print("Interface kinds that are files:")`

koopmant: I did that initially, but I forgot to escape the `\n` so the docstring generation failed. I'll escape them properly.

jmsmkn:
@koopmant This may have introduced flakeyness in the tests, see the tests on main:

https://github.com/comic/grand-challenge.org/actions/runs/18340579083/job/52234874253

I also see this after merging main into https://github.com/comic/grand-challenge.org/pull/4334

PR #4342, opened 2025-10-08
Async provisioning of component inputs

jmsmkn:
See https://github.com/DIAGNijmegen/rse-roadmap/issues/436

### DIFF
--- a/app/grandchallenge/components/backends/base.py
+++ b/app/grandchallenge/components/backends/base.py
@@ -1,4 +1,5 @@
 import asyncio
+import functools
 import io
 import json
 import logging
@@ -15,6 +16,7 @@ import aioboto3
 import boto3
 import botocore
 from asgiref.sync import async_to_sync
+from botocore.config import Config
 from django.conf import settings
 from django.core.exceptions import SuspiciousFileOperation, ValidationError
 from django.db import transaction
@@ -44,7 +46,9 @@ from grandchallenge.core.utils.error_messages import (
 logger = logging.getLogger(__name__)
 
 MAX_SPOOL_SIZE = 1_000_000_000  # 1GB
-CONCURRENCY = 128
+
+CONCURRENCY = 50
+BOTO_CONFIG = Config(max_pool_connections=120)
 
 
 class JobParams(NamedTuple):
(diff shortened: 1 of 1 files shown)


app/grandchallenge/components/backends/base.py
```diff
@@ -370,19 +370,7 @@ def _provision_inputs(self, *, input_civs, input_prefixes):
...
-                self._copy_input_file(src=civ.file, dest_key=key)
-            else:
-                with io.BytesIO() as f:
-                    f.write(json.dumps(civ.value).encode("utf-8"))
-                    f.seek(0)
-                    self._s3_client.upload_fileobj(
-                        Fileobj=f,
-                        Bucket=settings.COMPONENTS_INPUT_BUCKET_NAME,
-                        Key=key,
-                    )
```
jmsmkn, line 385: @koopmant watch out for things like this in the codebase! Looks like we don't use super_kind everywhere that we need to. I have refactored this one to look at the super kind.

app/grandchallenge/components/backends/base.py
```diff
@@ -411,38 +530,23 @@ def _with_inputs_json(self, *, input_civs):
...
-                        "output_prefix": self._io_prefix,
-                    }
-                ]
-            ).encode("utf-8")
-        )
-        self._s3_client.upload_fileobj(
-            f, settings.COMPONENTS_INPUT_BUCKET_NAME, self._invocation_key
+    @staticmethod
+    def _get_copy_input_object_task(*, src, target_key):
+        return functools.partial(
```
koopmant, line 535: I did not know this existed, good to know!

app/grandchallenge/components/backends/base.py
```diff
@@ -362,27 +404,43 @@ def _get_key_and_relative_path(self, *, civ, input_prefixes):
...
+            "s3", endpoint_url=settings.AWS_S3_ENDPOINT_URL, config=BOTO_CONFIG
+        ) as s3_client:
+            async with asyncio.TaskGroup() as task_group:
+                for task in tasks:
+                    task_group.create_task(
+                        task(
+                            semaphore=semaphore,
+                            s3_client=s3_client,
+                        )
+                    )
```
koopmant, line 422: I don't know much about this, so I'm assuming you're doing it correctly.

app/grandchallenge/components/backends/base.py
koopmant: Thanks for the heads up, will do! How did you find this?

app/grandchallenge/components/backends/base.py
jmsmkn: I only remembered about it this morning!

app/grandchallenge/components/backends/base.py
jmsmkn: Indeed, part of code review is to spread knowledge. I think this is the way to go, see:

- https://kissgyorgy.github.io/s3-asyncio-client/guide/performance/#session-reuse
- https://hynek.me/articles/waiting-in-asyncio/#task-groups

app/grandchallenge/components/backends/base.py
jmsmkn: I needed to modify this bit of code and noticed it. You could maybe search for uses of `.value`, `.file` or `.image`, but there are likely many instances of that.

PR #4347, opened 2025-10-10
Notify user when dicom upload fails

koopmant:


koopmant:
I removed all the cleanup for follows and notifications. 
These tests still pass:
`test_follow_clean_up_after_object_removal`
`test_follow_clean_up_after_forum_removal`
`test_follow_clean_up_after_challenge_removal`

The notification cleanup tests fail. 

I don't yet know why the test_follow_cleanup tests pass. It seems like follows get cleaned up automatically, but I don't know how.

PR #4351, opened 2025-10-15
Provision DICOM files for component jobs

jmsmkn:
See https://github.com/DIAGNijmegen/rse-roadmap/issues/436

### DIFF
--- a/app/tests/components_tests/test_backends.py
+++ b/app/tests/components_tests/test_backends.py
@@ -10,6 +10,7 @@ import aioboto3
 import botocore
 import pytest
 from botocore.auth import SigV4Auth
+from django.core.exceptions import SuspiciousFileOperation
 from django.template.defaultfilters import title
 
 from grandchallenge.components.backends.base import (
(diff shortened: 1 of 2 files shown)


app/tests/components_tests/test_backends.py
```diff
@@ -322,3 +334,378 @@ def test_invocation_json(settings):
...
+    dicom_civ = dicom_interface.create_instance(
+        image=ImageFactory(
+            dicom_image_set=DICOMImageSetFactory(
+                image_frame_metadata=[
+                    {
+                        "image_frame_id": "123",
+                        "frame_size_in_bytes": 1337,
+                        "study_instance_uid": "123",
+                        "series_instance_uid": "123",
+                        "sop_instance_uid": "../fds",
```
koopmant, line 695: Is this possible?

jmsmkn: Unsure, I don't know how the sop instance uids are generated (there is an open question for Chris on that) but better to be safe than sorry.

PR #4353, opened 2025-10-15
Move handling of linked object from upload session to error handler

koopmant:


app/tests/components_tests/test_tasks.py
```diff
@@ -463,35 +470,30 @@ def test_add_image_to_object_marks_job_as_failed_on_validation_fail(
...
     assert ComponentInterfaceValue.objects.filter(interface=ci).count() == 0
-    us.refresh_from_db()
-    assert us.status == RawImageUploadSession.FAILURE
-    assert us.error_message == error_message
-    obj.refresh_from_db()
-    assert obj.status == obj.CANCELLED
-    assert obj.error_message == "One or more of the inputs failed validation."
+    job.refresh_from_db()
+    assert job.status == job.CANCELLED
+    assert job.error_message == "One or more of the inputs failed validation."
```
jmsmkn, line 494: Why has the test on the upload session been removed? The upload session should still be marked as failed. A job can have multiple image inputs, and thus multiple upload sessions associated with it. The user would need to know which one of those upload sessions failed.

app/grandchallenge/core/error_handlers.py
```diff
@@ -93,37 +93,48 @@ class RawImageUploadSessionErrorHandler(ErrorHandler):
...
-        from grandchallenge.evaluation.models import Evaluation
 
-        if self._linked_object:
-            if isinstance(self._linked_object, Evaluation):
-                logger.error(error_message, exc_info=True)
-                error_message = "Input validation failed"
+        if isinstance(linked_object, ComponentJob):
+            self._linked_object = linked_object
+        else:
+            self._linked_object = None
```
jmsmkn: Why are only ComponentJobs handled?

app/tests/components_tests/test_tasks.py
koopmant: I moved it to `test_add_image_to_object_updates_upload_session_on_validation_fail`.

app/grandchallenge/core/error_handlers.py
koopmant: That is what the code was doing. It changed the error message for evaluations and marked `ComponentJob` instances as cancelled. That is what the `EvaluationCIVErrorHandler` and `JobCIVErrorHandler` handler do, so I'm calling those instead. 

If the linked object would be an archive item or display set, we'd get a `FallbackCIVValidationErrorHandler` below. That sends a notification to the user, but that is already done by the upload session here. So we want to skip in that case to avoid sending multiple notifications. 

I could move this check to line 130, would that make it clearer?

jmsmkn: Yes I think so, also adding your explanation as a comment would help future developers.

jmsmkn: I'm noticing that with Anne away we're missing a lot of the "why". I think that the "what" is usually clear in the codebase from the function names and organisation, and if you're using comments to explain "what" it usually means the code can be improved. I'm trying to add more "why"-type comments myself.

koopmant: Yes, I think that would be great. Will aim to add those more.

jmsmkn (approved):
LGTM!

PR #4354, opened 2025-10-15
Refactor ValuesForInterfacesMixin to LinkedComponentInterfaceMixin

koopmant:
Remove values from `ValuesForInterfacesMixin` because we only use the keys in `values_for_interfaces`, i.e. the component interfaces, and return the interfaces as a queryset instead of converting them to a list. 

Use `linked_component_interfaces`, which returns a queryset. 
Change `linked_component_interfaces` on `Algorithm` and `Phase` to return a queryset (was a set) as well, to avoid confusion over return type for properties with the same name.

### DIFF
--- a/app/grandchallenge/components/forms.py
+++ b/app/grandchallenge/components/forms.py
@@ -175,7 +175,9 @@ class MultipleCIVForm(Form):
 
         # add fields for all interfaces that already exist on
         # other display sets / archive items
-        for interface in base_obj.linked_component_interfaces:
+        for interface in base_obj.linked_component_interfaces.order_by(
+            "title"
+        ):
             current_value = None
 
             prefixed_interface_slug = (


app/grandchallenge/reader_studies/models.py
```diff
@@ -825,16 +825,14 @@ def create_civ_set_batch_url(self):
...
-            set(
-                self.questions.filter(interface__isnull=False).values_list(
-                    "interface__slug", flat=True
-                )
+    def linked_component_interfaces(self):
+        return super().linked_component_interfaces.union(
+            ComponentInterface.objects.filter(
+                pk__in=self.questions.filter(
+                    interface__isnull=False
+                ).values_list("interface__pk", flat=True)
```
jmsmkn: Isn't this equivalent?

```python3
ComponentInterface.objects.filter(
    question__in=self.questions
).distinct()
```

app/grandchallenge/components/forms.py
```diff
@@ -340,7 +341,9 @@ def __init__(
         qs = (
             ComponentInterface.objects.all()
             .filter(**socket_filter_kwargs)
-            .exclude(slug__in=base_obj.values_for_interfaces.keys())
+            .exclude(
+                slug__in=base_obj.linked_component_interfaces.values("slug")
```
jmsmkn, line 347: Doesn't calling `.values(...)` cause re-evaluation of the queryset and another hit?

app/grandchallenge/components/views.py
```diff
@@ -139,7 +139,9 @@ def get_queryset(self):
             qs = (
                 ComponentInterface.objects.all()
                 .filter(**extra_filter_kwargs)
-                .exclude(slug__in=obj.values_for_interfaces.keys())
+                .exclude(
+                    slug__in=obj.linked_component_interfaces.values("slug")
```
jmsmkn, line 143: Doesn't calling `.values(...)` cause re-evaluation of the queryset and another hit?

app/grandchallenge/components/forms.py
```diff
@@ -175,11 +175,12 @@ def __init__(self, *args, instance, base_obj, user, **kwargs):  # noqa C901
 
         # add fields for all interfaces that already exist on
         # other display sets / archive items
-        for slug in base_obj.values_for_interfaces.keys():
+        for interface in base_obj.linked_component_interfaces:
```
jmsmkn: Isn't ordering important here?

koopmant: I don't think they were ordered before, these fields appear in random order on the form. But yes, a consistent order would be nice.

app/grandchallenge/components/forms.py
koopmant: I don't think so? 

As for an alternative, I had used `.difference()` earlier, but I cannot use that because this queryset is filtered again later.

jmsmkn: If you have already iterated over the queryset before hitting this point then it will. It's also unclear to me if `linked_component_interfaces` is a set or a queryset, sometimes it is one and sometimes not, so unsure if you can consistently use `.values` here.

By iterating (like what was done before) then it doesn't matter if it is a set or queryset.

koopmant: To clarify, it's always a queryset. I meant `QuerySet.difference()`. (Note I also introduced usage of `QuerySet.union()` below. Both come with the caveat that the resulting queryset cannot be chained, so we might want to avoid them both.)

For my understanding, I turned to using the queryset directly because before we used the list of slugs and then had to do `interface = ComponentInterface.objects.filter(slug=slug).get()` for each one in a loop. 
With the queryset we have the risk of re-evaluating if we use it multiple times, but we should still save a number of hits because of the previous usage of the slugs, right?

koopmant: > To clarify, it's always a queryset.

I'm sorry, I see I am mistaken here.

koopmant: Right, so we have a property on `Algorithm` and `Phase` with the same name that returns a set. To be fair, this inconsistency has always been there, but `linked_component_interfaces` from this mixin wasn't used anywhere. 

Currently these properties are used in different places, so a different name should resolve the issue.

app/grandchallenge/reader_studies/models.py
koopmant: Yes, thanks!

app/grandchallenge/components/forms.py
koopmant: I opted for returning a queryset for `Algorithm.linked_component_interface` and `Phase.linked_component_interfaces` as well. So it is always a queryset now. 

> Note I also introduced usage of QuerySet.union() below. Both come with the caveat that the resulting queryset cannot be chained, so we might want to avoid them both.

I checked, and currently it is not a problem because we either use these querysets to filter other another queryset, or to iterate over them. We do not extend them anywhere.

jmsmkn: Ah okay - is it worth renaming them or are they really returning the same sort of thing? I was definitely confused by the different return types.

When using `union` isn't there a possibility that rows will be duplicated in the resulting queryset?

koopmant: Now that they have the same type, keeping the same name is fitting. It basically means "all component interfaces that an object is indirectly linked with through relations with other objects".

So for an archive or reader study it is the union of component interfaces for all archive items/display sets. For algorithms and phases it's for all possible inputs and outputs. 

> When using union isn't there a possibility that rows will be duplicated in the resulting queryset?

There shouldn't be. From the Django docs:

> To allow duplicate values, use the all=True argument.

app/grandchallenge/components/forms.py
koopmant: Ordering by title now.

app/grandchallenge/components/forms.py
jmsmkn: Nice, didn't see that!

jmsmkn (approved):
LGTM!

PR #4355, opened 2025-10-16
Add DICOM related fields to serializer

koopmant:
Add `image_name` and `user_uploads` fields to `ComponentInterfaceValuePostSerializer`

Also:
- Validate the provided fields in `ComponentInterfaceValuePostSerializer`
- Unify data conversion from `ComponentInterfaceValuePostSerializer` deserialized data to `CIVData` in `convert_deserialized_civ_data`.

Related to https://github.com/DIAGNijmegen/rse-roadmap/issues/433

### DIFF
--- a/app/tests/components_tests/test_serializers.py
+++ b/app/tests/components_tests/test_serializers.py
@@ -2,6 +2,7 @@ from contextlib import nullcontext
 
 import pytest
 from guardian.shortcuts import assign_perm
+from rest_framework.exceptions import ValidationError
 
 from grandchallenge.cases.models import RawImageUploadSession
 from grandchallenge.components.models import (
@@ -254,15 +255,21 @@ TEST_DATA = {
 
 
 def test_civ_post_validate_provided_fields():
-    values = [
-        {},  # no values is also valid
-        {"file": "https://some-api-url/"},
-        {"image": "https://some-api-url/"},
-        {"upload_session": "https://some-api-url/"},
-        {"user_upload": "https://some-api-url/"},
-        {"image_name": "foobar", "user_uploads": ["https://some-api-url/"]},
-        {"value": 42},
-    ]
+    with pytest.raises(
+        ValidationError, match="An interface must be specified"
+    ):
+        ComponentInterfaceValuePostSerializer._validate_provided_fields(
+            attrs={}
+        )
+
+    with pytest.raises(
+        ValidationError,
+        match="You must provide at least one of",
+    ):
+        ComponentInterfaceValuePostSerializer._validate_provided_fields(
+            attrs={"interface": "my-socket"}
+        )
+
     payload_empty = {
         "interface": "my-socket",
         "file": None,
@@ -274,12 +281,40 @@ def test_civ_post_validate_provided_fields():
         "value": None,
     }
 
+    # All None values is valid
+    with nullcontext():
+        ComponentInterfaceValuePostSerializer._validate_provided_fields(
+            attrs=payload_empty
+        )
+
+    values = [
+        {"file": "https://some-api-url/"},
+        {"image": "https://some-api-url/"},
+        {"upload_session": "https://some-api-url/"},
+        {"user_upload": "https://some-api-url/"},
+        {"image_name": "foobar", "user_uploads": ["https://some-api-url/"]},
+        {"value": 42},
+    ]
+
     for value in values:
         payload = {**payload_empty, **value}
+
+        # providing one value is valid
         with nullcontext():
             ComponentInterfaceValuePostSerializer._validate_provided_fields(
                 attrs=payload
             )
+
+        # providing multiple values is invalid
+        other_values = values.copy()
+        other_values.remove(value)
+        for other_value in other_values:
+            with pytest.raises(
+                ValidationError, match="You can only provide one of"
+            ):
+                ComponentInterfaceValuePostSerializer._validate_provided_fields(
+                    attrs={**payload, **other_value}
+                )
 
 
 @pytest.mark.django_db


app/grandchallenge/algorithms/serializers.py
```diff
@@ -310,35 +311,30 @@ def validate_inputs_and_return_matching_interface(self, *, inputs):
     @staticmethod
     def reformat_inputs(*, serialized_civs):
         """Takes serialized CIV data and returns list of CIVData objects."""
-        possible_keys = [
-            "image",
-            "value",
-            "file",
```
koopmant, line 316: I just noticed I removed "file" but did not add it below. Will add it again. None of the tests caught it, so will add a test too.

koopmant: @chrisvanrun Looking at the api, I see we never use file. Not even in `FileFromSVCreateStrategy`, where we also use `user_upload`:

```python
# Cannot link a file directly to file, so we download it first
```

Do you know if we actually support using `file` on `ComponentInterfaceValuePostSerializer`?

chrisvanrun: I have no clue. AFAIK, this needed to go via the `UserUpload`. There **seems** to be some validation for having file in the `ComponentInterfaceValuePostSerializer` but I am having a hard time to find any actual handling in the `CIVSetPostSerializerMixin`.

~I suspect it gets automagically handled via the actual `ComponentInterfaceValue.file` model field or its adaptation of that field for the serializer.~

~I wonder if the follow-up validators for passing a 'file' directly are hooked up correctly!~

I think it validates, but just nothing happens, but I have not traced it completely. (A lotta buts here)

koopmant: All right, thanks. What did you use to test it with?

chrisvanrun: I hadn't actually tested this. AFAIK gcapi generally doesn't allow you to add bytes directly to the params of the gcapi calls, rather there are some routines that push to pre-signed urls and create a UserUpload. I think that is really the way to go since sending large files in one piece is a no-go.

Tried my hand quickly on some curl calls but I have limited experience with adding bytes there.

app/grandchallenge/algorithms/serializers.py
```diff
@@ -310,35 +311,32 @@ def validate_inputs_and_return_matching_interface(self, *, inputs):
...
-
-            if not found_keys:
-                raise serializers.ValidationError(
-                    f"You must provide at least one of {possible_keys}"
-                )
-
-            if len(found_keys) > 1:
-                raise serializers.ValidationError(
-                    f"You can only provide one of {possible_keys} for each interface."
-                )
```
jmsmkn, line 335: Where has this validation gone?

app/grandchallenge/algorithms/serializers.py
```diff
@@ -310,35 +311,32 @@ def validate_inputs_and_return_matching_interface(self, *, inputs):
     @staticmethod
     def reformat_inputs(*, serialized_civs):
```
jmsmkn: Should reformat inputs be a separate method that can be used in `CIVSetPostSerializerMixin`?

app/grandchallenge/algorithms/serializers.py
koopmant: I thought I was handling this in `CIVData.validate`, but I didn't realize I am only passing along one of the values, so I will never catch it there...

app/grandchallenge/algorithms/serializers.py
koopmant: Yes, that's probably a good idea for now. I was planning on implementing it in the CIVPostSerializer itself, but that should not be done in this PR (if at all)

app/grandchallenge/algorithms/serializers.py
koopmant: Brought this back in `reformat_serialized_civ_data` and extended it.

app/grandchallenge/algorithms/serializers.py
koopmant: Moved, extended and tested.

app/grandchallenge/algorithms/serializers.py
koopmant: I kept running into issues while trying to test this. It's not important for this PR, so I'll drop it.

app/grandchallenge/algorithms/serializers.py
koopmant: I've renamed it to `convert_deserialized_civ_data`.

app/grandchallenge/algorithms/serializers.py
koopmant: I moved the validation to `ComponentInterfaceValuePostSerializer`

app/grandchallenge/components/models.py
```diff
@@ -446,7 +450,7 @@ def is_panimg_kind(self):
 
     @property
     def is_dicom_image_kind(self):
```
chrisvanrun, line 448: dicom_image_kind =/= dicom. What is the rational for doing this?

app/tests/algorithms_tests/test_serializers.py
```diff
@@ -240,22 +239,14 @@ def test_algorithm_job_post_serializer_create(
...
-            {
-                "interface": ci_img1.slug,
-                "upload_session": upload.api_url,
-                # Other attributes are optionally None
-                "user_upload": None,
-                "image": None,
-                "value": None,
-                "file": None,
-            },
+            {"interface": ci_img1.slug, "upload_session": upload.api_url},
```
chrisvanrun: Requests like this are to be expected from gcapi. Why was this removed?

app/tests/components_tests/test_serializers.py
```diff
@@ -304,34 +309,55 @@ def test_civ_post_value_validation(kind):
...
-    # setup
+def test_civ_post_value_required_validation(kind):
     interface = ComponentInterfaceFactory(
         kind=kind,
-        store_in_database=store_in_database,
+        store_in_database=True,
     )
 
-    payload = {"interface": interface.slug}
+    for kwargs in [
```
chrisvanrun, line 385: You could use a double parameterization for this as that better matches the testing style where we try to push things.

app/grandchallenge/components/models.py
koopmant: I figured I create the group now so I can already start using it in tests, then when we add new dicom kinds later (e.g. segmentations) they are automatically included in the tests.

app/tests/algorithms_tests/test_serializers.py
koopmant: We had two places where we handled the request data: one was `JobPostSerializer.reformat_inputs` and one was `CIVSetPostSerializerMixin.update`. In the job post serializer we had some validation that did not allow sending multiple keys (e.g. `upload_session` and also `image`). I harmonized it in `convert_deserialized_civ_data` where I now also apply this validation. That makes the request data in this test invalid because although the other attributes are None, they are still supplied. 

Should I allow that?

app/tests/components_tests/test_serializers.py
koopmant: That would have been my default until I got a comment from @jmsmkn in an earlier PR about making a test less efficient. But maybe I did not understand it correctly; when should I be worried about repeating the test setup?

app/tests/algorithms_tests/test_serializers.py
chrisvanrun: >  In the job post serializer we had some validation that did not allow sending multiple keys (e.g. upload_session and also image). I harmonized it in 

Good that you've harmonized them! The way that gcapi relies on the `ComponentInterfacePostRequest` object which (currently) requires every property to be set, so it sends them all set to None.

Having said that, afaik we are currently sending requests in the same way to the job end-point. I'll check.

app/grandchallenge/algorithms/serializers.py
```diff
@@ -306,42 +308,3 @@ def validate_inputs_and_return_matching_interface(self, *, inputs):
...
-            "value",
-            "file",
-            "user_upload",
-            "upload_session",
-        ]
-
-        data = []
-        for civ in serialized_civs:
-            found_keys = [
-                key for key in possible_keys if civ.get(key) is not None
```
chrisvanrun, line 324: This filters out keys that are `None`: the new version does not.

app/grandchallenge/components/models.py
chrisvanrun: Make sense, but then I wouldn't make `is_dicom_image_kind` such not be depend on membership to that group as that might break things in the future.

app/tests/algorithms_tests/test_serializers.py
chrisvanrun: The old `JobPostSerializer.reformat_inputs` did allow for values to be `None` (see other in-line comment).

app/tests/components_tests/test_serializers.py
chrisvanrun: AFAIK views are terribly expensive because of all the middleware and rendering. So if you can add several test cases in a single view call for testing that is a bit more efficient. Having said that, I don't know the original context of the other PR.

Perhaps the singular CIV factory call makes this construct a tiny bit more efficient? Comes at the cost though of not being able to distribute this test cases over multiple pytest runners. The latter might not matter much since we have so many tests already.

General guideline is: "Try and get tests into a Arrange, Act, Assert construct. Avoid long chains of mutations and asserts. ".

Imho your current construct covers that.

app/grandchallenge/algorithms/serializers.py
koopmant: Doh, you're right! Thanks for pointing it out. We do have the issue here of not accepting `None` values for json kind sockets...

app/grandchallenge/components/models.py
koopmant: > such not be depend on membership to that group

Sorry, I don't understand.

app/grandchallenge/algorithms/serializers.py
chrisvanrun: We do!

app/tests/algorithms_tests/test_serializers.py
koopmant: Thanks for checking. According to the schema, those properties are not required (only `interface` is):

```json
"ComponentInterfaceValuePostRequest": {
	"type": "object", 
	"description": "Serializes images with hyperlinks for external usage\nExpects interface_slug, to allow creating a ComponentInterfaceValue with an\nexisting ComponentInterface", 
	"properties": {
		"interface": {"type": "string", "minLength": 1}, 
		"value": {"nullable": true}, 
		"file": {"type": "string", "format": "binary", "nullable": true}, 
		"image": {"type": "string", "format": "uri", "nullable": true}, 
		"upload_session": {"type": "string", "format": "uri", "writeOnly": true, "nullable": true}, 
		"user_upload": {"type": "string", "format": "uri", "writeOnly": true, "nullable": true}
	}, 
	"required": ["interface"]
},
```

But as far as I understand it, datamodel-code-generator does not correctly convert this at the moment. So my first thought is we should fix it there?

app/tests/components_tests/test_serializers.py
jmsmkn: Big "it depends". Creating objects in the db is extremely slow. View testing is pretty slow. However, those are the most useful tests. If you have a lot of set up, then it is usually best not to parameterize so that you can avoid those set up costs. You can avoid the set up costs if you create fixtures and scope them to the module/session, but then that can introduce ordering dependendencies. Fixtures will default to the function scope and so be recreated for each parameter if you use that.

jmsmkn: If you're not using `@pytest.mark.django_db` then that probably means that parameterization is cleaner and just as fast.

koopmant: Thanks for the explanations both. I'll keep it like this then. 
(I had also tried earlier to write the test without creating the socket in the database, but the serializer field will look up the value, so that didn't work of course.)

app/tests/algorithms_tests/test_serializers.py
chrisvanrun: Not sure, we'd have to change things up to set non-required properties to some not `None` default (e.g. `MISSING = object()`) and during serialization on the gcapi side filter those out. The datamodel_code_generator doesn't seem to directly support that.

Likely doable, but the alternative is creating a more flexible and robust api.

Technically, the schema says we can send None values. As such, we'd break the current gcapi release if the current PR gets merged/deployed.

koopmant: > the schema would suggest it is allowed.

Oh yeah, completely forgot about that. Thanks, I'll change it to what we had with `CIVSetPostSerializerMixin.update`; there we would pass `None` as the value if all fields are `None`.

app/grandchallenge/components/models.py
chrisvanrun: The property `is_dicom_image_kind` is too specific for a dicom group. I can imagine we might have a dicom_report somewhere in the future, and then that would test positive with this property.

koopmant: I still don't completely follow. But let's tackle this when we actually need it. I'll remove the group for now, since it's unclear what it is supposed to represent.

app/tests/algorithms_tests/test_serializers.py
koopmant: Should be fixed now!

chrisvanrun (approved):
LGTM! Did you ever find out if the 'file' property is being used somewhere?

Also, should we add and fix the value=None option in a follow-up PR?

app/grandchallenge/components/serializers.py
```diff
@@ -25,6 +27,36 @@
 logger = logging.getLogger(__name__)
 
 
+def convert_deserialized_civ_data(*, deserialized_civ_data):
+    """Takes deserialized CIV data and returns list of CIVData objects."""
```
jmsmkn, line 31: Isn't the data serialized at this point? E.g. it had been serialized to a dictionary and you want to create `CIVData` objects?

app/grandchallenge/components/serializers.py
```diff
@@ -25,6 +27,36 @@
...
+
+        if keys_not_none == {"image_name", "user_uploads"}:
+            value = DICOMUploadWithName(
+                name=civ["image_name"],
+                user_uploads=civ["user_uploads"],
+            )
+        elif not keys_not_none:
+            value = None
+        else:
+            value = civ[keys_not_none.pop()]
```
jmsmkn: Can't this result in strange behaviour if `len(keys_not_none) != 1`?

app/grandchallenge/components/serializers.py
```diff
@@ -125,23 +159,31 @@ def __init__(self, *args, **kwargs):
                 codename="change_userupload",
             )
 
+            # Can't find a good way to set the dynamically generated queryset
+            # for a field that gets used on POST, so create it here
```
jmsmkn: I think that the way to do this is to create the field as normal but set the queryset to `UserUpload.objects.none()`, then update the queryset here.

app/grandchallenge/components/serializers.py
```diff
@@ -175,6 +221,58 @@ def validate(self, attrs):
...
+    @staticmethod
+    def _validate_panimg_image(*, attrs, interface):
+        if not any(
+            [
+                attrs.get("image"),
+                attrs.get("upload_session"),
+            ]
+        ):
+            raise serializers.ValidationError(
+                f"upload_session or image are required for interface "
```
jmsmkn, line 275: Didn't we also want to support `user_uploads` here for new versions of GCAPI, and then gracefully deprecate `upload_session`?

app/grandchallenge/components/serializers.py
koopmant: I tried that, but for some reason that doesn't work. I gave up after a while, although I'd like to understand the mechanics here.

app/grandchallenge/components/serializers.py
koopmant: This confused me a lot, that's why I renamed the variable as a hint. So I'm glad you asked. 

The data has already been past `to_internal_value` and has been converted from a dict of primitive datatypes to a dict of native values. So I'd say all fields have already been deserialized (right?), but we add another conversion to create a single CIVData object. 

I initially thought it would be nice if the serializer returns a CIVData object directly, but I'm not sure if that would be an improvement.

app/grandchallenge/components/serializers.py
koopmant: Yes, thanks. I'm relying on validation in `ComponentInterfaceValuePostSerializer._validate_provided_values`. I should probably just repeat the validation here.

app/grandchallenge/components/serializers.py
koopmant: Thanks for mentioning it, I almost forgot. Let me add this in another PR though, because I need to check if this requires more changes.

app/grandchallenge/components/serializers.py
jmsmkn: Serialization makes me think of DRF where you are going from Django objects to stuff (json, csv, excel, xml, etc). For me, you still do not have Django objects here, so the deserialization is at any point incomplete. Maybe `get_civ_data(*, form_inputs):`?

app/grandchallenge/components/serializers.py
jmsmkn: Yes, it's best not to rely on an earlier validation being called, but I don't think you have to repeat it, something like:

```python3
if keys_not_none = {"foo", "bar"}:
   do_stuff()
elif len(keys_not_none) == 1:
   value = civ[keys_not_none.pop()]
elif len(keys_not_none) == 0:
   value = None
else:
   raise ValueError(...)
```

app/grandchallenge/components/serializers.py
jmsmkn: Isn't it done here? https://github.com/comic/grand-challenge.org/blob/c1569b26827c402652e75288fc23fd9de810d4f1/app/grandchallenge/cases/serializers.py#L89-L115

koopmant: Yes, that's exactly what I tried. For some reason it worked fine with a field that has `many=False`, but with `many=True` it broke and no longer updated the queryset in __init__. If I started with `UserUpload.objects.none()` I would find no objects and if I started with `UserUpload.objects.all()` I would find objects the user does not have permission for. 

I then copied this section:
https://github.com/comic/grand-challenge.org/blob/c1569b26827c402652e75288fc23fd9de810d4f1/app/grandchallenge/cases/serializers.py#L139-L153

app/grandchallenge/components/serializers.py
chrisvanrun: Ah, I almost forgot as well! =P

app/grandchallenge/components/serializers.py
jmsmkn: Okay! Weird, I have no idea why there is a difference. If you find out please let me know.

app/grandchallenge/components/serializers.py
koopmant: > you still do not have Django objects here

I'm not sure if I follow. When I print the values that I receive in this function they have already been converted to django objects. For instance, `interface` is a `ComponentInterface` object at this point, and `image` an `Image` object.

jmsmkn: Ah okay, then my mistake!

app/grandchallenge/components/serializers.py
chrisvanrun: I think I found it while figuring out the API schema was off:

The `many=True` causes some wrapper magic via a `__new__` construct. I suspect that you are to use `child_relation` to access the wrapped.

```Python
self.fields["uploads"].child_relation.queryset=filter_by_permission( 
         queryset=UserUpload.objects.filter( 
             status=UserUpload.StatusChoices.COMPLETED 
         ), 
         user=user, 
         codename="change_userupload", 
)
```

chrisvanrun:
> I don't think that makes sense as then you would have two representations of nothing for `user_uploads`: `[]` and `None`. We should just stick to one which would be `[]`.

I agree, one option that signals: empty.

koopmant:
>
> ```
> image_name:
>   type: string
>   minLength: 1
> ```

Huh, that's not what I expected. But setting allow_null on a CharField has the same problem, we then have two representations of nothing: `""` and `None`.

koopmant:
> With `minLength: 1` we don't have `""` and `None` as nothing: just `None` as `""` would not be a valid value.

Ah, you're right. 

~~Although allow_null is false by default, so I think None in not allowed at the moment either?~~ (of course it isn't, that's why you suggested setting allow_null to true...)

Instead of allow_null I can add allow_blank to get this, would that work?
```
        user_uploads:
          type: array
          items:
            type: string
            format: uri
            writeOnly: true
          writeOnly: true
        image_name:
          type: string
```

koopmant:
Err, I'm also having a hard time validating the incoming data. If all None values mean "nothing" it's easy to filter them. 

I would be fine with allow_null then, but this hack is not working for me:
```python


    user_uploads = serializers.HyperlinkedRelatedField(
        queryset=UserUpload.objects.none(),
        view_name="api:upload-detail",
        required=False,
        write_only=True,
        many=True,
        allow_null=True,
    )

    class Meta:
        model = ComponentInterfaceValue
        fields = [
            ...
            "user_uploads",
        ]

        if "request" in self.context:
            self.fields["user_uploads"].allow_null = True
```

I'm setting `allow_null` twice, but it's not getting set.

app/grandchallenge/components/serializers.py
koopmant: Thanks! This works.

koopmant:
We currently have:

```yaml
user_uploads:
  type: array
  items:
    type: string
    format: uri
    writeOnly: true
  writeOnly: true
  nullable: true
image_name:
  type: string
  nullable: true
  minLength: 1
```

I've added some tests to make sure we allow `None` but not empty values (`""`. `[]`, `[None]`, `[""]`)

Regarding the allow_null field on the many relation. There is a list of kwargs that is passed on when `many=True` here:
https://github.com/encode/django-rest-framework/commit/49fae230005bba4607f425d90de77363d6b8659e#diff-37b984e11193bccbbb90121bf903bc5b94e2fa646888088ef8f84f995d0abb2aR16-R20

`allow_null` is not in this list, but I have found no information about why. I did find an old PR where it was implemented, just not for related fields:
https://github.com/encode/django-rest-framework/issues/2673#issuecomment-1548997998

app/tests/components_tests/test_serializers.py
```diff
@@ -246,6 +253,35 @@
...
+        "user_uploads": None,
+        "value": None,
+    }
+
+    for value in values:
+        payload = {**payload_empty, **value}
+        with nullcontext():
+            ComponentInterfaceValuePostSerializer._validate_provided_fields(
+                attrs=payload
+            )
```
chrisvanrun, line 306:
```suggestion
        ComponentInterfaceValuePostSerializer._validate_provided_fields(
            attrs=payload
        )
```

Alternatively, it might be good to add a 'negative' test here, where you check that any invalidity is really captured. Then again, I think a later test does this indirectly already.

koopmant: Thanks! I added some negative tests. 

I kept the `with nullcontext()`, because it indicates that we are asserting something there.

jmsmkn (approved):
LGTM!

PR #4360, opened 2025-10-17
Rename civ_data argument to avoid confusion with civ and CIVData type

koopmant:


app/grandchallenge/components/models.py
```diff
@@ -1521,13 +1521,13 @@ def only_completed(self):
         )
 
     @staticmethod
-    def retrieve_existing_civs(*, civ_data):
+    def retrieve_existing_civs(*, values):
```
jmsmkn: Isn't values an overloaded term?

koopmant: good point, thanks.

app/grandchallenge/algorithms/models.py
```diff
@@ -1256,7 +1256,7 @@ def get_civ_for_interface(self, interface):
         return self.inputs.get(interface=interface)
 
     def validate_values_and_execute_linked_task(
-        self, *, values, user, linked_task=None
+        self, *, civ_data_objects, user, linked_task=None
```
jmsmkn, line 1259: Should we update the method name too?

koopmant: yes

jmsmkn (approved):
LGTM!

PR #4363, opened 2025-10-17
Clean upload session object created in task

koopmant:
Also change `error_message` text field to allow blank but not null.


app/grandchallenge/cases/models.py
```diff
@@ -131,13 +131,13 @@ class RawImageUploadSession(UUIDModel):
         blank=True, null=True, default=None, editable=False
     )
 
-    error_message = models.TextField(blank=False, null=True, default=None)
+    error_message = models.TextField(blank=True)
```
jmsmkn, line 134: Not sure we can do this - as `None` is the default there will be fields in the database with this value. It will require a data migration.

app/grandchallenge/components/models.py
```diff
@@ -2541,9 +2541,9 @@ def _create_civ_for_image(  # noqa: C901
                         f"You need to provide a user along with the user upload "
                         f"queryset for interface {ci}"
                     )
-                upload_session = RawImageUploadSession.objects.create(
-                    creator=user
-                )
+                upload_session = RawImageUploadSession(creator=user)
+                upload_session.full_clean()
```
jmsmkn: Where will the validation errors be handled?

app/grandchallenge/cases/models.py
koopmant: Right, forgot about that. I added a data migration.

app/grandchallenge/components/models.py
koopmant: I somehow assumed they did not need handling. I see we only catch `CIVNotEditableException` elsewhere, but validation errors need to be reported with the relevant interface, so we need to handle them here. 

We also raise a RuntimeError if user is not supplied. This basically should never happen, but it is not handled in the case it does. I suppose I need to fix that too, right?

jmsmkn: If it is unhandled it should be raised through to Sentry, which I think is what we want?

koopmant: Ah right, and the user gets a server error in that case right?

jmsmkn: They do, which is fine - it is unhandled code that _should_ never happen.

koopmant: Then this should be complete.

jmsmkn (approved):
LGTM!

PR #4366, opened 2025-10-20
Adds hmac signing of the invocation result file

jmsmkn:
See https://github.com/DIAGNijmegen/rse-roadmap/issues/443

### DIFF
--- a/app/grandchallenge/components/models.py
+++ b/app/grandchallenge/components/models.py
@@ -1633,7 +1633,6 @@ class ComponentJob(FieldChangeMixin, UUIDModel):
         max_length=32,
         default=secrets.token_bytes,
         unique=True,
-        blank=False,
         editable=False,
         help_text="The key used to sign the inference result file",
     )


app/grandchallenge/components/models.py
```diff
@@ -1628,6 +1629,14 @@ class ComponentJob(FieldChangeMixin, UUIDModel):
             "/input/foo/bar/<relative_path>"
         ),
     )
+    signing_key = models.BinaryField(
+        max_length=32,
+        default=secrets.token_bytes,
+        unique=True,
+        blank=False,
+        editable=False,
```
koopmant: no need to provide `blank` if it's not editable

app/grandchallenge/components/models.py
```diff
@@ -1628,6 +1629,14 @@ class ComponentJob(FieldChangeMixin, UUIDModel):
             "/input/foo/bar/<relative_path>"
         ),
     )
+    signing_key = models.BinaryField(
+        max_length=32,
+        default=secrets.token_bytes,
+        unique=True,
+        blank=False,
```
jmsmkn:
```suggestion
```

PR #4372, opened 2025-10-22
Add exec and invocation durations

jmsmkn:
Adds two new fields:

- `exec_duration`
  - The duration of the execution, **if measured**. Excludes data validation, container pulling, model downloading, data downloading and data uploading times. Includes model loading time, input data loading time, processing time, output data writing time and **any delays from shared hardware issues**.
- `invoke_duration`
  - The duration of the execution, **if measured**. Excludes data validation, container pulling, model downloading, data downloading and data uploading times. **Potentially excludes model loading time, depending on the users implementation**. Includes input data loading time, processing time, output data writing time and **any delays from shared hardware issues**.
  - This will not be set for now and is for future use, it is there for challenge admins to get used to the fact that there are two fields.

One, both or neither will be set. Neither will be set for existing jobs.

Setting and serializing these values will follow in a future PR after all of the container images are upgraded.

Note: There is also `billing_duration` which includes everything, but we might have to change the implementation of that if we go to batch jobs so this will not be serialized for now.

See https://github.com/DIAGNijmegen/rse-roadmap/issues/441

### DIFF
--- a/app/grandchallenge/components/backends/amazon_sagemaker_training.py
+++ b/app/grandchallenge/components/backends/amazon_sagemaker_training.py
@@ -105,7 +105,7 @@ class AmazonSageMakerTrainingExecutor(AmazonSageMakerBaseExecutor):
             },
             StoppingCondition={
                 # https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_StoppingCondition.html
-                "MaxRuntimeInSeconds": self._time_limit,
+                "MaxRuntimeInSeconds": int(self._time_limit.total_seconds()),
             },
             Environment={
                 **self.invocation_environment,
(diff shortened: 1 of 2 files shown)


app/grandchallenge/components/backends/base.py
```diff
@@ -285,7 +307,7 @@ def __init__(
         self._job_id = job_id
         self._exec_image_repo_tag = exec_image_repo_tag
         self._memory_limit = memory_limit
-        self._time_limit = time_limit
+        self._time_limit = timedelta(seconds=time_limit)
```
koopmant, line 310: I see this is still used as an int in https://github.com/comic/grand-challenge.org/blob/5f1151639c0409993b56fb0fec9f97a5ae80369d/app/grandchallenge/components/backends/amazon_sagemaker_training.py#L106-L109

jmsmkn: Good catch!

jmsmkn: There was a test that should have caught that, but it was broken in #3904. Fixed now.

PR #4373, opened 2025-10-22
Avoid error when creating new item with empty value for a linked socket

### LINKED ISSUE #4371: CIV data with None value raises `NotImplementedError` for image kind sockets
Since https://github.com/comic/grand-challenge.org/commit/dd72c59c46006ba53dc4846b3f66031cceeb17d6 (from #4333)

Related comment: https://github.com/comic/grand-challenge.org/pull/4333#discussion_r2429104276

koopmant:
Close #4371

Creating new items (e.g. display sets or archive items) with an undefined value for one or more sockets would raise an error since #4333. This is a common use case and is allowed. 

Also add `__repr__` to `CIVData` for easier debugging.

### DIFF
--- a/app/grandchallenge/components/models.py
+++ b/app/grandchallenge/components/models.py
@@ -2263,6 +2263,9 @@ class CIVData:
         return self._dicom_upload_with_name
 
     def __repr__(self):
+        return f"CIVData(interface_slug={self._interface_slug!r}, value={self._initial_value!r})"
+
+    def __str__(self):
         return f"CIVData: {self.__dict__}"
 
     def __init__(self, *, interface_slug, value):


app/grandchallenge/components/models.py
```diff
@@ -2262,6 +2262,9 @@ def file_civ(self):
     def dicom_upload_with_name(self):
         return self._dicom_upload_with_name
 
+    def __repr__(self):
+        return f"CIVData: {self.__dict__}"
```
jmsmkn, line 2269: `__repr__` is for serialisation of the object to a string that can be used by python interpreters to construct a new instance of the object, that won't work with this value. `__str__` is for human readable representations, which would be appropriate for this value.

koopmant: You're right. I skipped half of the text when reading the documentation...

I need `__repr__` to print the representation for values in the stack trace. I checked with `__str__` and that will not be used so you still end up with something like `<grandchallenge.components.models.CIVData object at 0x...>`. 

PS
`__repr__` may also return a description in angled brackets that's useful for debugging, but only if it is not possible to return a representation that can be used to construct a new instance. [docs.python.org/...](https://docs.python.org/3/reference/datamodel.html#object.__repr__)

jmsmkn: Good to know!

PR #4378, opened 2025-10-23
Remove deprecated fields

koopmant:


app/tests/challenges_tests/test_forms.py
```diff
@@ -15,17 +14,45 @@
 )
 
 
+@pytest.mark.parametrize(
+    "required_field",
```
jmsmkn: Isn't this going to cause a slowdown of the tests as you have to repeat the setup each time? Seems like this would be easy to check in a single test.

koopmant: Thanks, I've changed this in #4264 now.

app/grandchallenge/archives/templates/archives/archive_cases_list.html
koopmant, line 1: I found these old templates which are no longer used anywhere. Their view has been removed in #3210

app/grandchallenge/archives/templates/archives/archive_cases_row.html
koopmant, line 1: I found these old templates which are no longer used anywhere. Their view has been removed in https://github.com/DIAGNijmegen/rse-grand-challenge/pull/3210

app/grandchallenge/cases/models.py
```diff
@@ -516,37 +515,6 @@ class Image(UUIDModel):
     color_space = models.CharField(
         max_length=5, blank=True, choices=COLOR_SPACES
     )
-    # Deprecated fields, see https://github.com/DIAGNijmegen/rse-roadmap/issues/446
```
koopmant, line 519: The serializer provides empty values for these. I think we want to keep that in place.

app/grandchallenge/archives/templates/archives/archive_cases_row.html
koopmant: I found this one because it references the deprecated field on `Image`.

Issue #4312, opened 2025-09-25
FlexibleImageWidget allows submitting without a selected image


amickan:
**Describe the bug**
When a user selects "Select an existing image", but then does not actually select an image, hitting the `Save` button on the form (algorithm job form, for example) result in the form being submitted and a server error being thrown to the user. What should happen instead is that the form gets returned with a validation error telling the user that they have to select an image. 

**To Reproduce**
1. Go to the "Try out algorithm" page of an algorithm with an image socket
2. Click on 'Select existing image'
3. Don't select an image and just hit the 'Save' button 
4. See ServerError (e.g.: https://grand-challenge.sentry.io/issues/6871969187/?query=is%3Aunresolved&referrer=issue-stream)

**Expected behavior**
The form should fail validation and the user should be prompted to select an image.

PR #4381, opened 2025-10-24
Avoid server error in FlexibleImageWidget validation

koopmant:
Fix error in `FlexibleImageWidget.decompress()` due to unexpected value. This now properly raises a "field required" validation error on the field. 

The possible data returned from the possible widgets of `FlexibleImageWidget` was not completely tested. This adds a test where we extract the form data from the response of the flexible image widget for each of the possible options the user can select.

Closes https://github.com/comic/grand-challenge.org/issues/4312

### DIFF
--- a/app/grandchallenge/cases/widgets.py
+++ b/app/grandchallenge/cases/widgets.py
@@ -63,15 +63,17 @@ class FlexibleImageWidget(MultiWidget):
         )
         super().__init__(widgets, attrs)
 
-    def decompress(self, value):
+    def decompress(self, value):  # noqa: C901
         if not value:
             return [None, None]
 
         if isinstance(value, (list, tuple)):
             if len(value) == 1:
                 item = value[0]
-                if not item:
+                if item == "":
                     return [None, None]
+                if not item:
+                    raise RuntimeError("Unexpected value")
                 if item in ImageWidgetChoices.names:
                     return [None, None]
                 if Image.objects.filter(pk=item).exists():


jmsmkn:
Does this solve the `“%(value)s” is not a valid UUID.` error?

> Closes to #4312

What does this mean?

app/grandchallenge/cases/widgets.py
```diff
@@ -70,6 +70,8 @@ def decompress(self, value):
         if isinstance(value, (list, tuple)):
             if len(value) == 1:
                 item = value[0]
+                if not item:
```
jmsmkn: Is this strict enough? This will catch anything that isn't truthy.

app/grandchallenge/cases/widgets.py
jmsmkn, line 82: I don't think this will ever be hit now.

Shouldn't this be switched to an `elif` statement and an error raised if not matched?

app/grandchallenge/cases/widgets.py
koopmant: I could turn it into `if item == ""`. I don't expect anything else here, so I suppose we could raise an error in that case?

app/grandchallenge/cases/widgets.py
koopmant: > I don't think this will ever be hit now.

This should still be reached if the item is a pk string for a user upload. I admit I wasn't sure if we can already filter for `UserUpload` objects at this point, so I simply did not touch this part.

app/grandchallenge/cases/widgets.py
jmsmkn: It's probably worth being strict here.

app/grandchallenge/cases/widgets.py
jmsmkn: I have no idea I'm afraid.

koopmant: I've been looking into this and we can add `if UserUpload.objects.filter(pk__in=value).exists():` as an extra check. However, when we raise errors in decompress, the user is presented with strange behaviour, that, as @pkcakeout mentioned, feels very buggy. 

What I believe happens is that the error is raised in during form validation, but outside the section where validation errors are caught. This interrupts the post request, so for the user nothing happens. 

This is a problem with our FlexibleImageWidget implementation which I cannot easily fix right now and needs to be picked up https://github.com/DIAGNijmegen/rse-roadmap/issues/435. 

I suggest to just fix the case where the value is an empty string and leave the rest as is for now.

app/grandchallenge/cases/widgets.py
```diff
@@ -63,13 +63,17 @@ def __init__(
...
         if not value:
             return [None, None]
 
         if isinstance(value, (list, tuple)):
             if len(value) == 1:
                 item = value[0]
+                if item == "":
+                    return [None, None]
+                if not item:
+                    raise RuntimeError("Unexpected value")
```
koopmant, line 76: I can also leave this out, it will then raise the same validation error ("value" is not a valid UUID)

app/grandchallenge/cases/widgets.py
jmsmkn: I agree, it is not the responsibility of the widget to raise validation errors here so no surprise the they are unhandled. Its responsibility is to adapt HTML form values to Python values.

PR #4383, opened 2025-10-27
Refactor GitHub repos list to use async requests

jmsmkn:
Reduce the time when fetching multiple pages by performing the requests asynchronously, using up to `ASYNC_CONCURRENCY` concurrent requests.

Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/664

### DIFF
--- a/app/grandchallenge/github/views.py
+++ b/app/grandchallenge/github/views.py
@@ -120,20 +120,18 @@ class GitHubInstallationRequiredMixin:
         return f"{settings.GITHUB_APP_INSTALL_URL}?state={self.github_state}"
 
     @cached_property
-    def github_request_kwargs(self):
+    def github_request_headers(self):
         return {
-            "headers": {
-                "Accept": "application/vnd.github+json",
-                "Authorization": f"token {self.github_user_token.access_token}",
-            },
-            "timeout": 10,
+            "Accept": "application/vnd.github+json",
+            "Authorization": f"token {self.github_user_token.access_token}",
         }
 
     @cached_property
     def installations(self):
-        response = requests.get(
+        response = httpx.get(
             "https://api.github.com/user/installations",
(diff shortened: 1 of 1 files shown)


app/grandchallenge/github/views.py
```diff
@@ -165,7 +167,22 @@ class RepositoriesList(
...
+    async def _get_page_github_repos(
+        self, *, installation_id, page, per_page, semaphore, httpx_client
+    ):
+        async with semaphore:
+            response = await httpx_client.get(
+                f"https://api.github.com/user/installations/{installation_id}/repositories",
+                params={"per_page": per_page, "page": page},
+                headers=self.github_request_headers,
+            )
+            response.raise_for_status()
```
koopmant, line 179: Why not do this by default for all requests through the httpx_client?

jmsmkn: We do, right?

koopmant: Sorry, I see we do it for each request. 
We could add `raise_for_status=True` in the contructor call in that case.

jmsmkn: I don't think that is implemented for httpx.

koopmant: You're right, I'm wrong again. (I am not sharp today...)

jmsmkn: Yes, don't know where that comes from:

> TypeError: AsyncClient.__init__() got an unexpected keyword argument 'raise_for_status'

> TypeError: AsyncClient.get() got an unexpected keyword argument 'raise_for_status'

koopmant: I was reading the documentation from another library. Almost feels like I'm an hallucinating LLM at this point.

PR #4386, opened 2025-10-28
Update image permission assignment through algorithm jobs

jmsmkn:
Avoids the use of the slow `update_viewer_groups_permissions` method by assigning the `view_image` permissions directly. The downside of this is that the permissions will need to be kept in sync in both `signals.py` and `_get_expected_job_viewer_groups` but we have good test coverage for this.

Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/663

### DIFF
--- a/app/grandchallenge/algorithms/signals.py
+++ b/app/grandchallenge/algorithms/signals.py
@@ -81,7 +81,7 @@ def _get_images_for_jobs(*, jobs):
 
 
 @receiver(m2m_changed, sender=Job.viewer_groups.through)
-def update_view_image_permissions_on_viewer_groups_change(  # noqa:C901
+def update_permissions_on_viewer_groups_change(  # noqa:C901
     *_, instance, action, reverse, model, pk_set, **__
 ):
     if action not in ["post_add", "post_remove", "pre_clear"]:


app/grandchallenge/algorithms/signals.py
```diff
@@ -21,56 +21,65 @@ def update_input_image_permissions(
...
+    if action == "post_add":
+        for job in jobs:
+            for group in job.viewer_groups.all():
+                assign_perm("view_image", group, images)
+
+    elif action in {"post_remove", "pre_clear"}:
+        for image in images:
+            # We cannot remove image permissions directly as the groups
+            # may have permissions through another object
+            image.update_viewer_groups_permissions(exclude_jobs=jobs)
```
koopmant: The jobs should not be excluded on `post_remove`, right?

app/grandchallenge/algorithms/signals.py
```diff
@@ -21,56 +21,65 @@ def update_input_image_permissions(
...
-        )
+
+def _get_images_for_jobs(*, jobs):
+    input_civs = Image.objects.filter(
+        componentinterfacevalue__algorithms_jobs_as_input__in=jobs
+    )
+    output_civs = Image.objects.filter(
+        componentinterfacevalue__algorithms_jobs_as_output__in=jobs
+    )
+    return input_civs.union(output_civs)
```
koopmant:
```suggestion
    input_images = Image.objects.filter(
        componentinterfacevalue__algorithms_jobs_as_input__in=jobs
    )
    output_images = Image.objects.filter(
        componentinterfacevalue__algorithms_jobs_as_output__in=jobs
    )
    return input_images.union(output_images)
```

app/grandchallenge/algorithms/signals.py
```diff
@@ -21,56 +21,65 @@ def update_input_image_permissions(
...
 
 
 @receiver(m2m_changed, sender=Job.viewer_groups.through)
-def update_group_permissions(
+def update_view_image_permissions_on_viewer_groups_change(  # noqa:C901
```
koopmant: This also updates the view job permissions. So maybe a different name?

app/grandchallenge/algorithms/signals.py
```diff
@@ -90,23 +99,32 @@ def update_group_permissions(
...
+    elif action in {"post_remove", "pre_clear"}:
         for group in groups:
-            operation("view_job", group, job)
+            for job in jobs:
+                remove_perm("view_job", group, job)
+
+        for image in images:
+            # We cannot remove image permissions directly as the groups
+            # may have permissions through another object
+            image.update_viewer_groups_permissions(exclude_jobs=jobs)
```
koopmant: Same as above, I think the jobs should only be excluded for `pre_clear`. Although it won't really change the outcome, I suppose.

app/grandchallenge/algorithms/signals.py
koopmant: addressed

PR #4387, opened 2025-10-29
Remove on commit signals for image permission updates

jmsmkn:
See DIAGNijmegen/rse-grand-challenge-admin#663

### DIFF
--- a/app/grandchallenge/reader_studies/signals.py
+++ b/app/grandchallenge/reader_studies/signals.py
@@ -35,10 +35,9 @@ def update_view_image_permissions_on_display_set_values_change(
         if pk_set is None:
             # When using a _clear action, pk_set is None
             # https://docs.djangoproject.com/en/2.2/ref/signals/#m2m-changed
-            images = [
-                civ.image
-                for civ in instance.values.filter(image__isnull=False)
-            ]
+            images = Image.objects.filter(
+                componentinterfacevalue__display_sets=instance
+            )
         else:
             images = Image.objects.filter(
                 componentinterfacevalue__pk__in=pk_set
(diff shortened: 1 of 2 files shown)


app/grandchallenge/reader_studies/signals.py
```diff
@@ -5,24 +5,33 @@
...
+            display_sets = instance.display_sets.all()
+        else:
+            display_sets = model.objects.filter(pk__in=pk_set)
+
     else:
+        display_sets = [instance]
+
         if pk_set is None:
             # When using a _clear action, pk_set is None
             # https://docs.djangoproject.com/en/2.2/ref/signals/#m2m-changed
```
koopmant, line 37: suggestion for line below this:

```
images = Image.objects.filter(
        componentinterfacevalue__display_sets=instance
    ).distinct()
```

app/grandchallenge/cases/models.py
```diff
@@ -683,10 +686,49 @@ def update_viewer_groups_permissions(  # noqa: C901
             when a many to many relationship is being cleared to remove this
             image from the results image set, and is used when the pre_clear
             signal is sent.
+        exclude_archive_items
+            As above but for Archive Items
```
koopmant: I found I found "exclude these from being considered" a bit confusing, as it sounds like they are excluded from the update but in fact they are excluded from the expected group. 
The code is pretty clear. I think I would remove this comment.

app/grandchallenge/reader_studies/signals.py
jmsmkn: Completely missed that that wasn't a query!

PR #4388, opened 2025-10-29
On commit cleanup

jmsmkn:
Ensures that:

- only celery tasks are used in `on_commit`
- the function passed to `on_commit` is a real method and not a lambda
- `apply_async` is never called directly

See https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/663

### DIFF
--- a/app/tests/verification_tests/test_forms.py
+++ b/app/tests/verification_tests/test_forms.py
@@ -7,6 +7,7 @@ from grandchallenge.verifications.forms import (
     VerificationForm,
 )
 from tests.factories import UserFactory
+from tests.utils import get_view_for_user
 from tests.verification_tests.factories import VerificationFactory
 
 
@@ -172,7 +173,9 @@ class TestConfirmEmailForm:
 
         assert form.is_valid()
 
-    def test_user_can_not_verify_other_token(self, settings):
+    def test_user_can_not_verify_other_token(
+        self, settings, client, django_capture_on_commit_callbacks
+    ):
         settings.task_eager_propagates = (True,)
         settings.task_always_eager = (True,)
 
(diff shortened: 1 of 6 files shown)


app/tests/verification_tests/test_forms.py
```diff
@@ -172,7 +173,9 @@ def test_user_can_verify(self):
 
         assert form.is_valid()
 
-    def test_user_can_not_verify_other_token(self, settings):
+    def test_user_can_not_verify_other_token(
```
koopmant, line 176: Propose to move the second part of this test to `test_views` if we move handling to the `form_invalid` method.

app/grandchallenge/verifications/forms.py
```diff
@@ -100,9 +101,11 @@ def clean_token(self):
...
 
         if not hasattr(self.user, "verification"):
-            deactivate_user.signature(
-                kwargs={"user_pk": self.user.pk}
-            ).apply_async()
+            on_commit(
+                deactivate_user.signature(
+                    kwargs={"user_pk": self.user.pk}
+                ).apply_async
+            )
```
koopmant, line 108: It's a bit strange that this is done in the form's clean method. Wouldn't it be better if this is handled in `ConfirmEmailView.form_invalid()`?

jmsmkn: The form can be invalid for many reasons and you only want to do it in this case.

koopmant: Yes, I understand. What I mean is, I feel like this should not be the responsibility of the form. I noticed it in the test, because aside from asserting that the form is invalid we also assert that there is the side-effect of the user being deactivated.

jmsmkn: Yes, the user submitted someone else's token, so they should be deactivated. The form is where we find they do that and act on it. The form does not do the deactivation, only creates the task to do it.

jmsmkn: This is anyway tangential to this PR. This PR is only about updating how the tasks are called for consistency across the codebase. If you want to change the logic then do it in a different PR.

koopmant: > The form does not do the deactivation, only creates the task to do it.

Fair enough

koopmant (approved):
LGTM!

PR #4392, opened 2025-11-05
Remove 'Accept-Encoding' header from metadata request

HarmvZ:
See explanation here: https://github.com/DIAGNijmegen/rse-roadmap/issues/438#issuecomment-3490358069

### DIFF
--- a/app/grandchallenge/components/backends/base.py
+++ b/app/grandchallenge/components/backends/base.py
@@ -97,7 +97,7 @@ def serialize_aws_request(request, *, unsigned_headers=None):
     External clients use this so the kwargs should not be changed.
     """
     headers = dict(request.headers.items())
-    
+
     if unsigned_headers:
         if intersection := headers.keys() & unsigned_headers.keys():
             raise ValueError(


app/grandchallenge/cases/views.py
```diff
@@ -181,7 +181,6 @@ def dicom(self, request, pk=None):
             url=f"{image_set_url}/getImageSetMetadata",
             data=json.dumps({"versionId": "1"}),
             headers={
-                "Accept-Encoding": "gzip",
```
jmsmkn, line 184: We don't just want to remove this as the client _should_ still send it. Can we add `unsigned_headers` to `serialize_aws_request` and then combine them into `headers`?

We would need to ensure that the request would still work when sending the additional unsigned header from python/c++.

Please also add a comment about _why_ Accept-Encoding cannot be signed.

HarmvZ: Removing it here just makes sure it is not used to sign the request so that it can be modified. Any additional headers are still allowed in the request. It's not required to explicitly define those here, right?

jmsmkn: The client should be sending this information and this is where we can centrally manage that they do. Then everything is up to date with the clients doing `httpx_client.stream(**request_kwargs)`.

pkcakeout: Quick note: I also need to adapt the HealthImagingLoader, because there I only copy the signed headers and sent those. I think this ends up being an absolutely bare-bones query with _only_ the flags from the signed header in the request, so I need to fix this by adding `Accept-Encoding` manually.

jmsmkn: If Harm implements my suggestions then you should not need to make any changes Paul as even the unsigned headers will be serialized. If you are parsing the signature to work out what is signed and unsigned that should be removed.

pkcakeout: Ah alright! That will work then. I am doing the dumb thing: Just taking the `data`, put it in as data (binary escaped copy) and then put in all the headers from the `header` field and call it quits. So ok, that should make things easier! Thanks for the correction!

jmsmkn: Yes, that was the intention, just in case the contract for the request changes and suddenly a new header or changed value is required, can manage it all in one place.

app/grandchallenge/components/backends/base.py
```diff
@@ -89,18 +89,21 @@ def duration_to_millicents(*, duration, usd_cents_per_hour):
...
+def serialize_aws_request(request, *, unsigned_headers=None):
     """
     The kwargs that will be passed to httpx.stream or httpx.request
     to generate a response from an AWSRequest instance.
 
     External clients use this so the kwargs should not be changed.
     """
+    headers = dict(request.headers.items())
+    if unsigned_headers:
+        headers.update(unsigned_headers)
```
jmsmkn:
```suggestion
    headers = dict(request.headers.items())
    
    if unsigned_headers:
        if intersection := headers.keys() & unsigned_headers.keys():
            raise ValueError(
                f"Signed and unsigned headers must be disjoint, {intersection=}"
            )
        else:
            headers.update(unsigned_headers)

```

PR #4393, opened 2025-11-05
Only get the current_value for FlexibleImage and FlexibleFile

koopmant:
Avoid RuntimeError due to taking the "current value" for dicom image kind. The FlexibleImageField is not used for dicom image kind, hence a current value is not needed here.


jmsmkn (approved):
I cannot really follow what is going on now but I trust you on this one!

PR #4394, opened 2025-11-06
Add tests for adding values to value sets

koopmant:

### DIFF
--- a/app/grandchallenge/algorithms/models.py
+++ b/app/grandchallenge/algorithms/models.py
@@ -1263,7 +1263,7 @@ class Job(CIVForObjectMixin, ComponentJob):
         )
 
         linked_task = execute_algorithm_job_for_inputs.signature(
-            kwargs={"job_pk": self.pk}, immutable=True
+            kwargs={"job_pk": str(self.pk)}, immutable=True
         )
 
         return super().validate_civ_data_objects_and_execute_linked_task(
--- a/app/grandchallenge/components/models.py
+++ b/app/grandchallenge/components/models.py
@@ -2654,9 +2654,9 @@ class CIVForObjectMixin:
                 kwargs={
                     "app_label": self._meta.app_label,
                     "model_name": self._meta.model_name,
-                    "object_pk": self.pk,
+                    "object_pk": str(self.pk),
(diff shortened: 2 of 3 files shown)


app/tests/components_tests/test_models.py
```diff
@@ -1876,3 +1885,237 @@ def test_signing_key_properties():
...
+)
+@pytest.mark.django_db
+def test_add_and_remove_civ(object_factory, kwargs, field_name):
+    obj = object_factory(**kwargs)
+    civ = ComponentInterfaceValueFactory()
+    n_values = getattr(obj, field_name).count()
+
+    obj.add_civ(civ=civ)
+
+    assert getattr(obj, field_name).count() == n_values + 1
```
chrisvanrun, line 1915: Why do a diff comparison here? Would a simple check on `0` and `1` not be sufficient?

koopmant: One of the factories (`AlgorithmJobFactory`) sets a socket value automatically post generation. This was the easiest solution.

chrisvanrun: Ah =/

app/grandchallenge/algorithms/models.py
```diff
@@ -1263,7 +1263,7 @@ def validate_civ_data_objects_and_execute_linked_task(
         )
 
         linked_task = execute_algorithm_job_for_inputs.signature(
-            kwargs={"job_pk": self.pk}, immutable=True
+            kwargs={"job_pk": str(self.pk)}, immutable=True
```
jmsmkn, line 1266: What is the reason for casting these to strings now? I think that this is old code that has been running for a while in prod.

koopmant: I'm saving this linked task in a JSON field later and it is not valid json if I don't make it a string. Do you think I should fix it at the point where I'm saving it to the field?

jmsmkn: No it's okay, you could do it with a custom JSON Encoder but casting to strings is also fine, we do it elsewhere, just forgot the reason.

PR #4395, opened 2025-11-07
Add slug to interface modal

chrisvanrun:
In the interface overview, we don't mention the slug of sockets anywhere. It can be found via the overview of sockets, but this PR makes it directly visible via the interface overview for both algorithms and challenge submissions.

Handy, if one needs to know the slugs of the input sockets to correctly call the API.

## Before
Screenshot 2025-11-07 at 11 03 39

## After
Screenshot 2025-11-07 at 11 12 37


koopmant (approved):
Nice!

PR #4398, opened 2025-11-10
Exclude the `deidentification.done` file from the dicom import job input files

koopmant:

### DIFF
--- a/app/grandchallenge/cases/models.py
+++ b/app/grandchallenge/cases/models.py
@@ -1190,12 +1190,12 @@ class DICOMImageSetUpload(UUIDModel):
         return f"inputs/{self.pk}"
 
     @property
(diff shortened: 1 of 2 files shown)


jmsmkn (dismissed):

app/grandchallenge/cases/models.py
```diff
@@ -1189,9 +1189,13 @@ def _import_job_name(self):
     def _input_prefix(self):
         return f"inputs/{self.pk}"
 
+    @property
+    def _input_files_location(self):
```
jmsmkn: As the test are failing a small suggestion - use `prefix` rather than `location` here as that is meaningful for S3.

jmsmkn (approved):
LGTM!

PR #4401, opened 2025-11-11
Quick fix for errors on running celery tasks without celeryd

chrisvanrun:
The uv-based integration tests for gcapi run without workers. This entails that the `task_prerun` and `task_postrun` listeners fire, but not the `celeryd_after_setup`. 

As a result, this throws an error several times during setup (but isolated so nothing breaks):
> AttributeError: 'Celery' object has no attribute 'is_solo_worker'

Instead of re-adding the worker, I thought it's OK to instead have the prerun/postrun to take into account the attr hadn't been set.

### DIFF
--- a/app/config/celery.py
+++ b/app/config/celery.py
@@ -68,8 +68,8 @@ def get_scale_in_protection_url():
 @task_prerun.connect()
 def set_ecs_scale_in_protection(*_, task, **__):
     if (
-        celery_app.is_solo_worker
-        and settings.ECS_ENABLE_CELERY_SCALE_IN_PROTECTION
+        settings.ECS_ENABLE_CELERY_SCALE_IN_PROTECTION
+        and celery_app.is_solo_worker
     ):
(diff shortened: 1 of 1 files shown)


app/config/celery.py
```diff
@@ -68,7 +68,7 @@ def get_scale_in_protection_url():
 @task_prerun.connect()
 def set_ecs_scale_in_protection(*_, task, **__):
     if (
-        celery_app.is_solo_worker
+        getattr(celery_app, "is_solo_worker", False)
         and settings.ECS_ENABLE_CELERY_SCALE_IN_PROTECTION
     ):
```
jmsmkn, line 73: `is_solo_worker` really should be set in a production environment. Could flipping them be cleaner, e.g.

```python3
    if (
        settings.ECS_ENABLE_CELERY_SCALE_IN_PROTECTION
        and celery_app.is_solo_worker
    ):
```

chrisvanrun: Ah yes, that would a bit of a safer option.

PR #4406, opened 2025-11-11
Ignore GET parameters for paginated tables

jmsmkn:
Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/620

### DIFF
--- a/app/tests/algorithms_tests/test_views.py
+++ b/app/tests/algorithms_tests/test_views.py
@@ -272,7 +272,7 @@ def test_algorithm_jobs_list_view(client):
         reverse_kwargs={"slug": slugify(alg.slug)},
         client=client,
         user=editor,
-        method=client.get,
+        method=client.post,
         follow=True,
         data={
             "length": 10,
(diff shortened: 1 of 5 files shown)


app/grandchallenge/datatables/views.py
```diff
@@ -44,25 +44,32 @@ def render_rows(self, *, object_list, request):
...
     def draw_response(self, *, request):
-        # TODO: I think we only have POST data now
-        form_data = request.POST or request.GET
-        start = int(form_data.get("start", 0))
-        page_size = int(form_data.get("length"))
-        search = form_data.get("search[value]")
+        start = int(request.POST.get("start", 0))
+        page_size = int(request.POST.get("length"))
+        search = request.POST.get("search[value]")
+        draw = int(request.POST.get("draw"))
```
koopmant, line 61: I see this hasn't changed, but for `page_size` and `draw` the `int()` will cause a TypeError if length and draw are not in the data. So I think we should either use `get()` with a valid default or not use `get()` and let a `KeyError` be raised if they are missing.

koopmant (approved):
Looks good, just a suggestion for change.

app/grandchallenge/datatables/views.py
jmsmkn: Yes I thought about that but figured I would leave it unchanged, either case results in a 5xx error for the user.

Issue #2412, opened 2022-04-20
Replace Django stdimage with Django Pictures


jmsmkn:
See https://github.com/codingjoe/django-stdimage/issues/290#issuecomment-1103824112

Now watching Django Pictures for the v1 release.

jmsmkn:
See the migration instructions here https://github.com/codingjoe/django-stdimage#migration-instructions

PR #4409, opened 2025-11-12
Upgrade to Django 5.2

jmsmkn:
Any code changes are to fix Deprecation Warnings. There are some new ones for Django 6.0 which can be tackled later.


jmsmkn:
We really have to do https://github.com/comic/grand-challenge.org/issues/2412 before this

PR #4410, opened 2025-11-12
Enable user uploads through serializer for image kinds

koopmant:
Related to https://github.com/DIAGNijmegen/rse-roadmap/issues/433

### DIFF
--- a/app/tests/reader_studies_tests/test_api.py
+++ b/app/tests/reader_studies_tests/test_api.py
@@ -1881,7 +1881,7 @@ def test_display_set_add_and_edit(
 
 
 @pytest.mark.django_db
-def test_display_set_add_and_edit_with_image_user_uploads(
+def test_display_set_update_with_image_user_uploads(
     client, settings, django_capture_on_commit_callbacks, mocker
 ):
     settings.task_eager_propagates = (True,)


app/grandchallenge/components/serializers.py
```diff
@@ -43,7 +43,13 @@ def convert_deserialized_civ_data(*, deserialized_civ_data):
...
             )
         elif len(keys_not_none) == 1:
-            value = civ[keys_not_none.pop()]
+            key = keys_not_none.pop()
+            value = civ[key]
+            if key == "user_uploads":
+                # Convert to queryset for type switch in CIVData
+                value = UserUpload.objects.filter(
+                    pk__in=[upload.pk for upload in value]
+                )
```
koopmant, line 52: This is the easiest way to do it at the moment

I don't really like the type switch in CIVData to be honest. It allows us to accept data from all possible widgets on a CIVForm, so it makes sense. But for data we get from the serializer we already know for which attribute the data is meant for. I think that should also be possible for the form data, although I'm not entirely sure. I could look into it in https://github.com/DIAGNijmegen/rse-roadmap/issues/435. 

With the added comment I think this should be fine for now?

jmsmkn: Yes, there's no good way to do this right now.

jmsmkn (approved):
LGTM!

PR #4411, opened 2025-11-13
Migrate from stdimage to django pictures

jmsmkn:
Following the migration guide at https://github.com/codingjoe/django-stdimage?tab=readme-ov-file#migration-instructions

This will need to be tested in staging to ensure that the micro short worker is taken down and has the correct permissions for the images.

A follow-up PR will need to be made to remove stdimage.

See #2412

### DIFF
--- a/app/tests/cases_tests/test_api.py
+++ b/app/tests/cases_tests/test_api.py
@@ -57,11 +57,11 @@ def test_image_api_fields(client):
 
     assert response.json() == {
         "api_url": image.api_url,
-        "color_space": "RGB",
-        "depth": None,
+        "color_space": image.color_space,
+        "depth": image.depth,
         "dicom_image_set": None,
-        "eye_choice": "OD",
-        "field_of_view": "F1M",
+        "eye_choice": image.eye_choice,
+        "field_of_view": image.field_of_view,
         "files": [
             {
                 "file": file.file.url,
@@ -71,10 +71,8 @@ def test_image_api_fields(client):
             }
             for file in image.files.all()
         ],
-        "height": 4,
-        "modality": {
-            "modality": "CF",
-        },
+        "height": image.height,
+        "modality": {"modality": image.modality.modality},
         "name": image.name,
         "patient_age": image.patient_age,
         "patient_birth_date": image.patient_birth_date.strftime("%Y-%m-%d"),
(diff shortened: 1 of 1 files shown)


app/grandchallenge/algorithms/migrations/0085_algorithm_social_image_height_and_more.py
```diff
@@ -0,0 +1,91 @@
...
+            model_name="algorithm",
+            name="social_image_height",
+            field=models.PositiveSmallIntegerField(editable=False, null=True),
+        ),
+        migrations.AddField(
+            model_name="algorithm",
+            name="social_image_width",
+            field=models.PositiveSmallIntegerField(editable=False, null=True),
+        ),
+        migrations.RunPython(
```
jmsmkn, line 42: We usually do these separately, but I am trying to see if this works altogether in production now. Will need the staging environment for this.

app/grandchallenge/algorithms/serializers.py
```diff
@@ -58,7 +58,6 @@ class Meta:
 
 class AlgorithmSerializer(serializers.ModelSerializer):
     average_duration = SerializerMethodField()
-    logo = URLField(source="logo.x20.url", read_only=True)
```
koopmant, line 61: These fields are removed because they aren't used?

app/config/settings.py
```diff
@@ -652,33 +653,26 @@ def get_private_ip():
...
-    "x15": (1665, None, False),
-    "x10": (1110, None, False),
+PICTURES = {
+    "BREAKPOINTS": {
+        # should be identical to the ones used in your CSS library.
+        "xs": 576,
+        "s": 768,
+        "m": 992,
+        "l": 1200,
+        "xl": 1400,
```
koopmant, line 667:
```suggestion
```

I think this is for bootstrap 5

app/config/settings.py
```diff
@@ -652,33 +653,26 @@ def get_private_ip():
...
+    },
+    "GRID_COLUMNS": 12,
+    "CONTAINER_WIDTH": 1200,
```
koopmant, line 670: I guess this should be 1140; not sure.

setup.cfg
```diff
@@ -39,6 +39,7 @@ ignore =
     B950
     # TODO SEP21 JM Exception chaining, unsure on the consequences
     B904
+    B907
```
koopmant, line 42: Add a comment?

app/tests/serializer_helpers.py
koopmant, line 1: I don't really understand what has changed here and why we no longer need this.

app/grandchallenge/algorithms/serializers.py
jmsmkn: They're still there and included in the tests. The custom serialiser for the field is just removed.

app/config/settings.py
jmsmkn: It is, but that is in the planning, it is close enough.

app/config/settings.py
jmsmkn: See above.

app/tests/serializer_helpers.py
jmsmkn: Tests are refactored into ones that make more sense, it's no longer used.

setup.cfg
jmsmkn: I disagree with the check, `!r` is for repr, we want the string but also the specific quotes wherever it is used.

koopmant: I understand, I meant should we add a comment here in the file to document that?

app/grandchallenge/algorithms/serializers.py
koopmant: Oh right, of course.

koopmant (approved):
LGTM!

Thanks for the explanations. Whether you add the comment about B907 is up to you.

setup.cfg
jmsmkn: I think it's okay. These things are a question of taste and if future developers have a different opinion on this on it can be re-enabled and the exceptions added back.

jmsmkn: Could you see my question on Teams?

koopmant: Sorry, I missed it.

PR #4422, opened 2025-11-18
Parts must be sorted

jmsmkn:
Fixes InvalidPartOrder: when calling the CompleteMultipartUpload operation: The list of parts was not in ascending order. Parts must be ordered by part number.


chrisvanrun (approved):
Symbolic: LGTM!

PR #4425, opened 2025-11-18
Unstick the navbar

jmsmkn:


jmsmkn:
Unsure, it looked a bit strange without the shadow to differentiate it on pages without breadcrumbs - the login pages for example.

koopmant:
Ah good catch. In that case we could go with `border-bottom` as an alternative. Or leave it as is.

PR #4426, opened 2025-11-18
Add dicom image set upload views

koopmant:
Add views for dicom image set uploads. 

Notifications about the upload now link to the detail view.

List view: 
Screenshot 2025-11-18 at 21 12 34

Notification:
Screenshot 2025-11-18 at 21 12 20

Detail view:
Screenshot 2025-11-18 at 21 11 51

Detail view completed upload with attached image:
Screenshot 2025-11-19 at 13 19 25

New menu item to list view:
Screenshot 2025-11-19 at 13 21 30

### DIFF
--- a/app/grandchallenge/core/templates/grandchallenge/partials/userlinks.html
+++ b/app/grandchallenge/core/templates/grandchallenge/partials/userlinks.html
@@ -90,6 +90,9 @@
             <a class="dropdown-item"
                href="{% url 'cases:raw-image-upload-session-list' %}">
                 Your Uploads</a>
+            <a class="dropdown-item"
+               href="{% url 'cases:dicom-image-set-upload-list' %}">
+                Your DICOM Uploads</a>
             {% if 'emails.view_email' in perms %}
                 <div class="dropdown-divider"></div>
                 <a class="dropdown-item"


koopmant:
The menu item your uploads links to the list view of raw image upload sessions. Should we create a view that combines the two list views?

Screenshot 2025-11-18 at 21 24 14

app/grandchallenge/cases/models.py
```diff
@@ -1514,3 +1517,26 @@ def convert_image_set_to_internal(self, *, image_set_id):
...
+        {"change_dicomimagesetupload", "view_dicomimagesetupload"}
+    )
+
+    content_object = models.ForeignKey(
+        DICOMImageSetUpload, on_delete=models.CASCADE
+    )
+
+
+class DICOMImageSetUploadGroupObjectPermission(GroupObjectPermissionBase):
+    allowed_permissions = frozenset()
```
koopmant, line 1536: Am I correct in thinking we add these models for performance reasons?

app/grandchallenge/cases/models.py
```diff
@@ -1196,6 +1196,9 @@ def save(self, *args, **kwargs):
 
         if adding:
             if self.creator:
+                assign_perm(
+                    f"view_{self._meta.model_name}", self.creator, self
+                )
```
koopmant, line 1201: Is it okay to give only the `view` permission to creators? I don't think `change` would mean anything right now, but it might make sense for creators to have it as well.

app/grandchallenge/cases/templates/cases/dicomimagesetupload_detail.html
```diff
@@ -0,0 +1,95 @@
...
+                    badge-info
+                {% endif %}">
+                    {{ object.get_status_display }}
+            </span>
+        </dd>
+    </dl>
+
+    {% if object.status == object.DICOMImageSetUploadStatusChoices.COMPLETED %}
+
+        <h2>Image</h2>
```
koopmant, line 56: Should I add a list of download links here, for downloading the dicom instances?

app/grandchallenge/cases/models.py
jmsmkn: Yes that is correct. For the best performance the two allowed_permissions sets should be disjoint. If they are not, then we face performance problems - see the image list view or reader study answer view.

app/grandchallenge/cases/models.py
jmsmkn: Only do what is necessary. If we need to add something in the future that is easier than changing it. In general - we do not know our future requirements, so it is better not to guess at them now.

app/grandchallenge/cases/templates/cases/dicomimagesetupload_detail.html
jmsmkn: That's going to be really ugly. We need to think about what the DICOM links will look like for CIVs and can deal with it later. My guess is that it will be a modal with instructions about how to get it with GCAPI, but that is for another PR.

app/grandchallenge/cases/models.py
```diff
@@ -1514,3 +1517,26 @@ def convert_image_set_to_internal(self, *, image_set_id):
...
+
+    def get_absolute_url(self):
+        return reverse(
+            "cases:dicom-image-set-upload-detail", kwargs={"pk": self.pk}
+        )
+
+
+class DICOMImageSetUploadUserObjectPermission(UserObjectPermissionBase):
+    allowed_permissions = frozenset(
+        {"change_dicomimagesetupload", "view_dicomimagesetupload"}
```
jmsmkn: Is anyone being assigned the change permission?

koopmant: No, I can just leave it out then?

jmsmkn: Yep!

PR #4427, opened 2025-11-18
Make raised validation error serializable

koopmant:
Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/674

### DIFF
--- a/app/grandchallenge/components/models.py
+++ b/app/grandchallenge/components/models.py
@@ -1479,7 +1479,7 @@ class ComponentInterfaceValue(models.Model, FieldChangeMixin):
                 except JSONDecodeError as error:
                     raise ValidationError(
                         f"The file is not valid JSON. {error}"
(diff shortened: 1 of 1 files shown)


app/grandchallenge/components/models.py
```diff
@@ -1477,7 +1477,9 @@ def validate_user_upload(self, user_upload):
                 try:
                     value = json.loads(user_upload.read_object())
                 except JSONDecodeError as error:
-                    raise ValidationError(error)
+                    raise ValidationError(
+                        f"The file is not valid JSON. {error}"
+                    )
```
jmsmkn:
```suggestion
                    ) from error
```

jmsmkn: We should be doing this everywhere this pattern occurs, so curious if this works here.

PR #4428, opened 2025-11-19
Add grand challenge forge

jmsmkn:
This PR moves Grand Challenge Forge into this repo. The only changes for integration are in a3ed073, the rest is as-is. The main change is that `grand_challenge_forge` is now `grandchallenge.forge`.

The history for grand challenge forge will continue to be available at https://github.com/DIAGNijmegen/rse-grand-challenge-forge. After this PR is merged that will be archived.

Due to the integration tests requiring a full docker environment they are run separately in CI. The helper tests were causing deadlocks as they invoke os.fork whilst running in a multiprocessing environment, so they are run sequentially. The code has been tidied up to clarify variables, functions that are only used in tests have been moved, dead code removed, and some functions replaced with django built ins.

See https://github.com/DIAGNijmegen/rse-roadmap/issues/457


chrisvanrun:
Cool! Thanks for picking this up James!

PR #4431, opened 2025-11-20
Ensure loading deidentification js for DICOM upload widget

koopmant:

### DIFF
--- a/app/grandchallenge/uploads/static/js/user_upload.js
+++ b/app/grandchallenge/uploads/static/js/user_upload.js
@@ -111,7 +111,8 @@
                     window.alert(
                         `Could not upload ${file.name} (${file.type}): ${e.message}`,
                     );
-                    // Set autoProceed to false to avoid race condition here.
+                    // `autoProceed` must be set to `false` to avoid race condition here. Otherwise, removeFile may get
+                    // called before uppy file initialization is complete, leading to errors.
                     uppy.removeFile(file.id);
                 }
 


koopmant:
Maybe we should fix this pattern, it seems a bit fragile as we've seen here.

koopmant:
It only activates for the dicom upload widget now.

app/grandchallenge/uploads/static/js/user_upload.js
```diff
@@ -99,22 +100,34 @@
...
-            if (globalThis.UPPY_FILE_PREPROCESSORS) {
-                for (const preprocessor of globalThis.UPPY_FILE_PREPROCESSORS) {
-                    if (await preprocessor.fileMatcher(file)) {
-                        try {
-                            const processedFile =
-                                await preprocessor.preprocessor(file.data);
-                            uppy.setFileState(file.id, { data: processedFile });
-                        } catch (e) {
-                            uppy.removeFile(file.id);
+        if (isDicomWidget) {
```
chrisvanrun, line 103: I am not sure about this check, there could be other preprocessors. Is it a better implementation to have checks on the content of `UPPY_FILE_PREPROCESSORS` rather?

chrisvanrun (changes requested):
Fixes it, and I think the error if no procedure is loaded is a good one. I wonder about the preprocessor handling though. Seems like a general construct and now we put it behind a switch if it's a specific widget.

app/grandchallenge/uploads/static/js/user_upload.js
koopmant: It was created as a general construct, but we only use it for the dicom preprocessor. To be honest, it was quite confusing. We call it uppy preprocessors, but uppy has other endpoints for preprocessing ([uppy.addPreProcesser(fn)](https://uppy.io/docs/uppy/#addpreprocessorfn)). A problem with those is that preprocessing takes place after some initial processing by uppy, just before uploading. Our current function expects the raw file, not whatever uppy transformed it into. So we want to hook into the pipeline before uppy touches it and therefore cannot use the `addPreProcessor` method. 

Anyway, to get back to your main point. Yes, this was generalised, but I don't think it should be until we actually need it. If we need it, we should probably use `addPreProcessor` properly instead of adding our own logic. 

Should I add a comment about why we're not using `addPreProcessor`?

chrisvanrun: > Should I add a comment about why we're not using addPreProcessor?

Yes, let us help our poor future selves to understand why we did it this way! 

And I agree with your point in perhaps generalizing too early in this case. Maybe we should just rename it DICOM_PREPROCESSING or something? That way we also don't introduce the idea of generalization here, preventing our future selves from thinking adding stuff to that will automagically have Uppy use it for preprocessing.

jmsmkn: > Maybe we should just rename it DICOM_PREPROCESSING or something?

Even more explicit: `CLIENT_SIDE_DICOM_DEIDENTIFICATION`

koopmant: > Yes, let us help our poor future selves to understand why we did it this way!

Haha, will do!

app/grandchallenge/uploads/static/js/file_preprocessors.js
```diff
@@ -113,27 +113,6 @@ function getDummyValue(vr) {
     }
 }
 
-// Helper to check if a file is a DICOM file by extension or magic byte
```
koopmant, line 116: I threw this away because dcmjs rejects non-dicom files already. And dicom files don't always have an extension, so filtering on that seems to strict. 

I could set allowed_file_types to `application/dicom` for the dicom widget to add a file restriction before calling dcmjs, should I do that?

app/grandchallenge/uploads/static/js/user_upload.js
koopmant: I removed the global variable, because we didn't need it at all.

app/grandchallenge/uploads/static/js/user_upload.js
```diff
@@ -99,22 +100,27 @@
...
+            // so we can deal with the raw file blob, instead of the uppy mutated file data.
+            uppy.on("file-added", async file => {
+                try {
+                    const processedFile = await preprocessDicomFile(file.data);
+                    uppy.setFileState(file.id, { data: processedFile });
+                } catch (e) {
+                    window.alert(
+                        `Could not upload ${file.name} (${file.type}): ${e.message}`,
+                    );
+                    // Set autoProceed to false to avoid race condition here.
```
chrisvanrun: I don't quite understand this comment. Could you please elaborate?

koopmant: I got errors when autoProceed was true, because `removeFile` was being called before uppy was finished initializing the file. Do you have a suggestion to make the comment clearer?

app/grandchallenge/uploads/static/js/file_preprocessors.js
koopmant: Okay, did that. But it turns out many dicom files actually have the mimetype `application/octet-stream`, so I've added that too.

app/grandchallenge/uploads/static/js/user_upload.js
koopmant: I updated it.

chrisvanrun: Thanks! I think that clears it up for me, and I hope for our future selves. We are sure keeping tabs on our future selves! =)

PR #4435, opened 2025-11-21
Cancel linked job on dicom import failure

koopmant:
Also add retry for HealthImaging `ConflictError("Requested ImageSet metadata is not consistent yet.")` during completed import job event handling.

The linked job and socket slug are saved on the DICOMImageSetUpload model so they can be recollected during error handling. The dicom import is event based, so the linked object cannot be handed over directly between tasks. 


app/grandchallenge/cases/models.py
```diff
@@ -1164,6 +1165,12 @@ class DICOMImageSetUpload(UUIDModel):
         editable=False,
         help_text="Serialized task that is run on job success",
     )
+    error_handling_data = models.JSONField(
```
jmsmkn: Why does this need to be saved on the model?

koopmant: When we handle dicom import job events, we only get enough information to get the `DICOMImageSetUpload` object. In the tasks for processing `RawImageUploadSession` we pass this information along, but in this case we have to get it from somewhere else. 

The only alternative I can think of is to work with signals. (Which would be cleaner, I expect.)

Do we have other options?

jmsmkn: I don't see how signals would be an alternative here.

Maybe I'm just confused about the variable `error_handling_data`. It doesn't seem to match any existing field/construct that we have. I don't see where it is set, or a schema. Why can't the existing task on success/task on failure constructs be used?

koopmant: Oh right, creating a task on failure would make a lot more sense! Thanks. 

As for signals, I haven't looked into it, but we create the job and the dicom image upload object. I think if the job listens to a signal from the dicom image upload object when it's status changes to failed, the upload object doesn't need to worry about passing along information to clean up a job.

jmsmkn: It might be worth reading the signals documentation, especially the warning about when they are and are not appropriate: https://docs.djangoproject.com/en/5.2/topics/signals/

They should only be used when code cannot be called directly (e.g. when you need to call user code from a 3rd party package), I don't think that is the problem you are trying to solve here.

koopmant: I went back to saving the linked object to the `DICOMImageSetUpload` model, but no longer in the undefined `error_handling_data` structure. It now has a generic foreign key for the linked object and a CharField for the linked socket slug. I used a ForeignKey initially, but that leads to a circular dependency in the migrations.

app/grandchallenge/cases/models.py
```diff
@@ -1164,6 +1167,24 @@ class DICOMImageSetUpload(UUIDModel):
...
+    linked_object_content_type = models.ForeignKey(
+        ContentType,
+        blank=True,
+        null=True,
+        related_name="dicomimagesetupload_linked_object",
+        on_delete=models.CASCADE,
+        db_index=True,
+    )
+    linked_object_object_id = models.CharField(
+        max_length=255, blank=True, null=True, db_index=True
```
jmsmkn: Both new fields probably should have `editable=False` rather than `blank=True`.

`db_index=True` should not be used, it is on the path to deprecation. An index defined in meta should be used. How that it set up will depend on how you are querying this data - maybe you will want separate ones, maybe a joint one is more efficient.

Why is `linked_object_object_id` a `CharField` and not a `UUIDField`?

If it should be a char field then it should not be nullable.

app/grandchallenge/cases/models.py
```diff
@@ -1164,6 +1167,24 @@ class DICOMImageSetUpload(UUIDModel):
...
+    )
+    linked_object = GenericForeignKey(
+        "linked_object_content_type", "linked_object_object_id"
+    )
+    linked_socket_slug = models.CharField(
```
jmsmkn: Why isn't this a foreign key to a socket? Again, probably should have `editable=False` rather than `blank=True`.

app/grandchallenge/cases/tasks.py
```diff
@@ -590,7 +629,56 @@ def handle_health_imaging_import_job_event(*, event):
...
+            )
+        else:
+            raise ValueError("Invalid job status")
+
+        upload.delete_input_files()
+    except (
+        botocore.exceptions.EndpointConnectionError,
+        health_imaging_client.exceptions.ThrottlingException,
+        health_imaging_client.exceptions.ServiceQuotaExceededException,
+    ) as e:
```
jmsmkn: Nit: would rather spell out that it is an `error` rather than using `e`. Sometimes things like this clash with the debugger commands, and I think it is clearer for other developers.

app/grandchallenge/cases/tasks.py
```diff
@@ -590,7 +629,56 @@ def handle_health_imaging_import_job_event(*, event):
     if upload.status != DICOMImageSetUploadStatusChoices.STARTED:
         return
 
-    upload.handle_event(event=event)
+    def _handle_error(*, error):
```
jmsmkn: The two _handle_error methods look very similar and I think upload deletion is missing here. Why not add this to the model:

```python
    def _handle_error(self, *, error_message):
        self.user_uploads.all().delete()
        self.delete_input_files()

        on_commit(
            handle_dicom_import_error.signature(
                kwargs={
                    "dicom_imageset_upload_pk": dicom_imageset_upload_pk,
                    "error_message": error_message,
                }
            ).apply_async
        )
```

and use that in both places? The `logger.error(error, exc_info=True)` for the expected errors can then be added where used.

app/grandchallenge/cases/models.py
koopmant: I used a ForeignKey initially, but that leads to a circular dependency in the migrations.

app/grandchallenge/cases/models.py
koopmant: Thanks. I copied this from the `Notification` model. Should I clean it up there as well?

jmsmkn: Not in this PR, there might be other considerations there (e.g. linking to things with non UUID pks).

app/grandchallenge/cases/models.py
jmsmkn: Hmm. The danger here is that the slug _could_ change (it shouldn't but there are sometimes requests for it and it is possible as an admin action). In that case it wouldn't be kept in sync. It would be safer to send and save the primary key which doesn't change (if it does you get a new object).

Does the error handler absolutely need the interface or can that be removed somehow?

app/grandchallenge/cases/tasks.py
koopmant: Thanks, looks a lot better.

app/grandchallenge/cases/models.py
koopmant: It is used to indicate which input socket failed validation for jobs. I'm not entirely sure if that is needed. The notification still links to the upload object; so it should be easy enough to identify. 

The logic has been copied from `RawImageUploadSession` which can receive a detailed error message with errors for multiple sockets. I'm guessing that was only possible through the API, right? 

The only place where we create `RawImageUploadSession` now is in `_update_civ_for_image`, where we create a session for a single socket.

jmsmkn: Unsure I'm afraid! There can be multiple upload sessions per linked object, but only one upload session per socket as sockets cannot be duplicated per linked object. That should be possible through the templates and the API.

app/grandchallenge/cases/models.py
jmsmkn: Nice catch of the `on_delete`, totally missed that one!

koopmant: Yeah, glad I saw it!

app/grandchallenge/cases/models.py
koopmant: All right, I've changed it to use the pk for now.

PR #4442, opened 2025-11-26
Use pydantic models within forge

jmsmkn:
Dictionaries are still used in the tests but they are doing some weird manipulation, so not touching those.


jmsmkn:
> LGTM! I wondered if we should not use some sort of named importing to differentiate the forge models from the main DB models, but I am kinda digging the 'Forge' prefix!

Yes thought about that, but then decided on the prefix due to the auto-importing from IDEs, you would have two instances of say `Phase` in the codebase and it is easy to pick the incorrect one. I do it all the time with `settings` as PyCharm prefers the local one that shouldn't be used.

PR #4443, opened 2025-11-26
Create forge models directly

jmsmkn:
I don't think it makes sense that the phase returns a `ForgeChallenge`, but forge doesn't work with phases directly. To be refactored later.


chrisvanrun:
True, it used to work with challenges and generate a per phase example. That way we would create a single challenge pack for a challenge. We then made it such that it was per phase, but never deconstructed the via challenge + multiple phases construct.

app/grandchallenge/archives/models.py
```diff
@@ -298,6 +299,13 @@ def allowed_socket_slugs(self):
         else:
             raise AttributeError
 
+    @cached_property
```
chrisvanrun, line 302: Oh, this construct is a thing of beauty. So elegant to put it into the main models.

jmsmkn: There are positives and negatives to both service layer vs fat models, but Django encourages the latter.

chrisvanrun:
> ImportError: cannot import name 'ComponentJob' from partially initialized module 'grandchallenge.components.models' (most likely due to a circular import) (/app/grandchallenge/components/models.py)

PR #4446, opened 2025-11-28
Forge skip the phase directory

chrisvanrun:
Changes:

- Minor mentions of challenge packs.
- Underscore starter-kit name / directory.
- Remove the phase-slug named directory, the examples are now directly rendered under the containing folder. Possible with the focus on phase instead of challenge.


jmsmkn (approved):
Made the same change but then reverted it as I wasn't sure it would be on the top level.

PR #4454, opened 2025-11-28
Create dicom deidentifier uppy plugin

### LINKED ISSUE #4447: Client side de-identifier does not process multiple files
**Describe the bug**
Only the first file is processed. 

**To Reproduce**
1. Make sure a dicom kind socket exists
2. Go to archive item create
3. Add value for new socket, select dicom
4. Drag/upload multiple files at once to the upload widget
5. Save
6. Receive a notification that the dicom upload failed because the StudyDate was inconsistent for the uploaded files

**Additional context**
I believe we need to follow the docs and build a simple Plugin: https://uppy.io/docs/guides/building-plugins/#baseplugin.

The issue with the current implementation is that it signals for uppy to upload after only the first file has been processed. The following files are uploaded unprocessed. 
See `user_upload.js` line 121. 

The underlying issue is that we hook into the `file-added` event. I don't think that event is meant for how we are using it. When we try to remove the file when an error happens, we run into a race condition if an alert has been shown. If we set autoProceed to false to prevent this, we don't have a way of proceeding with the upload.

koopmant:
Use an uppy plugin, replacing the event hook to `file-added`, to enable processing multiple files. 

Closes #4447

### DIFF
--- a/app/grandchallenge/uploads/static/js/dicom_deidentification.js
+++ b/app/grandchallenge/uploads/static/js/dicom_deidentification.js
@@ -336,7 +336,7 @@ class DicomDeidentifierPlugin extends Uppy.Core.BasePlugin {
     constructor(uppy, opts) {
         super(uppy, opts);
 
-        this.id = opts.id || "DicomDeidentifierPlugin";
+        this.id = opts?.id || "DicomDeidentifierPlugin";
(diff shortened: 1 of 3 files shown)


HarmvZ (approved):
Clean solution!

Issue #4452, opened 2025-11-28
Forge example algorithm uses old pytorch


chrisvanrun:
Forge currently uses `pytorch/pytorch` to base the examples on. However, [that tag](https://hub.docker.com/layers/pytorch/pytorch/latest/images/sha256-11691e035a3651d25a87116b4f6adc113a27a29d8f5a6a583f8569e0ee5ff897) hasn't been updated for the past 2 years. [Others](https://hub.docker.com/r/pytorch/pytorch/tags) have.

This wasn't a problem until we try to add the `pylibjpeg-libjpeg` dependency to the mix. This causes numpy version conflicts with the 'ancient' pytorch image.

<details>
<summary>Long stderr output. Basically says "A module that was compiled using NumPy 1.x cannot be run in
NumPy 2.2.6 as it may crash. "</summary>
<code>
A module that was compiled using NumPy 1.x cannot be run in
NumPy 2.2.6 as it may crash. To support both 1.x and 2.x
versions of NumPy, modules must be compiled with NumPy 2.0.
Some module may need to rebuild instead e.g. with 'pybind11>=2.12'.
 
If you are a user of the module, the easiest solution will be to
downgrade to 'numpy<2' or try to upgrade the affected module.
We expect that some modules will need time to support NumPy 2.
 
Traceback (most recent call last):  File "/opt/app/inference.py", line 26, in <module>
    import torch
  File "/opt/conda/lib/python3.10/site-packages/torch/__init__.py", line 1477, in <module>
    from .functional import *  # noqa: F403
  File "/opt/conda/lib/python3.10/site-packages/torch/functional.py", line 9, in <module>
    import torch.nn.functional as F
  File "/opt/conda/lib/python3.10/site-packages/torch/nn/__init__.py", line 1, in <module>
    from .modules import *  # noqa: F403
  File "/opt/conda/lib/python3.10/site-packages/torch/nn/modules/__init__.py", line 35, in <module>
    from .transformer import TransformerEncoder, TransformerDecoder, \
  File "/opt/conda/lib/python3.10/site-packages/torch/nn/modules/transformer.py", line 20, in <module>
    device: torch.device = torch.device(torch._C._get_default_device()),  # torch.device('cpu'),
[...]

PR #4455, opened 2025-12-01
Add DICOM Image Set examples to forge

chrisvanrun:
Close: #4452 

Adds DICOM Image set examples to forge.

In order to include a working dependency,  `pylibjpeg`, for decompression via `pydicom`, the base images have changed. While these are bigger, they are by default cached in SageMaker runtime. In theory, reducing pulling time. Unsure if we should cache these in the GitHub runners for when doing CI tests.

While I was changing the code I spotted a TODO for empty examples values, I've added it using `pydantic_core.MISSING`. No further additions had to be made for `predictions.json` generation, since the current construct for images does not differentiate between a DICOM Image Set and a panimg image.

### DIFF
--- a/app/tests/forge_tests/test_algorithm_template_generation.py
+++ b/app/tests/forge_tests/test_algorithm_template_generation.py
@@ -3,6 +3,7 @@ from uuid import uuid4
 
 import pytest
 
+from grandchallenge.components.models import INTERFACE_KIND_JSON_EXAMPLES
 from grandchallenge.forge.forge import generate_algorithm_template
 from grandchallenge.forge.models import ForgeAlgorithm
 from tests.forge_tests.utils import (
@@ -38,6 +39,50 @@ def test_for_algorithm_template_content(tmp_path):
         "test/input/interf1",
     ]:
         assert (template_path / filename).exists()
+
+
+@pytest.mark.parametrize(
+    "example_value",
+    INTERFACE_KIND_JSON_EXAMPLES.values(),
+)
+def test_algorithm_template_example_value_rendering(example_value, tmp_path):
+    testrun_zpath = Path(str(uuid4()))
+
+    context = algorithm_template_context_factory(
+        algorithm_interfaces=[
+            {
+                "inputs": [
+                    {
+                        "slug": "input-socket-slug",
+                        "relative_path": "input-value.json",
+                        "example_value": example_value.value,
+                        "is_file_kind": True,
+                        "is_json_kind": True,
+                    },
+                ],
+                "outputs": [
+                    {
+                        "slug": "output-socket-slug",
+                        "relative_path": "output-value.json",
+                        "example_value": example_value.value,
+                        "is_file_kind": True,
+                        "is_json_kind": True,
+                    },
+                ],
+            }
+        ],
+    )
+
+    # Black post-processing already tests that example values with escaped characters can
+    # correctly be parsed by a Python interpreter
+    with zipfile_to_filesystem(
+        output_path=tmp_path, preserve_permissions=False
+    ) as zip_file:
+        generate_algorithm_template(
+            algorithm=ForgeAlgorithm(**context),
+            output_zip_file=zip_file,
+            target_zpath=testrun_zpath,
+        )
 
 
 @pytest.mark.forge_integration


app/grandchallenge/forge/resources/example_1.dcm
jmsmkn, line 1: How was the example data generated? License okay, no PII etc?

app/grandchallenge/forge/templates/forge/partials/example_algorithm/inference.py.template
```diff
@@ -24,11 +24,17 @@ Happy programming!
 from pathlib import Path
 import json
 import torch
-{% if object.algorithm_input_sockets|has_image or object.algorithm_output_sockets|has_image %}
+
+{% if object.algorithm_input_sockets|has_image or object.algorithm_output_sockets|has_image or object.algorithm_input_sockets|has_dicom_image_set %}
```
jmsmkn: Missing check for `object.algorithm_output_sockets|has_dicom_image_set`? I know it's not implemented right now.

app/grandchallenge/forge/templates/forge/partials/example_algorithm/inference.py.template
```diff
@@ -24,11 +24,17 @@ Happy programming!
...
 from glob import glob
-import SimpleITK
 import numpy
 {% endif %}
+{% if object.algorithm_input_sockets|has_image or object.algorithm_output_sockets|has_image %}
+import SimpleITK
+{% endif %}
+{% if object.algorithm_input_sockets|has_dicom_image_set %}
```
jmsmkn: Same here

app/grandchallenge/forge/templates/forge/partials/example_algorithm/inference.py.template
```diff
@@ -160,6 +167,34 @@ def load_image_file_as_array(*, location):
 {% endif %}
 
 
+{% if object.algorithm_input_sockets|has_dicom_image_set %}
```
jmsmkn: And here

app/grandchallenge/forge/templates/forge/partials/example_algorithm/inference.py.template
```diff
@@ -160,6 +167,34 @@ def load_image_file_as_array(*, location):
...
+def load_dicom_image_set_as_array(*, location):
+    pass
+{% comment %}
```
jmsmkn: Doesn't this disappear from the rendered template?

app/grandchallenge/forge/templates/forge/partials/example_algorithm/requirements.txt.template
```diff
@@ -1,5 +1,13 @@
 {% load forge %}
+{% if object.algorithm_input_sockets|has_image or object.algorithm_output_sockets|has_image or object.algorithm_input_sockets|has_dicom_image_set %}
+numpy
+{% endif %}
 {% if object.algorithm_input_sockets|has_image or object.algorithm_output_sockets|has_image %}
 SimpleITK
-numpy
+{% endif %}
+{% if object.algorithm_input_sockets|has_dicom_image_set %}
```
jmsmkn: 2x here

app/grandchallenge/forge/models.py
```diff
@@ -32,15 +33,18 @@ class ForgeSocket(BaseModel):
     relative_path: str
     kind: ForgeKindEnum
     super_kind: ForgeSuperKindEnum
-    example_value: Any = None
+    example_value: Any = MISSING
```
jmsmkn, line 20: I spent ages searching for this last week, thanks!

app/grandchallenge/forge/templates/forge/partials/example_algorithm/inference.py.template
chrisvanrun: Oh no. I think I messed up re-inserting some code when doing debugging with the dependency + containers =/

app/grandchallenge/forge/resources/example_1.dcm
chrisvanrun: Totally fabricated with Pydicom.

app/grandchallenge/forge/templates/forge/partials/example_algorithm/inference.py.template
chrisvanrun: Indeed, output is being ignored right now. Would you say we'll implement it for the templating now?

app/grandchallenge/forge/models.py
```diff
@@ -49,10 +53,16 @@ def is_file(self):
             and not self.relative_path.endswith(".json")
         )
 
+    @property
+    def is_dicom_image_set(self):
+        return (
+            self.super_kind == ForgeSuperKindEnum.IMAGE
+            and self.kind == ForgeKindEnum.DICOM_IMAGE_SET
```
jmsmkn: We could take these properties from GC rather than calculating them here (is_dicom_image_set corresponds to is_image - these correspond to is_dicom_image_kind, is_panimg_kind and is_image_kind in the rest of the GC codebase, having the separate is_image_kind might save some of those long if statements). Even if not, it is more likely that we extend the DICOM image types rather than the PANIMG types.

app/grandchallenge/forge/templates/forge/partials/example_algorithm/inference.py.template
jmsmkn: Maybe, we would still need to do some work when it is actually supported but the if statements look a bit imbalanced right now.

app/grandchallenge/forge/models.py
chrisvanrun: Good idea. I'll trace them a bit since they might have drifted over time.

jmsmkn (approved):

chrisvanrun:
> LGTM! Suggestion for restructuring the templates - you could use template partials to remove the duplication of common code from evaluate.py/inference.py.

Definitely a good idea and one I had considered earlier. I am a bit doubtful of its final balance of trade-offs. The way I'd see it implemented is that you'd need a partial for each function, which themselves are generally just a few lines. I'd only put the main reading and writing for the additional inputs and outputs that the eval method and algorithm share. So you'd end up with a longish list of if statements, that each wraps an`include`. Which will make it more DRY but also couples them stronger and makes it harder (imho) to debug.

With the addition of the DICOM writing, it might have shifted the balance in favour of that abstraction. I might give it a go in the follow-up change set.

PR #4460, opened 2025-12-02
Ensure >1 interfaces when user confirms deletion

chrisvanrun:
Imho the via-form validation is enough: site admins should be able to unlink them when needed. Mostly am doubting 

## Preview
afbeelding


chrisvanrun:
> I think it should apply to everyone TBH - it otherwise leaves an algorithm in an unusable state. Site admins should be able to follow the same process.

Ok. Then I think it makes more sense to push it to the model/manger level.

PR #4465, opened 2025-12-05
Restructure forge templates with function partials

chrisvanrun:
This change set extracts common `load_X` and `write_X` functions to partial function templates and reduces duplication of these in evaluation methods and algorithms.

In addition, to clean up the if statements around the imports, they are included in the partial function templates. 

Subsequently, `isort` is used to float the imports to the top and clean up any duplicates.


uv.lock
```diff
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
+revision = 2
```
jmsmkn: @chrisvanrun You should update uv on your machine.

setup.cfg
```diff
@@ -18,6 +18,8 @@ banned-modules =
     celery.shared_task = Please use acks_late_2xlarge_task or acks_late_xlarge_task instead
     datetime.datetime.now = Please use django.utils.timezone.now instead
     grandchallenge.components.validate_scripts = Please do not import any validation script components
+exclude =
+  app/grandchallenge/forge/templates/forge/*
```
jmsmkn: No longer necessary

```suggestion
```

uv.lock
chrisvanrun: Brew was giving me some trouble with refusing to remove the old `uv` in my Path. Fixed now.

PR #4469, opened 2025-12-09
Allow setting the algorithm api method

jmsmkn:
Adds the ability to set the API method by adding a label on the container (`org.grand-challenge.api-method`). The value of this can be `exec` (default, set if label is missing) or `invoke` (new). The api method is then passed through to the instance at runtime as `GRAND_CHALLENGE_COMPONENT_API_METHOD` for pickup by SageMaker Shim.

See https://github.com/DIAGNijmegen/rse-roadmap/issues/400


app/grandchallenge/challenges/migrations/0001_initial.py
```diff
@@ -22,6 +22,7 @@ class Migration(migrations.Migration):
         migrations.swappable_dependency(settings.AUTH_USER_MODEL),
         ("task_categories", "0001_initial"),
         ("publications", "0003_auto_20201001_0758"),
+        ("modalities", "0001_initial"),
```
koopmant, line 25: Why is this necessary?

jmsmkn: There is a m2m relationship being created in this migration: https://github.com/DIAGNijmegen/rse-grand-challenge/blob/aa0d67e679bc34a77c6a0bd918b56f57e081cac7/app/grandchallenge/challenges/migrations/0001_initial.py#L199-L206

It showed up as the migrations were being applied in a different order on fresh databases with this PR. I don't know why the dependency wasn't there in the first place.

PR #4471, opened 2025-12-09
Load dicom deidentifier js as module

koopmant:
Load the dicom deidentifier js as an ES module. 
Run jest with `--experimental-vm-modules` to import from the module in the tests. This feature is experimental, but is running without issues. 

Additionally the `uidMap` is now defined as a global and not recreated if it already exists, making it available across multiple widgets on a page. 

This resolves the following errors (Shown in console on redisplay of form after validation errors)
- Syntax error: '_uidMap' is already defined. 
- Syntax error: 'DicomDeidentifierPlugin' is already defined.

Closes https://github.com/DIAGNijmegen/rse-grand-challenge/issues/4470

Alternative to #4472

### DIFF
--- a/app/grandchallenge/uploads/static/js/dicom_deidentification.mjs
+++ b/app/grandchallenge/uploads/static/js/dicom_deidentification.mjs
@@ -372,4 +372,4 @@ export class DicomDeidentifierPlugin extends Uppy.Core.BasePlugin {
 }
 
 // For testing only
-export { getDummyValue, preprocessDicomFile, uidMap };
+export { getDummyValue, preprocessDicomFile };


pkcakeout:
`require` is an older module that allows loading something that is _similar to modules_ predating the modules feature itself: https://requirejs.org/

I do not think that you can load a module with require.js, modules need to be loaded with the browser's module engine instead (caveat: a bit wonky knowledge base here for me).

From my knowledge that is why you have `<script type="module">` for which you are allowed to use import statements which then are also scoped while the scope feature is entirely missing otherwise and with that no `import` or `export` use is allowed.

koopmant:
Thanks! And good point, that is a good feature indeed. I hadn't even realized. I'll add it.

PR #4473, opened 2025-12-10
Remove use of auto_id without pattern for dynamically added sockets

koopmant:
Replace use of auto_id with a dedicated form_id. Supplying `auto_id` changes django's behaviour of setting the id on form elements (it removed the default "id_" prefix) as explained in the [docs here](https://docs.djangoproject.com/en/5.2/ref/forms/api/#configuring-form-elements-html-id-attributes-and-label-tags). Because we only use the value to set the form id, we can use our own context variable for that.

Also clean up some things:
- Rename `DICOMUploadWidgetSuffixes` to `DICOM_UPLOAD_WIDGET_SUFFIXES`
- Only check if socket exists for socket slugs that start with the right prefix (reduce calls to database when loading form fields for dynamically added sockets)

### DIFF
--- a/app/grandchallenge/components/forms.py
+++ b/app/grandchallenge/components/forms.py
@@ -218,8 +218,11 @@ class MultipleCIVForm(Form):
             interface_slug = self.parse_slug(slug=slug)
 
             if (
-                ComponentInterface.objects.filter(slug=interface_slug).exists()
+                interface_slug
                 and slug not in self.fields.keys()
+                and ComponentInterface.objects.filter(
+                    slug=interface_slug
+                ).exists()
             ):
                 interface = ComponentInterface.objects.filter(
                     slug=interface_slug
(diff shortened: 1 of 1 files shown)


app/grandchallenge/components/forms.py
```diff
@@ -332,10 +339,12 @@ def __init__(
         interface,
         base_obj,
         user,
+        form_id,
         htmx_url,
         **kwargs,
     ):
         super().__init__(*args, **kwargs)
+        self.id = form_id
```
chrisvanrun, line 347: `self.id` seems a bit pointless in isolation. I assume it is used in a follow PR?

koopmant: It's used in the template:
```html
id="form-{{ form.id }}"
```

chrisvanrun: Ah, ofc! Totally missed that!

PR #4474, opened 2025-12-10
Filter image search on dicom kind

koopmant:
Filter the image queryset for search results and field validation on dicom kind. 

The queryset filters whether `dicom_image_set` is set on the images, returning only images with dicom_image_set if the requested interface is of dicom image kind and returning only images without dicom_image_set if not. 

### DIFF
--- a/app/tests/cases_tests/test_views.py
+++ b/app/tests/cases_tests/test_views.py
@@ -135,7 +135,6 @@ def test_image_search_view(client):
         },
     )
     assert response.status_code == 200
-    assert len(response.context_data["object_list"]) == 1
     assert response.context_data["object_list"].get() == images[1]
 
     response = get_view_for_user(
@@ -165,7 +164,6 @@ def test_image_search_view(client):
         },
     )
     assert response.status_code == 200
-    assert len(response.context_data["object_list"]) == 1
     assert response.context_data["object_list"].get() == images_dicom[1]
 
     response = get_view_for_user(


app/grandchallenge/cases/views.py
```diff
@@ -346,6 +346,15 @@ class ImageSearchResultView(
     def get(self, request, *args, **kwargs):
         qs = self.get_queryset()
         prefixed_interface_slug = request.GET.get("prefixed-interface-slug")
+        interface = get_object_or_404(
+            ComponentInterface,
+            slug=prefixed_interface_slug.replace(
+                INTERFACE_FORM_FIELD_PREFIX, ""
+            ),
```
chrisvanrun, line 353: What if there is no prefixed_interface_slug. This would result in a server error. Should we handle it more gracefully? And give back a 404 as welll of sorts?

app/grandchallenge/cases/widgets.py
```diff
@@ -110,12 +110,15 @@ class FlexibleImageField(MultiValueField):
...
         *args,
+        interface,
         user=None,
         initial=None,
         **kwargs,
     ):
         image_search_queryset = filter_by_permission(
-            queryset=Image.objects.all(),
+            queryset=Image.objects.filter(
+                dicom_image_set__isnull=not interface.is_dicom_image_kind
```
chrisvanrun, line 120: Double negation, my poor head. I guess these odd queries are the result of our choice to make it a fundamental part of the existing `Image`.

More an observation than a comment.

app/tests/cases_tests/test_views.py
```diff
@@ -105,40 +106,84 @@ def test_permission_filtered_views(self, client):
...
+        client=client,
+        user=user,
+        data={
+            "prefixed-interface-slug": ci_dicom.slug,
+            f"query-{ci_dicom.slug}": "test",
+        },
+    )
+    assert response.status_code == 200
+    assert len(response.context_data["object_list"]) == 1
+    assert response.context_data["object_list"].get() == images_dicom[1]
```
chrisvanrun: `get()` already only works if there is 1, so the length check is not needed.

chrisvanrun (changes requested):
Minor suggestions.

app/grandchallenge/cases/widgets.py
koopmant: Yes, there is no way around this one that I can think of.

app/grandchallenge/cases/views.py
koopmant: Nice catch, thanks

koopmant: Returning a 404 now.

app/tests/cases_tests/test_views.py
koopmant: Thanks, removed it.

PR #4476, opened 2025-12-11
Forge: skip canceled futures in error collection

### LINKED ISSUE #4438: Forge: Templated evaluation container does not exit on error, error is not reported
Yesterday and today I spent a couple of hours debugging why an evaluation container stalled. I checked the logs on Sagemaker but there was nothing there. The container stopped doing stuff after mounting the inputs for the container and then just sat around doing apparently nothing for an hour, until Sagemaker killed the container because the time limit of 1 hour was reached. 

I assumed that error messages would be raised and appear in the logs of the container, like they do for algorithm jobs, and since that was not the case here, I assumed that something else was going on and was led down the wrong rabbit hole in debugging this. This container was using the recently added improvements for multiprocessing implemented here (https://github.com/DIAGNijmegen/rse-grand-challenge-forge/pull/41). We ultimately found the error by adding a simple print statement to the `handle_error()` function that prints the error. 

I'm unsure if we should add this print statement there, users should obviously be informed of errors in their scripts, but I assume that the solution is a bit more involved since the container should exit when it encounters an error, and that clearly did not happen in this case. Something with the termination of the child processes didn't work. @pkcakeout can hopefully take a look when he is back. 

Feel free to rename the issue once you know what the problem is. I might not be hitting the nail on its head here.

chrisvanrun:
Closes #4438 

In addition, it refactors `helpers.py` to `helpers.py.template` to ensure it isn't accidentally used in the GC code-base proper.

### DIFF
--- a/app/tests/forge_tests/test_evaluate_helpers.py
+++ b/app/tests/forge_tests/test_evaluate_helpers.py
@@ -2,6 +2,7 @@ import shutil
 import sys
 import tempfile
 import time
+from concurrent.futures import Future
 from functools import partial
 from multiprocessing import Process
 from pathlib import Path
@@ -168,9 +169,6 @@ def test_prediction_processing_canceling_processes_correctly(helpers):
         {"pk": "prediction2"},  # <- Should fail at this
         {"pk": "prediction3"},
         {"pk": "prediction4"},
-        {"pk": "prediction5"},
-        {"pk": "prediction6"},
-        {"pk": "prediction7"},
     ]
     with mock.patch.object(
         helpers,
@@ -186,3 +184,41 @@ def test_prediction_processing_canceling_processes_correctly(helpers):
         # Returning here on time is sufficient to ensure that things work as intended
         # Sanity: check that there is indeed only
         assert len(exc.value.predictions) == 1
+
+
+@pytest.mark.forge_integration
+def test_collect_errors(helpers):
+
+    future0 = Future()  # Finished no exception
+    future0.set_running_or_notify_cancel()
+    future0.set_result("foo")
+    assert future0.done()
+
+    future1 = Future()  # Finished with exception
+    future1.set_running_or_notify_cancel()
+    future1.set_exception(RuntimeError("Simulated failure"))
+    assert future1.done()
+
+    future2 = Future()  # Running
+    future2.set_running_or_notify_cancel()
+    assert future2.running()
+
+    future3 = Future()  # Canceled
+    future3.cancel()
+    assert future3.cancelled()
+
+    future4 = Future()  # Pending
+    assert not any([future4.done(), future4.running(), future4.cancelled()])
+
+    future_to_predictions = {
+        future0: {"pk": "prediction0"},
+        future1: {"pk": "prediction1"},
+        future2: {"pk": "prediction2"},
+        future3: {"pk": "prediction3"},
+        future4: {"pk": "prediction4"},
+    }
+
+    errors = {}
+    helpers._collect_errors(future_to_predictions, errors)
+
+    assert set(errors.keys()) == {"prediction1"}


app/grandchallenge/forge/templates/forge/partials/example_evaluation_method/helpers.py.template
```diff
@@ -175,7 +175,7 @@ def _collect_errors(future_to_predictions, errors):
     # Which prevents us relying solely on as_completed to catch exceptions
     def failed_futures():
         for f, p in future_to_predictions.items():
-            if f.done():
+            if f.done() and not f.cancelled():
```
chrisvanrun, line 178: This addition is the crucial fix.

app/tests/forge_tests/test_evaluate_helpers.py
```diff
@@ -132,19 +162,23 @@ def add_child_terminator(*args, **kwargs):
...
 @pytest.mark.timeout(4)
 @pytest.mark.forge_integration
-def test_prediction_processing_canceling_processes_correctly():
+def test_prediction_processing_canceling_processes_correctly(helpers):
     predictions = [
-        {"pk": "prediction1"},  # Should cancel this
-        {"pk": "prediction2"},  # Should fail at this
-        {"pk": "prediction3"},  # Should cancel this
-        {"pk": "prediction4"},  # Should cancel this
+        {"pk": "prediction1"},
```
chrisvanrun, line 168: Adding prediction7 finally adds a `PENDING` future that is converted into a `CANCELED` future, which is needed to show the behaviour. So the only way this would show up is if:
- an error occurs early
- there are sufficient jobs to work through that some are still pending
- we are in a few-CPU runtime environment so some futures are in the PENDING state and will be actually CANCELED

This was the case in #4438 latest seen occurrence.

I suspect that the reason the rest is in RUNNING state when the failing prediction2 is reached (despite there only being 2 workers via `max_workers`) is that the used pool seems to have some prep queue and that changes the states of the futures even if they haven't landed on a worker yet.

koopmant: For my understanding, this test failed without the added predictions? 
(I don't see how, but maybe it's my Friday brain missing something)

app/tests/forge_tests/test_evaluate_helpers.py
```diff
@@ -20,7 +45,7 @@ def working_process(p):
 
 
 def failing_process(*_):
-    raise RuntimeError("You have failed me for the last time")
+    raise RuntimeError("Stimulated Failing Process")
```
koopmant: I think you meant "Simulated" ?

koopmant (commented):
Looks good as far as I can tell (the futures thing goes a bit over my head). 

Question about the test. 
Seeing your comment on the issue:

> However, somehow the 2 which I expected to be in PENDING are in RUNNING state during collection, despite requesting everything to be cancelled during executor shutdown. This holds all the way till 6 predictions, and introducing the 7th finally introduces a pending future that has the collection fail.

Can we check their state explicitly in the test?

app/tests/forge_tests/test_evaluate_helpers.py
chrisvanrun: Haha. Maybe? ;-)

app/tests/forge_tests/test_evaluate_helpers.py
chrisvanrun: The test succeeded in the original configuration. This test is meant to test the error collection part of the pool worker. However, it only tested against RUNNING state futures and never against PENDING state futures. The latter is where things went wrong in practice.

koopmant: All right, makes sense. New test looks good!

PR #4477, opened 2025-12-12
Bind archive-item creation to upload permissions

chrisvanrun:
In addition, changed the composite permission codenames to explicit ones.

### DIFF
--- a/app/grandchallenge/archives/views.py
+++ b/app/grandchallenge/archives/views.py
@@ -358,7 +358,7 @@ class ArchiveItemUpdate(
     MultipleCIVProcessingBaseView,
 ):
     form_class = ArchiveItemUpdateForm
-    permission_required = "archiveitem.change_archiveitem"
+    permission_required = "archives.change_archiveitem"
     included_form_classes = (
         MultipleCIVForm,
         *MultipleCIVProcessingBaseView.included_form_classes,
--- a/app/grandchallenge/archives/apps.py
+++ b/app/grandchallenge/archives/apps.py
@@ -10,8 +10,8 @@ def init_archiveitem_permissions(*_, **__):
     g, _ = Group.objects.get_or_create(
         name=settings.REGISTERED_USERS_GROUP_NAME
     )
-    assign_perm("archiveitem.change_archiveitem", g)
-    assign_perm("archiveitem.add_archiveitem", g)
+    assign_perm("archives.change_archiveitem", g)
+    assign_perm("archives.add_archiveitem", g)
 
 
 class ArchivesConfig(AppConfig):


app/tests/archives_tests/test_views.py
```diff
@@ -1107,7 +1107,7 @@ def test_archive_item_create_permissions(client):
         client=client,
         user=uploader,
     )
-    assert response.status_code == 403
```
chrisvanrun, line 1110: fyi, in the original refactor for CIV creation, this tested status code being `403` was an oversight. Or so I think.

app/grandchallenge/archives/views.py
```diff
@@ -557,13 +541,9 @@ def get_success_url(self):
...
         if self.object:
-            return [
-                f"{Archive._meta.app_label}.change_{ArchiveItem._meta.model_name}"
-            ]
+            return ["archives.change_archiveitem"]
         else:
-            return [
-                f"{Archive._meta.app_label}.change_{Archive._meta.model_name}"
-            ]
+            return ["archives.upload_archive"]
```
chrisvanrun, line 546: Need to be able to create interfaces here when uploading.

app/grandchallenge/archives/views.py
```diff
@@ -366,9 +358,7 @@ class ArchiveItemUpdate(
     MultipleCIVProcessingBaseView,
 ):
     form_class = ArchiveItemUpdateForm
-    permission_required = (
-        f"{ArchiveItem._meta.app_label}.change_{ArchiveItem._meta.model_name}"
-    )
+    permission_required = f"{ArchiveItem._meta.app_label}.change_archiveitem"
```
koopmant:
```suggestion
    permission_required = "archives.change_archiveitem"
```

app/tests/archives_tests/test_views.py
koopmant: I think so too.

app/grandchallenge/archives/views.py
koopmant: Your comment threw me off a bit, but yes I think this is the right permission.

app/grandchallenge/archives/views.py
koopmant: There are two more in `archives/apps.py`, I think that should be all of them.

app/grandchallenge/archives/views.py
chrisvanrun: Ah, sorry! Meant to highlight that, here, it isn't a direct replacement of the `{X._meta.app_label}`.

app/grandchallenge/archives/views.py
chrisvanrun: Might as well change them there as well. On it!

chrisvanrun: Done!

PR #4479, opened 2025-12-16
Values for reserved socket slugs should not be allowed to be added

chrisvanrun:
This change set prevents the reserved socket values to:
- Be added via the API to an archive item, display set or Job
- Be added via the UI in:
  - Single CIV creation
  - Batch CIV creation
- Be looked up when selecting a via UI interface

It should not affect any sit admin actions.

Since most of these come in the shape of a mixin, I've opted to just test the archive route and not triplicate the tests to also apply to Jobs and Display Sets. Not sure if that is OK or not.

### DIFF
--- a/app/grandchallenge/components/models.py
+++ b/app/grandchallenge/components/models.py
@@ -2359,6 +2359,7 @@ class CIVData:
         self._dicom_upload_with_name = None
 
         ci = ComponentInterface.objects.get(slug=interface_slug)
+        self.interface = ci
 
         if ci.super_kind == ci.SuperKind.VALUE:
             self._init_json_civ_data()
@@ -2424,6 +2425,11 @@ class CIVData:
             self.dicom_upload_with_name,
             self.file_civ,
         ]
+
+        if self.interface_slug in RESERVED_SOCKET_SLUGS:
+            raise ValidationError(
+                f"Socket {self.interface.title!r} is reserved and cannot be used."
+            )
 
         # Ensure at most one of these properties is set
         # None can be an acceptable value, so 0 is ok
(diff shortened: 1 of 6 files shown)


app/grandchallenge/components/models.py
```diff
@@ -99,6 +99,15 @@
 logger = logging.getLogger(__name__)
 
 
+RESERVED_SOCKET_SLUGS = {
```
chrisvanrun, line 102: Put this in components as their role has changed but also because of circular imports.

app/grandchallenge/components/serializers.py
```diff
@@ -379,7 +379,7 @@ def update(self, instance, validated_data):
                 user=request.user,
             )
             logger.error(e, exc_info=True)
-        except CINotAllowedException as e:
+        except (CINotAllowedException, ValidationError) as e:
```
chrisvanrun: This allows the API to pick this up as a sensible error.

app/tests/archives_tests/test_views.py
```diff
@@ -228,21 +229,18 @@ def test_api_archive_api_detail_view(client):
 
 
 @pytest.mark.django_db
+@override_settings(task_eager_propagates=True, task_always_eager=True)
 def test_api_archive_item_allowed_sockets(
-    client, settings, django_capture_on_commit_callbacks
+    client, django_capture_on_commit_callbacks
```
chrisvanrun, line 234: Minor style update on this test. I initially had the reserved part here, but thought to keep the updated style.

app/grandchallenge/components/views.py
```diff
@@ -175,7 +180,12 @@ def base_object(self):
         raise NotImplementedError
 
     def form_valid(self, form):
-        form.process_object_data()
+        try:
+            form.process_object_data()
+        except ValidationError as e:
+            form.add_error(None, e)
+            return self.form_invalid(form)
+
```
chrisvanrun: This is my biggest doubt TBH.

We process the object data from cleaned data, which I think we should, but perhaps we should split the validation from the actual handling?

chrisvanrun: One workaround to this is correctly splitting the validation/cleaning of the `CIVData` with the actual process and linked tasks. I've done something like that here: https://github.com/DIAGNijmegen/rse-grand-challenge/pull/4480

chrisvanrun:
Merged the other PR into this, once this is approved it can be merged!

PR #4480, opened 2025-12-16
Split validation/cleaning and processing of civ_data_objects

chrisvanrun:
Splits the `validate_civ_data_objects_and_execute_linked_task` into:
- `validate_civ_data_objects`
- `process_civ_data_objects_and_execute_linked_task`

This allows to more cleanly handle validation errors being thrown by the `validate_civ_data_objects`

### DIFF
--- a/app/tests/archives_tests/test_models.py
+++ b/app/tests/archives_tests/test_models.py
@@ -10,7 +10,6 @@ from tests.components_tests.factories import (
     ComponentInterfaceValueFactory,
 )
 from tests.evaluation_tests.factories import PhaseFactory
-from tests.factories import UserFactory
 
 
 @pytest.mark.django_db
@@ -140,12 +139,11 @@ def test_archive_item_reserved_socket_slugs():
     with pytest.raises(
         ValidationError, match="is reserved and cannot be used."
     ):
-        ai.process_civ_data_objects_and_execute_linked_task(
-            civ_data_objects=[
+        ai.validate_civ_data_objects(
+            [
                 CIVData(
                     interface_slug=target_socket_slug,
                     value={},
                 )
             ],
-            user=UserFactory(),
         )


app/grandchallenge/components/views.py
```diff
@@ -180,12 +180,7 @@ def base_object(self):
         raise NotImplementedError
 
     def form_valid(self, form):
-        try:
-            form.process_object_data()
-        except ValidationError as e:
-            form.add_error(None, e)
-            return self.form_invalid(form)
-
```
chrisvanrun, line 188: This whole change set is done to remove this construct.

app/grandchallenge/components/forms.py
```diff
@@ -276,20 +276,27 @@ def parse_slug(*, slug):
...
-        for key, value in self.cleaned_data.items():
+        for key, value in cleaned_data.items():
             if key.startswith(INTERFACE_FORM_FIELD_PREFIX):
                 civ_data_objects.append(
                     CIVData(
                         interface_slug=key[len(INTERFACE_FORM_FIELD_PREFIX) :],
                         value=value,
                     )
                 )
+        self._civ_data_objects = civ_data_objects
```
koopmant, line 291: `Instance attribute _civ_data_objects defined outside __init__`

Wouldn't it make more sense to store this in `cleaned_data`?

app/grandchallenge/evaluation/models.py
```diff
@@ -1710,7 +1710,7 @@ def create_evaluation(self, *, additional_inputs):
                 )
 
         if additional_inputs:
-            evaluation.validate_civ_data_objects_and_execute_linked_task(
+            evaluation.process_civ_data_objects_and_execute_linked_task(
```
koopmant, line 1713: Shouldn't we validate here first?

app/grandchallenge/components/models.py
```diff
@@ -2493,22 +2493,20 @@ def remove_civ(self, *, civ):
...
-                user=user,
-                linked_task=linked_task,
-            )
-
-    def _update_civ(self, *, civ_data, user=None, linked_task=None):
+    def validate_civ_data_objects(self, /, civ_data_objects):
         if not self.is_editable:
             raise CIVNotEditableException(
                 f"{self} is not editable. CIVs cannot be added or removed from it.",
             )
```
koopmant, line 2500: I think this should remain in `_update_civ` or move to `process_civ_data_objects_and_execute_linked_task`.

app/grandchallenge/components/models.py
koopmant, line 2515: Now that we do it in validation, raising a `ValidationError` would be more appropriate I think.

PR #4483, opened 2025-12-17
Change "socket" to "interface" in forge template inputs.json

### LINKED ISSUE #4482: KeyError: 'socket' from latest forge template for algorithm
**Describe the bug**
Latest template assumes the key `socket` is in inputs.json, but the key is `interface`. 

**To Reproduce**
1. download the latest template for an algorithm
2. upload a container based on the template
3. Run a job
4. See error: KeyError, "socket" does not exist.

**Expected behavior**
containers made with the template should work without errors.

**Additional context**
Example content of inputs.json from a test run:

```
[
  {
    'interface': {
      'title': 'Cardiac OCT Image', 
      'description': 'Cardiac Optical Coherence Tomography image', 
      'slug': 'cardiac-oct', 
      'kind': 'Image', 
      'pk': 787, 
      'default_value': None, 
      'super_kind': 'Image', 
      'relative_path': 'images/cardiac-oct', 
      'overlay_segments': [], 
      'look_up_table': None
    }, 
    'value': None, 
    'file': None, 
    'image': {
      'pk': 'some-pk', 
      'name': '0033.rgb.nii.gz'
    }, 
    'pk': 3405148
  }
```

koopmant:
Closes #4482

### DIFF
--- a/app/grandchallenge/forge/templates/forge/partials/example_evaluation_method/evaluate.py.template
+++ b/app/grandchallenge/forge/templates/forge/partials/example_evaluation_method/evaluate.py.template
@@ -290,14 +290,14 @@ def read_predictions():
 
 def get_interface_key(job):
     # Each interface has a unique key that is the set of socket slugs given as input
-    socket_slugs = [sv["socket"]["slug"] for sv in job["inputs"]]
+    socket_slugs = [sv["interface"]["slug"] for sv in job["inputs"]]
     return tuple(sorted(socket_slugs))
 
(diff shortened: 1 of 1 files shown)


koopmant:
I wonder if I'm going in the right direction here. This might be a quick fix, but maybe we should generate `inputs.json` with both `socket` and `interface` keys instead?

chrisvanrun:
> I wonder if I'm going in the right direction here. This might be a quick fix, but maybe we should generate `inputs.json` with both `socket` and `interface` keys instead?

Yeah, I think that would likely be the better approach t.b.h.

PR #4484, opened 2025-12-17
Flexible DICOM Image fields

koopmant:
Add flexible fields for DICOM image kind sockets. 

These fields have the same functionality as the FlexibleImageField, but with a different implementation that avoids the use of HTMX and MultiValueField and MultiValueWidget. Instead, all fields are now added to the form. The search and upload fields are hidden by default and only shown though javascript when they are the selected choice in the source select widget. Form validation takes care of getting the data from the selected field. This enables us to let django handle all data routing to and from the correct fields. 

### DIFF
--- a/app/tests/cases_tests/test_views.py
+++ b/app/tests/cases_tests/test_views.py
@@ -2,8 +2,14 @@ import pytest
 from django.utils.html import format_html
 from guardian.shortcuts import assign_perm, remove_perm
 
-from grandchallenge.cases.widgets import ImageWidgetChoices
-from grandchallenge.components.forms import INTERFACE_FORM_FIELD_PREFIX
+from grandchallenge.cases.widgets import (
+    ImageSearchWidgetSuffixes,
+    ImageWidgetChoices,
+)
+from grandchallenge.components.forms import (
+    INTERFACE_FORM_FIELD_PREFIX,
+    FlexibleWidgetPrefixes,
+)
 from grandchallenge.components.models import ComponentInterface
 from tests.cases_tests.factories import (
     DICOMImageSetFactory,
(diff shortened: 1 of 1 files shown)


koopmant:
This should now be complete.

app/grandchallenge/cases/views.py
```diff
@@ -346,16 +353,17 @@ class ImageSearchResultView(
...
+        interface_slug = prefixed_interface_slug.replace(
+            INTERFACE_FORM_FIELD_PREFIX, ""
         )
+        interface = get_object_or_404(ComponentInterface, slug=interface_slug)
         qs = qs.filter(
             dicom_image_set__isnull=not interface.is_dicom_image_kind
         )
-        query = request.GET.get("query-" + prefixed_interface_slug)
+        query = request.GET.get(
+            f"{FlexibleWidgetPrefixes.SEARCH}{interface_slug}_{ImageSearchWidgetSuffixes.INPUT}",
```
chrisvanrun, line 364: Tracing some things and this feels really awkward. It is a composite data name which is partially automatic, partially manual and very 'far away' from the initial from that is the source from it. Easy to break this, especially since it is hard to test.

Looked into how we could simplify this, and getting the POST field here changed is quite simple (we just overwrite the 'name' attribute in the `ImageSearchInputWidget` widget). However, then we'd be messing with the data routing from the main form to the widgets with the final form. We could maybe `disable=True` on the `CharField` in `ImageSearchMultiField`.

What are your thoughts?

app/tests/cases_tests/test_views.py
chrisvanrun, line 1: This needs an addition to the `test_image_search_view` to cover the new query field.

app/grandchallenge/cases/views.py
koopmant: As you say, if we change the name of the widget to something simple and static, we break the data routing; so that not an option. 

> We could maybe disable=True on the CharField in ImageSearchMultiField.

I'm not sure what that achieves?

Regardless, although I agree with you that this is complicated, the code hasn't really changed here. I'm using the same logic that we already had, just with a different name. I can look into it, but it would have to be in a separate PR.

app/tests/cases_tests/test_views.py
koopmant: Yes, good find! I happen to already have that, because I needed to fix that test in my current branch where I replaced the FlexibleImageField.

app/grandchallenge/cases/views.py
chrisvanrun: The disabling of the `CharField` would practically disable the routing of the field: it'll just be set to the initial value. Writing that makes me also realize that will likely mess up the values when you have an invalid form returned =/.

Given that, the search view is not used anywhere else and is sort-of a companion view to the specific widget I think it'll work out.

koopmant: All right. I will also look into the autocomplete widget again later. If I get that to work I can remove this completely.

PR #4485, opened 2025-12-17
Let InterfaceFormFieldsFactory return a fields dictionary

koopmant:
Prepare for the introduction of the flexible dicom image widgets, which will be returned together.

### DIFF
--- a/app/tests/components_tests/test_forms.py
+++ b/app/tests/components_tests/test_forms.py
@@ -45,11 +45,11 @@ def test_interface_form_field_image_queryset_filter():
     upload1.status = UserUpload.StatusChoices.COMPLETED
     upload1.save()
     ci = ComponentInterfaceFactory(kind=ComponentInterface.Kind.PANIMG_IMAGE)
-    field = (
-        InterfaceFormFieldsMixin()
-        .get_fields_for_interface(interface=ci, user=user)
-        .popitem()[1]
+    fields = InterfaceFormFieldsMixin().get_fields_for_interface(
+        interface=ci, user=user
     )
+    assert len(fields) == 1
+    field = fields[f"{INTERFACE_FORM_FIELD_PREFIX}{ci.slug}"]
     assert im1 in field.fields[0].queryset.all()
     assert im2 not in field.fields[0].queryset.all()
     assert upload1 in field.fields[1].queryset.all()


koopmant:
> What about rewriting it to a form mixin and place it in the forms.py. 

Great idea, thanks!

app/tests/cases_tests/test_widget.py
```diff
@@ -204,26 +204,42 @@ def test_flexible_image_widget_prepopulated_value():
...
     civ = ComponentInterfaceValueFactory(interface=ci, image=im)
 
-    field = InterfaceFormFieldFactory(
-        interface=ci, user=user_with_perm, initial=civ
+    field = (
+        InterfaceFormFieldsMixin()
+        .get_fields_for_interface(
+            interface=ci, user=user_with_perm, initial=civ
+        )
+        .popitem()[1]
```
chrisvanrun: This implicitly adds an ordering to the test, while we don't really care about this. I think. Plus it's pretty hard to read.

What about this instead?

```Python
fields = InterfaceFormFieldsMixin().get_fields_for_interface(
     interface=ci, user=user_with_perm, initial=civ
)
assert len(fields) == 1
field = fields[f"{INTERFACE_FORM_FIELD_PREFIX}{ci.slug}"]
```

chrisvanrun (changes requested):
Small suggestion, otherwise it looks good!

koopmant:
> > > What about rewriting it to a form mixin and place it in the forms.py.
> > 
> > 
> > Great idea, thanks!
> 
> Ironically, I read your follow-up PR where you also introduced a mixin =)

Haha, I only just now realized it when I was updating the other PR.

app/tests/cases_tests/test_widget.py
koopmant: That would work. Although I wonder if I should just get the widget that I want directly, instead of using get_fields_for_interface?

chrisvanrun: Would also work, mostly depends on the test I guess?

koopmant: Indeed. I cleaned up the ones that are specifically about a certain field to use that field directly.

chrisvanrun (approved):
LGTM!

PR #4486, opened 2025-12-18
Gracefully fail add_image_to_answer task when answer is None

HarmvZ:
Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/699


chrisvanrun (approved):
LGTM, wonder if we need to add this to a test of sorts, just to have this behaviour covered?

HarmvZ:
I thought about that but a quick search revealed that the task was not really tested so I didn't add it.

PR #4488, opened 2025-12-19
Dicom image kind fields

koopmant:
Add form fields for flexible dicom image form input.

Data validation of the flexible widgets now happens in the InterfaceFormFieldMixin, where we select the data based on the widget choice. 

The form fields are still displayed all together, switching visibility will follow in a later PR. 

Stacked PR after #4485

### DIFF
--- a/app/grandchallenge/components/forms.py
+++ b/app/grandchallenge/components/forms.py
@@ -1,4 +1,5 @@
 import logging
+from enum import Enum
 
 from crispy_forms.helper import FormHelper
 from crispy_forms.layout import ButtonHolder, Layout, Submit
@@ -28,6 +29,7 @@ from grandchallenge.cases.widgets import (
     FlexibleImageWidget,
     ImageSearchWidget,
     ImageSourceChoiceField,
+    ImageWidgetChoices,
 )
 from grandchallenge.components.backends.exceptions import (
     CIVNotEditableException,
@@ -135,11 +137,12 @@ class ContainerImageForm(SaveFormInitMixin, ModelForm):
 
 
 INTERFACE_FORM_FIELD_PREFIX = "__INTERFACE_FIELD__"
-FLEXIBLE_WIDGET_PREFIXES = [
-    "flexible_widget_choice",
-    "flexible_upload",
-    "flexible_search",
-]
+
+
+class FlexibleWidgetPrefixes(Enum):
+    CHOICE = f"flexible_widget_choice{INTERFACE_FORM_FIELD_PREFIX}"
+    UPLOAD = f"flexible_upload{INTERFACE_FORM_FIELD_PREFIX}"
+    SEARCH = f"flexible_search{INTERFACE_FORM_FIELD_PREFIX}"
 
 
 class InterfaceFormFieldsMixin:
@@ -191,16 +194,16 @@ class InterfaceFormFieldsMixin:
             )
             if interface.is_dicom_image_kind:
                 return {
-                    f"flexible_widget_choice{prefixed_interface_slug}": ImageSourceChoiceField(
+                    f"{FlexibleWidgetPrefixes.CHOICE.value}{interface.slug}": ImageSourceChoiceField(
                         current_socket_value=current_socket_value,
                         **kwargs,
                     ),
-                    f"flexible_upload{prefixed_interface_slug}": DICOMUploadField(
+                    f"{FlexibleWidgetPrefixes.UPLOAD.value}{interface.slug}": DICOMUploadField(
                         user=user,
                         label="",
                         required=False,
                     ),
-                    f"flexible_search{prefixed_interface_slug}": ModelChoiceField(
+                    f"{FlexibleWidgetPrefixes.SEARCH.value}{interface.slug}": ModelChoiceField(
                         queryset=image_search_queryset,
                         label="",
                         required=False,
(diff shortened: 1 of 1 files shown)


app/grandchallenge/components/forms.py
```diff
@@ -111,7 +112,86 @@ class Meta:
...
+                        self[field_name].field.required = True
+
+        super().full_clean()
+
+        # Reset `required` to avoid javascript validation.
+        # Items may otherwise get a "Please fill out this field" tooltip
+        # blocking submission. This will lead to issues if this field is no
+        # longer the selected choice. (The widget is then not focusable.)
+        for field_name, required in fields_required.items():
+            self[field_name].field.required = required
```
chrisvanrun: Minor changes and comments for readability, mostly for helping our future selves.

```suggestion
    def full_clean(self):
        # Mark selected widgets as required for validation
        fields_required = {}
        
        try:
            for name in self.fields:
                if name.startswith(
                    "flexible_widget_choice" + INTERFACE_FORM_FIELD_PREFIX
                ):
                    base_name = name[len("flexible_widget_choice") :]
                    bound_field = self[name]
                    
                    widget_fields = {
                        "IMAGE_SEARCH": f"flexible_search{base_name}",
                        "IMAGE_UPLOAD": f"flexible_upload{base_name}",
                    }
                    
                    for choice, field_name in widget_fields.items():
                        if bound_field.data == choice:
                            # Store original required state and temporarily set to required
                            fields_required[field_name] = self[field_name].field.required
                            self[field_name].field.required = True

            super().full_clean()
        finally:
            # Reset `required` to avoid javascript validation.
            # Items may otherwise get a "Please fill out this field" tooltip
            # blocking submission. This will lead to issues if this field is no
            # longer the selected choice. (The widget is then not focusable.)
            for field_name, required in fields_required.items():
                self[field_name].field.required = required
```

app/grandchallenge/components/forms.py
```diff
@@ -111,7 +112,86 @@ class Meta:
...
+    def clean(self):
+        cleaned_data = super().clean()
+
+        keys_to_remove = []
+        data_to_add = {}
+
+        for key in cleaned_data.keys():
+            if any(
+                [
+                    key.startswith(flex_prefix + INTERFACE_FORM_FIELD_PREFIX)
```
chrisvanrun: Would it make sense to do this only once? Where we build the prefix together with the INTERFACE_FORM_FIELD_PREFIX?

app/grandchallenge/components/forms.py
```diff
@@ -111,7 +112,86 @@ class Meta:
...
+                            data_to_add[base_key] = cleaned_data[widget_key]
+                        except KeyError:
+                            pass
+                    else:
+                        if (
+                            widget_type in ["IMAGE_SEARCH", "IMAGE_UPLOAD"]
+                            and widget_key in self.errors
+                        ):
+                            # Ignore errors if it is not the selected choice.
+                            del self._errors[widget_key]
```
chrisvanrun: What sort of errors are we capturing / removing here?

chrisvanrun (commented):
A few questions.

app/grandchallenge/components/forms.py
koopmant: I don't follow, we have multiple flex prefixes. Do you mean we use the interfaceformfield prefix in the `FLEXIBLE_WIDGET_PREFIXES` directly?

app/grandchallenge/components/forms.py
koopmant: Basically any error that the widget could report. We are not interested in those errors if the widget is not the selected choice, because we won't do anything with that data. 

Do you have a suggestion for a clearer comment?

app/grandchallenge/components/forms.py
koopmant: Thanks!

app/grandchallenge/components/forms.py
chrisvanrun: Not sure, I am wondering if we could not prevent the form from generating them in the first place. I.e. remove the related fields.

app/grandchallenge/components/forms.py
chrisvanrun: Yes, something like this but maybe more in line with the style?

```Python
class InterfaceFormFieldsMixin:
    class FIELD_PREFIXES:
        CHOICE = f"flexible_widget_choice{INTERFACE_FORM_FIELD_PREFIX}"
        UPLOAD = f"flexible_upload{INTERFACE_FORM_FIELD_PREFIX}"
        SEARCH = f"flexible_search{INTERFACE_FORM_FIELD_PREFIX}"

    def full_clean(self):
        # Mark selected widgets as required for validation
        fields_required = {}
        for name in self.fields:
            if name.startswith(self.FIELD_PREFIXES.CHOICE):
                slug = name[len(self.FIELD_PREFIXES.CHOICE) :]
                bound_field = self[name]
                for choice, field_name in {
                    "IMAGE_SEARCH": f"{self.FIELD_PREFIXES.SEARCH}{slug}",
                    "IMAGE_UPLOAD": f"{self.FIELD_PREFIXES.UPLOAD}{slug}",
                }.items():
                    if bound_field.data == choice:
                        fields_required[field_name] = self[
                            field_name
                        ].field.required
                        self[field_name].field.required = True

        super().full_clean()

        # Reset `required` to avoid javascript validation.
        # Items may otherwise get a "Please fill out this field" tooltip
        # blocking submission. This will lead to issues if this field is no
        # longer the selected choice. (The widget is then not focusable.)
        for field_name, required in fields_required.items():
            self[field_name].field.required = required

    def clean(self):

        keys_to_remove = []
        data_to_add = {}

        for key in cleaned_data.keys():
            if any([key.startswith(prefix) for prefix in self.FIELD_PREFIXES]):
                keys_to_remove.append(key)

            if key.startswith(self.FIELD_PREFIXES.CHOICE):
                # Get the choice from the data because if it is "IMAGE_SELECTED"
                # the cleaned data becomes the current socket value
                choice = self[key].data
                slug = key[len(self.FIELD_PREFIXES.CHOICE) :]
                widget_fields = {
                    "IMAGE_SELECTED": key,
                    "IMAGE_SEARCH": f"{self.FIELD_PREFIXES.SEARCH}{slug}",
                    "IMAGE_UPLOAD": f"{self.FIELD_PREFIXES.UPLOAD}{slug}",
                }

                for widget_type, widget_key in widget_fields.items():
                    if choice == widget_type:
                        try:
                            data_to_add[
                                f"{INTERFACE_FORM_FIELD_PREFIX}{slug}"
                            ] = cleaned_data[widget_key]
                        except KeyError:
                            pass
                    else:
                        if (
                            widget_type in ["IMAGE_SEARCH", "IMAGE_UPLOAD"]
                            and widget_key in self.errors
                        ):
                            # Ignore errors if it is not the selected choice.
                            del self._errors[widget_key]

        cleaned_data.update(data_to_add)

        for key in keys_to_remove:
            del cleaned_data[key]

        return cleaned_data
```

app/grandchallenge/components/forms.py
koopmant: I tried that first, but it is more complicated. You would want to do this before any cleaning, get the selected choice first, (which needs to be cleaned first), then unbind those form fields somehow to avoid validation. It became very messy and it didn't work yet. I'm sure it's possible, but this seemed far simpler.

chrisvanrun: OK - simpler is my cup of tea, so let us stick with this then!

app/grandchallenge/components/forms.py
koopmant: Thanks, I've added an Enum for it and also used the `ImageWidgetChoices` instead of using the hardcoded values for those.

koopmant:
> I do wonder if we should, at one point, add a docstring that explains a bit what is going on here with fields, slugs et cetera.

Good point. I'll take a look at that before the final PR is merged.

PR #4489, opened 2025-12-22
Move reservation check to the CIVData validation

chrisvanrun:
Alternative to:
- #4480


koopmant (approved):
LGTM!

Smart to keep it to just this. We can extend this later.

PR #4495, opened 2026-01-02
Challenge Request form: use 'case' in place of 'image' and make output size a budget factor

chrisvanrun:
This PR renames 'images' to the more general term 'case'. 

In addition, it adds a field to the request form to describe the size of the prediction per case.

## Showcase in form
image

By default, which includes all the existing requests, the `average_size_of_prediction_in_mb` is set to `0`.


koopmant:
Let's combine this with #4264, or at least do it afterwards. Then this will possible for multiple tasks as well.

PR #4496, opened 2026-01-05
Invalidate email as first or last name

chrisvanrun:
Closes: https://github.com/DIAGNijmegen/rse-grand-challenge/issues/4254

I don't think we should disable email-like usernames, but instead add a quick contain check that covers an exact copy or a sneaky `'foo@example.test+1'` attempt.

### DIFF
--- a/app/grandchallenge/profiles/forms.py
+++ b/app/grandchallenge/profiles/forms.py
@@ -2,6 +2,7 @@ from crispy_forms.helper import FormHelper
 from crispy_forms.layout import Submit
 from django import forms
 from django.core.exceptions import ValidationError
+from django.core.validators import EmailValidator
 from django.forms import CheckboxInput, Select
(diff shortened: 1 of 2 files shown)


koopmant:
> I don't think we should disable email-like usernames

I think you mean names, not usernames. Because we already disable email like usernames. But why shouldn't we also disable email like names, I don't see how they could be valid?

chrisvanrun:
> LGTM. Technically the names can now still be invalid email addresses, but this should be good enough.

Yes, but adding a lot of checks increases the changes that some odd name would false positive on it. I thought just making '@' illegal, but then I remembered some South-African -American billionaire having fun in the naming department...

koopmant:
> I thought just making '@' illegal, but then I remembered some South-African -American billionaire having fun in the naming department...

Haha, even he didn't use the @ though, did he?

PR #4499, opened 2026-01-06
List size of component images and models in admin

koopmant:
Allow quick inspection of storage sizes.

### DIFF
--- a/app/grandchallenge/algorithms/admin.py
+++ b/app/grandchallenge/algorithms/admin.py
@@ -284,9 +284,10 @@ class AlgorithmModelAdmin(admin.ModelAdmin):
         "model",
     )
 
+    @admin.display(ordering="size_in_storage")
     def naturalsize_in_storage(self, obj):
         return humanize.naturalsize(obj.size_in_storage)
-naturalsize_in_storage.admin_order_field = "size_in_storage"
+
 
 @admin.register(AlgorithmInterface)
 class AlgorithmInterfaceAdmin(admin.ModelAdmin):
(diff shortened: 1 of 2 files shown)


app/grandchallenge/algorithms/admin.py
```diff
@@ -277,6 +284,9 @@ class AlgorithmModelAdmin(admin.ModelAdmin):
         "model",
     )
 
+    def naturalsize_in_storage(self, obj):
+        return humanize.naturalsize(obj.size_in_storage)
+
```
chrisvanrun: Might be nice to be able to sort this, albeit in practice it will likely all be GBs: https://books.agiliq.com/projects/django-admin-cookbook/en/latest/sorting_calculated_fields.html

Same holds for the other fields.

```suggestion
naturalsize_in_storage.admin_order_field = "size_in_storage"
```

app/grandchallenge/algorithms/admin.py
```diff
@@ -277,6 +284,10 @@ class AlgorithmModelAdmin(admin.ModelAdmin):
         "model",
     )
 
+    @admin.display(ordering="size_in_storage")
```
chrisvanrun, line 287: Ah yeah, that was the correct format!

chrisvanrun (approved):
LGTM!

app/grandchallenge/algorithms/admin.py
koopmant: precommit fixed it for me automatically 😄

PR #4502, opened 2026-01-09
Enable more phases for budget 1: Add and deprecate fields

koopmant:


chrisvanrun (approved):
LGTM, I am wondering what will happen with the deprecated fields in the follow up PRs =)

koopmant:
> I am wondering what will happen with the deprecated fields in the follow up PRs =)

Haha. I have a PR standby to clean them up. But that will have to be done in a separate deployment, so for now they'll have to stay.

PR #4505, opened 2026-01-09
Enable more phases for budget 4: Update challenge request budget update form

koopmant:


app/grandchallenge/challenges/forms.py
```diff
@@ -662,10 +628,92 @@ def __init__(self, *args, **kwargs):
         )
         self.helper.layout.append(Submit("save", "Save"))
 
-    def clean_algorithm_selectable_gpu_type_choices(self):
-        data = self.cleaned_data.get(
-            "algorithm_selectable_gpu_type_choices", []
+    def clean(self):
+        cleaned_data = super().clean()
+
+        if not self._clean_all_fields_nonempty_cleaned_data(cleaned_data):
```
chrisvanrun: Would it not make more sense to change this to validation per field rather than a form wide error?

app/tests/challenges_tests/test_forms.py
```diff
@@ -173,37 +172,98 @@ def test_budget_update_form():
...
-        "phase_1_number_of_submissions_per_team": 10,
-        "phase_2_number_of_submissions_per_team": 1,
-        "phase_1_number_of_test_images": 100,
-        "phase_2_number_of_test_images": 500,
-        "number_of_tasks": 1,
+        "task_ids": "[1, 2]",
+        "algorithm_maximum_settable_memory_gb_for_tasks": "[32, 32]",
+        "algorithm_selectable_gpu_type_choices_for_tasks": '[["", "T4"],["", "A10G", "T4"]]',
+        "average_size_test_case_mb_for_tasks": "[10, 100]",
+        # "inference_time_average_minutes_for_tasks": "[5, 10]",
```
chrisvanrun: ?

app/tests/challenges_tests/test_forms.py
```diff
@@ -173,37 +172,98 @@ def test_budget_update_form():
...
+        "number_of_submissions_per_team_for_phases": "[10, 1, 10, 1]",
+        "number_of_test_cases_for_phases": "[3, 100, 3, 100]",
     }
     form2 = ChallengeRequestBudgetUpdateForm(
         data=data2, instance=challenge_request
     )
     assert form2.is_valid()
+
+
+@pytest.mark.parametrize(
```
chrisvanrun, line 219: Love the comments next to the introduced defect.

app/grandchallenge/challenges/forms.py
chrisvanrun: Oh, I see why you have done this. Field validation had already failed and we are just checking here... IMHO we should not raise here but just not perform the other validations here.

app/tests/challenges_tests/test_forms.py
koopmant: I wanted to make it easy to see which one is missing... You'd prefer if I remove it?

app/tests/challenges_tests/test_forms.py
koopmant: Thanks! I needed them to keep track of things haha

app/grandchallenge/challenges/forms.py
koopmant: Hmm, yeah. It's also strange that I let `_clean_all_fields_nonempty_cleaned_data()` (what a name haha) return a value about whether all fields are nonempty. 

Although I think the general logic in clean is to raise or return cleaned data. And returning cleaned data here seems weird. Maybe I should just check if there are errors at this point and then return cleaned data knowing for sure that there are validation errors.

koopmant: > and then return cleaned data knowing for sure that there are validation errors.

Never mind, I see the return value may also be None.

koopmant: It's cleaned up. Thanks!

app/tests/challenges_tests/test_forms.py
koopmant: I removed it

koopmant:
I cancelled the CI workflow, because there will be failing tests anyway.

app/tests/challenges_tests/test_forms.py
chrisvanrun: Was mostly curious about it, but removing it is fine.

app/tests/challenges_tests/test_forms.py
chrisvanrun: Sid track: I've found the pattern where you add the actual description to the case and then do an:

```Python
assert not form.is_valid(), description
```

Also, very neat. That gives you an instant feedback which case broke instead of having to count the number of cases.

app/grandchallenge/challenges/forms.py
```diff
@@ -662,10 +628,87 @@ def __init__(self, *args, **kwargs):
...
+                "All fields defining phases must be of equal length."
+            )
+        self._clean_later_phases_not_more_teams_or_submissions(
+            task_id_for_phases, cleaned_data
         )
-        if HTMX_BLANK_CHOICE_KEY in data:
-            data[data.index(HTMX_BLANK_CHOICE_KEY)] = ""
-        return data
+
+        return cleaned_data
```
chrisvanrun: IMO this is cleaner, prevents the unwanted sentinel check, allows for other clean validation errors and gives a good key error if form data goes missing (instead of a red-herring down in the methods):

```suggestion
        cleaned_data = super().clean()

        if not self.errors:
            task_ids = cleaned_data["task_ids"]
            task_id_for_phases = cleaned_data["task_id_for_phases"]

            self._clean_task_lists_equal_length(cleaned_data)

            self._clean_task_id_for_phases(task_ids, task_id_for_phases)

            if not self._clean_phases_lists_equal_length(
                task_id_for_phases, cleaned_data
            ):
                raise ValidationError(
                    "All fields defining phases must be of equal length."
                )
            self._clean_later_phases_not_more_teams_or_submissions(
                task_id_for_phases, cleaned_data
            )

        return cleaned_data
```

app/grandchallenge/challenges/forms.py
```diff
@@ -662,10 +628,87 @@ def __init__(self, *args, **kwargs):
...
+        self, task_id_for_phases, cleaned_data
+    ):
+        all_phases_list_equal_length = True
+
+        for field_name in (
+            "number_of_teams_for_phases",
+            "number_of_submissions_per_team_for_phases",
+            "number_of_test_cases_for_phases",
+        ):
+            field_value = cleaned_data.get(field_name)
```
chrisvanrun:
```suggestion
            field_value = getattr(cleaned_data, field_name)
```

app/grandchallenge/challenges/forms.py
chrisvanrun: Also, related, but I think the `HTMX_BLANK_CHOICE_KEY` can just be straight up removed?

chrisvanrun (changes requested):
Small code style issue.

app/grandchallenge/challenges/forms.py
koopmant: > Also, related, but I think the HTMX_BLANK_CHOICE_KEY can just be straight up removed?

Yes, it already was removed. How are you still seeing it?

koopmant: > prevents the unwanted sentinel check

I'm a bit lost in translation here

koopmant: > gives a good key error if form data goes missing (instead of a red-herring down in the methods)

I'm missing the point here as well. If we return None, `cleaned_data` is never overwritten so this code is equivalent.

app/grandchallenge/challenges/forms.py
jmsmkn: Why?

koopmant: Why not use `get`? `cleaned_data` is a dict.

koopmant: Wait, this wouldn't even work

app/grandchallenge/challenges/forms.py
chrisvanrun: It is still defined as a constant `HTMX_BLANK_CHOICE_KEY` in the top of the module.

koopmant: I think we're still using `HTMX_BLANK_CHOICE_KEY ` in the challenge request form itself. Let me check

koopmant: Ah no, indeed. I'll remove it. Thanks! (I was confused by the generated diff above here.)

chrisvanrun: > I'm missing the point here as well. If we return None, cleaned_data is never overwritten so this code is equivalent.

Is it equivalent? Any `super().clean` errors will be picked up. Plus, the use of `if` fall throughs are discouraged in the code guidelines.

app/grandchallenge/challenges/forms.py
chrisvanrun: You are right! That was a silly mistake of mine. It should just be a `cleaned_data[field_name]`. Practically, it will not matter since the fields are required and the `BaseForm._clean_fields` will pick up if it is missing. So at this point in the clean it should really be present.

I made it a style note, as the `.get()` suggests that you take into account that it _might_ be missing here.

app/grandchallenge/challenges/forms.py
koopmant: > Any super().clean errors will be picked up. 

You're right, my bad. 

> Plus, the use of if fall throughs are discouraged in the code guidelines.

This one always gets me. Especially in cases like this where the if-statement contains a negation. 
`if errors -> stop`, seems much clearer to me. So in this case i'd make an exception.

koopmant: > So in this case I'd make an exception. 

On second thought, let's go for consistency.

app/grandchallenge/challenges/forms.py
koopmant: Ah good point.

app/grandchallenge/challenges/forms.py
chrisvanrun: > Ah no, indeed. I'll remove it. Thanks! (I was confused by the generated diff above here.)

Two of us!

app/grandchallenge/challenges/forms.py
```diff
@@ -689,7 +684,10 @@ def _clean_phases_lists_equal_length(
                 )
                 all_phases_list_equal_length = False
 
-        return all_phases_list_equal_length
+        if not all_phases_list_equal_length:
+            raise ValidationError(
+                "All fields defining phases must be of equal length."
+            )
```
chrisvanrun, line 712: Nice!

PR #4508, opened 2026-01-09
Dicom widget js

koopmant:
Add javascript to switch visibility of the flexible widgets. 

Stacked PR after #4488

### DIFF
--- a/app/grandchallenge/cases/static/cases/js/source_choice_widget.mjs
+++ b/app/grandchallenge/cases/static/cases/js/source_choice_widget.mjs
@@ -36,8 +36,6 @@ const initChoiceWidgets = () => {
     }
 };
 
-// Run on DOM changes (needed for re-rendered forms)
-new MutationObserver(initChoiceWidgets).observe(document.body, {
-    childList: true,
-    subtree: true,
-});
+document.addEventListener("htmx:load", initChoiceWidgets);
+
+initChoiceWidgets();


app/grandchallenge/cases/static/cases/js/source_choice_widget.mjs
```diff
@@ -0,0 +1,40 @@
...
+                    "d-none",
+                    this.value !== "IMAGE_UPLOAD",
+                );
+            });
+            el.dataset.eventListenerAdded = "1";
+        }
+    }
+};
+
+// Run on DOM changes (needed for re-rendered forms)
```
chrisvanrun: I am worried about performance here. This gets run so often. Wonder if there is someways we could scope this better? For instance, every key press when searching for an image or socket value fires this.

app/grandchallenge/cases/static/cases/js/source_choice_widget.mjs
```diff
@@ -0,0 +1,40 @@
...
+        );
+
+        if (!searchWidget || !uploadWidget) {
+            continue;
+        }
+
+        searchWidget.classList.toggle("d-none", el.value !== "IMAGE_SEARCH");
+        uploadWidget.classList.toggle("d-none", el.value !== "IMAGE_UPLOAD");
+
+        if (!el.dataset.eventListenerAdded) {
```
chrisvanrun, line 23: I think this can be checked earlier in the per element loop and renamed into something like `choiceWidgetInitialized`.

app/grandchallenge/cases/static/cases/js/source_choice_widget.mjs
```diff
@@ -0,0 +1,40 @@
...
+        );
+        const uploadWidget = document.getElementById(
+            `div_${el.id.replace(/^id_flexible_widget_choice/, "id_flexible_upload")}`,
+        );
+
+        if (!searchWidget || !uploadWidget) {
+            continue;
+        }
+
+        searchWidget.classList.toggle("d-none", el.value !== "IMAGE_SEARCH");
```
chrisvanrun, line 19: The toggle + force is elegant here.

app/grandchallenge/cases/static/cases/js/source_choice_widget.mjs
chrisvanrun: What if we'd simply hook it into htmx load, that would insert new parts to the DOM?

```
document.addEventListener("htmx:load", initDisableAfterSubmit);
```

koopmant: I thought of adding something like the mutationobserver below, but it doesn't work when the form is re-rendered when it's invalid. 

I've also investigated the timing and a call to initChoiceWidget was on average 0.0018 ms (100000 calls took 180 ms), while a call to the observer was on average 0.0004 ms. So it's faster, but you won't notice the difference anyway. I'm proposing to leave it like this. 

```javascript
const observer = new MutationObserver((mutations) => {
    for (const mutation of mutations) {
        if (mutation.type !== "childList") continue;

        for (const node of mutation.addedNodes) {
            if (
                node.nodeType === Node.ELEMENT_NODE &&
                (
                    node.matches?.("select[id^='id_flexible_widget_choice__INTERFACE_FIELD__']") ||
                    node.querySelector?.("select[id^='id_flexible_widget_choice__INTERFACE_FIELD__']")
                )
            ) {
                initChoiceWidgets();
                return;
            }
        }
    }
});

observer.observe(document.body, {
    childList: true,
    subtree: true,
});
```

koopmant:
> But afaik script-tag elements are purposefully NOT executed when new elements are inserted.

Yes, I believe so too.

app/grandchallenge/cases/static/cases/js/source_choice_widget.mjs
koopmant: It cannot. When I check it earlier, I skip the toggling for the visibility when the form is re-rendered on invalid input. The event listener is already set at that point, but we still need to toggle visibility manually after re-rendering.

app/grandchallenge/cases/static/cases/js/source_choice_widget.mjs
koopmant: > What if we'd simply hook it into htmx load, that would insert new parts to the DOM?
> 
> ```
> document.addEventListener("htmx:load", initDisableAfterSubmit);
> ```

We'd still need something for when the form is re-rendered though... I'm not sure how to catch that.

chrisvanrun: When is a form re-rendered? Not sure what scenario you are describing there, sorry.

chrisvanrun: Ah, on invalid input! Edit: wait, when what is invalid? Client side validation or POST server validation?

koopmant: Yes indeed! POST server validation.

koopmant: You were completely right about hooking into htmx events and running the `initChoiceWidgets` once when the javascript module is loaded. That works and I chose to go with that in the end.

PR #4512, opened 2026-01-15
Retain selected search result

koopmant:
Forward the form data with the search term and the selected image to a new image search widget.

This is needed because this is otherwise lost when there are validation errors on the form. 

In the FlexibleImageWidget (currently still used for non-dicom images) this selection is preserved by placing it as one of the choices in the "source choice widget" when the form is rendered after validation errors:
Screenshot 2026-01-15 at 18 26 20

We now preserve it in the search widget itself, with the used search term as well:
Screenshot 2026-01-15 at 18 22 24

### DIFF
--- a/app/grandchallenge/cases/widgets.py
+++ b/app/grandchallenge/cases/widgets.py
@@ -293,6 +293,9 @@ class ImageSearchInputWidget(TextInput):
         attrs["placeholder"] = "Search by pk or image name"
         context = super().get_context(name, value, attrs)
         css_class = context["widget"]["attrs"].get("class", "")
+        # When the MultiField marks the data invalid, the is-invalid class is
+        # added to all subwidgets; however, the search input is never "invalid"
+        # because that data will not be processed.
         context["widget"]["attrs"]["class"] = css_class.replace(
             "is-invalid", ""
         )


app/grandchallenge/cases/widgets.py
```diff
@@ -330,3 +330,85 @@ def compress(self, values: list[str, QuerySet[UserUpload]]):
...
+        return context
+
+
+class ImageSearchMultiWidget(MultiWidget):
+    template_name = "cases/image_search_multi_widget.html"
+
+    def __init__(self, attrs=None, prefixed_interface_slug=None):
+        widgets = {
+            IMAGE_SEARCH_WIDGET_SUFFIXES[0]: ImageSearchInputWidget(),
+            IMAGE_SEARCH_WIDGET_SUFFIXES[1]: ImageSearchChoiceWidget(
```
chrisvanrun, line 325: (nit) Given they are ever only used once, why not directly code them here?

app/grandchallenge/cases/widgets.py
```diff
@@ -330,3 +330,85 @@ def compress(self, values: list[str, QuerySet[UserUpload]]):
...
+        context["prefixed_interface_slug"] = self.prefixed_interface_slug
+        return context
+
+    def decompress(self, value):
+        if value:
+            return value
+        return ["", ""]
+
+
+class ImageSearchMultiField(MultiValueField):
```
chrisvanrun, line 343: (nit) should this not go to a `fields.py`?

app/grandchallenge/cases/widgets.py
```diff
@@ -330,3 +330,85 @@ def compress(self, values: list[str, QuerySet[UserUpload]]):
...
+
+IMAGE_SEARCH_WIDGET_SUFFIXES = ["search-term", "object-list"]
+
+
+class ImageSearchInputWidget(TextInput):
+    def get_context(self, name, value, attrs):
+        attrs["placeholder"] = "Search by pk or image name"
+        context = super().get_context(name, value, attrs)
+        css_class = context["widget"]["attrs"].get("class", "")
+        context["widget"]["attrs"]["class"] = css_class.replace(
```
chrisvanrun, line 299: Might be good to add a comment here on why we don't allow visibility of invalid here.

chrisvanrun (changes requested):
Generally looks good and a far better flow! A few nitpicks and I have the feeling we are missing a test that covers the uploading of an image, but I think that is already covered by another test.

app/grandchallenge/cases/widgets.py
koopmant: Yes, good point. A `form_fields.py` actually. I'll create it and move the other fields there in a cleanup PR.

app/grandchallenge/cases/widgets.py
koopmant: I'll use these later in some tests when I start using this to replace the `FlexibleImageField`.

PR #4514, opened 2026-01-19
Flexible dicom widget cleanup

koopmant:


app/grandchallenge/cases/widgets.py
```diff
@@ -321,8 +293,8 @@ class ImageSearchMultiWidget(MultiWidget):
 
     def __init__(self, attrs=None, prefixed_interface_slug=None):
         widgets = {
-            IMAGE_SEARCH_WIDGET_SUFFIXES[0]: ImageSearchInputWidget(),
-            IMAGE_SEARCH_WIDGET_SUFFIXES[1]: ImageSearchChoiceWidget(
+            ImageSearchWidgetSuffixes.INPUT.value: ImageSearchInputWidget(),
```
chrisvanrun, line 296: Just a general note, if we are not ever reading these values back (i.e. parse a random string into the enum) we can drop the `Enum` class which would absolve us from having to do `.value` every time we use it in an f-string.

koopmant: Yes, the .value is really annoying. If I drop the Enum completely, it's no longer an immutable set with unique values. However, I just found that if we use `StrEnum`, we can use it in f-strings without having to do `.value`.

koopmant: Thanks! It looks much better now

PR #4515, opened 2026-01-20
Add `created` index to ComponentJob

chrisvanrun:
The Job admin page list is [quite slow](https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/661). Default ordering is by created, however the existing index is a composite of  `status+created`.

Some quick ad hoc testing locally shows that this reduced the initial LIMIT list query from 22 ms to 0.6 ms on a 10k item table.


jmsmkn (approved):
LGTM!

PR #4518, opened 2026-01-21
Add event listener to stable ancestor

koopmant:
The form element is replaced after server side validation errors, so if the event listener was added to the original form element, the validation ui on new extra sockets is no longer invoked afterwards. We now add the listener to the document instead.


koopmant:
To clarify, this message was not getting displayes on newly added sockets for forms that has already returned with validation errors:
Screenshot 2026-01-21 at 17 06 20

PR #4519, opened 2026-01-23
Flexible fields for image kinds

koopmant:
Use the new flexible fields for all image kinds.

### DIFF
--- a/app/tests/cases_tests/test_widget.py
+++ b/app/tests/cases_tests/test_widget.py
@@ -230,35 +230,45 @@ def test_flexible_image_widget_prepopulated_value():
 @pytest.mark.django_db
 def test_image_upload_field_validation():
     user = UserFactory()
-    upload1 = UserUploadFactory(creator=user)
-    upload1.status = UserUpload.StatusChoices.COMPLETED
-    upload1.save()
-    upload2 = UserUploadFactory(creator=user)
-    upload2.status = UserUpload.StatusChoices.COMPLETED
-    upload2.save()
-    upload3 = UserUploadFactory()
-    upload3.status = UserUpload.StatusChoices.COMPLETED
-    upload3.save()
-    upload4 = UserUploadFactory(creator=user)
     ci = ComponentInterfaceFactory(kind=ComponentInterface.Kind.PANIMG_IMAGE)
     field = InterfaceFormFieldsMixin().get_fields_for_interface(
         interface=ci, user=user
     )[f"{FlexibleWidgetPrefixes.UPLOAD.value}{ci.slug}"]
-    data = [str(upload1.pk), str(upload2.pk)]
-    data_from_other_user = [str(upload3.pk)]
-    data_from_pending_upload = [str(upload4.pk)]
 
+    # Normal case: two uploads from the user with completed status
+    upload1, upload2 = UserUploadFactory.create_batch(
+        2,
+        creator=user,
+    )
+    upload1.status = UserUpload.StatusChoices.COMPLETED
+    upload1.save()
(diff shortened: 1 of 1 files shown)


app/tests/cases_tests/test_widget.py
```diff
@@ -223,6 +227,40 @@ def test_flexible_image_widget_prepopulated_value():
...
+
+    assert cleaned_data.count() == 2
+    assert upload1 in cleaned_data.all()
+    assert upload2 in cleaned_data.all()
+
+    with pytest.raises(ValidationError):
+        field.clean(data_from_other_user)
+
+    with pytest.raises(ValidationError):
+        field.clean(data_from_pending_upload)
```
chrisvanrun: This test was a bit messy so for my own reviewing I rewrote it a bit. Might as well help future readers by including it.

```suggestion
def test_image_upload_field_validation():
    user = UserFactory()
    ci = ComponentInterfaceFactory(kind=ComponentInterface.Kind.PANIMG_IMAGE)
    field = InterfaceFormFieldsMixin().get_fields_for_interface(
        interface=ci, user=user
    )[f"{FlexibleWidgetPrefixes.UPLOAD.value}{ci.slug}"]

    # Normal case: two uploads from the user with completed status
    upload1, upload2 = UserUploadFactory.create_batch(
        2,
        creator=user,
    )
    upload1.status = UserUpload.StatusChoices.COMPLETED
    upload1.save()
    upload2.status = UserUpload.StatusChoices.COMPLETED
    upload2.save()
    data = [str(upload1.pk), str(upload2.pk)]
    cleaned_data = field.clean(data)

    assert cleaned_data.count() == 2
    assert upload1 in cleaned_data.all()
    assert upload2 in cleaned_data.all()

    # Upload from another user
    other_user_upload = UserUploadFactory(
        creator=UserFactory(),
    )
    other_user_upload.status = UserUpload.StatusChoices.COMPLETED
    other_user_upload.save()
    data_from_other_user = [str(other_user_upload.pk)]
    with pytest.raises(ValidationError):
        field.clean(data_from_other_user)

    # Upload with non-completed status
    non_completed_upload = UserUploadFactory(
        creator=user,
    )
    assert non_completed_upload.status != UserUpload.StatusChoices.COMPLETED
    data_from_non_completed_upload = [str(non_completed_upload.pk)]
    with pytest.raises(ValidationError):
        field.clean(data_from_non_completed_upload)
```

app/tests/reader_studies_tests/test_views.py
```diff
@@ -796,41 +797,36 @@ def test_add_display_set_update_when_disabled(client):
...
+    for ci in (ci_value, ci_file, ci_image, ci_dicom):
+        response = get_view_for_user(
+            viewname="reader-studies:display-set-interfaces-create",
+            client=client,
+            reverse_kwargs={"pk": ds.pk, "slug": rs.slug},
+            data={"interface": str(ci.pk)},
+            user=u1,
+        )
+        assert response.status_code == 200
+        assert not response.context["form"].is_bound
```
chrisvanrun, line 829: Out of curiosity, why is this unbound at this point? Should it not be bound since there is data?

app/tests/cases_tests/test_widget.py
koopmant: I see your point. Thanks for rewriting!

app/tests/reader_studies_tests/test_views.py
koopmant: This is the view that returns the form fields for a new socket after the user has seached it. (After clicking the "Add value for new socket" button.) The data is used to select the right socket, but the form fields returned should be unbound.

PR #4520, opened 2026-01-26
Flexible fields for file kinds

koopmant:
Use the new flexible fields for file kinds.

### DIFF
--- a/app/tests/components_tests/test_form_fields.py
+++ b/app/tests/components_tests/test_form_fields.py
@@ -476,42 +476,6 @@ def test_file_search_field_validation_with_archive_items():
 
 
 @pytest.mark.django_db
-def test_file_source_select_prepopulated_value_algorithm_job():
-    creator, user = UserFactory.create_batch(2)
-    ci = ComponentInterfaceFactory(kind=FuzzyChoice(InterfaceKinds.file))
-    civ = ComponentInterfaceValueFactory(
-        interface=ci, file=factory.django.FileField()
-    )
-    job = AlgorithmJobFactory(creator=creator, time_limit=60)
-    job.inputs.set([civ])
-
-    field = FileSourceChoiceField(current_socket_value=civ)
-
-    assert field.current_socket_value == civ
-    assert field.choices == [
-        ("CURRENT", civ.title),
-        ("SEARCH", "Select an existing file"),
-        ("UPLOAD", "Upload a new file"),
-    ]
-    assert field.clean(SourceChoices.CURRENT.value) == civ
-    with pytest.raises(ValidationError, match="This field is required."):
-        field.clean(FileSourceChoices.UNDEFINED.value)
-
-    field = FileSourceChoiceField()
-
-    assert field.choices == [
-        ("", "Choose data source..."),
-        ("SEARCH", "Select an existing file"),
-        ("UPLOAD", "Upload a new file"),
-    ]
-    assert field.current_socket_value is None
-    with pytest.raises(ValidationError, match="Select a valid choice."):
-        field.clean(SourceChoices.CURRENT.value)
-    with pytest.raises(ValidationError, match="This field is required."):
-        field.clean(FileSourceChoices.UNDEFINED.value)
-
-
-@pytest.mark.django_db
 def test_file_source_select_prepopulated_value_display_set():
     editor, user = UserFactory.create_batch(2)
     ci = ComponentInterfaceFactory(kind=FuzzyChoice(InterfaceKinds.file))


app/grandchallenge/components/form_fields.py
```diff
@@ -111,3 +130,84 @@ def compress(self, values):
...
+                    FileSourceChoices.UNDEFINED.label,
+                )
+            )
+            choices.insert(
+                0,
+                (
+                    SourceChoices.CURRENT.value,
+                    current_socket_value.title,
+                ),
+            )
```
chrisvanrun, line 163: I think this entails we still can't detach CIVs from a CIVset (DisplaySet/ArchiveItem). Right?

app/tests/algorithms_tests/test_forms.py
```diff
@@ -251,8 +249,8 @@ def try_create_algorithm():
                 f'name="{FlexibleWidgetPrefixes.CHOICE}some-medical-image"',
                 f'name="{FlexibleWidgetPrefixes.UPLOAD}some-medical-image_{DICOMUploadWidgetSuffixes.NAME}"',
                 f'name="{FlexibleWidgetPrefixes.UPLOAD}some-medical-image_{DICOMUploadWidgetSuffixes.UPLOADS}"',
-                f'name="{FlexibleWidgetPrefixes.SEARCH}some-medical-image_{ImageSearchWidgetSuffixes.INPUT}"',
-                f'name="{FlexibleWidgetPrefixes.SEARCH}some-medical-image_{ImageSearchWidgetSuffixes.CHOICE}"',
+                f'name="{FlexibleWidgetPrefixes.SEARCH}some-medical-image_{SearchWidgetSuffixes.INPUT}"',
+                f'name="{FlexibleWidgetPrefixes.SEARCH}some-medical-image_{SearchWidgetSuffixes.CHOICE}"',
```
chrisvanrun, line 253: Way easier to rename if it's an Enum =)

app/tests/components_tests/test_form_fields.py
```diff
@@ -391,3 +397,189 @@ def test_flexible_file_widget_prepopulated_value_user_upload():
...
+    archive_item_without_perm = ArchiveItemFactory(archive=a2)
+    archive_item_with_perm.values.set([civ1])
+    archive_item_without_perm.values.set([civ2])
+    field = InterfaceFormFieldsMixin().get_fields_for_interface(
+        interface=ci, user=user
+    )[f"{FlexibleWidgetPrefixes.SEARCH}{ci.slug}"]
+
+    assert field.clean(["", f"{civ1.pk}"]) == civ1
+    with pytest.raises(ValidationError):
+        field.clean(["", f"{civ2.pk}"])
```
chrisvanrun, line 475: I like the just plain repeat of the pattern here instead of trying something complex to cover what could be considered duplication. Albeit it does get boring to read at this point =)

app/tests/components_tests/test_form_fields.py
```diff
@@ -391,3 +397,189 @@ def test_flexible_file_widget_prepopulated_value_user_upload():
...
+def test_file_source_select_prepopulated_value_algorithm_job():
+    creator, user = UserFactory.create_batch(2)
+    ci = ComponentInterfaceFactory(kind=FuzzyChoice(InterfaceKinds.file))
+    civ = ComponentInterfaceValueFactory(
+        interface=ci, file=factory.django.FileField()
+    )
+    job = AlgorithmJobFactory(creator=creator, time_limit=60)
+    job.inputs.set([civ])
+
+    field = FileSourceChoiceField(current_socket_value=civ)
```
chrisvanrun: I had to think when this would occur, but I think I am correct in saying: validation errors?

app/tests/components_tests/test_form_fields.py
```diff
@@ -391,3 +397,189 @@ def test_flexible_file_widget_prepopulated_value_user_upload():
...
+
+    field = FileSourceChoiceField()
+
+    assert field.choices == [
+        ("", "Choose data source..."),
+        ("SEARCH", "Select an existing file"),
+        ("UPLOAD", "Upload a new file"),
+    ]
+    assert field.current_socket_value is None
+    with pytest.raises(ValidationError, match="Select a valid choice."):
```
chrisvanrun, line 509: Good to use of `match=` here!

app/tests/reader_studies_tests/test_forms.py
```diff
@@ -1440,46 +1451,65 @@ def test_display_set_update_form(form_class):
...
+    ci_str = ComponentInterfaceFactory(kind=InterfaceKindChoices.STRING)
+    QuestionFactory(
+        reader_study=rs, answer_type=AnswerType.TEXT, interface=ci_str
     )
-    QuestionFactory(reader_study=rs, answer_type=AnswerType.TEXT, interface=ci)
     del rs.linked_component_interfaces
+    expected_fields[f"{INTERFACE_FORM_FIELD_PREFIX}{ci_str.slug}"] = (
+        CharField,
+        TextInput,
+    )
```
chrisvanrun, line 1502: Honestly, I am not fond of the tuple constructs. They are hard to read. Especially this line had me go: why does a `String` CI have two widgets/fields, only to realize they are meant to be the expected types of the field and widget in respective order.

What are your thoughts?

app/grandchallenge/components/form_fields.py
koopmant: Oh good point. I need to look into this because I believe we can detach CIV's if it's a file kind...

app/tests/algorithms_tests/test_forms.py
koopmant: Yes! 🤩

app/tests/components_tests/test_form_fields.py
koopmant: Duly noted, I'll keep that in mind! 

(In this case trying to generalize it would be very hard anyway. That is because of the many forms of socket values, see `get_component_interface_values_for_user`.)

app/tests/components_tests/test_form_fields.py
koopmant: Good question, and I'm not yet sure. When the form comes back with errors, the job inputs have not been set yet. With these new fields, the form data is preserved; so whatever the user did (select existing image or upload a new one) is what will be shown again. We don't transform that data and put it as an option in the source choice select widget, like we used to do. 

I was trying to cover the tests that I'm about to delete, in this case `test_flexible_file_widget_prepopulated_value_algorithm_job`. But with the FlexibleFileField we used `initial`. I'll need to check if we ever provided a way for algorithm editors of giving an initial file value for jobs. I don't think so. If it was just the trick of putting the "civ" as an option in the choices, than this test can go. 

It's a bit confusing, because `initial` can be used on forms, but is not meant as a fall-back value. So the way we used it was a little different from the intended usage (I think).

app/tests/reader_studies_tests/test_forms.py
koopmant: I get what you mean. I couldn't think of anything better. Now that we create multiple fields for some kinds, these tests sort of exploded; so I came up with the dictionary for expected fields with field type and widget type to make them a bit easier to write and read. 

TBH, I think testing for the types is overkill anyway, but we did that in some places so I started doing it on all the forms we test here and added testing of field type _and_ widget type, we sometimes only tested the widget type. So now testing for both for consistency. 
I'm also fine with just testing the expected field keys.

app/grandchallenge/components/form_fields.py
koopmant: Okay, so currently people can do it. Albeit through a work around (they'll have to select either "search existing" or "upload new" and then provide no data. The form will accept this and a file_civ of `None` will be processed, removing the civ.)
With the changes here, that workaround will not be possible. I'd like to add this in the correct way in a separate PR. I think we can already merge this; but I can also mark this branch as the target for that new PR.

app/tests/components_tests/test_form_fields.py
koopmant: Okay, still very confused about how we use `initial` in some places, but I'm certain this test can go. The new `current_socket_value` can only have a value if a Job already exists, so for the JobCreateForm, this will never happen. 
Our old way of using `initial` is no longer relevant. 

Sidenote, we still accept `initial` in `InterfaceFormFieldsMixin.get_fields_for_interface` because we use it in the `AdditionalInputsMixin`. That is now only passed for value kinds. This should not be a problem if I look at https://github.com/DIAGNijmegen/rse-roadmap/issues/357.

app/grandchallenge/components/form_fields.py
chrisvanrun: I don't think it is actively used. So it should be fine to do it via another PR. T.b.h. I wouldn't make large deal out of it because another workaround is create a new civ set, re-use old data but not the one you want to remove, then delete the old civ set.

app/tests/reader_studies_tests/test_forms.py
chrisvanrun: Do we ever switch widgets dynamically within a field, or are they bound to the Field class? If the latter, we could just test the Field types.

koopmant: We don't ever switch them dynamically, as far as I know. However, they are not bound to the Field class either. You can define the widget in the class; but you don't have to. For example, you can use a `ChoiceField` with a `Select` or a `RadioSelect` widget. You'd generally define that on the form.

PR #4524, opened 2026-01-27
Allow removing file civs

koopmant:
Add an option to remove file socket values in the select widget for file kinds. 

Screenshot 2026-01-27 at 16 34 30


chrisvanrun (approved):
Looks good! Made a small suggestion here: https://github.com/DIAGNijmegen/rse-grand-challenge/pull/4531

 Also, am I correct in think the actual files aren't removed but orphaned? AFAIK we currently don't clean them up via any periodic job or anything.

koopmant:
> Also, am I correct in think the actual files aren't removed but orphaned? AFAIK we currently don't clean them up via any periodic job or anything.

Yes, also made a note about that on the pitch. We have an issue about the same thing with image kinds somewhere. I'll make a reference of this there as well.

PR #4526, opened 2026-01-28
Make test cleanup robust

jmsmkn:
Refactors fixture which was added in #4476 so that it is better cleaned up.


chrisvanrun (approved):
LGTM!

PR #4527, opened 2026-01-28
Use `form_id` in place of `id`

jmsmkn:
Nit from #4473: prefer the use of an explicit `form_id` rather than `id` to avoid confusion with the `id()` built-in. This is why we use `pk` everywhere on Django models, rather than `id`.


koopmant (approved):
LGTM

> explicit form_id rather than id to avoid confusion with the id() built-in

noted!

PR #4528, opened 2026-01-28
Avoid passing empty string in flexible fields data validation

koopmant:
Remove the hidden input fields with the interface form field prefix to avoid passing these empty values. 

These fields were used for easy parsing in the `MultipleCIVForm`; however, they were not removed in `InterfaceFormFieldsMixin.clean` if the choice was `UNDEFINED`, effectively passing an empty string as the desired value for the socket.

We now recognise the flexible source choice field as an interface field, so we do not need the hidden fields anymore.

### DIFF
--- a/app/grandchallenge/components/forms.py
+++ b/app/grandchallenge/components/forms.py
@@ -456,31 +456,27 @@ class MultipleCIVForm(InterfaceFormFieldsMixin, Form):
                 )
             )
 
-        # Add fields for dynamically added new interfaces
-        for field_name in self.data.keys():
-            interface_slug = self.is_interface_field(field_name=field_name)
-
-            if (
-                interface_slug
-                and field_name not in self.fields.keys()
-                and ComponentInterface.objects.filter(
-                    slug=interface_slug
-                ).exists()
-            ):
-                interface = ComponentInterface.objects.filter(
-                    slug=interface_slug
-                ).get()
-
-                self.fields.update(
-                    self.get_fields_for_interface(
-                        interface=interface,
-                        user=self.user,
-                        required=False,
-                    )
+        for interface in self.get_dynamically_added_interfaces():
+            self.fields.update(
+                self.get_fields_for_interface(
+                    interface=interface,
+                    user=self.user,
+                    required=False,
                 )
+            )
+
+    def get_dynamically_added_interfaces(self):
+        new_interface_slugs = set()
+        for field_name in self.data.keys():
+            interface_slug = self.get_interface_slug(field_name=field_name)
+
+            if interface_slug and field_name not in self.fields.keys():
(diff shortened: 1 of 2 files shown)


app/grandchallenge/components/forms.py
```diff
@@ -491,12 +480,13 @@ def __init__(self, *args, instance, base_obj, user, **kwargs):  # noqa C901
                 )
 
     @staticmethod
-    def parse_slug(*, slug):
-        if not slug.startswith(INTERFACE_FORM_FIELD_PREFIX):
-            return None
-
-        interface_slug = slug[len(INTERFACE_FORM_FIELD_PREFIX) :]
-
+    def is_interface_field(*, field_name):
```
chrisvanrun: Reads like a boolean check, should really be a `get_interface_field` or `extract_interface_field` imho.

chrisvanrun: My suggestion would be to write a function that extracts all the interface slugs, based on that data and directly look through that, instead of a loop over the data, key extraction and then only act if it is None.

Also allows you to not just check for just the CHOICE but all the prefixes. That way, you could mimic the logic in the `clean`:

```Python
            if any(
                [
                    key.startswith(prefix.value)
                    for prefix in FlexibleWidgetPrefixes
                ]
            ):
```

koopmant: I still need to extract the interface slug though. I'm working on it, but it's just getting messier at the moment

koopmant: how is this?
```python
        for interface in self.get_dynamically_added_interfaces():
            self.fields.update(
                self.get_fields_for_interface(
                    interface=interface,
                    user=self.user,
                    required=False,
                )
            )

    def get_dynamically_added_interfaces(self):
        new_interface_slugs = set()
        for field_name in self.data.keys():
            interface_slug = self.get_interface_slug(field_name=field_name)

            if interface_slug and field_name not in self.fields.keys():
                new_interface_slugs.add(interface_slug)

        return list(
            ComponentInterface.objects.filter(slug__in=new_interface_slugs)
        )

    @staticmethod
    def get_interface_slug(*, field_name):
        if field_name.startswith(INTERFACE_FORM_FIELD_PREFIX):
            interface_slug = field_name[len(INTERFACE_FORM_FIELD_PREFIX) :]
        elif field_name.startswith(FlexibleWidgetPrefixes.CHOICE):
            interface_slug = field_name[len(FlexibleWidgetPrefixes.CHOICE) :]
        else:
            interface_slug = None
        return interface_slug
```

koopmant: I've refactored it. But I still only check for the CHOICE value because it always needs to be there. If there somehow is other flexible fields data, but the choice data does not exist, we should just ignore it.

PR #4529, opened 2026-01-28
Split civ data validation from processing

koopmant:
Split CIV data validation and processing of the validated CIV data objects. 

Validation of `allowed_socket_slugs`, which is used for archives, has moved from `CIVForObjectMixin` to `ArchiveItemPostSerializer` and `MultipleCIVForm` (used by ArchiveItem Create/Update and DisplaySet Create/Update). 

Related to:
- https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/685

### DIFF
--- a/app/grandchallenge/components/forms.py
+++ b/app/grandchallenge/components/forms.py
@@ -421,7 +421,7 @@ class AdditionalInputsMixin(UserMixin, InterfaceFormFieldsMixin):
         keys_to_remove = []
         inputs = []
         # Cannot call add_error in the for-loop because it updates cleaned_data,
-        # so safe errors to call add_error later.
+        # so save errors to call add_error later.
         errors = {}
 
         for key, value in cleaned_data.items():
@@ -514,7 +514,7 @@ class MultipleCIVForm(InterfaceFormFieldsMixin, Form):
         keys_to_remove = []
         inputs = []
         # Cannot call add_error in the for-loop because it updates cleaned_data,
-        # so safe errors to call add_error later.
+        # so save errors to call add_error later.
         errors = {}
 
         for key, value in cleaned_data.items():


app/grandchallenge/reader_studies/serializers.py
```diff
@@ -178,11 +177,6 @@ class DisplaySetPostSerializer(
         slug_field="slug", queryset=ReaderStudy.objects.none(), required=False
     )
 
-    def create(self, validated_data):
-        if validated_data.pop("values", []) != []:
-            raise DRFValidationError("Values can only be added via update")
-        return super().create(validated_data)
-
```
koopmant, line 185: This is already defined in `CIVSetPostSerializerMixin`.

app/grandchallenge/components/forms.py
```diff
@@ -502,21 +510,43 @@ def clean(self):
...
+                except ValidationError as error:
+                    errors[key] = error
+                else:
+                    inputs.append(civ_data)
 
         for key in keys_to_remove:
             cleaned_data.pop(key)
 
+        for key, error in errors.items():
+            self.add_error(key, error)
```
chrisvanrun, line 552: Why keep errors in a subset, and not add it directly via `.add_error`?

app/grandchallenge/components/forms.py
```diff
@@ -418,21 +419,28 @@ def clean(self):
...
```
chrisvanrun, line 444: Same here.

app/grandchallenge/archives/serializers.py
```diff
@@ -81,3 +81,29 @@ def __init__(self, *args, **kwargs):
                 user=user,
                 codename="upload_archive",
             )
+
+    def validate(self, data):
+        if self.instance:
+            archive = self.instance.archive
+        else:
+            archive = data["archive"]
```
chrisvanrun, line 89: There might not be an `"archive"` since it is not required. Does the `validate` return a nice error or just a 500 server error when that occurs?

chrisvanrun (changes requested):
We had a small meet and discussed why the `process_civ_data_objects_and_execute_linked_task` is generally called only if the `form_valid`, and not just as final part of the clean. Would be great if we could just throw the validation errors via the regular cleans. Tricky because the regular Django `ModelForm` only does a `.save()` on `form_valid()`, and we somewhat rely on 

I disagree that this splits the validation and processing fully. But it is a step in the right direction!

A few nits.

app/grandchallenge/components/forms.py
koopmant: I should have added a comment here. `add_error` takes the key out of cleaned_data, breaking the for-loop. So we add the errors later.

app/grandchallenge/archives/serializers.py
koopmant: The logic here is based on my understanding of the viewset: either we use the archive item update route, then there must be and instance; or we use the archive item create route, then there must be an archive in the data. 

If we try to create without giving an archive in the data we get:
```python
response = get_view_for_user(
        viewname="api:archives-item-list",
        method=client.post,
        data={"values": []},  # don't provide archive
    )
response.json()
> {'archive': ['This field is required.']}
```

Still, I agree this looks a bit funny. I'm not sure if this is the right way to do it.

app/grandchallenge/components/forms.py
chrisvanrun: Ah. Check!

app/grandchallenge/components/forms.py
```diff
@@ -510,8 +512,10 @@ def clean(self):
         cleaned_data = super().clean()
 
         keys_to_remove = []
-        errors = {}
         inputs = []
+        # Cannot call add_error in the for-loop because it updates cleaned_data,
+        # so safe errors to call add_error later.
```
jmsmkn: Should be "save" rather than "safe"?

app/grandchallenge/archives/serializers.py
chrisvanrun: It does look a bit funny, but I think it works. Even better: looking at the internals of `BaseSerializer` the `save()` choosing to either `update` or `create` routes use exactly that same switch on `self.instance`. Albeit, they check if it `is None` I don't see a direct boolean check to be potentially harmful here (unless someone decides to add a `__bool__` to the instance =P.

app/grandchallenge/components/forms.py
koopmant: Thank you

PR #4530, opened 2026-01-29
Validate is_editable on form

koopmant:
Add the validation for `is_editable` we have on the serializers in `CIVSetPostSerializerMixin` to the forms in `CIVSetUpdateFormMixin`.

### DIFF
--- a/app/grandchallenge/components/forms.py
+++ b/app/grandchallenge/components/forms.py
@@ -8,6 +8,7 @@ from dal.widgets import Select
 from django import forms
 from django.conf import settings
 from django.contrib.auth import get_user_model
+from django.core.exceptions import ValidationError
 from django.forms import (
     CheckboxSelectMultiple,
     Field,
@@ -553,6 +554,12 @@ class CIVSetCreateFormMixin:
 
 
 class CIVSetUpdateFormMixin:
+    def clean(self):
+        if not self.instance.is_editable:
+            raise ValidationError(self.instance.not_editable_error_message)
+
+        return super().clean()
+
     def process_object_data(self):
         instance = self.instance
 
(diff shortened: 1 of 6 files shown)


app/grandchallenge/components/forms.py
```diff
@@ -553,6 +554,12 @@ def process_object_data(self):
 
 
 class CIVSetUpdateFormMixin:
+    def clean(self):
+        if not self.instance.is_editable:
```
jmsmkn, line 558: Would this be cleaner if done on the model clean method?

app/grandchallenge/reader_studies/serializers.py
```diff
@@ -171,18 +170,10 @@ class DisplaySetPostSerializer(
...
     DisplaySetSerializer,
 ):
-    editability_error_message = (
-        "This display set cannot be changed, as answers for it already exist."
-    )
     reader_study = SlugRelatedField(
         slug_field="slug", queryset=ReaderStudy.objects.none(), required=False
     )
 
-    def create(self, validated_data):
```
jmsmkn, line 181: Why can this be removed?

koopmant: Ah sorry, I see I had this in two PR's and only placed a comment about it in the other one. 
This is already implemented in the parent class `CIVSetPostSerializerMixin`.

app/grandchallenge/components/forms.py
koopmant: It seems we are not using a ModelForm in this case. Maybe I should change that?

jmsmkn: I'm unsure - that might introduce more problems as this isn't the instance that is being updated/changed, right?

chrisvanrun: The instance might get updated, tricky bit is the `values` might only change here or is targeted. It is used in combination with `MultipleCIVForm` which has `Form` as basis.

imho that should really be a `ModelForm` but that is quite the can of worms to open.

PR #4537, opened 2026-02-04
Run image post processing in a virtual env

jmsmkn:
This will hopefully address some of the issues we have with memory and timeouts when using vips. We can completely remove panimg from the web dependencies if this works by migrating the conversion tasks too.

TODO
 - [ ] Replace virtualenv with a released version of panimg once https://github.com/DIAGNijmegen/rse-panimg/pull/122 is merged and released


chrisvanrun (commented):
Looks good! Provisional OK given the release of panimg is pending.

PR #4538, opened 2026-02-04
Remove add overlay segments task

jmsmkn:
This command has not been used in a long time and was out of date.

See #4537


chrisvanrun (approved):
AFAIK it is correctly removed. Haven't traced that it wasn't used anywhere but the other tests passing should suffice to conclude it isn't.

PR #4541, opened 2026-02-06
Remove panimg from main environment

jmsmkn:
This PR removes panimg as a direct dependency and puts everything into the virtual environment. As vips is no longer in the main environment this should solve the issue with pycurl installation on all platforms. Tests are removed that were only testing panimg functionality, I checked that they are present in the panimg repo - these were left over from the extraction of panimg from this codebase a few years ago.

If this all works then we should create another package for the shared models between panimg and grand challenge, for now they are copied in.

### DIFF
--- a/app/grandchallenge/cases/panimg.py
+++ b/app/grandchallenge/cases/panimg.py
@@ -8,6 +8,7 @@ from pydantic import TypeAdapter
 from grandchallenge.cases.panimg_models import (
     PanImgFile,
     PanImgResult,
+    PostProcessorOptions,
     PostProcessorResult,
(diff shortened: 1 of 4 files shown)


app/tests/cases_tests/test_mhd.py
chrisvanrun, line 1: Why did we have tests like this anyway? As a double check on panimg? =/

jmsmkn: These were left over from the extraction of panimg from this codebase a few years ago, they were kept as they still passed after the extraction.

PR #4547, opened 2026-02-10
Small improvements: Layout header and add fieldsets

chrisvanrun:
Fixes a bug where the header 'Update budget field' duplicating with validation errors (quite hilariously):

Screenshot 2026-02-10 at 11 43 58

In addition, changed the order of fields and added some field sets for sorting.

I did play around a bit with a description/title per task. That would add a bit of human readability if you have >2 tasks. Pushed it out of scope for now as I just want to focus on getting the PRs merged. Maybe something for the future.

<details>
<summary>
Screenshot of 3 tasks
</summary>
Screenshot 2026-02-10 at 12 15 31
</details>

### DIFF
--- a/app/grandchallenge/challenges/forms.py
+++ b/app/grandchallenge/challenges/forms.py
@@ -601,11 +601,11 @@ class ChallengeRequestBudgetUpdateForm(forms.ModelForm):
     class Meta:
         model = ChallengeRequest
         fields = (
+            "task_ids",
             "algorithm_selectable_gpu_type_choices_for_tasks",
             "algorithm_maximum_settable_memory_gb_for_tasks",
             "average_size_test_case_mb_for_tasks",
             "inference_time_average_minutes_for_tasks",
-            "task_ids",
             "task_id_for_phases",
             "number_of_teams_for_phases",
             "number_of_submissions_per_team_for_phases",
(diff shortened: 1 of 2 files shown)


app/grandchallenge/challenges/forms.py
```diff
@@ -601,11 +601,11 @@ class ChallengeRequestBudgetUpdateForm(forms.ModelForm):
     class Meta:
         model = ChallengeRequest
         fields = (
+            "task_ids",
```
koopmant, line 604: I initially placed it close to task_id_for_phases, but here is fine too.

app/grandchallenge/challenges/forms.py
```diff
@@ -626,7 +626,29 @@ def __init__(self, *args, **kwargs):
                 "hx-swap": "outerHTML",
             }
         )
-        self.helper.layout.append(Submit("save", "Save"))
+        self.helper.layout = Layout(
+            HTML("<h2>Update budget fields</h2>"),
```
koopmant, line 630: Ah, nice thanks! I see how that works.

app/grandchallenge/challenges/forms.py
chrisvanrun: I noticed we tend to be on odds on ordering of things ;-)

PR #4548, opened 2026-02-10
Move biopython to biom virtualenv

### LINKED ISSUE #4542: Move `biopython` to virtualenv
This is a hefty dependency that adds numpy to the webserver, I'm not sure that is completely necessary. Would be good to move this to its own environment as we have done for panimg and biom.

jmsmkn:
Closes #4542

### DIFF
--- a/app/grandchallenge/components/validators.py
+++ b/app/grandchallenge/components/validators.py
@@ -34,24 +34,27 @@ def validate_relative_path_not_reserved(value):
         raise ValidationError("This relative path is reserved")
 
 
-def _newick_parser(tree):
-    return NewickIO.Parser.from_string(tree)  # noqa: F821
-
-
-def validate_newick_tree_format(tree):
+def validate_newick_tree_format(*, file):
     """Validates a Newick tree by passing it through a parser"""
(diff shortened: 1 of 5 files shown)


app/grandchallenge/components/validators.py
```diff
@@ -35,24 +34,27 @@ def validate_relative_path_not_reserved(value):
...
     try:
-        for _ in parser.parse():
-            has_tree = True
-    except NewickIO.NewickError as e:
-        raise ValidationError(f"Invalid Newick tree format: {e}")
-
-    if not has_tree:
-        raise ValidationError("No Newick tree found")
+        run_script_in_venv(
+            venv_location=settings.COMPONENTS_VIRTUAL_ENV_BIOM_LOCATION,
```
chrisvanrun, line 43: What is the rationale for not creating a dedicated env for NEWICK? Optimalisation since they share a lot of deps?

jmsmkn: Yes, that was it. These seem to be fine together but we know that vips + biom do not place nicely together so have to be separate.

PR #4554, opened 2026-02-11
Add staff page to calculate challenge costs

### LINKED ISSUE #3787: Add a admin page to calculate challenge costs

koopmant:
Add staff page to calculate challenge costs under `/challenges/budget-calculator`. The form view returns the budget for an unsaved instance of `ChallengeRequest` based on the budget fields input. It updates when the input changes. 

Closes #3787

Screenshot:

Screenshot 2026-02-11 at 14 39 25

### DIFF
--- a/app/grandchallenge/challenges/forms.py
+++ b/app/grandchallenge/challenges/forms.py
@@ -11,7 +11,7 @@ from crispy_forms.layout import (
 from django import forms
 from django.conf import settings
 from django.core.exceptions import ValidationError
-from django.forms import TextInput
+from django.forms import Textarea
 from django.urls import reverse
 from django.utils.html import format_html
 from django.utils.text import format_lazy
@@ -614,7 +614,7 @@ class ChallengeRequestBudgetUpdateForm(forms.ModelForm):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         for field in self.fields.values():
-            field.widget = TextInput()
+            field.widget = Textarea(attrs={"rows": 1})
         self.helper = FormHelper(self)
         self.helper.form_id = "budget"
         self.helper.attrs.update(


app/grandchallenge/challenges/forms.py
```diff
@@ -612,6 +613,8 @@ class Meta:
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        for field in self.fields.values():
+            field.widget = TextInput()
```
chrisvanrun: Aren't these sensible defaults already? What do they default to?

app/grandchallenge/challenges/forms.py
```diff
@@ -728,3 +731,18 @@ def _clean_later_phases_not_more_teams_or_submissions(
                         field_name,
                         "Later phases in a task may not have more submissions than earlier phases.",
                     )
+
+
+class ChallengeRequestBudgetCalculatorForm(ChallengeRequestBudgetUpdateForm):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.helper.form_id = "budget-form"
```
chrisvanrun, line 740: Double check: we're replacing this because the original `"budget"` is already taken by the parent class, and we use `"budget"` here as target for the rendering in the template. Makes sense.

app/grandchallenge/challenges/forms.py
koopmant: They default to (enormous) text areas

app/grandchallenge/challenges/forms.py
koopmant: Exactly

app/grandchallenge/challenges/forms.py
chrisvanrun: IIRC a TextInput does not allow multiple lines? I think it would make sense to allow pasting a 'pretty' JSON. I originally thought to maybe have JSONEditor widget: but I discovered that doesn't work nicely for the inserted form in the challenge request details page. There are (ugly) workarounds but I think having a text field here would be better.

koopmant: I don't think so. The inputs are all a short list of items. The longest is for the GPU choices, but even that easily fits on a line, e.g. '[["", "T4", "A10G"], ["", "T4", "A10G"]]'. We could otherwise set rows for the text area, but honestly, I think we really don't need them. 

You can see what the text areas look like in the current edit budget fields form on the challenge requests detail page.

chrisvanrun: I get the practical aspects of this, we don't need to see all those empty lines. However, I disagree that breaking or limiting the JSON format is the good approach. Imho it is better to keep it at a TextField but limit the rows.

koopmant: All right

PR #4556, opened 2026-02-12
Monitor task memory

jmsmkn:
See https://github.com/DIAGNijmegen/rse-cloud-infrastructure/issues/212

### DIFF
--- a/app/config/celery.py
+++ b/app/config/celery.py
@@ -123,7 +123,7 @@ def log_memory(*_, task_id, task, **__):
         change = mem.percent - INITIAL_MEMORY_PERCENT
 
         logger.info(
-            f"Task {task.name} [{task_id}] Memory: {mem.percent}% used ({change:.1f}% change) | {mem.used / (1024 ** 3):.2f}/{mem.total / (1024 ** 3):.2f} GB"
+            f"Task {task.name} [{task_id}] Memory: {mem.percent}% used ({change:.1f}% change) | {mem.used / (1024 ** 3):.2f}/{mem.total / (1024 ** 3):.2f} GiB"
         )
 
         if change > 20:


app/config/celery.py
```diff
@@ -101,3 +102,29 @@ def remove_ecs_scale_in_protection(*_, **__):
...
+
+@task_postrun.connect()
+def log_memory(*_, task_id, task, **__):
+    if celery_app.is_solo_worker:
+        mem = psutil.virtual_memory()
+
+        change = mem.percent - INITIAL_MEMORY_PERCENT
+
+        logger.info(
+            f"Task {task.name} [{task_id}] Memory: {mem.percent}% used ({change:.1f}% change) | {mem.used / (1024 ** 3):.2f}/{mem.total / (1024 ** 3):.2f} GB"
```
chrisvanrun: Wow. That is one neat f-string.

pkcakeout:
```suggestion
            f"Task {task.name} [{task_id}] Memory: {mem.percent}% used ({change:.1f}% change) | {mem.used / (1024 ** 3):.2f}/{mem.total / (1024 ** 3):.2f} GiB"
```

It is a mouth full, but I would in fact only suggest fixing the unit.

pkcakeout (approved):
Fine by me, I wonder a bit though: The intent is to monitor the _entire node_? Or is it _this process_?

You could get a better estimate if you only want to observe this process by checking:

```
>>> pp = psutil.Process(os.getpid())
>>> pp.memory_info()
pmem(rss=22941696, vms=27140096, shared=8228864, text=3272704, lib=0, data=15937536, dirty=0)
```

`rss` tells you pretty much what you want to know here. Could maybe be better?

Approving it anyways in case this was meant to track the entire node...

pkcakeout:
Ok... accidentally approved this with a suggested change... I would fix the unit, but it is also a small thing, so...

PR #4560, opened 2026-02-18
Remove Traefik

jmsmkn:
Routing to the workstation container images is now done directly through docker. When the container is started we inspect the host address and mapping of the websocket and http ports. We check other running container instances to ensure that users are not mapped to other containers.

See https://github.com/DIAGNijmegen/rse-roadmap/issues/224

### DIFF
--- a/app/grandchallenge/components/backends/docker_client.py
+++ b/app/grandchallenge/components/backends/docker_client.py
@@ -181,6 +181,4 @@ def run_container(
     # Last two args must be the repo tag and optional command
     docker_args.append(repo_tag)
 
-    logger.info(f"Running: {docker_args}")
-
     return _run_docker_command(*docker_args)


app/grandchallenge/components/backends/docker_client.py
```diff
@@ -196,68 +161,26 @@ def run_container(  # noqa: C901
...
-                [
-                    "--publish",
-                    f"{bind_address}:{host_port}:{container_port}",
-                ]
-            )
+    for port in ports:
+        docker_args.extend(
+            [
+                "--publish",
+                f"0.0.0.0::{port}",
```
jmsmkn, line 177: ECS does not support host binding, so publish to all here.

app/grandchallenge/workstations/views.py
```diff
@@ -459,14 +459,27 @@ def get_context_data(self, **kwargs):
...
 
+    if (
+        not all(
+            [session.host_address, session.http_port, session.websocket_port]
+        )
+        or session.status != session.STARTED
+    ):
+        raise Http404
+
+    if "mlab4d4c4142" in path:
```
koopmant, line 474: Should this be a variable somewhere?

According to the docstring in WorkstationImage the path needs to start with this, not sure how strict we need to be with that... Probably good to remove that detail from the docstring anyway.

Makefile
```diff
@@ -49,7 +49,7 @@ build_web_dist:
 		--build-arg LOCKFILE_HASH=$(LOCKFILE_HASH) \
 		--build-arg GRAND_CHALLENGE_WEB_TEST_BASE_REPOSITORY_URI=$(GRAND_CHALLENGE_WEB_TEST_BASE_REPOSITORY_URI) \
 		--build-arg GRAND_CHALLENGE_WEB_BASE_REPOSITORY_URI=$(GRAND_CHALLENGE_WEB_BASE_REPOSITORY_URI) \
-		--network rse-grand-challenge_default \
+		--network grand-challengeorg_default \
```
koopmant, line 52: Not important, but why not stick with rse-grand-challenge_default?

app/tests/workstations_tests/test_models.py
```diff
@@ -105,23 +97,13 @@ def test_session_start(
 
         expected_labels = {
             "job": f"{s._meta.app_label}-{s._meta.model_name}-{s.pk}",
```
koopmant, line 99: Is this job label still being set? I cannot find it anywhere...

app/grandchallenge/workstations/views.py
jmsmkn: No, this is hardcoded within MeVisLab, won't be something that changes.

Makefile
jmsmkn: As there was another network being created in docker-compose.yml so went for consistency with the name there.

app/tests/workstations_tests/test_models.py
jmsmkn: The test is x-fail as it uses docker which is slated for removal, will be fixed later.

app/grandchallenge/workstations/views.py
jmsmkn: The starting with the string was removed due to #1360

PR #4562, opened 2026-02-20
Run workstation sessions on ECS

### LINKED ISSUE #4031: Docker session tests broken again
3 breakages in the past week. These should be re-written so that they do not rely on docker itself.

jmsmkn:
This PR:

- Changes the workstation session compute backend to ECS
- Moves all workstation lifecycle tasks to be run in the main region, which then uses the API of the rendering region. These now use celery tasks that are atomic with proper single transitions.
- Removes dependency on docker from workstation session runtime and tests

Closes #4031
See https://github.com/DIAGNijmegen/rse-roadmap/issues/224

### DIFF
--- a/app/grandchallenge/components/backends/amazon_ecs.py
+++ b/app/grandchallenge/components/backends/amazon_ecs.py
@@ -104,7 +104,9 @@ class ECSTaskOrchestrator:
     def _wait_for_task_running(self, *, task_arn):
         waiter = self._ecs_client.get_waiter("tasks_running")
         waiter.wait(
-            cluster=settings.COMPONENTS_SERVICE_CLUSTER_NAME, tasks=[task_arn]
+            cluster=settings.COMPONENTS_SERVICE_CLUSTER_NAME,
+            tasks=[task_arn],
+            WaiterConfig={"Delay": 5, "MaxAttempts": 36},  # 3 minutes
         )
 
     def _get_task_description(self, *, task_arn):


app/grandchallenge/workstations/models.py
```diff
@@ -530,7 +535,10 @@ def environment(self) -> dict:
             }
         )
 
-        if self.creator:
+        if (
+            self.creator
+            and settings.COMPONENTS_SERVICE_INCLUDE_CREATOR_AUTH_TOKEN
```
koopmant, line 540: Why did we add a flag for this?

app/grandchallenge/components/tasks.py
```diff
@@ -1052,39 +1053,140 @@ def deprovision_job(
...
 
-@shared_task
+@acks_late_micro_short_task(
+    retry_on=(
+        LockNotAcquiredException,
+        RetryStep,
+    )
+)
+@transaction.atomic
 def start_service(*, pk: uuid.UUID, app_label: str, model_name: str):
```
koopmant, line 1063: Please rename `service` in these three tasks to `session`.

app/grandchallenge/components/backends/amazon_ecs.py
```diff
@@ -0,0 +1,244 @@
...
+from grandchallenge.evaluation.utils import get
+
+
+class ConnectionInformation(NamedTuple):
+    host_address: str
+    http_port: int
+    websocket_port: int
+
+
+class ECSService:
```
koopmant: I find the name a bit confusing, I think something like `ECSTask` would be more appropriate?

app/grandchallenge/workstations/models.py
jmsmkn: Since the last release of CIRRUS it is unnecessary so it gives use a quick way to turn it on and off once the previous release of CIRRUS is retired at the end of this cycle. Additionally, it made writing the Stubber tests easier as it was not possible to know the generated variable prior to calling `start_service`.

app/grandchallenge/components/tasks.py
jmsmkn: We cannot easily rename celery tasks or change their args/kwargs, you need a migration cycle like you have for deprecating fields.

The naming and approach here mirrors `provision_job`, `execute_job` etc. above. The tasks take in the app label and model name so that they can work with multiple different implementations of "component service"s (that underneath use `ComponentImage`) just like implementations of `ComponentJob` use `ComponentImage`.

I would rather limit the use of `session` as it is too overloaded:

- The browser session (this is something we cannot rename and is associated with the users session cookie)
- Workstation session
- Case upload session
- more? PyCharm complains of too many results if I search the codebase for session!

Service is too, but not as much, and it is already qualified as Component Service. I would prefer to keep to `ComponentService` like `ComponentJob` - this is why I added environment variables with the prefix `COMPONENT_SERVICE_`. Thus, for the naming and migration reasons I would suggest not changing the task names. I will ensure the kwargs that have changed for `stop_expired_services` are handled.

app/grandchallenge/components/backends/amazon_ecs.py
jmsmkn: Yes, you also have ECS services so this isn't quite right. It is the ECS (Task) implementation of "component services" (see discussion on naming), so `ECSComponentService`?

jmsmkn: Or maybe we follow the pattern of `ComponentJobs` and have `Executor`, which is service land may be `Orchestrator`:

```python3
orchestrator = ECSTaskComponentServiceOrchestrator(**service.orchestrator_kwargs)
```

Which then saves having `service` and `ecs_service` in the tasks.

app/grandchallenge/components/tasks.py
koopmant: I see. That was not obvious to me though. I'm still a bit confused about the Component Service. In these tasks service could mean job, evaluation or session. I can see how service is a good enough name and indeed let's not bother renaming here. 
But the environment variables are meant for ECS tasks only, aren't they?

app/grandchallenge/components/backends/amazon_ecs.py
koopmant: That's quite long though. Maybe just `ECSTaskOrchestrator`?

app/grandchallenge/components/tasks.py
jmsmkn: Service in this case is only for implementations of "component service"s, which for now is only `workstations.Session` but there could be other concrete implementations (e.g. "Interactive Algorithms"). `algorithms.Job` and `evaluations.Evaluation` are implementations of `ComponentJob`, so are not handled here. It is the flip side to `ComponentJob`, which is why the pattern is very similar, but `ComponentJob` and "component service"s are orthogonal, they both are running `ComponentImage`s underneath, but doing different things.

The environment variables are not only for ECS Tasks - they are for any backend/orchestrator that runs the "component service"s. Like now we have the Docker one which we are just deleting, but for `ComponentJob`s we have `AmazonSageMakerBaseExecutor` and `IOCopyExecutor`/`UtilizationExecutor` in tests. It makes swapping out the backend easier both for testing and when we decide to change backends - which we do quite often: Docker -> AWS Batch -> ECS Tasks -> SageMaker Inference -> SageMaker Training for `ComponentJob`s, Docker -> ECS Tasks for "component service"s, and in the future this may be EKS/k8s/SageMaker Endpoints (e.g. for "Interactive Algorithms").

koopmant: > Service in this case is only for implementations of "component service"s, which for now is only workstations.Session but there could be other concrete implementations

I finally realized this when I read back my own comment. All right, this makes perfect sense. Thanks for the explanation!

app/tests/workstations_tests/test_tasks.py
```diff
@@ -1,11 +1,422 @@
...
+            },
+            service_response={
+                "tasks": [
+                    {"taskArn": "test-task-arn", "lastStatus": "RUNNING"}
+                ],
+                "failures": [],
+            },
+        )
+        ecs_stubber.add_response(
+            method="describe_tasks",
```
koopmant, line 264: Apparently we are calling `describe_tasks` twice, but I don't see how?

jmsmkn: The first one is called internally by:

```python3
waiter = self._ecs_client.get_waiter("tasks_running")
waiter.wait(...)
```

PR #4565, opened 2026-02-23
Fix ArchiveItem creation for challenge-bound archives

### LINKED ISSUE #4564: Error in archive item upload validation on archives with `allowed_component_interfaces`
**Describe the bug**

Uploading an archive item to an archive for a challenge (which has `allowed_component_interfaces` set) with only values that are asynchronously verified (e.g. images or json validation) causes 500 server errors if any of the data uploaded is invalid. It is unclear in these cases what has failed.

This piece of code seems to be responsible:

https://github.com/DIAGNijmegen/rse-grand-challenge/blob/8ee9d6919ba87696d9790ccc811f2c7757366ca6/app/grandchallenge/archives/serializers.py#L91-L104

The `data["values"]` will be unset in the scenario described causing a script error.

**To Reproduce**

Exact steps unknown, but in the simplest case, uploading an empty archive item via gcapi to an archive with `allowed_component_interfaces` will trigger the error.

**Expected behavior**

The correct validation error is returned, not a generic 500 server error.

chrisvanrun:
Closes #4564

Logic is only applied when `allowed_socket_slugs` is set on the archive.

### DIFF
--- a/app/grandchallenge/archives/serializers.py
+++ b/app/grandchallenge/archives/serializers.py
@@ -94,7 +94,9 @@ class ArchiveItemPostSerializer(
             pass
         else:
             errors = []
-            for civ in data["values"]:
+            for civ in data.get(
+                "values", []
+            ):  # "values" not required when creating an archive item
                 interface = civ["interface"]
                 if interface.slug not in allowed_socket_slugs:
                     errors.append(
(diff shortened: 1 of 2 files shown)


jmsmkn (approved):
LGTM

PR #4572, opened 2026-02-26
Challenge Request Draft

chrisvanrun:
This change set adds a new status to `ChallengeRequest`s: `DRAFT`.

With the workflow being:
`DRAFT` > `PENDING` > `ACCEPT`/`REJECT`.

The `ChallengeRequestStatusUpdateForm` is now no longer used to render an in-table and in-line form. Rather, it is redesigned to only respond to POSTs from buttons on the `ChallengeRequest` detail page: submit/accept/reject. The rational is that the in-table control isn't very friendly for users (it was OK for site admins) and IMHO it is better to have a single place where you update are done.

## Permissions

New `"challenges.review_challengerequest"` is added as to indicate that someone (or some group) can move `PENDING` to `ACCEPT`/`REJECT`. The existing `"challenges.change_challengerequest"` permission is used as criteria for `DRAFT` to `PENDING`, aka submitting. And in the near future will be used as criteria for updating the draft.

Manually, after merging, the `challenges.review_challengerequest` should be applied to the `Group.objects.get(name=settings.CHALLENGES_REVIEWERS_GROUP_NAME)`.

## Feature flag
Note: I've added an automatic `DRAFT` -> `PENDING` at creation for now, so this PR can directly be merged into `main`. This acts sort of a feature flag.

### DIFF
--- a/app/grandchallenge/challenges/templates/challenges/challengerequest_detail.html
+++ b/app/grandchallenge/challenges/templates/challenges/challengerequest_detail.html
@@ -65,7 +65,7 @@
                 {% if object.status == object.ChallengeRequestStatusChoices.DRAFT and 'change_challengerequest' in user_perms %}
                     <a class="btn btn-lg btn-primary"
                         href="#"
-                        hx-post="{% url 'challenges:requests-status-update' pk=object.pk %}"
+                        hx-post="{% url 'challenges:requests-submit' pk=object.pk %}"
                         hx-vals='{"status": "{{ object.ChallengeRequestStatusChoices.PENDING.value }}"}'
                         hx-confirm="Are you sure you want to submit the request?"
                     >
@@ -74,7 +74,7 @@
                 {% elif object.status == object.ChallengeRequestStatusChoices.PENDING and 'review_challengerequest' in user_perms %}
                     <a class="btn btn-success"
                         href="#"
-                        hx-post="{% url 'challenges:requests-status-update' pk=object.pk %}"
+                        hx-post="{% url 'challenges:requests-process' pk=object.pk %}"
                         hx-vals='{"status": "{{ object.ChallengeRequestStatusChoices.ACCEPTED.value }}"}'
                         hx-confirm="Are you sure you want to ACCEPT the request? A challenge will be created!"
                     >
@@ -82,7 +82,7 @@
                     </a>
                     <a class="btn btn-danger"
                         href="#"
-                        hx-post="{% url 'challenges:requests-status-update' pk=object.pk %}"
+                        hx-post="{% url 'challenges:requests-process' pk=object.pk %}"
                         hx-vals='{"status": "{{ object.ChallengeRequestStatusChoices.REJECTED.value }}"}'
                         hx-confirm="Are you sure you want to REJECT the request?"
                     >
--- a/app/grandchallenge/challenges/views.py
+++ b/app/grandchallenge/challenges/views.py
@@ -7,7 +7,7 @@ from django.contrib.auth.mixins import (
 )
 from django.contrib.messages.views import SuccessMessageMixin
 from django.db.models import F, Prefetch, Q
-from django.http import Http404, HttpResponse, HttpResponseBadRequest
+from django.http import Http404, HttpResponse
 from django.shortcuts import render
 from django.utils.html import format_html
 from django.views.generic import (
@@ -236,20 +236,6 @@ class ChallengeRequestStatusUpdate(
     form_class = ChallengeRequestStatusUpdateForm
     login_url = reverse_lazy("account_login")
 
-    def get_required_permissions(self, request=None):
-        if (
-            self.get_object().status
-            == self.get_object().ChallengeRequestStatusChoices.PENDING
-        ):
-            return ["review_challengerequest"]
-        elif (
-            self.get_object().status
-            == self.get_object().ChallengeRequestStatusChoices.DRAFT
-        ):
-            return ["change_challengerequest"]
-        else:
-            raise HttpResponseBadRequest()
-
     def form_valid(self, form):
         super().form_valid(form)
 
@@ -269,6 +255,14 @@ class ChallengeRequestStatusUpdate(
         response = HttpResponse()
         response["HX-Refresh"] = "true"
         return response
+
+
+class ChallengeRequestProcess(ChallengeRequestStatusUpdate):
+    permission_required = "review_challengerequest"
+
+
+class ChallengeRequestSubmit(ChallengeRequestStatusUpdate):
+    permission_required = "change_challengerequest"
 
 
 class ChallengeRequestBudgetUpdate(
--- a/app/grandchallenge/challenges/urls.py
+++ b/app/grandchallenge/challenges/urls.py
@@ -8,7 +8,8 @@ from grandchallenge.challenges.views import (
     ChallengeRequestCreate,
     ChallengeRequestDetail,
     ChallengeRequestList,
-    ChallengeRequestStatusUpdate,
+    ChallengeRequestProcess,
+    ChallengeRequestSubmit,
     UsersChallengeList,
 )
 
@@ -29,9 +30,14 @@ urlpatterns = [
         name="requests-detail",
     ),
     path(
-        "requests/<pk>/update/status/",
-        ChallengeRequestStatusUpdate.as_view(),
-        name="requests-status-update",
+        "requests/<pk>/update/status/submit/",
+        ChallengeRequestSubmit.as_view(),
+        name="requests-submit",
+    ),
+    path(
+        "requests/<pk>/update/status/accept-reject/",
+        ChallengeRequestProcess.as_view(),
+        name="requests-process",
     ),
     path(
         "requests/<pk>/update/budget/",
--- a/app/tests/challenges_tests/test_emails.py
+++ b/app/tests/challenges_tests/test_emails.py
@@ -54,7 +54,7 @@ def test_challenge_request_rejected_sent_email(client, challenge_reviewer):
     response = get_view_for_user(
         client=client,
         method=client.post,
-        viewname="challenges:requests-status-update",
+        viewname="challenges:requests-process",
         reverse_kwargs={"pk": request.pk},
         user=challenge_reviewer,
         data={
@@ -84,7 +84,7 @@ def test_challenge_request_accepted_sent_email_challenge_creation(
     response = get_view_for_user(
         client=client,
         method=client.post,
-        viewname="challenges:requests-status-update",
+        viewname="challenges:requests-process",
         reverse_kwargs={"pk": request.pk},
         user=challenge_reviewer,
         data={
--- a/app/tests/challenges_tests/test_forms.py
+++ b/app/tests/challenges_tests/test_forms.py
@@ -181,12 +181,12 @@ def test_accept_challenge_request_duplicate_shortname():
             ChallengeRequest.ChallengeRequestStatusChoices.PENDING,
             True,
         ),
-        (  # Accept
+        (  # Process - Accept
             ChallengeRequest.ChallengeRequestStatusChoices.PENDING,
             ChallengeRequest.ChallengeRequestStatusChoices.ACCEPTED,
             True,
         ),
-        (  # Reject
+        (  # Process - Reject
             ChallengeRequest.ChallengeRequestStatusChoices.PENDING,
             ChallengeRequest.ChallengeRequestStatusChoices.REJECTED,
             True,
--- a/app/tests/challenges_tests/test_permissions.py
+++ b/app/tests/challenges_tests/test_permissions.py
@@ -52,7 +52,8 @@ def test_request_challenge_only_when_verified(client):
     "viewname",
     [
         "challenges:requests-detail",
-        "challenges:requests-status-update",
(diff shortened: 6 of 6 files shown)


app/tests/challenges_tests/test_views.py
```diff
@@ -314,69 +309,6 @@ def test_challenge_card_status(
         assert expected_status in response.rendered_content
 
 
-@pytest.mark.django_db
-def test_challenge_request_workflow(
```
chrisvanrun, line 318: Covered by other tests, this test was a mix of view testing and emails.

app/grandchallenge/challenges/views.py
```diff
@@ -222,15 +232,35 @@ class ChallengeRequestStatusUpdate(
     UpdateView,
 ):
     model = ChallengeRequest
-    form_class = ChallengeRequestStatusUpdateForm
-    permission_required = "change_challengerequest"
-    template_name = "challenges/challengerequest_status_form.html"
     raise_exception = True
+    form_class = ChallengeRequestStatusUpdateForm
```
koopmant, line 236: I think it's probably cleaner to create separate views for this?

app/grandchallenge/challenges/templates/challenges/challengerequest_detail.html
```diff
@@ -20,13 +20,59 @@
 {% endblock %}
 
 {% block content %}
+    {% get_obj_perms request.user for object as "user_perms" %}
+    <div class="px-2 row d-flex">
+        <div class="d-flex ml-auto">
+        {% if object.status == object.ChallengeRequestStatusChoices.DRAFT and 'change_challengerequest' in user_perms %}
+            <a class="btn btn-lg btn-primary"
```
koopmant: I think I'd prefer these in this area somewhere, in regular size and without the rocket/thumbs-up/down icons:
Image

app/grandchallenge/challenges/templates/challenges/challengerequest_detail.html
```diff
@@ -20,13 +20,59 @@
...
+    {% get_obj_perms request.user for object as "user_perms" %}
+    <div class="px-2 row d-flex">
+        <div class="d-flex ml-auto">
+        {% if object.status == object.ChallengeRequestStatusChoices.DRAFT and 'change_challengerequest' in user_perms %}
+            <a class="btn btn-lg btn-primary"
+                href="#"
+                hx-post="{% url 'challenges:requests-status-update' pk=object.pk %}"
+                hx-target="this"
+                hx-vals='{"status": "{{ object.ChallengeRequestStatusChoices.PENDING.value }}"}'
+                hx-confirm="Are you sure you want to submit the request?"
```
koopmant: This is neat!

app/grandchallenge/challenges/templates/challenges/challengerequest_detail.html
```diff
@@ -35,7 +81,10 @@ <h3>[{{ object.short_name }}] {{ object.title  }}</h3>
         <div class="col-lg-6">
             <i class="fa fa-user text-primary pr-3"></i> Challenge request from {{ object.creator|user_profile_link }}<br>
             <i class="fa fa-envelope text-primary pr-3"></i> <a target="_blank" href="mailto:{{ object.contact_email }}">{{ object.contact_email }}</a><br>
-            <i class="fa fa-calendar text-primary pr-3"></i> Submitted on <span class="badge ml-1 px-2 {% if object.created|timedifference < 8 %} badge-success {% elif object.created|timedifference < 15 %} badge-warning {% else %} badge-danger {% endif %}">{{ object.created }}</span><br>
+            <i class="fa fa-calendar text-primary pr-3"></i> Created on <strong>{{ object.created|date:'F j, Y' }}</strong><br>
```
koopmant: Much better date format. It doesn't need to be strong though.

app/grandchallenge/challenges/views.py
koopmant, line 1: ChallengeRequestBudgetUpdate will need `review_challengerequest` permission as well?

app/grandchallenge/challenges/forms.py
koopmant, line 575: Do we need a permission check here? 
Edit: I think not because we only accept a form post from the creator if the status is DRAFT, right?

app/grandchallenge/challenges/models.py
```diff
@@ -963,11 +963,14 @@ class ChallengeRequestStatusChoices(models.TextChoices):
         ACCEPTED = "ACPT", _("Accepted")
         REJECTED = "RJCT", _("Rejected")
         PENDING = "PEND", _("Pending")
+        DRAFT = "DRFT", _("Draft")
+
+    submitted = models.DateTimeField(null=True)
```
koopmant: It's probably good to make this blank as well and then add a contraint that it needs to be set when the status is anthing other than DRAFT. `Invoice` has an example of this, e.g. `issued_on`.

app/grandchallenge/challenges/views.py
chrisvanrun: You mean one for accepting/rejecting, one for submitting? There will be some validation handling required that will be same for both (follow up PR). We could use inheritance for that, albeit you then hide the switch in the urls.py and the target view and the target for the buttons, and you'll still want a bad request if the state isn't matching what you'd expect.

Just to be clear for myself, your suggestion is to have these two urls, and related views?
- `/request/<pk>/submit/`
- `/request/<pk>/review/`

If so, I might play around with that to see if it is any cleaner in the follow up PR.

koopmant: > your suggestion is to have these two urls, and related views?

Yes, that is what I was thinking. Then you don't need an if-statement for the required permission. Indeed, one should probably inherit the other, since they are so similar. 

> If so, I might play around with that to see if it is any cleaner in the follow up PR.

👍

app/grandchallenge/challenges/templates/challenges/challengerequest_detail.html
chrisvanrun: I concur with the place, far better.

I prefer the icons as they indicate action instead of a static status indication. 

As for size, the accept/reject can be regular sized: **we** know where they are. However, I have to insist on keeping the submit for review large. Even though it hurs my own aethetic sensibility. If I could get away with making it flashing and have moving arrows to it, I would. It so such a crucial thing users will need to press:

image

image

koopmant: > instead of a static status indication

Ah good point, I did not realize that. 

> keeping the submit for review large. Even though it hurs my own aethetic sensibility. 

Haha, yeah it's very in your face. But okay!

app/grandchallenge/challenges/forms.py
chrisvanrun: Indeed, permission checks are done on the view level, not the form level.

app/grandchallenge/challenges/models.py
chrisvanrun: I had a scratch on my head on the need for blank. Then I realized I ran into the need when updating it from the site-admin perspective!(Or perhaps there is another one I missed?)

koopmant: > I ran into the need when updating it from the site-admin perspective!

Yes, me too!

app/grandchallenge/challenges/views.py
```diff
@@ -222,29 +232,39 @@ class ChallengeRequestStatusUpdate(
...
+            == self.get_object().ChallengeRequestStatusChoices.PENDING
+        ):
+            return ["review_challengerequest"]
+        elif (
+            self.get_object().status
+            == self.get_object().ChallengeRequestStatusChoices.DRAFT
+        ):
+            return ["change_challengerequest"]
+        else:
+            raise HttpResponseBadRequest()
```
koopmant: HttpResponseBadRequest() is not an exception

app/grandchallenge/challenges/templates/challenges/challengerequest_detail.html
```diff
@@ -20,22 +20,28 @@
...
     </div>
     <br>
     <div class="container-fluid row mb-4 ml-1 bg-light rounded p-3">
         <div class="col-lg-6">
             <i class="fa fa-user text-primary pr-3"></i> Challenge request from {{ object.creator|user_profile_link }}<br>
             <i class="fa fa-envelope text-primary pr-3"></i> <a target="_blank" href="mailto:{{ object.contact_email }}">{{ object.contact_email }}</a><br>
-            <i class="fa fa-calendar text-primary pr-3"></i> Submitted on <span class="badge ml-1 px-2 {% if object.created|timedifference < 8 %} badge-success {% elif object.created|timedifference < 15 %} badge-warning {% else %} badge-danger {% endif %}">{{ object.created }}</span><br>
+            <i class="fa fa-calendar text-primary pr-3"></i> Created on {{ object.created|date:'F j, Y' }}<br>
+            {% if object.submitted %}
+                <i class="fa fa-rocket text-primary pr-3"></i> Submitted on <span class="badge ml-1 px-2 {% if object.submitted|timedifference < 8 %} badge-success {% elif object.submitted|timedifference < 15 %} badge-warning {% else %} badge-danger {% endif %}">{{  object.submitted|date:'F j, Y'}}</span><br>
```
koopmant:
```suggestion
                <i class="fa fa-rocket text-primary pr-3"></i> Submitted on<span class="badge ml-1 px-2 {% if object.submitted|timedifference < 8 %} badge-success {% elif object.submitted|timedifference < 15 %} badge-warning {% else %} badge-danger {% endif %}">{{  object.submitted|date:'F j, Y'}}</span><br>
```
or without `ml-1`

app/grandchallenge/challenges/models.py
```diff
@@ -1205,6 +1214,19 @@ class ChallengeRequestStatusChoices(models.TextChoices):
...
+    class Meta:
+        permissions = [
+            ("review_challengerequest", "Can review challenge request"),
+        ]
+        constraints = [
+            models.CheckConstraint(
+                name="submitted_or_draft",
+                condition=Q(status=ChallengeRequestStatusChoices.DRAFT)
+                | Q(submitted__isnull=False),
+                violation_error_message="When setting the status to something else than a 'Draft', the 'Submitted' must also be set.",
```
koopmant:
```suggestion
                violation_error_message="When setting the status to something other than a 'Draft', 'Submitted' must also be set.",
```

app/grandchallenge/challenges/models.py
```diff
@@ -958,16 +958,25 @@ def submission_pdf_path(instance, filename):
...
+    ChallengeRequestStatusChoices = ChallengeRequestStatusChoices
 
     status = models.CharField(
         max_length=4,
-        choices=ChallengeRequestStatusChoices.choices,
-        default=ChallengeRequestStatusChoices.PENDING,
+        choices=ChallengeRequestStatusChoices,
+        default=ChallengeRequestStatusChoices.DRAFT,
+    )
+    submitted = models.DateTimeField(
```
koopmant:
```suggestion
    submitted_on = models.DateTimeField(
```

app/grandchallenge/challenges/models.py
```diff
@@ -1217,20 +1239,32 @@ def get_absolute_url(self):
 
     def save(self, *args, **kwargs):
         adding = self._state.adding
+        submitting = (
+            self._orig_status == self.ChallengeRequestStatusChoices.DRAFT
+            and self.status != self.ChallengeRequestStatusChoices.DRAFT
+        )
+        if submitting:
+            self.submitted = now()
```
koopmant: now doing this twice?

app/grandchallenge/challenges/models.py
chrisvanrun: Yeah, I agree there with you. Albeit we have 'created' I think adding the _on is a good indication that it is actually a timepoint and not a boolean check or something.

app/grandchallenge/challenges/views.py
chrisvanrun: It certainly isn't! This is very silly of me. I think I first had the `PermissionDenied` here, saw an alternative and went there. But why Chris of old ended up with a full request...

Edit: Wow. The first line sounds like an LLM!

koopmant: > Wow. The first line sounds like an LLM!

Hahahaha. It happens

app/grandchallenge/challenges/models.py
chrisvanrun: Of course, you HAVE to be sure it gets set! All jokes aside, nice catch. Forgot to remove that one.

app/grandchallenge/challenges/models.py
```diff
@@ -1217,20 +1239,31 @@ def get_absolute_url(self):
...
+            and self.status != self.ChallengeRequestStatusChoices.DRAFT
+        )
+        if submitting:
+            self.submitted_on = now()
+
         super().save(*args, **kwargs)
 
         if adding:
             self.assign_permissions()
+        elif submitting:
```
koopmant, line 1253: Will this work for the moment while new requests are immediately submitted? I think it does, right? Because we save it twice in the create view to update from draft to pending.

chrisvanrun: It should. I double-double-checked on the local instance and it still works.

PR #4580, opened 2026-03-03
Handle skipped inference result

koopmant:
Handle job events where the inference result is skipped due to failures during setup. 

Log unexpected errors in the `handle_event` task instead of raising them to ensure the job status is updated.

### DIFF
--- a/app/grandchallenge/components/backends/amazon_sagemaker_base.py
+++ b/app/grandchallenge/components/backends/amazon_sagemaker_base.py
@@ -570,14 +570,14 @@ class AmazonSageMakerBaseExecutor(Executor, ABC):
                 parsed_log = parse_structured_log(
                     log=log_event["message"].replace("\x00", "")
                 )
-                if parsed_log.inference_result_skipped is True:
-                    self._inference_result_skipped = True
                 timestamp = ms_timestamp_to_datetime(log_event["timestamp"])
             except (JSONDecodeError, KeyError, ValueError):
                 logger.warning("Could not parse log")
                 continue
 
             if parsed_log is not None:
+                if parsed_log.inference_result_skipped is True:
+                    self._inference_result_skipped = True
                 output = f"{timestamp.isoformat()} {parsed_log.message}"
                 if parsed_log.source == SourceChoices.STDOUT:
                     stdout.append(output)


app/grandchallenge/components/tasks.py
```diff
@@ -955,7 +955,6 @@ def handle_event(*, event, backend):
             error_message="An unexpected error occurred",
             **get_update_status_kwargs(executor=executor),
         )
-        raise
```
jmsmkn, line 958: We would lose the Sentry error here, so I think needs needs a 

```python3
except Exception as error:
    ...
    logger.error(str(error), exc_info=True)
    return
```

app/grandchallenge/components/backends/amazon_sagemaker_base.py
```diff
@@ -567,8 +568,13 @@ def _set_task_logs(self, *, event):
 
         for log_event in self._get_log_events(event=event):
             try:
+                structured_log = json.loads(
+                    log_event["message"].replace("\x00", "").strip()
+                )
+                if structured_log["inference_result_skipped"] is True:
```
jmsmkn:
```suggestion
                if structured_log.get("inference_result_skipped") is True:
```

jmsmkn: Both so we can parse existing logs and that I suggested making it an optional field for future ones.

jmsmkn: Actually - maybe shift this to `parse_structured_log` and `ParsedLog` so that it is unpacked in one place?

app/grandchallenge/components/tasks.py
koopmant: ah good point, yes

app/grandchallenge/components/backends/amazon_sagemaker_base.py
koopmant: Yes, I can do that. `parse_structured_log` then needs to become a method of `Executor` (or `AmazonSageMakerBaseExecutor`) so that I can set `self._inference_result_skipped`.

jmsmkn: I don't think that it is necessary?

```python3
class ParsedLog(NamedTuple):
    message: str
    source: SourceChoices
    inference_result_skipped: bool | None

def parse_structured_log(*, log: str) -> ParsedLog | None:
    """Parse the structured logs from SageMaker Shim"""
    structured_log = json.loads(log.strip())

    message = structured_log["log"]
    source = SourceChoices(structured_log["source"])
    inference_result_skipped = structured_log.get("inference_result_skipped")

    if structured_log["internal"] is False:
        # Defensive, in case the type of structured_log["internal"] is str
        return ParsedLog(message=message, source=source, inference_result_skipped= inference_result_skipped)

...

parsed_log = parse_structured_log(log=log_event["message"].replace("\x00", ""))

if parsed_log.inference_result_skipped is True:
    self._inference_result_skipped = True
```

koopmant: Ah like that. Okay, that works.

koopmant: Although the log line needs to be external now for us to set the `inference_result_skipped` on `Executor`. This skips internal log lines.

jmsmkn: Yes, I think that is what we want - it has to be set on external messages otherwise the user will not get an error message. Internal messages should never be sent to the user.

PR #4586, opened 2026-03-06
Challenge Request Updating and Submitting

chrisvanrun:
The heftiest of the change sets for friendlier challenge requests. There were two smaller ones prior.

This changset:
- adds an update page for drafts of challenge requests
- splits the challenge requests fields into two groups:
  - Required for submission
  - Required for acceptance (i.e. budget fields)

Most noteworthy change to the actual request is that the budget related fields are replaced by the more extensive use of 'Challenge Setup' that now holds a template for a guided description of the technical aspects of the challenge in order to adapt to the platform:

<details>
<summary> Challenge Setup Template </summary>

```
Please describe the runtime requirements, the amount of test data, and the participation limits for each phase.
This information will be used to calculate a budget estimate. See below example.

## Runtime Requirements
- Participant algorithm max. runtime: 5 minutes
- Participant algorithm max. vRAM: 32 GB
- Participant algorithm GPU types: A10G or T4 or None
- Evaluation method max. runtime: 42 minutes
- Evaluation method max. vRAM: 32 GB
- Evaluation method GPU types: A10G

If some phases have different requirements, please detail them.

## Test data
- Total size of all test cases: 5.4 GB
- Number of cases: 21
- Average output size per case: 100 MB

## Phase 1 - debug
- Expected total number of teams: 20
- Number of submissions per team: 3
- Number of test cases: 1

## Phase 2 - test
- Expected total number of teams: 20
- Number of submissions per team: 1
- Number of test cases: 20
```

</details>

<details>

<summary> Showcase - screenshots and captures (several) </summary>
First request creation page
afbeelding

Detail page
afbeelding

Direct submit (errors)
afbeelding

Update Challenge Request scroll through
<video src="https://github.com/user-attachments/assets/5394fb54-30c9-430a-ab4d-37c827c808cb"></video>

Submitting
<video src="https://github.com/user-attachments/assets/4ee5c039-53ee-4295-ae48-62df85244798"></video>

Accepting - errors
<video src="https://github.com/user-attachments/assets/4444a978-71c5-4c83-8652-a1872b372126"></video>

Accepting
<video src="https://github.com/user-attachments/assets/548e08b2-8ed6-4f67-bb8d-b1915ef964f6"></video>

</details>

### DIFF
--- a/app/grandchallenge/challenges/migrations/0070_assign_review_perm_to_reviewers_group.py
+++ b/app/grandchallenge/challenges/migrations/0070_assign_review_perm_to_reviewers_group.py
@@ -37,6 +37,35 @@ def assign_review_perm_to_reviewers_group(apps, schema_editor):
         )
 
 
+def assign_change_perm_to_creators(apps, schema_editor):
+    ChallengeRequest = apps.get_model(  # noqa: N806
+        "challenges", "ChallengeRequest"
+    )
+    ChallengeRequestUserObjectPermission = apps.get_model(  # noqa: N806
+        "challenges", "ChallengeRequestUserObjectPermission"
+    )
+    Permission = apps.get_model("auth", "Permission")  # noqa: N806
+
+    challenge_requests = ChallengeRequest.objects.select_related(
+        "creator"
+    ).all()
+
+    if not challenge_requests.exists():
+        return
+
+    change_permission = Permission.objects.get(
+        codename="change_challengerequest",
+        content_type__app_label="challenges",
+    )
+
+    for challenge_request in challenge_requests:
+        ChallengeRequestUserObjectPermission.objects.get_or_create(
+            content_object=challenge_request,
+            user=challenge_request.creator,
+            permission=change_permission,
+        )
+
+
 class Migration(migrations.Migration):
 
     dependencies = [
@@ -47,5 +76,9 @@ class Migration(migrations.Migration):
         migrations.RunPython(
             assign_review_perm_to_reviewers_group,
             elidable=True,
-        )
+        ),
+        migrations.RunPython(
+            assign_change_perm_to_creators,
+            elidable=True,
+        ),
     ]
--- a/app/tests/challenges_tests/test_migrations.py
+++ b/app/tests/challenges_tests/test_migrations.py
@@ -4,9 +4,14 @@ import pytest
 from django.apps import apps
 from django.conf import settings
 from django.contrib.auth.models import Group
-from guardian.shortcuts import get_group_perms, remove_perm
+from guardian.shortcuts import get_group_perms, get_user_perms, remove_perm
 
 from tests.factories import ChallengeRequestFactory
+
+_migration = importlib.import_module(
+    "grandchallenge.challenges.migrations"
+    ".0070_assign_review_perm_to_reviewers_group"
+)
 
 
 @pytest.mark.django_db
@@ -26,12 +31,31 @@ def test_assign_review_perm_to_reviewers_group():
         reviewers_group, challenge_request
     )
 
-    migration = importlib.import_module(
-        "grandchallenge.challenges.migrations"
-        ".0070_assign_review_perm_to_reviewers_group"
-    )
-    migration.assign_review_perm_to_reviewers_group(apps, schema_editor=None)
+    _migration.assign_review_perm_to_reviewers_group(apps, schema_editor=None)
 
     assert "review_challengerequest" in get_group_perms(
         reviewers_group, challenge_request
+    )
+
+
+@pytest.mark.django_db
+def test_assign_change_perm_to_creators():
+    challenge_request = ChallengeRequestFactory()
+
+    # The permission is assigned when the object is created; remove it to
+    # simulate the pre-migration state where it was missing.
+    assert "change_challengerequest" in get_user_perms(
+        challenge_request.creator, challenge_request
+    )
+    remove_perm(
+        "change_challengerequest", challenge_request.creator, challenge_request
+    )
+    assert "change_challengerequest" not in get_user_perms(
+        challenge_request.creator, challenge_request
+    )
+
+    _migration.assign_change_perm_to_creators(apps, schema_editor=None)
+
+    assert "change_challengerequest" in get_user_perms(
+        challenge_request.creator, challenge_request
     )


app/grandchallenge/challenges/views.py
```diff
@@ -219,13 +217,37 @@ def get_context_data(self, **kwargs):
         context.update(
             {
                 "fields": fields,
+                # Budget Estimate context
```
koopmant: What does this mean?

app/grandchallenge/challenges/views.py
```diff
@@ -189,25 +179,33 @@ class ChallengeRequestDetail(
...
+                "structures",
+                "challenge_setup",
+                "data_set",
+                "submission_assessment",
+                "challenge_publication",
+                "code_availability",
+                "algorithm_inputs",
+                "algorithm_outputs",
+            )
+            if getattr(self.object, field_name) is not None
```
koopmant: To be honest, I found it a bit odd to see some of these fields were missing on the detail page. Would it not be better to just show them? Then when you click "edit request" you know you will be able to edit all these fields. With this workflow they suddenly appear when you edit the request.

app/grandchallenge/challenges/models.py
```diff
@@ -1237,11 +1273,139 @@ def __init__(self, *args, **kwargs):
     def get_absolute_url(self):
         return reverse("challenges:requests-detail", kwargs={"pk": self.pk})
 
+    def clean(self):
+        super().clean()
+
+        if self.status == self.ChallengeRequestStatusChoices.PENDING:
+            self.clean_for_submission()
+        elif self.status == self.ChallengeRequestStatusChoices.ACCEPTED:
+            self.clean_for_submission()
```
koopmant, line 1305: why does it need to be cleaned for submission again?

app/grandchallenge/challenges/forms.py
```diff
@@ -149,118 +144,97 @@ def clean(self):
...
-        ),
-    )
+class ChallengeRequestForm(SaveFormInitMixin, forms.ModelForm):
+    class Meta:
+        model = ChallengeRequest
+        fields = (
+            "title",
+            "short_name",
+            "contact_email",
+            "abstract",
```
koopmant: Maybe move this one as well

app/grandchallenge/challenges/forms.py
```diff
@@ -331,52 +306,43 @@ class Meta:
...
-                f"please indicate if you plan to coordinate a publication "
-                f"of the challenge results."
-            ),
         }
 
-    def __init__(self, creator, *args, **kwargs):
+    def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.instance.creator = creator
-        self.fields["title"].required = True
```
koopmant, line 349: the title is no longer required now

app/grandchallenge/challenges/forms.py
```diff
@@ -331,52 +306,43 @@ class Meta:
...
-                        "support your challenge.</p>"
-                        "<p>Before you fill out this form, please read our <a href="
-                        "'https://grand-challenge.org/documentation/challenges/'"
-                        "target='_blank'>challenge documentation</a> and our <a href="
-                        "'https://grand-challenge.org/challenge-policy-and-pricing/'"
-                        "target='_blank'>challenge pricing policy</a>.</p><br>"
-                    ),
+                "General Information - Required",
+                *general_required_information,
+                "challenge_fee_agreement",
```
koopmant: maybe a silly question, but why is this not part of general required information?

app/grandchallenge/challenges/static/js/challenges/challenge_request_dynamic_fields.mjs
koopmant, line 1: This appears to no longer work. When "Yes" is selected, the explanation field is still required. 

I think this should also be possible with normal form validation btw, although it will look a bit less fancy.

app/grandchallenge/challenges/templates/challenges/challengerequest_form.html
```diff
@@ -10,17 +10,33 @@
...
+        Thank you for considering to host your challenge on our platform!
+    </p>
+
+    <p>
+        Please use this form to create a draft challenge request. You can update and save the draft before submitting it.
+    </p>
+
+    <p>
+        Your answers in the request will help our team of reviewers decide whether and in what way we can support your challenge.
+    </p>
```
koopmant, line 30:
```suggestion
```

app/grandchallenge/challenges/templates/challenges/challengerequest_update_form.html
```diff
@@ -0,0 +1,36 @@
...
+        <li class="breadcrumb-item active" aria-current="page">Create</li>
+    </ol>
+{% endblock %}
+
+{% block content %}
+    <h2>Update Challenge Request - {{ object.short_name }}</h2>
+
+    <p>
+        Your answers in the request will help our team of reviewers decide whether and in what way we can support your challenge.
+    </p>
```
koopmant, line 24:
```suggestion
```

app/grandchallenge/challenges/templates/challenges/challengerequest_detail.html
```diff
@@ -100,9 +103,15 @@ <h3>[{{ object.short_name }}] {{ object.title }}</h3>
         </div>
     </div>
     {% if object.status == object.ChallengeRequestStatusChoices.DRAFT %}
+        {% if object.can_be_submitted %}
         <div class="alert alert-warning">
-            <i class="fas fa-exclamation-triangle"></i> This challenge request is currently a draft. You can submit it for review using the button at the top of the page.
+            <i class="fas fa-exclamation-triangle"></i> This draft still needs to be submitted! You can submit it for review using the submit button above.
```
koopmant:
```suggestion
            <i class="fas fa-exclamation-triangle"></i>This challenge request is currently a draft. You can submit it for review using the submit button above.
```

app/grandchallenge/challenges/templates/challenges/challengerequest_detail.html
```diff
@@ -100,9 +103,15 @@ <h3>[{{ object.short_name }}] {{ object.title }}</h3>
...
     </div>
     {% if object.status == object.ChallengeRequestStatusChoices.DRAFT %}
+        {% if object.can_be_submitted %}
         <div class="alert alert-warning">
-            <i class="fas fa-exclamation-triangle"></i> This challenge request is currently a draft. You can submit it for review using the button at the top of the page.
+            <i class="fas fa-exclamation-triangle"></i> This draft still needs to be submitted! You can submit it for review using the submit button above.
         </div>
+        {% else %}
+        <div class="alert alert-warning">
+            <i class="fas fa-exclamation-triangle"></i> This draft needs to be submitted but is still incomplete! Edit the draft to complete the required information.
```
koopmant:
```suggestion
            <i class="fas fa-exclamation-triangle"></i> This challenge request is currently a draft and cannot yet be submitted for review because it is incomplete. Edit the request to complete the required information.
```

app/grandchallenge/challenges/forms.py
```diff
@@ -331,52 +313,43 @@ class Meta:
...
 
-    def __init__(self, creator, *args, **kwargs):
+    def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.instance.creator = creator
-        self.fields["title"].required = True
-        self.fields["challenge_fee_agreement"].required = True
+
+        if not self.instance.challenge_setup:
+            self.initial["challenge_setup"] = challenge_setup_template_text
```
chrisvanrun, line 261: A placeholder was purposefully not chosen, so users can copy or edit the template.

app/grandchallenge/challenges/views.py
```diff
@@ -151,26 +154,13 @@ class ChallengeRequestCreate(
 ):
     model = ChallengeRequest
     form_class = ChallengeRequestForm
-    # TODO-challenge-request-draft, update success message
-    # success_message = "A draft of your challenge request has been created!"
-    success_message = "Your challenge request has been created!"
+    success_message = "A draft of your challenge request has been created!"
```
chrisvanrun, line 156: Removal of the earlier 'feature flag'.

app/grandchallenge/challenges/views.py
```diff
@@ -189,25 +179,33 @@ class ChallengeRequestDetail(
...
-        "data_set",
-        "submission_assessment",
-        "challenge_publication",
-        "code_availability",
-        "algorithm_inputs",
-        "algorithm_outputs",
-    )
+
+    @cached_property
+    def detail_view_fields(self):
```
chrisvanrun, line 183: Cleans up the detail view.

app/grandchallenge/challenges/templates/challenges/challengerequest_update_form.html
```diff
@@ -0,0 +1,36 @@
...
+
+{% block title %}
+    Request - Challenges - {{ block.super }}
+{% endblock %}
+
+{% block breadcrumbs %}
+    <ol class="breadcrumb">
+        <li class="breadcrumb-item"><a href="{% url 'challenges:list' %}">Challenges</a></li>
+        <li class="breadcrumb-item"><a href="{% url 'challenges:requests-list' %}">Your Challenge Requests</a></li>
+        <li class="breadcrumb-item active" aria-current="page">Create</li>
```
koopmant: These aren't right.

app/tests/challenges_tests/test_models.py
```diff
@@ -447,3 +447,223 @@ def test_discussion_forum_permissions():
...
+    else:
+        # Should raise a ValidationError containing the expected error message
+        with pytest.raises(ValidationError) as exc_info:
+            challenge_request.clean()
+        error_messages = str(exc_info.value)
+        assert expected_error in error_messages
+
+
+@pytest.mark.django_db
+def test_challenge_request_submitted_field_set_on_status_update():
```
chrisvanrun, line 683: This was, oddly, at `profiles_tests/test_models.py`.

app/tests/factories.py
```diff
@@ -112,12 +112,26 @@ class Meta:
     creator = factory.SubFactory(UserFactory)
     short_name = factory.Sequence(lambda n: f"test-challenge-{n}")
     title = factory.fuzzy.FuzzyText()
-    start_date = fuzzy.FuzzyDate(
+    contact_email = factory.LazyAttribute(
```
chrisvanrun, line 115: Added values here to have a `ChallengeRequest`, by default, pass its own clean even when `ACCEPTED`.

app/grandchallenge/challenges/views.py
chrisvanrun: The context here is not directly used by the view's template but via an included template. Got put on the wrong foot here while tracing it. So thought to add a comment, clearly not clear enough though!

app/grandchallenge/challenges/models.py
chrisvanrun: Consistent state, mostly. If a challenge is accepted it should have and keep all the fields that you'd expect from a full request.

app/grandchallenge/challenges/forms.py
chrisvanrun: Not silly at all! This should be part of that, I think the set was earlier used for required status.

app/grandchallenge/challenges/forms.py
chrisvanrun: True, I don't really see the need to have it required by default. A value is required when submitting.

app/grandchallenge/challenges/views.py
chrisvanrun: I mostly hid them to prevent users focusing on the need to fill out those optional challenge details. What if we instead show all the core required information, but leave out the optional details?

So always show:
```
"title",
"short_name",
"start_date",
"end_date",
"organizers",
"abstract",
"challenge_setup"
 ```
 
 But hide these unless they have a value:
 ```
 "affiliated_event",
  "task_types",
  "modalities",
  "structures",
  "data_set",
  "submission_assessment",
  "challenge_publication",
  "code_availability",
  "algorithm_inputs",
  "algorithm_outputs",
 ```

File: app/grandchallenge/challenges/forms.py (hunk 5 above)
Comment (koopmant) [reply app/grandchallenge/challenges/forms.py:349]: It's required when you create the request though, so might as well keep it required.

File: app/grandchallenge/challenges/static/js/challenges/challenge_request_dynamic_fields.mjs
Comment (chrisvanrun) [reply app/grandchallenge/challenges/static/js/challenges/challenge_request_dynamic_fields.mjs:1]: Oh, in that case this might be finicky... since I just confirmed it works, but I remember that oddly this afternoon it didn't. Investigating!

File: app/grandchallenge/challenges/forms.py (hunk 4 above)
Comment (chrisvanrun) [reply app/grandchallenge/challenges/forms.py]: I think a few lines of description for the challenge is useful to have, even in the very first draft.

What if we change the help_text for this from:

```
 "Provide a summary of the challenge purpose. "
  "This should include a general introduction to the "
  "topic from both a biomedical as well as from a technical point of "
  "view."
```

to 

```
 "Provide a concise summary of the challenge purpose."
```

File: app/grandchallenge/challenges/templates/challenges/challengerequest_form.html (hunk 7 above)
Comment (chrisvanrun) [reply app/grandchallenge/challenges/templates/challenges/challengerequest_form.html:30]: Why would you suggest removing it? I like the few lines of welcome when someone plans to create a request. Sets the tone.

File: app/grandchallenge/challenges/templates/challenges/challengerequest_update_form.html (hunk 14 above)
Comment (chrisvanrun) [reply app/grandchallenge/challenges/templates/challenges/challengerequest_update_form.html]: Good catch!

File: app/grandchallenge/challenges/static/js/challenges/challenge_request_dynamic_fields.mjs
Comment (chrisvanrun) [reply app/grandchallenge/challenges/static/js/challenges/challenge_request_dynamic_fields.mjs:1]: Huh, can't get it NOT to work. Could you double check?

File: app/grandchallenge/challenges/views.py (hunk 2 above)
Comment (koopmant) [reply app/grandchallenge/challenges/views.py]: > What if we instead show all the core required information, but leave out the optional details?

Sounds good to me

File: app/grandchallenge/challenges/forms.py (hunk 5 above)
Comment (chrisvanrun) [reply app/grandchallenge/challenges/forms.py:349]: Fair point. Less noise if we keep it the same.

File: app/grandchallenge/challenges/models.py (hunk 3 above)
Comment (koopmant) [reply app/grandchallenge/challenges/models.py:1305]: All right. The validation error message will be slightly awkward though, because it mentions "to submit" explicitly.

File: app/grandchallenge/challenges/templates/challenges/challengerequest_form.html (hunk 7 above)
Comment (koopmant) [reply app/grandchallenge/challenges/templates/challenges/challengerequest_form.html:30]: It seemed rather obvious to me. (And in general I dislike single line paragraphs.)

But you can keep it if you prefer, no worries!

File: app/grandchallenge/challenges/static/js/challenges/challenge_request_dynamic_fields.mjs
Comment (koopmant) [reply app/grandchallenge/challenges/static/js/challenges/challenge_request_dynamic_fields.mjs:1]: Ah sorry, the page was using a cached version of the old script

File: app/grandchallenge/challenges/views.py (hunk 2 above)
Comment (chrisvanrun) [reply app/grandchallenge/challenges/views.py]: I've added it thus.

File: app/grandchallenge/challenges/forms.py (hunk 4 above)
Comment (koopmant) [reply app/grandchallenge/challenges/forms.py]: I meant move it from the create form to the update form. I agree it should be required, at least for submission. 

But yeah, I agree the help text can be improved as well. I think our users know what an abstract is, so your suggestion should be more than enough.

File: app/grandchallenge/challenges/models.py (hunk 3 above)
Comment (chrisvanrun) [reply app/grandchallenge/challenges/models.py:1305]: That is good point. Hadn't thought of that. We could rename it to 'a submitted'... but that seems odd as well. Since we only see this error in that light, and it should never occur unless someone meddles with a request in the backend I think it is OK?

File: app/grandchallenge/challenges/templates/challenges/challengerequest_form.html (hunk 7 above)
Comment (chrisvanrun) [reply app/grandchallenge/challenges/templates/challenges/challengerequest_form.html:30]: It's a small touch of Anne. So I vote to keep it =)

File: app/grandchallenge/challenges/forms.py (hunk 4 above)
Comment (chrisvanrun) [reply app/grandchallenge/challenges/forms.py]: > I meant move it from the create form to the update form. I agree it should be required, at least for submission.

I read it as such. IMO having a few lines for the general contextualisation of a draft is worth forcing the users to provide it when creating it in the first place. Could guide a follow-up of us about the idea of hosting.

Comment (koopmant) [reply app/grandchallenge/challenges/forms.py]: > is worth forcing the users to provide

I believe the point of the create form is to make it as easy as possible? Besides, when are we going to look at peoples drafts?

Anyway, if people don't want to fill this in, they won't. So whether we keep it here or not is not that important.

Comment (chrisvanrun) [reply app/grandchallenge/challenges/forms.py]: > I believe the point of the create form is to make it as easy as possible? Besides, when are we going to look at peoples drafts?

A valid point. Drafts can be overdue, so we'll be looking at them sometimes via that avenue. If an abstract then will help... likely not, we'll just email the user (hopefully automatically) asking them what to do with the draft.

File: app/grandchallenge/challenges/models.py (hunk 3 above)
Comment (koopmant) [reply app/grandchallenge/challenges/models.py:1305]: yes, it's fine

File: app/grandchallenge/challenges/forms.py (hunk 4 above)
Comment (koopmant) [reply app/grandchallenge/challenges/forms.py]: Ah okay. I thought we would just ignore everything in draft state.

File: app/grandchallenge/challenges/templates/challenges/challengerequest_update_form.html
Comment (koopmant) [inline app/grandchallenge/challenges/templates/challenges/challengerequest_update_form.html:1]: The result is at this point is somehow very busy. I don't have a suggestion though. It might be possible to present this in a better way. But if you don't have an alternative, it's fine. 

Image

File: app/grandchallenge/challenges/templates/challenges/challengerequest_detail.html
Comment (koopmant) [inline app/grandchallenge/challenges/templates/challenges/challengerequest_detail.html:1]: The order and title for the items is different from the update field. It would be nice if they are consistent. And I think the required fields are still hidden when they are None?

Image
Image

File: app/grandchallenge/challenges/views.py [hunk 17]
```diff
@@ -238,19 +253,22 @@ class ChallengeRequestStatusUpdate(
...
-        else:
-            challenge = None
-        send_challenge_status_update_email(
-            challengerequest=form.instance, challenge=challenge
-        )
+        response = HttpResponse()
+        response["HX-Refresh"] = "true"
+        return response
+
+    def form_invalid(self, form):
```
koopmant, line 274: I'm getting this validation error when I'm trying the manual detail items route (not a structered form):

"Either a structured challenge submission form needs to be uploaded or the following fields are required to submit a challenge request: Data License Extra"

app/grandchallenge/challenges/forms.py
```diff
@@ -111,35 +109,33 @@ def clean(self):
         return cleaned_data
 
 
-general_information_items_1 = (
+general_required_information = (
     "title",
     "short_name",
     "contact_email",
```
koopmant, line 115: Validation for invalid email does not show the tooltip, not sure if that's an issue for this form particularly. Maybe just create a separate issue for it. 

Image

app/grandchallenge/challenges/views.py
chrisvanrun: Huh. Investigating!

chrisvanrun: Ah, forgot that ofc was a boolean and should have been differently handled than the other fields. Fixed in the next push.

app/grandchallenge/challenges/forms.py
chrisvanrun: Works for me, both on Safari and Firefox, both the initial creation and the updating. Are you missing the tooltip shown below, or are you referring to something else?

https://github.com/user-attachments/assets/e926acf9-a768-4adc-9cd0-4d8fc7751993

app/grandchallenge/challenges/templates/challenges/challengerequest_detail.html
chrisvanrun: Good catch in the hiding of some fields. I found a switch I hadn't noticed yet in the template itself which was already there. It is now python-only that determines the order and label.

app/grandchallenge/challenges/templates/challenges/challengerequest_update_form.html
chrisvanrun: I agree with it being busy. I had noticed it but thought perhaps it was permissible. Marking a choice obvious in a form is always a bit tricky. If we both are annoyed by it, we should try to make it a bit friendlier on the eyes. That is the goal of the pitch after all.

chrisvanrun: Adding a collapsible or tabs here doesn't really fit with the style. It makes it more difficult to find any errors plus implementing it would be very verbose since it is currently `crispy.helper.layout` based. 

Instead, I played around a bit with spacing. How is this?

afbeelding

Would it help with adding some light background to bind the elements visually together? Or is that a total mess?:
afbeelding

koopmant: > Would it help with adding some light background to bind the elements visually together?

Yes, that's much better I think! We have a small issue with the legend text now, it being on the edge like that; although somehow it doesn't bother me.

app/grandchallenge/challenges/forms.py
koopmant: That's what I mean yes. It wasn't working for me this morning, but I haven't tested without extensions or anything. In any case, if it works on those browsers it's a different issue. I'll look into it later.

app/grandchallenge/challenges/static/js/challenges/challenge_request_dynamic_fields.mjs
```diff
@@ -2,18 +2,18 @@ function updateExtraField(fieldName, helpText) {
...
-    const checkbox = document.getElementById(checkboxFieldId);
-    if (checkbox.checked) {
+
+    const select = document.getElementById(checkboxFieldId);
+    if (select.value === "true") {
         document.getElementById(extraFieldDiv).style.display = "none";
         document.getElementById(extraFieldId).required = false;
     } else {
         document.getElementById(extraFieldDiv).style.display = "block";
-        document.getElementById(extraFieldId).required = true;
```
koopmant, line 11: I think it's probably good to keep this one to block updating the draft without it when the answer is no? But then saving a draft becomes a bit more annoying maybe...
Or we should write a better validation error message on submission when this is missing, because right now the message only mentions "data license extra" and we replace the title for the input there, so it's not obvious what data license extra means.

app/grandchallenge/challenges/templates/challenges/challengerequest_update_form.html
chrisvanrun: I spent slightly too long tweaking the colour. I think this is a nice subtle nudge:

afbeelding

app/grandchallenge/core/static/css/core.css
```diff
@@ -176,9 +177,20 @@ a[aria-expanded="false"] .fa-chevron-down {
...
 .htmx-request.htmx-indicator {
     display: inline-block;
 }
+
+/* Fixes a large whitespace for fieldset legends */
+fieldset > legend {
+    width: inherit;
+}
+
+.bg-light-gray {
```
koopmant: I think bootstraps bg-light is good enough? it's a little bit darker, but still readable.

app/grandchallenge/challenges/models.py
```diff
@@ -1013,21 +1035,31 @@ class ChallengeRequest(UUIDModel, ChallengeBase):
         max_length=255,
     )
     challenge_setup = models.TextField(
-        help_text="Describe the challenge set-up."
+        help_text="Describe the challenge set-up.",
```
koopmant: since we have a migration anyway, may I suggest this? (I know we aren't using this...)
```suggestion
        help_text="Describe the challenge setup.",
```

app/grandchallenge/challenges/forms.py
```diff
@@ -311,17 +270,17 @@ class Meta:
...
-                f"uploaded to Grand Challenge (read more about that <a href='https://grand-challenge.org/documentation/"
-                f"data-storage/' target='_blank'>here</a>)."
+                "Please describe the training and test datasets you are planning to "
+                "use. <br>In order to evaluate the submitted algorithms, the test dataset will need to be "
+                "uploaded to Grand Challenge (read more about that <a href='https://grand-challenge.org/documentation/"
+                "data-storage/' target='_blank'>here</a>)."
             ),
             "challenge_setup": (
-                "Describe the challenge set-up. How many tasks "
+                "Describe the challenge set-up from a technical stand-point. How many tasks "
```
koopmant:
```suggestion
                "Describe the challenge setup from a technical standpoint. How many tasks "
```

app/grandchallenge/challenges/forms.py
```diff
@@ -149,118 +145,93 @@ def clean(self):
...
+        model = ChallengeRequest
+        fields = (
+            "title",
+            "short_name",
+            "contact_email",
+        )
+        labels = {
+            "short_name": "Acronym",
+        }
+        help_texts = {
```
koopmant, line 159: Maybe we should update the field help texts in the model instead. Saves a little bit of duplication as well.

app/grandchallenge/core/static/css/core.css
chrisvanrun: Still readable, but IMO the help text loses too much contrast. And it is a bit painful on the eyes tbh. Note: this colour code comes from the `--light` var and isn't a random pick.

koopmant: oh, that's odd. Something is broken in our theme then somehow. because that should be the same I think?

For know we can override it like this I suppose, though we should at least use `var(--light);`

app/grandchallenge/challenges/static/js/challenges/challenge_request_dynamic_fields.mjs
chrisvanrun: > Or we should write a better validation error message on submission when this is missing, because right now the message only mentions "data license extra" and we replace the title for the input there, so it's not obvious what data license extra means.

Should be a simple solution.

app/grandchallenge/core/static/css/core.css
chrisvanrun: Ah, I hadn't noticed bootstrap uses multiple sources of vars (couldn't get the `var(--light)` to work nicely. The code is equal to $gray-100 on the scss context. So I'll just use that.

app/grandchallenge/challenges/forms.py
chrisvanrun: That is good point. Since we are migrating anyway, this is a nice moment to do so!

app/grandchallenge/core/static/css/core.css
koopmant: Yeah, I just noticed as well. Our Flatly theme updates the light color to gray-200. Maybe name the css class likewise (bg-gray-100 or something)?

app/grandchallenge/challenges/forms.py
chrisvanrun: Done. `ChallengeBase` is also used for `Challenge` so I've left the form-level labels for `ChallengeBase.short_name` and `ChallengeBase.title` untouched.

app/grandchallenge/core/static/css/core.css
chrisvanrun: Done!

app/grandchallenge/challenges/models.py
```diff
@@ -1013,21 +1037,52 @@ class ChallengeRequest(UUIDModel, ChallengeBase):
         max_length=255,
     )
     challenge_setup = models.TextField(
-        help_text="Describe the challenge set-up."
+        help_text=(
+            "Describe the challenge set-up from a technical stand-point. How many tasks "
```
koopmant:
```suggestion
            "Describe the challenge setup from a technical standpoint. How many tasks "
```

app/grandchallenge/challenges/forms.py
```diff
@@ -149,118 +145,93 @@ def clean(self):
...
+                "(e.g., https://{acronym}.grand-challenge.org/). No spaces and special "
+                "characters allowed. We prefer a single word with two digits at "
+                "the end indicating the year (e.g. LUNA26). See "
+                "<a href='https://www.grand-challenge.org/challenges' "
+                "target='_blank'>other challenges</a> for examples."
+            ),
+            "contact_email": (
+                "The email address that potential participants can use to contact you with questions about the challenge. "
+                "This email address will be visible on the challenge website, so please provide an address that you are comfortable sharing publicly."
+            ),
```
koopmant:
```suggestion
```

app/grandchallenge/challenges/models.py
```diff
@@ -1237,11 +1306,160 @@ def __init__(self, *args, **kwargs):
...
+        ):
+            if self.data_license is False and not self.data_license_extra:
+                raise ValidationError(
+                    "You need to explain why you are not willing/able to use a CC-BY license.",
+                )
+
+    def clean_challenge_fee_agreement(self):
+        if not self.challenge_fee_agreement:
+            raise ValidationError(
+                "You need to agree to the challenge fee agreement to submit a challenge request.",
```
koopmant:
```suggestion
                "You need to agree to the challenge pricing policy to submit a challenge request.",
```

app/grandchallenge/challenges/views.py
```diff
@@ -189,43 +178,80 @@ class ChallengeRequestDetail(
...
         context = super().get_context_data()
-        fields = {
-            field.verbose_name: field.value_to_string(self.object)
-            for field in self.object._meta.fields
-            if field.name in self.detail_view_fields
+        fields = {}
+        label_overrides = {
+            "challenge_setup": "Challenge technical setup",
+            "short_name": "Acronym",
+            "challenge_fee_agreement": "Read Pricing Policy",
```
koopmant:
```suggestion
            "challenge_fee_agreement": "Pricing Policy Agreement",
```

app/grandchallenge/challenges/templates/challenges/challengerequest_detail.html
```diff
@@ -100,19 +103,23 @@ <h3>[{{ object.short_name }}] {{ object.title }}</h3>
...
-            {% endif %}
+        {% for label, value in fields.items %}
+            <tr><td class="font-weight-bold">{{ label|title }}</td> <td>{{ value|linebreaks }}</td></tr>
         {% endfor %}
         {% if object.data_license %}
-            <tr><td class="font-weight-bold">Public data license</td> <td><i class="fas fa-check-circle text-success pr-2"></i>  The organizers agree to publish their public data set under a CC-BY license.</td></tr>
+            <tr><td class="font-weight-bold">Public data license</td> <td><i class="fas fa-check-circle text-success pr-2"></i> The organizers agree to publish their public data set under a CC-BY license.</td></tr>
         {% elif object.data_license == None %}
             <tr><td class="font-weight-bold">Public data license</td> <td>Unknown</td></tr>
         {% else %}
```
koopmant:
```suggestion
{% elif object.data_license == None %}
{% else %}
```

app/grandchallenge/challenges/models.py
```diff
@@ -1181,6 +1236,15 @@ class ChallengeRequest(UUIDModel, ChallengeBase):
     )
     data_license = models.BooleanField(
         default=False,
+        null=True,
```
koopmant, line 1232:
```suggestion
default=None,
        null=True,
```

app/tests/factories.py
```diff
@@ -1,12 +1,12 @@
 import datetime
 import hashlib
+import random
 
 import factory
```
koopmant:
```suggestion
import factory.fuzzy
```

This is needed, otherwise the "sort-of deprecated" fuzzy module is not imported.

app/grandchallenge/challenges/models.py
chrisvanrun: Aah. silly! That got changed back ofc since I didn't rebase it initially! >.<

app/tests/factories.py
chrisvanrun: Sort-of deprecated. Huh. Weird.

app/grandchallenge/challenges/models.py
chrisvanrun: Huh, good catch. Wonder why it was `False` in the first place.

Edit: because we added null=True here ofc, so it wasn't an option prior!

app/grandchallenge/challenges/templates/challenges/challengerequest_detail.html
chrisvanrun: Fair point. Fyi, I considered making the data_license something always required. However, it seems the structured submission form has a section on that, so it isn't really needed.

app/grandchallenge/challenges/models.py
koopmant, line 1: Just noticed this. None of the text based fields should have `null=True`. There should be ten of them, including the IdentifierField. I only noticed it because when I edit a new request and change nothing but save it, the "None" paragraphs on the detail view disappear.

app/grandchallenge/challenges/forms.py
```diff
@@ -392,147 +303,41 @@ def __init__(self, creator, *args, **kwargs):
...
+    def clean_challenge_setup(self):
+        challenge_setup = self.cleaned_data.get("challenge_setup")
 
-        return instance
+        if (
+            challenge_setup
+            and challenge_setup.strip().splitlines()
+            == challenge_setup_template_text.strip().splitlines()
+        ):
+            return None
```
koopmant: See the other comment. Without null=True this should become:
```suggestion
            return ""
```

app/grandchallenge/challenges/models.py
chrisvanrun: Ah. Classic mistake here I made with null/"" both representing empty. Need to set defaults here as well. On it!

koopmant: I think you won't need any defaults. If no value is provided it defaults to "" automatically.

chrisvanrun: Huh, hadn't realized that `blank=True, default=""` and `blank=True` are equivalent. TIL!

app/grandchallenge/challenges/forms.py
```diff
@@ -149,118 +145,89 @@ def clean(self):
...
+Please describe the runtime requirements, the amount of test data, and the participation limits for each phase.
+This information will be used to calculate a budget estimate. See below example.
+
+## Runtime Requirements
+- Participant algorithm max. runtime: 5 minutes
+- Participant algorithm max. vRAM: 32 GB
+- Participant algorithm GPU types: A10G or T4 or None
+- Evaluation method max. runtime: 42 minutes
+- Evaluation method max. vRAM: 32 GB
+- Evaluation method GPU types: A10G
```
koopmant, line 187:
```suggestion
- Participant algorithm max. runtime: 5 minutes
- Participant algorithm max. DRAM: 32 GB
- Participant algorithm GPU types: A10G or T4 or None
- Evaluation method max. runtime: 42 minutes
- Evaluation method max. DRAM: 32 GB
- Evaluation method GPU types: A10G
```

koopmant (approved):
✨ LGTM ✨

PR #4624, opened 2026-03-30
Optimise 'Your work' list queries

chrisvanrun:
Part of pitch:
- https://github.com/DIAGNijmegen/rse-roadmap/issues/229

This change set applies some optimalisations for DB calls in the views for 'Your Challenges', 'Your Algorithms', et cetera.

See the pitch for a breakdown of the different optimalisations profiling.

### DIFF
--- a/app/grandchallenge/algorithms/views.py
+++ b/app/grandchallenge/algorithms/views.py
@@ -228,7 +228,7 @@ class UsersAlgorithmList(
                 user_role_order=models.Case(
                     models.When(
                         editors_group__pk__in=user_group_pks,
-                        then=models.Value(3),
+                        then=models.Value(2),
                     ),
                     default=models.Value(1),
                     output_field=models.IntegerField(),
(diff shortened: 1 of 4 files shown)


app/grandchallenge/algorithms/views.py
```diff
@@ -215,17 +216,22 @@ def get_queryset(self):
             )
         )
 
-        user_groups = self.request.user.groups.all()
-
         if not self.request.user.is_superuser:
+            user_group_pks = list(
+                self.request.user.groups.values_list("pk", flat=True)
```
jmsmkn, line 221: I'm not sure that the profiling is representative as the group pks are serialised to Python. Did you try when the user is a member of 1000's of groups? That can happen where the user has lots of jobs. Testing locally will also miss out on the network hop for that.

chrisvanrun: > I'm not sure that the profiling is representative as the group pks are serialised to Python. Did you try when the user is a member of 1000's of groups?

In this case, the local user in question was part of ~5,000 archive-related groups (maybe +2 for registered user etc?). Execution wise it took <5 ms to get them. 

> Testing locally will also miss out on the network hop for that.

True, profiling the splitting of a query should also account for the 'flat' extra cost of a network hop to make actually deliver and return the query. Not sure what the typical round-trip latency to the database is?

jmsmkn: I've added a comment in the pitch for the full timings, but this PR is slower than the original just on a pure SQL level, never mind the network and python processing. I have added a suggestion in there that is the most efficient and keeps all the work in the database.

PR #4659, opened 2026-04-13
Resolve invalid combination of endpoint config options

koopmant:
This resolves the error: "ManagedInstanceScaling is not supported with the given EndpointConfig setup."


koopmant:
Found this in testing last Friday. This worked before, but it didn't do anything because the combination is not valid. AWS probably rolled out stricter validation.

amickan (approved):
LGTM

PR #4663, opened 2026-04-13
Avoid line break after euro sign

koopmant:
See https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/672

### DIFF
--- a/app/grandchallenge/core/templatetags/costs.py
+++ b/app/grandchallenge/core/templatetags/costs.py
@@ -1,5 +1,6 @@
 from django import template
 from django.conf import settings
+from django.utils.safestring import mark_safe
 
 register = template.Library()
 
@@ -7,7 +8,7 @@ register = template.Library()
 @register.filter
 def euro(value):
     try:
-        return f"€ {value:,.2f}"
+        return mark_safe(f"€&nbsp;{value:,.2f}")
     except ValueError:
         return ""
 
@@ -15,7 +16,7 @@ def euro(value):
 @register.filter
 def euro_no_cents(value):
     try:
-        return f"€ {value:,.0f}"
+        return mark_safe(f"€&nbsp;{value:,.0f}")
     except ValueError:
         return ""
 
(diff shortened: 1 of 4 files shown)


app/grandchallenge/core/templatetags/costs.py
```diff
@@ -1,21 +1,22 @@
...
+from django.utils.safestring import mark_safe
 
 register = template.Library()
 
 
 @register.filter
 def euro(value):
     try:
-        return f"€ {value:,.2f}"
+        return mark_safe(f"€&nbsp;{value:,.2f}")
```
jmsmkn: Please never use `mark_safe`, especially with user provided data! The way to do this is to use `format_html`.

jmsmkn (changes requested):

app/grandchallenge/core/templatetags/costs.py
jmsmkn: Reason is that this can potentially be a vector for CSS attacks, `format_html` will properly escape strings.

app/grandchallenge/core/templatetags/costs.py
```diff
@@ -1,21 +1,22 @@
 from django import template
 from django.conf import settings
+from django.utils.safestring import mark_safe
 
 register = template.Library()
 
 
 @register.filter
 def euro(value):
```
jmsmkn: Maybe add an argument that accepts the precision of the number that is displayed, can then be used for high precision unit costs and for replacing `euro_no_cents`.

app/grandchallenge/core/templatetags/costs.py
koopmant: *XSS right? 

Anyway, thank you!

jmsmkn: Oops, obviously not 100% still...

jmsmkn (approved):
LGTM!

PR #4668, opened 2026-04-14
Standardize showing notification created time

### LINKED ISSUE #4666: Standardize showing date and time for notification
When I go through my notifications it can be unclear from when the notification is. For most notifications it is somewhere in the title (.. weeks ago) but it is in different places in the sentence:

`Your upload from 1 month ago failed with the following error: Image validation for socket Generic Overlay failed with error: 1 file could not be imported.`

`Your registration request for test-DICOM-archive was accepted 2 months, 2 weeks ago.`

And for some it is not in the notification at all so it is unclear whether it was just triggered or old. This can be useful to know for these types of notification:

`Validation for socket Points of interest failed: The file could not be decoded`

It think it makes sense to standardize showing the date/time in notifications separate from the notification message.

HarmvZ:
- Shows notification created time in a separate right aligned column like on github and removes it from the notification text.
- Adds a fixture for creation of notifications for ease of testing.

Demo:

image

closes #4666


scripts/notification_fixtures.py
jmsmkn, line 1: Not sure about the addition of this - these scripts tend to rot quickly as they're not tested and rarely used, and are easy to add with LLMs now.

jmsmkn: It also reaches into the notification internals rather than having the notifications that are created through the actual code paths, so only really useful for styling. Also deletes notifications which could have been useful for development.

HarmvZ: Ok, your call. I generated it to easily test this PR and thought it might be useful for others.

jmsmkn: I think remove it, it's only really useful when changing all the notification rendering, this is the first time we've really done that in 5 years and it is anyway easy to generate with an LLM.

jmsmkn (approved):
LGTM!

PR #4671, opened 2026-04-15
Switch forge algorithm templates to use the invoke API

### LINKED ISSUE #4669: Switch the templates to use the invocation API
The invoke container API was added in https://github.com/DIAGNijmegen/rse-grand-challenge/pull/4469. It would be good if this is now made the default for templates so that algorithms and methods created this year are future proof.

koopmant:
This updates the algorithm templates to use the invoke API. 

Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/493

I used Kiro to improve the test_run script and to polish the (hand-)updated docstrings and comments. Tests were hand-written. All code has been reviewed by me.


koopmant:
I still had this branch from some time ago. It has (almost?) everything. I was still trying to improve the test_run script to only start the container once and invoke the model multiple times; right now it restarts the container for each interface. 

I had not yet figured out how to deal with the inputs and outputs when invoking multiple times. It might be easiest to add the test inputs and outputs folder to the container and swap the input for each interface to a single input folder and move the output to the outputs folder. Otherwise it's a headache with file permissions. 

I don't think there is a clean way of including a test script for exec mode. It would just be confusing. If a user wants to revert to exec mode, they'll have to adjust the test script themselves.

jmsmkn:
> It might be easiest to add the test inputs and outputs folder to the container and swap the input for each interface to a single input folder and move the output to the outputs folder.

In prod we do this with links from `/input` to the actual input location for each invocation, same with output.

> I don't think there is a clean way of including a test script for exec mode. It would just be confusing. If a user wants to revert to exec mode, they'll have to adjust the test script themselves.

We should not offer templates or an option for exec mode, there is no good reason to use it over invoke.

chrisvanrun:
The `LABEL` being used as switch finally had the 2 cents rolling down (or up!) on the way things are set up. =P

amickan:
After this is merged, we might want to update the onboarding powerpoint as well as the documentation to reflect the new default.

app/grandchallenge/forge/templates/forge/partials/example_algorithm/app.py.template
```diff
@@ -0,0 +1,139 @@
...
+
+The lifecycle on Grand Challenge is:
+
+  1. Container starts → your server boots and loads the model
+  2. /health is polled until it returns 200 (timeout: ~5 minutes)
+  3. For each case:
+       a. Inputs are placed at /input (previous inputs/outputs are cleared)
+       b. POST /invoke is called
+       c. Your code reads /input, processes, writes to /output, returns 201
+  4. Container is kept warm for a while in case more cases arrive
```
koopmant: Is it?

app/grandchallenge/forge/templates/forge/partials/example_algorithm/app.py.template
```diff
@@ -0,0 +1,139 @@
...
+
+Key points:
+  - Load your model during startup (in the lifespan context), not during
+    /invoke. Each /invoke call has its own timeout, and spending that time
+    loading a model will cause failures.
```
koopmant:
```suggestion
    loading a model may cause failures.
```

app/grandchallenge/forge/templates/forge/partials/example_algorithm/app.py.template
```diff
@@ -0,0 +1,139 @@
...
+
+@asynccontextmanager
+async def lifespan(app: FastAPI):
+    """Server lifespan: load the model at startup, release at shutdown.
+
+    Everything before `yield` runs before the server accepts requests.
+    This is where you should do expensive initialization (loading weights,
+    warming up the GPU, etc.) so that /invoke can run as fast as possible.
+    """
+    MODELS["answer_to_everything"] = init_model()
```
koopmant: I figured a clear example keeping the torch example with nn.Linear was important and I hadn't really completed this, but if we're really set on making this joke the model should always return 42. Something like this could work:
```
with torch.no_grad():
    model.weight.zero_()
    model.bias.fill_(42)

for p in model.parameters():
    p.requires_grad = False
```

app/grandchallenge/forge/templates/forge/partials/example_algorithm/app.py.template
```diff
@@ -0,0 +1,139 @@
...
+
+
+@app.post("/invoke")
+async def invoke():
+    """Invoke endpoint — called once per case.
+
+    When this is called, the case's input files are available at /input.
+    Read them, run inference, write results to /output, and return 201.
+
+    This may be called multiple times (once per case) without the container
```
koopmant: Will it be confusing that in general there will be only one case per job, so this is usually only called once? I'm not sure if we can rephrase this to avoid any potential confusion about that.

app/grandchallenge/forge/templates/forge/partials/example_algorithm/do_test_run.sh.template
```diff
@@ -1,90 +1,299 @@
...
-      --volume "$OUTPUT_DIR":/output \
-      --entrypoint /bin/sh \
-      $DOCKER_IMAGE_TAG \
-      -c "chmod -R -f o+rwX /output/* || true"
-
-    # Ensure volume is removed
-    docker volume rm "$DOCKER_NOOP_VOLUME" > /dev/null
-}
+# How long to wait for the container's /health endpoint to come up.
+# These match the timeouts used on Grand Challenge — do not change.
```
koopmant: I remember during testing that the 10 second timeout is quite annoying during testing, because the initial check will usually fail, even if you're loading a tiny model. Not sure if it's worth adding anything about that, but "do not change" seems like an empty statement; the user can do whatever they like.

app/grandchallenge/forge/templates/forge/partials/example_algorithm/do_test_run.sh.template
```diff
@@ -1,90 +1,299 @@
...
+HEALTH_CHECK_MAX_ATTEMPTS=30
+HEALTH_CHECK_DELAY_SECONDS=5
+HEALTH_CHECK_TIMEOUT_SECONDS=10
+
+# How long a single /invoke call is allowed to run.
+# Note that this does NOT correspond to the maximum runtime for the job set on Grand Challenge
+# The maximum runtime on GC includes I/O as well as model and auxiliary data loading
```
koopmant: Not sure if this is clear now. Might be good to get input from the 10% team.

app/grandchallenge/forge/templates/forge/partials/example_algorithm/do_test_run.sh.template
```diff
@@ -1,90 +1,299 @@
...
+start_container() {
+    log "Starting container"
+
+    # Extra arguments worth calling out:
+    #   --network <isolated>   no internet access; the tester sidecar issues
+    #                          HTTP requests to the container by name via
+    #                          Docker's embedded DNS
+    #   --volume <vol>:/tmp    scratch space (Grand Challenge disallows writes
+    #                          elsewhere outside the mounted directories)
+    #   --volume model:/opt/ml/model:ro   the (optional) model tarball
```
koopmant: nitpick, text not aligned here.

app/grandchallenge/forge/templates/forge/partials/example_algorithm/do_test_run.sh.template
```diff
@@ -1,90 +1,299 @@
...
+    local docker_run_args=(
+        --detach
+        --name "$CONTAINER_NAME"
+{% if not no_gpus %}        --gpus all{% endif %}
```
koopmant: Can we resolve the double negation while we're here? Not sure where no_gpus comes from...

app/grandchallenge/forge/templates/forge/partials/example_algorithm/do_test_run.sh.template
```diff
@@ -1,90 +1,299 @@
...
+    # Fix permissions so the host user can read the output files.
+    # The container may have written them as a different UID on Linux.
+    docker exec --user root "$CONTAINER_NAME" \
+        /bin/sh -c "chmod -R -f o+rX /output/*"
+
+    # Copy from the staging directory to the host output directory
+    cp -r "$STAGING_OUTPUT_DIR/." "${destination}/"
+}
+
+log() {
```
koopmant: Maybe set this to top like in do_save

koopmant (commented):
Looks good! Some questions and nitpicks.

app/grandchallenge/forge/templates/forge/partials/example_algorithm/app.py.template
jmsmkn: No it isn't.

app/grandchallenge/forge/templates/forge/partials/example_algorithm/app.py.template
```diff
@@ -0,0 +1,139 @@
...
+        print(f.read())
+
+    return model
+
+
+MODELS = {}
+
+
+@asynccontextmanager
+async def lifespan(app: FastAPI):
```
jmsmkn, line 56: Maybe make it clear that this only runs once and it should not be called directly?

app/grandchallenge/forge/templates/forge/partials/example_algorithm/app.py.template
jmsmkn: I think this is fine as it covers the future batch use case.

It is maybe missing the examples of how to read and write data (aren't there existing examples of this? Those can be called with the model passed along), it should be clear that anything other than the status code returned by this method is ignored.

app/grandchallenge/forge/templates/forge/partials/example_algorithm/app.py.template
```diff
@@ -0,0 +1,139 @@
...
+    print("Collecting Torch CUDA information")
+    print(f"Torch CUDA is available: {(available := torch.cuda.is_available())}")
+    if available:
+        print(f"\tnumber of devices: {torch.cuda.device_count()}")
+        print(f"\tcurrent device: { (current_device := torch.cuda.current_device())}")
+        print(f"\tproperties: {torch.cuda.get_device_properties(current_device)}")
+    print("=+=" * 10)
+
+
+def init_model():
```
jmsmkn: Maybe move this to another file? Users basically shouldn't touch anything else in the app. The lifecycle could then read init_model, and pass it to their inference method.

app/grandchallenge/forge/templates/forge/partials/example_algorithm/app.py.template
jmsmkn: Not yet at least, but may as well keep it there as in the future it will be if the admins opt in to it.

amickan: I wasn't sure about this. What I meant here is that it stays up and running until all cases have been processed. I guess I should just say that.

jmsmkn: Remember that the invoke container api is different from the invocations of algorithm endpoints. The container will only be kept alive for challenges when we implement batch mode, and will only be kept alive for algorithms if endpoints are in use. It will never be kept alive for the current algorithm jobs.

amickan: Ah yes, I keep forgetting this. Hm, then I think this point should be removed after all. I'll think about this some more tomorrow.

app/grandchallenge/forge/templates/forge/partials/example_algorithm/app.py.template
amickan: I was actually wondering if we need the dictionary approach here since there will always just be one model right? I pushed a change that simplifies this to just a `MODEL` variable.

koopmant: No we don't really need it. It came from the FastAPI example here: https://fastapi.tiangolo.com/advanced/events/?h=lifespan#lifespan

app/grandchallenge/forge/templates/forge/partials/example_algorithm/do_test_run.sh.template
amickan: Ok, I have not experienced issues with it, but probably good to make a comment here about it, and of course, you're right that people can do whatever they want anyway.

amickan: I think just removing the "do not change" is good enough.

app/grandchallenge/forge/templates/forge/partials/example_algorithm/do_test_run.sh.template
amickan: Yes. I have added some more information, but let's indeed wait and see what the 10% team thinks about this.

app/grandchallenge/forge/templates/forge/partials/example_algorithm/do_test_run.sh.template
amickan: It's only used in 2 places and set in one place, I've renamed it to `use_gpus`

app/grandchallenge/forge/templates/forge/partials/example_algorithm/app.py.template
amickan: After thinking about this some more, I think this will be confusing. I think it makes more sense to change it to describe the current way. Once we have implemented batch mode, we can update this accordingly.

app/grandchallenge/forge/templates/forge/partials/example_algorithm/app.py.template
amickan: I have removed "once per case". I think it is already true that it can be called multiple times, so fine to leave that as is, I would say. I have, however, adjusted the general lifecycle description at the beginning of the file to explain that we only process one case at a time. This part will then need to be updated once batch mode is enabled.

amickan: Regarding examples of how to read and write data, those are in inference.py

app/grandchallenge/forge/templates/forge/partials/example_algorithm/app.py.template
amickan: I like that idea. I think moving it to inference.py would make most sense. That's then the only file that users need to adapt.

amickan:
This is ready for re-review.

app/grandchallenge/forge/templates/forge/partials/example_algorithm/do_test_run.sh.template
koopmant: Yeah think so too.

app/grandchallenge/forge/templates/forge/partials/example_algorithm/app.py.template
```diff
@@ -0,0 +1,102 @@
...
+  - The server must listen on port 4743.
+  - /health must NOT redirect (302 is treated as a failure).
+  - /invoke must return exactly 201 on success. Any other status is a failure.
+  - Your container will have no internet access at runtime.
+
+Any implementation will do — FastAPI is used here as an example. The only
+requirements are the two endpoints and the LABEL in the Dockerfile.
+
+You should not need to edit this file. Instead, edit inference.py to implement
+your algorithm's init_model() and run() functions.
```
koopmant: I'd put this at the top.

koopmant (commented):
Small suggestion, but looks good otherwise. I cannot approve this as I'm the author.

app/grandchallenge/forge/templates/forge/partials/example_algorithm/app.py.template
```diff
@@ -0,0 +1,102 @@
...
+  2. /health is polled until it returns 200 (timeout: ~5 minutes)
+  3. Inputs are placed at /input
+  4. POST /invoke is called
+  5. Your code reads /input, does a forward pass, writes to /output, returns 201
+  6. Container stops and outputs are uploaded to Grand Challenge
+
+Key points:
+  - Load your model during startup (in the lifespan context), not during
+    /invoke. The /invoke call has its own timeout, and spending that time
+    loading a model may not leave sufficient time for processing. Separating model loading
```
chrisvanrun: I'd not reference future work. That just creates additional docs we'll forget to update.

app/grandchallenge/forge/templates/forge/partials/example_algorithm/app.py.template
```diff
@@ -0,0 +1,102 @@
...
+    from inference makes your container ready for future optimizations where multiple cases
+    may be processed without restarting the container.
+  - The server must listen on port 4743.
+  - /health must NOT redirect (302 is treated as a failure).
```
chrisvanrun: Return codes make perfect sense for us. They will not for the average user.

app/grandchallenge/forge/templates/forge/partials/example_algorithm/app.py.template
chrisvanrun: Signed!

app/grandchallenge/forge/templates/forge/partials/example_algorithm/do_test_run.sh.template
```diff
@@ -1,90 +1,300 @@
...
-    docker volume rm "$DOCKER_NOOP_VOLUME" > /dev/null
-}
+# How long to wait for the container's /health endpoint to come up.
+# These match the timeouts used on Grand Challenge.
+HEALTH_CHECK_MAX_ATTEMPTS=30
+HEALTH_CHECK_DELAY_SECONDS=5
+HEALTH_CHECK_TIMEOUT_SECONDS=10
+
+# How long a single /invoke call is allowed to run.
+# Note that this is NOT equivalent to the maximum runtime set for jobs on Grand Challenge
```
chrisvanrun, line 37: Smart to note this!

app/grandchallenge/forge/templates/forge/partials/example_algorithm/do_test_run.sh.template
```diff
@@ -1,90 +1,300 @@
...
+# The maximum runtime on GC includes I/O as well as model and auxiliary data loading.
```
chrisvanrun: I'd always use long notation.

app/grandchallenge/forge/templates/forge/partials/example_algorithm/do_test_run.sh.template
```diff
@@ -1,90 +1,300 @@
...
-      --entrypoint /bin/sh \
-      $DOCKER_IMAGE_TAG \
-      -c "rm -rf /output/* || true"
-else
-  mkdir -p -m o+rwX "${OUTPUT_DIR}/{{ interface_name }}"
-fi
+    provision "{{ interface_name }}"
+    invoke
+    collect_output "{{ interface_name }}"
+    log "Wrote results to ${OUTPUT_DIR}/{{ interface_name }}"
```
chrisvanrun: Should this log be part of the collect outputs?

app/grandchallenge/forge/templates/forge/partials/example_algorithm/app.py.template
amickan: I agree that the reference is not necessary, but these docs will need to be updated either way once the batch mode is implemented.

app/grandchallenge/forge/templates/forge/partials/example_algorithm/do_test_run.sh.template
amickan: That would make sense.

app/grandchallenge/forge/templates/forge/partials/example_algorithm/do_test_run.sh.template
amickan: What do you mean?

app/grandchallenge/forge/templates/forge/partials/example_algorithm/app.py.template
amickan: I have changed this in all places to, e.g. `HTTP 200 (OK)`

app/grandchallenge/forge/templates/forge/partials/example_algorithm/do_test_run.sh.template
chrisvanrun: Sorry. Forgot that github only highlights full lines. I meant 'GC'.

chrisvanrun (approved):
LGTM!

PR #4674, opened 2026-04-16
Rename "time" to "time_index" in annotations

chrisvanrun:
I think we are still on time(_index) with this.

In relation to:
- https://github.com/DIAGNijmegen/rse-roadmap/issues/472


chrisvanrun:
Per request of Paul, who is working a quick fix for CIRRUS.

jmsmkn (approved):
LGTM. We would need an admin follow up to remove any CIVs that use the old definition.

jmsmkn:
Please add a reference to the issue and pitch. Are you certain all locations are covered? I think that I would have expected to see more migrations.

chrisvanrun:
> Are you certain all locations are covered? I think that I would have expected to see more migrations.

AFAIK it is only buried in the JSON validation of the `Answer` validation. I followed the same route as the change set as the [original `"time"` PR](https://github.com/DIAGNijmegen/rse-grand-challenge/pull/4604/changes). A quick (double checking) cursory lookup for `"time"` has no related hits in the codebase.

CIV's and related don't show up in the migrations since the values/files are mainly validated away from the database.

pkcakeout:
Matching CIRRUS PR open and agreed upon with @HarmvZ - https://github.com/DIAGNijmegen/rse-cirrus-core/pull/3862

Needs more work but that "more work" will be done later and does not affect the result of the rename.

miriam-groeneveld:
I know Ward has been trying out this feature, so there will be some data with the `time` part, he is aware that we are changing it.

PR #4676, opened 2026-04-20
Update error message for corrupt container images

amickan:
This can happen for algorithms and evaluations. If it happens for a method container, saying that the "algorithm image" failed to start might be confusing.


jmsmkn (approved):
LGTM!

PR #4677, opened 2026-04-20
Update maximum segments length

### LINKED ISSUE #119: Increase support for 64+ segmentations.
A user would like to parse segmentations of [TotalSegmenter](https://github.com/wasserth/TotalSegmentator/tree/master). There are 117 classes to that segmentation and rse-panimg is only supporting up to 64 at the moment.

jmsmkn:
Closes https://github.com/DIAGNijmegen/rse-panimg/issues/119


app/grandchallenge/components/models.py
```diff
@@ -347,9 +346,9 @@ def _validate_voxel_values(self, image):
             raise ValidationError(
                 "Image segments could not be determined, ensure the voxel values "
                 "are integers and that it contains no more than "
-                f"{MAXIMUM_SEGMENTS_LENGTH} segments. Ensure the image has the "
-                "minimum and maximum voxel values set as tags if the image is a TIFF "
-                "file."
+                f"{len(self.overlay_segments_allowed_values)} segments. "
```
chrisvanrun, line 349: This error message will look odd if the `overlay_segments` aren't defined on every seg. CI.

IIRC some CIs don't have these set: setting the would be best but we can't always retrace what they should be.

I still think it is nice to add but we might need to build the error message up a bit creatively.

chrisvanrun (changes requested):

app/grandchallenge/components/models.py
jmsmkn: > This error message will look odd if the overlay_segments aren't defined on every seg. CI.

That is checked a couple of lines further up, they're not validated in that case:

```
if not self.overlay_segments:
    return
```

chrisvanrun (dismissed):
LGTM!

PR #4678, opened 2026-04-20
Update allowed file types for predictions file

amickan:
The allowed file types for user uploads on the `SubmissionForm` should match those checked in `prepare_and_execute_evaluation`, I think.


jmsmkn:
No we need to allow more - the browser can report different file types to the os (via libmagic).

PR #4681, opened 2026-04-22
Centralize common system error messages

amickan:
This replaces common currently hard-coded system error messages with an enum `SystemErrorMessages`

### DIFF
--- a/app/grandchallenge/algorithms/serializers.py
+++ b/app/grandchallenge/algorithms/serializers.py
@@ -298,7 +298,7 @@ class JobPostSerializer(JobSerializer):
             else:
                 error_handler = job.get_error_handler()
                 error_handler.handle_error(
-                    error_message=SystemErrorMessages.UNEXPECTED_ERROR,
+                    error_message=SystemErrorMessages.UNEXPECTED_ERROR.label,
                 )
                 logger.error(e, exc_info=True)
 
--- a/app/grandchallenge/algorithms/serializers.py
+++ b/app/grandchallenge/algorithms/serializers.py
@@ -33,7 +33,7 @@ from grandchallenge.components.serializers import (
     HyperlinkedComponentInterfaceValueSerializer,
     convert_deserialized_civ_data,
 )
-from grandchallenge.core.error_handlers import SystemErrorMessages
+from grandchallenge.core.error_messages import SystemErrorMessages
 from grandchallenge.core.guardian import filter_by_permission
 from grandchallenge.core.templatetags.remove_whitespace import oxford_comma
 from grandchallenge.hanging_protocols.serializers import (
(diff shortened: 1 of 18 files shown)


app/grandchallenge/algorithms/serializers.py
```diff
@@ -297,7 +298,7 @@ def create(self, validated_data):
             else:
                 error_handler = job.get_error_handler()
                 error_handler.handle_error(
-                    error_message="An unexpected error occurred",
+                    error_message=SystemErrorMessages.UNEXPECTED_ERROR.label,
```
jmsmkn: The developer having to remember to type `.label` each time I think is asking for trouble. Could this be simplified? Maybe defining `error_messages.py` and then `UNEXPECTED_ERROR = "..."`. Then it could be used with:

```python3
from grandchallenge.core.error_messages import UNEXPECTED_ERROR

...

error_handler.handle_error(error_message=UNEXPECTED_ERROR)
```

Bit cleaner. Maybe you would need some sort of structure for comparisons, or making sure that they can't be modified, something like:

```python3
>>> @dataclass(frozen=True)
... class _C:
...     M = "message"
...     
>>> C = _C()
>>> raise RuntimeError(C.M)
Traceback (most recent call last):
  File "<python-input-18>", line 1, in <module>
    raise RuntimeError(C.M)
RuntimeError: message
>>> C.M= "sadf"
Traceback (most recent call last):
  File "<python-input-19>", line 1, in <module>
    C.M= "sadf"
    ^^^
  File "<string>", line 15, in __setattr__
dataclasses.FrozenInstanceError: cannot assign to field 'M'
```

amickan: Yes, I was a bit irritated by having to type .label every time. I thought about just adding simple variables for each error message, but that felt too unstructured and then I defaulted to using what we always use for enums... but maybe in this case, we could just go with: 


```python
class SystemErrorMessages:
    UNEXPECTED_ERROR = "An unexpected error occurred"
    TIME_LIMIT_EXCEEDED = "Time limit exceeded"
    MEMORY_LIMIT_EXCEEDED = (
        "The container was killed as it exceeded its memory limit"
    )
```

Groups them together, but avoids having to use .label when using them.

This is less protected than the frozen dataclass approach you are suggesting, but that does feel a little overkill to me.

jmsmkn: That's true, but I would prefer to avoid someone potentially overwriting or changing the error message through assignment. We need to be careful with errors as that is the part of the codebase that is rarely tested, as the majority of the time the path is avoided.

amickan: Ok, fair enough. I think the `frozen=True` protects only instance attributes, not class-level attributes, so you could still do `_C.M= "sadf"` and that would not raise an error. I guess the risk that we would be using the protected class directly is very low, but I was googling and I think `StrEnum` does what we want and might be the best option?

jmsmkn: Yes, good find!

jmsmkn (approved):
LGTM!

PR #4682, opened 2026-04-23
Account for per-job output storage for budget estimations

chrisvanrun:
As mentioned here:
- https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/318

This adds an `average_size_job_output_mb_for_tasks` as a factor when estimating the required budget.

Existing budget estimations have this migrated to be set to zero.

Shown in the table as:
image


app/tests/challenges_tests/test_models.py
```diff
@@ -327,20 +351,26 @@ def test_challenge_request_budget_calculation(settings):
...
-        == 33309.53
+        == 34290.41
     )
 
     for i_phase in range(2):
         assert (
-            challenge_request.compute_and_storage_costs_euros_for_phases[
-                i_phase
-            ]
+            pytest.approx(
```
chrisvanrun, line 369: This was giving some false positives on differences. So added an approx here as well.

app/grandchallenge/challenges/models.py
```diff
@@ -1601,15 +1626,35 @@ def total_compute_time(self):
 
     @cached_property
     def data_storage_size_gb_for_phases(self):
```
chrisvanrun, line 1735: Not sure if it was better to split this into input/output. Opted to keep this in a single property for now.

chrisvanrun:
Ping!

koopmant (commented):
Pong! Looks good. Probably good to rebase first though. Should we update the `challenge_setup_template_text` on the challenge request form?

chrisvanrun:
> Pong! Looks good. Probably good to rebase first though. Should we update the `challenge_setup_template_text` on the challenge request form?

Thanks! I don't think we need to update the default text there, as it already has this snippet:

```
## Test data
- Total size of all test cases: 5.4 GB
- Number of cases: 21
- Average output size per case: 100 MB
```

chrisvanrun:
Force pushed a rebase.

koopmant:
Needs migration fix

chrisvanrun:
> Needs migration fix

It does! On it.

koopmant (dismissed):
LGTM!

chrisvanrun:
Hah, merge one branch. The other one gets stale!

koopmant (approved):

PR #4685, opened 2026-04-28
Centralize evaluation error messages

amickan:
In preparation for the action message builder I have been working on, it makes sense to also centralize the evaluation error messages. This class contains all error messages for which we want a different action message than for custom evaluation errors.  

### DIFF
--- a/app/grandchallenge/evaluation/tasks.py
+++ b/app/grandchallenge/evaluation/tasks.py
@@ -21,7 +21,10 @@ from grandchallenge.core.celery import (
     acks_late_2xlarge_task,
     acks_late_micro_short_task,
 )
-from grandchallenge.core.error_messages import SystemErrorMessages
+from grandchallenge.core.error_messages import (
+    EvaluationErrorMessages,
+    SystemErrorMessages,
+)
 from grandchallenge.core.exceptions import LockNotAcquiredException
 from grandchallenge.core.utils.query import check_lock_acquired
 from grandchallenge.core.validators import get_file_mimetype
@@ -91,9 +94,7 @@ def check_prerequisites_for_evaluation_execution(*, evaluation_pk):
     if any(jobs.values()):
         evaluation.update_status(
             status=Evaluation.CANCELLED,
-            error_message="There are non-successful jobs for this submission. "
-            "These need to be handled first before you can "
-            "re-evaluate. Please contact support.",
+            error_message=EvaluationErrorMessages.UNSUCCESSFUL_JOBS,
         )
         return
     else:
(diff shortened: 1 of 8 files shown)


amickan:
We could also add these to `SystemErrorMessages`, but they are specific to evaluations and are not system error messages in the literal sense, so I went with a separate class.

app/grandchallenge/evaluation/tasks.py
```diff
@@ -171,8 +172,8 @@ def prepare_and_execute_evaluation(*, evaluation_pk):
         else:
             evaluation.update_status(
                 status=Evaluation.FAILURE,
-                stderr=f"{mimetype} files are not supported.",
-                error_message=f"{mimetype} files are not supported.",
+                stderr=f"{mimetype} {EvaluationErrorMessages.UNSUPPORTED_INPUT}",
```
jmsmkn: This construction is a little strange. The user shouldn't get here as they should be rejected on upload. I don't think that it is worth having this exception to the rule, and we should just have one error message for this.

amickan: True, it is a rare error, but it does happen (which is why I had created this [PR](https://github.com/DIAGNijmegen/rse-grand-challenge/pull/4678) last week). 

> we should just have one error message for this

What do you mean with one error message? The reason I added it is that the fallback action message will be the one for custom method errors and for those we send admins to the evaluation logs. That would be a very strange / misleading thing to do for this error. Of course, it would only affect very few evaluations, but I figured it's a small effort to add the exception.

jmsmkn: I think having the error messages be consistent is a great benefit. Then you do not need to worry about partial matching, and everything works the same. "Input file type is not supported" would be fine I think.

amickan: I agree. I have removed the mimetype from the error message

jmsmkn (changes requested):
This branch is a target for #4686, so that needs to be merged first, no?

amickan:
> This branch is a target for #4686, so that needs to be merged first, no?

True. I just thought it's easier to review this one without the changes from the other PR merged into it, but I have merged the other branch now.

jmsmkn:
Then it may be better to target a clean branch so that changes can be merged in once reviewed? Then you don't have a mix of reviewed and unreviewed changes. Merging the previous PR would have dismissed my review, so with this setup it would anyway need to be reviewed again.

jmsmkn (approved):
LGTM!

amickan:
> Then it may be better to target a clean branch so that changes can be merged in once reviewed? Then you don't have a mix of reviewed and unreviewed changes. Merging the previous PR would have dismissed my review, so with this setup it would anyway need to be reviewed again.

You're right. A feature branch as target for both PRs would have been better. Next time! The remaining PRs for this can be sequential and will just target main, I think.

PR #4686, opened 2026-04-28
Add EvaluationActionMessageBuilder

amickan:
This adds the logic to build the action message for an evaluation given a specific user. 

### DIFF
--- a/app/grandchallenge/evaluation/models.py
+++ b/app/grandchallenge/evaluation/models.py
@@ -1946,16 +1946,6 @@ class EvaluationManager(ComponentJobManager):
 
 class EvaluationActionMessageBuilder:
 
-    HANDLERS = {
-        SystemErrorMessages.UNEXPECTED_ERROR: "_handle_unexpected",
-        SystemErrorMessages.MEMORY_LIMIT_EXCEEDED: "_handle_resource_limit",
-        SystemErrorMessages.TIME_LIMIT_EXCEEDED: "_handle_resource_limit",
-        EvaluationErrorMessages.ALGORITHM_FAILURE: "_handle_algorithm_failure",
-        EvaluationErrorMessages.UNSUCCESSFUL_JOBS: "_handle_reevaluation_blocker",
-        EvaluationErrorMessages.INTERFACE_MISMATCH: "_handle_reevaluation_blocker",
-        EvaluationErrorMessages.UNSUPPORTED_INPUT: "_handle_invalid_input",
-    }
-
     def __init__(self, *, evaluation, user):
         self.evaluation = evaluation
         self.phase = evaluation.submission.phase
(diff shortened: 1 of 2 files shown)


app/grandchallenge/evaluation/models.py
```diff
@@ -1941,6 +1944,198 @@ def get_evaluations_with_same_inputs(
...
+            Evaluation.CANCELLED,
+        ):
+            return
+
+        if not self.evaluation.error_message:
+            return
+
+        for error_message, handler_name in self.HANDLERS.items():
+            if error_message in self.evaluation.error_message:
+                return getattr(self, handler_name)()
```
jmsmkn: I find this construction complex. It uses strings to couple implementations, which IDE's will not follow. I think that it would be easier to follow and maintain with an `if`/`elif` statement.

app/grandchallenge/evaluation/models.py
```diff
@@ -1941,6 +1944,198 @@ def get_evaluations_with_same_inputs(
...
+        if self.is_admin:
+            return format_html(
+                "The participant needs to resubmit with the correct input format."
+            )
+        return format_html(
+            "{icon} Please make sure your submission is in the correct format and resubmit.",
+            icon=self.exclamation_mark,
+        )
+
+    def build_action_message(self, *, user):
```
jmsmkn: Unused?

app/grandchallenge/evaluation/models.py
```diff
@@ -1941,6 +1944,198 @@ def get_evaluations_with_same_inputs(
...
+
+    def _handle_custom(self):
+        if self.is_admin:
+            if self.is_algorithm_phase:
+                return format_html(
+                    "The participant's algorithm ran successfully but your evaluation "
+                    "method failed. Please {logs_message} if necessary.",
+                    logs_message=self.evaluation_logs_message,
+                )
+            return format_html(
```
jmsmkn: Nit: I don't like the style where the negatives just fall through. I find things easier to read if `else:` blocks are used, then the two actions from the if statement are clear from their indentation.

app/grandchallenge/evaluation/models.py
amickan: I used it for testing, but forgot to remove it before pushing. This will be added in the next PR that updates the templates.

app/grandchallenge/evaluation/models.py
amickan: I actually had that initially and then asked an LLM for improvement suggestions, haha... but I totally see your point and will change it

jmsmkn: I guessed. I think the things you need to ask from LLMs need to be specific, general "improvements" are almost always subjective, and if you've written something then that comes with the many years of context you have, and you certainly know better than an LLM.

app/grandchallenge/evaluation/models.py
```diff
@@ -1941,6 +1944,226 @@ def get_evaluations_with_same_inputs(
...
+            Evaluation.CANCELLED,
+        ):
+            return
+
+        if not self.evaluation.error_message:
+            return
+
+        if (
+            SystemErrorMessages.UNEXPECTED_ERROR
+            in self.evaluation.error_message
```
jmsmkn: Do these need to be `in`, or can they be `==`?

amickan: For most error messages we can use exact matching. The only 2 that would not match are `INVALID_INPUT` and `UNSUCCESSFUL_JOBS`. 

The other PR changes `INVALID_INPUT` entirely anyway, so doesn't matter anymore. 
`UNSUCCESSFUL_JOBS` are not super common and as you argued on the pitch, we don't need to worry about past evaluations so much, so I think, yes, we can just go for exact matching.

jmsmkn (approved):
LGTM!

PR #4692, opened 2026-04-29
Add Invocation model

koopmant:


app/config/settings.py
```diff
@@ -1170,6 +1170,7 @@ def get_private_ip():
     "ALGORITHM_ENDPOINTS_OUTPUT_BUCKET_NAME",
     "grand-challenge-algorithm-endpoints-outputs",
 )
+ALGORITHM_ENDPOINTS_MAXIMUM_INVOCATION_DURATION = 24 * 60 * 60  # 24 hours
```
jmsmkn: Isn't this a bit long? I think that the limit in AWS is 1 hour (https://docs.aws.amazon.com/sagemaker/latest/dg/async-inference-invoke-endpoint.html) but even that is more than what should be needed for an "interactive" algorithm.

app/grandchallenge/algorithms/models.py
```diff
@@ -1728,3 +1728,72 @@ class EndpointGroupObjectPermission(GroupObjectPermissionBase):
...
+    CANCELLED = "CANCELLED", _("Cancelled")
+
+
+class Invocation(UUIDModel):
+    StatusChoices = InvocationStatusChoices
+
+    endpoint = models.ForeignKey(Endpoint, on_delete=models.PROTECT)
+    status = models.CharField(
+        max_length=11,
+        choices=InvocationStatusChoices,
```
jmsmkn, line 1749: Should have a database constraint.

app/grandchallenge/algorithms/models.py
```diff
@@ -1728,3 +1728,72 @@ class EndpointGroupObjectPermission(GroupObjectPermissionBase):
...
+                limit_value=settings.ALGORITHM_ENDPOINTS_MAXIMUM_INVOCATION_DURATION
+            ),
+        ],
+    )
+    invoke_duration = models.DurationField(
+        null=True,
+        default=None,
+        blank=True,
+        help_text=(
+            "The duration of the invocation, if measured. "
```
jmsmkn: It will always be measured, right?

app/grandchallenge/algorithms/models.py
```diff
@@ -1728,3 +1728,72 @@ class EndpointGroupObjectPermission(GroupObjectPermissionBase):
...
+            "users implementation. "
+            "Includes input data loading time, "
+            "processing time, output data writing time and "
+            "any delays from shared hardware issues."
+        ),
+    )
+    error_message = models.CharField(max_length=1024, default="", blank=True)
+    inputs = models.ManyToManyField(
+        to=ComponentInterfaceValue,
+        related_name="algorithms_endpoints_as_input",
```
jmsmkn: The related name doesn't seem correct here - this is an invocation, not an endpoint?

app/grandchallenge/algorithms/models.py
```diff
@@ -1728,3 +1728,72 @@ class EndpointGroupObjectPermission(GroupObjectPermissionBase):
...
+    )
+    outputs = models.ManyToManyField(
+        to=ComponentInterfaceValue,
+        related_name="algorithms_endpoints_as_output",
```
jmsmkn: Same here.

app/grandchallenge/algorithms/models.py
jmsmkn: Maybe also an index for the cloudwatch stats?

app/grandchallenge/algorithms/models.py
```diff
@@ -1728,3 +1728,72 @@ class EndpointGroupObjectPermission(GroupObjectPermissionBase):
...
+    QUEUED = "QUEUED", _("Queued")
+    PROVISIONED = "PROVISIONED", _("Provisioned")
+    EXECUTING = "EXECUTING", _("Executing")
+    EXECUTED = "EXECUTED", _("Executed")
+    FAILURE = "FAILURE", _("Failure")
+    SUCCESS = "SUCCESS", _("Success")
+    CANCELLED = "CANCELLED", _("Cancelled")
+
+
+class Invocation(UUIDModel):
```
jmsmkn, line 1743: Should have a default sort on created along with an index for that?

app/config/settings.py
koopmant: Yes, I had meant to put a comment here to get a better limit. For an interactive algorithm an hour is far too long, but for an endpoint in general I suppose it would be fine. I can put it at 5 minutes until we have other use cases besides interactive algorithms.

app/grandchallenge/algorithms/models.py
koopmant: right

app/grandchallenge/algorithms/models.py
koopmant: Both added

app/grandchallenge/algorithms/models.py
koopmant: Thanks, changed it

app/grandchallenge/algorithms/models.py
koopmant: Indeed, thanks!

jmsmkn (approved):
LGTM!

PR #4693, opened 2026-04-30
Update submission detail page

amickan:
This updates the submission detail page in the following ways:
- Moves non-admin stuff out of the admin section
- Updates the evaluation table to include "next steps", "details" and "visibility" columns

Screenshot 2026-04-30 at 10 43 49

### DIFF
--- a/app/grandchallenge/evaluation/models.py
+++ b/app/grandchallenge/evaluation/models.py
@@ -2475,21 +2475,23 @@ class Evaluation(CIVForObjectMixin, ComponentJob):
     def visibility_icon(self):
         if self.status == self.SUCCESS:
             if self.published and self.rank > 0:
-                return format_html(
-                    '<i class="fas fa-eye text-success" title="This result is published on the leaderboard."></i>'
-                )
+                classes = "fas fa-eye text-success"
+                title = "This result is published on the leaderboard."
             elif self.published:
-                return format_html(
-                    '<i class="fas fa-eye text-warning" title="This result is not visible on the leaderboard yet because it has not been ranked yet or the ranking failed."></i>'
-                )
+                classes = "fas fa-eye text-warning"
+                title = "This result is not visible on the leaderboard yet because it has not been ranked yet or the ranking failed."
             else:
-                return format_html(
-                    '<i class="fas fa-hourglass-half text-warning" title="This result is under review by the challenge admins."></i>'
-                )
+                classes = "fas fa-hourglass-half text-warning"
+                title = "This result is under review by the challenge admins."
         else:
-            return format_html(
-                '<i class="fas fa-eye-slash text-danger" title="There is no result for this submission."></i>'
-            )
+            classes = "fas fa-eye-slash text-danger"
+            title = "There is no result for this submission."
+
+        return format_html(
+            '<i class="{classes}" title="{title}"></i>',
+            classes=classes,
+            title=title,
+        )
 
 
 class EvaluationUserObjectPermission(UserObjectPermissionBase):
--- a/app/grandchallenge/evaluation/models.py
+++ b/app/grandchallenge/evaluation/models.py
@@ -2107,7 +2107,7 @@ class EvaluationActionMessageBuilder:
                 )
             else:
                 return format_html(
-                    "{icon} Please {contact_message} for more information.",
+                    "{icon} {contact_message} for more information.",
                     icon=self.exclamation_mark,
                     contact_message=self.organiser_contact_message,
                 )
@@ -2465,6 +2465,31 @@ class Evaluation(CIVForObjectMixin, ComponentJob):
     @property
     def utilization(self):
         return self.evaluation_utilization
+
+    def build_action_message(self, *, user):
+        return EvaluationActionMessageBuilder(
+            evaluation=self, user=user
+        ).build()
+
+    @property
+    def visibility_icon(self):
+        if self.status == self.SUCCESS:
+            if self.published and self.rank > 0:
+                return format_html(
+                    '<i class="fas fa-eye text-success" title="This result is published on the leaderboard."></i>'
+                )
+            elif self.published:
+                return format_html(
+                    '<i class="fas fa-eye text-warning" title="This result is not visible on the leaderboard yet because it has not been ranked yet or the ranking failed."></i>'
+                )
+            else:
+                return format_html(
+                    '<i class="fas fa-hourglass-half text-warning" title="This result is under review by the challenge admins."></i>'
+                )
+        else:
+            return format_html(
+                '<i class="fas fa-eye-slash text-danger" title="There is no result for this submission."></i>'
+            )
 
 
 class EvaluationUserObjectPermission(UserObjectPermissionBase):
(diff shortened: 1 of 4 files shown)


app/grandchallenge/evaluation/templates/evaluation/submission_detail.html
```diff
@@ -55,6 +58,36 @@ <h2>Submission {{ object.pk }}</h2>
...
+                <a href="{{ object.supplementary_file.url }}"><i class="fa fa-file"></i> {{ object.supplementary_file.name }}</a>
+            </dd>
+        {% endif %}
+
+        {% if object.predictions_file %}
+            <dt>Submission File</dt>
+            <dd>
+                <a href="{{ object.predictions_file.url }}"><i class="fa fa-download"></i> {{ object.predictions_file.name }}</a>
+            </dd>
+        {% endif %}
```
jmsmkn: Shouldn't these only be shown conditionally based on the phase settings?

Image

app/grandchallenge/evaluation/templates/evaluation/submission_detail.html
```diff
@@ -55,6 +58,36 @@ <h2>Submission {{ object.pk }}</h2>
...
+
+        {% if object.supplementary_file %}
+            <dt>Supplementary File ({{ object.phase.supplementary_file_label }})</dt>
+            <dd>
+                <a href="{{ object.supplementary_file.url }}"><i class="fa fa-file"></i> {{ object.supplementary_file.name }}</a>
+            </dd>
+        {% endif %}
+
+        {% if object.predictions_file %}
```
jmsmkn, line 88: Isn't this only accessible by challenge admins?

app/grandchallenge/evaluation/templates/evaluation/submission_detail.html
```diff
@@ -66,34 +99,85 @@ <h3>Evaluations for this submission</h3>
...
-                <th>Result</th>
+                <th>Visibility</th>
+                <th class="nonSortable">Details</th>
+                <th class="nonSortable">Possible next step(s)</th>
             </tr>
             </thead>
             <tbody>
             {% for evaluation in object.evaluation_set.all %}
                 <tr>
+                    <td>{{ evaluation.pk }}</td>
```
jmsmkn, line 126: Maybe use `|split_first:"-"`?

app/grandchallenge/evaluation/templates/evaluation/submission_detail.html
```diff
@@ -66,34 +99,85 @@ <h3>Evaluations for this submission</h3>
...
                     <td data-order="{{ evaluation.created|date:"c" }}">{{ evaluation.created }}</td>
                     <td data-order="{{ evaluation.modified|date:"c" }}">{{ evaluation.modified }}</td>
-                    <td><a href="{{ evaluation.get_absolute_url }}">{{ evaluation.id }}</a>
+                    {% if "change_challenge" in challenge_perms %}
+                        <td>
+                            <a href="{{ evaluation.method.get_absolute_url }}">{{ evaluation.method.id|split_first:"-" }}</a>
```
jmsmkn:
```suggestion
                            <a href="{{ evaluation.method.get_absolute_url }}">{{ evaluation.method.pk|split_first:"-" }}</a>
```

app/grandchallenge/evaluation/templates/evaluation/submission_detail.html
jmsmkn: These have been moved out of a section that was only visible to admins, we have to be really careful when we change visibility for things like this.

app/grandchallenge/evaluation/templates/evaluation/submission_detail.html
amickan: But participants don't have a way of getting the full pk if we do that. They might need it to reference the evaluation when they contact support or the admins, right?

app/grandchallenge/evaluation/models.py
```diff
@@ -2466,6 +2466,31 @@ def create_utilization(self):
...
+    def build_action_message(self, *, user):
+        return EvaluationActionMessageBuilder(
+            evaluation=self, user=user
+        ).build()
+
+    @property
+    def visibility_icon(self):
+        if self.status == self.SUCCESS:
+            if self.published and self.rank > 0:
+                return format_html(
```
jmsmkn: Django will soon fail when using `format_html` with only one argument. There will be a warning about this if it is covered by a test. Maybe add variables here (`classes` and `title`), set them in the switch statement, then return one `format_html` at the end of the method.

app/grandchallenge/evaluation/templates/evaluation/submission_detail.html
amickan: It's accessible to anyone with view permission for the submission. At least that is the permission we check when serving the file.

app/grandchallenge/evaluation/templates/evaluation/submission_detail.html
amickan: Oh, I wasn't aware of those settings. For admins, we should always show them though.

app/grandchallenge/evaluation/templates/evaluation/submission_detail.html
jmsmkn: It's usually enough as putting that into the search in the table will hit it. But see my most recent comment on the pitch.

app/grandchallenge/evaluation/templates/evaluation/submission_detail.html
jmsmkn: Hmm, maybe if the creator has `view_submission` then it should always be displayed, didn't realise that.

jmsmkn: Those settings are under "Result detail". Submissions are not visible by others so maybe this is okay.

app/grandchallenge/evaluation/models.py
amickan: God to know! I'll update the property and also add a test.

app/grandchallenge/evaluation/templates/evaluation/submission_detail.html
amickan: Will leave as is for now until that additional PR is done.

app/grandchallenge/evaluation/templates/evaluation/submission_detail.html
amickan: > These have been moved out of a section that was only visible to admins, we have to be really careful when we change visibility for things like this.

Definitely, but as I said in my previous comment, it is accessible to anyone with view permission. If that is not the way it should be, we need to update the serving view.

jmsmkn: So long as the permissions match then this is fine.

jmsmkn:
There doesn't seem to be anything new to review.

amickan:
> There doesn't seem to be anything new to review.

Pre-lunch-brain-fog... I committed my changes, but never pushed them. Sorry! They are there now

jmsmkn (approved):
LGTM!

PR #4697, opened 2026-05-01
Copy_button template tag

amickan:
This adds a copy_button template tag that truncates a string, optionally wraps the string in an anchor tag given a link, and prepends a copy icon. 

See the evaluation and method columns here: 

Screenshot 2026-05-01 at 13 06 11

Clicking the copy button shows a tooltip as feedback that copying succeeded:

https://github.com/user-attachments/assets/44abdf48-9ca0-4a10-a372-32d99e1fb06d


app/grandchallenge/core/templatetags/copy_button.py
```diff
@@ -0,0 +1,37 @@
...
+
+register = template.Library()
+
+
+@register.filter
+def copy_pk_button(value, link=None):
+    try:
+        uuid.UUID(str(value))
+    except (ValueError, AttributeError):
+        return ""
```
amickan: Wasn't sure if it should error out here or fail silently.

app/grandchallenge/core/templatetags/copy_button.py
```diff
@@ -0,0 +1,37 @@
...
+    button = format_html(
+        "<button "
+        'type="button"'
+        'class="btn btn-link text-dark p-0 m-0 mr-1 copy-btn shadow-none" '
+        'data-copy="{}" '
+        'data-placement="left" '
+        'title="Copy to clipboard">'
+        '<i class="far fa-copy"></i>'
+        "</button>",
+        value,
```
jmsmkn: Please use kwargs with format_html

app/grandchallenge/core/templatetags/copy_button.py
```diff
@@ -0,0 +1,37 @@
...
+    )
+
+    shortened = split_first(value, "-")
```
jmsmkn: `split_first` can probably be removed.

app/grandchallenge/core/templatetags/copy_button.py
jmsmkn: I don't think this block is necessary at all, why not let it work with all strings?

app/grandchallenge/evaluation/templates/evaluation/evaluation_admin_list_row.html
```diff
@@ -2,13 +2,14 @@
 {% load user_profile_link from profiles %}
 {% load url %}
 {% load civ %}
+{% load copy_button %}
 
 
-<a href="{{ object.submission.get_absolute_url }}">{{ object.submission.id|split_first:"-" }}</a>
+{{ object.submission.id|copy_pk_button:object.submission.get_absolute_url }}
```
jmsmkn: Please always use `.pk` rather than `.id`, this is to avoid confusion with `id()`. If you are using an LLM this has to be in your agents.md. Same with all of our other conventions.

app/grandchallenge/core/templatetags/copy_button.py
```diff
@@ -0,0 +1,37 @@
...
+
+    if link:
+        return button + format_html('<a href="{}">{}</a>', link, shortened)
+    else:
+        return button + format_html("{}", shortened)
```
jmsmkn: format_html should use kwargs. You could remove the double format_html calling by defaulting the link to `"#"`.

app/grandchallenge/core/templatetags/copy_button.py
```diff
@@ -0,0 +1,37 @@
...
+
+from grandchallenge.evaluation.templatetags.evaluation_extras import (
+    split_first,
+)
+
+register = template.Library()
+
+
+@register.filter
+def copy_pk_button(value, link=None):
```
jmsmkn: Could use `@stringfilter` here: https://docs.djangoproject.com/en/6.0/howto/custom-template-tags/#template-filters-that-expect-strings

app/grandchallenge/evaluation/templates/evaluation/evaluation_admin_list_row.html
```diff
@@ -2,13 +2,14 @@
...
 
 <split></split>
 
-<a href="{{ object.get_absolute_url }}">{{ object.id|split_first:"-" }}</a>
+{{ object.id|copy_pk_button:object.get_absolute_url }}
```
jmsmkn: `.id` used again here.

app/grandchallenge/evaluation/templates/evaluation/evaluation_admin_list_row.html
amickan: Good catch. I hadn't noticed that it was `id`. I just added the template tag to what was already there. Should have checked more carefully.

app/grandchallenge/core/templatetags/copy_button.py
amickan: Also fine. Strings without a dash would then not be truncated. If we ever want to use this with non-pk strings that contain a dash, we can extend the template tag to drop the truncation.

app/grandchallenge/core/templatetags/copy_button.py
amickan: I think I prefer to not show a link when the string does not link to a new page.

app/grandchallenge/evaluation/templates/evaluation/evaluation_admin_list_row.html
jmsmkn: Ah sorry, I see now!

app/grandchallenge/core/templatetags/copy_button.py
amickan: I could foresee us wanting to use this for pks without truncating them. For example on the evaluation detail page for the submission, evaluation and method pks. 

I could just turn this into a simple tag and add another boolean variable `truncate` that conditionally splits the value at the first dash or leaves it as is. What do you think?

jmsmkn: No, do not add the option, if it is needed then it can be added then (YAGNI).

app/grandchallenge/core/templatetags/copy_button.py
```diff
@@ -0,0 +1,32 @@
...
+    shortened = str(value).split("-", 1)[0]
+
+    if link:
+        return button + format_html(
+            '<a href="{link}">{display_value}</a>',
+            link=link,
+            display_value=shortened,
+        )
+    else:
+        return button + format_html("{display_value}", display_value=shortened)
```
jmsmkn, line 32: Maybe I'm missing something here but why not do:

```suggestion
        return button + shortened
```

jmsmkn: Ah, because both need to be safe stings to return a safe string.

jmsmkn (commented):
LGTM but needs @HarmvZ to look at the JS

app/grandchallenge/core/templatetags/copy_button.py
```diff
@@ -0,0 +1,32 @@
...
+
+register = template.Library()
+
+
+@register.filter
+@stringfilter
+def copy_button(value, link=None):
+    button = format_html(
+        "<button "
+        'type="button"'
```
HarmvZ:
```suggestion
        'type="button" '
```

app/grandchallenge/core/static/js/copy_to_clipboard.mjs
```diff
@@ -0,0 +1,25 @@
...
+    navigator.clipboard.writeText(id).then(() => {
+        $(btn)
+            .tooltip({ trigger: "manual" })
+            .attr("data-original-title", "Copied to clipboard!")
+            .tooltip("show");
+
+        activeTimers[id] = setTimeout(() => {
+            $(btn).attr("title", originalTitles[id]).tooltip("hide");
+        }, 2000);
+    });
```
HarmvZ: You could add a `.catch()` here and show a tooltip indicating to the user that the copy failed. (nice to have)

HarmvZ (changes requested):

HarmvZ (approved):

pkcakeout:
I know that it is merged and everything, but just to prevent things from a design perspective, you should likely choose a location for "action buttons" on grand-challenge as a whole:

image

Some action buttons ("Copy") are now to the left of containers, others are to the right of containers ("Hide this result"). Long-term this will lead to "button soup syndrome". I would choose a side.

jmsmkn:
Please make an issue if it is an issue.

PR #4698, opened 2026-05-01
Update evaluation list and detail pages

amickan:
Small follow-up PR that adds the action message to the evaluation detail page. Also cleans up the evaluation table to only show the columns that are relevant for the given phase. 

### DIFF
--- a/app/grandchallenge/evaluation/templates/evaluation/evaluation_admin_list_row.html
+++ b/app/grandchallenge/evaluation/templates/evaluation/evaluation_admin_list_row.html
@@ -27,15 +27,17 @@
 
 <split></split>
 
-{% for input in object.additional_inputs.all|sort_civs %}
-    <dd>
-        {% include 'components/partials/civ.html' with object=input display_inline=True only %}
-    </dd>
-{% empty %}
-    <span class="text-muted">Empty</span>
-{% endfor %}
+{% if phase.additional_evaluation_inputs.all %}
+    {% for input in object.additional_inputs.all|sort_civs %}
+        <dd>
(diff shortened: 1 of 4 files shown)


app/grandchallenge/evaluation/templates/evaluation/evaluation_admin_list_row.html
```diff
@@ -82,4 +83,6 @@
 
 <split></split>
 
-{{ object.submission.comment }}
+{% if object.allow_submission_comments %}
+    {{ object.submission.comment }}
```
jmsmkn, line 87: Should the split thing be in here?

amickan: We could use `truncatechars` indeed. `split_first` would only have an effect if the comment contains a dash, and I don't think we need the copy button from the other PR for this either.

jmsmkn: Oh sorry, I meant `<split></split>` as this is an optional column.

amickan: It's the last column of the table, so not needed as far as I understand

amickan: Having `<split></split>` after the last column or not, doesn't matter

jmsmkn: Okay!

jmsmkn (approved):
LGTM!

PR #4701, opened 2026-05-04
Add expiry date to invoices

chrisvanrun:
This PR adds the expiry date property to the `Invoices`.

Part of pitch:

Adding the property first, so we can utilize it in the more complex calculations directly (instead of inserting it at a later stage). Also allows parallel development of the second part of the pitch.

~A follow-up PR will make the property `null=False`, This to account for deployment timings and prevent DB errors, albeit we manually add Invoices, so the chances are very slim.~


app/grandchallenge/pages/templates/pages/challenge_statistics.html
```diff
@@ -68,6 +68,7 @@ <h4>Invoices</h4>
                     <tr>
                         <th>Number</th>
                         <th>Issued on</th>
+                        <th>Expires on</th>
```
amickan, line 71: Do we really want the UI changes already if we're not actually taking it into account yet? I would kept these changes for a later PR. Same goes for the annotation then.

app/grandchallenge/invoices/models.py
```diff
@@ -283,8 +301,19 @@ def clean(self):
...
+
     def save(self, *args, **kwargs):
         adding = self._state.adding
+        if not self.expires_on:
+            if adding:
+                self.created = now()
+            self.expires_on = self.created.date() + relativedelta(
+                years=settings.CHALLENGE_INVOICES_DEFAULT_EXPIRE_AFTER_YEARS
+            )
         super().save(*args, **kwargs)
```
amickan: I think you could do something like this:

```python 
def default_expiry():
    return date.today() + timedelta(years=5)

expires_on = models.DateField(
        help_text="The date when the invoice expires",
        default=default_expiry
    )
```

For the migration this would mean that you can't check for `expires_on__isnull=True` anymore, but since we don't have that many invoices, I think it would fine to loop over all invoices. You could then also make it required from the beginning.

app/grandchallenge/invoices/models.py
```diff
@@ -283,8 +301,19 @@ def clean(self):
                     "The total amount may not change. (You may only redistribute costs.)"
                 )
 
+        if self.pk and not self.expires_on:
+            raise ValidationError(
+                "The expiry date may not be empty when updating an invoice."
+            )
```
amickan: If you do what I suggest in my previous comment, this would be redundant.

app/grandchallenge/invoices/models.py
```diff
@@ -88,6 +101,11 @@ class Invoice(models.Model, FieldChangeMixin):
     created = models.DateTimeField(auto_now_add=True)
     updated = models.DateTimeField(auto_now=True)
 
+    expires_on = models.DateField(
+        help_text="The date when the invoice expires",
+        blank=True,
```
jmsmkn: This should always be set - so why allow blank?

app/grandchallenge/invoices/models.py
```diff
@@ -283,8 +301,19 @@ def clean(self):
...
+        if self.pk and not self.expires_on:
+            raise ValidationError(
+                "The expiry date may not be empty when updating an invoice."
+            )
+
     def save(self, *args, **kwargs):
         adding = self._state.adding
+        if not self.expires_on:
+            if adding:
+                self.created = now()
```
jmsmkn: Never manually set created, this is done by `auto_now_add`

app/grandchallenge/invoices/migrations/0011_invoice_expires_on.py
```diff
@@ -0,0 +1,35 @@
...
+        migrations.AddField(
+            model_name="invoice",
+            name="expires_on",
+            field=models.DateField(
+                help_text="The date when the invoice expires",
+                blank=True,
+                null=True,
+            ),
+        ),
+        migrations.RunPython(set_expires_on, elidable=True),
```
jmsmkn: Data migrations should be separate from table changes.

app/grandchallenge/invoices/models.py
```diff
@@ -88,6 +101,11 @@ class Invoice(models.Model, FieldChangeMixin):
     created = models.DateTimeField(auto_now_add=True)
     updated = models.DateTimeField(auto_now=True)
 
+    expires_on = models.DateField(
+        help_text="The date when the invoice expires",
+        blank=True,
+        null=True,
```
jmsmkn: Needs a 3rd migration to remove the null.

app/grandchallenge/invoices/migrations/0011_invoice_expires_on.py
```diff
@@ -0,0 +1,35 @@
...
+from django.db import migrations, models
+
+
+def set_expires_on(apps, schema_editor):
+    Invoice = apps.get_model("invoices", "Invoice")  # noqa: N806
+    for invoice in Invoice.objects.filter(expires_on__isnull=True):
+        invoice.expires_on = invoice.created.date() + relativedelta(
+            years=settings.CHALLENGE_INVOICES_DEFAULT_EXPIRE_AFTER_YEARS
+        )
+        invoice.save(update_fields=["expires_on"])
```
jmsmkn: Can be done in a single pass:

```python3
Invoice.objects.filter(expires_on__isnull=True).update(
    expires_on=ExpressionWrapper(
        TruncDate(F("created")) + timedelta(days=365 * settings.CHALLENGE_INVOICES_DEFAULT_EXPIRE_AFTER_YEARS),
        output_field=DateField(),
    )
)
```

app/grandchallenge/invoices/models.py
chrisvanrun: This was to cover creating via the admin page, however... I think Anne's proposed solution via a default function would work better. (Or have it set at the time the initial form gets created.)

app/grandchallenge/invoices/models.py
chrisvanrun: In this case, it hadn't been set yet. But I think having a default value works better overall. So this whole section will go.

app/grandchallenge/invoices/models.py
chrisvanrun: Smart! This also solves the blank=True issue I had! Wonderful.

app/grandchallenge/pages/templates/pages/challenge_statistics.html
chrisvanrun: It is currently only visible for us. I can pick it out, and push it back later, but I spent quite some time on getting the elements right here. I'd rather just keep it.

app/grandchallenge/invoices/models.py
chrisvanrun: Indeed!

app/grandchallenge/invoices/migrations/0011_invoice_expires_on.py
chrisvanrun: Why was this the case again? Was this to do with failure states and things being reversible?

app/grandchallenge/invoices/migrations/0011_invoice_expires_on.py
chrisvanrun: It can! Though the `expires_on__isnull` is now gone. I think that is fine since the addition of the field and overwriting the (then) existing `expires_on` is bundled in one PR/deployment. Do you agree?

app/grandchallenge/pages/templates/pages/challenge_statistics.html
amickan: Ah right, I missed that this is only for us. Fine to keep it then!

app/grandchallenge/invoices/migrations/0012_migrate_invoice_expires_on.py
```diff
@@ -0,0 +1,34 @@
...
+    Invoice.objects.update(
+        expires_on=ExpressionWrapper(
+            TruncDate(F("created"))
+            + timedelta(
+                days=365
+                * settings.CHALLENGE_INVOICES_DEFAULT_EXPIRE_AFTER_YEARS
+            ),
+            output_field=DateField(),
+        )
+    )
```
amickan, line 20: Am I reading this right that this sets the expiry date for all existing invoices to creation date + 1 year? That is not what we want, right? I think technically, the expiry date for invoices created before 2026 (or was there a specific date for the policy change) is December 31 of the year they were created in. For invoices created this year, it should be created date + 5 years. Right?

amickan: I definitely need a coffee! I read it wrong! Sorry. But the second part of my comment still stands, not sure if we need it to be so accurate though. For a simpler heuristic, we could also just set the expiry date for invoices before 2026 to Dec 31 2025. @miriam-groeneveld can you comment on whether that is a correct assumption?

app/grandchallenge/invoices/models.py
```diff
@@ -72,6 +73,18 @@ def with_overdue_status(self):
...
+
+        return self.annotate(
+            is_expired=ExpressionWrapper(
+                Q(
+                    expires_on__lt=today,
+                ),
+                output_field=models.BooleanField(),
+            ),
+        )
+
```
amickan, line 87: Why does this need to be annotated on the queryset? If we only need this in the template, you could just check this in the template itself, right? `{% if invoice.expires_at < now %}`

app/grandchallenge/invoices/models.py
amickan: This can go now.

app/tests/invoices_tests/test_models.py
```diff
@@ -460,3 +463,20 @@ def test_invoices_cannot_be_deleted():
         Invoice.objects.filter(pk=invoice.pk).delete()
 
     assert Invoice.objects.filter(pk=invoice.pk).exists()
+
+
+@pytest.mark.django_db
+def test_invoice_default_expires_on(settings, mocker):
```
amickan, line 469: Given that `expires_on` is just a `DateField`, I think this test would be fine without the mocking of the date (unless the test runs at midnight, but what are the odds). Took me a second to understand why you were doing this, but also fine to keep it this way.

app/grandchallenge/invoices/migrations/0012_migrate_invoice_expires_on.py
chrisvanrun: I think we should use a cut-off point like you mentioned. I feel that should be a manual shell call after this PR is live.

app/grandchallenge/invoices/migrations/0011_invoice_expires_on.py
jmsmkn: Okay - then that is a mistake with how the migration is being done. The first migration will not work on a database that has existing invoices. You are adding a field without a default. Django would have warned you about that when creating the migration. You will need to set a default, which could be a date but that would need to be static. The way to do this is:

- 1st migration: table - add the field, allow null, default to null
- 2nd migration: data - populate the rows with null values
- 3rd migration: table - disallow null and remove the default

app/grandchallenge/invoices/models.py
chrisvanrun: We could, benefit here would be the DB speed-up, albeit we'll never have enough invoices for that to really matter. Main reason here is that it mimics the `with_overdue_status`. I just realized it has one crucial fix though: a singular reference time/date. Technically, we should use a single reference time for all parts of a possibly combined (lazy) query. Instead of using a reference from Python we could use the Now on DB level:

```Python
def with_overdue_status(self):
    return self.with_due_date().annotate(
        is_overdue=ExpressionWrapper(
            Q(
                payment_type__in=[
                    Invoice.PaymentTypeChoices.PREPAID,
                    Invoice.PaymentTypeChoices.POSTPAID,
                ],
                payment_status=Invoice.PaymentStatusChoices.ISSUED,
                due_date__lt=TruncDate(Now()),
            ),
            output_field=models.BooleanField(),
        ),
        is_due=ExpressionWrapper(
            Q(
                payment_type__in=[
                    Invoice.PaymentTypeChoices.PREPAID,
                    Invoice.PaymentTypeChoices.POSTPAID,
                ],
                payment_status=Invoice.PaymentStatusChoices.ISSUED,
                due_date__gte=TruncDate(Now()),
                issued_on__lte=TruncDate(Now()),
            ),
            output_field=models.BooleanField(),
        ),
    )

def with_expired_status(self):
    return self.annotate(
        is_expired=ExpressionWrapper(
            Q(
                expires_on__lt=TruncDate(Now()),
            ),
            output_field=models.BooleanField(),
        ),
    )
```

jmsmkn: `now` is only available as a template tag - I think defining this as an annotation is a good way to do it as we will eventually need it in the available compute calculations too. Then it is defined in one place, less chance of a logic drift.

jmsmkn: > Technically, we should use a single reference time for all queries.

The original solution does that, the only difference is that you would be using the time on the python server rather than the time on the database server, no meaningful difference.

app/tests/invoices_tests/test_migrations.py
```diff
@@ -0,0 +1,38 @@
...
+
+from tests.invoices_tests.factories import InvoiceFactory
+
+_migration = importlib.import_module(
+    "grandchallenge.invoices.migrations.0012_migrate_invoice_expires_on"
+)
+
+
+@pytest.mark.django_db
+def test_set_expires_on_migration(settings):
```
jmsmkn: This test has little value, I would remove it.

app/grandchallenge/invoices/migrations/0012_migrate_invoice_expires_on.py
jmsmkn: I think keep it at 5 years too, otherwise we would cut off a bunch of challenges without warning once the business logic is implemented.

app/grandchallenge/invoices/models.py
amickan: > I think defining this as an annotation is a good way to do it as we will eventually need it in the available compute calculations too

Ah good point. Makes sense then!

app/tests/invoices_tests/test_models.py
chrisvanrun: With timing, I always prefer to use mocked fixed time points. If it works once, it'll keep working irrespective of when the test runs. I'll leave the timezone, leap years and other fun bits to the datetime devs =).

app/grandchallenge/invoices/models.py
chrisvanrun: Actually, having looked at this a bit more. I think it is a better to use the Python reference: the database might be in a completely different timezone and mixing the render node time and that on a distance (DB)... is likely a recipe for VERY weird bugs. Better to have some incongruity if someone happens to load the page on exactly the UTC 00:00.

app/grandchallenge/invoices/migrations/0012_migrate_invoice_expires_on.py
chrisvanrun: Rough shell call:

```Python
from datetime import date, timedelta

from django.db.models import DateField, ExpressionWrapper, F
from django.db.models.functions import TruncDate

from grandchallenge.invoices.models import Invoice

Invoice.objects.filter(created__gte=date(2025, 12, 31)).update(
    expires_on=ExpressionWrapper(
        TruncDate(F("created"))
        + timedelta(
            days=365
        ),
        output_field=DateField(),
    )
)
```

app/grandchallenge/invoices/models.py
chrisvanrun: Huh, GitHub has hidden some comments from me while I was working on things:
afbeelding

> The original solution does that, the only difference is that you would be using the time on the python server rather than the time on the database server, no meaningful difference.

Except for that it would be the exact same timestamp for all `Now` comparisons in the single db query, combining multiple annotations would then use the same timestamp. As far as I understand the `with_overdue_status` and `with_expired_status` could _theoretically_ have a different date. Practically not and the effects are so minor that we can let it slide I think.

app/grandchallenge/invoices/migrations/0011_invoice_expires_on.py
chrisvanrun: I had expected (and locally it seems to work through existing invoices) that it would use the current default generating function? I.e. today + offset? Django was stringent on the type of callable it was, but otherwise it gave me no warnings. Note: I changed that after Anne commented that it might work.

app/grandchallenge/invoices/models.py
jmsmkn: > the database might be in a completely different timezone and mixing the render node time and that on a distance (DB)

Not a concern. The point of using now is that it is timezone agnostic.

> As far as I understand the with_overdue_status and with_expired_status could theoretically have a different date

Yes, if the two calls span UTC midnight boundary. Given the entire response should occur within 100ms, likely these two calls would happen within a couple of ms, so only a problem if the code path would occur at ~23:59:59.990 UTC. Non-issue if you ask me. Our codebase is probably littered with such things. If you really think that it is a problem then cache the date using `@cached_property` on the queryset.

That issue would also be the same for using the database `Now`, but in that case you would use `TransactionNow`.

app/grandchallenge/invoices/migrations/0011_invoice_expires_on.py
amickan: I think it "works" in this case because the callable that is used as the default value returns the same value (today's date) for each row. So it boils down to the same as using a static default.

app/grandchallenge/invoices/migrations/0012_migrate_invoice_expires_on.py
jmsmkn: Why have a manual data migration step when there is already a data migration?

app/grandchallenge/invoices/migrations/0011_invoice_expires_on.py
jmsmkn: My mistake, missed the default, sorry!

app/grandchallenge/invoices/migrations/0012_migrate_invoice_expires_on.py
amickan: I think Chris missed your comment to keep it at 5 years. @chrisvanrun I had not thought about the effect it would have on running challenges, so I agree with James that we should just keep it at 5 years for all invoices.

app/grandchallenge/invoices/models.py
amickan: This still needs to go, otherwise I think the PR LGTM.

app/grandchallenge/invoices/migrations/0012_migrate_invoice_expires_on.py
chrisvanrun: > Why have a manual data migration step when there is already a data migration?

Because the cut-off date is something relevant to our specific Radboudumc bookkeeping. IMO I still think we should push that expiry date for old challenges since leaving it leaves a potential money drain on the table for the next few years. I can see it's not nice to do a hard cut-off like this out of the blue. We can push it somewhere in the (near) future where they'll get email warning of it happening.

amickan (approved):
LGTM!

PR #4702, opened 2026-05-04
Add invocation input provision task

koopmant:

### DIFF
--- a/app/grandchallenge/algorithms/models.py
+++ b/app/grandchallenge/algorithms/models.py
@@ -43,10 +43,7 @@ from grandchallenge.components.models import (  # noqa: F401
     Tarball,
 )
 from grandchallenge.components.schemas import GPUTypeChoices
-from grandchallenge.components.tasks import (
-    provision_invocation_input_data,
-    start_endpoint,
-)
+from grandchallenge.components.tasks import start_endpoint
 from grandchallenge.core.guardian import (
     GroupObjectPermissionBase,
     UserObjectPermissionBase,
@@ -1820,12 +1817,6 @@ class Invocation(UUIDModel):
         if adding:
             if self.endpoint.status != self.endpoint.StatusChoices.RUNNING:
                 raise ValidationError("Endpoint is not running")
-
-            on_commit(
-                provision_invocation_input_data.signature(
-                    kwargs=self.task_kwargs
-                ).apply_async
-            )
 
         super().save(*args, **kwargs)
 
--- a/app/grandchallenge/algorithms/models.py
+++ b/app/grandchallenge/algorithms/models.py
@@ -1828,6 +1828,14 @@ class Invocation(UUIDModel):
         )
 
     @property
+    def task_kwargs(self):
+        return {
+            "app_label": self._meta.app_label,
+            "model_name": self._meta.model_name,
+            "pk": self.pk,
+        }
+
+    @property
     def orchestrator_kwargs(self):
         kwargs = self.endpoint.orchestrator_kwargs
         kwargs["invocation_id"] = (
@@ -1839,3 +1847,30 @@ class Invocation(UUIDModel):
     @property
     def orchestrator(self):
         return EndpointOrchestrator(**self.orchestrator_kwargs)
+
+    @cached_property
+    def inputs_complete(self):
+        # check if all inputs are present and if they all have a value
+        # interfaces that do not require a value will be considered complete regardless
+        return {
+            civ.interface
+            for civ in self.inputs.all()
+            if civ.has_value or not civ.interface.value_required
+        } == {*self.algorithm_interface.inputs.all()}
+
+    def update_status(
+        self,
+        *,
+        status: InvocationStatusChoices,
+        error_message="",
+        invoke_duration=None,
+    ):
+        self.status = status
+
+        if error_message:
+            self.error_message = error_message[:1024]
+
+        if invoke_duration is not None:
+            self.invoke_duration = invoke_duration
+
+        self.save()
--- a/app/grandchallenge/algorithms/models.py
+++ b/app/grandchallenge/algorithms/models.py
@@ -43,7 +43,10 @@ from grandchallenge.components.models import (  # noqa: F401
     Tarball,
 )
 from grandchallenge.components.schemas import GPUTypeChoices
-from grandchallenge.components.tasks import start_endpoint
+from grandchallenge.components.tasks import (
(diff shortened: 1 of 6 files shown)


app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
```diff
@@ -16,23 +16,30 @@ def __init__(
...
         signing_key,
+        invocation_id=None,
+        time_limit=settings.ALGORITHM_ENDPOINTS_MAXIMUM_INVOCATION_DURATION,
         algorithm_model=None,
     ):
+        if invocation_id:
+            job_id = invocation_id
+        else:
+            job_id = endpoint_id
         self._executor = AmazonSageMakerTrainingExecutor(
```
koopmant, line 26: The usage of the executor on this class is becoming more complex:
- setting bucket names
- not setting algorithm_model to avoid auxiliary_data tasks (so we can still use the provision method on the executor)
- setting job_id based on either endpoint_id or invocation_id

Both `Invocation` and `Endpoint` now have an orchestrator property, returning an orchestrator with the required (different) properties. It works, but it's opaque. I haven't been able to come up with an alternative though.

koopmant:
Reverted the last commit as I realized that's exactly the opposite of what we want. See https://github.com/DIAGNijmegen/rse-roadmap/issues/481#issuecomment-4359036134

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
```diff
@@ -16,23 +16,30 @@ def __init__(
...
         api_method,
         signing_key,
+        invocation_id=None,
+        time_limit=settings.ALGORITHM_ENDPOINTS_MAXIMUM_INVOCATION_DURATION,
         algorithm_model=None,
     ):
+        if invocation_id:
+            job_id = invocation_id
+        else:
+            job_id = endpoint_id
```
jmsmkn: Neither `invocation_id` or `endpoint_id` are further used as far as I can tell, maybe only have one required kwarg: `job_id` or `task_id`?

jmsmkn:
Looks good! Would be good to have Anne take a look.

app/grandchallenge/algorithms/models.py
```diff
@@ -1823,3 +1826,51 @@ def inference_id(self):
...
             f"{settings.COMPONENTS_REGISTRY_PREFIX}-alg-endp-invoc-{self.pk}"
         )
+
+    @property
+    def task_kwargs(self):
+        return {
+            "app_label": self._meta.app_label,
+            "model_name": self._meta.model_name,
+            "pk": self.pk,
+        }
```
amickan, line 1836: Is this used anywhere?

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
amickan: > Both `Invocation` and `Endpoint` now have an orchestrator property, returning an orchestrator with the required (different) properties. It works, but it's opaque.

On initial read, I was confused by that as well, but I think it's clear enough after all.

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
amickan: That sounds like a good simplification.

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
```diff
@@ -16,23 +16,30 @@ def __init__(
...
+            job_id=job_id,
             exec_image_repo_tag=exec_image_repo_tag,
             memory_limit=memory_limit,
             time_limit=time_limit,
             requires_gpu_type=requires_gpu_type,
             use_warm_pool=False,
             signing_key=signing_key,
             api_method=api_method,
-            algorithm_model=algorithm_model,
+            algorithm_model=None,
```
amickan, line 35: For my understanding: why do we set the model to `None` here?

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
koopmant: Indeed, you can't tell. I wrote it like this so it's more obvious that the job_id is different depending on where you get the orchestrator: if you get it from the endpoint we use the endpoint_id as job_id; if you get it from the invocation, we use the invocation_id as job_id. 

Alternatively I can handle this in `orchestrator_kwargs`, and provide the job_id directly. I feel like it wouldn't be as obvious then though.

app/grandchallenge/algorithms/models.py
koopmant: Not anymore, I was scheduling the task on creation of the invocation. But I no longer do that. We will need this for later tasks anyway though.

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
jmsmkn: I think it's better to do it in the `orchestrator_kwargs`. If you did it here in the init method you would really need more validation to ensure that exactly one of `job_id` or `endpoint_id` is set, and as far as the orchestrator is concerned it doesn't make a meaningful difference.

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
koopmant: This is what I meant in the other comment. We set it the `None` to ensure the auxiliary data tasks are empty. You cannot see it from the code, but this way we can call `_executor.provision()`. It handles the provisioning but also includes the auxiliary data, which is already loaded for invocations. It seems a bit fragile, although I don't expect the code in Executor to change. I have included a test to ensure the tasks are empty and a test to check which provisioning tasks are used.

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
koopmant: All right

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
amickan: Ah right, makes sense. I guess it's fine then. Good to have the tests to alert future us in case of changes.

amickan (approved):
LGTM!

PR #4704, opened 2026-05-05
Clean up endpoint models

koopmant:
- add signing_key to `Endpoint`
- update `InvocationStatusChoices`
- add `algorithm_interface` to `Invocation`


app/grandchallenge/algorithms/models.py
```diff
@@ -1773,6 +1783,9 @@ class Invocation(UUIDModel):
         ),
     )
     error_message = models.CharField(max_length=1024, default="", blank=True)
+    algorithm_interface = models.ForeignKey(
+        AlgorithmInterface, on_delete=models.PROTECT
```
jmsmkn, line 1787: This will not be set for jobs that were created before algorithm interfaces were implemented, so cannot be done without a data migration.

koopmant: Isn't this enough?

```
        migrations.AlterField(
            model_name="job",
            name="algorithm_interface",
            field=models.ForeignKey(
                default=None,
                on_delete=django.db.models.deletion.PROTECT,
                to="algorithms.algorithminterface",
            ),
            preserve_default=False,
        ),
```
It's a one-time default of None. Older rows will be invalid, but is that an issue?

koopmant: I just tested it, yes that is an issue.

jmsmkn: Yes, if you do not allow null values in a column then none of the values can be null. Why are you touching the job model as part of your pitch?

koopmant: I'll revert this part then, since there is no good default for jobs. 

For invocations I'll keep it as is (doing the same thing there), which should not be a problem because there are no rows yet.

koopmant: > Why are you touching the job model as part of your pitch?

It was a confusing part in the job model, since algorithm_interface on a job should never be None. I discussed it with Anne and it was a leftover from earlier work. I thought I saw an easy fix, so I figured I'd include it here. I reverted it just now.

jmsmkn (approved):
LGTM!

PR #4705, opened 2026-05-05
Revert "Clean up endpoint models"

jmsmkn:
Reverts DIAGNijmegen/rse-grand-challenge#4704

Due to `│ "django.db.utils.IntegrityError: could not create unique index \"algorithms_endpoint_signing_key_key\""
│ "DETAIL:  Key (signing_key)=(...) is duplicated."`


jmsmkn:
I undid this as the problem was only on staging.

PR #4706, opened 2026-05-06
Add invoke endpoint task

koopmant:

### DIFF
--- a/app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
+++ b/app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
@@ -84,6 +84,14 @@ class EndpointOrchestrator:
         return f"s3://{settings.ALGORITHM_ENDPOINTS_OUTPUT_BUCKET_NAME}/{self._io_prefix}/failures"
 
     @property
+    def _invocation_key(self):
+        return self._executor._invocation_key
+
+    @property
+    def _invocation_s3_uri(self):
+        return f"s3://{settings.ALGORITHM_ENDPOINTS_INPUT_BUCKET_NAME}/{self._invocation_key}"
+
+    @property
     def invocation_environment(self):
         env = self._executor.invocation_environment
 
@@ -212,3 +220,11 @@ class EndpointOrchestrator:
 
     def provision_invocation_input_data(self, *, input_civs):
         self._executor.provision(input_civs=input_civs, input_prefixes={})
+
+    def invoke_endpoint(self, *, inference_id):
+        self._sagemaker_runtime_client.invoke_endpoint_async(
+            EndpointName=self._endpoint_name,
+            ContentType="application/json",
+            InputLocation=self._invocation_s3_uri,
+            InferenceId=inference_id,
+        )
--- a/app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
+++ b/app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
@@ -118,6 +118,10 @@ class EndpointOrchestrator:
         else:
             return 30
 
+    @property
+    def _time_limit(self):
+        return self._executor._time_limit
+
     def provision_auxiliary_data(self):
         if self._algorithm_model:
             self._s3_client.copy(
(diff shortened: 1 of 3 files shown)


app/grandchallenge/components/tasks.py
```diff
@@ -1775,4 +1775,49 @@ def provision_invocation_input_data(
...
+
+    orchestrator = invocation.orchestrator
+
+    try:
+        orchestrator.invoke_endpoint(inference_id=invocation.inference_id)
+    except Exception:
+        logger.error("Could not invoke endpoint", exc_info=True)
+
+        on_commit(
+            stop_endpoint.signature(
```
jmsmkn: Do you really want to stop the endpoint if a single invocation fails? I would have thought that you would still want to keep the endpoint running for future invocations?

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
```diff
@@ -212,3 +220,11 @@ def attempt(method):
 
     def provision_invocation_input_data(self, *, input_civs):
         self._executor.provision(input_civs=input_civs, input_prefixes={})
+
+    def invoke_endpoint(self, *, inference_id):
+        self._sagemaker_runtime_client.invoke_endpoint_async(
+            EndpointName=self._endpoint_name,
+            ContentType="application/json",
+            InputLocation=self._invocation_s3_uri,
+            InferenceId=inference_id,
```
jmsmkn, line 233: See the earlier discussion, I think that this is missing a timeout.

app/grandchallenge/components/tasks.py
koopmant: This is the call to the runtime client to start the invocation asynchronously, if it fails I'm unsure what exactly failed at this point and I assume we'll need to investigate.

jmsmkn: I think that you're right that we should get the error, but I don't think that stopping at this point is necessary. It can go through the normal stopping channels.

I missed this for the provisioning task by the way. For invocations I don't think that their failure should change the state of the endpoint.

koopmant: All right; I'll adjust the other task as well.

koopmant: done

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
koopmant: Added!

jmsmkn (approved):
LGTM!

app/grandchallenge/components/tasks.py
```diff
@@ -1761,18 +1761,51 @@ def provision_invocation_input_data(
...
+
+    else:
+        invocation.update_status(status=invocation.StatusChoices.PROVISIONED)
+
         on_commit(
-            stop_endpoint.signature(
-                kwargs=invocation.endpoint.task_kwargs,
+            invoke_endpoint.signature(
+                kwargs=invocation.task_kwargs,
             ).apply_async
```
amickan, line 1775: Don't you still want to stop the endpoint in case of a failure to provision the inputs?

app/grandchallenge/components/tasks.py
```diff
@@ -1761,18 +1761,51 @@ def provision_invocation_input_data(
...
+
+    try:
+        orchestrator.invoke_endpoint(inference_id=invocation.inference_id)
+    except Exception:
+        logger.error("Could not invoke endpoint", exc_info=True)
+
         invocation.update_status(
             status=invocation.StatusChoices.FAILURE,
             error_message=SystemErrorMessages.UNEXPECTED_ERROR,
         )
```
amickan, line 1808: And here also: don't we need to stop the endpoint in this case, or in some cases at least, depending on the error?

app/grandchallenge/components/tasks.py
amickan: Sorry, I see now that there was a discussion about that.

app/grandchallenge/components/tasks.py
amickan: Disregard this comment as well.

amickan (approved):
LGTM

Issue #4408, opened 2025-11-12
Kombu 5.6 breaks SQS messages


jmsmkn:

jmsmkn:
I was right to be suspicious, something breaks with every minor release of kombu/celery.

We get:

```
[2025-11-12 11:24:24,298: WARNING/MainProcess] Received and deleted unknown message. Wrong destination?!?


The full contents of the message body was: body: '{\n "id": "<uuid>",\n "content-encoding": "utf-8",\n "content-type": "application/json",\n "task": "grandchallenge.components.tasks.handle_event",\n "kwargs": {\n "event":... (4110b)
```

jmsmkn:
We now no longer use the celery messages on SQS, and have migrated those to lambda tasks. This means that kombu and celery could be upgraded, but as we are replacing them we do not need to bother unless there is a security issue.

jmsmkn:
Celery task durations:

```
grandchallenge.components.tasks.validate_docker_image: 2253.107348
grandchallenge.cases.tasks.execute_post_process_image_task: 1797.55272

Celery time limit 900s

grandchallenge.cases.tasks.build_images: 762.317973
grandchallenge.components.tasks.upload_to_registry_and_sagemaker: 728.88234
grandchallenge.challenges.tasks.update_challenge_storage_size: 516.57187
grandchallenge.emails.tasks.send_bulk_email: 270.280197
grandchallenge.components.tasks.add_file_to_object: 270.127843
grandchallenge.components.tasks.provision_job: 244.215467

LAMBDA_TASKS_DEFAULT_SOFT_TIMEOUT = 240

grandchallenge.challenges.tasks.update_challenge_compute_costs: 160.240743
grandchallenge.evaluation.tasks.set_evaluation_inputs: 148.858345
grandchallenge.cases.tasks.import_dicom_to_health_imaging: 119.378672
grandchallenge.publications.tasks.update_publication_metadata: 102.1591
grandchallenge.challenges.tasks.update_challenge_results_cache: 101.086274
grandchallenge.utilization.tasks.create_job_warm_pool_utilizations: 92.913014
grandchallenge.components.tasks.assign_docker_image_from_upload: 88.776667
grandchallenge.statistics.tasks.update_site_statistics_cache: 85.556998
grandchallenge.discussion_forums.tasks.create_forum_notifications: 77.643491
grandchallenge.components.tasks.assign_tarball_from_upload: 75.97088
[...]

PR #4707, opened 2026-05-07
Move all event handlers to lambda tasks

jmsmkn:
See https://github.com/DIAGNijmegen/rse-cloud-infrastructure/issues/373
See #4408

### DIFF
--- a/app/config/settings.py
+++ b/app/config/settings.py
@@ -1302,10 +1302,6 @@ CELERY_BEAT_SCHEDULE = {
         "task": "grandchallenge.evaluation.tasks.cancel_external_evaluations_past_timeout",
         "schedule": timedelta(hours=1),
     },
-    "push_metrics_to_cloudwatch": {
-        "task": "grandchallenge.core.tasks.put_cloudwatch_metrics",
-        "schedule": timedelta(seconds=30),
-    },
     "stop_expired_services": {
         "task": "grandchallenge.components.tasks.stop_expired_services",
         "kwargs": {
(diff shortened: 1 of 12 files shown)


app/config/settings.py
```diff
@@ -1302,10 +1302,6 @@ def get_private_ip():
         "task": "grandchallenge.evaluation.tasks.cancel_external_evaluations_past_timeout",
         "schedule": timedelta(hours=1),
     },
-    "push_metrics_to_cloudwatch": {
```
koopmant, line 1305: How does this get scheduled now?

jmsmkn: That's something I need to work on in staging - there is functionality for this in event bridge. Shouldn't block merging though as removing the periodic task from settings does not remove it from the database, so it will still run until we manually remove that.

koopmant (approved):
LGTM!

app/config/settings.py
koopmant: I see. All right!

amickan:
Just to double check / confirm my assumptions: we are aiming to entirely remove celery from our stack eventually, right? So all @acks_late_micro_short_task decorators will eventually be replaced with @lambda_task, correct? For the @acks_late_2x_large_task lambda would not work because of the time constraints, for that we'd need step functions or something else.

jmsmkn:
Yes, that is the plan. Unfortunately something with Celery breaks with every minor release, it is far too complex for what we need it for. There is only one task that would not fit on lambda (validate_docker_image), but we can come up with something else for that. See #4408, https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/253#issuecomment-4046618787.

PR #4708, opened 2026-05-07
Add follow_up_on field to Invoice

amickan:
Adds a `follow_up_on` field to the Invoice model. This date cannot be in the past or more than 1 year into the future and needs to be before the expiry date of the invoice. 

In a future PR the field will be made required for post-paid invoices. Can't do that yet since the migration would fail otherwise. So we have to deploy this, fill in dates for all existing post-paid invoices and then add the extra constraint. 

Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/475


app/grandchallenge/invoices/models.py
```diff
@@ -122,10 +124,15 @@ class Invoice(models.Model, FieldChangeMixin):
         null=True,
     )
     last_checked_on = models.DateField(
-        help_text="The date when the invoice status was last checked",
+        help_text="The date when the status of issued invoices was last checked",
         blank=True,
         null=True,
     )
+    follow_up_on = models.DateField(
+        help_text="The date when a post-paid invoice will be issued if necessary (required for post-paid invoices).",
```
chrisvanrun, line 132: Should we make this more general and say that it's a generic time point action might be required for the invoice. I would like to set this equal the expiry date for regular invoices, if they want to be informed about it expiring or not.

app/tests/invoices_tests/test_models.py
```diff
@@ -480,3 +481,39 @@ def test_invoice_default_expires_on(settings, mocker):
     )
     invoice = InvoiceFactory()
     assert invoice.expires_on == date(2027, 3, 1)
+
+
+@pytest.mark.django_db
```
chrisvanrun, line 493: Nice, clean tests. 

I would suggest adding a positive test here as well. AFAIK adding the attribute to the `InvoiceFactory` by default it does not cover the `full_clean` OK-ness of the property.

chrisvanrun (changes requested):

app/grandchallenge/invoices/models.py
amickan: It cannot be set equal to the expiry date. It needs to be within a year and before the expiry date.

amickan: If we want to enable email notifications about the expiry date, we should do that separately. Otherwise, all the constraints need to go. I don't think loosening those constraints is an option since those were added with the specific goal to get a grip on post-paid invoice handling.

amickan: Thinking about this some more, I understand the request to keep the field more general. We would need to add another condition to each constraint to make the date restrictions I added here only apply to post-paid invoices. Then the field could potentially be used for other purposes in the future.

If the goal with that is to send a reminder about the expiry date for invoices though, I think that could better be achieved by simply adding a task that checks the expiry date directly. Even if we were to use the follow-up date for this purpose, we'd need to add a new, separate task to send that email, so there is no benefit to using the follow-up field IMO (and that's not part of this anyway already delayed pitch). 

Moreover, semantically, the only case where we want to _follow-up_ on an invoice is for initialized post-paid invoices. The other invoices don't need a scheduled follow-up and informing challenge orgs of the expiry date of an invoice is not a follow-up, it's just a notification of a fact.

chrisvanrun: I agree with your logic. Let's keep it out of scope for now, we can always add it later.

chrisvanrun (dismissed):
I had checked the constraints in my first review round. And if we purely focus on the post-paid this LGTM!

app/grandchallenge/invoices/models.py
amickan: Ok. I'll wait for @jmsmkn to comment on this / approve the PR.

app/grandchallenge/invoices/models.py
```diff
@@ -269,6 +276,24 @@ class Meta:
...
+                name="follow_up_on_not_more_than_a_year_in_future",
+                condition=Q(follow_up_on__isnull=True)
+                | Q(follow_up_on__lte=Now() + timedelta(days=365)),
+                violation_error_message="Follow-up date cannot be more than a year into the future.",
+            ),
+            models.CheckConstraint(
+                name="follow_up_on_not_in_past",
+                condition=Q(follow_up_on__isnull=True)
+                | Q(follow_up_on__gte=Now()),
+                violation_error_message="Follow-up date cannot be in the past.",
```
jmsmkn: Won't this mean that the object can never be updated once the follow up date is in the past?

app/grandchallenge/invoices/models.py
jmsmkn: I agree with not letting the scope creep.

app/grandchallenge/invoices/models.py
amickan: That is a good point. You're right! That needs to go then...

jmsmkn (approved):
LGTM!

PR #4709, opened 2026-05-11
Unwrap invocation json contents for invocations

koopmant:
The invocation.json content for jobs are inference task(s) in a list. For an endpoint invocation there can be only one inference task. The content should not be a list, otherwise the invocation fails.  

### DIFF
--- a/app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
+++ b/app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
@@ -1,6 +1,7 @@
 import boto3
 from botocore.exceptions import ClientError
 from django.conf import settings
+from pydantic_core import to_json
 
 from grandchallenge.components.backends.amazon_sagemaker_training import (
     AmazonSageMakerTrainingExecutor,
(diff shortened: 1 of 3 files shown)


app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
```diff
@@ -223,8 +224,13 @@ def attempt(method):
         self.deprovision_auxiliary_data()
 
     def provision_invocation_input_data(self, *, input_civs):
+        self._executor._get_invocation_json = self._get_invocation_json
```
koopmant: The endpoint orchestrator and the executor need to be refactored so we don't have to do this. For now I think this is fine?

jmsmkn: This is really difficult to follow, and I think will be quite brittle. Maybe a boolean on the executor (`use_task_list: bool = True`?) would be easier to implement, follow and maintain?

koopmant: Yes, that will do nicely. Thanks!

jmsmkn (dismissed):
LGTM!

koopmant:
Accidentally committed to the wrong branch, dismissing your review @jmsmkn. Could you approve again?

jmsmkn (approved):

PR #4710, opened 2026-05-11
Update endpoint models

koopmant:
- Add missing fields to `Endpoint` and `Invocation` models
- Show fields as readonly in admin
- Use editable=False consistently
- Shorten inference_id so it is not too long on staging


app/grandchallenge/algorithms/models.py
```diff
@@ -1581,6 +1581,9 @@ class Endpoint(FieldChangeMixin, UUIDModel):
         choices=EndpointStatusChoices,
         default=EndpointStatusChoices.QUEUED,
     )
+    stdout = models.TextField(editable=False)
+    stderr = models.TextField(editable=False)
```
jmsmkn: Shouldn't these have defaults as they will not get filled when the objects are created, usually we only set the logs when the task ends.

app/grandchallenge/algorithms/models.py
```diff
@@ -1785,7 +1788,11 @@ class Invocation(UUIDModel):
             "any delays from shared hardware issues."
         ),
     )
-    error_message = models.CharField(max_length=1024, default="", blank=True)
+    stdout = models.TextField(editable=False)
+    stderr = models.TextField(editable=False)
```
jmsmkn: Shouldn't these have defaults as they will not get filled when the objects are created, usually we only set the logs when the task ends.

app/grandchallenge/algorithms/models.py
```diff
@@ -1785,7 +1788,11 @@ class Invocation(UUIDModel):
...
+    runtime_metrics = models.JSONField(default=dict, editable=False)
+    error_message = models.CharField(max_length=1024, editable=False)
```
jmsmkn: Shouldn't this have a default too?

app/grandchallenge/algorithms/models.py
```diff
@@ -1823,9 +1830,7 @@ def save(self, *args, **kwargs):
     @property
     def inference_id(self):
         # Sagemaker requires this to be max 64 chars
-        return (
-            f"{settings.COMPONENTS_REGISTRY_PREFIX}-alg-endp-invoc-{self.pk}"
-        )
+        return f"{settings.COMPONENTS_REGISTRY_PREFIX}-alg-endp-inv-{self.pk}"
```
jmsmkn: Is this going to be short enough in production too? You may want to consider shortening both this and the endpoint ids, for the training jobs we use a single letter to give us the most headroom.

koopmant: > Is this going to be short enough in production too?

Yes, it already was. I extended the test for both staging and production. I could shorten it more, but I'm not sure why we'd need it. I could align with the training jobs, but should I also add it to `ModelChoices` in that case? I wasn't sure if I should, since these are a different category.

jmsmkn: It's technically not necessary to have the middle bit at all - this is not being used for multiple models like we do for algorithms vs evaluations. It is good to give users more freedom to set `settings.COMPONENTS_REGISTRY_PREFIX`, say we add another identifier to identify my resources in staging vs yours, or use a longer date etc. It's an annoying limitation to hit. Using `-AE-` and `-AEI-` would give sufficient differentiation and the most freedom in prefix length.

app/grandchallenge/algorithms/models.py
koopmant: I was confused about the defaults because on Jobs the stderr has one, but stdout has not. When a row gets created the value will be empty indeed. Because null=False that empty value will be an empty string for this field. But I suppose that is all implicit; I'll set the defaults. 

I'll leave the stdout field on Job untouched.

app/grandchallenge/algorithms/models.py
koopmant: Good point. All right, I will set it to that. I'll leave it out of ModelChoices.

app/grandchallenge/algorithms/models.py
jmsmkn: Ah, weird, yes I think it would be clearer with a default. You could add it to the Job field and update the migration where the field was last changed as it would be an SQL no-op.

koopmant: Added to the field on ComponentJob and updated the migrations.

jmsmkn (approved):
LGTM!

PR #4711, opened 2026-05-11
Invoice annotation for approved compute

chrisvanrun:
Part of the pitch:

This adds a cached `compute_costs_utilized_euros_millicents`, but does not provide logic to fill it out yet. That comes when invoice and utilization will be linked. Which will be a follow-up.

In addition, it adds the tests and logic that uses the `Invoice` properties:
- `expiry`
- `payment status`
- `payment_type`
- `compute_costs_euro`
- `compute_costs_utilized_euros_millicents`

To calculate what amounts to the net or balance of the invoice. It stores the result in: `compute_costs_balance_euros_millicents`. This property can later be used to display the state of an invoice and to select an invoice for a utilization charge.


app/grandchallenge/invoices/models.py
```diff
@@ -94,6 +104,69 @@ def status_aggregates(self):
             num_is_due=Count("is_due", filter=Q(is_due=True), distinct=True),
         )
 
+    def with_approved_compute(self):
```
amickan: Would maybe `available_compute` be a more suitable title?

app/grandchallenge/invoices/models.py
```diff
@@ -94,6 +104,69 @@ def status_aggregates(self):
...
+                    ),
+                    When(
+                        payment_type=PaymentTypeChoices.PREPAID,
+                        payment_status=PaymentStatusChoices.PAID,
+                        then=F("utilized_compute_cost_euro_millicents"),
+                    ),
+                    default=Value(0),
+                    output_field=models.PositiveIntegerField(),
+                ),
+            ),
```
amickan: I'm confused. Shouldn't the available/approved compute always be 0 when it's expired, regardless of the type and status of the invoice? 

I guess I misunderstand what this annotation will be used for. It's a bit difficult to review without seeing how it will be used down the line.

chrisvanrun: The Quick preview:

```
   def with_compute_balance(self):
        return self.with_approved_compute().annotate(
            balance_compute_euros_millicents=ExpressionWrapper(
                (
                    F("approved_compute_euros_millicents")
                    - F("utilized_compute_cost_euro_millicents")
                ),
                output_field=models.BigIntegerField(),
            ),
        )
```

chrisvanrun:
We had a sync discussion about this, and I am going to change this a bit to only show the balance.

app/grandchallenge/invoices/models.py
```diff
@@ -94,6 +104,75 @@ def status_aggregates(self):
...
+            compute_costs_euros__gt=0,
+            payment_type=PaymentTypeChoices.PREPAID,
+            payment_status=PaymentStatusChoices.PAID,
+        )
+
+        qs = self.with_expired_status().annotate(
+            compute_costs_diff_euros_millicents=ExpressionWrapper(
+                F("compute_costs_euros") * 100_000
+                - F("compute_costs_utilized_euros_millicents"),
+                output_field=models.PositiveIntegerField(),
```
amickan: This can be negative, so you probably want to use `IntegerField` here, right?

amickan: Looks like I reviewed too early. This just changed.

app/grandchallenge/invoices/models.py
```diff
@@ -94,6 +104,74 @@ def status_aggregates(self):
...
+            payment_status=PaymentStatusChoices.PAID,
+        )
+
+        qs = self.with_expired_status()
+
+        neg_utilized_costs = F("compute_costs_utilized_euros_millicents") * -1
+        diff = F("compute_costs_euros") * 100_000 - F(
+            "compute_costs_utilized_euros_millicents"
+        )
+        capped_diff = Least(diff, Value(0))
```
chrisvanrun: I found the query with these inserts more readable. Not sure if everyone shares that observation?

app/grandchallenge/invoices/models.py
```diff
@@ -94,6 +104,74 @@ def status_aggregates(self):
...
+                        payment_type=PaymentTypeChoices.POSTPAID,
+                        then=capped_diff,
+                    ),
+                    When(
+                        payment_type=PaymentTypeChoices.PREPAID,
+                        payment_status=PaymentStatusChoices.PAID,
+                        then=capped_diff,
+                    ),
+                    default=neg_utilized_costs,
+                    output_field=models.PositiveIntegerField(),
```
amickan: This can be negative, so I think an `IntegerField` would be more suitable here? Same below

app/grandchallenge/invoices/models.py
```diff
@@ -94,6 +104,74 @@ def status_aggregates(self):
...
+                then=Case(
+                    # Cap at zero on expiry: approved invoices can't go positive
+                    When(
+                        payment_type=PaymentTypeChoices.COMPLIMENTARY,
+                        then=capped_diff,
+                    ),
+                    When(
+                        payment_type=PaymentTypeChoices.POSTPAID,
+                        then=capped_diff,
+                    ),
```
amickan: Is that right? Shouldn't this be similar to prepaid and only `capped_diff` when paid, otherwise `neg_utilized`?

app/grandchallenge/invoices/models.py
chrisvanrun: BigIntegerField even. Good catch!

app/grandchallenge/invoices/models.py
chrisvanrun: Interesting case. If there is a POSTPAID, and it is expired. IMO it should either be 0 or represent negative balance. 

Take for instance a POSTPAID that is fully approved (i.e. initialized) and then it transitions into expiry. Should it then suddenly act as a full negative drain on the budget? I would instead see it as neutral then: you can't pin new costs to it, and if it was overcharged we should take that into account in the overall challenge-wide considerations. That is what the capped_diff establishes.

app/grandchallenge/invoices/models.py
chrisvanrun: Huh, why didn't this show up in the tests? There are negative values there...

app/grandchallenge/invoices/models.py
chrisvanrun: See `test_postpaid_overutilized` and `test_postpaid_overutilized_but_expired`.

amickan: Hm, but `capped_diff` for expired POSTPAID means any utilized compute under the budget effectively becomes free, right?

chrisvanrun: POSTPAID is already free: we don't require payment of that up front. If we ask orgs to pay, but orgs refuse, we can cancel the invoice. Until that point, an expired POSTPAID with some utilized compute should not be blocking any submissions while there is another invoice that could be charged.

Realistically we shouldn't see that very often: but I can imagine two different POSTPAIDs (different grants for instance) and we'd first use up one, and then the other.

amickan: Ok, true, so long as the utilization on the post-paid is taken into in the overall calculation to determine whether submissions are still possible, this works.

chrisvanrun: Not sure if I follow. The utilization on the post-paid is taken into account only under certain conditions. 

The balance is meant for 3 things:
- positive, please use me to charge costs!
- zero, utilization and compute_costs are in balance (or when expired, the utilization is covered).
- negative, ups someone overcharged and I should have a say to block charges landing on other invoices.


Currently, my plan is to have the go-no/go for a charge on the challenge budget be a simple sum of all balances. If that is positive, we have a go, otherwise no-go.

amickan: Chris and I discussed this synchronously, and I understand it better now and agree with the approach taken here.

amickan (dismissed):
LGTM

app/grandchallenge/invoices/models.py
```diff
@@ -94,6 +104,74 @@ def status_aggregates(self):
             num_is_due=Count("is_due", filter=Q(is_due=True), distinct=True),
         )
 
+    def with_compute_balance(self):
```
chrisvanrun: Hej @jmsmkn, would you mind having a look at this query when you get a chance? Curious if you spot any room for optimization.

chrisvanrun:
Rebased (following invoice migration merge onto main)

app/grandchallenge/invoices/models.py
jmsmkn: There are a number of footguns with using database annotations, in particular when using subqueries and Q filtering. You would need to test this with all combinations of invoice type, payment state, compute costs, and multiple challenges.

Honestly, I do not see the need to do this in the database and I think that this is premature optimisation. If I were you I would implement the logic on the challenge level - that is the only place we need it and avoids the need to repeat the challenge invoice subquery. I would also do it in python as there it is much clearer what is happening, and as there are on the order of 10's of invoices per challenge there is no significant performance gain.

chrisvanrun:
Pulled the computation of the balance fully into Python. @amickan, the tests are almost identical except that I added one with two challenges (just to ensure there is no cross contamination of prepaid status).

app/grandchallenge/challenges/models.py
```diff
@@ -929,6 +929,57 @@ def forge_model(self):
...
+            for invoice in invoices
+        )
+        for invoice in invoices:
+            invoice.compute_costs_balance_euros_millicents = (
+                self._get_invoice_compute_costs_balance(
+                    invoice=invoice,
+                    has_paid_prepaid_invoice=has_paid_prepaid_invoice,
+                )
+            )
+            yield invoice
```
jmsmkn: Why return a generator here? You're already iterating over all invoices above.

chrisvanrun: Not per see. The `any` is lazy. With the general number of invoices, it is perfectly fine to return a list here - I think. At the same time I couldn't really come up with a good reason not to yield it.

jmsmkn: Because we're using Django and that if you're passing around objects you should be passing around querysets.

app/grandchallenge/challenges/models.py
```diff
@@ -929,6 +929,57 @@ def forge_model(self):
...
+
+    @staticmethod
+    def _get_invoice_compute_costs_balance(
+        *, invoice, has_paid_prepaid_invoice
```
jmsmkn: You're passing the invoice to set a property on the invoice, this suggests that it should be a method on the invoice, even better would be a cached property.

chrisvanrun: Solid point. I considered that but the overall check on existing (paid) prepaid invoices had me pull it up to the challenge level. Re-using the `is_expired` annotation is a nice bonus.

Re-examining that decision I realize that existing prepaid paid check can be made into a cached property on the challenge level... that way this property can be pushed to the correct model.

app/grandchallenge/challenges/models.py
chrisvanrun: Good argument!

app/grandchallenge/challenges/models.py
chrisvanrun: The `is_expired` annotation now seems a bit superfluous. The purist in me wants to fix what the platform considers 'today' and always compare the same date, but I think having a cached function somewhere to base it off is a bit too complex of a solution if its only goal is to cover the miniscule chance of incongruity in list views.

app/grandchallenge/invoices/models.py
```diff
@@ -310,6 +303,42 @@ def _current_state(self):
...
+    def is_expired(self):
+        return self.expires_on < now().date()
+
+    @cached_property
+    def compute_costs_balance_euros_millicents(self):
+        utilized = self.compute_costs_utilized_euros_millicents
+        diff = self.compute_costs_euros * 1000 * 100 - utilized
+
+        if self.payment_status == PaymentStatusChoices.CANCELLED:
+            return -utilized
```
amickan: This is technically already covered by `if not is_eligible:`, right?

chrisvanrun: Nope. COMPL with status CANCELLED would pass. But you did give me inspiration to pull out that boolean check, push it into a property. IMHO it is about as clean as it is going to get.

app/grandchallenge/invoices/models.py
```diff
@@ -307,30 +307,32 @@ def _current_state(self):
     def is_expired(self):
         return self.expires_on < now().date()
 
+    @property
+    def budget_is_utilizable(self):
```
amickan: Good to make this a separate property, but I think the name is misleading. I would expect the expiry date to be taken into account in determining whether budget is utilizable. I liked the `is_eligible` you had initially. Maybe make it even more explicit and call it `eligible_for_utilization_charge` or something like that?

amickan: Hm, reading this again, I think the name I suggested has the same issue, actually, that it suggests that expiry is taken into account.

jmsmkn:
I find this far easier to follow, thanks!

app/grandchallenge/invoices/models.py
chrisvanrun: `is_budget_active` ?

chrisvanrun: No, that has the same problem... ehm. `is_budget_authorized`? I think that hits the spot!

jmsmkn (approved):
LGTM!

amickan (approved):
LGTM too

PR #4713, opened 2026-05-11
Weekly open invoice email to staff

amickan:
This adds a weekly task that checks if there are post-paid invoices that require our attention or invoices in intermediate state (issues, requested). If any of those invoices exist, it sends an email to staff listing those invoices. 

Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/475

### DIFF
--- a/app/grandchallenge/invoices/models.py
+++ b/app/grandchallenge/invoices/models.py
@@ -107,7 +107,7 @@ class InvoiceQuerySet(models.QuerySet):
             | Q(
                 payment_type=Invoice.PaymentTypeChoices.POSTPAID,
                 payment_status=Invoice.PaymentStatusChoices.INITIALIZED,
-                follow_up_on__lt=now().date(),
+                follow_up_on__lte=now().date(),
             )
         )
 
--- a/app/grandchallenge/invoices/models.py
+++ b/app/grandchallenge/invoices/models.py
@@ -96,6 +96,21 @@ class InvoiceQuerySet(models.QuerySet):
             num_is_due=Count("is_due", filter=Q(is_due=True), distinct=True),
         )
 
+    def to_check(self):
+        return self.filter(
+            Q(payment_status=Invoice.PaymentStatusChoices.REQUESTED)
+            | Q(payment_status=Invoice.PaymentStatusChoices.ISSUED)
+            | Q(
+                payment_type=Invoice.PaymentTypeChoices.PREPAID,
+                payment_status=Invoice.PaymentStatusChoices.INITIALIZED,
+            )
+            | Q(
+                payment_type=Invoice.PaymentTypeChoices.POSTPAID,
+                payment_status=Invoice.PaymentStatusChoices.INITIALIZED,
+                follow_up_on__lt=now().date(),
+            )
+        )
+
 
 def default_invoice_expiry():
     return now().date() + relativedelta(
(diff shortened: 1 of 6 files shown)


app/grandchallenge/invoices/tasks.py
```diff
@@ -29,3 +32,51 @@ def send_challenge_invoice_issued_notification_emails(*, pk):
...
+@transaction.atomic
+def send_open_invoices_email():
+    from grandchallenge.invoices.models import Invoice
+
+    subject = "Invoices to check"
+
+    post_paid_invoices_to_follow_up = Invoice.objects.filter(
+        payment_type=Invoice.PaymentTypeChoices.POSTPAID,
+        payment_status=Invoice.PaymentStatusChoices.INITIALIZED,
+        follow_up_on__lte=now().date(),
```
jmsmkn, line 47: Shouldn't they only be followed up when the date has passed?

jmsmkn (commented):
Just an idea, but you could have a filter in the admin and link to that:

```python3
class InvoiceQuerySet(models.QuerySet):

    def to_check(self):
        return self.filter(
            Q(payment_status=Invoice.PaymentStatusChoices.REQUESTED)
            | Q(payment_status=Invoice.PaymentStatusChoices.ISSUED)
            | Q(
                payment_type=Invoice.PaymentTypeChoices.PREPAID,
                payment_status=Invoice.PaymentStatusChoices.INITIALIZED,
            )
            | Q(
                payment_type=Invoice.PaymentTypeChoices.POSTPAID,
                payment_status=Invoice.PaymentStatusChoices.INITIALIZED,
                follow_up_on__gt=now().date(),
            )
        )

class Invoice(models.Model):
    objects = InvoiceQuerySet.as_manager()
    ...
```

```python3
class ToCheckFilter(admin.SimpleListFilter):
    title = "to check"
    parameter_name = "to_check"

    def lookups(self, request, model_admin):
        return [("1", "Yes")]

    def queryset(self, request, queryset):
        return queryset.to_check()


@admin.register(Invoice)
class InvoiceAdmin(admin.ModelAdmin):
    list_filter = [ToCheckFilter]
```

app/grandchallenge/invoices/tasks.py
amickan: Yes, but I thought since we send this email only once a week and since following up anyway probably takes us a day or two, including invoices with a follow-up "today" would be fine. 

But maybe it's safer to leave those out after all.

app/grandchallenge/invoices/models.py
```diff
@@ -96,6 +96,21 @@ def status_aggregates(self):
...
+            Q(payment_status=Invoice.PaymentStatusChoices.REQUESTED)
+            | Q(payment_status=Invoice.PaymentStatusChoices.ISSUED)
+            | Q(
+                payment_type=Invoice.PaymentTypeChoices.PREPAID,
+                payment_status=Invoice.PaymentStatusChoices.INITIALIZED,
+            )
+            | Q(
+                payment_type=Invoice.PaymentTypeChoices.POSTPAID,
+                payment_status=Invoice.PaymentStatusChoices.INITIALIZED,
+                follow_up_on__lt=now().date(),
```
jmsmkn: Maybe I wasn't clear enough, it wasn't the equals that was the issue. This should be greater than (or greater than or equal) in order to follow up, not less than, no?

```suggestion
                follow_up_on__gt=now().date(),
```

amickan: The follow-up date should be in the past, so I think it should be less than today. Greater than would mean it's in the future, no?

jmsmkn: Sorry! Of course. My mistake. Then you could add the `e` back if you like, up to you.

jmsmkn (approved):
LGTM!

PR #4717, opened 2026-05-13
Handle endpoint invocation events

koopmant:


app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
```diff
@@ -8,6 +13,17 @@
 from grandchallenge.components.backends.base import (
     list_and_delete_objects_from_prefix,
 )
+from grandchallenge.components.backends.exceptions import ComponentException
+from grandchallenge.components.backends.utils import UUID4_REGEX
+from grandchallenge.core.error_messages import SystemErrorMessages
+
+logger = logging.getLogger(__name__)
```
jmsmkn, line 21: If this is only being used from lambda tasks then that has its own logger: https://github.com/DIAGNijmegen/rse-django-lambda-tasks#logging

app/grandchallenge/components/tasks.py
```diff
@@ -1805,3 +1805,60 @@ def invoke_endpoint(*, pk: uuid.UUID, app_label: str, model_name: str):
 
     else:
         invocation.update_status(status=invocation.StatusChoices.EXECUTING)
+
+
+@lambda_task(retry_on=(LockNotAcquiredException,))
```
jmsmkn, line 1811: Have you checked that the permissions fit without having to refactor how they are set on lambda tasks? I think the existing policy is almost at the 6000 character limit.

koopmant: If I look at the json of LAMBDA_rumc-gc-s-1802-lambda-tasks_TASK policy, it's now at more than 10,000 chars. But in testing it worked fine? How can that be?

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
koopmant: Using that instead now

app/grandchallenge/components/tasks.py
koopmant: Discussed synchronously, the count included spaces.

app/grandchallenge/components/tasks.py
```diff
@@ -1805,3 +1806,59 @@ def invoke_endpoint(*, pk: uuid.UUID, app_label: str, model_name: str):
...
+            pk=invocation_params.pk
+        )
+
+    if invocation.status == invocation.StatusChoices.CANCELLED:
+        # Nothing to do
+        return
+    elif invocation.status != invocation.StatusChoices.EXECUTING:
+        raise RuntimeError(
+            "Invocation is not in the expected state for handling events"
+        )
```
amickan: Do you really want to raise an error here? I think we usually just lo the error to sentry. 
And why handle cancelled separately?

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
```diff
@@ -234,3 +259,57 @@ def invoke_endpoint(self, *, inference_id):
...
+    def _handle_expired_invocation(self, *, event):
+        # Requires investigation
+        task_logger.info(event)
+        task_logger.error("Endpoint invocation expired")
+        raise ComponentException(SystemErrorMessages.UNEXPECTED_ERROR)
+
+    def _handle_failed_invocation(self, *, event):
+        # Requires investigation
+        task_logger.info(event)
+        task_logger.error("Endpoint invocation failed")
```
amickan, line 314: Just for my understanding: user errors do not result in failed invocations but would show up in completed invocations, right?

app/grandchallenge/components/tasks.py
koopmant: Ah this is the one time where we actually expect multiple calls to the task. I guess we don't even want to log an error then, we don't for the job events task either. 
In case we do log an error, we handle `cancelled` separately, because that should never log an error; we just ignore it.

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
koopmant: I believe so. Currently testing to be sure.

app/grandchallenge/components/tasks.py
amickan: Yeah, I think in line with how we do things for jobs, I would not log an error actually.

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
koopmant: User errors would indeed show up in `_handle_completed_invocation`. (There we inspect the `users_process_exit_code` and if it isn't 0, we mark the invocation as failed.) So the invocation would ultimately be marked as failed, but the invocationStatus we receive from Sagemaker would be `Completed`.

amickan: Right, that is how it works for jobs as well, just wanted to double check.

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
```diff
@@ -234,3 +259,57 @@ def invoke_endpoint(self, *, inference_id):
...
+        else:
+            raise ValueError("Invalid invocation status")
+
+    def _handle_completed_invocation(self):
+        self._executor._handle_completed_job()
+
+    def _handle_expired_invocation(self, *, event):
+        # Requires investigation
+        task_logger.info(event)
+        task_logger.error("Endpoint invocation expired")
```
amickan, line 308: Would that give us enough information in sentry to debug? Maybe include the inference_id just to be sure? Or maybe sentry would show us the `task_logger.info(event)` call as a breadcrumb?

amickan (approved):
LGTM! Just a small question about logging to sentry.

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
koopmant: > Or maybe sentry would show us the task_logger.info(event) call as a breadcrumb?

I assumed it would. I'm not actually sure. Will need to test...

koopmant: I'll test it later today. Will merge so I can rebase the branch with the parse outputs task to have it ready for review.

PR #4718, opened 2026-05-15
Link `ComponentJobUtilization` and `Invoice`

chrisvanrun:
Part of pitch:

This changeset introduces a link between `ComponentJobUtilization` and a challenge's `Invoice`, enabling compute costs to be tracked and summed on a per-invoice basis.

## Invoice Linking

Active invoices are resolved via `Challenge.active_invoice`, which raises an error if the challenge has insufficient funds.

**`EvaluationUtilization`** is linked at evaluation creation time (`create_evaluation`), which is triggered by:
- Creating a submission
- Re-evaluating via Django admin (site admin)
- Re-evaluating via view (editor)

In all three cases, the user receives an appropriate error message if the challenge has insufficient funds.

**`JobUtilization`** is linked when creating jobs for an evaluation. A fresh active invoice is fetched per job — if none is found, the `Job` is marked `FAILURE` with a descriptive error message.

**`WarmJobUtilization`** is linked after `Job` completion.

## Compute Cost Caching

`update_challenge_compute_costs` now sums linked utilizations on a per-invoice basis. The existing `Challenge`-level `compute_cost_euro_millicents` field is kept intact for now — removing it and its related paths will be addressed in a follow-up PR.

## Management Command: `link_challenge_utilization_to_invoice`

Since existing challenges have no invoice-utilization linkage, a management command is introduced to backfill this (rather than a one-off migration), with the longer-term goal of enforcing a constraint that `phase`, `challenge`, and `invoice` are always set together on utilizations.

The command reports cleanly which challenges lack invoices.


app/grandchallenge/algorithms/models.py
```diff
@@ -1439,8 +1439,8 @@ def get_or_create_display_set(self, *, reader_study):
 
         return display_set
 
-    def create_utilization(self):
-        JobUtilization.objects.create(job=self)
+    def create_utilization(self, *, invoice):
+        JobUtilization.objects.create(job=self, invoice=invoice)
```
chrisvanrun: Technically this is not used, but IMO it should match the abstract signature.

app/grandchallenge/challenges/models.py
```diff
@@ -937,6 +938,37 @@ def has_paid_prepaid_invoice(self):
...
+    def compute_costs_balance_euros_millicents(self):
+        return self._get_compute_costs_balance_euros_millicents(
+            invoices=self.invoices.all()
+        )
+
+    def _get_compute_costs_balance_euros_millicents(self, *, invoices):
+        return sum(
+            invoice.compute_costs_balance_euros_millicents
+            for invoice in invoices
+        )
```
chrisvanrun: Not sure about this construct. I mainly introduced it to prevent double database hits in retrieving the `active_invoice`.

app/grandchallenge/challenges/tasks.py
```diff
@@ -78,7 +79,13 @@ def update_challenge_results_cache():
     )
 
 
-def retry_with_backoff(exceptions, max_attempts=5, base_delay=1, max_delay=10):
+def retry_with_backoff(
+    *,
```
chrisvanrun: Since I was touching this, I thought it was better to pull this up to regular kwargs only.

app/grandchallenge/challenges/tasks.py
```diff
@@ -102,23 +109,34 @@ def wrapper(*args, **kwargs):
...
 
-            @retry_with_backoff((LockNotAvailable,))
-            def save_challenge():
+            @retry_with_backoff(exceptions=(LockNotAvailable,))
+            def save_invoice():
+                invoice.save(
+                    update_fields=("compute_costs_utilized_euros_millicents",)
+                )
+
+                # TODO: remove this once compute_cost_euro_millicents is removed or deprecated
```
chrisvanrun, line 114: Very inefficient, obviously. But OK for now I think.

app/tests/invoices_tests/factories.py
```diff
@@ -30,5 +30,19 @@ class InvoiceFactory(factory.django.DjangoModelFactory):
         )
     )
 
+    @classmethod
+    def create_valid(cls, **kwargs):
```
chrisvanrun, line 34: Given the need to suddenly create a valid Invoice all over the code base, I thought it cleaner to create a quick method to spawn a valid one with compute (and other costs) instead of re-doing it every time.

chrisvanrun:
For unclear reasons, the tests are failing on S3 uploads. Might be a dependency update? Maybe something transient? I've kicked the tests... if they fail again, I'll play a better detective.

app/grandchallenge/utilization/management/commands/link_challenge_utilization_to_invoice.py
```diff
@@ -0,0 +1,119 @@
...
+        parser.add_argument(
+            "--all_challenges",
+            action="store_true",
+            help="Link all challenge utilizations to invoices",
+        )
+
+        parser.add_argument(
+            "--create_missing_invoice",
+            action="store_true",
+            help="Create an invoice if none exists for a challenge",
```
amickan: I thought it was decided that we don't want to do this and instead create invoices manually?

app/grandchallenge/utilization/management/commands/link_challenge_utilization_to_invoice.py
```diff
@@ -0,0 +1,119 @@
...
+            "challenge_short_names",
+            nargs="*",
+            type=str,
+        )
+
+        parser.add_argument(
+            "--all_challenges",
+            action="store_true",
+            help="Link all challenge utilizations to invoices",
+        )
```
amickan: I don't think these arguments are necessary. We should be able to run this command multiple times as is (without changing flags) and it should continue where it stopped last. It will run as an [ECS task](https://github.com/DIAGNijmegen/rse-cloud-infrastructure/pull/261), and we don't want to update the code every time we run it.

app/grandchallenge/utilization/management/commands/link_challenge_utilization_to_invoice.py
```diff
@@ -0,0 +1,119 @@
...
+        else:
+            challenges = Challenge.objects.filter(
+                short_name__in=options["challenge_short_names"]
+            )
+
+        for challenge in challenges:
+            self._link_challenge_utilization(
+                challenge=challenge,
+                create_missing_invoice=options["create_missing_invoice"],
+            )
```
amickan, line 42: You could instead loop over utilizations (and filter early on `invoice__isnull=True`). If you do that, you'd need to precompute the invoice per challenge to not create an N+1 problem, of course.

app/grandchallenge/utilization/management/commands/link_challenge_utilization_to_invoice.py
```diff
@@ -0,0 +1,119 @@
...
+            self.stdout.write(
+                self.style.WARNING(f"{prefix} Found unlinked utilizations.")
+            )
+
+        invoice = (
+            Invoice.objects.filter(challenge=challenge)
+            .order_by("expires_on", "created")
+            .first()
+        )
+
```
amickan: This now uses a single invoice per challenge for all utilizations. I think that is ok given that the overall budget check on submission creation takes all invoices and their potential over-charge into account, just double checking though...

app/grandchallenge/algorithms/models.py
amickan: What do you mean that this is not used? You're calling it in the create method you added, right?

app/grandchallenge/challenges/models.py
amickan: But you're not using the cached property in `active_invoice`, right?

app/grandchallenge/challenges/models.py
```diff
@@ -937,6 +938,37 @@ def has_paid_prepaid_invoice(self):
...
+
+    @property
+    def active_invoice(self):
+        invoices = self.invoices.order_by("expires_on", "created")
+
+        if (
```
amickan: You could use the cached property here.

app/grandchallenge/challenges/models.py
```diff
@@ -937,6 +938,37 @@ def has_paid_prepaid_invoice(self):
...
+            raise InsufficientBudgetError
+
+        for invoice in invoices:
+            if invoice.compute_costs_balance_euros_millicents > 0:
+                return invoice
+
+        # Defensive catch
+        raise ValueError(
+            "Unexpected: this challenge has no available budget in any invoice despite a positive total balance."
+        )
```
amickan: This cannot happen, right? `_get_compute_costs_balance_euros_millicents` checks `compute_costs_balance_euros_millicents` for each invoice already.

app/grandchallenge/challenges/tasks.py
amickan: I think our policy is to only make changes when absolutely necessary, so I would say leave this as is.

amickan: Making this keyword-only, makes the use of the decorator much more verbose... James might have had other reasons for using positional arguments here, not sure. In any case, change this in a separate PR if you think this should be changed ;-)

app/grandchallenge/challenges/costs.py
```diff
@@ -13,30 +14,30 @@
...
-        challenge=challenge
-    )
-
-    update_compute_cost_euro_millicents(
-        obj=challenge,
-        algorithm_job_utilizations=algorithm_job_utilizations,
-        job_warm_pool_utilizations=job_warm_pool_utilizations,
-        evaluation_job_utilizations=evaluation_job_utilizations,
+def annotate_compute_costs(*, invoice):
+    result = Invoice.objects.filter(pk=invoice.pk).aggregate(
```
amickan: Does this really work as expected? I think when you have multiple Sum() aggregations across different related tables in a single aggregate() call, Django performs all the joins simultaneously, which causes rows to be multiplied. I haven't checked your tests yet, just dropping this thought here before I forget.

amickan: I see there are no detailed tests for this method. I would suggest to add one to double check that the aggregation works. And if it does not, just do 3 separate queries and sum in python.

app/grandchallenge/components/models.py
```diff
@@ -1661,6 +1661,11 @@ def retrieve_existing_civs(*, civ_data_objects):
 
         return existing_civs
 
+    def create(self, *args, utilization_invoice=None, **kwargs):
+        component_job = super().create(*args, **kwargs)
+        component_job.create_utilization(invoice=utilization_invoice)
+        return component_job
```
amickan: If I remember correctly, this will only cover direct ComponentJob.objects.create() calls. A direct save on a new instance (`job = ComponentJob()` folllowed by `job.save()`) would bypass this. We might not be creating jobs like that at the moment, but I think it's still safer to leave the call to `create_utilization()` in the model's `save` method.

app/grandchallenge/algorithms/tasks.py
```diff
@@ -164,6 +165,21 @@ def create_algorithm_jobs(
                 input_civ_set=ai.values.all(),
                 use_warm_pool=use_warm_pool,
             )
+            jobs.append(
+                job
+            )  # Keep track of created jobs, even if utilization setup has failed
+
+            if job_utilization_challenge is not None:
```
amickan: What if `job_utilization_challenge` is `None`? That should raise an error as well, right?

amickan: And wait, I'm confused. With the manager method you added, shouldn't you be able to just add `utilization_invoice=job_utilization_challenge.active_invoice` to the create call instead of doing this here? Or why have the manager method at all if you're doing it here?

app/grandchallenge/evaluation/forms.py
```diff
@@ -741,17 +751,14 @@ def clean(self):
...
 
-        # Fetch from the db to get the cost annotations
-        challenge = (
-            Challenge.objects.filter(
-                pk=cleaned_data["submission"].phase.challenge.pk
+        try:
+            invoice = cleaned_data["submission"].phase.challenge.active_invoice
+        except InsufficientBudgetError:
+            raise ValidationError(
+                "Challenge has insufficient budget. Please contact the challenge organizers."
```
amickan: This is a view that is for challenge admins only. So the message should encourage them to contact support to add more budget.

app/grandchallenge/components/models.py
amickan: I see why you pulled it out of the save() method - you don't have access to the invoice there. 

But looking at where you set the invoice, I think you technically don't need this manager method, do you? You can assign the invoice to the utilization in the same way that the phase and challenge are set on the utilization in `create_algorithm_jobs` (same can be done in `create_evaluation`). Or what am I missing?

app/grandchallenge/algorithms/models.py
chrisvanrun: Yes, but the `JobUtilization` is first created without any linking. Then the linking is added (invoice/challenge/phase) and then it is saved. So we never actually use the kwargs here for the `JobUtilization`

app/grandchallenge/algorithms/tasks.py
chrisvanrun: `job_utilization_challenge` is default set to `None` if not provided, I am leaning into that and sort of assumed that meant it gets used when not applicable to challenges.

Tracing it, it seems like there isn't any real use of it outside the challenge-related scheduling. Maybe an older construct with auto running Jobs from archive items. I am hesitant to add to the change set by updating all the related tests. So prefer to keep this as is for now.

As for not adding it to the create call: the intent here is to create a `Job` and mark it as a failure if no invoice can be found. Adding it to the create call would prevent creation. Not withstanding my idea to later add a constraint to utilizations that ensures `.invoice` + `.challenge` + `.phase` to be set in tandem. Saving it with only `.invoice` would be counterproductive in that regard, better to have it all set in one go.

Edit: I am not sure how to go about creating a 'sister' object neatly. Putting it in the base class' `save()` sort-of keeps it in the model? It is certainly 'closer' than a signal. The `objects` manager is a tiny bit further but like you mentioned: it isn't hit if you create an object and then `save()` it. I've reverted the objects manager changes, and have kept the change minimum for now: when adding the constraint we should think about populating the `EvaluationUtilization` properties.

app/grandchallenge/challenges/models.py
chrisvanrun: Would that not hit the database twice?

app/grandchallenge/challenges/models.py
chrisvanrun: It shouldn't. I think it's even mathematically impossible. I was a bit too careful here.

app/grandchallenge/utilization/management/commands/link_challenge_utilization_to_invoice.py
chrisvanrun: https://github.com/DIAGNijmegen/rse-roadmap/issues/475#issuecomment-4397274919

Ah! Re-reading James' comment I see that the idea is to add the invoices manually. Totally misinterpreted that the first time I read that.

app/grandchallenge/utilization/management/commands/link_challenge_utilization_to_invoice.py
chrisvanrun: Indeed, that seemed simpler instead of writing a priority queue to post them to and then switch when one is full.

app/grandchallenge/components/models.py
chrisvanrun: No, the utilization creation is already a bit odd. Given the ad hoc setting I'll pull the Evaluation up to have the same manner as the `JobUlitization`. That way they are at least similar in that regard.

app/grandchallenge/utilization/management/commands/link_challenge_utilization_to_invoice.py
chrisvanrun: Would that not be a dramatic number of items to loop over? This way we update all utilizations in 3 update calls.

chrisvanrun: I've heavily refactored things. Let me know what you think.

app/grandchallenge/challenges/costs.py
chrisvanrun: Ah, there ARE tests. But it never occurred to me how the underlying JOINs and Sums would work. Updated!

app/grandchallenge/challenges/models.py
chrisvanrun: I've removed the cached property. For another day.

app/grandchallenge/utilization/management/commands/link_challenge_utilization_to_invoice.py
```diff
@@ -0,0 +1,79 @@
...
+class Command(BaseCommand):
+    help = "Links challenge utilizations to invoices"
+
+    def handle(self, *_, **__):
+        challenges_with_unlinked_utilization = Challenge.objects.filter(
+            Q(job_utilizations__invoice__isnull=True)
+            | Q(evaluation_utilizations__invoice__isnull=True)
+            | Q(job_warm_pool_utilizations__invoice__isnull=True)
+        ).distinct()
```
amickan: This is going to be a pretty expensive query. Unsure how bad it really is, but maybe replacing it with 3 simple indexed scans on individual tables would be better? Something like this:

```python
challenge_ids = set()
for utilization_model in CHALLENGE_UTILIZATION_MODELS:
    ids = (
        utilization_model.objects
        .filter(invoice__isnull=True)
         values_list("challenge_id", flat=True)
         .distinct()
    )
    challenge_ids.update(ids)

challenges = Challenge.objects.filter(pk__in=challenge_ids)
for challenge in challenges:
```

amickan: To make this more efficient, you could pre-compute the challenge to invoice mapping:

```python
invoice_by_challenge = {
    inv.challenge_id: inv
    for inv in Invoice.objects.order_by("challenge_id", "expires_on", "created")
}
```

That saves you from repeating the same invoice lookup 300+ times.

amickan: I'm also still wondering if looping over utilizations would be better. It would drastically reduce the number of DB queries to execute, but the downside would be that it would require batching /to work, which is less easy to read/follow.

amickan: But actually, if you precompute the challenge-to-invoice mapping, you don't need your expensive initial query. You can just loop over all challenges regardless. So I think that is the most efficient way of going about this after all. 

```python
for challenge in Challenge.objects.all():
    invoice = invoice_by_challenge.get(challenge.pk)

    if not invoice:
        missing_invoice_challenges.append(challenge.short_name)
        continue

    for model in CHALLENGE_UTILIZATION_MODELS:
        model.objects.filter(
            challenge=challenge,
            invoice__isnull=True,
        ).update(invoice=invoice)
```

app/grandchallenge/algorithms/tasks.py
amickan: > So prefer to keep this as is for now.

Makes sense. 

> when adding the constraint we should think about populating the `EvaluationUtilization` properties

Not sure I follow. All 3 are currently set at the same time, so if you add a constraint, the constraint would be satisfied by how we currently do things, right?

amickan:
I still have some comments about the management call, but otherwise this looks good to me now. Would still be good if @koopmant could take a look as well.

app/grandchallenge/algorithms/tasks.py
chrisvanrun: No. The `EvaluationUtilization` has `challenge`, `phase` et cetera extracted from the evaluation during `save()`. In a secondary save the `invoice` is set.

amickan: Oh, I'm only seeing that now. It's a bit confusing that the 3 utilizations have their properties assigned at different times... I can't think of a better way though.

amickan: Maybe we need to link invoices to AlgorithmJobs and Evaluations directly as well, then you can do all the assigning in the save() method of the utlization, and it can be the same for all 3 types of utilizations. But it's duplicate links...

chrisvanrun: Hence, my idea to supercharge the `create` of the object manager with those properties.

app/grandchallenge/utilization/management/commands/link_challenge_utilization_to_invoice.py
chrisvanrun: Updated, and I think resolved.

chrisvanrun: @amickan quick ping. The command has been updated, and about as minimal as it'll get.

amickan: I _think_ this looks good. Still waiting for @koopmant 's 2 cents though. Good to have someone else look at this with a fresh perspective. 

I assume you have tested the command, but of course those tests will never be a true representation of the production DB, so it's hard to test the efficiency of the command.

Maybe to summarize for Thomas: the current approach results in 3 update calls per challenge, so 3*416 = 1248 round trips to the DB (plus the initial challenge-to-invoice mapping call). The number of queries is the same on subsequent runs of the command. Each of those DB calls touches a humungous table (the utilization tables have millions of rows), filters on indexed fields and hence reduces the number of rows to update, but the number of rows to subsequently update can still be in the order of 100.000 rows. I do not know how performant that is, but the assumption here is that since the filtering and updating is all done in the DB, that it would be fine? 

We also discussed looping over utilizations instead of challenges. That would need to be done in batches since it requires loading objects into memory in Python. You'd end up with one update per batch of say 1000, so ultimately more DB roundtrips calls than when looping over challenges, but each nice and small. On subsequent runs of the command, you would skip rows that already have an invoice, so don't redo update calls.

I'm not experienced enough with DB queries to judge which of these is ultimately better.

chrisvanrun: > I assume you have tested the command, but of course those tests will never be a true representation of the production DB, so it's hard to test the efficiency of the command.

Indeed. Tbh, I did not bother adding 100k worth of utilizations as the local dev instance does not match the production in terms of performance. Rather I focused on variants of having invoices and missing invoices.

I think we can reduce the subsequent call number of round trips to the DB if we get a cheap set of challenge_ids that have empty invoice FKs in the 3 utilization tables. The 1248 round trips on every call seems like a lot to me.

chrisvanrun: > I think we can reduce the subsequent call number of round trips to the DB if we get a cheap set of challenge_ids that have empty invoice FKs in the 3 utilization tables. The 1248 round trips on every call seems like a lot to me.

I don't think we can do a 'quick' scan to gather all challenge_ids. The first time around it will be near full table scan =/ per utilization table.

amickan: > I don't think we can do a 'quick' scan to gather all challenge_ids. The first time around it will be near full table scan =/ per utilization table.

Exactly, I don't think that is possible. 

I think 1248 DB round trips is not too bad, as long as each call is manageable. I believe it is ok given that the filter can use indexes, but I don't know for sure. 

It might be nice to print out how many rows the update call touched. Then we know that it did something or (on subsequent calls of the command) that it didn't.

chrisvanrun: Smart. Done.

koopmant:
> Would still be good if @koopmant could take a look as well.

> Still waiting for @koopmant 's 2 cents though. Good to have someone else look at this with a fresh perspective.

Sorry, I thought I would be able to look at this sooner, but kept getting stuck in the weeds in my own pitch. I can look at this next week.

app/grandchallenge/algorithms/tasks.py
koopmant, line 172: For my understanding, is this link still used?

app/grandchallenge/algorithms/tasks.py
koopmant: > job_utilization_challenge is default set to None if not provided, I am leaning into that and sort of assumed that meant it gets used when not applicable to challenges.
> 
> Tracing it, it seems like there isn't any real use of it outside the challenge-related scheduling. Maybe an older construct with auto running Jobs from archive items. I am hesitant to add to the change set by updating all the related tests. So prefer to keep this as is for now.

Fine with me, but let's create an issue for it. I believe James added it when I went on holiday for that pitch. Not sure if there is a reason for setting the default to None.

app/grandchallenge/algorithms/tasks.py
```diff
@@ -165,6 +166,21 @@ def create_algorithm_jobs(
                 input_civ_set=ai.values.all(),
                 use_warm_pool=use_warm_pool,
             )
+            jobs.append(
+                job
+            )  # Keep track of created jobs, even if utilization setup has failed
```
koopmant:
```suggestion
            )
```
The comment might be confusing. I started rewriting it until I realized this comment is not really necessary at all, the code explains itself.

koopmant (commented):
It would have been better if this had been split up into separate PR's. The three headings in the description nicely describe three possible PR's. There is a lot going on here and it was a bit much for me.

Some thoughts:
Do we _have_ to assign the invoice on creation? What if we set the invoice when we set the duration on the utilization; if there is no available budget at that time, we set it at the last invoice. (I don't assume we can quickly update the budget for an invoice at that time?) It might still be a small improvement over assigning early?
If we do assign it on creation, I think adding it as an argument for create() would be nicer, see this next comment. 

> the intent here is to create a Job and mark it as a failure if no invoice can be found

Maybe this is discussed on the pitch, but I don't fully understand why this is necessary? I.e., why not let job creation fail and update the evaluation with the error message? I like the implementation of getting the active invoice or an error if there are no funds. By adding invoice it to the Job create call, we'd prevent being able to create jobs when there are no funds. Now you have to check afterwards and not forget it... Once a job is created we will not check the funds again; will we?

Regarding management commands in general. When do we use a command and when a data migration? Data migrations can be run multiple times as well, can't they?

app/grandchallenge/algorithms/tasks.py
```diff
@@ -165,6 +166,21 @@ def create_algorithm_jobs(
...
+                try:
+                    job.utilization.invoice = (
+                        job_utilization_challenge.active_invoice
+                    )
+                except InsufficientBudgetError:
+                    job.update_status(
+                        status=Job.FAILURE,
+                        error_message="Job cannot be executed. The challenge has insufficient budget to run this job.",
+                    )
+                    break
```
amickan: With James' comment [here](https://github.com/DIAGNijmegen/rse-roadmap/issues/475#issuecomment-4552777686), this will need to be changed. The invoice will need to be passed along as a kwarg to this function.

amickan:
> Do we _have_ to assign the invoice on creation? What if we set the invoice when we set the duration on the utilization; if there is no available budget at that time, we set it at the last invoice. (I don't assume we can quickly update the budget for an invoice at that time?) It might still be a small improvement over assigning early?

Setting it after the fact would potentially allow for more unbudgeted submissions (e.g. on deadline day, parallel submissions), right?

amickan:
> Regarding management commands in general. When do we use a command and when a data migration? Data migrations can be run multiple times as well, can't they?

A data migration would only run once. They are not meant to be run multiple times.

app/grandchallenge/utilization/management/commands/link_challenge_utilization_to_invoice.py
amickan: I discussed the migration approach synchronously with James. Looping over utilizations is the better approach here since the filter on `invoice__isnull=True` is very cheap and allows us to skip already processed rows on subsequent runs. When looping over challenges, we filter utilizations by invoice and challenge. Even though both fields are indexed, they are not indexed together (there is no composite index for these two fields), and so the filter is much less efficient. That combined with the fact that we'd redo these queries unnecessarily on subsequent runs, makes this approach less efficient.

app/grandchallenge/algorithms/tasks.py
jmsmkn: This code expects that the returned jobs are the ones that have been scheduled, so the utilisation set up must not fail.

app/grandchallenge/challenges/costs.py
```diff
@@ -63,6 +65,7 @@ def annotate_job_duration_and_compute_costs(*, phase):
 def update_compute_cost_euro_millicents(
     *,
     obj,
+    obj_field,
```
jmsmkn: Why is this new kwarg necessary?

app/grandchallenge/algorithms/tasks.py
jmsmkn: This can eventually be removed as it is available through the invoice or phase links, no need to shortcut it once the compute costs are managed on the invoice rather than the challenge.

app/grandchallenge/challenges/costs.py
jmsmkn: I see now. The construction indicates a design problem. Please rename `compute_costs_utilized_euros_millicents` to `compute_cost_euro_millicents` on invoices.

This is for several reasons:
- This is the name suggested by the method being used here: `update_compute_cost_euro_millicents`
- Internal consistency: attributes that represent the same thing should be called the same thing
- Historical consistency: we already have an existing name for this property
- Simplified implementation: reduced complexity in passing and setting these attributes, no need for extra kwargs or `setattr`
- Reduced cognitive load: developers do not need to thing about what objects they are dealing with when accessing this attribute

koopmant:
> > Do we _have_ to assign the invoice on creation? What if we set the invoice when we set the duration on the utilization; if there is no available budget at that time, we set it at the last invoice. (I don't assume we can quickly update the budget for an invoice at that time?) It might still be a small improvement over assigning early?
> 
> Setting it after the fact would potentially allow for more unbudgeted submissions (e.g. on deadline day, parallel submissions), right?

Not really, because just linking the utilization does not change the budget. Only once the utilization has a value for what's been utilized (i.e. when we set the duration) and only after the invoice has been updated will the budget change.

chrisvanrun:
> It would have been better if this had been split up into separate PR's. The three headings in the description nicely describe three possible PR's. There is a lot going on here and it was a bit much for me.

Maybe a feature branch then? The coupling could have gone in a different (tiny) PR, but the rest would need to go in in one go.

> Not really, because just linking the utilization does not change the budget. Only once the utilization has a value for what's been utilized (i.e. when we set the duration) and only after the invoice has been updated will the budget change.

I agree with that note. We could add an increment of the overall utilization at the time of setting the duration to the utilized invoice. That would not absolve you from doing a periodic consolidation as that is far simpler than covering all corner cases where the overall sum should be adjusted.

With the change that everything involving a single evaluation will be booked on the same invoice, it's really a lot easier to set it at the time of creation.

chrisvanrun:
Quick to-do here:
- [x] Rename cache value.
- [x] Use evaluation invoice for all jobs. Don't have jobs fail on account of insufficient budget.
- [x] (optional) update management command to loop over utilizations.

app/grandchallenge/challenges/costs.py
chrisvanrun: The name was chosen deliberately to contrast it with the existing `Invoice.compute_costs_euros`. Having these side by side is easy to confuse:

- `Invoice.compute_costs_euros`
- `Invoice.compute_cost_euro_millicents`

`Challenge.compute_cost_euro_millicents` will eventually be removed as it can be replaced by a simple query on the `Invoice` table.

Rather than renaming the `Invoice.compute_costs_utilized_euros_millicents`, would it make more sense to rename the `Phase.compute_cost_euro_millicents` to align with it (i.e. `Phase.compute_costs_utilized_euros_millicents`)?

jmsmkn: That cannot be done without downtime and/or a data migration, so rename the new one.

chrisvanrun:
Given its changed directions, this PR grew widely impractical. The first part is merged via a different PR, the second part is now up for review: https://github.com/DIAGNijmegen/rse-grand-challenge/commit/8b56ac0f5ff75f75a8a00d80c3d9a7b6fd439956

PR #4724, opened 2026-05-20
Invocation serializer

amickan:
This add serializers for the invocation model, a `InvocationPostSerializer` to create invocations and a `HyperlinkedInvocationSerializer` to represent invocations via the public API. 

The post serializer still needs a custom create method that takes care of processing the inputs. And the viewset is also still missing. 

Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/481

### DIFF
--- a/app/grandchallenge/algorithms/serializers.py
+++ b/app/grandchallenge/algorithms/serializers.py
@@ -334,7 +334,7 @@ class HyperlinkedInvocationSerializer(serializers.ModelSerializer):
 
     endpoint = HyperlinkedRelatedField(
         queryset=Endpoint.objects.none(),
-        view_name="api:endpoint-detail",
+        view_name="api:algorithms-endpoint-detail",
         required=True,
     )
     inputs = HyperlinkedComponentInterfaceValueSerializer(many=True)
@@ -350,7 +350,7 @@ class InvocationPostSerializer(serializers.ModelSerializer):
 
     endpoint = HyperlinkedRelatedField(
         queryset=Endpoint.objects.none(),
-        view_name="api:endpoint-detail",
+        view_name="api:algorithms-endpoint-detail",
         required=True,
     )
 
@@ -368,7 +368,9 @@ class InvocationPostSerializer(serializers.ModelSerializer):
             user = self.context["request"].user
 
             self.fields["endpoint"].queryset = filter_by_permission(
-                queryset=Endpoint.objects.all(),
+                queryset=Endpoint.objects.filter(
+                    status=Endpoint.StatusChoices.RUNNING
+                ),
                 user=user,
                 codename="invoke_endpoint",
             )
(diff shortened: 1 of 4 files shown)


app/grandchallenge/algorithms/serializers.py
```diff
@@ -326,3 +327,87 @@ def validate_inputs_and_return_matching_interface(self, *, inputs):
...
+                input_count=len(provided_inputs),
+            )
+            return interface
+        except ObjectDoesNotExist:
+            raise serializers.ValidationError(
+                f"The set of inputs provided does not match "
+                f"any of the endpoint's algorithm's interfaces. This algorithm supports the "
+                f"following input combinations: "
+                f"{oxford_comma([f'Interface {n}: {oxford_comma(interface.inputs.all())}' for n, interface in enumerate(allowed_algorithm_interfaces, start=1)])}"
+            )
```
amickan: I think this can be pulled out of the serializer and can be an independent method that both the job and the invocation serializers use. Just want to double check first that the functionality is correct.

app/grandchallenge/algorithms/serializers.py
```diff
@@ -326,3 +327,87 @@ def validate_inputs_and_return_matching_interface(self, *, inputs):
...
+        data["algorithm_interface"] = (
+            self.validate_inputs_and_return_matching_interface(inputs=inputs)
+        )
+        data["civ_data_objects"] = convert_deserialized_civ_data(
+            deserialized_civ_data=inputs
+        )
+
+        return data
+
+    def validate_inputs_and_return_matching_interface(self, *, inputs):
```
koopmant: Why don't we just pass algorithm instead of the self._endpoint trick? Or perhaps even better to pass allowed_algorithm_interfaces to this function? 
We could do the same in the JobPostSerializer.

app/grandchallenge/algorithms/serializers.py
koopmant: Functionality seems correct to me! 

When we create an independent method we can pass `allowed_algorithm_interfaces` and no longer define instance attributes outside __init__. (self._endpoint here and self._algorithm in JobPostSerializer)

koopmant (dismissed):
LGTM

app/grandchallenge/algorithms/serializers.py
amickan: Exactly. That's what I was going to do.

koopmant: Great minds...

amickan: I decided to push this update to this PR after all. Needs re-approval now. Sorry.

koopmant (approved):
LGTM!

app/grandchallenge/algorithms/serializers.py
koopmant: No worries!

koopmant: Don't forget to update the PR description.

amickan: The PR description is still accurate, isn't it?

koopmant: > The post serializer still needs a custom create method that takes care of processing the inputs. 

This is already added now, right?

amickan: No, the create method is still missing.

PR #4725, opened 2026-05-21
Get invocation logs after notification event

koopmant:


koopmant:
Tested in staging and the logs are not always available yet when the event is handled. Will need a different solution.

PR #4726, opened 2026-05-21
Filter structured logs on task

koopmant:
Structured logs from algorithm endpoints can contain messages from multiple invocations. By filtering on `task` these can be separated. 

If no task is provided, all logs will be included, i.e. also the logs from the invocations. 

Related to https://github.com/DIAGNijmegen/rse-roadmap/issues/481


koopmant:
> If no task is provided, all logs will be included, i.e. also the logs from the invocations.

I noted this because the value for `task` in logs not tied to a task is `None`. But when you provide task=None, we don't filter the logs on only those messages, but include all message regardless of the task value.

jmsmkn (approved):
LGTM, but is this actually being used anywhere?

koopmant:
> LGTM, but is this actually being used anywhere?

Not yet! But needed for filtering the invocation logs later

PR #4727, opened 2026-05-21
Add create method for invocation serializer

koopmant:

### DIFF
--- a/app/grandchallenge/algorithms/serializers.py
+++ b/app/grandchallenge/algorithms/serializers.py
@@ -381,7 +381,7 @@ class InvocationPostSerializer(serializers.ModelSerializer):
             )
 
     def validate(self, data):
-        endpoint = data.pop("endpoint")
+        endpoint = data["endpoint"]
         inputs = data.pop("inputs")
         data["algorithm_interface"] = (
             validate_inputs_and_return_matching_interface(
--- a/app/grandchallenge/algorithms/serializers.py
+++ b/app/grandchallenge/algorithms/serializers.py
@@ -21,6 +21,7 @@ from grandchallenge.algorithms.models import (
     AlgorithmModel,
     Endpoint,
     Invocation,
+    InvocationStatusChoices,
     Job,
     annotate_input_output_counts,
 )
@@ -399,3 +400,31 @@ class InvocationPostSerializer(serializers.ModelSerializer):
         )
 
         return data
+
+    def create(self, validated_data):
+        civ_data_objects = validated_data.pop("civ_data_objects", [])
+
+        invocation = Invocation.objects.create(
+            **validated_data,
+            status=InvocationStatusChoices.VALIDATING_INPUTS,
+        )
+
+        try:
+            invocation.process_civ_data_objects_and_execute_linked_task(
+                civ_data_objects=civ_data_objects,
+                user=self.context["request"].user,
+            )
+        except CIVNotEditableException as e:
+            invocation.refresh_from_db()
+            if invocation.status == invocation.StatusChoices.CANCELLED:
+                # this can happen for jobs with multiple inputs
+                # if one of them fails validation
+                pass
+            else:
+                error_handler = invocation.get_error_handler()
+                error_handler.handle_error(
+                    error_message=SystemErrorMessages.UNEXPECTED_ERROR,
+                )
+                logger.error(e, exc_info=True)
+
+        return invocation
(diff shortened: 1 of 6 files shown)


app/grandchallenge/algorithms/serializers.py
```diff
@@ -361,7 +362,12 @@ class InvocationPostSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = Invocation
-        fields = ["pk", "endpoint", "inputs", "status"]
+        fields = ["endpoint", "time_limit", "inputs"]
```
amickan: Are we sure that we don't need the pk to be returned? Cirrus (or potential other callers in the future) don't need to know the pks of the created invocations? For jobs we return this so users can query the status of the jobs after creating them.

amickan:
LGTM! The only thing you're missing now is the `InvocationViewSet`, but I guess you're planning on adding that in another PR?

app/grandchallenge/algorithms/serializers.py
koopmant: Is that how that works... I thought they we're input fields, because this is a PostSerializer. It didn't make sense to me. Are they _also_ input fields when I put them back?

amickan: The fields defined in the serializer are both what's accepted on POST and what's returned in the response. Since the pk is not editable though it is automatically marked as a read_only_field, and so not required on POST (but still serialized in the response if included in the fields).

amickan: If the status should be returned, you would need to mark it as read_only as well: `status = CharField(source="get_status_display", read_only=True)`

app/tests/algorithms_tests/test_serializers.py
```diff
@@ -721,3 +721,94 @@ def test_input_validation_on_invocation_serializer(inputs, interface, rf):
             in str(serializer.errors)
         )
         assert "algorithm_interface" not in serializer.validated_data
+
+
+@pytest.mark.django_db
+def test_time_limit_validation_on_invocation_post_serializer(settings):
```
amickan: Maybe good to add a check that setting it to something less than `ALGORITHM_ENDPOINTS_MAXIMUM_INVOCATION_DURATION` works and that that user-provided time limit is not overriden by the default?

app/grandchallenge/algorithms/serializers.py
koopmant: Thanks for the explanation, starting to understand!

app/tests/algorithms_tests/test_serializers.py
koopmant: I'm wondering if I should make it settable at all. They can only reduce the time limit at the moment, which does not seem very useful.

amickan: True. At the moment it's only for use in Cirrus and I guess there are no plans for Cirrus to change this? Leave it out then.

koopmant: haha, just finished the test. All right, good call. I've removed the field.

koopmant:
> LGTM! The only thing you're missing now is the `InvocationViewSet`, but I guess you're planning on adding that in another PR?

Yes, I'll add the viewset in a separate one.

app/tests/algorithms_tests/test_serializers.py
amickan: Sorry! We can always add it back if we need it later.

app/grandchallenge/algorithms/serializers.py
```diff
@@ -362,6 +363,7 @@ class InvocationPostSerializer(serializers.ModelSerializer):
     class Meta:
         model = Invocation
         fields = ["pk", "endpoint", "inputs", "status"]
+        read_only_fields = ["status"]
```
amickan: Have you checked what the serialized status looks like? I think it would return the all caps status. What we usually do instead is define the field like this: `status = CharField(source="get_status_display", read_only=True)`. That will print the display label that we define in TextChoices (i.e. `Validating inputs` rather than `VALIDATING_INPUTS` for example).

koopmant: Ah good point, you're right: "status": "VALIDATING_INPUTS"

koopmant: Thanks. Updated.

amickan (approved):
LGTM

PR #4728, opened 2026-05-21
Admin phase links

chrisvanrun:
As a non superuser, but being able to support challenges I found I can't readily download starter kits. That is not very handy.

This change set adds a link and sets to use global permissions on the view that downloads starter kits for a challenge phase.

Moreover, it adds a link to configure the algorithm interfaces to the Phase admin, which I've missed a few times already.

### DIFF
--- a/app/grandchallenge/evaluation/admin.py
+++ b/app/grandchallenge/evaluation/admin.py
@@ -102,7 +102,8 @@ class PhaseAdmin(admin.ModelAdmin):
     readonly_fields = (
         "give_algorithm_editors_job_view_permissions",
         "algorithm_interfaces",
-        "algorithm_interface_configuration_links",
+        "algorithm_interface_configuration_link",
+        "download_starter_kit_link",
     )
     form = PhaseAdminForm
 
@@ -111,7 +112,7 @@ class PhaseAdmin(admin.ModelAdmin):
         return instance.open_for_submissions
 
     @staticmethod
-    def algorithm_interface_configuration_links(obj):
+    def algorithm_interface_configuration_link(obj):
         if obj.submission_kind != SubmissionKindChoices.ALGORITHM:
             return "-"
         else:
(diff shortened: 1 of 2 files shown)


app/grandchallenge/evaluation/admin.py
```diff
@@ -100,13 +102,50 @@ class PhaseAdmin(admin.ModelAdmin):
...
+        "download_starter_kit_link",
     )
     form = PhaseAdminForm
 
     @admin.display(boolean=True)
     def open_for_submissions(self, instance):
         return instance.open_for_submissions
 
+    @staticmethod
+    def algorithm_interface_configuration_link(obj):
```
amickan, line 115: These links are already available in the Challenge admin. No harm in repeating them here technically, but maybe you were not aware of that?

app/grandchallenge/evaluation/views/__init__.py
```diff
@@ -1582,6 +1582,7 @@ class PhaseStarterKitDownload(
     View,
 ):
     permission_required = "evaluation.change_phase"
+    accept_global_perms = True
```
amickan: I think the better way to give staff users permission for this would be by adding a check on the user to the `dispatch` method.

app/grandchallenge/evaluation/admin.py
chrisvanrun: I was aware. But there they are linked with only the original slug to go on. Which was horrible for large challenges. UNICORN had numbers. So this route feels safer, albeit technically it's a duplicate. I think having them on two locations is not terrible.

app/grandchallenge/evaluation/views/__init__.py
chrisvanrun: One would have to weave that in via `check_permissions` (or maybe have `get_required_permissions` return an empty list). Seems a bit ugly.

 Is it not the general goal to be as precise with permissions as possible? I thought the global permissions `change_phase` match with what is required to view the `PhaseAdmin` view in the first place. I might be wrong.

amickan: We check object level permissions pretty much everywhere, and the `ObjectPermissionRequiredMixin` that this view uses explicitly sets `accept_global_perms=False`. Accessing this view seems similar to accessing the interface view. Maybe check how we do it there.

amickan (approved):
LGTM

PR #4729, opened 2026-05-22
Refactor reminders for draft challenge requests

### LINKED ISSUE #4721: Draft request reminders not send often enough.
Depending on the creation date the current crontab has it sometimes send only a single reminder.

We should also start sending reminders earlier: after 2 weeks.

More-over, the current date range check doesn't work quite nicely. IMHO it is likely less complex to keep a counter on the request.

chrisvanrun:
Closes #4721 

Unlike the follow-up date, the draft `ChallengeRequest` reminders is operating on a time window which doesn't quite work.

Thanks to Miriam for spotting this:
> If the request was made on 21-4 then there is a check on 1-5 (too young: 10 days), on 14-5 (23 days: reminder), and then on 1-6 (40 days: too old).

Instead of breaking my head over a window date range that might work, I simplified things with a straight forward counter. 

Frequency of them being sent is the frequency of the task, unchanged and is every 1st and 14th of the month.

It also reduces the age at which reminders starting to get sent to a minimum of 7 days.

Any current requests still in draft will receive a few extra reminders for the next few weeks. Which IMHO is fine. I intentionally did not add a migration step to counter that.


app/grandchallenge/challenges/tasks.py
```diff
@@ -244,8 +244,9 @@ def send_challenge_request_draft_reminder_emails():
     for c in ChallengeRequest.objects.filter(
         status=ChallengeRequest.ChallengeRequestStatusChoices.DRAFT,
         created__lte=now()
-        - settings.CHALLENGE_REQUEST_AGE_START_REMINDER_CUTOFF,
-        created__gte=now()
-        - settings.CHALLENGE_REQUEST_AGE_END_REMINDER_CUTOFF,
+        - settings.CHALLENGE_REQUEST_AGE_START_DRAFT_REMINDER_CUTOFF,
+        draft_reminder_count__lt=settings.CHALLENGE_REQUEST_MAX_DRAFT_REMINDERS,
```
amickan, line 235: I think this is not actually defined in settings, is it?

chrisvanrun: Ah silly, had 2 commits in the tube! That is what you get when doing a lot of context switching =)

app/grandchallenge/challenges/models.py
```diff
@@ -1029,6 +1029,12 @@ class ChallengeRequest(UUIDModel, ChallengeBase):
             "This email address will be visible on the challenge website, so please provide an address that you are comfortable sharing publicly."
         ),
     )
+    draft_reminder_count = models.PositiveSmallIntegerField(
+        default=0,
```
jmsmkn, line 1095: This field should not be editable.

app/grandchallenge/challenges/tasks.py
```diff
@@ -244,8 +244,9 @@ def send_challenge_request_draft_reminder_emails():
...
     ):
         send_challenge_requests_draft_reminder(challenge_request=c)
+        c.draft_reminder_count += 1
+        c.save(update_fields=["draft_reminder_count"])
```
jmsmkn: The task is not a singleton so it would be more robust to do this with a single update:

```python3
requests = ChallengeRequest.objects.filter(
        status=ChallengeRequest.ChallengeRequestStatusChoices.DRAFT,
        created__lte=now()
        - settings.CHALLENGE_REQUEST_AGE_START_DRAFT_REMINDER_CUTOFF,
        draft_reminder_count__lt=settings.CHALLENGE_REQUEST_MAX_DRAFT_REMINDERS,Expand commentComment on line R248Resolved
    )

for request in requests:
    send_challenge_requests_draft_reminder(challenge_request=request)

requests.update(draft_reminder_count=F("draft_reminder_count") + 1)
```

chrisvanrun: It is superior in that it would prevent an additional DB query per reminder! I am not sure why you would say it is more robust? Any failure would be caught by the transaction wrapper, and doing the update in a single call wouldn't catch other spawned (same) tasks to also send the email afaik.

jmsmkn: The task is not a singleton, so n-tasks can run at the same time. Updating the value in python takes the read value and increments it. If n tasks run at the same time, they read `draft_reminder_count` as 0, and then all set it to 1. But n emails have been sent. Using `F(...)` does the update in the database, so the counter gets updated correctly. See https://docs.djangoproject.com/en/6.0/ref/models/expressions/#f-expressions

chrisvanrun: Aaaaah! The increment is applied at the commit of a transaction if using an `F()` construct, and a pure increment in that it's not calculating the new value during the transaction but only at the time of commit in the DB. Yep, way robuster!

Today I learned something new! Thank you!

chrisvanrun:
This had gone a bit stale + including the fix commits I had locally but not yet tested. Did a rebase and forced push to get it fresh.

jmsmkn:
The migration needs to be re-generated due to the addition of `editable=False`. Does this all still make sense with the cron definition for this task? From our terraform:

```
task_name = "grandchallenge.challenges.tasks.send_challenge_request_draft_reminder_emails"
schedule  = "cron(0 6 1,14 * ? *)"
```

chrisvanrun:
>  Does this all still make sense with the cron definition for this task?

Yes.

app/grandchallenge/challenges/tasks.py
```diff
@@ -228,11 +228,14 @@ def send_onboarding_task_reminder_emails():
...
-    ):
-        send_challenge_requests_draft_reminder(challenge_request=c)
+        - settings.CHALLENGE_REQUEST_AGE_START_DRAFT_REMINDER_CUTOFF,
+        draft_reminder_count__lt=settings.CHALLENGE_REQUEST_MAX_DRAFT_REMINDERS,
+    )
+
+    for request in requests:
+        send_challenge_requests_draft_reminder(challenge_request=request)
+
+    requests.update(draft_reminder_count=F("draft_reminder_count") + 1)
```
amickan, line 241: Do we need to lock the request for updating?

amickan: I see an earlier discussion on this PR about this construct, and if I understand the conversation correctly, locking is not necessary.

amickan (approved):
LGTM

app/grandchallenge/challenges/tasks.py
chrisvanrun: Indeed, there is no need for a lock since the update is done incrementally on the DB level.

PR #4730, opened 2026-05-22
Move copy button to the right

### LINKED ISSUE #4714: Design consistency: Move all buttons to the left or right
_Originally posted by @pkcakeout in https://github.com/DIAGNijmegen/rse-grand-challenge/issues/4697#issuecomment-4421691449_:

I know that it is merged and everything, but just to prevent things from a design perspective, you should likely choose a location for "action buttons" on grand-challenge as a whole:

image

Some action buttons ("Copy") are now to the left of containers, others are to the right of containers ("Hide this result"). Long-term this will lead to "button soup syndrome". I would choose a side.

---------------

As requested by @jmsmkn, converted into an issue. I am not 100% sure _what_ to do about this. This is kind of a mix between a "grand vision UI"-decision and then combing through the application as needed to adapt interfaces.

But, just in general, I think that it is worth it to choose one place for - all kinds of buttons, be it the left, or the right. That trains the mind of users and leads to less confusion and mental load while for us when adding new buttons or for the users when using the buttons.

amickan:
Closes #4714 

Screenshot 2026-05-22 at 13 09 25

### DIFF
--- a/app/grandchallenge/core/templatetags/copy_button.py
+++ b/app/grandchallenge/core/templatetags/copy_button.py
@@ -11,7 +11,7 @@ def copy_button(value, link=None):
     button = format_html(
         "<button "
         'type="button" '
-        'class="btn btn-link text-dark p-0 m-0 mr-1 copy-btn shadow-none" '
+        'class="btn btn-link text-dark p-0 mt-0 mb-1 mx-1 copy-btn shadow-none" '
         'data-copy="{value}" '
         'data-placement="left" '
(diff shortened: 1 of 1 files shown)


chrisvanrun (approved):
LGTM!

pkcakeout:
*Respect* :+1: - In CIRRUS-land this would be lieing on the issue stack for 2 years until someone finds time because it gets burried under other viewer issues that prevent users from using he workstation :cry:

PR #4733, opened 2026-05-22
Add invocation REST API viewset

koopmant:

### DIFF
--- a/app/grandchallenge/algorithms/views.py
+++ b/app/grandchallenge/algorithms/views.py
@@ -33,10 +33,7 @@ from rest_framework.mixins import (
     ListModelMixin,
     RetrieveModelMixin,
 )
-from rest_framework.permissions import (
-    DjangoModelPermissions,
-    DjangoObjectPermissions,
-)
+from rest_framework.permissions import DjangoObjectPermissions
 from rest_framework.viewsets import GenericViewSet, ReadOnlyModelViewSet
 
 from grandchallenge.algorithms.filters import (
@@ -926,7 +923,7 @@ class InvocationViewSet(
     CreateModelMixin, RetrieveModelMixin, ListModelMixin, GenericViewSet
 ):
     queryset = Invocation.objects.all()
-    permission_classes = [DjangoModelPermissions]
+    permission_classes = [DjangoObjectPermissions]
     filter_backends = [DjangoFilterBackend, InvocationObjectPermissionsFilter]
     filterset_class = InvocationFilter
 
--- a/app/grandchallenge/algorithms/views.py
+++ b/app/grandchallenge/algorithms/views.py
@@ -42,8 +42,8 @@ from rest_framework.viewsets import GenericViewSet, ReadOnlyModelViewSet
 from grandchallenge.algorithms.filters import (
     AlgorithmFilter,
     EndpointFilter,
-    EndpointObjectPermissionsFilter,
     InvocationFilter,
+    InvocationObjectPermissionsFilter,
     JobViewsetFilter,
 )
 from grandchallenge.algorithms.forms import (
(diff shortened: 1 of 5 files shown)


app/tests/algorithms_tests/test_api.py
```diff
@@ -806,3 +814,291 @@ def test_delete_not_allowed(self, client):
...
+                "inputs": [
+                    {"interface": ci_string.slug, "value": "foo"},
+                ],
+            },
+        )
+
+        assert (
+            response.status_code == status.HTTP_400_BAD_REQUEST
+        ), response.data
+        assert response.data["endpoint"][0].code == "does_not_exist"
```
koopmant, line 1063: I think an explanation that the endpoint exists, but has the wrong state would be very nice. But this is how we generally do it I think?

app/grandchallenge/algorithms/views.py
```diff
@@ -914,6 +922,21 @@ class EndpointViewSet(ReadOnlyModelViewSet):
     filterset_class = EndpointFilter
 
 
+class InvocationViewSet(
+    CreateModelMixin, RetrieveModelMixin, ListModelMixin, GenericViewSet
+):
+    queryset = Invocation.objects.all()
+    permission_classes = [DjangoModelPermissions]
+    filter_backends = [DjangoFilterBackend, InvocationObjectPermissionsFilter]
```
koopmant: It took me forever to understand I needed to use DjangoModelPermissions here. We use that because we don't have object permissions for invocations, the permissions are handled via the endpoints. See the `InvocationObjectPermissionsFilter`. 
We still check for an `add_invocation` permission on the model level.

jmsmkn: Indeed, that permissions structure is confusing and prone to bugs. If you have followed what is being done for jobs and covered this by tests it should be safe.

koopmant: All right, this should be good then. I think the tests are quite thorough.

amickan: Can you explain why it works to use `DjangoObjectPermissions` for the `JobViewSet` and doesn't work here? In both cases, the permissions go via parent objects (algorithm vs. endpoint), right?

app/tests/algorithms_tests/test_api.py
amickan: This is indeed how we usually do it. I see what you mean though. We could instead add a check to the validate method on the serializer and raise a more descriptive error message. A little less elegant, but it would still serve the same purpose.

app/grandchallenge/algorithms/views.py
koopmant: Err, not really. I just found out that it does work here; I'm not sure how I got to the conclusion that `DjangoObjectPermissions` does not work. It's true that we don't have object permissions for invocations, but it's irrelevant. I think I got confused because the `perms_map` in `DjangoObjectPermissions` has the `add_...` permission for POST requests, but there is no object to check in a post request. So I'm not sure why that's there at all. 
```python
    perms_map = {
        'GET': [],
        'OPTIONS': [],
        'HEAD': [],
        'POST': ['%(app_label)s.add_%(model_name)s'],
        'PUT': ['%(app_label)s.change_%(model_name)s'],
        'PATCH': ['%(app_label)s.change_%(model_name)s'],
        'DELETE': ['%(app_label)s.delete_%(model_name)s'],
    }
```

As you can see there is no required object permisssion to view the object; so that works fine as well. 

I'll change it back to `DjangoObjectPermissions`.

app/tests/algorithms_tests/test_api.py
koopmant: I'll keep it as is then.

app/grandchallenge/algorithms/views.py
amickan: Ok, that's what I thought. Thanks for double checking. I also don't understand why the perms map includes a permission for post requests. The DjangoModelPermissions already checks the global permission, and as you said there is no object to check the object permission against for post requests... oh well, your tests cover all cases well, so I think this is good now.

amickan (approved):
LGTM!

PR #4739, opened 2026-05-27
Rename is_launchable to has_budget

amickan:
In preparation for the changes for https://github.com/DIAGNijmegen/rse-roadmap/issues/469, this renames the `is_launchable` property on reader studies to `has_budget`. That describes the property better. With the changes in this pitch, reader studies will still be launchable for editors even when they are out of budget.


jmsmkn (approved):
LGTM!

PR #4741, opened 2026-05-27
Only preload interactive algorithms for reader studies with budget

amickan:
Only preload interactive algorithms of reader studies with budget. 

Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/469

### DIFF
--- a/app/grandchallenge/components/tasks.py
+++ b/app/grandchallenge/components/tasks.py
@@ -1312,9 +1312,12 @@ def preload_interactive_algorithms():
 
     reader_studies_out_of_budget = [
         rs.pk
-        for rs in ReaderStudy.objects.exclude(max_credits__isnull=True).only(
-            "pk", "max_credits"
+        for rs in ReaderStudy.objects.exclude(max_credits__isnull=True)
+        .prefetch_related(
+            "session_utilizations__reader_studies",
+            "endpoint_utilizations__reader_studies",
         )
+        .only("pk", "max_credits")
         if not rs.has_budget
     ]
 
--- a/app/grandchallenge/components/tasks.py
+++ b/app/grandchallenge/components/tasks.py
@@ -1307,8 +1307,16 @@ class InteractiveAlgorithmLambda:
 @acks_late_micro_short_task
 @transaction.atomic
 def preload_interactive_algorithms():
-    from grandchallenge.reader_studies.models import Question
+    from grandchallenge.reader_studies.models import Question, ReaderStudy
     from grandchallenge.workstations.models import Session
+
+    reader_studies_out_of_budget = [
+        rs.pk
+        for rs in ReaderStudy.objects.exclude(max_credits__isnull=True).only(
+            "pk", "max_credits"
+        )
+        if not rs.has_budget
+    ]
 
     active_interactive_algorithms = (
         Question.objects.filter(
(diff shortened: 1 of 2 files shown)


app/grandchallenge/components/tasks.py
```diff
@@ -1307,9 +1307,17 @@ def delete_stale_provisioned_concurrency_configs(self):
...
     from grandchallenge.workstations.models import Session
 
+    reader_studies_out_of_budget = [
+        rs.pk
+        for rs in ReaderStudy.objects.exclude(max_credits__isnull=True).only(
+            "pk", "max_credits"
+        )
+        if not rs.has_budget
+    ]
+
```
amickan: This query has an N+1 problem (due to expensive queries in `credits_consumed`). It's an asynchronous task, but still. Currently looking at annotating `credits_consumed` in the DB to make this more efficient.

amickan: Struggling with turning `credits_consumed` into an annotation. It involves nested aggregates, and gets quite complex. Maybe prefetching the related objects is good enough here? In general, I would like to know how bad the performance is before I continue trying to optimize this. 

@jmsmkn could you compare timings for the following 2 queries: 

```python
import time
from grand_challenge.reader_studies.models import ReaderStudy


def query_without_prefetch():
    return [
        rs.pk
        for rs in ReaderStudy.objects.exclude(max_credits__isnull=True).only(
            "pk", "max_credits"
        )
        if not rs.has_budget
    ]


def query_with_prefetch():
    return [
        rs.pk
        for rs in ReaderStudy.objects.exclude(max_credits__isnull=True)
        .prefetch_related(
            "session_utilizations__reader_studies",
            "endpoint_utilizations__reader_studies",
        )
        .only("pk", "max_credits")
        if not rs.has_budget
    ]


def benchmark(label, func):
    start = time.perf_counter()
    result = func()
    end = time.perf_counter()

    return {
        "label": label,
        "time_seconds": end - start,
    }


def main():
    results = [
        benchmark("without_prefetch", query_without_prefetch),
        benchmark("with_prefetch", query_with_prefetch),
    ]

    for r in results:
        print(f"{r['label']}:")
        print(f"  time: {r['time_seconds']:.6f}s")


if __name__ == "__main__":
    main()
```

jmsmkn:
```
without_prefetch:
  time: 0.294184s
with_prefetch:
  time: 0.203041s
```

amickan: Thanks. Those timings seem fine to me for an asynchronous task and not worth the added complexity of moving the consumed credits into a DB annotation.

app/grandchallenge/components/tasks.py
```diff
@@ -1318,6 +1326,7 @@ def preload_interactive_algorithms():
                 Session.RUNNING,
             ],
         )
+        .exclude(reader_study__pk__in=reader_studies_out_of_budget)
```
jmsmkn: Is there a reason to use the negative here? Filtering is going to be more efficient than excluding.

app/grandchallenge/components/tasks.py
jmsmkn: Would still be worth including the prefetching if those objects are going to be used.

app/grandchallenge/components/tasks.py
amickan: No. This seemed easier to read to me and I didn't know that filtering is more efficient than excluding. I'll change it to filter().

app/grandchallenge/components/tasks.py
amickan: Very true, I had that as a local commit, but forgot to push!

jmsmkn (approved):
LGTM!

PR #4742, opened 2026-05-28
Remove and migrate more celery tasks

jmsmkn:
These have now been migrated to `lambda_tasks` and the periodic tasks set up.

See #4408

### DIFF
--- a/app/config/settings.py
+++ b/app/config/settings.py
@@ -1229,10 +1229,6 @@ CELERY_BEAT_SCHEDULE = {
         "task": "grandchallenge.algorithms.tasks.update_associated_challenges",
         "schedule": crontab(hour=3, minute=0),
     },
-    "send_new_unread_direct_messages_emails": {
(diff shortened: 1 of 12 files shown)


amickan (approved):
LGTM! 

I have a general question though: how do you decide which tasks to migrate to lambda tasks when / in what order? And how does the period task set-up work these days (if we add new periodic tasks, should these use the lambda decorator and how do we define when they should run)?

jmsmkn:
> how do you decide which tasks to migrate to lambda tasks when / in what order?

I'm slowly going through and seeing what adds new features to test - this time it's calling new functions and mixed signatures. Trying to do things that will migrate entire apps at a time. Also avoiding anything people might be working on right now (invoices, challenges, evaluations, algorithms, reader studies).

> And how does the period task set-up work these days

It's done using event bridge scheduler. See the latest commit in the cloud infrastructure repo.

PR #4745, opened 2026-05-29
Add endpoint status events task

koopmant:
Set endpoints to `STARTED` when it is being created and add a task to receive endpoint state change events, setting the endpoint to `RUNNING` when it is in service. 


koopmant:
We could choose to change the status choices to `CREATING` and `IN_SERVICE` to align with the aws states, instead of `STARTED` and `RUNNING`; but I've opted to align with worksession status choices.

jmsmkn (approved):
LGTM!

PR #4746, opened 2026-06-01
Link utilization and challenge invoices - existing utilizations

chrisvanrun:
Alternative to: https://github.com/DIAGNijmegen/rse-grand-challenge/pull/4718

Part of pitch:

Following a discussion, we've opted to ignore expired invoice balances (even if they are overcharged): as long as there is a non-expired Invoice with available compute costs it should be charged.
 
This throws several wrenches in the logic across the earlier PR. To keep things a bit cleaner, I've opted to split the earlier assigning invoices during regular submissions from the rest. It also includes several renames that were requested in the original PR.

This change set now only handles existing utilizations, including a command to migrate existing utilizations. The previous PR had a command that pushed everything to the first invoice. It now does smart routing, 'spreading' the costs over the existing invoices.

I envision that, if we need to reroute utilizations in the future: we'll unset all invoices for utilizations related to that challenge and run the command. As such, I've opted to go the extra mile and take expiration date into account, albeit we have 0 expired invoices in production at the moment.

In addition, this change set adds a column to the challenge cost overview. This serves as a check in production to see if the migration ran correctly. All values should be close the current (challenge-level) available compute costs, but not match perfect since we might have some surcharges on earlier invoices, which no longer count towards the overall available compute for a challenge. 

Otherwise, this change set should have no influence on running processes.  The linking of new utilizations will happen via a follow-up PR.


app/grandchallenge/utilization/management/commands/route_challenge_utilizations_to_invoices.py
```diff
@@ -0,0 +1,168 @@
...
+
+class NoAuthorizedInvoiceError(Exception):
+    pass
+
+
+class ChallengeInvoiceRouter:
+    def __init__(self):
+        # Final update of compute costs to ensure we have the latest information on the invoices
+        # before we start routing utilizations to them.
+        update_challenge_compute_costs()
```
amickan: Could two of these tasks then theoretically run in parallel? 

I don't think this is necessary to be honest since we'll run this command multiple times anyway, until all utlizations have been linked.

app/grandchallenge/utilization/management/commands/route_challenge_utilizations_to_invoices.py
```diff
@@ -0,0 +1,168 @@
...
+                    batch_size=UPDATE_BATCH_SIZE,
+                )
+                objects_to_update = []
+                self.stdout.write(
+                    self.style.SUCCESS(
+                        f"Progress: {updated} {model.__name__} updated."
+                    )
+                )
+
+            queryset = model.objects.filter(
```
chrisvanrun, line 99: This a deplorable performant query ... I am unsure how to optimize this without doing a per-row scan on the Python side. Maybe we need an index for the combine triplet? AFAIK, that would only be useful for this command and can't readily be used for over-time display of invoices: https://github.com/DIAGNijmegen/rse-roadmap/issues/465#issuecomment-4441711039

 It only needs to run thrice for this command, so perhaps it is OK?

app/grandchallenge/utilization/management/commands/route_challenge_utilizations_to_invoices.py
chrisvanrun: My rational for adding this is workflow as follows.

You run this command, it finishes, and you collect the challenges that still need an invoice. You add the needed invoice and re-run the command. Ups! The `invoice.compute_cost_euro_millicents` haven't been updated from your previous run, any newly added utilizations for earlier handled challenges will be incorrectly assigned.

One option would be to flush the only-in-Python `Invoice.compute_cost_euro_millicents` at the end of the command to the database (via an incremental update). In essence, that would be a proxy for the actual `update_challenge_compute_costs`. This is quite efficient, since it'll take more and more time to run `update_challenge_compute_costs` in the future and the command should maybe not be coupled to it so tightly.

chrisvanrun:
afbeelding

I've cancelled them for now.

app/grandchallenge/utilization/management/commands/route_challenge_utilizations_to_invoices.py
```diff
@@ -0,0 +1,168 @@
...
+
+            updated = 0
+            objects_to_update = []
+
+            def push_bulk_update():
+                nonlocal updated
+                nonlocal objects_to_update
+                updated += model.objects.bulk_update(
+                    objs=objects_to_update,
+                    fields=["invoice"],
```
amickan: Does this also take care of updating `compute_cost_euro_millicents` on the invoice?

app/grandchallenge/utilization/management/commands/route_challenge_utilizations_to_invoices.py
amickan: The per-row scan would be for free if you prefetch the challenge. On the downside, not filtering on challenge means you'd read unnecessary objects into Python memory, but I think that is preferred (and fine since you're batching) over adding an index that will never be used outside of the management command.

amickan: But I don't know how bad the query is in the first place. Maybe James can profile it in production on the largest utilization table.

app/grandchallenge/utilization/management/commands/route_challenge_utilizations_to_invoices.py
chrisvanrun: On the database level? No. We're keeping a shadow bookkeeping on the Python side where it is tallied. However, I think we might want to flush things to the database as to not rely on the large consolidation task. Since we are doing that we should also get the latest on the invoices at that time. I'll whip something together.

app/grandchallenge/utilization/management/commands/route_challenge_utilizations_to_invoices.py
amickan: Maybe I'm missing something, but why are you setting `invoice.compute_cost_euro_millicents` if you're not pushing it to the DB?

app/grandchallenge/utilization/management/commands/route_challenge_utilizations_to_invoices.py
```diff
@@ -0,0 +1,201 @@
...
+
+ITER_BATCH_SIZE = 10_000
+UPDATE_BATCH_SIZE = 1_000
+
+
+class NoAuthorizedInvoiceError(Exception):
+    pass
+
+
+class ChallengeInvoiceRouter:
```
amickan: I honestly found the command a bit hard to read / review because of this set-up - I had to go back and forth when reading this to understand what's happening. Since this is just a one-off command (or well, run a handful of times), I would have preferred to just have all the logic in one place.

amickan: Same goes for the `bulk_update` function.

chrisvanrun: Thanks for your honesty. My initial idea was to have a single entry point: `route` and to split the whole iter/bulk_update loop from the control logic.

I wonder how it would be if you combine it all. TBH, I think it would be total mess. The `bulk_update` innards can be repeated, I don't think that is too bad.

amickan: I think the main problem I had with it are the hidden side-effects from `route()` - it mutates both the invoice and the utilization, but neither of that is obvious when you just look at what happens in handle. I guess I would have expected this function to return the mutated object(s) at the very least.

app/grandchallenge/utilization/management/commands/route_challenge_utilizations_to_invoices.py
chrisvanrun: You need to keep a running tally to know when you have exhausted an invoice's compute costs when routing utilizations to it. You can't update the DB on every route since you'd hammer the DB. Then it becomes the question: do you NEED to flush (i.e. push the invoice updates) at one point?

I thought not: using the heavy consolidation task to update it in between command calls. However, I now think it's likely best to get an invoice update when pushing a new batch of updated utilizations. There might be some race condition there.... need to think about that. We might need to lock things.

app/grandchallenge/utilization/management/commands/route_challenge_utilizations_to_invoices.py
amickan: And I didn't mean put everything inline. The methods on the class could be plain functions called from the handle method. I just think that there is no need for an abstraction layer here, and the logic and data flow would be more visible and easier to follow without it.

amickan: Inlining the bulk-update is a nice change though, thanks!

app/grandchallenge/utilization/management/commands/route_challenge_utilizations_to_invoices.py
amickan: Seems efficient to me to update the invoices when pushing a batch of utilizations. I think the approach with using the increment works even if there have been updates in between. Not sure about concurrent writes, but I guess those are pretty unlikely. The only other process that updates `invoice.compute_cost_euro_millicents` is the periodic task, right? We have full control over when we run this command, so we can make sure to run it inbetween runs of the periodic task. It's a bit fragile, but how often are we going to run this command anyway? Hopefully only a handful of times, right?

chrisvanrun: Yeah, the periodic update might write in the meanwhile. But if the periodic update runs right AFTER we pushed the incremental update but haven't refreshed the Invoices: we end-up with a too low of something. Can we lock the table in between?

Only a handful of times, but I can sort of imagine it'll be useful if we have a totally messed up invoicing for a challenge and need to re-arrange things. It shouldn't (TM) happen because paid invoices are paid invoices...

amickan: > But if the periodic update runs right AFTER we pushed the incremental update but haven't refreshed the Invoices:

I don't follow. What do you mean with refreshing after pushing? The period task pulls the latest state of the DB when it runs, and if the update finished, that state will contain the updated values.

chrisvanrun: > I don't follow. What do you mean with refreshing after pushing? The period task pulls the latest state of the DB when it runs, and if the update finished, that state will contain the updated values.

There is room between the flush from this command, and when it receives a refreshed set of invoices. In between that time, it is possible the flushed compute_costs get overwritten. Maybe a transaction fixes it?

chrisvanrun: Yes. I believe a transaction is exactly something that can fix this. Now added and hence it should remove any fragility on this.

app/grandchallenge/utilization/management/commands/route_challenge_utilizations_to_invoices.py
chrisvanrun: I've rewritten things using plain functions. I tend to rely on OOP a lot, so I hope this helps reading it!

amickan: Much easier to read, thanks!

app/grandchallenge/utilization/management/commands/route_challenge_utilizations_to_invoices.py
chrisvanrun: Since `Invoice` are not linkable yet, we can't run this on production. We could add it via different PR, but I don't see much reason to not just give this command a go on production.

app/grandchallenge/utilization/management/commands/route_challenge_utilizations_to_invoices.py
```diff
@@ -0,0 +1,183 @@
...
+        for model in CHALLENGE_UTILIZATION_MODELS:
+            self.stdout.write(
+                self.style.SUCCESS(
+                    f"Progress: started working on {model.__name__}"
+                )
+            )
+
+            updated = 0
+            objects_to_update = []
+            invoices_map = get_invoice_map()
```
koopmant: We only need to get this once, not for every model right? We're updating it after every transaction anyway. 
It doesn't really matter, just checking if I understand.

app/grandchallenge/utilization/management/commands/route_challenge_utilizations_to_invoices.py
```diff
@@ -0,0 +1,183 @@
...
+
+        for model in CHALLENGE_UTILIZATION_MODELS:
+            self.stdout.write(
+                self.style.SUCCESS(
+                    f"Progress: started working on {model.__name__}"
+                )
+            )
+
+            updated = 0
+            objects_to_update = []
```
koopmant, line 97: doing this again below

app/tests/challenges_tests/test_models.py
```diff
@@ -802,3 +804,77 @@ def test_challenge_queryset_with_user_roles_multiple_challenges():
...
+        payment_status=Invoice.PaymentStatusChoices.PAID,
+        payment_type=Invoice.PaymentTypeChoices.COMPLIMENTARY,
+        compute_costs_euros=2,
+    )
+
+    challenge = (
+        Challenge.objects.with_invoices_with_budget_authorization().get(
+            pk=challenge.pk
+        )
+    )
```
koopmant: could this be too early in the test? (prefetching before the other invoices are created)

koopmant (commented):
Looks good! Some nitpicks only

app/tests/challenges_tests/test_models.py
chrisvanrun: You are right, the refresh is a reset of some cached properties. So this should come later.

app/grandchallenge/utilization/management/commands/route_challenge_utilizations_to_invoices.py
chrisvanrun: This makes me realized that in between model loops, the invoice_map wasn't updated correctly if the flush happened outside the batch-loop. I've made it now such that it is done ONCE before the loop starts, and is correctly updated whenever we push batched data to the server. This was incorrect previously because it wasn't within a transaction which would open up the process to a race condition of other updating tasks.

app/grandchallenge/utilization/management/commands/route_challenge_utilizations_to_invoices.py
chrisvanrun: It did, which was superfluous. It is gone now!

koopmant (approved):
LGTM

PR #4747, opened 2026-06-01
Update email app to use transactions and lambda tasks

jmsmkn:
The email app is modified to use transactions. The tasks are updated so that:

- `send_bulk_email` will schedule 1 email at a time, iterating over the pages of recipients
- `send_raw_emails` will schedule batches of individual emails to send, limited by the SES limits or `EMAILS_MAX_SENT_PER_MINUTE` setting
- `send_raw_email` will send an individual email

A status field is added to `Email` and `RawEmail` for tracking, and locks are now used to ensure only one thing is acting on the object at a time.


app/grandchallenge/background_tasks/tasks.py
```diff
@@ -12,15 +12,15 @@
...
-from lambda_tasks.timeouts import SoftTimeLimitExceeded
+from redis.exceptions import LockError
 
 from grandchallenge.background_tasks.models import CeleryTaskDailyStats
 
 
 @lambda_task(
     singleton=True,
-    # No need to retry here as the periodic task call this again
-    ignore_errors=(SoftTimeLimitExceeded,),
```
amickan, line 23: Why do we no longer ignore soft time limit errors?

app/grandchallenge/emails/tasks.py
```diff
@@ -78,106 +77,141 @@ def get_receivers(action):
...
 def send_raw_emails():
-    if settings.DEBUG:
-        client = None
-        min_send_duration = timedelta(seconds=1)
-    else:
-        client = boto3.client("ses", region_name=settings.AWS_SES_REGION_NAME)
-        min_send_duration = timedelta(
-            seconds=1 / int(client.get_send_quota()["MaxSendRate"])
-        )
+    max_emails_per_minute = get_max_emails_per_minute()
```
amickan, line 155: Does this task still get scheduled every 30 seconds? With the limit per minute here, wouldn't it make more sense to only schedule it once every minute?

amickan: Ah not with the `emails_in_flight` being taken into account. Had not seen that yet.

app/grandchallenge/background_tasks/tasks.py
jmsmkn: As the task is now wrapped in a transaction it will mean that the changes are not committed, so we need to know when this task is exceeding the time limit.

app/grandchallenge/emails/tasks.py
jmsmkn: The minimum rate for the scheduler is once per minute, unfortunately cannot schedule any faster than that without doing internal calls.

app/grandchallenge/background_tasks/tasks.py
amickan: Ah, good to know. All lambda tasks are automatically wrapped in a transaction, nice!

amickan (approved):
LGTM

Issue #4734, opened 2026-05-26
GC test runners flaky


chrisvanrun:
A few times, we've seen boto client error out on the GC's test runners. The test run for 2h26m: might be a red herring, and it is actually a problem with the local S3 client.

> botocore.exceptions.ClientError: An error occurred (InternalError) when calling the PutObject operation (reached max retries: 4): We encountered an internal error, please try again.

For instance on this test run: https://github.com/DIAGNijmegen/rse-grand-challenge/actions/runs/26291766767/job/77393600337?pr=4718

Currently scrolling through the log to see if anything catches my eye.

chrisvanrun:
Function-scoped fixture `algorithm_with_multiple_inputs` which seems to be giving some trouble. Possibly a rate limit or connection limit on the client.

chrisvanrun:
Another example of a failed run: https://github.com/DIAGNijmegen/rse-grand-challenge/actions/runs/26291736493/job/77393490962?pr=4729#step:6:40

jmsmkn:
Note that there is currently an outage of GitHub actions. Those errors happened a few days ago but the availability of GHA has been very poor recently, could be related.

chrisvanrun:
I've seen it last week as well. Maybe there have been problems, and it is possible this was caused by something blocking an execution thread... so system load or simply failing schedulers might be to ultimately blame. I haven't really found anything in the seaweedFS docs that we could set that would mediate it. All timeouts are very reasonable, and it is supposed to easily handle many, many more connections and processes.

jmsmkn:
This is happening still

chrisvanrun:
The frequency of this is getting annoying, and we're burning through hours: the 30-minute timeout is sensible. 

The error message is very generic, and just suggests an internal error on SeaweedFS. 

Approximating the setup on the runner (randomy-seed, and number of workers, ~14 min) I can't get to replicate this. 

The `weed mini` we are using for seaweed is throwing everything on one process:

> This command starts all components in one process (master, volume, filer,
S3 gateway, WebDAV gateway, and Admin UI).

It is odd that it is always the same 124 tests failing, and it is always against 'grand-challenge-protected'. The latter could be just because we mostly use this for things that are typically tested (AlgorithmImages, Image file etc.).

chrisvanrun:
Most likely the volumes are running out of space. (Volumes then become read-only, which would explain why the PutObject action is failing.)

jmsmkn:
[...]

PR #4752, opened 2026-06-02
Add `df` calls after tests for debugging

chrisvanrun:
Will help to figure out why:
- #4734 

Now I am rooting this PR's test run is going to bug out.


jmsmkn (changes requested):
This step will not run if the tests fail so won't provide any insights. The seeds should not be hardcoded. The logs for `s3.localhost-1` are what is needed. Given that actions runners have ~20GB free, and the tests are creating a few mb of data at most this is highly unlikely to be due to space.

PR #4753, opened 2026-06-02
Use generated fields and annotations for reader study budgets

jmsmkn:
See https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/802


koopmant (commented):
LGTM

PR #4755, opened 2026-06-02
Link utilization and challenge invoices - new utilizations

chrisvanrun:
Related to pitch:

Largely a re-do of: https://github.com/DIAGNijmegen/rse-roadmap/issues/475

With the exception that the base is way different and that we ignore any negative balances on other invoices in the challenge.

In short: this change set adds an `Challenge.active_invoice` which is used via `Submission.create_evaluation` to have it assigned to the `EvaluationUtilization` and subsequently all related jobs.

Note: this involves quite some tests suddenly needing a valid invoice.

### DIFF
--- a/app/grandchallenge/evaluation/forms.py
+++ b/app/grandchallenge/evaluation/forms.py
@@ -24,7 +24,7 @@ from django.utils.text import format_lazy
 
 from grandchallenge.algorithms.forms import UserAlgorithmsForPhaseMixin
 from grandchallenge.algorithms.models import Job
-from grandchallenge.challenges.models import Challenge
+from grandchallenge.challenges.exceptions import InsufficientBudgetError
 from grandchallenge.components.forms import (
     AdditionalInputsMixin,
     ContainerImageForm,
@@ -471,6 +471,15 @@ class SubmissionForm(
                 "You must confirm that you want to submit to this phase."
             )
 
+        try:
+            invoice = self._phase.challenge.active_invoice
+        except InsufficientBudgetError:
+            raise ValidationError(
+                "Challenge has insufficient budget. Please contact the challenge organizers."
+            )
+        else:
+            cleaned_data["invoice"] = invoice
+
         return cleaned_data
 
     def clean_phase(self):
@@ -655,7 +664,8 @@ class SubmissionForm(
         instance = super().save(*args, **kwargs)
 
         instance.create_evaluation(
-            additional_inputs=self.cleaned_data["additional_inputs"]
+            additional_inputs=self.cleaned_data["additional_inputs"],
+            invoice=self.cleaned_data["invoice"],
         )
 
         return instance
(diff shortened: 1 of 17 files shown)


app/tests/invoices_tests/factories.py
```diff
@@ -30,5 +30,19 @@ class InvoiceFactory(factory.django.DjangoModelFactory):
         )
     )
 
+    @classmethod
```
koopmant: Why not just make this the default for the factory?

koopmant (dismissed):
LGTM, just a question

app/tests/invoices_tests/factories.py
chrisvanrun: Actually... I don't remember. There WAS a reason. I think the reason was likely it would blow up the already sizeable change set even further, so I was pushing things left and right to keep changes to a minimum where possible. I think having a valid invoice by default is a good thing. I'll update things.

chrisvanrun:
Fyi. One of the push/pull CI Django test runs failed due to: 

- #4734 

I think: the test-failing pattern matches and the timeout got hit. The other with an actual error which was just fixed.

chrisvanrun:
Given https://github.com/DIAGNijmegen/rse-roadmap/issues/475#issuecomment-4610622909 , this is now ready to be merged. 

Needs a final look by @koopmant.

koopmant (approved):
LGTM!

app/grandchallenge/evaluation/forms.py
```diff
@@ -471,6 +471,15 @@ def clean(self):
                 "You must confirm that you want to submit to this phase."
             )
 
+        try:
+            invoice = self._phase.challenge.active_invoice
+        except InsufficientBudgetError:
+            raise ValidationError(
+                "Challenge has insufficient budget. Please contact the challenge organizers."
```
jmsmkn, line 478: Shouldn't this already be covered by `open_for_submissions`?

chrisvanrun: This had me dive a bit into the gating of submissions.

As far as I can tell, the `open_for_submission` is currently only used to gate `Algorithm` **creation** for phases via the ill-named `UserCanSubmitAlgorithmToPhaseMixin`. It also toggles a few warnings and information alerts in view templates.

However, it is currently not used to gate `Submission` creation. The budget-check responsibility is carried by a `challenge.available_compute_euro_millicents` check in `SubmissionForm.clean_creator`. Which, honestly, I had missed for this change set. 

If the budget runs out: there will now be two errors shown on the submission create view: `"You cannot create a new submission at this time"` + `"Challenge has insufficient budget. Please contact the challenge organizers."`. 

My planned next and final step for the pitch is to go through all the usages of `challenge.available_compute_euro_millicents` and replace it by the computed property via the available invoice (i.e. with >0 available budget).  At that step, I want to double-check if the usage is still correct as the new meaning is subtly different. Followed by a final remove of the cached value field, annotations, task logic, et cetera.

I suspect the `SubmissionForm.clean_creator` check should be replaced by some application of `Phase.open_for_submission`. Which I'll employ in the final change set.

Is that OK? Or should I stack things atop of this PR?

jmsmkn: Don't we currently keep submissions open if the budget is exhausted? That way we gather the submissions but don't perform the evaluations. I remember this being a problem on deadline days when challenges would run out of budget.

chrisvanrun: > Don't we currently keep submissions open if the budget is exhausted? 

I don't think so, `open_for_submissions` checks for budget, this prevents `Algorithm` creation.

https://github.com/DIAGNijmegen/rse-grand-challenge/blob/8cfda1d92c5df95ced860d4900cb9a96dbc0d436/app/grandchallenge/evaluation/models.py#L1107-L1114


And the current submission-creation form should give an error if there is insufficient budget (note: `raise_submission_limit_error` raises).

https://github.com/DIAGNijmegen/rse-grand-challenge/blob/8cfda1d92c5df95ced860d4900cb9a96dbc0d436/app/grandchallenge/evaluation/forms.py#L559-L605

Keeping them in a queue would maybe make sense in the sketched `EvaluationRequest` implementation, but as far as I can tell (and test), the current forms are invalid if POSTed when there is no budget.

amickan: Chris is right. We don't queue submissions when there is no budget.

jmsmkn: Okay - thought we did, the behaviour should not change so that is fine.

chrisvanrun: Can this PR then be merged?

jmsmkn: This doesn't change the behaviour, so yes.

PR #4757, opened 2026-06-03
Enable launching rs without budget for editors

amickan:
This enables launching out-of-budget reader studies for editors. Saving answers is disabled. No endpoints are created and no interactive algorithms are preloaded either. 

Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/469


app/grandchallenge/reader_studies/serializers.py
```diff
@@ -296,6 +296,14 @@ def validate(self, attrs):
             display_set = attrs.get("display_set")
             creator = self.context.get("request").user
 
+        reader_study = ReaderStudy.objects.with_has_budget().get(
```
jmsmkn: Rather than double fetching the reader study you could update the question queryset for this view to fetch the reader study along with the budget using `Prefetch`, for instance: https://github.com/DIAGNijmegen/grand-challenge/blob/8b56ac0f5ff75f75a8a00d80c3d9a7b6fd439956/app/grandchallenge/challenges/models.py#L156-L162

amickan: I had tried that but it didn't work because querysets defined on the viewset are only used on GET calls. I think what I could do instead is override the question field's queryset on the serializer. I'll try that.

amickan: Hm, that works for POST requests, but still fails for PATCH requests because the question is taken from the instance there. So then I'll need to update both the queryset on the viewset and on the serializer field.

jmsmkn (approved):
LGTM!

app/grandchallenge/workstations/forms.py
```diff
@@ -57,18 +57,22 @@ class SessionForm(ModelForm):
...
 
     def clean(self):
-        if self.__reader_study and not self.__reader_study.has_budget:
+        if self.__reader_study and (
+            not self.__reader_study.has_budget
+            and not self.__reader_study.is_editor(self.__user)
+        ):
             raise ValidationError(
                 "Reader study is out of budget and cannot be launched."
             )
```
pkcakeout, line 78: [There was a post before this one which I deleted claiming that this was a bug of some kind]

A bit more detail: Maybe there is a complimentary check missing for groups that have permissions to launch reader-studies as will (like DIAG and computational pathology etc?)

jmsmkn: Please do not comment on old PRs about issues. If there are new issues please make a new issue and do all of the commenting in one place there. You can link from the issue to lines in this PR.

pkcakeout: Ah crap... and I even added a back-link here which is not super-neat... I will delete my post above.

PR #4758, opened 2026-06-03
Run lambda tasks in background for the development server

jmsmkn:
Users who run without Docker will need to set LAMBDA_TASKS_LOCAL_WORKERS in their environment.

See #4408


amickan (approved):
Thanks! LGTM!

PR #4763, opened 2026-06-04
Migrate upload session linked tasks

jmsmkn:
See #4408

### DIFF
--- a/app/grandchallenge/cases/tasks.py
+++ b/app/grandchallenge/cases/tasks.py
@@ -6,6 +6,7 @@ from pathlib import Path
 from shutil import rmtree
 from subprocess import CalledProcessError
 from tempfile import TemporaryDirectory
+from uuid import UUID
 
 import boto3
 import botocore.exceptions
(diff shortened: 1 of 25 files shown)


app/tests/cases_tests/test_tasks.py
```diff
@@ -41,28 +43,34 @@
...
-        session.process_images(linked_task=local_linked_task.signature())
+        session.process_images(
+            upload_session_complete_task=UploadSessionCompleteTask(
+                task=local_linked_task, kwargs={}
+            )
+        )
 
-    assert called == {"upload_session_pk": session.pk}
+    global LOCAL_LINKED_TASK_CALLED
+    assert LOCAL_LINKED_TASK_CALLED == {"upload_session_pk": session.pk}
```
koopmant:
```suggestion
    assert LOCAL_LINKED_TASK_CALLED == {"upload_session_pk": str(session.pk)}
```

app/grandchallenge/cases/tasks.py
```diff
@@ -112,16 +118,29 @@ def extract_files(*, source_path: Path, checked_paths=None):
...
+def build_images_celery(**kwargs):
+    # TODO: 4408 Remove, this is still here to handle existing tasks on SQS
+    return build_images(**kwargs)
+
+
+@lambda_task(
+    queue=LambdaTaskQueueChoices.MEM8G,
+    retry_on=(LockNotAcquiredException,),
+    soft_timeout=LONG_TASK_SOFT_TIMEOUT,
+    hard_timeout=LONG_TASK_HARD_TIMEOUT,
```
koopmant, line 134: Is this the first long task? And do we intend to use the same timeout for "long" tasks? 
I'm wondering if we shouldn't just set the time limit value here...

koopmant (dismissed):
LGTM. 

Question + a fix in one test.

app/grandchallenge/cases/tasks.py
jmsmkn: Yes - we intend to use the same. The limit for AWS lambda is 15 minutes so just hardcoding that assumption here in case they increase it further.

PR #4764, opened 2026-06-04
Add admin action to update postpaid costs

### LINKED ISSUE #4732: Admin action to update postpaid invoice costs
Add an admin action that [calculates the compute and storage costs based on actual challenge usage](https://github.com/DIAGNijmegen/rse-grand-challenge-admin/pull/779) and updates the post-paid invoice with the proper amounts. 

The admin action should only work for post-paid invoices in initialized state.

amickan:
Closes #4732

### DIFF
--- a/app/grandchallenge/invoices/admin.py
+++ b/app/grandchallenge/invoices/admin.py
@@ -1,6 +1,12 @@
-from django.contrib import admin
-from django.db import models
+import math
 
+from django.contrib import admin, messages
+from django.db import models
+from django.db.models import Q, Sum
+from django.db.models.functions import Coalesce
+from django.template.defaultfilters import pluralize
+
+from grandchallenge.challenges.models import ChallengeRequest
 from grandchallenge.core.admin import (
     GroupObjectPermissionAdmin,
     UserObjectPermissionAdmin,
@@ -49,6 +55,7 @@ class ToCheckFilter(admin.SimpleListFilter):
 @admin.register(Invoice)
 class InvoiceAdmin(admin.ModelAdmin):
     list_display = (
+        "pk",
         "challenge",
         "issued_on",
         "expires_on",
@@ -72,6 +79,7 @@ class InvoiceAdmin(admin.ModelAdmin):
     )
     autocomplete_fields = ("challenge",)
     readonly_fields = ["invoice_request_text"]
+    actions = ["calculate_postpaid_costs"]
 
     def has_delete_permission(self, request, obj=None):
         # invoices cannot be deleted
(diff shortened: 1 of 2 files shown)


app/grandchallenge/invoices/admin.py
```diff
@@ -115,6 +123,112 @@ def invoice_request_text(self, obj):
...
+            unpaid_compute = total_compute
+            unpaid_storage = total_storage - paid_storage
+            unpaid_total = unpaid_compute + unpaid_storage
+
+            if unpaid_total <= 0:
+                unused_budget_postpaid_invoices.append(invoice)
+                continue
+
+            post_paid_total = math.ceil(unpaid_total / 250.0) * 250
+            compute_ratio = total_compute / total_consumed
```
amickan, line 181: Not sure if this is right. `total_consumed` is a mix of all storage costs incurred for the challenge for 5 years and the compute costs charged to this specific invoice (i.e. not all compute costs of the challenge). Maybe the ratios should be calculated based on the unpaid amounts (i.e. `unpaid_compute / unpaid_total` and `unpaid_storage / unpaid_total`)?

jmsmkn:
The task is doing quite a lot. Design wise maybe this would be better as a normal invoice update view for authorised users, so the values can be displayed and sanity checked before clicking save to update the invoice?

app/grandchallenge/invoices/admin.py
```diff
@@ -115,6 +123,112 @@ def invoice_request_text(self, obj):
...
+        storage_costs_euros_per_gb = (
+            ChallengeRequest().storage_costs_euros_per_gb
+        )
+        unused_budget_postpaid_invoices = []
+        updated_invoices = []
+
+        for invoice in valid_invoices:
+            challenge = invoice.challenge
+            total_compute = invoice.compute_cost_euro_millicents / 1000 / 100
+            total_storage = storage_costs_euros_per_gb * (
```
koopmant, line 155: I think we'll want to use `storage_bytes_to_euro_per_year` and `registry_bytes_to_euro_per_year` here.

app/grandchallenge/invoices/admin.py
```diff
@@ -115,6 +123,112 @@ def invoice_request_text(self, obj):
...
+
+        storage_costs_euros_per_gb = (
+            ChallengeRequest().storage_costs_euros_per_gb
+        )
+        unused_budget_postpaid_invoices = []
+        updated_invoices = []
+
+        for invoice in valid_invoices:
+            challenge = invoice.challenge
+            total_compute = invoice.compute_cost_euro_millicents / 1000 / 100
```
koopmant, line 154:
```suggestion
            unpaid_compute = invoice.compute_cost_euro_millicents / 1000 / 100
            total_compute = challenge.compute_cost_euro_millicents / 1000 / 100
```

app/grandchallenge/invoices/admin.py
```diff
@@ -115,6 +123,112 @@ def invoice_request_text(self, obj):
...
+                )
+                .exclude(pk=invoice.pk)
+                .aggregate(
+                    paid_storage=Coalesce(Sum("storage_costs_euros"), 0),
+                )
+            )["paid_storage"]
+
+            unpaid_compute = total_compute
+            unpaid_storage = total_storage - paid_storage
+            unpaid_total = unpaid_compute + unpaid_storage
```
koopmant, line 174: I just realize that with a prepaid invoice that has too little storage budget but still some compute budget left, this calculation will issue a new postpaid invoice just for the storage. We need to account for this as well. So we'll need to calculate a potential new split for an already paid earlier invoice. Maybe this action is not something we should automate after all?

app/grandchallenge/invoices/admin.py
koopmant: This isn't right indeed. We'll need to take the total compute, see above.

koopmant (commented):
I stopped reviewing when I realized we need to take into account that the compute/storage split on an earlier invoice may not be optimal and needs to be adjusted, see my comment below. I think it may be easier if we keep this manual for now after all. 

We could make an update view as James suggested. There we can present all the information we need to make the right decision. But that might be easier to make once we've done it a few times manually ourselves?

amickan:
> I stopped reviewing when I realized we need to take into account that the compute/storage split on an earlier invoice may not be optimal and needs to be adjusted

Good point. In such a scenario you would want to adjust the split on the earlier invoice, and only then do these calculations here. Maybe there are even more cornercases that we haven't thought of yet. Let's not automate this for now, and wait until we have done this a few times and know what kind of a view we'd want.

PR #4768, opened 2026-06-05
Report number updated utilizations in command

chrisvanrun:
Tiny addition that better shows the work done. Will let us know if there were any unlinked utilizations.


amickan (approved):
LGTM

PR #4774, opened 2026-06-08
Update invoice-table, add detail view and make related section in statistics staff-only

chrisvanrun:
Part of pitch:

Updated:

This removes the invoice list from the per-challenge statistics page and makes the previous challenge review information staff-only.

In addition, it adds a reference to a detail page of an invoice. This detail page contains limited information now but will be updated in a follow-up to show-case the consumed budget. The detail paged is accessed via a new `external_pk` field as to prevent enumeration attacks.

Finally, the badge HTML tag was pulled into the model and updated to be more consistent and informative. The due and overdue status are removed: IMHO they are shown sufficiently via the due date warnings UI.

## Showcase

afbeelding

afbeelding


app/grandchallenge/invoices/templates/invoices/invoice_detail.html
```diff
@@ -0,0 +1,60 @@
...
+            {{ object.due_date|default_if_none:"-" }}
+            {% if object.is_overdue %}
+                <i class="fas fa-exclamation-triangle text-danger" title="Invoice is overdue."></i>
+            {% elif object.is_due %}
+                <i class="fas fa-exclamation-triangle text-warning" title="Invoice is due."></i>
+            {% endif %}
+         </dd>
+
+        <dt>Follow-up on</dt>
+        <dd>{{ object.follow_up_on|default_if_none:"-" }}</dd>
```
amickan: I would only show this for post-paid invoices.

app/grandchallenge/invoices/models.py
```diff
@@ -356,6 +357,58 @@ def _current_state(self):
...
+        payment_type = self.payment_type
+        payment_status = self.payment_status
+
+        if self.is_expired:
+            return "Expired"
+        elif (
+            payment_type == PaymentTypeChoices.COMPLIMENTARY
+            and payment_status != PaymentStatusChoices.CANCELLED
+        ):
+            return "Paid"
```
amickan: Unsure if we should have this here. Complimentary invoices should always be paid (unless cancelled). Maybe we should instead add a DB constraint for that. Then you could just fall back to `self.get_payment_status_display()`. One exception less here.

app/grandchallenge/invoices/templates/invoices/invoice_detail.html
chrisvanrun: Agreed. Updated and I have also done the same for the Due date: only show it if the state is ISSUED.

app/grandchallenge/invoices/models.py
chrisvanrun: I did a quick check, and it seems like all existing `Comlimentary` types are always set as PAID. Let's simplify this by adding a DB rule.

amickan: Seperate PR I would say

app/grandchallenge/invoices/views.py
```diff
@@ -34,3 +35,26 @@ def get_queryset(self):
...
+    model = Invoice
+    raise_exception = True
+    login_url = reverse_lazy("account_login")
+
+    def dispatch(self, request, *args, **kwargs):
+        invoice = self.get_object()
+        if request.user.has_perm(
+            "challenges.view_challengerequest"
+        ) or request.user.has_perm("invoices.view_invoice", invoice):
+            return super().dispatch(request, *args, **kwargs)
```
amickan: I thought about this for a while. It struck me as odd to "hide" an object permission check in an AccessMixin like that, but I think this is the way to do it after all given that you only want to check for object-level permissions if the user does not have the global perm.

chrisvanrun: Perhaps we should lean into that odd feeling and update the `"challenges.view_challengerequest"` in general?

Create a group of challenge reviewers and assign `"view_invoice"` to that group?

jmsmkn: There is already a challenge reviewers group, see `settings.CHALLENGES_REVIEWERS_GROUP_NAME`. They should be assigned the `view_invoice` object permission if necessary.

jmsmkn: But either way I do not see why challenge reviewers need to see this page, given their job is to review challenges.

Any administrator that needs to see the invoices should get all of the information they need from the admin, so should not need access to this page.

This then simplifies this view to using object permissions only.

app/grandchallenge/invoices/models.py
chrisvanrun: See: https://github.com/DIAGNijmegen/rse-grand-challenge/pull/4776

app/grandchallenge/invoices/templates/invoices/invoice_list.html
```diff
@@ -38,64 +38,50 @@ <h2>Invoices for {{ challenge.short_name }}</h2>
...
-                    <th>Issued on</th>
-                    <th>Due date</th>
-                    <th>Amount</th>
-                    <th>Type</th>
-                    <th>Status</th>
+                    <td>
+                        <a href="{% url 'invoices:detail' challenge_short_name=invoice.challenge.short_name pk=invoice.pk %}" title="View invoice details"><i class="fas fa-info-circle"></i></a>
+                    </td>
+                    <td>{% firstof invoice.internal_invoice_number '-' %}</td>
+                    <td data-order="{{ invoice.issued_on|date:"c" }}">{{ invoice.issued_on }}</td>
```
amickan: Shouldn't we use `default_if_none` here and above (invoice number) for consistency?

amickan (dismissed):
The invoice table was now removed from the challenge stats page. Was that intentional? I kinda liked having it there for reference. 

Otherwise, this looks good to me. Make sure to update the PR description (you're no longer adding a partial template) ;-)

app/grandchallenge/invoices/models.py
amickan: This can go with the database constraint that you're adding in the other PR, right?

app/grandchallenge/invoices/templates/invoices/invoice_list.html
chrisvanrun: The issued_on would be nicely consistent. Yes. The internal_invoice_number already has it: but since it is a char field...

app/grandchallenge/invoices/models.py
chrisvanrun: Mhhh, I'd rather merge that one first and then update things here.

chrisvanrun:
> The invoice table was now removed from the challenge stats page. Was that intentional? I kinda liked having it there for reference.

Yes, as the detail page is inaccessible. I also limited that section to staff-only. My plan is to update the invoice (Django) admin to show the needed fields.

app/grandchallenge/invoices/urls.py
```diff
@@ -1,9 +1,10 @@
 from django.urls import path
 
-from grandchallenge.invoices.views import InvoiceList
+from grandchallenge.invoices.views import InvoiceDetail, InvoiceList
 
 app_name = "invoices"
 
 urlpatterns = [
     path("all/", InvoiceList.as_view(), name="list"),
+    path("<int:pk>/", InvoiceDetail.as_view(), name="detail"),
```
jmsmkn: We shouldn't expose integer primary keys to prevent enumeration attacks. Please add an `external_pk = models.UUIDField(unique=True, default=uuid.uuid4, editable=False)` along with an index for this field, to invoices and use this to lookup here.

amickan: Ah, good to know! We should add this to our coding guidelines doc. 

I think we also do this for emails and teams.

jmsmkn: Not such a big issue there - emails are behind the global "emails.view_email", and neither contain protected information like invoices do.

chrisvanrun: (And CIVs. I think?)

jmsmkn: Yes CIVs should have it too. Would need to use `default_db` for that for generation but it's possible:

```
migrations.RunSQL(
    sql="UPDATE componentinterfacevalue SET external_pk = gen_random_uuid() WHERE uid IS NULL;",
    reverse_sql=migrations.RunSQL.noop,
),
```

Larger issue would be the breakages in gcapi, all of the external applications and potentially algorithms and evaluation code who are currently working with integers.

chrisvanrun: https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/810

Will fix `Invoice` in via this PR.

app/grandchallenge/invoices/models.py
chrisvanrun: Done.

app/grandchallenge/invoices/urls.py
chrisvanrun: Done. Required a 3-step migration.

app/grandchallenge/invoices/migrations/0020_auto_20260609_0849.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("invoices", "0019_invoice_external_pk_and_more"),
+    ]
+
+    operations = [
+        migrations.RunPython(gen_uuid),
```
jmsmkn: Should have `elidable=True` set as there is nothing to do if no data are present.

app/grandchallenge/invoices/views.py
```diff
@@ -34,3 +34,23 @@ def get_queryset(self):
...
+):
+    model = Invoice
+    permission_required = "view_invoice"
+    raise_exception = True
+    login_url = reverse_lazy("account_login")
+
+    def get_object(self, queryset=None):
+        if queryset is None:
+            queryset = self.get_queryset()
+        return queryset.get(external_pk=self.kwargs["external_pk"])
```
jmsmkn: Rather than doing this I think you can set:

```
slug_field = "external_pk"
slug_url_kwarg = "external_pk"
```

on the view.

jmsmkn:
Couple of minor things, rest looks good.

app/grandchallenge/invoices/migrations/0020_auto_20260609_0849.py
chrisvanrun: Ah yes, forgot about that one. Thanks!

app/grandchallenge/invoices/views.py
chrisvanrun: Ah, yes! I was staring at the `pk` logic, missing that the slug route actually has a settable field name.

jmsmkn (approved):
LGTM!

PR #4775, opened 2026-06-08
Challenge-level budget clean up

chrisvanrun:
Part of pitch:
- https://github.com/DIAGNijmegen/rse-roadmap/issues/475

This change results in:
- Removal of the `Challenge.compute_cost_euro_millicents` field
- Removal of the related logic in periodic consolidation task
- Removal of the annotation `ChallengeQuerySet.with_available_compute`
- Migrations of any final usage of fields calculated via `with_available_compute` to `Invoice` derivates


app/grandchallenge/challenges/models.py
```diff
@@ -476,13 +410,6 @@ class Challenge(ChallengeBase, FieldChangeMixin):
...
     )
 
-    compute_cost_euro_millicents = models.PositiveBigIntegerField(
-        # We store euro here as the costs were incurred at a time when
-        # the exchange rate may have been different
-        editable=False,
-        default=0,
-        help_text="The total compute cost for this challenge in Euro Cents, including Tax",
-    )
     size_in_storage = models.PositiveBigIntegerField(
```
amickan, line 486: The field needs to be deprecated first, and removed in a separate PR, I think

chrisvanrun: What is our go-to construct for deprecation?

chrisvanrun: General warning if it gets accessed. And removal at a follow-up deployment?

Roughly like:

```Python
from warnings import warn

class MyModel(models.Model):
    _old_field = models.CharField(max_length=100, db_column="old_field")

    @property
    def old_field(self):
        warn(
            "old_field is deprecated, use new_field instead.",
            DeprecationWarning,
            stacklevel=2,
        )
        return self._old_field
```

amickan: We use https://pypi.org/project/django-deprecate-fields/

You just wrap the field in `deprecate_field()` and run `makemigrations`

chrisvanrun: Done.

amickan (approved):
LGTM

app/grandchallenge/challenges/models.py
```diff
@@ -476,12 +411,13 @@ class Challenge(ChallengeBase, FieldChangeMixin):
...
-        editable=False,
-        default=0,
-        help_text="The total compute cost for this challenge in Euro Cents, including Tax",
+    compute_cost_euro_millicents = deprecate_field(
+        models.PositiveBigIntegerField(
+            editable=False,
+            default=0,
+            help_text="The total compute cost for this challenge in Euro Cents, including Tax",
+        ),
+        raise_on_access=True,
```
amickan, line 420: Hm, not sure about this. I don't think we have set this to True in the past.

chrisvanrun: I think we want this. Or else we'll have some red-herring with `None` values downstream somewhere.

It is done at one other place:
https://github.com/DIAGNijmegen/rse-grand-challenge/blob/43ca13de5e8d205ee52aeb09ecb167b961ebba39/app/grandchallenge/workstations/models.py#L447

amickan: I think you're right and raising is ok here.

amickan: Actually, wait, isn't the whole point of deprecating it like this to not raise (but just log) an error if an old process tries to access the field?

chrisvanrun: @jmsmkn , given deployment mystics. What is generally wanted for a deprecated field?

jmsmkn: Set `raise_on_access=True` - hopefully all usages are covered in the tests so you'll hear about it there but if you have some uncovered python rows or templates you'll hear about it in prod this way. We generally do not grep the logs for usages.

jmsmkn (approved):
LGTM!

PR #4776, opened 2026-06-08
Add DB constraint that prevents non-paid (or cancelled) complimentary…

chrisvanrun:
… invoices

See: https://github.com/DIAGNijmegen/rse-grand-challenge/pull/4774#discussion_r3372729408

Current existing `Invoice` do not break this constraint. (Double-checked the backend.)


chrisvanrun:
Hah. I focused so hard on the production side that I forgot to run other tests locally.

Edit: should now pass.

amickan (approved):
LGTM! 

Maybe @koopmant or @jmsmkn have thoughts on adding this constraint?

chrisvanrun:
> Maybe @koopmant or @jmsmkn have thoughts on adding this constraint?

To contextualize. This was used sometimes in the tests. IMHO it should not be the case for 'real' invoices.

jmsmkn:
No seems fine to me

koopmant:
Fine with me! Not sure when we'd want a cancelled complimentary one though? It could just expire, right?

chrisvanrun:
> Not sure when we'd want a cancelled complimentary one though?

A valid point. Practically, that will have the same effect. There is a semantic difference between cancelling and expiring, though. I prefer a default way of handling 'ups' invoices: cancelling them. Moving the `expires_on` date feel a bit hacky to use it for that prupose.

PR #4778, opened 2026-06-08
Fix task errors for add_image_to_answer

jmsmkn:
`add_image_to_answer` would raise an error if the answer upload session and the requested upload session that has just finished processing did not match. This happens when a user saves an annotation answer in rapid succession.

This should not be a task error as the task behaved correctly, so now we return a status object with what happened, which can be seen in the Django admin.


amickan (approved):
Makes sense and looks good to me.

PR #4779, opened 2026-06-09
Add Broken Links App

jmsmkn:
New broken_links app that replaces `django.middleware.common.BrokenLinkEmailsMiddleware`. Instead of sending emails on 404s, broken links are saved to the database.

An interactive Kiro session was used to generate these commits. All lines have been reviewed by me as they were generated.

See https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/808


app/grandchallenge/broken_links/models.py
```diff
@@ -0,0 +1,34 @@
...
+    domain = models.CharField(max_length=255)
+    path = models.TextField()
+    referer = models.TextField()
+    user_agent = models.TextField()
+    ip_address = models.GenericIPAddressField(null=True)
+    is_internal = models.BooleanField()
+
+    class Meta(UUIDModel.Meta):
+        indexes = [
+            models.Index(fields=["-created"]),
```
koopmant, line 18: I don't think the index itself has an order, so the "-" is not necessary here.

koopmant (commented):
The code looks good to me, but testing this locally by setting debug to False breaks the site layout as it can no longer find the styling resources. Broken links are then added like `/unknown/CACHE/css/output.b75ef920ac8c.css`. Please confirm if that's the case for you too, it could be an environment issue.

app/grandchallenge/broken_links/models.py
jmsmkn: It only matters in multi-column indices, but keeping it there indicates to future developers that this is how we are querying this column.

jmsmkn:
You would need to run `collectstatic`, and set `STATIC_ROOT=../.static/` (or somewhere writable).

koopmant (dismissed):
LGTM

PR #4780, opened 2026-06-09
Use lambda tasks for `update_challenge_compute_costs`

jmsmkn:
Uses fanout to update the invoices and phases associated only with one challenge at a time, removing the internal retries to use the external scheduler.

See #4408 
See https://github.com/DIAGNijmegen/rse-roadmap/issues/475#issuecomment-4657516393


chrisvanrun (approved):
Looks solid to me!

PR #4782, opened 2026-06-09
Show budget consumption in `Invoice` admin

chrisvanrun:
Part of pitch (sort of):

Since the invoice detail page is meant for challenge organizers, the site admins need to rely on the backend display of information on `Invoice` objects. This includes current usage and status.

This change set re-arranges shown fields in the admin into field sets (plenty of fields to show) and adds these fields:
- `Is active` (i.e. not expired)
- `Consumed budget` (% consumed)
- `Available Compute Cost` (Euros)
- `Approved Compute Cost` (Euros)
- `Consumed Compute Cost` (Euros)
- `Write Off Compute Cost` (Euros)

It also removes some fields from the list view to make it more compact, leaving the most essential (IMO).

## Showcase

<details>
<summary>A few screenshots</summary>

### List View 

afbeelding

### Detail View

(split over three screenshots)

afbeelding

afbeelding

afbeelding

</details>

### DIFF
--- a/app/grandchallenge/invoices/admin.py
+++ b/app/grandchallenge/invoices/admin.py
@@ -47,18 +47,18 @@ class ToCheckFilter(admin.SimpleListFilter):
         return queryset
 
 
-class IsActiveFilter(admin.SimpleListFilter):
-    title = "is active"
-    parameter_name = "is_active"
+class IsExpiredFilter(admin.SimpleListFilter):
+    title = "is expired"
+    parameter_name = "is_expired"
 
     def lookups(self, request, model_admin):
         return [("1", "Yes"), ("0", "No")]
 
     def queryset(self, request, queryset):
         if self.value() == "1":
-            return queryset.exclude(is_expired=True)
+            return queryset.filter(is_expired=True)
         elif self.value() == "0":
-            return queryset.exclude(is_expired=False)
+            return queryset.filter(is_expired=False)
         return queryset
 
 
@@ -74,13 +74,13 @@ class InvoiceAdmin(admin.ModelAdmin):
         "issued_on",
         "expires_on",
         "last_checked_on",
-        "is_active",
+        "is_not_expired",
         "internal_comments",
     )
     list_filter = (
         OverdueListFilter,
         ToCheckFilter,
-        IsActiveFilter,
+        IsExpiredFilter,
         "payment_status",
         "payment_type",
         "challenge__short_name",
@@ -111,7 +111,7 @@ class InvoiceAdmin(admin.ModelAdmin):
                     "follow_up_on",
                     (
                         "expires_on",
-                        "is_active",
+                        "is_not_expired",
                     ),
                 ]
             },
@@ -162,7 +162,7 @@ class InvoiceAdmin(admin.ModelAdmin):
         "consumed_compute_cost",
         "write_off_compute_cost",
         "total_amount_euros",
-        "is_active",
+        "is_not_expired",
     ]
 
     ordering = ["expires_on", "created"]
@@ -179,7 +179,7 @@ class InvoiceAdmin(admin.ModelAdmin):
         return obj.internal_invoice_number
 
     @admin.display(boolean=True)
-    def is_active(self, obj):
+    def is_not_expired(self, obj):
         return not obj.is_expired
 
     def available_compute_cost(self, obj):
--- a/app/grandchallenge/invoices/admin.py
+++ b/app/grandchallenge/invoices/admin.py
@@ -58,6 +58,7 @@ class InvoiceAdmin(admin.ModelAdmin):
         "percent_budget_consumed_display",
         "issued_on",
         "follow_up_on",
+        "is_active",
         "internal_comments",
     )
     list_filter = (
@@ -90,7 +91,10 @@ class InvoiceAdmin(admin.ModelAdmin):
                     "paid_on",
                     "last_checked_on",
                     "follow_up_on",
-                    "expires_on",
+                    (
+                        "expires_on",
+                        "is_active",
+                    ),
                 ]
             },
         ),
@@ -139,6 +143,7 @@ class InvoiceAdmin(admin.ModelAdmin):
         "consumed_compute_cost",
         "write_off_compute_cost",
         "total_amount_euros",
+        "is_active",
     ]
 
     @admin.display(description="Total")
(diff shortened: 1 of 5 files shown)


app/grandchallenge/invoices/admin.py
```diff
@@ -70,8 +68,114 @@ class InvoiceAdmin(admin.ModelAdmin):
...
+    ]
+
+    @admin.display(description="Total")
+    def total_amount_euros(self, obj):
+        return euro(obj.total_amount_euros, decimal_places=0)
+
+    @admin.display(description="Number")
+    def internal_invoice_number_display(self, obj):
+        return obj.internal_invoice_number
+
```
amickan: What is the added benefit of doing this?

app/grandchallenge/invoices/admin.py
```diff
@@ -50,17 +51,14 @@ def queryset(self, request, queryset):
...
-        "follow_up_on",
-        "internal_invoice_number",
-        "internal_client_number",
-        "contact_email",
-        "total_amount_euros",
+        "internal_invoice_number_display",
         "payment_type",
         "payment_status",
-        "paid_on",
-        "last_checked_on",
```
amickan, line 63: I would like to keep this in the list view since I use it to double check that I have updated all invoices at the end of my weekly check.

app/grandchallenge/invoices/admin.py
```diff
@@ -70,8 +68,114 @@ class InvoiceAdmin(admin.ModelAdmin):
...
+    @admin.display(description="Total")
+    def total_amount_euros(self, obj):
+        return euro(obj.total_amount_euros, decimal_places=0)
+
+    @admin.display(description="Number")
+    def internal_invoice_number_display(self, obj):
+        return obj.internal_invoice_number
+
+    @admin.display(boolean=True)
+    def is_active(self, obj):
```
amickan: Might be nice to be able to filter on this on the list view.

amickan (commented):
Looks good, just some minor comments. 

As mentioned [here](https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/809#issuecomment-4668101923), I think it would be useful to add a few more fields (compute/storage ratio for the challenge and total consumed storage and compute costs of the linked challenge), but this can be done in a separate PR as well.

app/grandchallenge/invoices/admin.py
amickan: And with "this" I mean "last_checked_on".

app/grandchallenge/invoices/admin.py
```diff
@@ -70,8 +68,114 @@ class InvoiceAdmin(admin.ModelAdmin):
...
+        "approved_compute_cost",
+        "consumed_compute_cost",
+        "write_off_compute_cost",
+        "total_amount_euros",
+        "is_active",
+    ]
+
+    @admin.display(description="Total")
+    def total_amount_euros(self, obj):
+        return euro(obj.total_amount_euros, decimal_places=0)
```
amickan: I think this will break adding invoices since `total_amount_euros` returns `None` when the costs are not set yet. Should be changed to: 

```suggestion
        if obj.total_amount_euros:
            return euro(obj.total_amount_euros, decimal_places=0)
        else:
            return ''
```

app/grandchallenge/invoices/admin.py
chrisvanrun: Quick fix for column width: 'Internal invoice number' was nearly twice the actual number.

app/grandchallenge/invoices/admin.py
chrisvanrun: Ah, I had fixed that. But reverted it slightly later in a review since I couldn't remember why I had added it in the first place. >.<

chrisvanrun:
> I think it would be useful to add a few more fields (compute/storage ratio for the challenge and total consumed storage and compute costs of the linked challenge), but this can be done in a separate PR as well.

I prefer not to scope creep this too much. This is a first attempt that is bound to be missing some practical things. The first review has already given some steering from a practical view. Let's give this new admin a spin first and see where we land.

chrisvanrun:
(Force pushed a rebased branch as the `is_expired` touches some other code).

amickan:
> > I think it would be useful to add a few more fields (compute/storage ratio for the challenge and total consumed storage and compute costs of the linked challenge), but this can be done in a separate PR as well.
> 
> I prefer not to scope creep this too much. This is a first attempt that is bound to be missing some practical things. The first review has already given some steering from a practical view. Let's give this new admin a spin first and see where we land.

I agree. Additions should be made via a new PR and I was working on one already, but I'll wait until this PR is merged.

app/grandchallenge/invoices/admin.py
```diff
@@ -46,32 +47,159 @@ def queryset(self, request, queryset):
         return queryset
 
 
+class IsExpiredFilter(admin.SimpleListFilter):
```
jmsmkn, line 50: Why is this necessary? You should be able to add `"is_expired"` directly to `list_filter`

app/grandchallenge/invoices/admin.py
```diff
@@ -46,32 +47,159 @@ def queryset(self, request, queryset):
...
+        else:
+            return ""
+
+    @admin.display(description="Number")  # Reduce column width
+    def internal_invoice_number_display(self, obj):
+        return obj.internal_invoice_number
+
+    @admin.display(boolean=True)
+    def is_not_expired(self, obj):
+        return not obj.is_expired
```
amickan, line 183: Why not just use `is_expired`?

app/grandchallenge/invoices/models.py
```diff
@@ -121,6 +121,9 @@ def with_budget_authorization(self):
             )
         )
 
+    def with_is_expired(self):
+        return self.annotate(is_expired=Q(expires_on__lt=Now()))
+
```
amickan, line 126: Out of curiosity: Why did this have to change from being a property to an annotation?

app/grandchallenge/invoices/admin.py
chrisvanrun: AFAIK you can't directly add annotated fields to the `list_filter`. Directly adding it gives a:

`<class 'grandchallenge.invoices.admin.InvoiceAdmin'>: (admin.E116) The value of 'list_filter[2]' refers to 'is_expired', which does not refer to a Field.`

Perhaps if one adds it to one of the other field lists on the class it might work, but I haven't been able to find that.

app/grandchallenge/invoices/models.py
chrisvanrun: AFAIK it is required to be able to filter on it as filtering is an action on the database level.

app/grandchallenge/invoices/admin.py
jmsmkn: Ah okay, then yes this is the way to do it for an annotation.

app/grandchallenge/invoices/admin.py
chrisvanrun: Green is good. I prefer the visual cue and allowed for some, at first glance, weird negation.

afbeelding

vs

afbeelding

amickan (approved):
LGTM

PR #4783, opened 2026-06-09
Add a copy button for image pks

jmsmkn:
Separates out the button from the rendered link, and moves that to a Django template partial as the format_html was getting a bit complex. Adds a span so the links and buttons do not line break.

Examples:

Screenshot 2026-06-09 at 16 23 16

Screenshot 2026-06-09 at 16 23 00

Closes https://github.com/DIAGNijmegen/rse-roadmap/issues/491


amickan (approved):
LGTM

PR #4784, opened 2026-06-09
Fix empty column in cost overviews

chrisvanrun:
See: https://github.com/DIAGNijmegen/rse-roadmap/issues/475#issuecomment-4660820243


app/grandchallenge/pages/templates/pages/challenge_statistics.html
```diff
@@ -65,6 +65,14 @@ <h4>Challenge Costs</h4>
                     <tr>
                         <th>Challenge Status</th>
                         <th>Should be Open</th>
+                        <th>
+                            % Active Budget Consumed
```
jmsmkn: Adding this back means that the number of columns in the header and the number of columns in the rows do not match. `% Active Budget Consumed` is already present below this block.

chrisvanrun: I thought that reverting the initial-fix commit would have pushed that back. Fixed now.

app/grandchallenge/pages/templates/pages/challenge_statistics.html
```diff
@@ -65,18 +65,10 @@ <h4>Challenge Costs</h4>
...
-                                title="Consumed costs as a percentage of the approved costs of the currently active invoices."
-                                class="text-muted">
-                                <i class="fas fa-question-circle"></i>
-                            </span>
-                        </th>
                         <th>
                             % Active Budget Consumed
                             <span data-toggle="tooltip" data-placement="right"
-                                title="Consumed costs as a percentage of the approved costs of the currently active (i.e., non-expired) invoices."
+                                title="Consumed costs as a percentage of the approved costs of the currently active invoices."
```
jmsmkn: This now no longer matches the title in the challenge statistics, should that be updated too?

jmsmkn (approved):
LGTM

app/grandchallenge/pages/templates/pages/challenge_statistics.html
chrisvanrun: Quick note here: the tooltip got updated (reverted) again. It already matched but missed an additional clarifying comment.

PR #4788, opened 2026-06-12
Refactor utilization priority of invoices

chrisvanrun:
Adds a new annotation to `Invoice` which shows the utilization priority. The lower the number, the sooner it will get utilized (if it has available budget). 

The priority is made such that already `PAID` invoices will have higher priority than those that have available budget but haven't been paid (i.e. postpaid vs prepaid). Priority ignores current incurred costs on the invoice.

See: https://github.com/DIAGNijmegen/rse-roadmap/issues/475#issuecomment-4681336214

In addition, it adds the priority in a sortable manner to the admin invoice list as to make it absolutely obvious in which order existing invoices will be utilized.

afbeelding

### DIFF
--- a/app/grandchallenge/invoices/models.py
+++ b/app/grandchallenge/invoices/models.py
@@ -6,8 +6,16 @@ from dateutil.relativedelta import relativedelta
 from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.db import models
-from django.db.models import Count, Exists, ExpressionWrapper, F, OuterRef, Q
-from django.db.models.functions import Cast, Now
+from django.db.models import (
+    Count,
+    Exists,
+    ExpressionWrapper,
+    F,
+    OuterRef,
+    Q,
+    Window,
+)
+from django.db.models.functions import Cast, Now, RowNumber
 from django.utils.html import format_html
 from django.utils.timezone import now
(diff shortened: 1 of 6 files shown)


amickan (approved):
LGTM! 

If we rerun the command now, the mis-assigned utilizations would not be reassigned though, right? Have we decided that we don't want to do that?

app/grandchallenge/invoices/models.py
```diff
@@ -139,6 +147,17 @@ def to_check(self):
...
+    def with_utilization_priority_per_challenge(self):
+        return self.annotate(
+            is_paid=Q(payment_status=Invoice.PaymentStatusChoices.PAID)
+        ).annotate(
+            utilization_priority=Window(
+                expression=RowNumber(),
+                partition_by=[F("challenge_id")],
+                order_by=["-is_paid", "expires_on", "created"],
+            )
+        )
```
chrisvanrun, line 159: @jmsmkn, I would appreciate your 2 cents on this query. Given the number of invoices I think this query is perfectly fine? I split the `is_paid` annotation for readability.

jmsmkn: Looks fine to me.

PR #4790, opened 2026-06-12
Correct file permissions in dockerfile

koopmant:
Resolves permission denied error in `update_all_container_image_shims ` lambda task

### DIFF
--- a/dockerfiles/web-base/Dockerfile
+++ b/dockerfiles/web-base/Dockerfile
@@ -62,13 +62,10 @@ RUN mkdir -p /opt/git-lfs \
 ENV PYTHONUNBUFFERED=1\
     AWS_XRAY_SDK_ENABLED=false\
     COMPONENTS_SAGEMAKER_SHIM_VERSION=0.7.0\
-    PATH="/opt/grand-challenge/.venv/bin:/home/django/.local/bin:${PATH}"
+    PATH="/opt/grand-challenge/.venv/bin:${PATH}"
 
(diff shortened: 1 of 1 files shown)


dockerfiles/web-base/Dockerfile
```diff
@@ -62,13 +62,10 @@ RUN mkdir -p /opt/git-lfs \
 ENV PYTHONUNBUFFERED=1\
     AWS_XRAY_SDK_ENABLED=false\
     COMPONENTS_SAGEMAKER_SHIM_VERSION=0.7.0\
-    PATH="/opt/grand-challenge/.venv/bin:/home/django/.local/bin:${PATH}"
+    PATH="/opt/grand-challenge/.venv/bin:${PATH}"
 
 RUN mkdir -p /opt/grand-challenge /app /static /opt/sagemaker-shim /opt/virtualenvs \
-    && groupadd -r django && useradd -m -r -g django django \
```
koopmant, line 68: I've removed the django user because I think it's not used anymore. But to be honest, I'm not entirely sure.

jmsmkn (changes requested):
This removes the django user from the container, this is required so that it does not run as root.

dockerfiles/web-base/Dockerfile
jmsmkn: It is required, see https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user

koopmant: Thank you!

jmsmkn (approved):
LGTM!

PR #4791, opened 2026-06-15
Postpaid admin calculations

amickan:
This adds fields the Invoice admin that calculate the suggested compute and storage amounts for postpaid invoices based on the costs a challenge has incurred and what has already been paid. 

Related to https://github.com/DIAGNijmegen/rse-grand-challenge-admin/pull/779


amickan:
It's possible that @chrisvanrun is currently working on adding some of the same properties (or annotations) to the challenge or invoice models for his pitch. That's why I've left this in draft mode for now. I can rebase later.

app/grandchallenge/challenges/models.py
```diff
@@ -814,6 +816,65 @@ def percent_active_budget_consumed(self):
...
             return None
 
+    @cached_property
+    def total_projected_storage_cost_euros(self):
+        return (
+            self.size_in_storage / 1024**3 + self.size_in_registry / 1024**3
+        ) * ChallengeRequest().storage_costs_euros_per_gb
+
+    @cached_property
+    def total_consumed_compute_costs_with_write_off_euros(self):
```
chrisvanrun: Doesn't this just equal the `compute_cost_euro_millicents`?

amickan: Oh yes, of course, you're right! Not sure how I missed that...

chrisvanrun: Well, there are only 4 different fields with slightly different meanings ;-)

amickan: Just realized that this field was deprecated! So I think I do need a new property on the Challenge model to represent the consumed costs including write-off. The way to calculate this can just be via the `compute_cost_euro_millicents` of all linked invoices.

app/grandchallenge/challenges/models.py
```diff
@@ -814,6 +816,65 @@ def percent_active_budget_consumed(self):
...
+
+    @cached_property
+    def total_consumed_costs_euros(self):
+        return (
+            self.total_consumed_compute_costs_with_write_off_euros
+            + self.total_projected_storage_cost_euros
+        )
+
+    @cached_property
+    def total_paid_storage_costs_euros(self):
```
amickan: Note that the property that Chris added [here](https://github.com/DIAGNijmegen/rse-grand-challenge/pull/4792/changes#diff-bbf63a098e2626450b97fc9c187f51e22f6f4a98362db4732ab875799a5ec3aeR471-R478) cannot be used here since we want to sum only paid invoices.

amickan:
I think this can be reviewed and potentially merged before https://github.com/DIAGNijmegen/rse-grand-challenge/pull/4792. Either of these need to be rebased after merging of the other, so doesn't really matter which one goes first.

app/grandchallenge/challenges/models.py
koopmant: This link is not working for me

app/grandchallenge/invoices/admin.py
```diff
@@ -199,6 +213,69 @@ def consumed_compute_cost(self, obj):
     def write_off_compute_cost(self, obj):
         return millicents_to_euro(obj.write_off_compute_cost_euro_millicents)
 
+    @admin.display(description="Consumed compute cost charged to this invoice")
+    def invoice_consumed_compute_cost(self, obj):
+        return self.consumed_compute_cost(obj)
```
koopmant, line 218: What does this do?

app/grandchallenge/invoices/admin.py
```diff
@@ -199,6 +213,69 @@ def consumed_compute_cost(self, obj):
...
+        return euro(obj.challenge.unpaid_storage_costs_euros)
+
+    def total_unpaid_costs(self, obj):
+        if obj.total_unpaid_costs_euros:
+            return euro(obj.total_unpaid_costs_euros)
+        else:
+            return None
+
+    def suggested_total_postpaid_amount(self, obj):
+        if obj.suggested_total_postpaid_amount >= 0:
```
koopmant: I think this can be None, so the comparison would fail

app/grandchallenge/invoices/admin.py
```diff
@@ -199,6 +213,69 @@ def consumed_compute_cost(self, obj):
...
+            return euro(obj.suggested_total_postpaid_amount)
+        else:
+            return None
+
+    def surplus(self, obj):
+        if obj.surplus >= 0:
```
koopmant: Same, can be None

koopmant: Maybe check the others as well

koopmant: Ah they aren't in the fieldset when they are None, so I guess it's not an issue.

koopmant (commented):
I don't see any issues, just a question.

app/grandchallenge/invoices/admin.py
amickan: It changes the display name from `Invoice consumed compute cost` to the more verbose (but in my opinion clearer) `Consumed compute cost charged to this invoice`.

app/grandchallenge/challenges/models.py
amickan: Sorry, this is the property I meant: https://github.com/DIAGNijmegen/rse-grand-challenge/blob/896cf58ba365eb5f13560303603c510ecfece5ff/app/grandchallenge/invoices/models.py#L471-L477

app/grandchallenge/challenges/models.py
```diff
@@ -842,6 +844,65 @@ def percent_active_compute_budget_consumed(self):
         else:
             return None
 
+    @cached_property
+    def total_projected_storage_cost_euros(self):
+        return (
+            self.size_in_storage / 1024**3 + self.size_in_registry / 1024**3
```
chrisvanrun: I think we have a `settings.GIGABYTE` that is more commonly used.

app/grandchallenge/challenges/models.py
```diff
@@ -842,6 +844,65 @@ def percent_active_compute_budget_consumed(self):
...
+        ) * ChallengeRequest().storage_costs_euros_per_gb
```
chrisvanrun: This seems very odd, would it work to change the property to a staticmethod on the `ChallengeRequest`.

app/grandchallenge/challenges/models.py
```diff
@@ -842,6 +844,65 @@ def percent_active_compute_budget_consumed(self):
...
+
+    @cached_property
+    def total_consumed_compute_costs_with_write_off_euros(self):
```
chrisvanrun: General note on these calculations. For accuracy and precision we generally calculate everything on the millicent level. We only show euros when humans are involved. Any within-system calcs are best done with millicents.

I suggest we change these properties to the euro_millicent level.

app/grandchallenge/challenges/models.py
koopmant: clear!

app/grandchallenge/invoices/models.py
```diff
@@ -502,6 +503,81 @@ def percent_compute_budget_consumed(self):
...
+                + self.consumed_compute_cost_euro_millicents / 1000 / 100
+            )
+
+    @cached_property
+    def suggested_total_postpaid_amount(self):
+        if (
+            not self.payment_type == PaymentTypeChoices.POSTPAID
+            or not self.payment_status == PaymentStatusChoices.INITIALIZED
+        ):
+            return None
```
chrisvanrun: Should we not raise something here instead, as we should never get here?

app/grandchallenge/invoices/admin.py
koopmant: Ah all right. Does it have to have a different name? `invoice_consumed_compute_cost` vs `consumed_compute_cost`. That confused me initially.

koopmant: Not a big deal though

koopmant (dismissed):
LGTM, name change of property optional

app/grandchallenge/invoices/admin.py
amickan: Oh, I see. Yes, because the field is already displayed earlier in the admin under its original name.

app/grandchallenge/invoices/admin.py
```diff
@@ -253,11 +330,46 @@ def get_queryset(self, request):
...
         )
 
+    def get_fieldsets(self, request, obj=None):
+        fieldsets = list(self.BASE_FIELDSETS)
+        if (
+            obj
+            and obj.payment_type == PaymentTypeChoices.POSTPAID
+            and obj.payment_status == PaymentStatusChoices.INITIALIZED
+        ):
+            fieldsets.append(
```
chrisvanrun, line 357: I see this as a somewhat OK addition.

However, the extra fields are throwing confusion in the mix with respect to the actual budget fields. Further more, they require switches for POSTPAID+INITIALIZED.

I'd suggest a dedicated view for this. That would allow us to namescope it somewhat? We can also push the challenge-level properties to this? We could show the link to the page only for postpaid invoices (as the switch you've made here).

app/grandchallenge/challenges/models.py
```diff
@@ -842,6 +844,65 @@ def percent_active_compute_budget_consumed(self):
...
+            sum(
+                invoice.compute_cost_euro_millicents
+                for invoice in self.invoices.all()
+            )
+            / 1000
+            / 100
+        )
+
+    @cached_property
+    def total_consumed_costs_euros(self):
```
chrisvanrun: 'consumed' feels like a missnomen here. Because you also have the projected (i.e. not consumed yet) in here.

app/grandchallenge/challenges/models.py
```diff
@@ -842,6 +844,65 @@ def percent_active_compute_budget_consumed(self):
...
+    @cached_property
+    def unpaid_storage_costs_euros(self):
+        return max(
+            self.total_projected_storage_cost_euros
+            - self.total_paid_storage_costs_euros,
+            0,
+        )
+
+    @cached_property
+    def compute_ratio(self):
```
chrisvanrun: This is not a ratio, rather a proportion/fraction/share. A ratio is a common USA thing where we as Europeans have very little intuition for. Steer clear.

app/tests/invoices_tests/test_models.py
```diff
@@ -449,6 +451,47 @@ def test_complimentary_invoice_status_badge(invoice_kwargs, badge):
     assert invoice.get_status_badge() == badge
 
 
+@pytest.mark.django_db
+def test_postpaid_suggested_costs_properties():
```
chrisvanrun: This tests it somewhat, but doesn't hit cornercases that I think might screw up things.

For instance, pushing the projected storage costs beyond what is covered by the prepaid.

Or what if we have two postpaid invoices? Does this still work?

chrisvanrun (changes requested):
A few nitpicks, and a general worry about the confusing mix of different totals in the models.

app/grandchallenge/invoices/admin.py
amickan: I don't entirely follow what you mean by a separate view and how that would solve the confusion. This section here is only visible for postpaid, initialized invoices. For the person navigating around in the admin, viewing invoices, I would say, there is no confusion. I do agree that having all of these extra properties on the models makes things more complex in the codebase for developers - I had that same thought while adding them - but I don't see how a separate view would fix that. Can you elaborate a bit?

app/grandchallenge/challenges/models.py
amickan: I think that should work. Makes sense.

app/grandchallenge/challenges/models.py
amickan: Maybe just total_costs then?

amickan: or should we be explicit and call it total_consumed_and_projected_costs?

app/grandchallenge/invoices/models.py
amickan: Makes sense. I've changed it to raise a NotImplementedError.

app/grandchallenge/challenges/models.py
amickan: Realized that I'm only using this property as a switch for whether or not to calculate the compute share. Does not need to be a property then. I will remove it.

app/grandchallenge/invoices/admin.py
amickan: Had a synchronous call about this and agreed that a section in the admin (as implemented here) is the way to go. To minimize the confusion with the extra model properties, we agreed to group them together and separate them from the other model properties through comments.

app/tests/invoices_tests/test_models.py
amickan: I have added a whole bunch of extra tests, which hopefully cover the cornercases well enough.

app/grandchallenge/challenges/models.py
```diff
@@ -927,6 +929,76 @@ def forge_model(self):
...
+
+    @cached_property
+    def total_projected_storage_cost_euro_millicents(self):
+        return (
+            self.size_in_storage / settings.GIGABYTE
+            + self.size_in_registry / settings.GIGABYTE
+        ) * storage_cost_euro_millicents_per_gb()
+
+    @cached_property
+    def total_consumed_compute_costs_with_write_off_euro_millicents(self):
```
amickan: Just FYI Chris: I opted to keep this name as is (over renaming it to `total_utilized_compute_cost`. I personally find the term `utilized` hard to grasp and prefer this explicit / descriptive name. Is that ok with you? The name is consistent with your terminology in my eyes (you chose to call compute costs without write-off `consumed_compute_cost`).

chrisvanrun: I had to call the split parts, but why not stick to the generic `Challenge.compute_cost_euro_millicents`? That is the generally aggreed name for this property on each object in the challenge-sphere.

For instance, you could also, alternatively, summate the `Phase.compute_cost_euro_millicents` and get to the same number.

app/grandchallenge/challenges/models.py
```diff
@@ -975,6 +1047,17 @@ def submission_pdf_path(instance, filename):
     )
 
 
+def storage_cost_euro_millicents_per_gb():
```
chrisvanrun, line 1032: Just a brain fart: making this a module-based variable that gets calculated once would likely be fragile since the module might stick in memory while the env vars are updated. As such, having this as a function is a good call I think.

app/tests/challenges_tests/test_models.py
```diff
@@ -936,6 +940,208 @@ def test_budget_properties():
...
+        == 67
+    )
+    assert (
+        round(challenge.total_paid_storage_costs_euro_millicents / 1000 / 100)
+        == 1
+    )
+    assert (
+        round(challenge.unpaid_storage_costs_euro_millicents / 1000 / 100)
+        == 66
+    )
```
chrisvanrun, line 1049: Style nitpick, but generally the tests with millicents has this inverted which saves you from having to round it:

I.e. 

```Python
assert challenge.unpaid_storage_costs_euro_millicents == 66 * 1000 * 100
```

app/tests/challenges_tests/test_models.py
```diff
@@ -936,6 +940,208 @@ def test_budget_properties():
...
+    ),
+)
+@pytest.mark.django_db
+def test_unpaid_storage_costs_euro_millicents_capped_at_0(
+    already_paid_amount, to_be_paid_amount
+):
+    challenge = ChallengeFactory(
+        size_in_storage=30 * 1024**3,
+        size_in_registry=70 * 1024**3,
+    )  # 67 Euro utilized storage
```
chrisvanrun: Just a wonder: would it make sense to add an assert somewhere that checks this? As, at one point we'll be updating the size-to-euro conversion. Or maybe fix this in a settings overwrite at the start of these tests. That way we solidify the test?

chrisvanrun (dismissed):
A few nitpicks, otherwise it looks good! I hope I haven't sent you down the rabbit hole with requesting additional tests. So many corners!

app/grandchallenge/challenges/models.py
amickan: True. Let's use that.

app/grandchallenge/challenges/models.py
amickan: Indeed.

app/grandchallenge/challenges/models.py
amickan: Hm the challenge currently still has the deprecated field with that same name. We should remove that first. Can that be done now, @chrisvanrun ?

app/tests/challenges_tests/test_models.py
amickan: In this specific case I would not need to round, but in a lot of other places I still would need to round. This is anyway outdated, I added a function that does the rounding and division and use that everywhere now.

app/tests/challenges_tests/test_models.py
amickan: That is already part of the tests. I check the value of `total_projected_storage_cost_euro_millicents` in a couple of tests.

chrisvanrun: Ok. Let's hope that hints our future selves sufficiently.

app/grandchallenge/challenges/models.py
chrisvanrun: Yep, see PR!

chrisvanrun (approved):
LGTM!

PR #4792, opened 2026-06-15
Show challenge budget utilization to challenge organizers

chrisvanrun:
Part of pitch:

In addition of adding the utilization markup:
- Renamed `percent_budget_consumed` to `percent_compute_budget_consumed` to better match the content.
- Added an approved variant of the storage costs that is bound by budget authorization.
 
## Showcase
afbeelding

afbeelding

afbeelding

### DIFF
--- a/app/grandchallenge/invoices/templates/invoices/invoice_detail.html
+++ b/app/grandchallenge/invoices/templates/invoices/invoice_detail.html
@@ -91,7 +91,7 @@
 
     <hr>
 
-    <h3>Compute Costs</h3>
+    <h3 class="mt-3">Compute Costs</h3>
 
 
     {% if not object.is_budget_authorized %}
@@ -100,8 +100,7 @@
         </span>
     {% else %}
 
-
-    <dl class="mt-3 px-3 border-left rounded">
+    <dl class="mt-1">
         <dt>Compute budget spent</dt>
         <dd>
             <div class="progress my-1">
--- a/app/grandchallenge/invoices/templates/invoices/invoice_detail.html
+++ b/app/grandchallenge/invoices/templates/invoices/invoice_detail.html
@@ -81,10 +81,7 @@
                 <dt>Storage capacity reservation</dt>
                 <dd class="mb-0">€&nbsp;{{ object.storage_costs_euros }}</dd>
 
-                <div class="d-flex align-items-center">
-                    <hr class="flex-grow-1 my-0 mr-2" style="width: 100px;">
-                    <span>+</span>
-                </div>
+                <hr class="mt-1 mb-3 ml-0" style="width: 60%">
 
                 <dt>Total Amount</dt>
                 <dd>{{ invoice.total_amount_euros|euro:0 }}</dd>
--- a/app/grandchallenge/invoices/templates/invoices/invoice_detail.html
+++ b/app/grandchallenge/invoices/templates/invoices/invoice_detail.html
@@ -70,7 +70,7 @@
                 {% endif %}
             </dl>
         </div>
-        <div class="col-md-6">
+        <div class="col-md-6 pl-3 border-left rounded">
             <dl>
                 <dt>Base contribution</dt>
                 <dd>€&nbsp;{{ object.support_costs_euros }}</dd>
@@ -81,9 +81,9 @@
                 <dt>Storage capacity reservation</dt>
                 <dd class="mb-0">€&nbsp;{{ object.storage_costs_euros }}</dd>
 
-                <hr class="mt-1 mb-3 ml-0" style="width: 60%">
+                <hr class="mt-3 ml-0" style="width: 44%">
 
-                <dt>Total Amount</dt>
+                <dt>Total amount</dt>
                 <dd>{{ invoice.total_amount_euros|euro:0 }}</dd>
             </dl>
         </div>
@@ -91,7 +91,7 @@
 
     <hr>
 
-    <h3>Compute Budget Consumption</h3>
+    <h3>Compute Costs</h3>
 
 
     {% if not object.is_budget_authorized %}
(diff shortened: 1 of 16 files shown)


app/grandchallenge/invoices/templates/invoices/invoice_detail.html
```diff
@@ -17,60 +17,108 @@
...
+                        <dd>{{ object.follow_up_on|default_if_none:"-" }}</dd>
+                    {% endif %}
 
+                    <dt>Expires on</dt>
+                    <dd>
+                        {{ object.expires_on }}
+                        {% if object.is_expired %}
+                            <i class="fas fa-exclamation-triangle text-danger" title="Invoice is expired."></i>
+                        {% endif %}
+                    </dd>
```
amickan, line 69: I guess this only makes sense for paid invoices?

app/grandchallenge/invoices/templates/invoices/invoice_detail.html
```diff
@@ -17,60 +17,108 @@
...
+            </dl>
+        </div>
+    </div>
+
+    <hr>
+
+    <h3 class="mt-3">Compute Costs</h3>
+
+
+    {% if not object.is_budget_authorized %}
```
amickan: Does this also work for past invoices? Or maybe put differently, why not check `compute_cost_euro_millicents`?

amickan (dismissed):
LGTM. Just 2 minor comments.

app/grandchallenge/invoices/templates/invoices/invoice_detail.html
chrisvanrun: You mean the 'expires on'? It is now shown for all invoices that are **not** cancelled. Imho it should also as soon as it is fixed as to inform users, even if they pay the expiration date is already set. (No use postponing your payment there!)

amickan: Ok, fair enough.

app/grandchallenge/invoices/templates/invoices/invoice_detail.html
chrisvanrun: The moment you can link a utilization to the invoice, that is the moment this section should show. Switching on `compute_cost_euro_millicents` would mean it only shows _after_ the first utilization gets linked.

No linkage should be possible for non-authorized invoices. So I think it is the best effective gate?

chrisvanrun: The idea being that when organizers go check their invoices, they find it paid and see the usage. That DOES make me realize I haven't actually tested this in a NOOP setting where there isn't any budget / invoice.

amickan: Hm, and what about expired invoices? Don't we want to show their (albeit outdated) budget consumption information?

chrisvanrun: Expired invoices and budget authorized are seperate things. Expiry comes into play when calculating the available budget. When expired this is always 0, else it is dependent on how much was consumed.

Buget authorization is a bookkeeping things, expired and available budget is the workfloor =).

As for not testing things without an invoice: it seems someone spent 'None' budget. You mean nothing? No, I mean 'None%':

afbeelding

chrisvanrun: I think a switch for `approved_compute_cost_euro_millicents` is best here.

amickan: > Expired invoices and budget authorized are seperate things. Expiry comes into play when calculating the available budget. When expired this is always 0, else it is dependent on how much was consumed.

Yes indeed, but you could still show how much budget an expired invoice consumed when it was still authorized. But that complicates the display, and anyway, I believe we discussed on the pitch that we do explicitly not want to show the budget consumption for past (i.e. expired) challenges. So never mind!

amickan (dismissed):

app/grandchallenge/invoices/templates/invoices/invoice_detail.html
chrisvanrun: Now I am confused a bit. I think that is the case?

AFAIK. The `approved_compute_cost_euro_millicents` and `consumed_compute_cost_euro_millicents` are not changed if ~a challenge~ an invoice expires. So the progress bar and the total consumed EURO amount remain (`consumed_compute[...]`). Only the available compute budget drops to 0.

chrisvanrun: Oh wait, I think you are referring to showing all the now expired budget consumption on the challenge level. Yeah, no... those are indeed hidden on purpose.

amickan: Right, I somehow thought that the approved_compute_cost_euro_millicents would be 0 for expired invoices. Thanks for clarifying. 

I still remember a discussion about that on the pitch: read [this](https://github.com/DIAGNijmegen/rse-roadmap/issues/465#issuecomment-4108617254) comment and the ones before it for context. @miriam-groeneveld should make a final call, I guess.

chrisvanrun: > I still remember a discussion about that on the pitch: read https://github.com/DIAGNijmegen/rse-roadmap/issues/465#issuecomment-4108617254 comment and the ones before it for context. @miriam-groeneveld should make a final call, I guess.

Afaik that is about not merging this until the expiry dates have been set.

miriam-groeneveld: Not sure that I understand the question, for expired invoices we don't want to show how much was consumed (so we avoid all challenge orgs asking for their money back). That's why we will set the expiry dates for the older challenges, to avoid those questions

chrisvanrun:
Note: I've taken the semantic decision to name non-expired invoices 'active' and rename the `Challenge.active_invoice` to `Challenge.utilization_invoice`.

amickan (approved):
LGTM

PR #4796, opened 2026-06-15
Add view_plane to json-annotations

pkcakeout:
Add the new optional `view_plane` json data to our annotations.

_Note:_ This is a little deviation from our original planned naming convention which was only "plane". But when I saw the json in from of me for the first time, I did not quite like the list of generic names that underneath each other:

"corners", "boxes", "lines", ... all make sense in the context of an annotation, but "plane" felt a bit too generic and like it needed some context.

Added a single test to verify that plane-data indeed is an optional additional parameters now. Please tell me if something is missing or needs changing!


jmsmkn (approved):
LGTM!

PR #4797, opened 2026-06-15
Remove need for log parsing on sagemaker execution errors

koopmant:
Requires sagemaker-shim to be updated to 0.8.1

Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/828

### DIFF
--- a/app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
+++ b/app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
@@ -27,6 +27,8 @@ class ObjectParams(NamedTuple):
     pk: UUID
 
 
+# TODO: refactor EndpointOrchestrator, AmazonSageMakerEndpointExecutor,
+#  AmazonSageMakerTrainingExecutor and AmazonSageMakerBaseExecutor to simplify
 class AmazonSageMakerEndpointExecutor(AmazonSageMakerTrainingExecutor):
     def __init__(self, *args, runtime_setup_result_key=None, **kwargs):
         super().__init__(*args, **kwargs)
--- a/app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
+++ b/app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
@@ -27,6 +27,19 @@ class ObjectParams(NamedTuple):
     pk: UUID
 
 
+class AmazonSageMakerEndpointExecutor(AmazonSageMakerTrainingExecutor):
+    def __init__(self, *args, runtime_setup_result_key=None, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.__runtime_setup_result_key = runtime_setup_result_key
+
+    @property
+    def runtime_setup_result_key(self):
+        if self.__runtime_setup_result_key:
+            return self.__runtime_setup_result_key
+        else:
+            return super().runtime_setup_result_key
+
+
 class EndpointOrchestrator:
     def __init__(
         self,
@@ -39,8 +52,9 @@ class EndpointOrchestrator:
         signing_key,
         time_limit=settings.ALGORITHM_ENDPOINTS_MAXIMUM_INVOCATION_DURATION,
         algorithm_model=None,
+        runtime_setup_result_key=None,
     ):
-        self._executor = AmazonSageMakerTrainingExecutor(
+        self._executor = AmazonSageMakerEndpointExecutor(
             job_id=job_id,
             exec_image_repo_tag=exec_image_repo_tag,
             memory_limit=memory_limit,
@@ -53,6 +67,7 @@ class EndpointOrchestrator:
             input_bucket_name=settings.ALGORITHM_ENDPOINTS_INPUT_BUCKET_NAME,
             output_bucket_name=settings.ALGORITHM_ENDPOINTS_OUTPUT_BUCKET_NAME,
             use_task_list=False,
+            runtime_setup_result_key=runtime_setup_result_key,
         )
         self._endpoint_name = endpoint_name
         self._exec_image_repo_tag = exec_image_repo_tag
@@ -143,6 +158,10 @@ class EndpointOrchestrator:
     @property
     def _time_limit(self):
         return self._executor._time_limit
+
+    @property
+    def runtime_setup_result_key(self):
+        return self._executor.runtime_setup_result_key
 
     def provision_auxiliary_data(self):
         if self._algorithm_model:
--- a/app/grandchallenge/components/backends/base.py
+++ b/app/grandchallenge/components/backends/base.py
@@ -40,6 +40,10 @@ from grandchallenge.components.backends.exceptions import (
     ComponentException,
     UncleanExit,
 )
+from grandchallenge.components.backends.utils import (
+    NO_ERRORS_IN_LOG_MESSAGE,
+    user_error,
+)
 from grandchallenge.components.models import (
     APIMethodChoices,
     ComponentInterface,
@@ -899,11 +903,14 @@ class Executor(ABC):
         error_message = inference_result.user_safe_error_message
         if not error_message:
             if inference_result.user_process_last_stderr_lines:
-                error_message = (
-                    inference_result.user_process_last_stderr_lines[-1]
-                )
-        if not error_message and inference_result.return_code != 0:
-            error_message = "User process returned non-zero exit code without error message"
+                for line in reversed(
+                    inference_result.user_process_last_stderr_lines
+                ):
+                    error_message = user_error(line)
+                    if error_message != NO_ERRORS_IN_LOG_MESSAGE:
+                        break
+            else:
+                error_message = NO_ERRORS_IN_LOG_MESSAGE
         return error_message
 
     def _handle_completed_job(self):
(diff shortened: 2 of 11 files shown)


dockerfiles/web-base/Dockerfile
```diff
@@ -73,7 +73,7 @@ USER django:django
 # Fetch and install sagemaker shim for shimming containers
 RUN mkdir -p /opt/sagemaker-shim \
     && wget "https://github.com/DIAGNijmegen/rse-sagemaker-shim/releases/download/v${COMPONENTS_SAGEMAKER_SHIM_VERSION}/sagemaker-shim-${COMPONENTS_SAGEMAKER_SHIM_VERSION}-Linux-x86_64.tar.gz" -P /opt/sagemaker-shim/ \
-    && echo "aec4bc64f4ce90bd4368e89ebb8db571c84b10efe3908419bdd7fe17480f9ce0  /opt/sagemaker-shim/sagemaker-shim-${COMPONENTS_SAGEMAKER_SHIM_VERSION}-Linux-x86_64.tar.gz" | shasum -c - || exit 1 \
+    && echo "189df2402f6592e72ab1366d1cc52ea5547909f8515d62960094bbbd22eae68d  /opt/sagemaker-shim/sagemaker-shim-${COMPONENTS_SAGEMAKER_SHIM_VERSION}-Linux-x86_64.tar.gz" | shasum -c - || exit 1 \
```
jmsmkn, line 76: We need to think about the upgrade strategy here. When this is deployed the containers will still be on the old version of sagemaker shim. There will be a time when both versions of the shim are running. We also need to take into account that some containers cannot be upgraded as they have existing jobs defined using the old shim version.

Perhaps:
1. Make a new PR where SageMaker shim is upgraded
2. We deploy, and upgrade the containers, making no assumptions that the `runtime_setup_result_key` will be present. There may be some strange errors for a while if we need `inference_result_skipped` but the new version of the shim does not provide it, but, those are anyway failed jobs. Upgrading the containers will take 24-36 hours to fully complete.
3. Make a 2nd PR where `runtime_setup_result_key` is interpreted, and we know that all the containers _should_ produce it.

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
```diff
@@ -27,6 +27,19 @@ class ObjectParams(NamedTuple):
     pk: UUID
 
 
+class AmazonSageMakerEndpointExecutor(AmazonSageMakerTrainingExecutor):
+    def __init__(self, *args, runtime_setup_result_key=None, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.__runtime_setup_result_key = runtime_setup_result_key
+
+    @property
+    def runtime_setup_result_key(self):
```
jmsmkn, line 38: Why this construction?

dockerfiles/web-base/Dockerfile
koopmant: Ah yes, good point. I think that strategy should work.

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
koopmant: I could not think of anything else. I need to override the property because for invocations the key should still use the endpoint id, not the invocation id. A property can only be replaced by a child class property. 

I know this construction needs cleaning up afterwards, but I didn't want to overhaul everything while we are half way. I have a todo to clean up EndpointOrchestrator, AmazonSagemakerTrainingExecutor, and AmazonSagemakerBaseExecutor already. 

Do you have an alternative?

dockerfiles/web-base/Dockerfile
koopmant: Created a separate PR for the update. We can merge this one once all containers are updated.

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
jmsmkn: Okay - maybe add that as a TODO?

app/grandchallenge/components/backends/base.py
```diff
@@ -819,6 +833,51 @@ def _get_upload_input_content_task(self, *, content, key):
             key=key,
         )
 
+    def _get_runtime_setup_result(self):
```
jmsmkn: A lot of this seems duplicated from _get_inference_result. Maybe create another method for fetching and validating an object with the digest and use that in both places?

Loading and validating the json from each object is sufficiently specific to each method so the duplication there is fine.

koopmant: Thanks, I've added a new method for it.

app/grandchallenge/components/backends/base.py
```diff
@@ -886,7 +914,7 @@ def _handle_completed_job(self):
         elif users_process_exit_code == 137:
             raise ComponentException(SystemErrorMessages.MEMORY_LIMIT_EXCEEDED)
         else:
-            raise ComponentException(user_error(self.stderr))
+            raise ComponentException(inference_result.user_safe_error_message)
```
jmsmkn: We only want the last, non-empty line here. Does `inference_result.user_safe_error_message` guarantee that? I thought it was a string that could contain newlines?

jmsmkn: Hmm, looking back at https://github.com/DIAGNijmegen/rse-sagemaker-shim/pull/73 - does `user_safe_error_message` get set when the users process raises an exception? I only see it being set when the time limit is exceeded.

koopmant: Right, sorry this still relied on the shim setting only the last line, but we changed that. 

It catches UserSafeError here: https://github.com/DIAGNijmegen/rse-sagemaker-shim/pull/73/changes#diff-43235cc7fa7b5d545be1b1c23d8f5d47bfee70f9c617d002a436c725a1d99591L1107-L1124

That should send all the errors and we need to process those here with `user_error`

koopmant: Okay, I did not test that `str(exception_group)` we added in the shim and turns out it does not show the error messages; it only prints `(1 sub-exception)`. That will need to be fixed first...

jmsmkn: That github link does not work for me - could you link to the location on main?

koopmant: https://github.com/DIAGNijmegen/rse-sagemaker-shim/blob/main/sagemaker_shim/models.py#L1154-L1173

jmsmkn: Thanks!

koopmant: https://github.com/DIAGNijmegen/rse-sagemaker-shim/pull/75

koopmant: Added `_get_error_message` to handle the `user_safe_error_message` and `user_process_last_stderr_lines` returned in inference result to select a suitable error message.

app/grandchallenge/components/backends/base.py
```diff
@@ -893,6 +894,18 @@ def _get_inference_result(self):
...
+    @staticmethod
+    def _get_error_message(*, inference_result):
+        error_message = inference_result.user_safe_error_message
+        if not error_message:
+            if inference_result.user_process_last_stderr_lines:
+                error_message = (
+                    inference_result.user_process_last_stderr_lines[-1]
+                )
+        if not error_message and inference_result.return_code != 0:
+            error_message = "User process returned non-zero exit code without error message"
```
jmsmkn: We already use `"No errors were reported in the logs."` elsewhere.

koopmant: Ah thanks

app/grandchallenge/components/backends/base.py
```diff
@@ -893,6 +894,18 @@ def _get_inference_result(self):
 
         return inference_result
 
+    @staticmethod
+    def _get_error_message(*, inference_result):
+        error_message = inference_result.user_safe_error_message
+        if not error_message:
+            if inference_result.user_process_last_stderr_lines:
```
jmsmkn, line 905: We probably want to iterate over the `inference_result.user_process_last_stderr_lines` in reverse, for each line: splitlines, for each subline in reverse: strip, if anything left: return the first non-empty match. Just in case they produce something like:

user_process_last_stderr_lines = ["pre error", "\nmy error here \n\r\n", " ", "\n\n \n \r\n "].

Should return "my error here".

koopmant: Done

app/grandchallenge/components/backends/base.py
koopmant: Using `user_error()` now

jmsmkn:
This looks good to me. The defaults for `user_process_last_stderr_lines: list[str] = []` (2 places) and `user_safe_error_message: str = ""` can be dropped as they will always be present once 8.0.1 is deployed, right?

We will need to do the same strategy of a separate PR for patching so that the error parsing continues to work until 8.0.1 is fully deployed.

koopmant:
> The defaults for user_process_last_stderr_lines: list[str] = [] (2 places) and user_safe_error_message: str = "" can be dropped as they will always be present once 8.0.1 is deployed, right?

True. Removed those.

jmsmkn (approved):
LGTM!

PR #4799, opened 2026-06-15
Update serializers for invoke API

koopmant:
See https://github.com/DIAGNijmegen/rse-roadmap/issues/477

### DIFF
--- a/app/grandchallenge/reader_studies/serializers.py
+++ b/app/grandchallenge/reader_studies/serializers.py
@@ -14,6 +14,7 @@ from rest_framework.serializers import (
     SerializerMethodField,
 )
 
+from grandchallenge.algorithms.serializers import AlgorithmInterfaceSerializer
 from grandchallenge.components.schemas import ANSWER_TYPE_SCHEMA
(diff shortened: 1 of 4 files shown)


app/grandchallenge/reader_studies/serializers.py
```diff
@@ -55,6 +77,9 @@ class QuestionSerializer(HyperlinkedModelSerializer):
     look_up_table = LookUpTableSerializer(read_only=True, allow_null=True)
     widget = CharField(source="get_widget_display", read_only=True)
     interactive_algorithms = SerializerMethodField()
+    algorithms = ReaderStudyAlgorithmImplementationSerializer(
```
jmsmkn, line 80: Please check that there is not an n+1 problem with this when using the question list API view, I imagine that the algorithms will need to be pre-fetched.

koopmant: Thanks! Should be resolved now.

jmsmkn (approved):
LGTM!

PR #4803, opened 2026-06-17
Do not raise errors on early exit from update_container_image_shim

jmsmkn:
We can return here as there is nothing to do. As we now run `update_all_container_image_shims` as a periodic task it will be caught again later. We can keep that periodic task so that the containers will be automatically updated when SageMaker shim is updated. We will know when the upgrade is complete when the `update_all_container_image_shims` task returns `0`, indicating that no more upgrade tasks were created.


koopmant (dismissed):
LGTM

Should we add tests for tasks like this?

jmsmkn:
Tests added with kiro.

koopmant (approved):

jmsmkn:
Failed on flaky test

PR #4806, opened 2026-06-17
Submission detail improvements

amickan:
This adds a column with links to the evaluation detail pages. The link is disabled if the user does not have view permission. 
The Details column no longer contains links to the evaluation detail page.
Screenshot 2026-06-17 at 11 03 26

Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/824
Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/823
Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/822

### DIFF
--- a/app/config/settings.py
+++ b/app/config/settings.py
@@ -1396,7 +1396,7 @@ if strtobool(os.environ.get("CSP_REPORT_ONLY", "False")):
     CONTENT_SECURITY_POLICY_REPORT_ONLY = CONTENT_SECURITY_POLICY
     del CONTENT_SECURITY_POLICY
 
-ENABLE_DEBUG_TOOLBAR = True
+ENABLE_DEBUG_TOOLBAR = False
 
 if DEBUG:
     # Allow localhost in development
--- a/app/config/settings.py
+++ b/app/config/settings.py
@@ -1396,7 +1396,7 @@ if strtobool(os.environ.get("CSP_REPORT_ONLY", "False")):
     CONTENT_SECURITY_POLICY_REPORT_ONLY = CONTENT_SECURITY_POLICY
     del CONTENT_SECURITY_POLICY
 
-ENABLE_DEBUG_TOOLBAR = False
+ENABLE_DEBUG_TOOLBAR = True
 
 if DEBUG:
     # Allow localhost in development
--- a/app/grandchallenge/evaluation/templates/evaluation/submission_detail.html
+++ b/app/grandchallenge/evaluation/templates/evaluation/submission_detail.html
@@ -125,7 +125,7 @@
             </thead>
             <tbody>
             {% for evaluation in object.evaluation_set.all %}
-                {% get_obj_perms request.user for evaluation as "evaluation_perms" %}
+                {% get_obj_perms request.user for evaluation as "evaluation_perms" checker %}
                 <tr>
                     <td>
                         <span title="{% if "view_evaluation" in evaluation_perms %}View evaluation details{% else %}You cannot view the details of this evaluation.{% endif %}">
@@ -152,7 +152,9 @@
                     </td>
                     <td>
                         {% if evaluation.status == evaluation.SUCCESS %}
-                            {% if not evaluation.published %}
+                            {% if evaluation.published %}
+                                Evaluation is published.
+                            {% else %}
                                 Evaluation is under review by the challenge admins.
                             {% endif %}
                         {% elif evaluation.status == evaluation.FAILURE or evaluation.status == evaluation.CANCELLED %}
--- a/app/grandchallenge/evaluation/templates/evaluation/submission_detail.html
+++ b/app/grandchallenge/evaluation/templates/evaluation/submission_detail.html
@@ -154,8 +154,6 @@
                         {% if evaluation.status == evaluation.SUCCESS %}
                             {% if not evaluation.published %}
                                 Evaluation is under review by the challenge admins.
-                            {% else %}
-                                -
                             {% endif %}
                         {% elif evaluation.status == evaluation.FAILURE or evaluation.status == evaluation.CANCELLED %}
                             {{ evaluation.error_message }}
--- a/app/grandchallenge/evaluation/templates/evaluation/submission_detail.html
+++ b/app/grandchallenge/evaluation/templates/evaluation/submission_detail.html
@@ -109,6 +109,7 @@
         >
             <thead class="thead-light">
             <tr>
+                <th></th>
                 <th>Evaluation Id</th>
                 <th>Created</th>
                 <th>Updated</th>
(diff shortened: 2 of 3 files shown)


app/grandchallenge/evaluation/templates/evaluation/submission_detail.html
```diff
@@ -140,17 +152,12 @@ <h3>Evaluations for this submission</h3>
...
                         {% if evaluation.status == evaluation.SUCCESS %}
-                            {% if evaluation.published %}
-                                <a href="{{ evaluation.get_absolute_url }}"><i class="fas fa-info-circle mr-1"></i></a>View public result
-                            {% elif "change_challenge" in challenge_perms %}
-                                <a href="{{ evaluation.get_absolute_url }}"><i class="fas fa-info-circle mr-1"></i></a>View private result
-                            {% else %}
+                            {% if not evaluation.published %}
                                 Evaluation is under review by the challenge admins.
+                            {% else %}
+                                -
```
amickan: Not entirely sure what to display here for successful evaluations. Technically, the visibility information is already shown in the visibility column (via an icon + tooltip), so we could opt to rename this column to `Error message` and only populate it for failed or cancelled evaluations. It did feel useful to have `Evaluation is under review by the challenge admins.` there though, since that information is otherwise hidden in a tooltip in the visibility column, but not sure.

jmsmkn: Maybe "Evaluation is published"?

app/grandchallenge/evaluation/templates/evaluation/submission_detail.html
```diff
@@ -123,7 +125,17 @@ <h3>Evaluations for this submission</h3>
             </thead>
             <tbody>
             {% for evaluation in object.evaluation_set.all %}
+                {% get_obj_perms request.user for evaluation as "evaluation_perms" %}
```
jmsmkn: This is going to add an n+1 issue right?

app/grandchallenge/evaluation/templates/evaluation/submission_detail.html
amickan: We could do that. It would be duplicate information to some extent since that information is already in the visibility column.

app/grandchallenge/evaluation/templates/evaluation/submission_detail.html
jmsmkn: See `ObjectPermissionCheckerMixin` for a solution.

app/grandchallenge/evaluation/templates/evaluation/submission_detail.html
jmsmkn: True, but rounds everything off. The empty details column seems strange, I feel like it needs a "this page has intentionally been left blank" or something. Unless it becomes Error details, but then "Evaluation is under review by the challenge admins." does not make sense there.

jmsmkn: Or we go "Error Details" and "Evaluation is under review by the challenge admins." moves to the "Possible Next Steps" for participants, but then the indication that the admins must do something is less clear.

app/grandchallenge/evaluation/templates/evaluation/submission_detail.html
amickan: Good catch, thanks!

app/grandchallenge/evaluation/templates/evaluation/submission_detail.html
amickan: Duplication is ok in this case, I think.

app/grandchallenge/evaluation/templates/evaluation/submission_detail.html
jmsmkn: Did you test that works?

jmsmkn: Ah I see, I was reading it as "as checker", not seeing that this is an extra arg.

app/grandchallenge/evaluation/views/__init__.py
```diff
@@ -469,6 +471,12 @@ def get_queryset(self):
             .prefetch_related("phase__optional_hanging_protocols")
         )
 
+    def get_context_data(self, *args, object_list=None, **kwargs):
```
jmsmkn: For my information, why does `(self, *args, **kwargs)` not work here?

amickan: I thought it was because of the weird signature that `CachedPhaseMixin` uses, but I can't reproduce the error I got earlier when I removed `object_list` from the arguments here. I think we should update `CachedPhaseMixin` from `def get_context_data(self, *, object_list=None, **kwargs):` to `def get_context_data(self, *args, **kwargs):`

app/grandchallenge/evaluation/views/__init__.py
```diff
@@ -93,7 +94,7 @@ def phase(self):
             Phase, slug=self.kwargs["slug"], challenge=self.request.challenge
         )
 
-    def get_context_data(self, *, object_list=None, **kwargs):
+    def get_context_data(self, *args, **kwargs):
         context = super().get_context_data()
```
jmsmkn: Existing bug: the super call is missing passing on the args and kwargs.

app/grandchallenge/evaluation/views/__init__.py
jmsmkn: Agreed, we should always use that style in case the signatures change (unlikely in Django)

app/grandchallenge/evaluation/views/__init__.py
amickan: Should have seen that! Fixing now.

app/config/settings.py
```diff
@@ -1396,7 +1396,7 @@ def get_private_ip():
     CONTENT_SECURITY_POLICY_REPORT_ONLY = CONTENT_SECURITY_POLICY
     del CONTENT_SECURITY_POLICY
 
-ENABLE_DEBUG_TOOLBAR = False
+ENABLE_DEBUG_TOOLBAR = True
```
jmsmkn:
```suggestion
ENABLE_DEBUG_TOOLBAR = False
```

amickan: oh wow!

jmsmkn: Thankfully leaving it on is noisy enough!

jmsmkn (approved):
LGTM!

app/config/settings.py
amickan: Indeed!

PR #4811, opened 2026-06-17
Get invocation logs task

koopmant:
Add separate task to retrieve the logs after an invocation.

### DIFF
--- a/app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
+++ b/app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
@@ -381,9 +381,6 @@ class EndpointOrchestrator:
     def handle_event(self, *, event):
         invocation_status = self._get_invocation_status(event=event)
 
-        self._set_task_logs(event=event)
-        # TODO: set runtime metrics
-
         if invocation_status == "Completed":
             self._handle_completed_invocation()
         elif invocation_status == "Expired":
@@ -411,5 +408,5 @@ class EndpointOrchestrator:
     def get_outputs(self, *, output_interfaces):
         return self._executor.get_outputs(output_interfaces=output_interfaces)
 
-    def _set_task_logs(self, *, event):
+    def set_task_logs(self, *, event):
         self._executor._set_task_logs(event=event, task=self._executor._job_id)
--- a/app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
+++ b/app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
@@ -199,10 +199,6 @@ class EndpointOrchestrator:
     def stderr(self):
         return self._executor.stderr
 
-    @property
-    def logs_complete(self):
-        return self._executor.logs_complete
-
     def provision_auxiliary_data(self):
         if self._algorithm_model:
             self._s3_client.copy(
--- a/app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
+++ b/app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
@@ -1,6 +1,6 @@
 import logging
 import re
-from datetime import datetime
+from datetime import datetime, timedelta
 from typing import NamedTuple
 from uuid import UUID
 
@@ -54,9 +54,10 @@ class AmazonSageMakerEndpointExecutor(AmazonSageMakerTrainingExecutor):
         )
 
     def _get_end_time(self, *, event):
-        return int(
-            datetime.fromisoformat(event.get("eventTime")).timestamp() * 1000
-        )
+        event_time = datetime.fromisoformat(event.get("eventTime"))
+        # Add buffer time to allow logs to complete
+        log_end_time = event_time + timedelta(seconds=10)
+        return int(log_end_time.timestamp() * 1000)
 
     @property
     def runtime_setup_result_key(self):
--- a/app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
+++ b/app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
@@ -199,6 +199,10 @@ class EndpointOrchestrator:
     def stderr(self):
         return self._executor.stderr
 
+    @property
+    def logs_complete(self):
+        return self._executor.logs_complete
+
     def provision_auxiliary_data(self):
         if self._algorithm_model:
             self._s3_client.copy(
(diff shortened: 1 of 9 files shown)


koopmant:
Dependent on https://github.com/DIAGNijmegen/rse-grand-challenge/pull/4797

Will need to be rebased after that is merged and some conflicts resolved.

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
```diff
@@ -27,6 +28,35 @@ class ObjectParams(NamedTuple):
...
+    def _get_start_time(self, *, event):
+        return int(
+            datetime.fromisoformat(event.get("receivedTime")).timestamp()
+            * 1000
+        )
+
+    def _get_end_time(self, *, event):
+        event_time = datetime.fromisoformat(event.get("eventTime"))
+        # Add buffer time to allow logs to complete
+        log_end_time = event_time + timedelta(seconds=10)
```
koopmant, line 59: The log times are often have timestamps later than the reported eventTime. 10 seconds is very long, but in terms of log-lines it's probably not that much. 
(Latest logs I checked had a delay of only 50ms)

app/grandchallenge/components/tasks.py
```diff
@@ -1988,3 +1991,39 @@ def parse_endpoint_invocation_outputs(
     else:
         invocation.outputs.add(*outputs)
         invocation.update_status(status=invocation.StatusChoices.SUCCESS)
+    finally:
+        parse_endpoint_invocation_logs.execute_on_commit(
+            **invocation.task_kwargs, event=event
+        )
+
+
+@lambda_task(retry_on=(LockNotAcquiredException, RetryStep), retry_delay=10)
```
koopmant: Do I have to count retries on the object if I want to limit the retries? I don't want to let this run endlessly, in case the "invocation complete" message is missing for some reason.

jmsmkn: Now that SageMaker shim is sending the last loglines directly I'm wondering about the whole need for this. Do we actually need to cache the log lines on the model?

Ultimately the logs are stored in CloudWatch, so we could get them from the admin side. If the user needs them we could have a view for them that gets the logs later directly from CloudWatch (would need to use an HTMX inner view to get them as it would be slow). Then, I don't think that we'd need to store anything beyond the log stream id, event start time and event end time.

koopmant: We're only sending the stderr loglines though. 

For how long do we store them on cloudwatch?

jmsmkn: Yes, but we only need stderr for the tasks that the platform needs to do - extracting an error message.

The logs are stored for a year, which I think is sufficient for debugging. We made this same migration recently with Workstation sessions.

I think it would simplify things, reduce double storage and downloading of logs every time you fetch a job object.

koopmant: All right, sounds good!

Should I rework the task to return the logs so we can use it in the future view?

jmsmkn: Sure, but remove the `RetryStep`, if they're incomplete now then that's fine. You can call the task with `_delay=120` or something.

koopmant: When we make a view for it, we'll probably want to store the event times. Alternatively we could use out own timestamps of created and modified. Is that a problem for later, or should I solve it within the pitch?

jmsmkn: Problem for later, this would be a new pitch.

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
```diff
@@ -30,10 +31,34 @@ class ObjectParams(NamedTuple):
...
+    @property
+    def _log_group_name(self):
+        # Hardcoded by AWS
+        return f"/aws/sagemaker/Endpoints/{self.__endpoint_name}"
+
+    def _get_start_time(self, *, event):
+        return int(
+            datetime.fromisoformat(event.get("receivedTime")).timestamp()
+            * 1000
+        )
```
jmsmkn, line 54: There is `ms_timestamp_to_datetime` available, would that help here?

jmsmkn: Ah wait, `event.get("receivedTime")` are normal timestamps?

app/grandchallenge/components/tasks.py
```diff
@@ -1927,31 +1927,34 @@ def handle_endpoint_invocation_event(*, event: dict):
...
+            **invocation.task_kwargs, event=event
         )
 
 
 @lambda_task(
     queue=LambdaTaskQueueChoices.MEM8G, retry_on=(LockNotAcquiredException,)
 )
 def parse_endpoint_invocation_outputs(
-    *, pk: str | UUID, app_label: str, model_name: str
+    *, pk: str | UUID, app_label: str, model_name: str, event: dict
```
jmsmkn, line 1957: FYI - when changing kwargs to async tasks you also need to consider jobs that are already on the queue. In this case, existing jobs will not have `event` set, so you would need to set a default here, deploy, wait for all pending tasks to have `event` set, then remove the default.

As there are unlikely to be any pending tasks right now then this is fine, but just FYI.

koopmant: Thanks. Had not realized!

app/grandchallenge/components/tasks.py
```diff
@@ -1988,3 +1991,36 @@ def parse_endpoint_invocation_outputs(
...
+        raise RuntimeError("Invocation is not ready for log parsing")
+
+    orchestrator = invocation.orchestrator
+
+    orchestrator.set_task_logs(event=event)
+
+    invocation.set_logs(
+        stdout=orchestrator.stdout,
+        stderr=orchestrator.stderr,
+        # TODO: set runtime metrics
```
jmsmkn: Again we can probably forget setting metrics and get it all directly from Cloudwatch. Likely the metrics should be set at the endpoint level anyway.

jmsmkn (dismissed):
LGTM!

app/grandchallenge/components/tasks.py
koopmant: Yes I was thinking the same thing. Runtime metrics of a short invocation would not really show much anyway. 

I'll remove the todo.

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
koopmant: No, this uses the isoformat timestamp. I could add a helper function `iso_timestamp_to_datetime`, But I'm going to an ms_timestamp again after that, to connect with the expected format in `AmazonSageMakerBaseExecutor` (which is where we use ms_timestamp_to_datetime)

This can be refactored when touching those classes

jmsmkn (approved):

PR #4812, opened 2026-06-17
Add stop endpoint action

koopmant:


app/grandchallenge/algorithms/admin.py
```diff
@@ -373,25 +373,18 @@ class EndpointAdmin(admin.ModelAdmin):
...
 
     @staticmethod
     def algorithm(obj):
         return obj.algorithm_image.algorithm
 
-    def has_add_permission(self, request, obj=None):
-        return False
+    def get_readonly_fields(self, request, obj=None):
+        if obj:
+            return [f.name for f in self.model._meta.fields]
```
jmsmkn: Why this construction? Can't add, and also if there is an object then nothing is editable? Why not set `editable=False` on the fields on the model?

koopmant: Do you want to be able to add through the admin?

And why should anything be editable here? Many of these fields all already non-editable on the model; when the change permissions was False by default, these fields would be shown as read-only. When the change permission is enabled, these fields disappear, so we'll have to define them as readonly fields to show them. 

I figured this is an easy way to not forget any fields. I only permit the change permission for stopping the endpoint.

jmsmkn: > When the change permission is enabled, these fields disappear, so we'll have to define them as readonly fields to show them.

This is the Django way to do it, I would rather we follow the practices in the framework than to try and fight it.

koopmant: Well, that's exactly what I did, didn't I? I've defined them as readonly fields. 
I fought it initially by not using the change permission at all, then realized that's not the way to go. 

I'm guessing I should set the read_only fields by listing them explicitly?

jmsmkn: Yes, please list them explicitly.

koopmant: All right, thanks. That's done. 

> Can't add

Should I make a note about this on the pitch?

jmsmkn: Sure

jmsmkn (approved):
LGTM!

PR #4813, opened 2026-06-18
Allow uploading of failure result on invocation timeout

koopmant:
If the timeout of the invocation is equal to the `InvocationMaximumTimeout` seconds, there will not be any time to upload the result, and the invocation state will be failed without a clear explanation. Add buffer time for the uploading to complete.


jmsmkn (dismissed):
LGTM!

jmsmkn (approved):

PR #4814, opened 2026-06-18
Add a claimed at field to workstations

jmsmkn:
This allows sessions to be started without a creator. The session duration and stopping now works with the claimed time, rather than the created time.

See https://github.com/DIAGNijmegen/rse-roadmap/issues/458


app/grandchallenge/workstations/models.py
```diff
@@ -676,7 +690,7 @@ def handle_maximum_duration_changed(self):
     def handle_session_stopped(self):
         SessionUtilization.objects.create(
             session=self,
-            duration=now() - self.created,
+            duration=now() - self.claimed_at,
```
jmsmkn, line 693: I'm assuming the `SessionUtilization.duration` means "billing duration", and the time for the unclaimed session will not be billed. I think we have the same meaning for that for things like algorithm jobs.

This way, the reader study credit use still matches up, but the unclaimed session time will not be accounted for. This does not matter with our current architecture (running on EC2 compute pools), but if the switch to say Fargate was made then there would be additional costs for idling.

koopmant: > I'm assuming the SessionUtilization.duration means "billing duration". 

Indeed we use the duration set on utilizations to indicate duration relevant for costs. 

> I think we have the same meaning for that for things like algorithm jobs.

I think for jobs we use the event times to calculate the total time and for warm pools we use ResourceRetainedBillableTimeInSeconds. I think you're referring to the latter?

koopmant (approved):
LGTM!

app/grandchallenge/workstations/models.py
jmsmkn: Yes for both. For training jobs we use TrainingStartTime and TrainingEndTime, which corresponds to the billable time that is displayed in AWS. For some reason BillableTimeInSeconds is not set in the event.

jmsmkn:
Save merging this until I am back from the weekend.

PR #4823, opened 2026-06-22
Add a task to maintain unclaimed sessions

jmsmkn:
The task `consolidate_unclaimed_sessions` should be run periodically. It will:

- Stop sessions that have been unclaimed for too long, or are for the wrong workstation image
- Start unclaimed workstations in each region such that the minumum number of active unclaimed sessions is maintained, subject to the ceiling of the maximum number of workstations per region

This is done in the workstations application as it relies on the default workstation implementation.

The initial implementation was done by hand, used Kiro to clean it up and generate the tests.

See DIAGNijmegen/rse-roadmap#458


amickan (approved):
LGTM! 

I should start using Kiro for my tests as well. Very comprehensive set of tests and easy to read!

PR #4824, opened 2026-06-22
Prefer to claim running sessions

jmsmkn:
If there is not an existing session for this use and there are unclaimed sessions in this region, prefer to use those.

See DIAGNijmegen/rse-roadmap#458


amickan (approved):
LGTM

Issue #4834, opened 2026-06-25
Update challenge request GPU settings


jmsmkn:
`algorithm_maximum_settable_memory_gb_for_tasks` and `algorithm_selectable_gpu_type_choices_for_tasks` should be set per challenge, not per task, and then used for each task. There has never been a case of someone wanting to specify this per task.

`algorithm_selectable_gpu_type_choices_for_tasks` should allow all lesser instance types when creating phases.

See https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/835

amickan:
I agree that it's pretty unlikely for challenges to want different GPU choices for different tasks (and most challenges only have one task anyway), but what's the harm in having the option to define this per task?

jmsmkn:
Complication of the form and codebase, another point for mistakes to come in, YAGNI.

PR #4835, opened 2026-06-25
Migrate algorithm memory and gpu types to the challenge level

jmsmkn:
`algorithm_maximum_settable_memory_gb` and `algorithm_selectable_gpu_type_choices` are now set per challenge request, not per task. `algorithm_maximum_settable_memory_gb` is now an integer (default `32`) and `algorithm_selectable_gpu_type_choices` is now a list (not a list of lists, default `["", "T4"]`).

We use field validation to ensure that NO GPU is always available, that if an A10G is set that T4s are also set, that if an A100 is set that A10G and T4 are also set. We have no particular requirements around V100 or K80 (there for completeness but not used in reality). 

See #4834


app/grandchallenge/challenges/migrations/0076_remove_challengerequest_algorithm_maximum_settable_memory_gb_for_tasks_and_more.py
```diff
@@ -0,0 +1,57 @@
...
+        migrations.AlterField(
+            model_name="challengerequest",
+            name="algorithm_selectable_gpu_type_choices",
+            field=models.JSONField(
+                default=grandchallenge.challenges.models.default_selectable_gpu_types,
+                help_text='The GPU type choices that participants will be able to select for their algorithm inference jobs. Options are ["", "A100", "A10G", "V100", "K80", "T4"].',
+                validators=[
+                    grandchallenge.core.validators.JSONValidator(
+                        schema={
+                            "$schema": "http://json-schema.org/draft-07/schema",
```
amickan, line 37: Maybe I'm missing something, but what changes here? The schema looks identical to what's done in migration 0062.

jmsmkn: Sets the default to `["", "T4"]` rather than an empty list, gives the user something to work with. Should be an SQL no-op anyway.

amickan: Ah right, I see it now.

amickan (approved):
LGTM

PR #4837, opened 2026-06-25
Always use warm pools

jmsmkn:
Now that we use lambda tasks we can generate new jobs more quickly, so can always keep the warm pools full. From using these with A10G instances we know that the shim cleans up sufficiently. This will result in time and cost savings regardless of the instance type used.


amickan (approved):
Nice!

PR #4839, opened 2026-06-26
Add instance type information to submission create

### LINKED ISSUE #4216: Add what GPU types and memory limits are accepted to each phase to the submission page
Currently unclear about what is permissible to admins and participants.

jmsmkn:
Screenshot 2026-06-26 at 10 07 06

Closes #4216

### DIFF
--- a/app/grandchallenge/evaluation/templates/evaluation/submission_form.html
+++ b/app/grandchallenge/evaluation/templates/evaluation/submission_form.html
@@ -59,13 +59,16 @@
                 <a href="{% url 'evaluation:phase-update' challenge_short_name=challenge.short_name slug=phase.slug %}">
                     Phase Settings</a>.
             </p>
-            {% if phase.submission_kind == phase.SubmissionKindChoices.ALGORITHM and not phase.external_evaluation %}
+            {% if phase.submission_kind == phase.SubmissionKindChoices.ALGORITHM %}
                 <p>
                     For each submission, {{ phase.jobs_to_schedule_per_submission }}
                     algorithm job{{ phase.jobs_to_schedule_per_submission|pluralize }} will be created
                     from the {{ phase.jobs_to_schedule_per_submission }} valid item{{ phase.jobs_to_schedule_per_submission|pluralize }}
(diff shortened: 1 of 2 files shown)


app/grandchallenge/evaluation/templates/evaluation/submission_form.html
```diff
@@ -59,13 +59,16 @@ <h3>Create a new submission</h3>
                 <a href="{% url 'evaluation:phase-update' challenge_short_name=challenge.short_name slug=phase.slug %}">
                     Phase Settings</a>.
             </p>
-            {% if phase.submission_kind == phase.SubmissionKindChoices.ALGORITHM and not phase.external_evaluation %}
```
jmsmkn, line 62: I've removed this as it should be clear from the text that no jobs are going to be created as it should be an empty archive, and it puts all the algorithm information in one place. The external evaluators should set `phase.algorithm_time_limit` to communicate this to the user.

amickan (approved):
LGTM

app/grandchallenge/evaluation/templates/evaluation/submission_form.html
amickan: That's fine. None of that information is relevant for external evaluations though, right?

amickan: As in: they are not actually restricted to the configured GPU types options.

jmsmkn: `clean_algorithm` still runs if `external_evaluation` is True, and that enforces the instance types and memory configuration.

amickan: Ok, yes, but I mean the infrastructure that the algorithm runs on might be different since it does not run on GC.

jmsmkn: That's correct, we have no insight into that, but it still matters for the users as they can only select algorithms that match the phase configuration.

PR #4840, opened 2026-06-26
Default new algorithms for phases to select T4 or No GPU

jmsmkn:
See https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/835


amickan (approved):
LGTM. 

I would have thought that this is already the case implicitly given the order of `GPUTypeChoices` but good to make this explicit in case that changes.

jmsmkn:
It depends what is first in the list that is configured on the phase, that is usually NO_GPU. Hopefully this covers most situations and people don't even need to open the dropdown.

Issue #4816, opened 2026-06-18
Create logs and runtime metrics page for algorithm endpoints


koopmant:
This would require the following:
- handle endpoint status event to mark endpoint as stopped
- save stdout and stderr to endpoint after a stopped status event
- save runtime metrics to endpoint after a stopped status event

koopmant:
Alternatively we could keep the logs and metrics only in cloudwatch, see https://github.com/DIAGNijmegen/rse-grand-challenge/pull/4811#discussion_r3430880071

jmsmkn:
I agree that this should be retrieved from cloud watch as necessary. Then we would on need to store on the model:

- The cloudwatch log stream name (maybe not even, I don't know how well `describe_log_streams` works with `logStreamNamePrefix`)
- Logging start time
- Logging end time

Then a three part view:

- Detail page (Django Template) with two inner htmx views:
  - Logs (start off with get the most recent n)
  - Metrics

jmsmkn:
This is working well and does not need HTMX. The nice thing is that users can now get their logs and metrics whilst their algorithm is running, and see when we're waiting for capacity.

We can remove the log setting on the component jobs now.

jmsmkn:
Downside is that the metrics disappear after 15 days, and the logs after the logs group retention time, but I think this is fine.

jmsmkn:
The endpoints need:

```python3
    @property
    def _log_group_name(self):
        # Hardcoded by AWS
        return f"/aws/sagemaker/Endpoints/{self.__endpoint_name}"
```

For logs tests see `app/tests/components_tests/test_amazon_sagemaker_endpoint_backend.py` and `app/tests/components_tests/test_tasks.py` in #4831

jmsmkn:
We need to work out first _who_ gets access to the logs and metrics. For algorithm jobs the logic is very complex: https://grand-challenge.org/documentation/try-out-your-algorithm/#job-permissions

It depends on how the job was created.

We can ignore "Permission when the job is created as part of a challenge submission". Aside: I think using the Job model for Challenge submissions was a mistake as it introduced these complications. That could be corrected with Batching (https://github.com/DIAGNijmegen/rse-roadmap/issues/374).

There is a viewers group, so we need to update the Endpoint permissions so that they match what is done for Jobs:


| User/Group	| Permission | 
| --- | --- |
| Job Creator	| Change this Endpoint, Invoke this Endpoint | 
[...]

PR #4841, opened 2026-06-26
Link directly through to the evaluation logs

jmsmkn:
There is a link in the evaluation admin section but may as well take them straight through to the dedicated logs page.

See #4816


amickan (approved):
LGTM!

PR #4842, opened 2026-06-26
Fix invoice due count

### LINKED ISSUE #4691: Incorrect number of invoices due
The count of due invoices appears incorrect on the badge and in the warning message on the invoices list:

Image

amickan:
This fixes the due and overdue count annotation for invoices. 

The previous code counted distinct values of the is_overdue/is_due boolean field among the filtered rows — since every row passing the filter already has that field set to True, there's only ever one distinct value, so the count collapsed to 1 (or 0 if no rows matched) instead of reflecting the actual number of matching invoices.

Closes #4691


jmsmkn (approved):
LGTM!

PR #4844, opened 2026-06-29
Add client tokens for idempotent AWS calls

### LINKED ISSUE #4843: Set clientToken in AWS create calls
Required for API idempotency. This can be `{app_name}-{model_name}-{pk}`. Methods:

- health imaging start_dicom_import_job
- ecs run_task

jmsmkn:
For idempotent calls to AWS services. These are the only ones that support them, for many other calls they are not necessary, but if they are supported they should be set.

Closes #4843


amickan (approved):
LGTM!

PR #4848, opened 2026-07-01
Remove text_align from datatables defaults

### LINKED ISSUE #4558: Remove default alignment override for paginated datatables
Default alignment for paginated datatables overrides the datatables defaults with `align-middle` and `text-center`. This can give poorly readable output:

Image

Suggestion:
- Remove defaults (Remove `text_align` from `PaginatedTableListView`, and remove `columnDefs` from `list.mjs`)
- Check results for all uses of paginated datatables
- Add css classes where necessary

jmsmkn:
I went through as many tables as I could find and they all looked fine. Will need to double check in production.

Closes #4558


amickan (approved):
LGTM

PR #4850, opened 2026-07-01
Set algorithm image import status as cancelled when the build does not succeed

### LINKED ISSUE #4068: Initialised container image imports should be cancelled if the build fails
Image

jmsmkn:
Algorithm image imports would be left in an initialized state if the codebuild build fails. This marks them as cancelled, and also performs a data migration to update the existing ones.

There were no tests for the entire codebuild app, so new ones have been added.

This PR has been generated interactively with Kiro and all code has been reviewed by me.

Closes #4068


amickan (approved):
LGTM

PR #4852, opened 2026-07-01
Move follow creation from notifications to discussion_forums

### LINKED ISSUE #4058: Move permission check for FollowForm to FollowCreate view
Currently, we check forum and topic view permission in the `FollowForm` [clean method](https://github.com/comic/grand-challenge.org/blob/e4e1bca08dd2fc2a1382258e2302741a5676e8f7/app/grandchallenge/notifications/forms.py#L41). This should be moved to the `FollowCreate` view.

jmsmkn:
Replace the generic FollowCreate view and FollowForm in the notifications app with two specific views in discussion_forums: ForumFollowCreate and ForumTopicFollowCreate.

Permission checks (view_forum, view_forumtopic) are now enforced by ObjectPermissionRequiredMixin on the view rather than in form validation.

This PR has been implemented iteratively with kiro, all code has been pre-reviewed by me.

Closes #4058

### DIFF
--- a/app/grandchallenge/discussion_forums/models.py
+++ b/app/grandchallenge/discussion_forums/models.py
@@ -51,6 +51,13 @@ class Forum(UUIDModel):
             kwargs={"challenge_short_name": self.parent_object.short_name},
         )
 
+    @property
+    def follow_create_url(self):
+        return reverse(
+            "discussion-forums:forum-follow-create",
+            kwargs={"challenge_short_name": self.parent_object.short_name},
+        )
+
 
 class ForumTopic(FieldChangeMixin, UUIDModel):
     forum = models.ForeignKey(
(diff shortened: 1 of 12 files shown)


amickan (dismissed):
The move form a generic follow create to a forum / topic specific one makes sense. However, it means that we can't just add follow buttons for other objects (without creating custom create forms for those object follows, or changing how this works again), but I don't think we want that at the moment anyway, so this LGTM.

jmsmkn:
Indeed, the form is re-usable now, I think that can actually go back into the follow stuff. The views are concerned with authorisation which is now correct, and adding new ones of those is no big deal. It could be made less to type out with mixins but I think that is unnecessary coupling. I will re-work it a little.

jmsmkn:
Okay - I've moved the form back to the notifications app as that is now reusable, and the concerns about the permissions are now pushed to the views. I think that makes sense that for new subscription creations then new views need to be added to check the permissions.

I don't know why the rendering of the PR is so off on GitHub. Not that much has changed in the notification forms and template tags. I rebased as maybe it had something to do with a deleted file, but it is still odd. Sorry.

amickan (approved):
LGTM!

PR #4853, opened 2026-07-01
Add create to Endpoint REST API

koopmant:


app/grandchallenge/algorithms/serializers.py
```diff
@@ -134,6 +134,66 @@ class Meta:
         fields = ["api_url", "pk", "algorithm", "status"]
 
 
+class EndpointPostSerializer(serializers.ModelSerializer):
+    algorithm = HyperlinkedRelatedField(
+        queryset=Algorithm.objects.none(),
+        view_name="api:algorithm-detail",
+        write_only=True,
```
jmsmkn, line 142: Why is this write only? The user wouldn't want to read it back?

jmsmkn: Ah I see now

app/grandchallenge/algorithms/serializers.py
```diff
@@ -134,6 +134,66 @@ class Meta:
...
+    def validate(self, data):
+        algorithm = data.pop("algorithm")
+        user = self.context["request"].user
+
+        if not algorithm.active_image:
+            raise serializers.ValidationError(
+                "Algorithm image is not ready to be used"
+            )
+
+        # TODO: check for remaining algorithm user credits
```
jmsmkn: This needs to be implemented before it can be merged!

app/grandchallenge/algorithms/serializers.py
koopmant: It's the same as in the JobPostSerializer. Not sure why it's write only there. I can change it in both places.

app/grandchallenge/algorithms/serializers.py
```diff
@@ -134,6 +134,66 @@ class Meta:
...
+                queryset=Algorithm.objects.all(),
+                user=user,
+                codename="execute_algorithm",
+            )
+
+    def validate(self, data):
+        algorithm = data.pop("algorithm")
+        user = self.context["request"].user
+
+        if not algorithm.active_image:
```
jmsmkn, line 166: Don't we also need to check that the active image uses the correct api method?

app/grandchallenge/algorithms/serializers.py
koopmant: There should be nobody with add permission, so there are no users who can reach this at the moment. I can still work on this first though.

app/grandchallenge/algorithms/serializers.py
koopmant: Good point!

app/grandchallenge/algorithms/serializers.py
jmsmkn: I think this is correct - you don't actually keep the algorithm on the resulting model, it is unpicked later.

app/grandchallenge/algorithms/serializers.py
jmsmkn: Ah that's true, still, just in case.

app/grandchallenge/algorithms/serializers.py
koopmant: Ah we take it out of the data of course.

app/grandchallenge/algorithms/serializers.py
koopmant: added

app/grandchallenge/algorithms/serializers.py
koopmant: Added the validation step for credits

koopmant:
This is ready for re-review. I've also added a creators group with `add_endpoint` permission.

jmsmkn:
> I've also added a creators group with add_endpoint permission.

Why is that necessary? I think this pattern should be removed. Those sorts of permissions should be left for the site admins to set up based on say, organisation membership.

koopmant:
Hmm, yeah. I added the group because we mentioned some organizations in the pitch that are not created yet on grand-challenge. Creating new organizations would make more sense. Let me revert.

jmsmkn (approved):
LGTM!

PR #4857, opened 2026-07-03
Add score to `Answer` serialization and optimise ground-truth queries

chrisvanrun:
(Only) part of pitch:
- https://github.com/DIAGNijmegen/rse-roadmap/issues/395

This adds a `score` to the serialization of `Answer`.

While re-acquanting myself with this part of the code base I found a few glarringly inefficient snippets (3xN+1s). This, with a prefetch/select_related reduced the needed DB queries in the test example from 41 to 33.

 I also thought of a key index that should speed up the retrieval of a ground-truth considerably (and hence also speed up the saving of an `Answer`). We currently have nearly 2 million rows in the `Answer` table. With it, I think the addition of `calculate_score` to the `Answer.save` is no longer required as an optimisation. It is removed.

Note that`score` does not need to be made read-only: it always gets updated on every `Answer.save()`.


app/grandchallenge/reader_studies/models.py
```diff
@@ -2089,6 +2089,11 @@ class Meta:
         unique_together = (
             ("creator", "display_set", "question", "is_ground_truth"),
         )
+        indexes = [
+            models.Index(
+                fields=["is_ground_truth", "display_set", "question"]
+            ),
+        ]
```
jmsmkn: Indexes should be most to least selective left to right, so ground truth should be last, and given that there are many questions per display set then that should be first.

As there should only be one ground truth per display set and question a smaller partial index might be better:

```python
indexes = [
    models.Index(
        fields=["display_set", "question"],
        name="unique_ground_truth_per_question",
        condition=models.Q(is_ground_truth=True),
    ),
]
```

Depends on the query pattern though. Has this index actually been tested to see if it provides a benefit?

chrisvanrun: I haven't tested all different query variants. The test database is pretty empty. The query patterns is a bit wide spread but generally there is a filter for `is_ground_truth=False`. 

I theorized that putting the `is_ground_truth` first will also allow those queries to use the index. However, my main goal is to get the ground-truth lookup near instant. A partial index will do that while at the same time not have an impact on writing regular answers.

I'll do a quick comparison between the two against a bit more objects but fully suspect the partial will win.

jmsmkn: However, my hunch is that the existing indexes on the display set or question are already enough, and that narrows things down enough that a linear search over ground truth is trivial. Removing N+1 queries has a far higher impact.

My advice here would be to remove the index, deploy and see if it is a problem. Otherwise we are prematurely optimising for problems that may not exist.

chrisvanrun: Local testing environment isn't really conductive for this. Still, there is some lessons to be extracted.

With ~10_000 `Answer`s accross 10 `Question`s across 200 `DisplaySet`s there isn't really any benefit for the GT retrieval query: `Answer.objects.filter(is_ground_truth=True, display_set=ds, question=q)`.

| Index Strategy | N | Mean (ms) | Std Dev (ms) | Stderr (ms) |
|---|---|---|---|---|
| No index | 4 | 0.103 | 0.020 | 0.010 |
| 3-part order (original order) | 8 | 0.067 | 0.025 | 0.009 |
| 3-part order updated | 7 | 0.026 | 0.014 | 0.005 |
| Partial index | 9 | 0.086 | 0.024 | 0.008 |

Looking at the general explained plans I now realize that the 'no index' strategy uses a Bitmap to index scan both display_set and question and then combines the candidates. Even with 2 mil. rows in the `Answer` table the resulting candidate rows will remain roughly the same and don't degrade much with the total number or rows, so the final scan for ground_truth is pretty 'cheap'. Assuming production would follow roughly the same plan I actually think there isn't much benefit from adding this index after all.

app/grandchallenge/reader_studies/models.py
```diff
@@ -2147,7 +2147,7 @@ def validate(  # noqa: C901
         if not creator.has_perm("read_readerstudy", question.reader_study):
             raise ValidationError("This user is not a reader for this study.")
 
-        valid_options = question.options.values_list("id", flat=True)
+        valid_options = [option.id for option in question.options.all()]
```
amickan, line 2150: Why this change? If anything, I would assume the original to be more efficient.

amickan: Oh because you prefetch the options now!

app/grandchallenge/reader_studies/views.py
```diff
@@ -1193,20 +1193,25 @@ def ground_truth(self, request, pk=None, case_pk=None):
...
             raise Http404()
-        answers = Answer.objects.filter(
-            display_set_id=case_pk,
-            question__reader_study=reader_study,
-            is_ground_truth=True,
+        answers = (
+            Answer.objects.filter(
+                display_set_id=case_pk,
+                is_ground_truth=True,
+            )
```
amickan, line 1201: You got rid of the reader study filter here, and technically it should be redundant, but maybe it's good to add a check later that the reader study that the answer belongs to is indeed the expected one?

app/grandchallenge/reader_studies/models.py
chrisvanrun: Already prefetched, doing the options.values_list does not use the queryset cached values and does a new query every time.

app/grandchallenge/reader_studies/views.py
chrisvanrun: My inital idea was to ensure the index is used here (hence I constructed the composite index slightly weird). It runs in a task so it does not really matter. Imho if anyone doubts, we should just leave it: the ds.reader_study and question.reader_study being equal is not a DB restraint (just a validation check).

Restored.

app/grandchallenge/reader_studies/models.py
chrisvanrun: Your hunch was correct James!

amickan (approved):
LGTM!

PR #4858, opened 2026-07-03
Update the evaluation template to use the invoke API

amickan:
Updates the evaluation template to use the invoke API.

Part of https://github.com/DIAGNijmegen/rse-roadmap/issues/493

### DIFF
--- a/app/grandchallenge/forge/templates/forge/partials/example_evaluation_method/app.py.template
+++ b/app/grandchallenge/forge/templates/forge/partials/example_evaluation_method/app.py.template
@@ -1,0 +1,88 @@
+{% load forge %}
+"""
+Evaluation method inference server for Grand Challenge.
+
+You should not need to edit this file. Instead, edit evaluate.py to implement
+your evaluation logic.
+
+Grand Challenge runs your evaluation method as an HTTP server with two endpoints:
+
+  GET  /health  — Called repeatedly after the container starts. Once this
+                  responds with HTTP 200 (OK), Grand Challenge considers
+                  the server ready.
+  POST /invoke  — Called once. Your evaluation should read its input from
+                  /input, compute metrics, write results to /output, and
+                  respond with HTTP 201 (Created) when done.
+
+The lifecycle on Grand Challenge is:
+
+  1. Container starts → your server boots
+  2. /health is polled until it responds with HTTP 200 (timeout: ~5 minutes)
+  3. Inputs (predictions.json and algorithm outputs) are placed at /input
+  4. POST /invoke is called
+  5. Your code reads /input, computes metrics, writes to /output, responds
+     with HTTP 201
+  6. Container stops and outputs are uploaded to Grand Challenge
+
+Key points:
+  - The server must listen on port 4743.
+  - /health must NOT redirect — Grand Challenge treats a redirect as a failure.
+  - /invoke must respond with HTTP 201 (Created) when done. Any other response
+    is treated as a failure.
+  - Your container will have no internet access at runtime.
+
+Any implementation will do — FastAPI is used here as an example. The only
+requirements are the two endpoints and the LABEL in the Dockerfile.
+"""
+from contextlib import asynccontextmanager
+
+from fastapi import FastAPI, Response, status
+import uvicorn
+
+import evaluate
+
+from uvicorn.config import LOGGING_CONFIG
+
+
+@asynccontextmanager
+async def lifespan(app: FastAPI):
+    """Server lifespan.
+
+    This is managed by FastAPI and runs exactly once — do not call it directly.
+    If your evaluation method needs to load resources at startup, do it here
+    before the yield.
+    """
+    yield
+
+
+app = FastAPI(lifespan=lifespan)
+
+
+@app.get("/health")
+async def health():
+    """Health check endpoint.
+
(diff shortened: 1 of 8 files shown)


app/grandchallenge/forge/templates/forge/partials/example_evaluation_method/app.py.template
```diff
@@ -0,0 +1,88 @@
...
+
+@asynccontextmanager
+async def lifespan(app: FastAPI):
+    """Server lifespan.
+
+    This is managed by FastAPI and runs exactly once — do not call it directly.
+    If your evaluation method needs to load resources at startup, do it here
+    before the yield.
+    """
+    yield
```
koopmant, line 55: Is this clear enough? I can imagine the empty example could be confusing... Then again, it's probably fine?

koopmant (approved):
LGTM, just a question

app/grandchallenge/forge/templates/forge/partials/example_evaluation_method/app.py.template
amickan: Not sure either. I guess we will find out when Vilma and Alonso look at it.

chrisvanrun: I had considered suggesting to add a single `setup` function here, as some evaluation methods have started using models for reviewing results and those they'd like to load somewhere only once. However, generally, they'll have the example algorithm to work from 2, plus the one time I saw a model used they directly wanted to use 3 and those did not fit into memory all at once. That one still had to load the models sequencially so would not have used it.

/end brain dump.

app/grandchallenge/forge/templates/forge/partials/example_evaluation_method/do_save.sh.template
```diff
@@ -31,25 +44,14 @@ output_filename="${DOCKER_IMAGE_TAG}_${formatted_build_info}.tar.gz"
...
+tar -czf "$output_tarball_name" -C "${SCRIPT_DIR}/ground_truth" .
+log "Saved as: ground_truth.tar.gz"
 
-tar -czf $output_tarball_name -C "${SCRIPT_DIR}/ground_truth" .
-printf "Saved as: \e[32mground_truth.tar.gz\e[0m\n"
-
-echo "==== Done"
-echo ""
-
-printf "\e[31mIMPORTANT: Please upload the ground_truth.tar.gz as seperate Ground Truth to your Phase!\e[0m\n"
```
chrisvanrun, line 55: Do note that this is a slightly different colour to have it stand out. Not sure if we want to keep this?

app/grandchallenge/forge/templates/forge/partials/example_evaluation_method/evaluate.py.template
```diff
@@ -1,32 +1,29 @@
...
-  ./do_test_run.sh
+  1. Read algorithm outputs and associate them with ground truths via predictions.json
+  2. Calculate metrics by comparing the algorithm output to the ground truth
+  3. Repeat for all algorithm jobs that ran for this submission
+  4. Aggregate the calculated metrics
+  5. Save the metrics to /output/metrics.json
 
-This will start the evaluation and reads from ./test/input and writes to ./test/output
+To test locally:  ./do_test_run.sh
+To save for upload to GC:   ./do_save.sh
```
chrisvanrun:
```suggestion
To save for upload to Grand Challenge:   ./do_save.sh
```

chrisvanrun (approved):
LGTM, two small nitpicks.

app/grandchallenge/forge/templates/forge/partials/example_evaluation_method/do_save.sh.template
amickan: I thought it was not important enough to introduce a second log function, or to leave this printf statement in here. But if y'all think differently, I can change it back of course. Same would then hold for the algorithm model.

app/grandchallenge/forge/templates/forge/partials/example_evaluation_method/app.py.template
amickan: Yeah, I would say, let's leave it as is in. It's a nice placeholder if someone wants to load a model, and indeed they have the algorithm template as separate example indeed, so I doubt it will be confusing. But we'll see what the PhDs say.

app/grandchallenge/forge/templates/forge/partials/example_evaluation_method/evaluate.py.template
amickan: Sorry for the oversight/

app/grandchallenge/forge/templates/forge/partials/example_evaluation_method/do_save.sh.template
chrisvanrun: I thought perhaps add the colour as second argument to the function. The reason this was made an intial other colour was because people kept forgetting to upload the model and I wanted to heap on the cues.

amickan: Fair enough. I'll make that fix with the other improvements in the upcoming PR(s).

PR #4859, opened 2026-07-03
Add keep_alive action for algorithm endpoints

koopmant:

### DIFF
--- a/app/tests/algorithms_tests/test_api.py
+++ b/app/tests/algorithms_tests/test_api.py
@@ -803,11 +803,6 @@ class TestEndpointKeepAlive:
         assert response.status_code == 400, response.data
         assert response.json() == {"status": "Endpoint duration limit reached"}
 
-        endpoint.refresh_from_db()
-        assert (
-            endpoint.maximum_duration == endpoint.endpoint_utilization.duration
-        )
-
 
 @pytest.mark.django_db
 class TestInvocationList:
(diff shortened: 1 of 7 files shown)


app/tests/algorithms_tests/test_api.py
```diff
@@ -781,23 +783,54 @@ def test_post_not_allowed(self, client):
...
         client.force_login(user=endpoint.creator)
-        response = client.put(f"{self.url}{endpoint.pk}/", data={})
+        response = client.delete(f"{self.url}{endpoint.pk}/")
         assert response.status_code == status.HTTP_403_FORBIDDEN
 
-    def test_patch_not_allowed(self, client):
-        endpoint = EndpointFactory()
+
+@pytest.mark.django_db
+class TestEndpointKeepAlive:
```
amickan, line 794: Is there a reason that you can't use the `get_view_for_user()` helper method for these tests? That takes care of logging the user in, and of reversing the url from the view name.

amickan (approved):
LGTM, just a question about the tests.

app/tests/algorithms_tests/test_api.py
koopmant: Not really, I could have used that as well. I started using this after James added tests in this format at some point. The urls shouldn't change anyway, so I think it should not matter much if we reverse them or not.

amickan: Ok. Fine with me to leave them like this.

PR #4862, opened 2026-07-07
Extend endpoint maximum duration on invocation

koopmant:
Extend the maximum duration for an endpoint:
- when using the InvocationCreate REST API
- again when invoking the algorithm endpoint (to adjust for any time passed during provisioning)

If an endpoint is linked to a reader study, the maximum duration is not extended; lifetime management for these endpoints is linked to the workstation session.

### DIFF
--- a/app/grandchallenge/algorithms/serializers.py
+++ b/app/grandchallenge/algorithms/serializers.py
@@ -510,6 +510,11 @@ class InvocationPostSerializer(serializers.ModelSerializer):
             status=InvocationStatusChoices.VALIDATING_INPUTS,
         )
 
+        if not invocation.endpoint.is_linked_to_reader_study:
+            invocation.endpoint.keep_alive(
+                seconds=invocation.orchestrator.time_limit.total_seconds()
+            )
+
         try:
             invocation.process_civ_data_objects_and_execute_linked_task(
                 civ_data_objects=civ_data_objects,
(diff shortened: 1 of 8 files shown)


jmsmkn:
I feel like this is the third PR that touches the same lines of code. Not sure what I should review here?

koopmant:
> I feel like this is the third PR that touches the same lines of code. Not sure what I should review here?

There was some duplication in the change from:
```
def keep_alive(self, *, duration):
```
to 
```
def keep_alive(self, *, seconds):
```
Not sure if that's what you mean?

The keep_alive method was already added, this PR adds calling that method when invoking the endpoint.

app/grandchallenge/algorithms/serializers.py
```diff
@@ -413,6 +413,11 @@ def create(self, validated_data):
             status=InvocationStatusChoices.VALIDATING_INPUTS,
         )
 
+        if not invocation.endpoint.is_linked_to_reader_study:
+            invocation.endpoint.keep_alive(
+                duration=invocation.orchestrator.time_limit
+            )
+
```
koopmant: I just found out in testing that when creating invocations too quickly, this get a lock_timeout on the endpoint. Can I do something similar as we do in tasks with `check_lock_acquired`?

jmsmkn: Yes, as all requests are wrapped in transactions. But I'm not sure what should be done - you wouldn't want to wait for the lock, the user would anyway need to try again if you errored out.

koopmant: Indeed. I can just catch it and return a clear error message.

koopmant: The problem is locking the endpoint early during the transaction. I now moved it to after `process_civ_data_objects_and_execute_linked_task`. I could also do it after the transaction with `on_commit`, but I think this should remain atomic.

jmsmkn (approved):
LGTM!

PR #4863, opened 2026-07-07
Parametrize endpoint keep_alive action

koopmant:

### DIFF
--- a/app/grandchallenge/algorithms/serializers.py
+++ b/app/grandchallenge/algorithms/serializers.py
@@ -121,6 +121,15 @@ class AlgorithmModelSerializer(serializers.ModelSerializer):
         fields = ["pk", "algorithm", "created", "import_status", "model"]
 
 
+class KeepAliveSerializer(serializers.Serializer):
+    duration_minutes = serializers.IntegerField(
+        required=False,
+        default=5,
+        min_value=1,
+        max_value=15,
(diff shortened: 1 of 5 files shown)


koopmant:
I recognize this is not strictly necessary, but seems like a nice to have.

app/grandchallenge/algorithms/serializers.py
```diff
@@ -121,6 +121,15 @@ class Meta:
         fields = ["pk", "algorithm", "created", "import_status", "model"]
 
 
+class KeepAliveSerializer(serializers.Serializer):
+    duration_minutes = serializers.IntegerField(
```
jmsmkn, line 125: Why not use a duration field here? https://www.django-rest-framework.org/api-guide/fields/#durationfield

jmsmkn:
> I recognize this is not strictly necessary, but seems like a nice to have.

I'm not so sure, now it behaves differently to the workstation keep alive. I think that a consistent API is preferred.

app/grandchallenge/algorithms/serializers.py
koopmant: I figured an integerfield would be a bit more user friendly.

koopmant:
> I think that a consistent API is preferred.

All right, let's close this.

PR #4866, opened 2026-07-08
Add new field to display set data "description_safe"

pkcakeout:
See CIRRUS bug `#3991`.

We are trying to clearly mark fields that are "safely escaped html" with the `_safe` postfix in our json structures. This was not done for the description field, so this is fixing this problem.

Needs to be a two-step removal - in two months we can remove the old `description` field, when CIRRUS was updated twice.

### DIFF
--- a/app/grandchallenge/reader_studies/models.py
+++ b/app/grandchallenge/reader_studies/models.py
@@ -1083,6 +1083,11 @@ class DisplaySet(
 
     @property
     def description(self) -> str:
+        # Note: Remove this after August 2026
+        return self.description_safe
+
+    @property
+    def description_safe(self) -> str:
         case_text = self.reader_study.case_text
 
         if case_text:
--- a/app/grandchallenge/reader_studies/serializers.py
+++ b/app/grandchallenge/reader_studies/serializers.py
@@ -178,6 +178,7 @@ class DisplaySetSerializer(HyperlinkedModelSerializer):
             "optional_hanging_protocols",
             "view_content",
             "description",
+            "description_safe",
             "index",
         )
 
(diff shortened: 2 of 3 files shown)


app/grandchallenge/reader_studies/models.py
```diff
@@ -1083,6 +1083,11 @@ def workstation_url(self):
 
     @property
     def description(self) -> str:
+        # Note: Remove this after August 2026
```
chrisvanrun, line 1086: I suggest adding a release marker for CIRRUS, that makes it a bit clearer.

app/grandchallenge/reader_studies/models.py
```diff
@@ -1083,6 +1083,11 @@ def workstation_url(self):
...
+        return self.description_safe
+
+    @property
+    def description_safe(self) -> str:
```
jmsmkn, line 1090: Using string concatenation means that the output will be unsafe here (`str` + `SafeStr` = `str`). Can this method be reworked? Maybe:

```python

output = ""

if case_text:
            seen_names = set()

            for val in self.values.all():
                try:
                    name = val.image.name
                except AttributeError:
                    continue

                if name in case_text and name not in seen_names:
                    output += case_text[name]
                    seen_names.add(name)

return md2html(output)
```

Then it is obvious that the return value is safe.

jmsmkn: We can then also add a test that the property returns a `SafeStr`. Perhaps that should be done for all the other `_safe` attributes that are sent to CIRRUS.

pkcakeout: I thought that concatenation was a bit weird as well and actually did another version of this PR where I serialize a dict of the individually serialized `case_text` strings per image. That would allow for more templating options on the CIRRUS side as well. Would that be a good idea maybe as well? I dismissed it as "overcomplicated" - though if we want to rework this function anyways, this would likely be an option as well? I will open the PR for that branch!

pkcakeout: See #4867

jmsmkn:
Continued in #4867

PR #4867, opened 2026-07-09
Structured `descriptions_map_safe` for case texts.

pkcakeout:
Same as #4866 but with a structured map instead of a concatenated string. Opened this because of James' comment about the concatenation being messy/potentially unsafe. This avoids this because it serializes everything as a mapping, so that we get the safe strings as values inside the map.

### DIFF
--- a/app/grandchallenge/reader_studies/models.py
+++ b/app/grandchallenge/reader_studies/models.py
@@ -26,6 +26,7 @@ from django.db.models.functions import Coalesce
 from django.db.models.signals import post_delete
 from django.dispatch import receiver
 from django.utils.functional import cached_property
+from django.utils.safestring import SafeString
 from django_extensions.db.models import TitleSlugDescriptionModel
 from guardian.shortcuts import assign_perm, remove_perm
 from pictures.models import PictureField
@@ -1104,22 +1105,20 @@ class DisplaySet(
             return ""
 
     @property
-    def descriptions_map_safe(self) -> dict[str]:
+    def descriptions_map_safe(self) -> dict[str, SafeString]:
         case_text = self.reader_study.case_text
+        if not case_text:
+            return {}
+
         result = {}
-        if case_text:
-            seen_names = set()
-            output = ""
+        for val in self.values.all():
+            try:
+                name = val.image.name
+            except AttributeError:
+                continue
 
-            for val in self.values.all():
-                try:
-                    name = val.image.name
-                except AttributeError:
-                    continue
-
-                if name in case_text and name not in seen_names:
-                    seen_names.add(name)
-                    result[name] = md2html(case_text[name])
+            if name in case_text and name not in result:
+                result[name] = md2html(case_text[name])
         return result
 
     @property
--- a/app/grandchallenge/reader_studies/models.py
+++ b/app/grandchallenge/reader_studies/models.py
@@ -1104,6 +1104,25 @@ class DisplaySet(
             return ""
 
     @property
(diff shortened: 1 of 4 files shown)


app/grandchallenge/reader_studies/models.py
```diff
@@ -1103,6 +1103,25 @@ def description(self) -> str:
         else:
             return ""
 
+    @property
+    def descriptions_map_safe(self) -> dict[str]:
+        case_text = self.reader_study.case_text
+        result = {}
+        if case_text:
+            seen_names = set()
+            output = ""
```
jmsmkn: `output` is unused and `name not in seen_names` can be replaced with `name not in result` and `seen_names` removed.

Could also do with the tests for `SafeString`s being returned for both this and the other `_safe` attributes.

pkcakeout: Refactoring done, also added a model-test for the `_safe`-dict. I did not create anything of modify the existing `description`-field. Should I?

jmsmkn: That's fine. Was just thinking this would also be a useful test to have for the other `_safe` attributes, such as `help_text_safe`, `title_safe` etc etc.

app/grandchallenge/reader_studies/serializers.py
```diff
@@ -165,6 +167,15 @@ def get_index(self, obj) -> int | None:
     def get_title_safe(self, obj) -> str:
         return clean(obj.title, no_tags=True)
 
+    @extend_schema_field(
+        {
+            "type": "object",
+            "additionalProperties": {"type": "string"},
+        }
+    )
+    def get_descriptions_map_safe(self, obj) -> dict[str, str]:
```
pkcakeout, line 176: I am not sure about this. This _works_ but is annoying. Annotating strings via type-hints as "SafeString" makes `drf-spectacular` think that we want to serialize a custom type, which it correctly resolves, but spits out a warning about. We would need to annotate all `SafeString` types with a custom metadata... does this scale - I doubt it.

The most general way to fix this is, I think, is by attaching a custom serializer to the SafeString builtin type which then gets picked up by `drf-spectacular` so it knows how to handle a SafeString... but this PR is kind of spinning out of control at that point.

jmsmkn: This is how to do it. You could define a separate serializer instead (`DescriptionsMapSerializer`) but this would do the same thing in a roundabout way.

jmsmkn (approved):
LGTM!

app/grandchallenge/reader_studies/serializers.py
pkcakeout: Ok, thanks for confirming. I take it that since you approved the PR that this is ok to do it like this, so :shrug:

jmsmkn: Yes it's all much of a muchness.

You can merge this if you're happy with the data structure. Usually I would say that a list of objects would be better - with a map the meaning of the key is implicit (in this case it is the image name, not pk), same with the value in this case. With a list of objects you can assign meaning to the contents for communication, and it's easier to extend as you can add extra attributes in case you need them later, such as the image pk.

pkcakeout: > Usually I would say that a list of objects would be better

True, I actually considered that. However, in this case the current representation mirrors the `case_text`-data structure from the reader study configuration directly. Kind of nice to to simply re-use that representation, I think.

PR #4871, opened 2026-07-09
Cancel active invocations when stopping endpoint

koopmant:
Endpoints will stop regardless of any active invocations. Mark those invocations as cancelled. 

### DIFF
--- a/app/grandchallenge/components/tasks.py
+++ b/app/grandchallenge/components/tasks.py
@@ -1794,7 +1794,7 @@ def stop_endpoint(*, pk: str | UUID, app_label: str, model_name: str):
 
     endpoint.orchestrator.deprovision()
     endpoint.update_status(status=endpoint.StatusChoices.STOPPED)
-    cancel_related_invocations.execute_on_commit(
+    cancel_active_invocations.execute_on_commit(
         endpoint_pk=endpoint.pk, _delay=300
     )
 
@@ -1819,7 +1819,7 @@ def stop_expired_endpoints(*, app_label: str, model_name: str):
 
 
 @lambda_task
-def cancel_related_invocations(*, endpoint_pk: str | UUID):
+def cancel_active_invocations(*, endpoint_pk: str | UUID):
     from grandchallenge.algorithms.models import Invocation
 
     invocations = list(
--- a/app/grandchallenge/components/tasks.py
+++ b/app/grandchallenge/components/tasks.py
@@ -1818,7 +1818,7 @@ def stop_expired_endpoints(*, app_label: str, model_name: str):
         stop_endpoint.execute_on_commit(**endpoint.task_kwargs)
 
 
-@lambda_task(retry_on=(LockNotAcquiredException,))
+@lambda_task
 def cancel_related_invocations(*, endpoint_pk: str | UUID):
     from grandchallenge.algorithms.models import Invocation
 
@@ -1833,7 +1833,7 @@ def cancel_related_invocations(*, endpoint_pk: str | UUID):
     )
 
     if Invocation.objects.active().filter(endpoint=endpoint_pk).exists():
-        raise LockNotAcquiredException(
+        task_logger.error(
             "Some invocations were locked and could not be cancelled"
         )
 
--- a/app/grandchallenge/components/tasks.py
+++ b/app/grandchallenge/components/tasks.py
@@ -1818,7 +1818,7 @@ def stop_expired_endpoints(*, app_label: str, model_name: str):
         stop_endpoint.execute_on_commit(**endpoint.task_kwargs)
 
 
-@lambda_task
+@lambda_task(retry_on=(LockNotAcquiredException,))
 def cancel_active_invocations(*, endpoint_pk: str | UUID):
     from grandchallenge.algorithms.models import Invocation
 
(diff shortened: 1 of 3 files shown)


app/grandchallenge/components/tasks.py
```diff
@@ -1794,6 +1794,9 @@ def stop_endpoint(*, pk: str | UUID, app_label: str, model_name: str):
 
     endpoint.orchestrator.deprovision()
     endpoint.update_status(status=endpoint.StatusChoices.STOPPED)
+    cancel_active_invocations.execute_on_commit(
+        endpoint_pk=endpoint.pk, _delay=300
```
jmsmkn: The delay of 300 seems a little arbitrary. Maybe there is a better way of doing this? Don't we have event bridge events to see if the endpoint failed to start?  We could extend that to when it stops and cancel the invocations then.

Same with setting the status to STOPPED here, does it actually stop immediately?

app/grandchallenge/components/tasks.py
```diff
@@ -1815,6 +1818,26 @@ def stop_expired_endpoints(*, app_label: str, model_name: str):
...
+        .filter(endpoint=endpoint_pk)
+        .values_list("pk", flat=True)
+    )
+    Invocation.objects.filter(pk__in=invocations).update(
+        status=Invocation.StatusChoices.CANCELLED
+    )
+
+    if Invocation.objects.active().filter(endpoint=endpoint_pk).exists():
+        task_logger.error(
+            "Some invocations were locked and could not be cancelled"
```
jmsmkn: We'd then have to manually go in and cancel them, right? Maybe it would be better to use `check_lock_acquired` and retry on `LockNotAcquired` so that everything does get cancelled.

app/grandchallenge/components/tasks.py
koopmant: I'm not sure if could extend the status event handling for stopping (there is a DELETING status, but after that the endpoint is just deleted). When we set the status to stopped, the [delete_endpoint](https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_DeleteEndpoint.html) function has successfully returned, so the endpoint should already be deleted. 

The point of the delay is to allow for processing of the invocation on our side. Basically, if the endpoint just completed right before the endpoint is stopped, we still need to process the outputs. 300 seconds is probably way too much, but I figured it doesn't really matter.

app/grandchallenge/components/tasks.py
koopmant: Yes, I had that initially. But I was worried about many retries. Should I be worried about that? Trying this for 2880 times seems like a lot, although it seems unlikely to happen.

jmsmkn: It's a pattern we often use so is fine. Highest number so far is 24.

app/grandchallenge/components/tasks.py
jmsmkn: It would be best if the state on Grand Challenge accurately reflects what is running on AWS as we really need to know if endpoints are left running. You should be able to check the event bridge documentation to see what filters are available. If not, you can remove the status filter and see what is sent.

koopmant: Hmm, I see I misinterpreted the AWS documentation. Their wording threw me off. 

I'll add a filter for the `DELETING` status and start a task that repeatedly polls to see if the endpoint still exists, marking it as deleted when it's no longer found.

jmsmkn: I don't think polling is the way to go, it is better to work with events. Otherwise you can get rate limited and you are performing unnecessary work.

koopmant: There is no final event for when the endpoint is actually deleted. `DELETING` is the last event before that, which basically means `DeleteEndpoint` is in progress. So I think if we want to accurately reflect what's on AWS, we need to check if the endpoint is still there. In my tests so far, deleting has always been very quick. So we could also just check once, (say after 10 seconds?) and if it's still there we raise it for support to handle it?

jmsmkn: Ah that's annoying. We need something that is reliable and single alerts can be missed. Can we set a cloudwatch alarm somehow? You will anyway need to export statistics on how many endpoints we think are running for the dashboard, so we could look at the difference between that and the actual usage as a metric to monitor.

koopmant: Okay, I've created a separate issue for this: https://github.com/DIAGNijmegen/rse-grand-challenge/issues/4880

What is you take on the 300 seconds delay to allow output parsing for the just executed invocations?

app/grandchallenge/components/tasks.py
koopmant: Using `check_lock_acquired` and retry on `LockNotAcquired` now.

app/grandchallenge/components/tasks.py
jmsmkn: I think just kill it without delay as the endpoint stops immediately. The user would then need to make sure themselves that everything is done before they call stop, they can make the choice whether to wait or not. The task will cover anything which is locked for processing already.

koopmant: All right 👍

jmsmkn (approved):
LGTM!

PR #4872, opened 2026-07-10
Remove newsletter modal

jmsmkn:
Added in #2179 when the newsletter was introduced. Existing users have now had almost 5 years to sign up, if they've not done it already they're not going to do it. They have to make a decision about it when signing up anyway.


amickan (approved):
LGTM!

PR #4883, opened 2026-07-16
Refactor SaveFormInitMixin

### LINKED ISSUE #4267: `SaveFormInitMixin` does not work with dynamically created fields or custom layouts
With `SaveFormInitMixin` all dynamically created fields need to be present before calling `super().__init__`. Sometimes that is a problem, e.g. in #4257.

Additionally, if someone sets up a custom layout the button is then clobbered and the save blocking functionality does not work e.g. https://github.com/comic/grand-challenge.org/blob/86d9e564ce5b3680918218f34dc21e4e0d7f1327/app/grandchallenge/evaluation/forms.py#L168-L177

I don't know what the solution to this is - maybe a custom `FormHelper` that is then used in place of the mixin, but then you have issues with adding the JS required by the button. I guess we could include the JS in the base template.

amickan:
Replaces the eager field-snapshotting in `SaveFormInitMixin` with a `SaveFormHelper` class that generates its layout lazily at render time. Dynamic fields added after super().__init__() are now always included regardless of MRO ordering and without having to call `init_form_helper()` again.

Forms with custom layouts now no longer need to create a fresh `FormHelper`, which in turn means that the gc-disabled-after-submit attribute is kept. Moving the js to the base template means it works regardless of form media declarations. 

Closes #4267 

This PR has been implemented iteratively with kiro, all code has been reviewed by me.

### DIFF
--- a/app/grandchallenge/core/forms.py
+++ b/app/grandchallenge/core/forms.py
@@ -28,7 +28,7 @@ class SaveFormHelper(FormHelper):
     """
     - Sets the gc-disable-after-submit attribute on the form element
     - Generates a default layout lazily at render time
-      (rather than snapshotting fields are construction time)
+      (rather than snapshotting fields at construction time)
     """
 
     def __init__(self, form=None):
--- a/app/grandchallenge/core/forms.py
+++ b/app/grandchallenge/core/forms.py
@@ -26,14 +26,9 @@ class PhaseMixin:
 
 class SaveFormHelper(FormHelper):
     """
-    A FormHelper subclass that:
-      - Sets the gc-disable-after-submit attribute on the form element
-      - Generates a default layout (Fieldset + Save button) lazily at
-        render time from the form's current fields
-
-    The lazy layout means field ordering and dynamic field addition
-    work seamlessly — whatever fields exist when the form is rendered
-    will be included.
+    - Sets the gc-disable-after-submit attribute on the form element
+    - Generates a default layout lazily at render time
+      (rather than snapshotting fields are construction time)
     """
 
     def __init__(self, form=None):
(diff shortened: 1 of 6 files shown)


koopmant:
Couple of questions before I look at this
- Could you add a description to the PR? I've looked at the linked issue but it links to other issues and it's not super clear to me what the problem is that we're solving here. 
- I also see there are some docstrings, I have not read them in detail but are they necessary?
- Should this have a test to show what is fixed?

amickan:
>  Could you add a description to the PR? I've looked at the linked issue but it links to other issues and it's not super clear to me what the problem is that we're solving here.

Done!

>  I also see there are some docstrings, I have not read them in detail but are they necessary?

I have simplified them, and do find them useful, but I'm also fine with removing them entirely. 

> Should this have a test to show what is fixed?

I have added a few tests to illustrate what this solves.

app/grandchallenge/core/forms.py
```diff
@@ -24,38 +24,47 @@ def __init__(self, *args, phase, **kwargs):
...
     """
-    Mixin that adds some save features to a form via init:
-      - a 'Save' button
-      - disabling fieldsets after the form is submitted
-
-    When used in a form with dynamically created fields,
-    this mixin needs to be placed before any other mixins.
+    - Sets the gc-disable-after-submit attribute on the form element
+    - Generates a default layout lazily at render time
+      (rather than snapshotting fields are construction time)
```
koopmant:
```suggestion
      (rather than snapshotting fields at construction time)
```

koopmant (commented):
Small typo, but looks good otherwise!

koopmant:
> I have simplified them, and do find them useful, but I'm also fine with removing them entirely.

I agree, they are useful!

koopmant (approved):
LGTM!

PR #4885, opened 2026-07-17
Include get_file_name in evaluation template for json file sockets

### LINKED ISSUE #4884: Forge evaluation-method templates are missing get_file_name
Having an interface with non database stored and `Anything`-kind socket for algorithms should include `get_file_name` function (and include an example usage).

It is currently not included.

Readon is that its inclusion currently relies on `is_file_kind` which does not include the JSON sockets that are uploaded via a file upload.

Currently affects chimera-agent, primarily. As I think other active challenges mostly rely on images for identification.

amickan:
Closes #4884


chrisvanrun (approved):
LGTM!

PR #4886, opened 2026-07-17
Refactor endpoints REST API

koopmant:
Include the api_url in the response after creating an Endpoint for convenience and clarify the reponse for the keep_alive action when the endpoint is no longer active.


amickan (approved):
LGTM!

PR #4888, opened 2026-07-20
Add comment field for verifications

amickan:
As discussed during the RSE meeting, this adds a comment field to verifications for internal remarks.


chrisvanrun (commented):
Workflow wise, would it make sense to also allow the verification status to be editable in the detail view, that way we can reject or accept and update the comment in one go?

amickan:
> Workflow wise, would it make sense to also allow the verification status to be editable in the detail view, that way we can reject or accept and update the comment in one go?

I see why you would want that, but it's a bit tricky because verifying updates 2 fields on the model (verified_at and is_verified). You always want to set those together atomically, so I think the only way to do this via the change form would be a custom button for the admin action that we already have on the list view. But that would not solve the "adding a comment and verifying in one go" problem you mention. So overall, I'd say let's keep it as is.

chrisvanrun (approved):
LGTM! Let's see how this unpacks in practice!

PR #4889, opened 2026-07-20
Show description when visualizing Vega charts

chrisvanrun:
A request came from PRISMA challenge organizers: to add a description when showing Vega charts.

Adding a description in the chart itself is tedious: If it becomes longer than just a subtitle you have to manually cut lines and put it as marks. Instead, I opted to add a simple HTML tag instead to the civ rendering.

### Before

Screenshot 2026-07-20 at 16 56 45

### After
Screenshot 2026-07-20 at 17 00 16


chrisvanrun:
Huh, I totally missunderstood some core JSON schema workings here.

amickan (approved):

PR #4891, opened 2026-07-21
Chain output job output parsing

chrisvanrun:
Closes: https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/868

The increase from 4 minutes to 14 minutes runtime was insufficient. Logs indicate that each output files takes around 1m42s to process.

This PR, in the most safe way possible, splits the processing of job outputs from one task, to a chain of tasks.

### DIFF
--- a/app/grandchallenge/components/tasks.py
+++ b/app/grandchallenge/components/tasks.py
@@ -1045,10 +1045,7 @@ def handle_event(*, event: dict, backend: str):
 
 
 @lambda_task(
-    queue=LambdaTaskQueueChoices.MEM8G,
-    retry_on=(LockNotAcquiredException,),
-    soft_timeout=LONG_TASK_SOFT_TIMEOUT,
-    hard_timeout=LONG_TASK_HARD_TIMEOUT,
+    queue=LambdaTaskQueueChoices.MEM8G, retry_on=(LockNotAcquiredException,)
 )
 def parse_job_outputs(
     *,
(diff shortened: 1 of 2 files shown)


chrisvanrun:
This was tested on staging with an algorith that has 3 output sockets.

app/grandchallenge/components/tasks.py
```diff
@@ -1062,18 +1057,67 @@ def parse_job_outputs(
...
         raise RuntimeError("Job is not ready for output parsing")
 
     if job.outputs.exists():
         raise RuntimeError("Job already has outputs")
 
+    interface_pks = list(
+        job.output_interfaces.order_by("pk").values_list("pk", flat=True)
+    )
+
+    if not interface_pks:
```
koopmant: Technically, this cannot happen because an algorithm interface must always have an output. 

I'd say ignore this here. And raise an error in the other task if it is somehow called with an empty list of interface_pks.

koopmant (changes requested):
This needs a test. One other comment.

chrisvanrun:
> This needs a test. One other comment.

Is this not already covered via `algorithm_tests.test_tasks.test_algorithm`? Or are you requesting specific tests for this task that dives into the different error states?

koopmant:
> > This needs a test. One other comment.
> 
> Is this not already covered via `algorithm_tests.test_tasks.test_algorithm`? Or are you requesting specific tests for this task that dives into the different error states?

Indeed, the important tests already exist. I just meant the error states; or do you think that is overkill?

chrisvanrun:
> Indeed, the important tests already exist. I just meant the error states; or do you think that is overkill?

I'm a stickler for tests and this is prime logic in data processing. The specific state checks are primary defensive with respect to sanity checks. Still, I'll see about adding some additional tests that checks them specifically.

chrisvanrun:
@koopmant I've tried my hand at creating a test that checks the defensive error states. Because this is the tail of the job execution, there are plenty of states to mock for an executor. To simplify the tests I've opted to mock the executor where applicable.

koopmant (approved):
LGTM!

PR #4892, opened 2026-07-22
Remove external refs from chart schema

### LINKED ISSUE #4680: Chart data cannot be added to display sets via UI
**Describe the bug**
When creating or updating a display set, selecting a socket of kind `Chart` renders no usable widget to add the data.

**To Reproduce**
1. Go to the Cases of any reader study
2. Click on Edit display set or Add display set
3. Select a socket with kind `Chart`, for example ROC Curve or FROC curve. 
4. No widget is rendered.

**Expected behavior**
A widget to upload JSON data.

**Screenshots**

Image

**Additional context**
Sockets with kind `Chart` that expect a JSON file, for example the ECG socket, do show the correct (image select or image upload) widget. The difference being the setting for `store_in_database`.
Uploading the JSON data via gcapi does work, this only concerns upload via UI.

amickan:
Closes #4680

### DIFF
--- a/app/grandchallenge/components/forms.py
+++ b/app/grandchallenge/components/forms.py
@@ -43,7 +43,10 @@ from grandchallenge.components.models import (
     ComponentInterface,
     SourceChoices,
 )
-from grandchallenge.components.schemas import generate_component_json_schema
+from grandchallenge.components.schemas import (
+    generate_component_json_schema,
+    generate_widget_json_schema,
+)
 from grandchallenge.components.widgets import (
     FileSearchInputWidget,
     SearchSelect,
@@ -155,27 +158,6 @@ class FlexibleWidgetPrefixes(StrEnum):
     CHOICE = f"flexible_widget_choice{INTERFACE_FORM_FIELD_PREFIX}"
     UPLOAD = f"flexible_upload{INTERFACE_FORM_FIELD_PREFIX}"
     SEARCH = f"flexible_search{INTERFACE_FORM_FIELD_PREFIX}"
-
-
-def _strip_external_refs(*, schema):
-    """Return a copy of the schema with external $ref URIs removed from
-    definitions. The browser-side JSON editor cannot resolve external
-    URIs; full validation still happens server-side."""
-    return {
-        **schema,
-        "definitions": {
-            key: (
-                {
-                    k: v
-                    for k, v in definition.items()
-                    if k != "$ref" or not v.startswith("http")
-                }
-                if isinstance(definition, dict)
-                else definition
-            )
-            for key, definition in schema["definitions"].items()
-        },
-    }
 
 
 class InterfaceFormFieldsMixin:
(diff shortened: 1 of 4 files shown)


amickan:
Looks like server-side validation of charts is also broken (at least in development). Needs further investigation.

amickan:
> Looks like server-side validation of charts is also broken (at least in development). Needs further investigation.

Never mind. Server-side validation does work, I had just used non-conforming json to test things....

app/grandchallenge/components/forms.py
```diff
@@ -157,6 +157,27 @@ class FlexibleWidgetPrefixes(StrEnum):
     SEARCH = f"flexible_search{INTERFACE_FORM_FIELD_PREFIX}"
 
 
+def _strip_external_refs(*, schema):
```
chrisvanrun: Slight missmatch between the name, as it suggests that all external refs are removed.

I am not sure about this one to be honest. Is it a better fix to try to resolve all `$refs` here via the registry from `get_json_schema_registry`? Together with a tweak to `generate_component_json_schema` to have it generate a more dressed out version of the required definitions to ensure we don't do this resolving everytime?

amickan: Good idea to refine the name. 

RE implementation: I'm also not sure, but for charts specifically, my reasoning was that the vega schema is quite big and inlining it would bloat the definitions dictionary. I would expect page loading to be slower and I generally wonder how useful the schema would be when inlined. I don't expect people to be building their vega-lite chart on GC. I would assume that they just copy-paste the working json. Reserving the schema validation to server-side validation seemed like a pragmatic fix to me.

chrisvanrun: Would it make more sense to then only target Chart? Have another function that generated json for user widgets and add a filter that kicks out the Chart definitions. The current filter and removal seems to broad: we can easily comment on the chart filtering to rationalize it.

amickan: Looking at this again after a break, I agree with you. I have changed it to specifically handle chart interfaces. However, since external refs will always be an issue, I have added a test that checks for those in our interface schemas and that fails if there are external refs. That way future developers will be alerted when they add a new schema.

app/grandchallenge/components/schemas.py
```diff
@@ -775,3 +775,25 @@ def generate_component_json_schema(*, component_interface, required):
...
+    schema = generate_component_json_schema(
+        component_interface=component_interface,
+        required=required,
+    )
+
+    schema = {
+        **schema,
+        "definitions": {
+            **schema["definitions"],
+            "CHART": {"type": "object"},
```
chrisvanrun, line 796:
```suggestion
            # Vega-Lite charts reference an external schema, which does not work with the JSON editor
            "CHART": {"type": "object"},
```

Perhaps add an explicit comment to the addition?

amickan: There already is a comment in the docstring of the method. I will remove it there, otherwise it's duplicate.

chrisvanrun (approved):
LGTM!

PR #4893, opened 2026-07-22
Distinguish between panimg and dicom image imports on job progress page

### LINKED ISSUE #4453: Images import url on job progress page only links to upload session, not to dicom uploads
The link leads to `cases/uploads`, but if the image import is actually a dicom import, that isn't shown on `cases/uploads` but on `cases/dicom-uploads`.

amickan:
Closes #4453


koopmant (approved):
LGTM!

PR #4894, opened 2026-07-23
Deduplicate display set QS before shuffling

### LINKED ISSUE #4209: Display sets duplicated when shuffle hanging list is true and the user is both reader and editor
To reproduce:

- Create a reader study with shuffle_hanging_list False
- Add a user as both reader and editor
- Add display sets
- Go to https://grand-challenge.org/api/v1/reader-studies/display-sets/?reader_study=uuid
- Note that the number of display sets is correct
- Set shuffle_hanging_list to True
- Go to https://grand-challenge.org/api/v1/reader-studies/display-sets/?reader_study=uuid
- Note that the number of display sets has doubled, it should be the same

amickan:
Closes #4209


koopmant (approved):
LGTM!

PR #4895, opened 2026-07-23
Fix negative general credits when algorithm specific credits expire

### LINKED ISSUE #4132: Remaining general credits for user become negative
**Describe the bug**
When `AlgorithmUserCredits` have just expired, the remaining general credits for a user can be negative for up to a month. 

**To Reproduce**
A user needs to have spent AlgorithmUserCredits that have expired in the last month. 
When we calculate the remaining general credits, the sum of all spent credits last month is deducted with the user credits that are currently active. Since they have expired, currently active is zero. 

**Screenshots**

Image

**Additional context**

https://github.com/comic/grand-challenge.org/blob/868ea54c1e9b17fd9d43a11fab1fb4d8369b4157/app/grandchallenge/algorithms/models.py#L838-L853

amickan:
Closes #4132


koopmant (approved):
LGTM!

PR #4896, opened 2026-07-24
Fix admin pages for models, gt and component images

amickan:
The admin pages for component images, models and ground truths were inaccessible because including the image calls the `url` property on the file field to generate a clickable link. Direct links to files in protected storage are not allowed though, making the admin pages inaccessible. 

I don't think we ever need these links, so just excluding them from those pages seems good enough. 

https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/873#issuecomment-5060032863


chrisvanrun (approved):
LGTM!

PR #4897, opened 2026-07-24
Make job output parsing idempotent

chrisvanrun:
Slight rework of the tasks involved. Should now be idempotent in that calling it multiple times will not mess things up.

Related: https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/868

### DIFF
--- a/app/grandchallenge/components/tasks.py
+++ b/app/grandchallenge/components/tasks.py
@@ -1080,7 +1080,7 @@ def parse_job_outputs(
     queue=LambdaTaskQueueChoices.MEM8G,
     retry_on=(LockNotAcquiredException,),
 )
-def parse_singular_job_output(  # noqa: C901
+def parse_singular_job_output(
     *,
     job_pk: str | UUID,
     job_app_label: str,
--- a/app/grandchallenge/components/tasks.py
+++ b/app/grandchallenge/components/tasks.py
@@ -1057,14 +1057,14 @@ def parse_job_outputs(
     with check_lock_acquired():
         job = model.objects.select_for_update(nowait=True).get(pk=job_pk)
 
-    if job.status != job.PARSING:
-        if job.status not in (job.EXECUTED, job.SUCCESS):
-            raise RuntimeError("Job is not ready for output parsing")
-        else:
-            task_logger.info(
-                f"Job with pk={job_pk} already marked as completely parsed: reparsing"
-            )
-        job.update_status(status=job.PARSING)
+    if job.status in (job.SUCCESS, job.FAILURE, job.PARSING):
+        # Nothing to do
+        return
+
+    if job.status != job.EXECUTED:
+        raise RuntimeError("Job is not ready for output parsing")
+
+    job.update_status(status=job.PARSING)
 
     interface_pks = list(
         job.output_interfaces.order_by("pk").values_list("pk", flat=True)
@@ -1084,7 +1084,7 @@ def parse_job_outputs(
     queue=LambdaTaskQueueChoices.MEM8G,
     retry_on=(LockNotAcquiredException,),
 )
-def parse_singular_job_output(
+def parse_singular_job_output(  # noqa: C901
     *,
     job_pk: str | UUID,
     job_app_label: str,
@@ -1099,8 +1099,9 @@ def parse_singular_job_output(
     with check_lock_acquired():
         job = model.objects.select_for_update(nowait=True).get(pk=job_pk)
 
-    if job.status == job.SUCCESS:
-        return  # Nothing to do
+    if job.status in (job.SUCCESS, job.FAILURE):
+        return  # Nothing to do: reached a terminal point
+
     if job.status != job.PARSING:
         raise RuntimeError("Job is not in parsing state")
 
@@ -1121,27 +1122,7 @@ def parse_singular_job_output(
         task_logger.info(
             f"Output interface with pk={interface_pk} already parsed for job with pk={job_pk}"
         )
-    else:
-        if not _parse_output_interface(
-            job=job, backend=backend, interface=interface
-        ):
-            return  # marked as failed, stop here
-
-    if remaining_interface_pks:
-        parse_singular_job_output.execute_on_commit(
-            job_pk=job_pk,
-            job_app_label=job_app_label,
-            job_model_name=job_model_name,
-            backend=backend,
-            interface_pks=remaining_interface_pks,
-        )
-    else:
-        job.update_status(status=job.SUCCESS)
-
-
-def _parse_output_interface(*, job, backend, interface) -> bool:
-    """Parse a single output interface. Returns False if the job was
-    marked as failed (caller should stop), True otherwise."""
+        return  # Nothing to do here
 
     executor = job.get_executor(backend=backend)
 
@@ -1153,17 +1134,27 @@ def _parse_output_interface(*, job, backend, interface) -> bool:
             error_message=str(e),
             detailed_error_message=e.message_details,
         )
-        return False
+        return
     except Exception:
         job.update_status(
             status=job.FAILURE,
             error_message=SystemErrorMessages.UNEXPECTED_ERROR,
         )
         task_logger.error("Could not parse outputs", exc_info=True)
-        return False
+        return
+
+    job.outputs.add(*outputs)
+
+    if remaining_interface_pks:
+        parse_singular_job_output.execute_on_commit(
+            job_pk=job_pk,
+            job_app_label=job_app_label,
+            job_model_name=job_model_name,
+            backend=backend,
+            interface_pks=remaining_interface_pks,
+        )
     else:
-        job.outputs.add(*outputs)
-        return True
+        job.update_status(status=job.SUCCESS)
 
 
 @lambda_task(retry_on=(RetryStep,))
--- a/app/grandchallenge/components/tasks.py
+++ b/app/grandchallenge/components/tasks.py
@@ -1105,8 +1105,7 @@ def parse_singular_job_output(  # noqa: C901
     if job.status != job.PARSING:
         raise RuntimeError("Job is not in parsing state")
 
-    interface_pk = interface_pks[0]
-    remaining_interface_pks = interface_pks[1:]
+    interface_pk, *remaining_interface_pks = interface_pks
 
     interface_model = job.output_interfaces.model
     try:
--- a/app/grandchallenge/components/tasks.py
+++ b/app/grandchallenge/components/tasks.py
@@ -1057,17 +1057,18 @@ def parse_job_outputs(
     with check_lock_acquired():
         job = model.objects.select_for_update(nowait=True).get(pk=job_pk)
 
-    if job.status != job.EXECUTED:
-        raise RuntimeError("Job is not ready for output parsing")
-
-    if job.outputs.exists():
-        raise RuntimeError("Job already has outputs")
+    if job.status != job.PARSING:
+        if job.status not in (job.EXECUTED, job.SUCCESS):
+            raise RuntimeError("Job is not ready for output parsing")
+        else:
+            task_logger.info(
+                f"Job with pk={job_pk} already marked as completely parsed: reparsing"
+            )
+        job.update_status(status=job.PARSING)
 
     interface_pks = list(
         job.output_interfaces.order_by("pk").values_list("pk", flat=True)
     )
-
-    job.update_status(status=job.PARSING)
 
     # Kick off the parsing chain
     parse_singular_job_output.execute_on_commit(
@@ -1080,7 +1081,8 @@ def parse_job_outputs(
 
 
 @lambda_task(
-    queue=LambdaTaskQueueChoices.MEM8G, retry_on=(LockNotAcquiredException,)
+    queue=LambdaTaskQueueChoices.MEM8G,
+    retry_on=(LockNotAcquiredException,),
 )
 def parse_singular_job_output(
     *,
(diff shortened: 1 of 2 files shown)


app/grandchallenge/components/tasks.py
```diff
@@ -1057,18 +1057,19 @@ def parse_job_outputs(
...
-    if job.outputs.exists():
-        raise RuntimeError("Job already has outputs")
+    if job.status != job.PARSING:
+        if job.status not in (job.EXECUTED, job.SUCCESS):
+            raise RuntimeError("Job is not ready for output parsing")
+        else:
+            task_logger.info(
+                f"Job with pk={job_pk} already marked as completely parsed: reparsing"
+            )
+        job.update_status(status=job.PARSING)
```
amickan: Why is this necessary? Now successful jobs can be reparsed? I don't think we want that. 

>f"Job with pk={job_pk} already marked as completely parsed: reparsing"

And this is not correct if the status is "EXECUTED".

app/grandchallenge/components/tasks.py
```diff
@@ -1115,6 +1117,34 @@ def parse_singular_job_output(
         )
         return
 
+    if job.outputs.filter(interface=interface_pk).exists():
+        task_logger.info(
+            f"Output interface with pk={interface_pk} already parsed for job with pk={job_pk}"
+        )
```
koopmant, line 1119: I think we can just return at this point and leave the rest as it was, no? Either, 1) the previous job failed or succeeded, then the job is no longer in the parsing state and we never get here; or, 2) the job for this output has already run, so we stop here and don't do anything.

app/grandchallenge/components/tasks.py
```diff
@@ -1090,20 +1092,20 @@ def parse_singular_job_output(
...
 
+    model = apps.get_model(app_label=job_app_label, model_name=job_model_name)
     with check_lock_acquired():
         job = model.objects.select_for_update(nowait=True).get(pk=job_pk)
 
-    if not interface_pks:
-        raise RuntimeError("No output interfaces to parse")
-
+    if job.status == job.SUCCESS:
+        return  # Nothing to do
```
amickan: This should never happen. If it does we want to know about it, I think, so that we can investigate. So I would keep it at the status check right after this one.

app/tests/components_tests/test_tasks.py
```diff
@@ -2209,6 +2206,17 @@ def get_outputs(self, output_interfaces):
             interface_pks=[int_socket_1.pk, int_socket_2.pk, int_socket_3.pk],
         )
 
+    if test_idempotent:
+        with django_capture_on_commit_callbacks(execute=True):
+            parse_singular_job_output(
```
koopmant, line 2335: I think we can use mocker spy here to inspect that it never calls `executor.get_outputs`

app/grandchallenge/components/tasks.py
```diff
@@ -1115,6 +1117,34 @@ def parse_singular_job_output(
...
+            job_app_label=job_app_label,
+            job_model_name=job_model_name,
+            backend=backend,
+            interface_pks=remaining_interface_pks,
+        )
+    else:
+        job.update_status(status=job.SUCCESS)
+
+
+def _parse_output_interface(*, job, backend, interface) -> bool:
```
amickan: Why this separate function? All this does is get the executor. The interface is not used. And are you sure that this would fail if the job was marked as failed?

amickan: Oh wait, I'm reading the change log wrong.

amickan: I understand what's happening now, but I'd suggest inlining this into `parse_singular_job_output`. It's only called from one place, and the boolean return value that silently control the caller's flow is easy to miss. Inlining makes the failure handling visible where it happens rather than requiring a jump to a separate function to understand what's going on.

app/tests/components_tests/test_tasks.py
```diff
@@ -2152,19 +2157,11 @@ def test_parse_job_outputs_incorrect_state(
...
 
-    job.status = job.EXECUTED
-    job.save()
-
-    job.outputs.add(ComponentInterfaceValueFactory())
-
-    with pytest.raises(RuntimeError, match="Job already has outputs"):
-        with django_capture_on_commit_callbacks(execute=True):
-            parse_job_outputs(**job.task_kwargs)
-
```
amickan, line 2163: Why was this removed?

app/tests/components_tests/test_tasks.py
```diff
@@ -2081,8 +2081,9 @@ def test_invoke_endpoint_skips_keep_alive_for_reader_study_endpoint(mocker):
 
 
 @pytest.mark.django_db
+@pytest.mark.parametrize("test_idempotent", [True, False])
```
amickan: I think it would be clearer to write a separate test for idempotency.

app/tests/components_tests/test_tasks.py
```diff
@@ -2209,6 +2206,17 @@ def get_outputs(self, output_interfaces):
...
+        with django_capture_on_commit_callbacks(execute=True):
+            parse_singular_job_output(
+                **job.task_kwargs,
+                interface_pks=[
+                    int_socket_1.pk,
+                    int_socket_2.pk,
+                    int_socket_3.pk,
+                ],
+            )
+
```
amickan, line 2332: The idempotency test should check that no duplicate CIVs were created.

app/grandchallenge/components/tasks.py
amickan: Agreed. My comment below aims at the same.

app/tests/components_tests/test_tasks.py
chrisvanrun: Is that not already the case below, where we check the number of outputs? Or do you mean in general, the total number of CIVs present in the database.

amickan: Yes, I meant in the database. Something like `ComponentInterfaceValue.objects.filter(interface=...).count()==1`

app/tests/components_tests/test_tasks.py
chrisvanrun: No longer applies, if the job already has outputs: we still want to check if each of the interfaces is actually there.

app/grandchallenge/components/tasks.py
chrisvanrun: You are right, we should only handle the expected states.

app/grandchallenge/components/tasks.py
chrisvanrun: If this is re-ran and we are in one of the terminal states that we get in once we finish parsing we can just silently let this go I think? At least, that is what I understand under idempotency.

app/grandchallenge/components/tasks.py
chrisvanrun: Ok!

app/grandchallenge/components/tasks.py
chrisvanrun: It was an attempt to make the function less complex (as it is currently not passing the linter). However, I think it is a bit of a false negative since there are some guardian checks which are pretty simple. I'll flag it as to be ignored.

app/tests/components_tests/test_tasks.py
chrisvanrun: That would just involve downloading things right? How would that be different from checking no additional CIVs are created?

app/grandchallenge/components/tasks.py
```diff
@@ -1057,18 +1057,19 @@ def parse_job_outputs(
     with check_lock_acquired():
         job = model.objects.select_for_update(nowait=True).get(pk=job_pk)
 
+    if job.status in (job.SUCCESS, job.FAILURE, job.PARSING):
+        # Nothing to do
+        return
+
     if job.status != job.EXECUTED:
         raise RuntimeError("Job is not ready for output parsing")
```
koopmant: I think the idea is that we don't expect this task to be called with a different state. We only defend in case it happens, so the database is not affected. So raising the runtime error when status is not `executed` is correct. Silently returning on success, failure and parsing is not. We only silently return when it's expected that a task can be called multiple times with the same state.

app/grandchallenge/components/tasks.py
```diff
@@ -1080,30 +1081,32 @@ def parse_job_outputs(
...
+        raise RuntimeError("No output interfaces to parse")
 
+    model = apps.get_model(app_label=job_app_label, model_name=job_model_name)
     with check_lock_acquired():
         job = model.objects.select_for_update(nowait=True).get(pk=job_pk)
 
-    if not interface_pks:
-        raise RuntimeError("No output interfaces to parse")
+    if job.status in (job.SUCCESS, job.FAILURE):
+        return  # Nothing to do: reached a terminal point
```
koopmant: Same comment as above; should not silently return, let the runtime error be raised below.

app/grandchallenge/components/tasks.py
```diff
@@ -1080,30 +1081,32 @@ def parse_job_outputs(
...
 
     if job.status != job.PARSING:
         raise RuntimeError("Job is not in parsing state")
 
-    executor = job.get_executor(backend=backend)
-
-    interface_pk = interface_pks.pop(0)
+    interface_pk = interface_pks[0]
+    remaining_interface_pks = interface_pks[1:]
```
koopmant: Do you know if this is needed now? I'd assume with the arguments being passed between two independent tasks, the `interface_pks` is a new object when the task starts. So `.pop()` would be just fine?

app/tests/components_tests/test_tasks.py
koopmant: Yes, you're right. It would be too close to the implementation.

app/grandchallenge/components/tasks.py
koopmant: See my other two comments. If we get into a task with a state we do not expect, we want to be informed.

koopmant (changes requested):
Small issue and a question

app/grandchallenge/components/tasks.py
chrisvanrun: What is your idea of idempotency then? I thought the idea was that you could call a task multiple times and not have it return a failure (or leave a bad state).

app/grandchallenge/components/tasks.py
chrisvanrun: I agree that it is not needed to protect for accidental mutation! However, I added it in this format because I prefer to reference the remaining as `remaining_interface_pks` going forward. Ofc, you could call it as such:

```Python
interface_pk = interface_pks.pop(0)
remaining_interface_pks = intererface_pks
```

But I thought the used annotation was a bit more elegant. I found an even more elegant way btw:

`interface_pk, *remaining_interface_pks = interface_pks`

Is that even better? I might be biased because automatic unpacking is used a lot in JS.

app/grandchallenge/components/tasks.py
chrisvanrun: Coding guidelines (albeit this refers Celery still) has it as follows:
> Be idempotent – they should be able to run twice without any issues 

Raising runtime errors the second time around seems to go against that?

koopmant: I believe it is more about the bad state than raising a failure, although I do remember an extra detail now. 

This is how James explained it to me: 
If we expect that the task could be called with a certain or some states, we log that it was called again and then return without raising an error.
If a task is called and the state is unexpected we raise an error. 

So it depends on whether you expect this task to be called again with these states, and I think the answer is no?

But otherwise we could use the task logger so we are still informed.

app/grandchallenge/components/tasks.py
koopmant: Ah okay. On elegance, it's not something we see often, but I have to admit it is kind of neat. However you prefer is fine with me!

chrisvanrun: Rule of cool!

chrisvanrun:
Rerunning tests because of flaky test: https://github.com/DIAGNijmegen/rse-grand-challenge/issues/4903

app/grandchallenge/components/tasks.py
chrisvanrun: Had a small sync discussion about this. We've reached the conclusion that throwing runtime errors is fine: if we see this happen a lot we'll make the choice (at a later stage) to either systemically ignore it via an implemented silent escape or keep it in as a sanity-check system to register WHY tasks are beings scheduled when unexpected.

app/grandchallenge/components/tasks.py
```diff
@@ -1080,30 +1077,28 @@ def parse_job_outputs(
 
 
 @lambda_task(
-    queue=LambdaTaskQueueChoices.MEM8G, retry_on=(LockNotAcquiredException,)
+    queue=LambdaTaskQueueChoices.MEM8G,
+    retry_on=(LockNotAcquiredException,),
 )
-def parse_singular_job_output(
+def parse_singular_job_output(  # noqa: C901
```
koopmant: Is C901 still applicable?

koopmant (dismissed):
LGTM, just a question

app/grandchallenge/components/tasks.py
chrisvanrun: Nope! Good catch!

app/grandchallenge/components/tasks.py
```diff
@@ -1060,15 +1060,12 @@ def parse_job_outputs(
     if job.status != job.EXECUTED:
         raise RuntimeError("Job is not ready for output parsing")
 
-    if job.outputs.exists():
-        raise RuntimeError("Job already has outputs")
```
amickan, line 1064: I think this would be good to leave in since this task should only be called when there are no outputs yet, right?

amickan:
One more comment / question, otherwise this looks good to me.

app/grandchallenge/components/tasks.py
koopmant: Yes, indeed. I think this task should not change. The status update to `PARSING` should also remain.

chrisvanrun: Check!

amickan (approved):

PR #4899, opened 2026-07-27
Add check for required objects on notifications

koopmant:


app/tests/notifications_tests/test_views.py
```diff
@@ -389,6 +389,42 @@ def test_follow_deletion_through_api(client):
...
+    [
+        Notification.Type.FORUM_POST,
+        Notification.Type.FORUM_POST_REPLY,
+        Notification.Type.ACCESS_REQUEST,
+        Notification.Type.REQUEST_UPDATE,
+        Notification.Type.NEW_ADMIN,
+        Notification.Type.EVALUATION_STATUS,
+        Notification.Type.MISSING_METHOD,
+        Notification.Type.IMAGE_IMPORT_STATUS,
+    ],
```
koopmant: Not sure why we are skipping `JOB_STATUS`, `FILE_COPY_STATUS` and `CIV_VALIDATION`... I think I'll add those here and in the property.

amickan: No good reason, just brain fog when I wrote this test ;-)

koopmant: All right, thanks. Added those.

app/grandchallenge/notifications/models.py
```diff
@@ -242,7 +242,52 @@ def get_receivers(*, kind, actor, action_object, target):  # noqa: C901
...
+            return self.target is not None
+        elif self.type == NotificationTypeChoices.EVALUATION_STATUS:
+            return (
+                self.action_object is not None
+                and self.target is not None
+                and getattr(self.action_object, "submission", None) is not None
+                and getattr(self.target, "challenge", None) is not None
+            )
+        elif self.type == NotificationTypeChoices.JOB_STATUS:
+            return self.actor is not None and self.target is not None
```
chrisvanrun:
```suggestion
            return self.target is not None
```

JOB_STATUS has handling of not existing actors

chrisvanrun (changes requested):
Small tibit for the JOB_STATUS case.

app/grandchallenge/notifications/models.py
```diff
@@ -269,8 +269,15 @@ def _required_objects_exist(self) -> bool:
...
             )
+        elif self.type == NotificationTypeChoices.JOB_STATUS:
+            return self.actor is not None and self.target is not None
         elif self.type == NotificationTypeChoices.IMAGE_IMPORT_STATUS:
             return self.action_object is not None
+        elif self.type in (
+            NotificationTypeChoices.FILE_COPY_STATUS,
+            NotificationTypeChoices.CIV_VALIDATION,
+        ):
+            return self.actor is not None
```
chrisvanrun: Does this require an actor?

koopmant: I think so. The actor is the receiver.

chrisvanrun: It's a bit of a spagetti to be honest, but I don't think we've seen errors about other functions than print_notification erroring out. Not sure if we then need to also cover the sending of the notifications: just the list view.

I'd rather not send the email if we need the actor and it is missing.

koopmant: You're right, for the failure we are seeing, this is not necessary. I wonder if I should change the approach to EAFP. I have a changeset ready for that.

chrisvanrun: I think the only proper way to do that is to overwrite the `GenericForeignKey` behaviour with respect to returning None values. A simple shadow that throws an ObjectNotFound when the super() call returns None?

What do you think?

chrisvanrun: I think your current approach is too broad with the exception.

koopmant: > I think the only proper way to do that is to overwrite the `GenericForeignKey` behaviour with respect to returning None values. A simple shadow that throws an ObjectNotFound when the super() call returns None?
> 
> What do you think?

Sorry, I don't understand what you're saying here.

koopmant: > I think your current approach is too broad with the exception.

Okay, I'll revert to the property. But I'll keep the test checking all notification types and only test the 200 response.

koopmant: We discussed synchronously. Agreed to keep Anne's original property. I just simplified the test to check the view is working for all notification types.

app/tests/notifications_tests/test_views.py
```diff
@@ -389,6 +392,28 @@ def test_follow_deletion_through_api(client):
     assert len(Follow.objects.all()) == num_follows - 1
 
 
+@pytest.mark.parametrize("notification_type", NotificationTypeChoices)
+@pytest.mark.django_db
+def test_notification_list_with_deleted_attributes(client, notification_type):
```
chrisvanrun, line 397: I had a slight alternative here that specifically tests the field per type of notification. What do you think?

```Python
from tests.notifications_tests.test_signals import (
    MODELS_FOR_NOTIFICATIONS_CLEANUP,
)


@pytest.mark.django_db
@pytest.mark.parametrize(
    "model, object_reference, kind, factories",
    MODELS_FOR_NOTIFICATIONS_CLEANUP,
)
def test_print_notification_missing_reference(
    model,
    object_reference,
    kind,
    factories,
    settings,
):
    settings.LAMBDA_TASKS_EAGER = True

    u = UserFactory()

    Notification.objects.all().delete()

    notification_kwargs = {
        "user": u,
        "type": kind,
        "message": "foo",
        "actor": None,
        "target": None,
        "action_object": None,
    }

    object_kwargs = {
        k: v["factory"](**v["kwargs"]) for k, v in factories.items()
    }
    notification_kwargs.update(object_kwargs)
    notification = NotificationFactory(**notification_kwargs)

    assert getattr(notification, object_reference) is not None

    notification.print_notification(user=u)  # Should not raise an exception

    setattr(notification, object_reference, None)
    notification.save()

    notification.refresh_from_db()

    assert getattr(notification, object_reference) is None

    printed_notification = notification.print_notification(user=u)
    if printed_notification:
        assert (
            "This notification refers to an object that has been deleted."
            in printed_notification
        ), f"Expected message not found in printed notification for model {model}"

```

koopmant: seems a bit complicated to be honest.

chrisvanrun: The setup was a copy from the signal tests, but the view test is likely cleaner and a more secure test.

chrisvanrun (approved):
LGTM!

PR #4900, opened 2026-07-27
Missing gfk filter for notifications

amickan:
Adds a filter to identify orphaned notifications, see https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/876

### DIFF
--- a/app/grandchallenge/notifications/admin.py
+++ b/app/grandchallenge/notifications/admin.py
@@ -1,5 +1,6 @@
 from actstream.models import Follow
 from django.contrib import admin
+from django.contrib.contenttypes.models import ContentType
 
 from grandchallenge.core.admin import (
     GroupObjectPermissionAdmin,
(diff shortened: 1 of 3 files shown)


app/grandchallenge/notifications/admin.py
```diff
@@ -14,6 +14,27 @@
...
+        ]
+
+    def queryset(self, request, queryset):
+        if self.value() != "1":
+            return queryset
+
+        pks = [
+            notification.pk
+            for notification in queryset.iterator()
+            if not notification._required_objects_exist
```
koopmant: Not sure how performant this is. Would a direct filter be much faster?

amickan: A direct query does not work here because of the GFKs. I do share your concerns about performance. I have pushed an alternative, which I think would be more efficient - thanks, Kiro ;-)

amickan: Downside is that we now have 2 logics in place for checking for orphaned notifications. I wanted to keep the logic in one place, but I think both fixes are temporary anyway, so fine IMO.

app/grandchallenge/notifications/admin.py
```diff
@@ -14,6 +15,96 @@
...
+                .values_list(ct_field, flat=True)
+                .distinct()
+            )
+
+        content_types_by_id = {
+            ct.pk: ct for ct in ContentType.objects.filter(pk__in=all_ct_ids)
+        }
+
+        for ct_field, id_field in gfk_fields:
+            for content_type in content_types_by_id.values():
```
koopmant: maybe I'm missing something, but if we don't look up by pk, then why do we need the dict?

koopmant (commented):
Wow, I find this really hard to follow now. If you say it works, I believe you. Had a question though.

app/grandchallenge/notifications/admin.py
amickan: Good point. You're right. Oversight on my part when I was editing what Kiro spit out. I've simplified it to a list.

amickan:
> Wow, I find this really hard to follow now. If you say it works, I believe you. Had a question though.

I agree. I hope the comments help a bit though? I think the filter should anyway be removed after we've used it and have fixed the underlying problem.

koopmant:
> I think the filter should anyway be removed after we've used it and have fixed the underlying problem.

Agreed

koopmant (approved):
LGTM

PR #4902, opened 2026-07-28
Handle deleted forum objects when notifying

chrisvanrun:
Relates to: https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/876


chrisvanrun:
~Still need to add tests here.~ Done!

chrisvanrun:
TBH, I am still in doubt about when to throw errors in a task. For now: I've let the task raise errors when the object in question has been deleted (either early, or late at time of locking.).

chrisvanrun:
Quick side note: I double checked all other locations that `Notification.send` is used: all objects related to those aren't often deleted (except for some shell action or otherwise site-admin based clean ups).

For CIVSets, it's always the holder of the uploaded files that is used and not the actual CIVSets.

As such, I think this is the only place where this really needs to be deployed.

app/grandchallenge/discussion_forums/tasks.py
```diff
@@ -63,3 +66,8 @@ def create_forum_notifications(
             action_object=obj,
             target=obj.forum,
         )
+
+    # To prevent orphaned notifications we do a final check on the existence
+    # of the action object and lock it for the remainder of the transaction.
```
koopmant, line 71: The part about the "remainder of the transaction" is a bit confusing, since there is nothing after this.

app/grandchallenge/discussion_forums/tasks.py
```diff
@@ -63,3 +66,8 @@ def create_forum_notifications(
...
+    with check_lock_acquired():
+        model.objects.select_for_update(nowait=True).get(pk=object_pk)
```
koopmant, line 73: What happens if we cannot acquire the lock for some reason and this task is retried? Have the emails already been sent at this point? Because then they would be sent again...

app/grandchallenge/discussion_forums/tasks.py
```diff
@@ -30,6 +31,8 @@ def create_forum_notifications(
         )
         return
 
+    # Do not lock here as notifications (and emails) might take a while to be created,
+    # rather, we apply a lock at the end of this task.
```
koopmant, line 35: Might it be worthwhile to schedule a separate task to send the emails, so we can simply lock the object for this task? I know that is something we do in other tasks.

koopmant (commented):
Some questions

app/grandchallenge/discussion_forums/tasks.py
chrisvanrun: The transaction still has to be finished right? Does that not fall under the transaction itself? Maybe a better term would be "until the transaction is applied".

app/grandchallenge/discussion_forums/tasks.py
chrisvanrun: AFAIK emails are also database based, as such they'd rollback.

app/grandchallenge/discussion_forums/tasks.py
chrisvanrun: There are several things that need to happen prior to emails. I don't think we can create bulk notifications: needs permission setting and all. Also, the work is undetermined as the number of participants are not fixed. As such, we should lock smartly anyway.

app/grandchallenge/discussion_forums/tasks.py
amickan: The emails are stored in the DB, but it looks like they get dispatched immediately once created, and that would still happen even if you rollback the DB changes, so I think you'd end up with duplicate emails. A test for this would clarify whether that's the case.

app/grandchallenge/discussion_forums/tasks.py
koopmant: I was just a bit puzzled. Bit indeed, it is not wrong.

app/grandchallenge/discussion_forums/tasks.py
chrisvanrun: Ok - emails ARE being sent (there is a second way of sending emails, which IS database blocked. But this is a direct email sending call. Currently working on an alternative test + implementation.

chrisvanrun: I initially had the email sending functions wrapping in an on_commit. However, I was afraid this would be counted as a spam action. The original task had an object creation and a datatable-row lock in between each of the calls for emailing. Caching the emailing all to a single on_commit moment would have likely led to spam. So instead I rewrote it to use a lambda task instead for sending the emails.

However, I now realize that the normal function `send_unread_notification_emails` has done this exact thing for the longest time: so maybe the rate was OK... so instead of a single task per notification: I've created a batch task for instant email notifications but one that includes locking as it'll be fired off a lot more than the regular daily-summary task.

chrisvanrun:
The tests are failing because of the flaky test (fixed on main). Rebasing to get things to pass on the current branch.

Edit: Done.

app/grandchallenge/notifications/tasks.py
```diff
@@ -17,23 +18,33 @@ def send_unread_notification_emails():
...
+
+
+@lambda_task(retry_on=(LockNotAcquiredException,))
+def send_unread_notification_instant_emails(*, user_profile_ids: list[int]):
+    site = Site.objects.get_current()
+
+    with check_lock_acquired():
+        UserProfile.objects.select_for_update(nowait=True).filter(
+            pk__in=user_profile_ids
+        )
```
amickan, line 41: Does this locking work? The queryset here is not actually evaluated, so I would assume the rows are never locked either.

amickan: And then below you're re-doing the query. I think you need to join the two queries, and then iterate over the queryset further below for dispatching.

chrisvanrun: Ah good point. I should prob add a dedicated test for this!

chrisvanrun: Darn, I tried to write a test to double check if it is correctly locked: that had me going down the rabbit hole of transactions, DB connections and trying to write a test for that. Imho, way to complicated to pull of and likely the reason we don't regularly test this.

koopmant: > Imho, way to complicated to pull of and likely the reason we don't regularly test this.

Yes, been there...

app/grandchallenge/notifications/tasks.py
```diff
@@ -17,23 +18,33 @@ def send_unread_notification_emails():
...
+        UserProfile.objects.select_for_update(nowait=True).filter(
+            pk__in=user_profile_ids
+        ).exists()  # Exists resolves the queryset
+
+    profiles = (
+        UserProfile.objects.filter(
+            pk__in=user_profile_ids,
+            notification_email_choice=NotificationEmailOptions.INSTANT,
+            user__is_active=True,
         )
```
amickan, line 48: You're still doing partially the same query twice. Why not combine them into one?

amickan: I also think that using exists() would not lock all rows, but only the first match.

chrisvanrun: > You're still doing partially the same query twice. Why not combine them into one?

Reason is the required `DISTINCT` in the second one, and locking does not combine with that.

chrisvanrun: > I also think that using exists() would not lock all rows, but only the first match.

Ah, that is just cheating! You are fully right, it applies a `LIMIT` and that only locks the hit rows.

amickan: > Reason is the required `DISTINCT` in the second one, and locking does not combine with that.

Ah it looked like that was gone. Ok, maybe `values_list` to get the pks of the locked rows?

chrisvanrun: Yes! I had added that as well: and a `len` so nothing is pushed into models.

koopmant (approved):
LGTM!

PR #4904, opened 2026-07-28
Fix flaky display-set index test

### LINKED ISSUE #4903: Extra flaky test: `test_display_set_index`
Even re-running it trice has it fail sometimes (seen it a few times now in a few unrelated PRs).

For instance in this run: https://github.com/DIAGNijmegen/rse-grand-challenge/actions/runs/30341850106/job/90218992879

chrisvanrun:
Closes: #4903

In short: because of _subtle_ differences between the randomization implementation and the test implementation of the shuffeling differed the execution plan (theory) sometimes switched. Subtle in that adding a filter on reader_study has influence. Perhaps a better approach is extracting the shuffeling and re-use it in the test. 

There are cons to that though, and because I want to do other things I think it is fine to leave it as is?

### DIFF
--- a/app/tests/reader_studies_tests/test_api.py
+++ b/app/tests/reader_studies_tests/test_api.py
@@ -2088,13 +2088,16 @@ def test_display_sets_shuffled_per_user(client):
     ]
 
 
-@pytest.mark.flaky(reruns=3)
 @pytest.mark.django_db
 def test_display_set_index(client):
     n_display_sets = 10
     reader_study = ReaderStudyFactory()
     user = UserFactory()
     reader_study.add_reader(user)
+
+    assert not DisplaySet.objects.filter(
+        reader_study=reader_study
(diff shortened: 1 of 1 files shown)


koopmant:
What is the desired behavior and what is the test trying to prove? When the view is retrieved again, the shuffled order should be the same? Maybe we can get the expected index from `shuffled_order` in the test?

chrisvanrun:
> `shuffled_order`

Mhh, that might work. I THINK it is trying to determine if the detail retrieval' `index` follows the general list retrieval' `index`. Given that, we can indeed source it from the list retrieval. On it.

chrisvanrun:
~Ghuh, I might actually have found a bug here =/. Parking this until I can properly fix this.~

Nope, just me incorrectly applying a fixed seed.

koopmant (approved):
LGTM!

PR #4910, opened 2026-07-30
Match endpoints active on SageMaker to endpoints active in database

koopmant:
See https://github.com/DIAGNijmegen/rse-grand-challenge/issues/4880

### DIFF
--- a/app/grandchallenge/core/tasks.py
+++ b/app/grandchallenge/core/tasks.py
@@ -118,9 +118,9 @@ def _get_metrics():
 
     metric_data.append(
         {
-            "MetricName": "EndpointsOnSagemaker",
+            "MetricName": "LeftoverEndpointsOnSagemaker",
             "Dimensions": [{"Name": "Model", "Value": Endpoint.__name__}],
-            "Value": _get_endpoints_count_from_sagemaker(),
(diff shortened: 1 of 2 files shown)


koopmant:
I've kept it included in the `put_cloudwatch_metrics` task instead of creating a separate task. I think there is a small risk of the task failing when the boto client gives transient errors, so I'm going to turn it into a separate task.

amickan:
> I think there is a small risk of the task failing when the boto client gives transient errors, so I'm going to turn it into a separate task.

That is the case for the `put_cloudwatch_metrics` task as well though, right? They both use the boto client, or what am I missing?

amickan (approved):
LGTM! Just a question.

koopmant:
> > I think there is a small risk of the task failing when the boto client gives transient errors, so I'm going to turn it into a separate task.
> 
> That is the case for the `put_cloudwatch_metrics` task as well though, right? They both use the boto client, or what am I missing?

Yes, that's true. If `PutMetricData` gives a client error the task will also fail.  But now it's an additional client for the sagemaker service. It's probably overly cautious, but I didn't want to risk failing the entire task when just `ListEndpoints` fails, because then we are missing all metrics.

PR #4916, opened 2026-08-04
Limit requested EBS volume size

jmsmkn:
Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/874


app/config/settings.py
```diff
@@ -960,6 +960,9 @@ def get_private_ip():
 COMPONENTS_MAXIMUM_IMAGE_SIZE = 10 * GIGABYTE
 COMPONENTS_MINIMUM_JOB_DURATION = 5 * 60  # 5 minutes
 COMPONENTS_MAXIMUM_JOB_DURATION = 24 * 60 * 60  # 24 hours
+COMPONENTS_EBS_VOLUME_SIZE_LIMIT = int(
```
koopmant:
```suggestion
COMPONENTS_EBS_VOLUME_SIZE_LIMIT_GB = int(
```

koopmant:
Looks good to me. Not sure if it closes 874 though. Shouldn't we fail the job when we hit a `botocore.errorfactory.ResourceLimitExceeded` error?

jmsmkn:
> Shouldn't we fail the job when we hit a botocore.errorfactory.ResourceLimitExceeded error?

No, as the job should be retried until resources (GPU instances) are available.

koopmant (approved):
LGTM!

Issue #4918, opened 2026-08-04
Use fan-out rather than chaining for output parsing


jmsmkn:
Output parsing was changed to use a single task per output, which are chained. This causes increased delays. It would be better to do this work in parallel, like what is done for the input validation.

Additionally, the intermediate `parse_job_outputs` can be removed with 

https://github.com/DIAGNijmegen/grand-challenge/blob/a0ed03e8d2b78b7b5b055d7e70de7f39a7d8cfc7/app/grandchallenge/components/tasks.py#L1066-L1079

pushed up to the success path of `handle_event`

https://github.com/DIAGNijmegen/grand-challenge/blob/a0ed03e8d2b78b7b5b055d7e70de7f39a7d8cfc7/app/grandchallenge/components/tasks.py#L1040-L1044

PR #4919, opened 2026-08-05
Skip intermediate `parse_job_outputs`

jmsmkn:
This PR removes calling an intermediate lambda task. `parse_job_outputs` can then be removed from the codebase once deployed.

Also from looking at the code I saw that internal errors were being attached to the job objects which are then visible to the user, so these were changed to `SystemErrorMessages.UNEXPECTED_ERROR`. I also updated the internal API so that the caller is now responsible for iterating over the outputs, rather than sending a list with one entry.

See #4918

### DIFF
--- a/app/grandchallenge/components/tasks.py
+++ b/app/grandchallenge/components/tasks.py
@@ -1117,25 +1117,33 @@ def parse_singular_job_output(
     except ComponentInterface.DoesNotExist:
         job.update_status(
             status=job.FAILURE,
-            error_message=f"Output interface with pk={interface_pk} does not exist",
+            error_message=SystemErrorMessages.UNEXPECTED_ERROR,
+        )
+        task_logger.error(
+            "Could not parse outputs - interface does not exist", exc_info=True
         )
         return
 
     if job.outputs.filter(interface=interface_pk).exists():
-        task_logger.info(
-            f"Output interface with pk={interface_pk} already parsed for job with pk={job_pk}"
+        job.update_status(
+            status=job.FAILURE,
+            error_message=SystemErrorMessages.UNEXPECTED_ERROR,
         )
-        return  # Nothing to do here
+        task_logger.error(
+            "Could not parse outputs - interface already parsed", exc_info=True
+        )
+        return
 
     executor = job.get_executor(backend=backend)
 
     try:
-        outputs = executor.get_outputs(output_interfaces=[interface])
-    except ComponentException as e:
+        val = executor.create_value_for_output(interface=interface)
+        job.outputs.add(val)
+    except ComponentException as error:
         job.update_status(
             status=job.FAILURE,
-            error_message=str(e),
-            detailed_error_message=e.message_details,
+            error_message=str(error),
+            detailed_error_message=error.message_details,
         )
         return
     except Exception:
@@ -1145,8 +1153,6 @@ def parse_singular_job_output(
         )
         task_logger.error("Could not parse outputs", exc_info=True)
         return
-
-    job.outputs.add(*outputs)
 
     if remaining_interface_pks:
         parse_singular_job_output.execute_on_commit(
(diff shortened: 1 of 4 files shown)


app/grandchallenge/components/tasks.py
```diff
@@ -1103,31 +1112,38 @@ def parse_singular_job_output(
...
+            "Could not parse outputs - interface does not exist", exc_info=True
         )
         return
 
     if job.outputs.filter(interface=interface_pk).exists():
-        task_logger.info(
-            f"Output interface with pk={interface_pk} already parsed for job with pk={job_pk}"
+        job.update_status(
+            status=job.FAILURE,
+            error_message=SystemErrorMessages.UNEXPECTED_ERROR,
```
koopmant: This was added to make the task idempotent, it should not fail the job.

app/grandchallenge/components/tasks.py
```diff
@@ -1103,31 +1112,38 @@ def parse_singular_job_output(
...
+        )
+        return
 
     executor = job.get_executor(backend=backend)
 
     try:
-        outputs = executor.get_outputs(output_interfaces=[interface])
-    except ComponentException as e:
+        val = executor.create_value_for_output(interface=interface)
+        job.outputs.add(val)
```
koopmant: Why move this line to within the try-except block?

koopmant (changes requested):

app/grandchallenge/components/tasks.py
jmsmkn: If it does not fail then the job would be left in the parsing state as the tasks are chained.

app/grandchallenge/components/tasks.py
jmsmkn: So that we can catch and communicate any errors that arise from it.

app/grandchallenge/components/tasks.py
koopmant: If this output has already been added, then the task with this interface has already run, hasn't it? So the chain continues with the next interface and the task that fails here is a duplicate. That was our understanding anyway.

jmsmkn: That cannot be assumed. Operations wise it is better to raise an error and be notified than have the job dangling in a "stuck" state. We need to know about this as if it occurs then it is an application problem on our side that would need to be fixed.

koopmant: I see. We agree this task is not idempotent then, right? Is this something we don't (or less) have to worry about since using lambda tasks?

koopmant (dismissed):
LGTM

app/grandchallenge/components/tasks.py
jmsmkn: Tasks still need to be idempotent, but with the way the task was implemented it was not possible to have both. I have updated it so that it recurses over any unparsed interfaces instead.

koopmant: Indeed, that's why I approved it already. Getting the remaining interfaces from the job is smart though!

app/grandchallenge/components/tasks.py
```diff
@@ -1089,68 +1079,55 @@ def parse_singular_job_output(
...
-        )
+    if remaining_interfaces:
+        parse_singular_job_output.execute_on_commit(**job.task_kwargs)
     else:
         job.update_status(status=job.SUCCESS)
 
+    if parsed_interface:
+        return parsed_interface.slug
+    else:
+        return
```
koopmant, line 1129: Where does this end up?

jmsmkn: On `lambda_tasks.models.TaskRecord.result`, visible in the Django admin. See https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/858

koopmant: Thanks.

koopmant (approved):
LGTM

PR #4921, opened 2026-08-05
Parse outputs in parallel

jmsmkn:
Uses the same pattern as input validation for output parsing.

See #4918

### DIFF
--- a/app/tests/components_tests/test_tasks.py
+++ b/app/tests/components_tests/test_tasks.py
@@ -14,6 +14,7 @@ from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.core.files.base import ContentFile
 from django.utils.timezone import now
 from lambda_tasks.decorators import lambda_task
+from lambda_tasks.models import TaskRecord
 from requests import put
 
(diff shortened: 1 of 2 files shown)


jmsmkn:
@koopmant I've tested this locally with multiple outputs and it seems to work as expected.  Trying stacked PRs for the first time, don't know if it makes it any easier!

app/tests/components_tests/test_tasks.py
```diff
@@ -2140,7 +2145,7 @@ def test_parse_singular_job_output(
 
 
 @pytest.mark.django_db
-def test_parse_singular_job_output_idempotent(
+def test_parse_job_output_idempotent(
```
koopmant, line 2148: Could add another test for calling the task with the same interface while not all interfaces have been processed yet. It should return {"status": f"{interface_slug} already exists for job"}

koopmant (dismissed):
Looks great! 

Suggestion for extra test, but not strictly necessary.

koopmant (approved):
LGTM!

PR #4927, opened 2026-08-06
Fix phase admin slowness

jmsmkn:
Adds search fields to the related models so that they can be used as autocomplete fields so that they are only loaded on demand.

Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/783


koopmant (approved):
LGTM

PR #4930, opened 2026-08-06
Do not raise PriorStepFailed

jmsmkn:
Raising means that the transaction is not committed, the task did its work so exit cleanly with a status instead.

See https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/858


koopmant (approved):
LGTM!

PR #4931, opened 2026-08-07
Expire sessions that have been stopped on ECS

jmsmkn:
If sessions on ECS are stopped then we need to update the status in Grand Challenge, this iterates over all of the active unclaimed sessions and marks them as expired if they have been stopped.

See https://github.com/DIAGNijmegen/rse-roadmap/issues/458

### DIFF
--- a/app/tests/workstations_tests/test_tasks.py
+++ b/app/tests/workstations_tests/test_tasks.py
@@ -3,6 +3,7 @@ from datetime import timedelta
 
 import pytest
 from botocore.stub import Stubber
+from django.core.exceptions import ObjectDoesNotExist
 from django.utils.timezone import now
 
 from grandchallenge.components.backends.amazon_ecs import ECSTaskOrchestrator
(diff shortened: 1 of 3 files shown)


app/tests/workstations_tests/test_tasks.py
```diff
@@ -694,3 +695,227 @@ def test_claimed_sessions_count_toward_maximum(
...
+
+        assert result == {"n_sessions_stopped": 1, "n_sessions_started": 3}
+
+        session.refresh_from_db()
+        assert session.status == Session.EXPIRED
+
+        callback_reprs = [repr(c) for c in callbacks]
+        assert _stop_task_repr(pk=session.pk) in callback_reprs
+
+    def test_does_not_stop_session_when_ecs_task_is_running(
```
koopmant: This is already covered by `test_does_not_stop_session_for_active_ecs_statuses` below?

app/tests/workstations_tests/test_tasks.py
```diff
@@ -694,3 +695,227 @@ def test_claimed_sessions_count_toward_maximum(
             .count()
             == 2
         )
+
+    def test_stops_session_when_ecs_task_is_stopped(
```
koopmant: Already covered by `test_stops_session_for_all_terminal_ecs_statuses` below?

koopmant (dismissed):
LGTM, just two possibly duplicate tests.

koopmant (approved):
LGTM!

PR #4933, opened 2026-08-07
Handle requests to sort by non-sortable column

jmsmkn:
This only occurs with scripters or mismatched GET/POST requests where the column sortable state has changed.


koopmant (approved):
LGTM!

PR #4938, opened 2026-08-11
Fix unhandled rejection on upload 403

HarmvZ:
When generate-presigned-urls returns 403 (like upload limit reached), the client now treats it as a non-retryable error: alerts the  user with the detail message, cancels all Uppy uploads to reset the UI, and returns a pending promise to prevent Uppy from processing the response.

Previously the 403 was fed back into Uppy's retry loop, causing infinite retries, a stuck progress bar, and an UnhandledRejection in Sentry.

https://github.com/user-attachments/assets/e08a130e-7511-4b2d-92ad-aec19dbd11ee

Fixes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/895


jmsmkn (approved):
LGTM!

PR #4939, opened 2026-08-12
Catch throttling error when removing images from registry

jmsmkn:


koopmant (approved):
LGTM

(I was staring at the tests trying to figure out why the task was getting called twice, not realizing the previous two calls are included in the final four...)

I see you haven't added a test for the retrying on the throttling error. The test would have to mock the throttling, so I think there is little value in adding the test?

Issue #4817, opened 2026-06-18
Refactor EndpointOrchestrator, AmazonSageMakerEndpointExecutor, AmazonSagemakerTrainingExecutor, and AmazonSagemakerBaseExecutor


koopmant:
After adding the orchestrator for endpoints and endpoint invocations, this has become a mess. 

During development we decided to keep an instance inside the EndpointOrchestrator, so we would not create a new child class. Later AmazonSageMakerEndpointExecutor was added as a new child class anyway, to override some properties.

koopmant:
It's a bit difficult to know without diving into the spaghetti, but I think we should be able to end up with the following:

`Executor(ABC)`
`AmazonSagemakerTrainingExecutor(Executor)`
`AmazonSageMakerEndpointOrchestrator(Executor)`

`Executor` will then handle provisioning and output creation, while the other two deal with object lifecycle and events

| Concern | Executor | AmazonSagemaker-TrainingExecutor | AmazonSagemaker-EndpointOrchestrator |
|---------|:--------:|:------------:|:----------------:|
| S3 client | ✓ | | | |
| SageMaker/Runtime/CloudWatch clients | | ✓ | ✓ |
| Path/key calculation | ✓ | | |
| Instance type selection | ✓ | | |
| Input provisioning | ✓ | | |
| Auxiliary data provisioning | ✓ | | |
| Invocation JSON | ✓ | | |
| Environment variables | ✓ | | |
| Output creation | ✓ | | |
| Inference result handling | ✓ | | |
| Result validation + signing | ✓ | | |
| Cost/duration | ✓ | | |
| Job param parsing | | ✓ | |
| Warm pool | | ✓ | |
| Job lifecycle (create/stop) | | ✓ | |
| Endpoint lifecycle (create/stop) | | | ✓ |
| Invocation lifecycle (create) | | | ✓ |
| Job event handling | | ✓ | |
| Endpoint status event handling | | | ✓ |
| Invocation event handling | | | ✓ |
| Stopped job handling | | ✓ | |
| Failed job handling | | ✓ | |
| Expired invocation handling | | | ✓ |
| Failed invocation handling | | | ✓ |
| Deprovisioning | ✓ (io+aux) | extends (stop + delete training outputs) | extends (endpoint+config+model) |

@jmsmkn do you think this division of concerns makes sense?

jmsmkn:
[...]

PR #4946, opened 2026-08-14
Remove abstract methods from base classes

koopmant:
See #4817

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/app/grandchallenge/components/backends/base.py
+++ b/app/grandchallenge/components/backends/base.py
@@ -414,15 +414,6 @@ class Executor(ABC):
             prefix=self._auxiliary_data_prefix,
         )
 
-    @staticmethod
-    @abstractmethod
-    def get_job_name(*, event):
-        pass
-
-    @staticmethod
-    @abstractmethod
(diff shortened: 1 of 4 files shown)


app/grandchallenge/components/backends/base.py
```diff
@@ -443,10 +434,6 @@ def usd_cents_per_hour(self): ...
     @abstractmethod
     def external_admin_url(self): ...
 
-    @property
-    @abstractmethod
-    def warm_pool_retained_billable_time_in_seconds(self): ...
```
jmsmkn, line 448: This is used in `create_job_warm_pool_utilizations`, which fetches the `settings.COMPONENTS_DEFAULT_BACKEND`. That would now only work with the training executor, changes raising a `NotImplementedError` to an `AttributeError`. I guess that is okay, future developers can still work that one out.

jmsmkn (approved):
LGTM!

PR #4950, opened 2026-08-14
Move utilization_duration methods out of `AmazonSageMakerBaseExecutor`

koopmant:
Utilization and cost tracking is implemented differently for endpoints. By moving it out of `AmazonSageMakerBaseExecutor` to `AmazonSageMakerTrainingExecutor`, this difference is now explicit.

See #4817

### DIFF
--- a/app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
+++ b/app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
@@ -1,6 +1,6 @@
 import logging
 import re
-from datetime import datetime, timedelta
+from datetime import timedelta
 from typing import NamedTuple
 from uuid import UUID
 
@@ -77,16 +77,13 @@ class AmazonSageMakerEndpointOrchestrator(AmazonSageMakerBaseExecutor):
             or settings.AWS_DEFAULT_REGION,
         )
 
-    def _get_start_time(self, *, event):
-        return int(
-            datetime.fromisoformat(event.get("receivedTime")).timestamp()
-            * 1000
-        )
+    @property
+    def utilization_duration(self):
+        raise NotImplementedError
 
-    def _get_end_time(self, *, event):
-        return int(
-            datetime.fromisoformat(event.get("eventTime")).timestamp() * 1000
-        )
+    @property
+    def compute_cost_euro_millicents(self):
+        raise NotImplementedError
 
     @property
     def runtime_setup_result_key(self):
--- a/app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
+++ b/app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
@@ -79,10 +79,12 @@ class AmazonSageMakerEndpointOrchestrator(AmazonSageMakerBaseExecutor):
 
     @property
     def utilization_duration(self):
+        # The duration is tracked on the Endpoint model.
         raise NotImplementedError
 
     @property
     def compute_cost_euro_millicents(self):
+        # The cost is calculated on the Endpoint model.
         raise NotImplementedError
 
     @property
--- a/app/grandchallenge/components/backends/amazon_sagemaker_training.py
+++ b/app/grandchallenge/components/backends/amazon_sagemaker_training.py
@@ -370,7 +370,7 @@ class AmazonSageMakerTrainingExecutor(AmazonSageMakerBaseExecutor):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
-        self.__utilization_duration = None
+        self._utilization_duration = None
 
     @property
     def _training_output_prefix(self):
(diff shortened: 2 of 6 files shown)


app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
```diff
@@ -77,16 +77,13 @@ def external_admin_url(self):
...
-            * 1000
-        )
+    @property
+    def utilization_duration(self):
+        raise NotImplementedError
 
-    def _get_end_time(self, *, event):
-        return int(
-            datetime.fromisoformat(event.get("eventTime")).timestamp() * 1000
-        )
```
koopmant, line 89: These were not being used. I think I had intended them for retrieving a section of the logs for each invocation, but that is not currently implemented. If we implement that, we can put them back.

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
```diff
@@ -77,16 +77,13 @@ def external_admin_url(self):
...
+    @property
+    def compute_cost_euro_millicents(self):
+        raise NotImplementedError
```
koopmant, line 88: This property on `Executor` is not abstract, so overriding with a `NotImplementedError` feels a little strange, but I think it's okay? I wanted to show this property should not be used, although it's not strictly necessary; if this gets called it will raise a NotImplementedError through the utilization_duration property.

app/grandchallenge/components/backends/amazon_sagemaker_training.py
```diff
@@ -364,6 +364,11 @@ def runtime_metrics_chart(self):
 
 
 class AmazonSageMakerTrainingExecutor(AmazonSageMakerBaseExecutor):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.__utilization_duration = None
```
jmsmkn: I think that it is better to avoid using `__`, this makes the attribute impossible to override in subclasses. Why use this over a single one?

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
jmsmkn: How is this calculated for billing then?

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
jmsmkn: I think that it would be better to move that to the base class, or leave it here and make it an abstract method, then it's clear that it is intentional (although a comment about why wouldn't go amiss).

app/grandchallenge/components/backends/amazon_sagemaker_training.py
koopmant: I'm not sure why a double one was used originally. I agree it's not necessary here.

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
koopmant: These are start and end times of invocations. But endpoints are billed on how long they run, regardless of the invocations. We handle that on the model itself when the endpoint is stopped and when `keep_alive` is called.

jmsmkn: Ah okay, shouldn't we still keep the invocation duration?

koopmant: We don't use that. For jobs we set it based on the event, to know when it started and ended. For endpoints, we use the timestamps in the database. Showing it is not implemented here should make that a bit clearer. I've also added a comment.

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
koopmant: Okay, made it abstract on `Executor` and added a comment.

app/grandchallenge/components/backends/amazon_sagemaker_training.py
koopmant: Fixed!

app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
jmsmkn: But what about `Invocation.invoke_duration`? Or is that what you mean with jobs?

koopmant: Ah sorry, I misread your comment. I thought we were talking about utilization duration. Yes, the invoke duration we do use. For jobs and invocations we set that in `Executor._handle_completed_job` (which we call directly in `AmazonSageMakerEndpointOrchestrator._handle_completed_invocation`). 

To clarify
- The invoke and exec durations come from the inference result. 
- The component job utiliization duration comes from the job event
- The endpoint utilization duration is based on the database timestamp `created` and whenever the endpoint is kept alive or stopped. 

I think that is everything, right?

jmsmkn: Yes that clarifies it, thanks!

jmsmkn (approved):
LGTM!

PR #4951, opened 2026-08-14
Disallow creating verification with site email

jmsmkn:
Also adds an admin action to create a verification for the current users email address, that can then be approved in the usual workflow.

See https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/902


koopmant (dismissed):
LGTM

PR #4959, opened 2026-08-19
Replace bleach with JustHTML

### LINKED ISSUE #2755: Replace Bleach
Bleach is now deprecated (https://github.com/mozilla/bleach/issues/698). There are some existing bindings around [ammonia ](https://github.com/rust-ammonia/ammonia) which could be useful, e.g. https://github.com/messense/nh3.

jmsmkn:
Closes #2755

### DIFF
--- a/app/grandchallenge/core/templatetags/bleach.py
+++ b/app/grandchallenge/core/templatetags/bleach.py
@@ -2,7 +2,7 @@ from django import template
 from django.conf import settings
 from django.template.loader import render_to_string
 from django.utils.safestring import SafeString
-from justhtml import JustHTML
+from justhtml import JustHTML, SanitizationPolicy, UrlPolicy, UrlRule
 from markdown import markdown as render_markdown
 from markdown.extensions.toc import TocExtension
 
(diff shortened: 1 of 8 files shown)


app/grandchallenge/core/templatetags/bleach.py
```diff
@@ -17,19 +16,88 @@
...
-    cleaned_html = bleach.clean(
-        html,
-        tags=tags,
-        attributes=settings.BLEACH_ALLOWED_ATTRIBUTES,
-        css_sanitizer=CSSSanitizer(
-            allowed_css_properties=settings.BLEACH_ALLOWED_STYLES
-        ),
-        protocols=settings.BLEACH_ALLOWED_PROTOCOLS,
-        strip=settings.BLEACH_STRIP,
+        allowed_tags = frozenset(
```
jmsmkn, line 24: I moved these definitions here as they're only used in one place, and the construction of the allow_rules is closely related to the contents of `allowed_tags` and `allowed_attributes`.

app/grandchallenge/core/templatetags/bleach.py
koopmant, line 1: Can we rename this file?

koopmant (changes requested):

app/grandchallenge/core/templatetags/bleach.py
jmsmkn: Maybe, but would be a much bigger change to imports both in templates (`{% load clean from bleach %}`, `{% load bleach %}`, `{% load foo,bar,bleach,baz %}`, maybe others) and python imports. The template imports are difficult to cover completely so it is less risky to keep the name the same, suggest moving the rename to a different PR if necessary.

koopmant: Okay!

koopmant (approved):
LGTM!

PR #4960, opened 2026-08-20
Add noopener to blank targets 

jmsmkn:
See DIAGNijmegen/rse-grand-challenge-admin#908


koopmant (approved):
LGTM!

Why not add `noreferrer` as well though?

PR #4961, opened 2026-08-21
Remove beautiful soup

jmsmkn:
We needed to switch from bleach to JustHTML for sanitization, and this library can also perform the work of BeautifulSoup. So, replace that and we have a couple fewer dependencies.


koopmant (approved):
LGTM!

PR #4967, opened 2026-08-25
Fix algorithm evaluation fixtures

### LINKED ISSUE #4963: Algorithm evaluation fixtures broken
Due to sha matching for the container image.

jmsmkn:
Solution is to skip the container upload, this is not used during the evaluation process anyway. The other changes are just a reorganisation as this resulted in some methods not being used in the modules where they were implemented, and this resulted in some circular imports.

Closes #4963


koopmant (approved):
LGTM

PR #4968, opened 2026-08-27
Fix deleted log event handling

jmsmkn:
If the logs have expired then there is nothing else to fetch, so break the loop.

Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/913


koopmant (approved):
LGTM

PR #4970, opened 2026-08-27
Send email when prepaid invoice gets paid

### LINKED ISSUE #4935: Send a notification when a challenge's budget first becomes available
When the first invoice is paid and organizers can start testing, send an email so they don't end up waiting for nothing.

amickan:
Closes #4935


app/grandchallenge/invoices/templates/invoices/partials/challenge_prepaid_invoice_paid_notification_email.md
```diff
@@ -0,0 +1,12 @@
+We are pleased to let you know that we have received payment for a prepaid invoice for your challenge {{ invoice.challenge.short_name }}.
+
+**Invoice details:**
+
+- Amount: {{ invoice.total_amount_euros }} Euro
+- Compute capacity reservation: {{ invoice.compute_costs_euros }} Euro
+- Storage capacity reservation: {{ invoice.storage_costs_euros }} Euro
+
+With this payment, a compute budget of {{ invoice.compute_costs_euros }} Euro has become available for your challenge.
+You can now use this budget to process submissions.
```
koopmant, line 12: Technically, a bigger budget will usually be available, since it includes postpaid budget, right? Maybe just leave the part with the amount out.

koopmant, line 12: I initially figured we should only send this when the budget for the challenge goes above zero for the first time. And then send a message that they can start testing. 

I guess this works as well. Although we will send multiple messages when there are multiple prepaid invoices. I suppose that's acceptable?

Should we still somehow indicate they can start testing?

koopmant (commented):
some questions

app/grandchallenge/invoices/templates/invoices/partials/challenge_prepaid_invoice_paid_notification_email.md
amickan: True. We could a generic sentence along the lines of: "If you have a post-paid capacity reservation as well, that amount is now also available for your challenge, but will be invoiced later"?

Or indeed leave the amounts out entirely.

amickan: > I initially figured we should only send this when the budget for the challenge goes above zero for the first time.

There is no single event that this can be tied to other than the status change, right? And then to figure out whether it's the first time the budget goes above 0 requires additional queries. I figured that it was easier and good enough to just trigger an email on status change. It is nice to be informed about a paid invoice regardless of whether it's the first, I think. 

> Should we still somehow indicate they can start testing?

I thought that was covered by the sentence "You can now use this budget to process submissions." How would you phrase it?

koopmant: Yes, you're right. This is more straightforward. 

> How would you phrase it?

If it had been when the budget first becomes available, I would have said something like "you can start testing and open your phases". But since it is more generic, I think your message makes more sense.

koopmant: Hmm, thinking about it, it will be weird if it is a prepaid invoice with zero compute budget. Because then they still won't be able to process submissions.

amickan: Hm, true, that is a corner case I had not considered. I can include this sentence conditionally.

amickan: I have changed it to only say the part about being able to process submissions when there is compute budget. But I still list the budget split above for completeness. I have added support cost as well for completeness.

koopmant: Nice!

Perhaps call it "base contribution", though? That's how we refer to it in the pricing policy.

amickan: Good idea!

app/grandchallenge/invoices/templates/invoices/partials/challenge_prepaid_invoice_paid_notification_email.md
```diff
@@ -0,0 +1,15 @@
+We are pleased to let you know that we have received payment for a prepaid invoice for your challenge {{ invoice.challenge.short_name }}.
+
+**Invoice details:**
+
+- Amount: {{ invoice.total_amount_euros }} Euro
+- Base cost: {{ invoice.support_costs_euros }} Euro
```
koopmant:
```suggestion
- Base contribution: {{ invoice.support_costs_euros }} Euro
```
This is what we call it on the invoice detail page.

koopmant (commented):
Final nitpick.

koopmant (approved):
LGTM!

PR #4972, opened 2026-08-27
Update SageMaker Shim to v0.8.3

### LINKED ISSUE #4971: Update sagemaker-shim
When shim 0.8.3 is released and we update it, the error message "Invoke endpoint returned status 500, expected 201" will be replaced by the last error message of the user's process. For most algorithms currently used in challenges, the stderr is empty however, so it will become "No errors were reported in the logs."

Should we wait until after MICCAI before we update?

koopmant:
Closes #4971


jmsmkn (approved):
LGTM!

PR #4975, opened 2026-08-28
Add remove option for images

### LINKED ISSUE #3211: FlexibleImageWidget should have a null option

koopmant:
Closes #3211

Removed images may become permanently inaccessible for the user. The widget displays a warning about that.

### DIFF
--- a/app/grandchallenge/cases/form_fields.py
+++ b/app/grandchallenge/cases/form_fields.py
@@ -13,9 +13,9 @@ from grandchallenge.cases.widgets import (
     DICOMUploadWidget,
     DICOMUploadWithName,
     ImageSearchMultiWidget,
+    ImageSourceSelect,
 )
 from grandchallenge.components.models import SourceChoices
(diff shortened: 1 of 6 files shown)


amickan (approved):
LGTM!

PR #4977, opened 2026-08-31
Use `no_key` locks when starting and updating workstation sessions

jmsmkn:
When launching multiple reader studies the through table needs to be updated with new entries, but these are locked when the session is starting. The session itself is not modified by the create view.

Closes https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/917


amickan (approved):
LGTM

PR #4982, opened 2026-09-09
Move inputs and outputs to Job and Evaluation models

amickan:
This moves the `inputs` , `outputs` , `invoke_duration`, `exec_duration` and `input_prefixes` from  `ComponentJob` to `Evaluation` and `Job` in preparation for the batch jobs models. I confirmed that this is a no-op migration, hence no migration files. 

### DIFF
--- a/app/grandchallenge/algorithms/models.py
+++ b/app/grandchallenge/algorithms/models.py
@@ -1182,6 +1182,43 @@ class Job(CIVForObjectMixin, ComponentJob):
         to=ComponentInterfaceValue,
         related_name="%(app_label)s_%(class)ss_as_output",
     )
+    exec_duration = models.DurationField(
+        null=True,
+        default=None,
+        editable=False,
+        help_text=(
+            "The duration of the execution, if measured. "
+            "Excludes data validation, container pulling, model downloading, "
+            "data downloading and data uploading times. "
+            "Includes model loading time, input data loading time, "
+            "processing time, output data writing time and "
+            "any delays from shared hardware issues."
+        ),
+    )
+    invoke_duration = models.DurationField(
+        null=True,
+        default=None,
+        editable=False,
+        help_text=(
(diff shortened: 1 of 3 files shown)


app/grandchallenge/algorithms/models.py
```diff
@@ -1174,6 +1174,14 @@ class Job(CIVForObjectMixin, ComponentJob):
     algorithm_interface = models.ForeignKey(
         AlgorithmInterface, on_delete=models.PROTECT, null=True, blank=True
     )
+    inputs = models.ManyToManyField(
+        to=ComponentInterfaceValue,
+        related_name="%(app_label)s_%(class)ss_as_input",
```
jmsmkn, line 1179: We could be explicit about the related names now to make it easier to trace in the codebase.

Also, as you said, should the exec and invoke duration move too?

amickan: I have just moved the durations and the `input_prefixes` field. If we are explicit about the related names, Django does generate a migration file for some reason

app/grandchallenge/algorithms/models.py
```diff
@@ -1174,6 +1174,51 @@ class Job(CIVForObjectMixin, ComponentJob):
...
+            "Excludes data validation, container pulling, model downloading, "
+            "data downloading and data uploading times. "
+            "Potentially excludes model loading time, depending on the "
+            "users implementation. "
+            "Includes input data loading time, "
+            "processing time, output data writing time and "
+            "any delays from shared hardware issues."
+        ),
+    )
+    input_prefixes = models.JSONField(
```
amickan, line 1213: This is only relevant for evaluations as far as I understand, but since the executor currently assumes that this field is present on the Job model, we need to move it here as well. `BatchJob` won't need it and hence should not inherit it IMO.

jmsmkn (approved):
LGTM!

PR #4984, opened 2026-09-09
Make algorithm_interface required for algorithm jobs

### LINKED ISSUE #4983: Algorithm interface should be required for algorithm jobs
The algorithm interface is currently optional for algorithm jobs on the DB level: 

```python
algorithm_interface = models.ForeignKey(
        AlgorithmInterface, on_delete=models.PROTECT, null=True, blank=True
)
```

`null=True, blank=True` should be removed. I think this was a leftover task that was forgotten after implementing algorithm interfaces. We should be able to just make this change since all jobs should have an interface, but it would be good to double check in the shell. The following should return 0:

```python
from grandchallenge.algorithms.models import Job
Job.objects.filter(algorithm_interface__isnull=True).count()
```

amickan:
Closes #4983


app/grandchallenge/algorithms/migrations/0104_alter_job_algorithm_interface.py
```diff
@@ -0,0 +1,24 @@
...
+    dependencies = [
+        ("algorithms", "0103_algorithm_algorithms_algorithm_unique_slug"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="job",
+            name="algorithm_interface",
+            field=models.ForeignKey(
+                default=0,
```
jmsmkn: Why does this have a default and preserve default?

amickan: Misunderstanding from my side. When running `makemigrations` it asks for a default, so I gave it one thinking that if I didn't, this same prompt would pop up again when running the migration at deploy time. I realize now that that is of course not the case (we run `migrate`, not `makemigrations` when deploying and we know that there are no empty rows, so no need for a default). Sorry!

jmsmkn (approved):
LGTM!

PR #4985, opened 2026-09-09
Add BatchJob and BatchJobTask models

amickan:
This adds the `BatchJob` and `BatchJobTask` models as described in the [pitch](https://github.com/DIAGNijmegen/rse-roadmap/issues/374).

The models are still missing a bunch of methods / properties, which I will add in later PRs as I add the code that uses these. 
I did have to add a dummy `create_utilization` method already or else the tests would fail.


jmsmkn (approved):
LGTM!

PR #4986, opened 2026-09-10
Update executor to provision a list of inference tasks

amickan:
This enables the executor to provision multiple `InferenceTask`s and adds a separate provisioning entry point for batch jobs. 

This PR has been implemented iteratively with kiro, all code has been pre-reviewed by me.

### DIFF
--- a/app/grandchallenge/components/backends/base.py
+++ b/app/grandchallenge/components/backends/base.py
@@ -9,6 +9,7 @@ import logging
 import os
 import secrets
 from abc import ABC, abstractmethod
+from collections.abc import Iterable
 from datetime import timedelta
 from json import JSONDecodeError
 from math import ceil
@@ -82,6 +83,13 @@ class JobParams(NamedTuple):
 class CIVProvisioningTask(NamedTuple):
     key: str
     task: functools.partial
+
+
+class ProvisioningTask(NamedTuple):
+    pk: str
+    input_civs: Iterable[ComponentInterfaceValue]
+    input_prefixes: dict[str, str]
+    output_prefix: str
 
 
 def duration_to_euro_millicents(*, duration, usd_cents_per_hour):
@@ -377,14 +385,31 @@ class Executor(ABC):
         self.__s3_client = None
 
     def provision(self, *, input_civs, input_prefixes):
-        # We cannot run everything async as it requires database access.
-        # So first we gather the async tasks that need to be run,
-        # then execute them in the event loop for the current thread
-        # using a method wrapped in @async_to_sync.
-        provisioning_tasks = self._get_provisioning_tasks(
-            input_civs=input_civs, input_prefixes=input_prefixes
-        )
-        self._provision(tasks=provisioning_tasks)
+        tasks = [
+            ProvisioningTask(
+                pk=self._job_id,
+                input_civs=input_civs,
+                input_prefixes=input_prefixes,
+                output_prefix=self._io_prefix,
+            )
+        ]
+        self._provision(tasks=self._get_provisioning_tasks(tasks=tasks))
+
+    def provision_batch_job(self, *, batch_job):
+        tasks = [
+            ProvisioningTask(
+                pk=f"{self._job_id}-{task.pk}",
+                input_civs=task.inputs.prefetch_related(
+                    "interface", "image__files"
+                ).all(),
+                input_prefixes={},
+                output_prefix=self._output_prefix_for_task(
+                    task_pk=str(task.pk)
+                ),
+            )
+            for task in batch_job.tasks.all()
+        ]
+        self._provision(tasks=self._get_provisioning_tasks(tasks=tasks))
 
     @abstractmethod
     def execute(self): ...
@@ -484,6 +509,9 @@ class Executor(ABC):
     def _io_prefix(self):
         return safe_join("/io", *self.job_path_parts)
 
+    def _output_prefix_for_task(self, *, task_pk):
+        return safe_join(self._io_prefix, task_pk)
+
     @property
     def _invocation_prefix(self):
         return safe_join("/invocations", *self.job_path_parts)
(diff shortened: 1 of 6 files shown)


app/grandchallenge/components/backends/base.py
```diff
@@ -84,6 +85,13 @@ class CIVProvisioningTask(NamedTuple):
     task: functools.partial
 
 
+class InferenceTaskSpec(NamedTuple):
+    pk: str
+    input_civs: Iterable[ComponentInterfaceValue]
+    input_prefixes: dict[str, str]
+    output_prefix: str
```
amickan, line 92: Open to suggestions for better names for this class.

app/grandchallenge/components/backends/base.py
```diff
@@ -582,31 +614,47 @@ async def _provision(self, *, tasks):
...
+                        )
                     )
+
+            inference_tasks.append(
+                InferenceTask(
+                    pk=task_spec.pk,
+                    inputs=invocation_inputs,
+                    output_bucket_name=self._output_bucket_name,
+                    output_prefix=task_spec.output_prefix,
+                    timeout=self._time_limit,
```
amickan, line 642: Unsure about what the time_limit set on the executor for batch jobs should reflect: the maximum runtime for the `BatchJob` as a whole, or the maximum runtime for each of its subtasks? The current implementation assumes the latter since the shim seems to set this as timeout for a single invocation.

app/grandchallenge/components/backends/base.py
```diff
@@ -377,14 +385,35 @@ def __init__(
         self.__s3_client = None
 
     def provision(self, *, input_civs, input_prefixes):
-        # We cannot run everything async as it requires database access.
```
jmsmkn, line 380: Is this comment no longer useful?

app/grandchallenge/components/backends/base.py
```diff
@@ -377,14 +385,35 @@ def __init__(
...
+                    input_civs=input_civs,
+                    input_prefixes=input_prefixes,
+                    output_prefix=self._io_prefix,
+                )
+            ]
         )
-        self._provision(tasks=provisioning_tasks)
+        self._provision(tasks=tasks)
+
+    def provision_batch_job(self, *, batch_job):
```
jmsmkn: This increases the API footprint and is mostly duplicate of `provision`. Can this be rolled up into `provision` somehow so that it works for both individual and batch tasks?

app/grandchallenge/components/backends/base.py
```diff
@@ -377,14 +385,35 @@ def __init__(
...
+        tasks = self._get_provisioning_tasks(
+            task_specs=[
+                InferenceTaskSpec(
+                    pk=f"{self._job_id}-{task.pk}",
+                    input_civs=task.inputs.prefetch_related(
```
jmsmkn: Shouldn't this all be prefetched on the first query and not in each row?

app/grandchallenge/components/backends/base.py
jmsmkn: The timeout here should be the one that is for the `BatchJobTask`. The time limit on the `BatchJob`, which is sent to and enforced by SageMaker, should be the sum of these, plus the container pulling, process startup time and data download/upload time.

app/grandchallenge/components/backends/base.py
```diff
@@ -704,19 +761,15 @@ def _get_civ_provisioning_tasks(self, *, civ, input_prefixes):
...
-        )
-
+    def _get_create_invocation_json_task(self, *, inference_tasks):
         if self._use_task_list:
-            content = to_json([inference_task])
+            content = to_json(inference_tasks)
         else:
-            content = to_json(inference_task)
+            if len(inference_tasks) != 1:
+                raise NotImplementedError(
```
jmsmkn: I think that a `ValueError` would be more appropriate, and nit style wise would be to add the else block.

app/grandchallenge/components/backends/base.py
jmsmkn: Seems as good as any.

app/grandchallenge/components/backends/base.py
amickan: I had opted to keep things separate for clarity and so that I didn't have to touch the current job scheduling calls anywhere. I see your point though about API footprint. I'll work on unifying the two into one single provision method.

app/grandchallenge/components/backends/base.py
amickan: Still useful. It had just gotten lost in moving back and forth between implementations.

app/grandchallenge/components/backends/base.py
amickan: I pushed a version with a single provisioning entrypoint for all models.

jmsmkn (dismissed):
LGTM!

app/grandchallenge/components/backends/base.py
amickan: Thanks!

Just to confirm I understood this correctly: Reading the pitch, I understand that we want to set the time limit on the `BatchJob` through an environment variable `EVALUATION_MAXIMUM_BATCH_JOB_DURATION`. So the enforced time limit for a batch job will always be the same. For single tasks within a batch, we use the phase-level defined `algorithm_time_limit`. We divide `EVALUATION_MAXIMUM_BATCH_JOB_DURATION` by the phase-level defined time-limit to get the number of `BatchJobTasks` we can attach to a batch job. 

We could derive the per-task time out here in the executor, but I think it would be cleaner to add a
`time_limit` field to the `BatchJobTask` model and pass that along as part of `InferenceTaskSpec` (or otherwise pass the time limit along rather than calculating it here). Would you agree?

jmsmkn: `EVALUATION_MAXIMUM_BATCH_JOB_DURATION` just determines the maximum number of BatchJobTasks per BatchJob. You could include just 1 BatchJobTask in a BatchJob, say if you did 999 in the first BatchJob, then there is just 1 left to schedule in the second BatchJob.

If you fix the value of the timeout then you run the risk of letting the task go too long - say if there is an exit failure after a timeout in SageMaker shim, we do not know how robust that mechanism is.

I think that is better to keep this responsibility of determining what the total allowed time should be with the executor taking into account the included timeouts. That saves having an additional field as we already know the timeout of the individual `BatchJobTask` from the phase. Then there is no need to keep those two things in sync and you have a single place to define what the time limit should be. The executor is then not concerned with `EVALUATION_MAXIMUM_BATCH_JOB_DURATION`.

jmsmkn (approved):

app/grandchallenge/components/backends/base.py
amickan: > If you fix the value of the timeout then you run the risk of letting the task go too long - say if there is an exit failure after a timeout in SageMaker shim, we do not know how robust that mechanism is.

That's a good argument. 

> I think that is better to keep this responsibility of determining what the total allowed time should be with the executor taking into account the included timeouts.

I'm not sure I know what you mean here. The `BatchJob` has a time_limit field, which is what is currently used for both the `MaxRuntimeInSeconds` set for the Sagemaker job _and_ the individual invocations. I'll still need to fix that. I think it makes sense that the time limit set on the BatchJob refers to the total allowed time. This can be calculated when creating the `BatchJob` and its tasks by the asynchronous task that takes care of the scheduling. The individual invocation time outs are already known and can be passed as additional argument to `InferenceTaskSpec` to the executor. That seems like the cleanest solution to me. The executor then indeed does not need to be concerned with `EVALUATION_MAXIMUM_BATCH_JOB_DURATION`, however, it also does not "determine what the total allowed time should be" - so maybe you disagree with me assessment. I'll merge this PR and make a new PR just for the small change that I think is still necessary to set the timeouts all correctly. We can continue discussing things there.

PR #4987, opened 2026-09-11
Set timeout for individual batch job tasks

amickan:
This makes it possible to set individual time limits for the batch job tasks. This adds a computed `total_time_limit` on the executor to use when scheduling the Sagemaker job, while keeping individual time limits for the sub-tasks in the `InferenceTaskSpec`. 

### DIFF
--- a/app/grandchallenge/components/backends/base.py
+++ b/app/grandchallenge/components/backends/base.py
@@ -90,6 +90,7 @@ class InferenceTaskSpec(NamedTuple):
     input_civs: Iterable[ComponentInterfaceValue]
     input_prefixes: dict[str, str]
     output_prefix: str
+    timeout: timedelta
 
 
 def duration_to_euro_millicents(*, duration, usd_cents_per_hour):
@@ -393,7 +394,12 @@ class Executor(ABC):
         self._provision(tasks=tasks)
 
     def build_inference_task_spec(
-        self, *, input_civs, input_prefixes=None, task_pk=None
+        self,
+        *,
+        input_civs,
+        input_prefixes=None,
+        task_pk=None,
+        task_timeout=None,
     ):
         return InferenceTaskSpec(
             pk=f"{self._job_id}-{task_pk}" if task_pk else self._job_id,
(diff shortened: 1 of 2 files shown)


app/grandchallenge/components/backends/amazon_sagemaker_endpoint.py
```diff
@@ -269,9 +267,14 @@ def handle_status_event(self, *, event):
         else:
             raise ValueError("Invalid endpoint status")
 
-    def provision_invocation_input_data(self, *, input_civs):
+    def provision_invocation_input_data(self, *, input_civs, time_limit):
```
amickan, line 264: The mix of `time_limit` and `timeout` terminology is slightly annoying. Should we choose one of those terms and use that everywhere? The model fields are called `time_limit`, so that would be the natural choice.

jmsmkn: Agreed - it will probably be too difficult to change within SageMaker Shim as it is part of the API there (https://github.com/DIAGNijmegen/rse-sagemaker-shim#sagemaker-shim-serve), but we can use time limit in all places internally until that object is created.

jmsmkn (dismissed):
LGTM!

app/tests/components_tests/test_amazon_sagemaker_endpoint_backend.py
```diff
@@ -530,12 +530,14 @@ def test_endpoint_orchestrator_provision_invocation_input_data_tasks(
         ],
         "output_bucket_name": "algorithm-endpoints-output",
         "output_prefix": f"/io/algorithms/invocation/{invocation.pk}",
-        "timeout": "PT42S",
+        "time_limit": "PT42S",
```
jmsmkn: The invocation json no longer matches the specifications from SageMaker Shim: https://github.com/DIAGNijmegen/rse-sagemaker-shim#sagemaker-shim-serve

jmsmkn (approved):
LGTM!

PR #4988, opened 2026-09-14
Remove compute_costs_euros_per_hour_for_tasks

amickan:
Bumped into this while working on https://github.com/DIAGNijmegen/rse-grand-challenge/pull/4987

The challenge request budget calculator calculates compute costs per hour by looping over tasks. The hourly compute costs, however, no longer differ per task since they depend solely on things that are set on the entire challenge request level. I think this changed a while ago and this part of the code was probably just not updated.


jmsmkn (approved):
LGTM!

PR #4990, opened 2026-09-16
Use a GitHub app for dependency updates

jmsmkn:
This allows the normal GitHub actions to be run to save forcing through a PR each week.


jmsmkn:
I have created an GitHub app for this and installed it for all of DIAG Nijmegen. To use it in other repos you need to set the actions secrets in the target repo `DEPENDENCY_UPDATE_APP_CLIENT_ID`, which can be found on https://github.com/organizations/DIAGNijmegen/settings/apps/diag-dependency-updates as Client ID under About, and `DEPENDENCY_UPDATE_APP_PRIVATE_KEY`, which is the full contents of a pem file (encluding the ---- begin and ---- end lines) which can be generated under "Private keys" on the same page.

PR #4994, opened 2026-09-16
Pin actions workflows using commit ids

jmsmkn:
This protects against supply chain attacks. The ids are kept up to date weekly by dependabot.


pkcakeout:
Good idea! I will bookmark this for CIRRUS as well!

PR #4995, opened 2026-09-17
Provision task specs at executor creation time

amickan:
The task-specs are needed both at provisioning time and at execution time now that we derive the overall time limit for the boto job from the task specs. This updates the executor so that the task definitions are passed to it at creation time.  

Fix for https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/936


app/grandchallenge/algorithms/models.py
```diff
@@ -1472,6 +1473,17 @@ def executor_kwargs(self):
             executor_kwargs["algorithm_model"] = self.algorithm_model.model
         return executor_kwargs
 
+    def get_inference_task_definitions(self):
```
jmsmkn: Nit: `get_` with only `self` indicates that this should be a `@property` `inference_task_definitions`. Could then use `@cached_property` if the generation is expensive.

app/grandchallenge/components/backends/base.py
```diff
@@ -619,26 +598,34 @@ async def _provision(self, *, tasks):
                             )
                         )
 
-    def _get_provisioning_tasks(self, *, task_specs):
+    def _get_provisioning_tasks(self):
```
jmsmkn: Same here.

jmsmkn (dismissed):
LGTM! Up to you about whether to fix the nits.

app/grandchallenge/algorithms/models.py
amickan: Makes a lot of sense! I'll make the change.

jmsmkn (approved):
LGTM!

PR #4996, opened 2026-09-17
Use uv to manage biom and panimg environments

### LINKED ISSUE #4535: `biom` and `panimg` virtualenvs should use uv lockfile for installs
Currently uses pip, where the dependencies are untracked. The lockfile would need automatic updating.

jmsmkn:
Closes #4535


amickan (approved):
LGTM

PR #4997, opened 2026-09-18
Add a contact-us app/form

chrisvanrun:
This PR adds a contact-us form which does not require an authenticated user.

When submitted, an email containing the message (+ email) is sent to the managers. 

Follow-ups.  It is currently not linked anywhere, the idea is to provide links to this in various places with a preset message for ease.

Relates to pitch:
- https://github.com/DIAGNijmegen/rse-roadmap/issues/501

### Showcase

afbeelding

### AI Disclaimer
This changeset was written with the aid of AI (Kiro), steered by me. The code was subsequently double-checked, reviewed and cleaned up by me.


app/grandchallenge/contact/emails.py
```diff
@@ -0,0 +1,45 @@
...
+def send_contact_email(*, cleaned_data, request):
+    """Send a contact form submission to the site managers.
+
+    The email contains all the form fields, the ``referer`` and some metadata
+    about the request (client IP, user agent and referer) to help triage.
+    """
+
+    forwarded_for = request.headers.get("x-forwarded-for", "<none>")
+    user_agent = request.headers.get("user-agent", "<none>")
+    host = request.get_host()
```
jmsmkn: This is protected information under GDPR and cannot be associated with an individual as it is not part of our privacy policy, why is this necessary to collect?

app/grandchallenge/contact/templates/contact/contact_form.html
```diff
@@ -0,0 +1,32 @@
...
+{% block content %}
+    <h1>Send us a message</h1>
+
+    {% if honeypot_tripped %}
+        <p class="alert alert-danger">
+            We couldn't send your message automatically. Please retry or email
+            <a href="mailto:{{ SUPPORT_EMAIL }}">{{ SUPPORT_EMAIL }}</a>
+            directly and we will get back to you.
+        </p>
+    {% endif %}
```
jmsmkn: Why tell users about this?

app/grandchallenge/contact/static/contact/js/contact_challenge.mjs
```diff
@@ -0,0 +1,84 @@
+/**
```
jmsmkn: This file is not being minified, so all of the comments will be included. An LLM could work out what is happening anyway. I'm not sure on the value of this, it's adding a lot of complexity and is a never ending battle which we will never win. Might be better to keep the implementation here simple and leave the rest to AWS WAF?

app/grandchallenge/contact/forms.py
```diff
@@ -0,0 +1,90 @@
...
+    )
+    email = forms.EmailField(
+        label="Your email",
+        widget=forms.TextInput(attrs={"placeholder": "Your email"}),
+    )
+    message = forms.CharField(
+        label="Your message",
+        widget=forms.Textarea(attrs={"placeholder": "Your message"}),
+    )
+    accept_privacy_policy = forms.BooleanField(
```
jmsmkn: If we remove the name field then this also is not necessary.

app/grandchallenge/contact/forms.py
```diff
@@ -0,0 +1,90 @@
...
+    # and is never used anywhere (not included in the email).
+    subject = forms.CharField(
+        required=False,
+        label="Subject",
+        widget=forms.HiddenInput,
+    )
+
+    # Hidden reference field, allows us to know where the user reached the
+    # form from. Can be prefilled via a URL argument.
+    referer = forms.CharField(
```
jmsmkn: Why is this necessary?

app/grandchallenge/contact/views.py
```diff
@@ -0,0 +1,40 @@
...
+    def get_initial(self):
+        initial = super().get_initial()
+        for field in ["message", "referer"]:
+            value = self.request.GET.get(field)
+            if value is not None:
+                initial[field] = value
+        return initial
+
+    def get_success_url(self):
+        return reverse("home")
```
jmsmkn, line 22: Maybe add a success message?

jmsmkn: Ah, I see that it is in form valid. Usually we use `SuccessMessageMixin`

app/grandchallenge/contact/views.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def form_valid(self, form):
+        if form.honeypot_tripped or not form.challenge_passed:
```
jmsmkn: Form valid should only emit events and save the form, it should not render anything else.

app/grandchallenge/contact/forms.py
```diff
@@ -0,0 +1,90 @@
...
+        )(),
+    )
+
+    # Honeypot field: hidden from real users, but likely to be filled in by
+    # bots. It is deliberately named "subject" to look like a plausible field
+    # and is never used anywhere (not included in the email).
+    subject = forms.CharField(
+        required=False,
+        label="Subject",
+        widget=forms.HiddenInput,
```
jmsmkn: This is unlikely to be effective as hidden fields are ignored. For reference, see how this is implemented in Django AllAuth.

app/grandchallenge/contact/emails.py
chrisvanrun: My rational is our dealings with spam and how to handle things when it spikes so much we need deploy some blocks. We'd be collecting it to provide the service.

jmsmkn: That is not necessary. All of that information is available via server logs, but those are not tied to an individual like you are trying to do here, so therefore not part of GDPR.

app/grandchallenge/contact/templates/contact/contact_form.html
chrisvanrun: This should also show when the challenge fails (I could/should add that). It is more of a catch for false positives and nudge users to use email instead.

app/grandchallenge/contact/static/contact/js/contact_challenge.mjs
chrisvanrun: It does add some complexity. On the scale of anti-bot-tampering however, this is a relative straight forward challenge. Per advice of Paul, who had some dealing with keeping bots away from a form, these sort of challenges are very effective. He also used key-strokes on input fields and other fingerprints. I thought that was over the top and kept it a bit simpler

app/grandchallenge/contact/forms.py
chrisvanrun: There won't be a single entry point to this form. We could rely on the message content itself, but knowing which 'contact-us' button sees the most milage can be informative for design.

Knowing the verbosity of the average users I thought it wise to provide some measure of context on the message via the which button they used. For instance: book a reader study demo, or host your challenge, et cetera.

app/grandchallenge/contact/forms.py
chrisvanrun: I do agree that it's very unlikely to be effective. I'll have a look at how AllAuth has weaponized it.

app/grandchallenge/contact/forms.py
chrisvanrun: I'd be afraid of losing some human touch here? Apart from that, isn't the email address in itself already something we'd need to cover in the privacy policy?

app/grandchallenge/contact/forms.py
jmsmkn: I think this falls under "You ain't gonna need it" and I really cannot see the value of it. The user would almost certainly tell you about what they want in their message. If their message is so terse as "Yes interested", they're probably not worth bothering with anyway. Either way, the topic can be asked for when replying if it is unclear.

app/grandchallenge/contact/static/contact/js/contact_challenge.mjs
chrisvanrun: > An LLM could work out what is happening anyway. 

Certainly! Even if we'd obfuscate this: I am sure an targeted LLM loop with some context resolving form exploits will not blink at anything I can throw at it.

jmsmkn: I don't think we need to bother with the challenge or honeypot. Keep the implementation simple here and solve that on the WAF level, which is already implemented now: https://github.com/DIAGNijmegen/rse-cloud-infrastructure/pull/443

If that is insufficient, then think about adding complexity here. I recommend going for the most straightforward solution first.

chrisvanrun:
I will rework this PR into a minimum form where just the message is prefilled.

app/grandchallenge/contact/forms.py
jmsmkn: Consent is not always necessary. This falls under one of the other legal bases - "Data processing is necessary for the performance of a contract", so long as we only collect the data absolutely necessary (conservatively, email and message) for establishing this:

> Legal basis of contract
> One of the legal bases in the GDPR is that processing of personal data is necessary for the performance of a contract. You may rely on this legal basis if you have a contract with someone and you are unable to perform that contract without processing personal data as well.
> 
> This legal basis also applies for the stage prior to the conclusion of a contract. For example: when you draw up a quotation on request.
> 
> Make sure that you do not process personal data that are not necessary for the performance of the contract. If you do so, you will need consent or another legal basis for processing such data. Unless it concerns a compatible further processing.

See https://www.autoriteitpersoonsgegevens.nl/en/themes/basic-gdpr/gdpr-basics/legal-bases-from-the-gdpr-explained

app/grandchallenge/contact/forms.py
chrisvanrun: > Either way, the topic can be asked for when replying if it is unclear.

True.

app/grandchallenge/contact/forms.py
chrisvanrun: > This legal basis also applies for the stage prior to the conclusion of a contract. For example: when you draw up a quotation on request.

Yep, that seems to cover it. I had half a thought to add the policy link to the template as a polite reminder but it's already in the footer of the website: visibile enough for this imho.

chrisvanrun:
Cut this changeset to a minimum. I have updated the PR description with a new screenshot of the render.

app/grandchallenge/contact/emails.py
```diff
@@ -0,0 +1,19 @@
...
+    body = format_html(
+        "Email: {email}\n\nMessage:\n{message}\n",
+        email=email,
+        message=message,
+    )
+
+    mail_managers(
+        subject=subject,
+        message=body,
+    )
```
jmsmkn: How about setting reply_to in order to make contacting straightforward for the managers? Would then probably remove the users email address from the subject. It's not supported by `mail_managers` but can be achieved with:

```python3
site = Site.objects.get_current()
domain = site.domain.lower()

# this check is taken from Django 5.2, the format of the manager emails is changed in Django 6.0
if not all(isinstance(a, (list, tuple)) and len(a) == 2 for a in settings.MANAGERS):
    raise ValueError("The MANAGERS setting must be a list of 2-tuples.")

EmailMessage(
    subject=f"[{domain}] New contact message",
    body=message,
    to=[a[1] for a in settings.MANAGERS],
    reply_to=[email],
).send()
```

app/tests/contact_tests/test_views.py
```diff
@@ -0,0 +1,64 @@
...
+    form = response.context["form"]
+    assert form.initial["message"] == "Prefilled Message"
+
+    content = response.rendered_content
+    assert "Prefilled Message" in content
+
+
+@pytest.mark.django_db
+def test_valid_submission_redirects_to_homepage(client, settings):
+    settings.SUPPORT_EMAIL = TEST_SUPPORT_EMAIL
```
jmsmkn: What does this do?

chrisvanrun: Ah, a left over from the honey pot hit 'error' message. Missed that in the cleanup.

app/grandchallenge/contact/emails.py
chrisvanrun: I like the initial idea! 

I had this first but switched to keep a uniform way of sending emails. Would make it slightly more simple for us to reply to, but at the same time I am slightly worried about it playing a part in the emails getting flagged for spam. The sender and reply not matching is a signal for junk. AFAIK this only happens on a per-message basis, so it'd have to get flagged by our support inbox, which we have some measure of control over. So that will likely be fine.

Another thing I am unsure about is the subject, as you mentioned. Having a bunch of "Re: [CONTACT-US] Message" isn't very informative. At the same time, with a new email inbox we might have conversations and a reply-to might bind things nicely!

I'll implement it.

jmsmkn (approved):
LGTM!

PR #4998, opened 2026-09-18
Update executor to handle multiple inference results

amickan:
This updates the executor to handle a list of `InferenceResults`. Updates to the output parsing methods will be made in a separate PR. 


app/grandchallenge/evaluation/models.py
```diff
@@ -1906,6 +1906,61 @@ def executor_kwargs(self):
             executor_kwargs["algorithm_model"] = self.algorithm_model.model
         return executor_kwargs
 
+    def update_status(  # noqa:C901
```
jmsmkn: Isn't this duplicated from somewhere else? There isn't a way to re-use the common parts? Just thinking about future maintenance.

app/grandchallenge/components/backends/base.py
```diff
@@ -513,10 +518,11 @@ def runtime_setup_result_key(self):
             self._io_prefix, ".sagemaker_shim", "runtime_setup_result.json"
         )
 
-    @property
-    def _inference_result_key(self):
+    def _inference_result_key(self, *, task_pk=None):
```
jmsmkn: Probably should be `_get_inference_result_key` now.

app/grandchallenge/evaluation/models.py
amickan: Oh, this is a pretty stupid oversight on my part. Yes, this can definitely be re-used. I copied this from the `ComponentJob` model without realizing that that is the base for `BatchJobs` as well....

jmsmkn: Glad these reviews are still useful :-)

app/grandchallenge/components/backends/base.py
```diff
@@ -435,12 +434,8 @@ def utilization_duration(self): ...
...
 
     @property
-    def exec_duration(self):
-        return self._exec_duration
-
-    @property
-    def invoke_duration(self):
-        return self._invoke_duration
+    def inference_results(self):
+        return self._inference_results
```
jmsmkn, line 438: This relies on ordering of operations, that `_handle_completed_job` is done first. Maybe the work should be moved here? Somethings like:

```python
@cached_property
def inference_results(self):
    return [
        self._get_inference_result(task_pk=task.task_pk)
        for task in self._task_definitions
    ]
```

app/grandchallenge/components/backends/base.py
```diff
@@ -499,6 +494,16 @@ def _io_prefix(self):
...
         return safe_join(self._io_prefix, task_pk)
 
+    def _output_prefix(self, *, task_pk=None):
+        return (
+            self._output_prefix_for_task(task_pk=task_pk)
+            if task_pk
+            else self._io_prefix
+        )
+
+    def _inference_task_pk(self, *, task_pk=None):
```
jmsmkn, line 504: Nit: maybe these two new methods and `_output_prefix_for_task` should have a `_get_` prefix too?

app/grandchallenge/components/backends/base.py
amickan: Not sure. It only makes sense to fetch the inference results for completed jobs, right? If we move the fetching from S3 to the property and someone accesses it when there are no results (yet), it would error out. That would be strange, right? I think you would want it to return an empty list then.

amickan: On second thought, I understand your suggestion better, but I'm not sure what the appropriate behavior would be if `_get_inference_result` fails for one of the tasks. The property is currently called by `get_update_status_kwargs`, which is called for completed jobs, but also for failed and stopped boto jobs. I assume that for those jobs, there would not (always) be a result file to fetch and that would be fine, we don't really care about the result files for those jobs, right?

amickan: In other words: for failed and stopped jobs, we want to ignore exceptions raised by `_get_inference_result` (in fact, that is what we currently do for failed jobs), but for completed jobs, we don't want to ignore them. Is that assumption correct?

jmsmkn: > In other words: for failed and stopped jobs, we want to ignore exceptions raised by _get_inference_result (in fact, that is what we currently do for failed jobs), but for completed jobs, we don't want to ignore them. Is that assumption correct?

I think so, yes. For failed/stopped jobs there _should_ be a results file, but we also accept that generating it reliably in those situations is difficult. I think that we emit a message to Sentry for those though?

amickan: Thanks for confirming. I've made a suggestion. Is this better?

> I think that we emit a message to Sentry for those though?

We did not do this, but I've changed it so we do now.

jmsmkn (approved):
LGTM!

PR #4999, opened 2026-09-21
Remove the sign up honeypot field

jmsmkn:
This was preventing sign ups from genuine users who use auto filling in their browser. CAPTCHA is now added at the WAF level.

See DIAGNijmegen/rse-roadmap#501


chrisvanrun (approved):
LGTM!

PR #5000, opened 2026-09-21
Output parsing for batchjobs

amickan:
This moves the job output parsing logic out of the async tasks and onto methods on the job models, and in doing so, enables parsing outputs for multiple tasks. 


amickan:
I'm puzzled by the tests that are failing here. The changes here shouldn't affect the tests that are failing, and even more strangely: running them locally, they also fail on `main` (even though my previous PR passed all tests).

jmsmkn:
I saw this on another branch, try re-running them.

scripts/local_s3.py
```diff
@@ -39,6 +39,8 @@ def _create_buckets(*, client):
         settings.UPLOADS_S3_BUCKET_NAME,
         settings.COMPONENTS_INPUT_BUCKET_NAME,
         settings.COMPONENTS_OUTPUT_BUCKET_NAME,
+        settings.ALGORITHM_ENDPOINTS_INPUT_BUCKET_NAME,
+        settings.ALGORITHM_ENDPOINTS_OUTPUT_BUCKET_NAME,
```
amickan, line 43: This was necessary to make the tests pass. It's a mystery to me though why these tests fail only now and not on my previous PR where I introduced `_results_or_empty`

amickan:
> I saw this on another branch, try re-running them.

Ah, ok. I can revert my latest commit and see if they pass. They consistently fail locally though.

jmsmkn:
Check what your seaweed fs container is logging.

jmsmkn:
I think you found the issue and have merged main which should hopefully solve it.

scripts/local_s3.py
jmsmkn: I think this was the issue. Maybe something to do with test ordering? Anyway, added in #5003

app/grandchallenge/algorithms/models.py
```diff
@@ -1274,6 +1274,32 @@ def container(self):
...
 
+    def schedule_output_parsing(self):
+        # Local import to avoid a circular dependency
+        from grandchallenge.components.tasks import parse_job_output
+
+        for interface in self.output_interfaces.all():
+            parse_job_output.execute_on_commit(
+                **self.task_kwargs,
+                interface_slug=interface.slug,
+            )
```
jmsmkn: This fragment is duplicated twice and only slightly different on the third. Maybe there is a way to do this that reduces that?

app/grandchallenge/algorithms/models.py
```diff
@@ -1274,6 +1274,32 @@ def container(self):
...
+    def add_output_value(self, *, value, task_pk=None):
+        if task_pk:
+            raise ValueError("This job does not have subtasks.")
+        self.outputs.add(value)
+
+    @property
+    def output_parsing_complete(self):
+        expected_interfaces = {*self.output_interfaces.all()}
+        parsed_interfaces = {output.interface for output in self.outputs.all()}
+        return not (expected_interfaces - parsed_interfaces)
```
jmsmkn: Same with these. Maybe think about what is different. Can we treat everything as operating on multiple subtasks? Then, `Job` and `Evaluation` happen to have a single one.

amickan: Found a way! I do think this adds slightly more complexity, but I agree it's worth it to avoid duplicate code and to have a single place for the logic.

jmsmkn (approved):
LGTM!

